PR #4247, opened 2022-03-01
Fix Typo in Spark-writes sample code

singhpk234:
We should use `partitionedBy` rather than `partitionBy` as mentioned before the sample code.


hililiwei (approved):
good.

rajarshisarkar (approved):
LGTM.

jackye1995 (approved):
thanks for the fix!

PR #4248, opened 2022-03-01
AWS : use additionalLocations in glue table to store location info

singhpk234:
- The new additionalLocations field added in Glue [CodePointer](https://github.com/aws/aws-sdk-java-v2/blob/master/services/glue/src/main/resources/codegen-resources/service-2.json#L12226-L12229)
- store location info present in table properties in this param.


aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -70,6 +73,12 @@
       .orNoop()
       .build();
 
+  // Attempt to set additionalLocations if available on the given AWS SDK version
+  private static final DynMethods.UnboundMethod ADDITIONAL_LOCATIONS = DynMethods.builder("additionalLocations")
+      .hiddenImpl("software.amazon.awssdk.services.glue.model.UpdateTableRequest$Builder", List.class)
```
rajarshisarkar, line 78: `additionalLocations` should be in `software.amazon.awssdk.services.glue.model.StorageDescriptor`.

rajarshisarkar: In that case, we set the storage description only in `IcebergToGlueConverter` class.

rajarshisarkar: In that case, we set the `StorageDescriptor` only in `IcebergToGlueConverter` class.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -207,6 +216,15 @@ void persistGlueTable(Table glueTable, Map<String, String> parameters, TableMeta
         SET_VERSION_ID.invoke(updateTableRequest, glueTable.versionId());
       }
 
+      if (!ADDITIONAL_LOCATIONS.isNoop()) {
+        ADDITIONAL_LOCATIONS.invoke(updateTableRequest,
```
rajarshisarkar, line 220: Also, is `additionalLocations` available in the previous SDK version? We might not need reflection in that case.

singhpk234:

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
```diff
@@ -61,6 +65,12 @@ private IcebergToGlueConverter() {
   public static final String ICEBERG_FIELD_OPTIONAL = "iceberg.field.optional";
   public static final String ICEBERG_FIELD_CURRENT = "iceberg.field.current";
 
+  // Attempt to set additionalLocations if available on the given AWS SDK version
+  private static final DynMethods.UnboundMethod ADDITIONAL_LOCATIONS = DynMethods.builder("additionalLocations")
```
jackye1995, line 69: nit: prefer verb for setter, `SET_ADDITIONAL_LOCATIONS`

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
```diff
@@ -175,12 +185,25 @@ static void validateTableIdentifier(TableIdentifier tableIdentifier) {
...
-  static void setTableInputInformation(TableInput.Builder tableInputBuilder, TableMetadata metadata) {
+  static void setTableInputInformation(
+      TableInput.Builder tableInputBuilder,
+      Map<String, String> tableProps,
+      TableMetadata metadata) {
     try {
+      StorageDescriptor.Builder storageDescriptor = StorageDescriptor.builder();
+      if (!ADDITIONAL_LOCATIONS.isNoop()) {
+        ADDITIONAL_LOCATIONS.invoke(storageDescriptor.additionalLocations(),
+            ImmutableList.of(
```
jackye1995, line 199: this should be a set, also have we tested if it's fine to have `ImmutableList.of(...)` with null values? I think it's going to fail.

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
```diff
@@ -175,12 +185,25 @@ static void validateTableIdentifier(TableIdentifier tableIdentifier) {
...
+                tableProps.get(TableProperties.WRITE_DATA_LOCATION),
```
jackye1995, line 200: this is likely wrong, the property is not a part of Glue parameters, but `metadata.properties()`

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
singhpk234: +1
yup just checked ImmutableList.of() fails with null values... 
- added a test case as well where in the additionalLocation the properties we are fetching values from is not present. 

replaced this with location's with not null filtering, as it's not worth storing null as well.

```java
  private static final List<String> PROPERTIES_FOR_GLUE_ADDITIONAL_LOCATIONS = ImmutableList.of(
      TableProperties.WRITE_DATA_LOCATION,
      TableProperties.WRITE_METADATA_LOCATION,
      TableProperties.OBJECT_STORE_PATH,
      TableProperties.WRITE_FOLDER_STORAGE_LOCATION
  );

  private static List<String> getAdditionalLocations(TableMetadata tableMetadata) {
    return PROPERTIES_FOR_GLUE_ADDITIONAL_LOCATIONS.stream()
        .map(tableMetadata.properties()::get)
        .filter(Objects::nonNull)
        .collect(Collectors.toList());
  }
```

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
```diff
@@ -179,8 +196,12 @@ static void validateTableIdentifier(TableIdentifier tableIdentifier) {
    */
   static void setTableInputInformation(TableInput.Builder tableInputBuilder, TableMetadata metadata) {
     try {
+      StorageDescriptor.Builder storageDescriptor = StorageDescriptor.builder();
+      if (!SET_ADDITIONAL_LOCATIONS.isNoop()) {
+        SET_ADDITIONAL_LOCATIONS.invoke(storageDescriptor, getAdditionalLocations(metadata));
+      }
```
rajarshisarkar, line 202: nit: newline after if block

aws/src/test/java/org/apache/iceberg/aws/glue/TestIcebergToGlueConverter.java
```diff
@@ -40,6 +42,11 @@
 
 public class TestIcebergToGlueConverter {
 
+  private final Map<String, String> tableMetaDataProperties = ImmutableMap.of(
+      TableProperties.WRITE_DATA_LOCATION, "s3://writeDataLoc",
+      TableProperties.WRITE_METADATA_LOCATION, "s3://writeMetaDataLoc",
+      TableProperties.WRITE_FOLDER_STORAGE_LOCATION, "s3://writeFolderStorageLoc");
```
rajarshisarkar, line 48: Let's test a key will null values.

singhpk234: I think, null as value is handled here we only have 3 keys ...  rather than [4](https://github.com/apache/iceberg/pull/4248/files#diff-b97b3e8bad418d1e6cdee1bc8b35c5db6f5df45b8d714c8640b51532bc86ecfbR68) 
4th Key i.e `TableProperties.OBJECT_STORE_PATH` is not present in this hashMap hence we will get Null when we do a hashMap lookup and it will be filtered out and not sent in additionalLocations.

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
```diff
@@ -60,6 +65,18 @@ private IcebergToGlueConverter() {
   public static final String ICEBERG_FIELD_ID = "iceberg.field.id";
   public static final String ICEBERG_FIELD_OPTIONAL = "iceberg.field.optional";
   public static final String ICEBERG_FIELD_CURRENT = "iceberg.field.current";
+  private static final List<String> PROPERTIES_FOR_GLUE_ADDITIONAL_LOCATIONS = ImmutableList.of(
```
jackye1995, line 68: variable name a bit convoluted, what about `ADDITIONAL_LOCATION_PROPERTIES`

aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java
```diff
@@ -276,4 +298,11 @@ private static void addColumnWithDedupe(List<Column> columns, Set<String> dedupe
       dedupe.add(field.name());
     }
   }
+
+  private static Set<String> getAdditionalLocations(TableMetadata tableMetadata) {
```
jackye1995, line 302: I think this method is simple enough that we can just put it in the main method, no need to have a separated helper

aws/src/test/java/org/apache/iceberg/aws/glue/TestIcebergToGlueConverter.java
```diff
@@ -216,6 +229,10 @@ public void testSetTableInputInformationWithRemovedColumns() {
             .build())
         .build();
 
+    Assert.assertEquals(
+        "additionalLocations do not match",
```
jackye1995, line 233: the message here should describe the expected behavior, instead of the wrong behavior, like `"additionalLocations should match"`. It seems like we are also doing this wrong at L237 and L241, could you update those messages as well?

jackye1995:
thanks, mostly look good to me! There is also a test in integration tests related to this, could you update that as well?

singhpk234:
Thanks @jackye1995 !! 
- updated the integration Test for the same and verified it by running locally

---

ready for another pass

jackye1995 (approved):
looks good to me! Thanks for all the work!

rajarshisarkar (approved):
LGTM. Thanks, @singhpk234!

amogh-jahagirdar (approved):
Looks good to me, thanks for this!

aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogTable.java
```diff
@@ -74,6 +74,8 @@ public void testCreateTable() {
     Assert.assertTrue(response.table().parameters().containsKey(BaseMetastoreTableOperations.METADATA_LOCATION_PROP));
     Assert.assertEquals(schema.columns().size(), response.table().storageDescriptor().columns().size());
     Assert.assertEquals(partitionSpec.fields().size(), response.table().partitionKeys().size());
+    Assert.assertEquals("additionalLocations should match", tableLocationProperties.values(),
```
xiaoxuandev, line 77: Is this test passing? it fails on my side with:
Expected :[s3://xioxuan-test/writeDataLoc, s3://xioxuan-test/writeMetaDataLoc, s3://xioxuan-test/writeFolderStorageLoc]
Actual   :[s3://xioxuan-test/writeMetaDataLoc, s3://xioxuan-test/writeDataLoc, s3://xioxuan-test/writeFolderStorageLoc]

singhpk234: yes I checked running the test case against master right now it passes :
```shell
pk/iceberg - (master) > ./gradlew :iceberg-aws:integrationTest --tests "org.apache.iceberg.aws.glue.TestGlueCatalogTable"

BUILD SUCCESSFUL in 1m 44s
14 actionable tasks: 1 executed, 13 up-to-date

```

Looks like the order is of the output is messed up are you testing this in your branch ? if yes some change might have caused the diff in order of keys of HashMap.

xiaoxuandev: Yes. I am testing is against master branch. Let me double check again.

xiaoxuandev: it is still failing when I run clean build on master branch. I added this fix and it works. https://github.com/apache/iceberg/pull/4355

singhpk234: This is very strange, if possible can you please share the java version you are in. 

P.S thankyou for making the change to fix it, have added a feedback, do let me know what you think of it,

xiaoxuandev: openjdk version "1.8.0_322"

PR #4254, opened 2022-03-02
Core: support dynamic MetricsContext initialization

kunal0829:
With MetricsContext being recently added, it is important to be able to dynamically initialize what implementation of MetricsContext the user prefers. This PR adds that option with the `io.metrics-impl` property for future MetricsContext implementations. Moved `IO_METRICS_IMPL` and `DEFAULT_METRICS_IMPL` to `CatalogProperties`  to keep consistency with the repo.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
@@ -29,6 +29,21 @@ import java.util.Optional;
  * like thread safety and serialization.
  */
 public interface MetricsContext extends Serializable {
+  MetricsContext NULL_METRICS = new MetricsContext() {
+    @Override
+    public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit unit) {
+      return new Counter<T>() {
+        @Override
+        public void increment() {
+        }
+
+        @Override
+        public void increment(T amount) {
+        }
+      };
+    }
+  };
+
   enum Unit {
     UNDEFINED("undefined"),
     BYTES("bytes"),
@@ -94,19 +109,6 @@ public interface MetricsContext extends Serializable {
    * @return a non-recording metrics context
    */
   static MetricsContext nullMetrics() {
-    return new MetricsContext() {
-      @Override
-      public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit unit) {
-        return new Counter<T>() {
-          @Override
-          public void increment() {
-          }
-
-          @Override
-          public void increment(T amount) {
-          }
-        };
-      }
-    };
+    return NULL_METRICS;
   }
 }
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -29,6 +29,7 @@ import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.metrics.MetricsContext;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.util.SerializableSupplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -50,6 +51,8 @@ public class S3FileIO implements FileIO {
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
+
+  private static final String S3 = "s3";
 
   /**
    * No-arg constructor to load the FileIO dynamically.
@@ -119,10 +122,20 @@ public class S3FileIO implements FileIO {
     // Report Hadoop metrics if Hadoop is available
     String metricsImpl = properties.getOrDefault(
             CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
-    properties.put(CatalogProperties.IO_METRICS_SCHEME, properties.getOrDefault(
-            CatalogProperties.IO_METRICS_SCHEME, "s3"));
 
-    this.metrics = CatalogUtil.loadFileIOMetricsContext(metricsImpl, properties);
+    ImmutableMap<String, String> metricContextProperties;
+    if (properties.containsKey(CatalogProperties.IO_METRICS_SCHEME)) {
+      metricContextProperties = ImmutableMap.<String, String>builder()
+              .putAll(properties)
+              .build();
+    } else {
+      metricContextProperties = ImmutableMap.<String, String>builder()
+              .putAll(properties)
+              .put(CatalogProperties.IO_METRICS_SCHEME, S3)
+              .build();
+    }
+
+    this.metrics = CatalogUtil.loadFileIOMetricsContext(metricsImpl, metricContextProperties, null);
   }
 
   @Override
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -45,14 +45,13 @@ import software.amazon.awssdk.services.s3.model.DeleteObjectRequest;
  */
 public class S3FileIO implements FileIO {
   private static final Logger LOG = LoggerFactory.getLogger(S3FileIO.class);
+  private static final String IO_METRICS_SCHEME_S3 = "s3";
 
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
-
-  private static final String IO_METRICS_SCHEME_S3 = "s3";
 
   /**
    * No-arg constructor to load the FileIO dynamically.
@@ -123,7 +122,7 @@ public class S3FileIO implements FileIO {
     String metricsImpl = properties.getOrDefault(
             CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
 
-    ImmutableMap<String, String> metricContextProperties = ImmutableMap.copyOf(properties);
+    Map<String, String> metricContextProperties = properties;
     if (!properties.containsKey(CatalogProperties.IO_METRICS_SCHEME)) {
       metricContextProperties = ImmutableMap.<String, String>builder()
               .putAll(properties)
@@ -131,7 +130,8 @@ public class S3FileIO implements FileIO {
               .build();
     }
 
-    this.metrics = CatalogUtil.loadFileIOMetricsContext(metricsImpl, metricContextProperties, null);
+    this.metrics = CatalogUtil.loadFileIOMetricsContext(
+            metricsImpl, metricContextProperties, null /* no Hadoop config */);
   }
 
   @Override
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.aws.s3;
 
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.aws.AwsClientFactories;
 import org.apache.iceberg.aws.AwsProperties;
 import org.apache.iceberg.common.DynConstructors;
@@ -43,7 +44,6 @@ import software.amazon.awssdk.services.s3.model.DeleteObjectRequest;
  */
 public class S3FileIO implements FileIO {
   private static final Logger LOG = LoggerFactory.getLogger(S3FileIO.class);
-  private static final String DEFAULT_METRICS_IMPL = "org.apache.iceberg.hadoop.HadoopMetricsContext";
 
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
@@ -119,12 +119,16 @@ public class S3FileIO implements FileIO {
     // Report Hadoop metrics if Hadoop is available
     try {
       DynConstructors.Ctor<MetricsContext> ctor =
-          DynConstructors.builder(MetricsContext.class).hiddenImpl(DEFAULT_METRICS_IMPL, String.class).buildChecked();
+              DynConstructors.builder(MetricsContext.class)
+                      .impl(properties.getOrDefault(
+                              CatalogProperties.IO_METRICS_IMPL, CatalogProperties.DEFAULT_METRICS_IMPL))
+                      .buildChecked();
       this.metrics = ctor.newInstance("s3");
 
       metrics.initialize(properties);
     } catch (NoSuchMethodException | ClassCastException e) {
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
+      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics",
+              CatalogProperties.DEFAULT_METRICS_IMPL, e);
     }
   }
 
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -118,7 +118,6 @@ public class S3FileIO implements FileIO {
       this.s3 = AwsClientFactories.from(properties)::s3;
     }
 
-    // Report Hadoop metrics if Hadoop is available
     String metricsImpl = properties.getOrDefault(
             CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
 
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -117,18 +117,16 @@ public class S3FileIO implements FileIO {
     }
 
     // Report Hadoop metrics if Hadoop is available
+    String metricsImpl = properties.getOrDefault(
+            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.DEFAULT_METRICS_IMPL);
     try {
       DynConstructors.Ctor<MetricsContext> ctor =
-              DynConstructors.builder(MetricsContext.class)
-                      .impl(properties.getOrDefault(
-                              CatalogProperties.IO_METRICS_IMPL, CatalogProperties.DEFAULT_METRICS_IMPL))
-                      .buildChecked();
+              DynConstructors.builder(MetricsContext.class).impl(metricsImpl, String.class).buildChecked();
       this.metrics = ctor.newInstance("s3");
 
       metrics.initialize(properties);
     } catch (NoSuchMethodException | ClassCastException e) {
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics",
-              CatalogProperties.DEFAULT_METRICS_IMPL, e);
+      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", metricsImpl, e);
     }
   }
 
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -22,9 +22,9 @@ package org.apache.iceberg.aws.s3;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.CatalogUtil;
 import org.apache.iceberg.aws.AwsClientFactories;
 import org.apache.iceberg.aws.AwsProperties;
-import org.apache.iceberg.common.DynConstructors;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
@@ -118,16 +118,11 @@ public class S3FileIO implements FileIO {
 
     // Report Hadoop metrics if Hadoop is available
     String metricsImpl = properties.getOrDefault(
-            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.DEFAULT_METRICS_IMPL);
-    try {
-      DynConstructors.Ctor<MetricsContext> ctor =
-              DynConstructors.builder(MetricsContext.class).impl(metricsImpl, String.class).buildChecked();
-      this.metrics = ctor.newInstance("s3");
+            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
+    properties.put(CatalogProperties.IO_METRICS_SCHEME, properties.getOrDefault(
+            CatalogProperties.IO_METRICS_SCHEME, "s3"));
 
-      metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", metricsImpl, e);
-    }
+    this.metrics = CatalogUtil.loadFileIOMetricsContext(metricsImpl, properties);
   }
 
   @Override
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -52,7 +52,7 @@ public class S3FileIO implements FileIO {
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
 
-  private static final String S3 = "s3";
+  private static final String IO_METRICS_SCHEME_S3 = "s3";
 
   /**
    * No-arg constructor to load the FileIO dynamically.
@@ -123,15 +123,11 @@ public class S3FileIO implements FileIO {
     String metricsImpl = properties.getOrDefault(
             CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
 
-    ImmutableMap<String, String> metricContextProperties;
-    if (properties.containsKey(CatalogProperties.IO_METRICS_SCHEME)) {
+    ImmutableMap<String, String> metricContextProperties = ImmutableMap.copyOf(properties);
+    if (!properties.containsKey(CatalogProperties.IO_METRICS_SCHEME)) {
       metricContextProperties = ImmutableMap.<String, String>builder()
               .putAll(properties)
-              .build();
-    } else {
-      metricContextProperties = ImmutableMap.<String, String>builder()
-              .putAll(properties)
-              .put(CatalogProperties.IO_METRICS_SCHEME, S3)
+              .put(CatalogProperties.IO_METRICS_SCHEME, IO_METRICS_SCHEME_S3)
               .build();
     }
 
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -81,4 +81,8 @@ private CatalogProperties() {
   public static final String APP_ID = "app-id";
   public static final String USER = "user";
 
+  public static final String IO_METRICS_IMPL = "io-metrics-impl";
+  public static final String DEFAULT_METRICS_IMPL = "org.apache.iceberg.hadoop.HadoopMetricsContext";
```
jackye1995, line 85: nit: IO_METRICS_IMPL_DEFAULT

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,14 +117,16 @@ public void initialize(Map<String, String> properties) {
     }
 
     // Report Hadoop metrics if Hadoop is available
+    String metricsImpl = properties.getOrDefault(
```
jackye1995, line 120: can we move this chunk of initialization logic to `CatalogUtil` next to `loadFileIO`?

core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
```diff
@@ -49,7 +49,7 @@ public HadoopMetricsContext(String scheme) {
   public void initialize(Map<String, String> properties) {
     // FileIO has no specific implementation class, but Hadoop will
     // still track and report for the provided scheme.
-    this.scheme = properties.getOrDefault(SCHEME, scheme);
+    this.scheme = properties.getOrDefault(CatalogProperties.IO_METRICS_SCHEME, scheme);
```
jackye1995, line 52: I think we no longer need constructor `public HadoopMetricsContext(String scheme)`, we should use a no-arg constructor because we are already initializing the scheme here. The scheme should be passed in from S3FileIO.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,15 +117,12 @@ public void initialize(Map<String, String> properties) {
...
-          DynConstructors.builder(MetricsContext.class).hiddenImpl(DEFAULT_METRICS_IMPL, String.class).buildChecked();
-      this.metrics = ctor.newInstance("s3");
-
-      metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
-    }
+    String metricsImpl = properties.getOrDefault(
+            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
+    properties.put(CatalogProperties.IO_METRICS_SCHEME, properties.getOrDefault(
```
jackye1995, line 122: we cannot guarantee the properties map is mutable. we should create a new copy to put the metrics scheme

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,15 +117,12 @@ public void initialize(Map<String, String> properties) {
...
+            CatalogProperties.IO_METRICS_SCHEME, "s3"));
```
jackye1995, line 123: s3 should be a static variable

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -275,6 +276,25 @@ public static FileIO loadFileIO(
     return fileIO;
   }
 
+  public static MetricsContext loadFileIOMetricsContext(
+      String impl,
+      Map<String, String> properties) {
```
jackye1995, line 281: this should also have the `Object hadoopConf` in the parameter like `loadFileIO`, we will pass in `null` in S3FileIO but other implementations might leverage this.

core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
```diff
@@ -33,23 +34,18 @@
...
-    this.scheme = scheme;
+  public HadoopMetricsContext() {
   }
 
   @Override
   public void initialize(Map<String, String> properties) {
     // FileIO has no specific implementation class, but Hadoop will
     // still track and report for the provided scheme.
-    this.scheme = properties.getOrDefault(SCHEME, scheme);
+    this.scheme = properties.get(CatalogProperties.IO_METRICS_SCHEME);
```
jackye1995, line 48: assert not null

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,15 +120,22 @@ public void initialize(Map<String, String> properties) {
...
-          DynConstructors.builder(MetricsContext.class).hiddenImpl(DEFAULT_METRICS_IMPL, String.class).buildChecked();
-      this.metrics = ctor.newInstance("s3");
-
-      metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
+    String metricsImpl = properties.getOrDefault(
+            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
+
+    ImmutableMap<String, String> metricContextProperties;
```
jackye1995, line 126: should never use concrete class for map as the type.

Also can just use `Map<String, String> metricContextProperties = properties` and use a single if statement.

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -275,6 +276,26 @@ public static FileIO loadFileIO(
     return fileIO;
   }
 
+  public static MetricsContext loadFileIOMetricsContext(
+      String impl,
+      Map<String, String> properties,
+      Object hadoopConf) {
```
jackye1995, line 282: we need to check and initialize `hadoopConf` like in `loadFileIO` (L373)

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -43,14 +45,15 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(S3FileIO.class);
-  private static final String DEFAULT_METRICS_IMPL = "org.apache.iceberg.hadoop.HadoopMetricsContext";
 
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
 
+  private static final String S3 = "s3";
```
jackye1995, line 55: variable name should be more specific for the purpose of usage, like `IO_METRICS_SCHEME_DEFAULT`

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -275,6 +276,26 @@ public static FileIO loadFileIO(
     return fileIO;
   }
 
+  public static MetricsContext loadFileIOMetricsContext(
```
jackye1995, line 279: please add a test for this

core/src/test/java/org/apache/iceberg/TestCatalogUtil.java
```diff
@@ -402,4 +416,24 @@ public String getArg() {
...
+
+  public static class TestFileIOMetricsContext implements FileIOMetricsContext {
+    private String scheme;
+
+    public TestFileIOMetricsContext() {
+    }
+
+    @Override
+    public void initialize(Map<String, String> properties) {
+      // FileIO has no specific implementation class, but Hadoop will
```
jackye1995, line 428: nit: remove unnecessary stuffs for test usage, like L428, 429, 431, 432

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,15 +120,18 @@ public void initialize(Map<String, String> properties) {
...
+
+    ImmutableMap<String, String> metricContextProperties = ImmutableMap.copyOf(properties);
+    if (!properties.containsKey(CatalogProperties.IO_METRICS_SCHEME)) {
+      metricContextProperties = ImmutableMap.<String, String>builder()
+              .putAll(properties)
+              .put(CatalogProperties.IO_METRICS_SCHEME, IO_METRICS_SCHEME_S3)
+              .build();
     }
+
+    this.metrics = CatalogUtil.loadFileIOMetricsContext(metricsImpl, metricContextProperties, null);
```
jackye1995, line 134: nit: add an inline comment for null input:

```java
CatalogUtil.loadFileIOMetricsContext(metricsImpl, metricContextProperties, null /* no Hadoop config */);
```

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -43,14 +45,15 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(S3FileIO.class);
-  private static final String DEFAULT_METRICS_IMPL = "org.apache.iceberg.hadoop.HadoopMetricsContext";
 
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
 
+  private static final String IO_METRICS_SCHEME_S3 = "s3";
```
jackye1995, line 55: class variables should be placed before instance variables

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,15 +120,18 @@ public void initialize(Map<String, String> properties) {
...
-          DynConstructors.builder(MetricsContext.class).hiddenImpl(DEFAULT_METRICS_IMPL, String.class).buildChecked();
-      this.metrics = ctor.newInstance("s3");
-
-      metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
+    String metricsImpl = properties.getOrDefault(
+            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
+
+    ImmutableMap<String, String> metricContextProperties = ImmutableMap.copyOf(properties);
```
jackye1995, line 126: nit: no need to do copy of, use `Map<String, String>`

jackye1995:
@danielcweeks as we discussed in the community sync, this is the first step to add the metrics context support in other FileIOs, just move things to `CatalogProperties` to initialize the metrics context like other dynamically loaded classes, any comments?

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -275,6 +276,28 @@ public static FileIO loadFileIO(
...
+      return null;
+    }
+
+    configureHadoopConf(metricsContext, hadoopConf);
+
+    metricsContext.initialize(properties);
+    return metricsContext;
+  }
+
+
```
jackye1995, line 300: nit: extra newline

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -117,15 +119,19 @@ public void initialize(Map<String, String> properties) {
     }
 
     // Report Hadoop metrics if Hadoop is available
```
jackye1995, line 121: nit: this comment can be removed

jackye1995 (approved):
looks good to me!

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -275,6 +276,27 @@ public static FileIO loadFileIO(
...
+      Object hadoopConf) {
+    LOG.info("Loading custom MetricsContext implementation: {}", impl);
+    MetricsContext metricsContext;
+    try {
+      DynConstructors.Ctor<MetricsContext> ctor =
+              DynConstructors.builder(MetricsContext.class).impl(impl).buildChecked();
+      metricsContext = ctor.newInstance();
+    } catch (NoSuchMethodException | ClassCastException e) {
+      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", impl, e);
+      return null;
```
danielcweeks, line 291: We should return the null metrics impl not `null` here because engines without hadoop (e.g. Flink) will then NPE in the S3FileIO.

danielcweeks: Specifically `MetricsContext.nullMetrics()`

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -81,4 +81,8 @@ private CatalogProperties() {
   public static final String APP_ID = "app-id";
   public static final String USER = "user";
 
+  public static final String IO_METRICS_IMPL = "io.metrics-impl";
+  public static final String IO_METRICS_IMPL_DEFAULT = "org.apache.iceberg.hadoop.HadoopMetricsContext";
+  public static final String IO_METRICS_SCHEME = "io.metrics-scheme";
```
danielcweeks, line 86: `io.metrics-scheme` is hadoop specific since it's tied to a hadoop filesystem scheme in terms of reporting.  Do we really need to move it here?

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -116,16 +118,19 @@ public void initialize(Map<String, String> properties) {
...
-      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
+    Map<String, String> metricContextProperties = properties;
+    if (!properties.containsKey(CatalogProperties.IO_METRICS_SCHEME)) {
+      metricContextProperties = ImmutableMap.<String, String>builder()
+              .putAll(properties)
+              .put(CatalogProperties.IO_METRICS_SCHEME, IO_METRICS_SCHEME_S3)
+              .build();
     }
+
+    this.metrics = CatalogUtil.loadFileIOMetricsContext(
```
danielcweeks, line 132: this overrides the default `MetricsContext.nullMetrics()` that gets set in the variable initialization win will result in NPE (see other note in catalog util)

danielcweeks (changes requested):

core/src/main/java/org/apache/iceberg/CatalogUtil.java
jackye1995: yes good point I missed that, this should not return null

core/src/main/java/org/apache/iceberg/CatalogProperties.java
jackye1995: I think there still needs to be a way for users to input some sort of grouping concept in the metrics, for example users who use `S3FileIO` for `gs://` should expect metrics to land to some different metrics namespace from `s3://`. Scheme seems to be a good way to do that. For our CloudWatch metrics integration. we will match this to the CloudWatch namespace concept, where the metrics will by default be emitted to a namepace `iceberg/<scheme>` such as `iceberg/s3`

jackye1995: Maybe we can give a more generic name for this than `scheme`, something like `namespace`? @kunal0829 what is the terminology used in statsd and collectd for this?

kunal0829: Both statsd and prometheus it is called namespace. Collectd takes a combination of host and plugin to create a "namespace".

core/src/main/java/org/apache/iceberg/CatalogUtil.java
kunal0829: Thank you for that. I have fixed that

core/src/main/java/org/apache/iceberg/CatalogProperties.java
jackye1995: sounds like using a term `io.metrics.namespace` seems like a more generic solution? @danielcweeks

rajarshisarkar (approved):
Looks good to me. Thanks @kunal0829!

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -116,16 +118,19 @@ public void initialize(Map<String, String> properties) {
       this.s3 = AwsClientFactories.from(properties)::s3;
     }
 
-    // Report Hadoop metrics if Hadoop is available
-    try {
-      DynConstructors.Ctor<MetricsContext> ctor =
-          DynConstructors.builder(MetricsContext.class).hiddenImpl(DEFAULT_METRICS_IMPL, String.class).buildChecked();
-      this.metrics = ctor.newInstance("s3");
+    String metricsImpl = properties.getOrDefault(
+            CatalogProperties.IO_METRICS_IMPL, CatalogProperties.IO_METRICS_IMPL_DEFAULT);
```
rdblue, line 122: Looks like indentation is off in this PR. Can you fix it?

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
```diff
@@ -29,6 +29,21 @@
  * like thread safety and serialization.
  */
 public interface MetricsContext extends Serializable {
+  MetricsContext NULL_METRICS = new MetricsContext() {
```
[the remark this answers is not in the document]
rdblue: Then we should move it outside of the interface? No need to have both NULL_METRICS and `nullMetrics()`.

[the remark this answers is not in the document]
jackye1995: Don't really have a good place to put this... what about this?

```java
  /**
   * Utility method for producing no metrics.
   *
   * @return a non-recording metrics context
   */
  static MetricsContext nullMetrics() {
    return NullMetricsContext.INSTANCE;
  }

  class NullMetricsContext implements MetricsContext {

    private static final MetricsContext INSTANCE = new NullMetricsContext();

    @Override
    public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit unit) {
      return new Counter<T>() {
        @Override
        public void increment() {
        }

        @Override
        public void increment(T amount) {
        }
      };
    }
  }
```

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -275,6 +276,27 @@ public static FileIO loadFileIO(
...
+      Object hadoopConf) {
+    LOG.info("Loading custom MetricsContext implementation: {}", impl);
+    MetricsContext metricsContext;
+    try {
+      DynConstructors.Ctor<MetricsContext> ctor =
+              DynConstructors.builder(MetricsContext.class).impl(impl).buildChecked();
+      metricsContext = ctor.newInstance();
+    } catch (NoSuchMethodException | ClassCastException e) {
+      LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", impl, e);
+      metricsContext = MetricsContext.nullMetrics();
```
jackye1995, line 291: nit: can directly return null metrics

core/src/test/java/org/apache/iceberg/TestCatalogUtil.java
```diff
@@ -158,6 +160,26 @@ public void loadCustomFileIO_badClass() {
...
+    Assertions.assertThat(metricsContext).isInstanceOf(TestFileIOMetricsContext.class);
+    Assert.assertEquals("testScheme", testMetricsContext.scheme());
+  }
+
+  @Test public void loadBadFileIOMetricsContext() {
+    Map<String, String> properties = Maps.newHashMap();
+    MetricsContext metricsContext = CatalogUtil.loadFileIOMetricsContext(
+            "BadMetricsContext", properties, null);
+
+    MetricsContext nullMetricsContext = MetricsContext.nullMetrics();
```
jackye1995, line 179: nit: can directly do `Assert.assertEquals(MetricsContext.nullMetrics(), metricsContext);`

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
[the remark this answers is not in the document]
rdblue: Yeah, exactly. I think that should live in a separate file, though.

What was wrong with using the anonymous class?

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -81,4 +81,8 @@ private CatalogProperties() {
   public static final String APP_ID = "app-id";
   public static final String USER = "user";
 
+  public static final String IO_METRICS_IMPL = "io.metrics-impl";
+  public static final String IO_METRICS_IMPL_DEFAULT = "org.apache.iceberg.hadoop.HadoopMetricsContext";
+  public static final String IO_METRICS_NAMESPACE = "io.metrics.namespace";
```
rdblue, line 86: I find this name confusing because I don't think of a metrics namespace. This is specific to Hadoop, right?

core/src/test/java/org/apache/iceberg/TestCatalogUtil.java
```diff
@@ -158,6 +160,26 @@ public void loadCustomFileIO_badClass() {
...
+    properties.put(CatalogProperties.IO_METRICS_NAMESPACE, "testScheme");
+    MetricsContext metricsContext = CatalogUtil.loadFileIOMetricsContext(
+            TestFileIOMetricsContext.class.getName(), properties, null);
+    TestFileIOMetricsContext testMetricsContext = (TestFileIOMetricsContext) metricsContext;
+
+    Assertions.assertThat(metricsContext).isInstanceOf(TestFileIOMetricsContext.class);
+    Assert.assertEquals("testScheme", testMetricsContext.scheme());
+  }
+
+  @Test public void loadBadFileIOMetricsContext() {
```
rdblue, line 174: Annotations should be on the previous line.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rdblue: I think the best solution here is to not define a public property for "namespace". I think the IO classes should just pass a "scheme" option into the metrics class. No need to define a contract or to try to make this generic.

rdblue:
@danielcweeks and @jackye1995, are we sure that dynamic loading is appropriate here?

I don't think that this fits with where we've used dynamic loading in the past. Dynamic loading is used where we want to be able to customize the implementation from the catalog. For example, a catalog may dynamically choose GCS or S3 or another FileIO implementation based on the catalog settings and table details (like root location). But the metrics reporting that this affects are being passed back to a processing engine, which is determined by the environment.

What we've been discussing for Flink is to have some way to connect FileIO metrics to Flink jobs and tasks, which likely means passing in a MetricsContext from Flink. This conflicts with the idea of passing down configuration that defines it. It doesn't make sense to me that we might allow a catalog or table to override something that is engine-specific.

One thing that Iceberg has been fairly successful at so far is trying to put configuration in the right place. I think this is really similar: we should not use table or catalog configuration for engine concerns.

jackye1995:
> But the metrics reporting that this affects are being passed back to a processing engine, which is determined by the environment.

Yes I agree this is more related to the processing engine than catalog, rather than the catalog. The question here becomes how do we support integration to different engines without dynamically loading? We do not really have a way of configuring FileIO outside catalog.

The issue with HadoopMetricsContext is that we are going back to be tied to Hadoop. We have seen users trying to use Iceberg environments without Hadoop dependency (e.g. AWS Kinesis serverless Flink) and people don't want to load Hadoop dependencies just for getting metrics. So I think we definitely need some other implementations.

Our plan with this PR is to later introduce metrics context such as `CollectdMetricsContext`, and that could be then tied to any monitoring services that typically has an agent process running to collect metrics at background. Our target is AWS CloudWatch integration, but I think most products offer similar integration pattern.

rdblue:
What I was thinking is a mix-in interface for `FileIO` that allows the engine to set the metrics context. We could default to Hadoop since that's what happens with `HadoopFileIO` automatically, but if you want to get passed metrics directly then you can set the context:

```java
  FileIO io = table.io();
  if (io instanceof MetricsProducer) {
    ((MetricsProducer) io).setMetricsContext(engineContext);
  }
```

What do you think, @jackye1995?

jackye1995:
> I was thinking is a mix-in interface

Yes that could work!

> wouldn't you integrate with the engine rather than with low-level libraries

Correct, however the current `HadoopMetricsContext` seems to be doing this across all the engines, and this offers a unique benefit for `HadoopFileIO` users that they are able to see their FileIO metrics across all compute platforms aggregate through the Hadoop FileSystem interface. 

There is benefit in at least providing an alternative such as `CollectdMetricsContext`, because we have been advocating `S3FileIO` to be not Hadoop dependent, and it is awkward that we now tell people we have to bring back the dependency and use Hadoop FileSystem to report metrics. This also applies to other non-Hadoop FileIOs.

Because of that, I think it still makes sense to have some sort of default behavior controlled through catalog level property, maybe in the name of `io.metrics.default.type` which is limited to a few specific types including `hadoop` and `collectd`, and `io.metrics.default.namespace` provides the namespace of the metrics.

(regarding metrics namespace, it was discussed in https://github.com/apache/iceberg/pull/4254#discussion_r819178970, it seems to be the terminology that most metrics system use, let me know what you think)

rdblue:
> (regarding metrics namespace, it was discussed in https://github.com/apache/iceberg/pull/4254#discussion_r819178970, it seems to be the terminology that most metrics system use, let me know what you think)

I think the problem is that we're standardizing something that doesn't need to be standard. This is specific to Hadoop metrics collection. I don't think we are ready to add a feature like prefixing metric names more generally.

Since a lot of this is unknown, I would lean toward not exposing things publicly or building a lot of expectations right now.

rdblue:
> Correct, however the current HadoopMetricsContext seems to be doing this across all the engines, and this offers a unique benefit for HadoopFileIO users that they are able to see their FileIO metrics across all compute platforms aggregate through the Hadoop FileSystem interface.

This is just a convenient way to get metrics to engines that already use Hadoop. I think we all agree that we want to be able to extend this, but it is a reasonable default for now.

>  think it still makes sense to have some sort of default behavior controlled through catalog level property, maybe in the name of io.metrics.default.type which is limited to a few specific types including hadoop and collectd, and io.metrics.default.namespace provides the namespace of the metrics.

I don't think we are at a point where there's a good argument for catalog-level configuration. Catalog configuration would be in conflict with what engines need and we would have to find a more complicated way to avoid this problem.

jackye1995:
> This is just a convenient way to get metrics to engines that already use Hadoop. I think we all agree that we want to be able to extend this, but it is a reasonable default for now.

Yes, in that case let's first try to add the `MetricsProducer` mixin interface. I agree the catalog-based solution is more like a workaround, let's just implement the right thing.

In this model, how would we eventually make this not just a S3FileIO specific implementation?

I am thinking about something like the following in SparkCatalog

```java
private Pair<Table, Long> load(Identifier ident) {
  ...
  Table table;
  try {
    table = icebergCatalog.loadTable(namespaceAsIdent);
    FileIO io = table.io();
    if (io instanceof MetricsProducer) {
        // Report Hadoop metrics by default, will add logic and option for SparkMetricsContext in the future when that is available
        try {
          // how would scheme be set for ResolvingFileIO? should we take table.location() as input to constructor to derive the scheme?
          MetricsContext metrics = ctor.newInstance(io.scheme());
        } catch (NoSuchMethodException | ClassCastException e) {
        }
      ((MetricsProducer) io).setMetricsContext(metrics);
    }
  } catch (Exception ignored) {
    // the namespace does not identify a table, so it cannot be a table with a snapshot selector
    // throw the original exception
    throw e;
  }
  ...
}
```

Would this make sense to you?

rdblue:
I think the `MetricsProducer` interface that carries `setMetricsContext` makes sense. We may not even need anything in `Catalog`. I think it is likely that the Flink implementation would create a `MetricsContext` in each task and set it on the IO when setting up execution. The dynamic Hadoop handling is a good default, but we probably only need to expose a way to set this.

rdblue:
It sounds like we agree about not setting up metrics with dynamic config, so I'm going to go ahead and close this one. Thanks for working on it, @kunal0829!

PR #4256, opened 2022-03-03
Docs: Fix typo in Java API Quickstart

wypoon:
Sample code declares a variable as `Catalog` instead of `HiveCatalog` but calls a method that is not in `Catalog`, only in `HiveCatalog`.

### DIFF
--- a/docs/versioned/api/java-api-quickstart.md
+++ b/docs/versioned/api/java-api-quickstart.md
@@ -39,7 +39,7 @@ You can initialize a Hive catalog with a name and some properties.
 ```java
 import org.apache.iceberg.hive.HiveCatalog;
 
-Catalog catalog = new HiveCatalog();
+HiveCatalog catalog = new HiveCatalog();
 catalog.setConf(spark.sparkContext().hadoopConfiguration());  // Configure using Spark's Hadoop configuration
 
 Map <String, String> properties = new HashMap<String, String>();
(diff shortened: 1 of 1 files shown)


docs/versioned/api/java-api-quickstart.md
```diff
@@ -39,7 +39,7 @@ You can initialize a Hive catalog with a name and some properties.
 ```java
 import org.apache.iceberg.hive.HiveCatalog;
```
wypoon, line 40: Note: We imported `HiveCatalog` here, not `Catalog`.

docs/versioned/api/java-api-quickstart.md
```diff
@@ -39,7 +39,7 @@ You can initialize a Hive catalog with a name and some properties.
...
 
-Catalog catalog = new HiveCatalog();
+HiveCatalog catalog = new HiveCatalog();
 catalog.setConf(spark.sparkContext().hadoopConfiguration());  // Configure using Spark's Hadoop configuration
```
wypoon, line 43: `setConf` is not a method of `Catalog`, it is a method of `org.apache.hadoop.conf.Configurable`, which `HiveCatalog` also implements.

openinx (approved):
LGTM, Thanks @wypoon for the contribution, and thanks @kbendick for the checking.

PR #4011, opened 2022-01-30
Allow table defaults to be configured and/ or enforced at catalog level using catalog properties.

SinghAsDev:
Allow table defaults to be configured and/ or enforced at catalog level using catalog properties. This resolves https://github.com/apache/iceberg/issues/3994

PR #4259, opened 2022-03-03
AWS: Add S3 Tagging

rajarshisarkar:
This change adds [S3 Tags](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html) to the objects while writing using `S3FileIO`. Users can pass their custom tags as part of the catalog properties.

Spark SQL launch command:
```
    --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key2=my_val2
```

Tags added in S3:

```
aws s3api get-object-tagging --bucket iceberg-warehouse --key s3-tagging/data/00000-0-593dfa9d-872f-4683-986f-ea220b7c875d-00001.parquet
{
    "TagSet": [
        {
            "Key": "my_key2",
            "Value": "my_val2"
        },
        {
            "Key": "my_key",
            "Value": "my_val"
        }
    ]
}
```
----

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
@@ -218,9 +218,10 @@ public class AwsProperties implements Serializable {
 
   /**
    * Used by {@link S3FileIO} to tag objects when writing. To set, we can pass a catalog property.
-   * Example in Spark: --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key=my_val
    * <p>
    * For more details, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   * <p>
+   * Example in Spark: --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key=my_val
    */
   public static final String S3_WRITE_TAGS_PREFIX = "s3.write.tags.";
 
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
@@ -22,6 +22,7 @@ package org.apache.iceberg.aws;
 import java.io.Serializable;
 import java.util.Map;
 import org.apache.iceberg.aws.dynamodb.DynamoDbCatalog;
+import org.apache.iceberg.aws.s3.S3FileIO;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.util.PropertyUtil;
 import software.amazon.awssdk.services.s3.model.ObjectCannedACL;
@@ -214,6 +215,14 @@ public class AwsProperties implements Serializable {
    * For more details, see https://docs.aws.amazon.com/general/latest/gr/rande.html
    */
   public static final String CLIENT_ASSUME_ROLE_REGION = "client.assume-role.region";
+
+  /**
+   * Used by {@link S3FileIO} to tag objects when writing. To set, we can pass a catalog property.
+   * Example in Spark: --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key=my_val
+   * <p>
+   * For more details, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   */
+  public static final String S3_WRITE_TAGS_PREFIX = "s3.write.tags.";
 
   /**
    * @deprecated will be removed at 0.15.0, please use {@link #S3_CHECKSUM_ENABLED_DEFAULT} instead
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/BaseS3File.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/BaseS3File.java
@@ -19,13 +19,16 @@
 
 package org.apache.iceberg.aws.s3;
 
+import java.util.Set;
 import org.apache.iceberg.aws.AwsProperties;
 import org.apache.iceberg.metrics.MetricsContext;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import software.amazon.awssdk.http.HttpStatusCode;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
 import software.amazon.awssdk.services.s3.model.HeadObjectResponse;
 import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.model.Tag;
 
 abstract class BaseS3File {
   private final S3Client client;
@@ -33,12 +36,23 @@ abstract class BaseS3File {
   private final AwsProperties awsProperties;
   private HeadObjectResponse metadata;
   private final MetricsContext metrics;
+  private final Set<Tag> writeTags;
 
   BaseS3File(S3Client client, S3URI uri, AwsProperties awsProperties, MetricsContext metrics) {
     this.client = client;
     this.uri = uri;
     this.awsProperties = awsProperties;
     this.metrics = metrics;
+    this.writeTags = Sets.newHashSet();
+  }
+
+  BaseS3File(S3Client client, S3URI uri, AwsProperties awsProperties, MetricsContext metrics,
+      Set<Tag> writeTags) {
+    this.client = client;
+    this.uri = uri;
+    this.awsProperties = awsProperties;
+    this.metrics = metrics;
+    this.writeTags = writeTags;
   }
 
   public String location() {
@@ -59,6 +73,10 @@ abstract class BaseS3File {
 
   protected MetricsContext metrics() {
     return metrics;
+  }
+
+  public Set<Tag> writeTags() {
+    return writeTags;
   }
 
   /**
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -115,8 +115,7 @@ public class S3FileIO implements FileIO {
   @Override
   public void initialize(Map<String, String> properties) {
     this.awsProperties = new AwsProperties(properties);
-    this.writeTags = toTags(
-        PropertyUtil.propertiesWithPrefix(properties, AwsProperties.S3_WRITE_TAGS_PREFIX));
+    this.writeTags = toTags(properties);
 
     // Do not override s3 client if it was provided
     if (s3 == null) {
@@ -145,8 +144,9 @@ public class S3FileIO implements FileIO {
     }
   }
 
-  private Set<Tag> toTags(Map<String, String> tagKeyToTagValue) {
-    return tagKeyToTagValue.entrySet().stream()
+  private Set<Tag> toTags(Map<String, String> properties) {
+    return PropertyUtil.propertiesWithPrefix(properties, AwsProperties.S3_WRITE_TAGS_PREFIX)
+        .entrySet().stream()
         .map(e -> Tag.builder().key(e.getKey()).value(e.getValue()).build())
         .collect(Collectors.toSet());
   }
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -20,7 +20,9 @@
 package org.apache.iceberg.aws.s3;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import org.apache.iceberg.aws.AwsClientFactories;
 import org.apache.iceberg.aws.AwsProperties;
 import org.apache.iceberg.common.DynConstructors;
@@ -28,11 +30,13 @@ import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.metrics.MetricsContext;
+import org.apache.iceberg.util.PropertyUtil;
 import org.apache.iceberg.util.SerializableSupplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.DeleteObjectRequest;
+import software.amazon.awssdk.services.s3.model.Tag;
 
 /**
  * FileIO implementation backed by S3.
@@ -50,6 +54,7 @@ public class S3FileIO implements FileIO {
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
+  private Set<Tag> writeTags;
 
   /**
    * No-arg constructor to load the FileIO dynamically.
@@ -88,7 +93,7 @@ public class S3FileIO implements FileIO {
 
   @Override
   public OutputFile newOutputFile(String path) {
-    return S3OutputFile.fromLocation(path, client(), awsProperties, metrics);
+    return S3OutputFile.fromLocation(path, client(), awsProperties, metrics, writeTags);
   }
 
   @Override
@@ -110,6 +115,8 @@ public class S3FileIO implements FileIO {
   @Override
   public void initialize(Map<String, String> properties) {
     this.awsProperties = new AwsProperties(properties);
+    this.writeTags = toTags(
+        PropertyUtil.propertiesWithPrefix(properties, AwsProperties.S3_WRITE_TAGS_PREFIX));
 
     // Do not override s3 client if it was provided
     if (s3 == null) {
@@ -136,5 +143,11 @@ public class S3FileIO implements FileIO {
         client.close();
       }
     }
+  }
+
+  private Set<Tag> toTags(Map<String, String> tagKeyToTagValue) {
+    return tagKeyToTagValue.entrySet().stream()
+        .map(e -> Tag.builder().key(e.getKey()).value(e.getValue()).build())
+        .collect(Collectors.toSet());
   }
 }
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
@@ -28,10 +28,16 @@ import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.io.PositionOutputStream;
 import org.apache.iceberg.metrics.MetricsContext;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.Tag;
 
 public class S3OutputFile extends BaseS3File implements OutputFile {
+  public static S3OutputFile fromLocation(String location, S3Client client, AwsProperties awsProperties,
+      MetricsContext metrics) {
+    return new S3OutputFile(client, new S3URI(location), awsProperties, metrics, Sets.newHashSet());
+  }
+
   public static S3OutputFile fromLocation(String location, S3Client client, AwsProperties awsProperties,
       MetricsContext metrics, Set<Tag> writeTags) {
     return new S3OutputFile(client, new S3URI(location), awsProperties, metrics, writeTags);
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.aws.s3;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.Set;
 import org.apache.iceberg.aws.AwsProperties;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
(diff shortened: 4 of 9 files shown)


aws/src/main/java/org/apache/iceberg/aws/s3/S3Tag.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.iceberg.aws.s3;
+
+import java.util.Collection;
+import java.util.Map;
+import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import software.amazon.awssdk.services.s3.model.Tag;
+
+public class S3Tag {
```
jackye1995, line 29: the logic here can just be moved to Glue catalog, I don' t think it's worth adding a new class for it

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -50,6 +52,7 @@
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
+  private Collection<Tag> s3Tags;
```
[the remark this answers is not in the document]
jackye1995: things in S3FileIO must be serializable with stable serial ID, can you check if this is? If not, prefer to use simple data structure like `Map<String, String>` for this

aws/src/main/java/org/apache/iceberg/aws/s3/S3Tag.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public S3Tag() {
+  }
+
+  public boolean isS3FileIOConfigured(Map<String, String> catalogProperties) {
+    String fileIOImpl = catalogProperties.get(CatalogProperties.FILE_IO_IMPL);
+    return fileIOImpl == null || fileIOImpl.equals(S3_FILE_IO_IMPL);
+  }
+
+  public Collection<Tag> toTags(TableIdentifier tableIdentifier) {
```
jackye1995, line 41: I think we can hold on this feature later, let's first add `s3.tags.<key>=<val>` as a generic way to specify s3 tags, and in a subsequent PR we can add `s3.table-tags-enabled`

rajarshisarkar: Sounds good.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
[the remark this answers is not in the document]
rajarshisarkar: Changed it to `List<Tag>`.

aws/src/main/java/org/apache/iceberg/aws/s3/S3Tag.java
rajarshisarkar: I have made the changes.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -162,7 +173,28 @@ private String cleanWarehousePath(String path) {
 
   @Override
   protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
-    return new GlueTableOperations(glue, lockManager, catalogName, awsProperties, fileIO, tableIdentifier);
+    if (isS3FileIOConfigured()) {
+      // Create different S3FileIO to use different credentials in future
+      FileIO s3FileIO = new S3FileIO(toTags(tableIdentifier));
+      s3FileIO.initialize(this.catalogProperties);
```
amogh-jahagirdar, line 179: Nit: I think the practice in ICeberg is to only use this when actually mutating the field, not needed when reading.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -73,11 +74,17 @@
...
 import software.amazon.awssdk.services.glue.model.TableInput;
 import software.amazon.awssdk.services.glue.model.UpdateDatabaseRequest;
+import software.amazon.awssdk.services.s3.model.Tag;
 
 public class GlueCatalog extends BaseMetastoreCatalog
     implements Closeable, SupportsNamespaces, Configurable<Configuration> {
 
   private static final Logger LOG = LoggerFactory.getLogger(GlueCatalog.class);
+  private static final String S3_FILE_IO_IMPL = "org.apache.iceberg.aws.s3.S3FileIO";
+  private static final String S3_TAGS_FORMAT = "s3.tags.format";
```
jackye1995, line 84: Let's first provide a generic solution, before adding specific tag keys for users. I think it's better if in `AwsProperties` we introduce the property prefix `s3.write.tags`, so that users can define any free-form prefixes, for example:

```
spark-sql --packages $DEPENDENCIES \
    --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCatalog \
    --conf spark.sql.catalog.my_catalog.warehouse=s3://my-bucket/my/key/prefix \
    --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key2=my_val2
```

This can be passed to `GlueCatalog.initialize`, and we can extract the tags `my_key->my_val` and `my_key2->my_val2`, and stored in S3FileIO to be used during a write.

`write` in `s3.write.tags` is useful in case we would like to introduce other tagging, such as `s3.delete.tags`.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -73,11 +74,17 @@
...
+  private static final String S3_TAGS_NAMESPACE = "s3.tags.namespace";
+  private static final String S3_TAGS_TABLE = "s3.tags.table";
```
jackye1995, line 86: also adding tags for table and namespace names might not be ideal, because people can rename table and these information could change. With the generic tagging solution described above, we can in the future add tags using some stable Iceberg table properties, such as the Iceberg table UUID.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -162,7 +173,28 @@ private String cleanWarehousePath(String path) {
...
+      FileIO s3FileIO = new S3FileIO(toTags(tableIdentifier));
+      s3FileIO.initialize(this.catalogProperties);
+      closeableGroup.addCloseable(s3FileIO);
+      return new GlueTableOperations(glue, lockManager, catalogName, awsProperties, s3FileIO, tableIdentifier);
+    } else {
+      return new GlueTableOperations(glue, lockManager, catalogName, awsProperties, fileIO, tableIdentifier);
+    }
+  }
+
+  private boolean isS3FileIOConfigured() {
```
jackye1995, line 187: having a specific case for S3FileIO should not be needed in a generic solution. `GlueCatalog` can just pass in the properties list to S3FileIO, and S3FileIO tries to parse any properties begin with `s3.write.tags` to gather all the tags needed.

In the future, we might want to initialize a different S3FileIO for each `newTableOps` to have table-specific tags, but I think that's an enhancement we can do later, it would be much cleaner if we first offer the generic solution.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: Thanks, I have made the changes.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: This makes sense, I have made the changes to allow users to define any free-form prefixes.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: Agreed, I have made the changes.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: Sure, I have made the changes accordingly.

rajarshisarkar:
I took the idea of parsing the properties beginning with a prefix from https://github.com/apache/iceberg/pull/4011

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -215,6 +216,11 @@
    */
   public static final String CLIENT_ASSUME_ROLE_REGION = "client.assume-role.region";
 
+  /**
+   * S3 Tag prefix used by {@link S3FileIO}
```
jackye1995, line 220: can be more descriptive, we need to at least say this is used to tag objects when writing. Also would be great to provide:
1. an example of how to set write tags using prefix
2. a link to S3 tagging documentation

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -124,8 +125,8 @@ private LockManager initializeLockManager(Map<String, String> properties) {
 
   private FileIO initializeFileIO(Map<String, String> properties) {
     String fileIOImpl = properties.get(CatalogProperties.FILE_IO_IMPL);
-    if (fileIOImpl == null) {
-      FileIO io = new S3FileIO();
+    if (fileIOImpl == null || fileIOImpl.equals(S3_FILE_IO_IMPL)) {
+      FileIO io = new S3FileIO(properties);
```
jackye1995, line 129: I don't think you need to do this, properties will be passed in during `io.initialize(properties);` at next line anyway

aws/src/main/java/org/apache/iceberg/aws/s3/BaseS3File.java
```diff
@@ -19,20 +19,23 @@
...
 import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.model.Tag;
 
 abstract class BaseS3File {
   private final S3Client client;
   private final S3URI uri;
   private final AwsProperties awsProperties;
   private HeadObjectResponse metadata;
   private final MetricsContext metrics;
+  private List<Tag> tags;
```
jackye1995, line 38: private final, also should be named `writeTags`

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -81,14 +86,38 @@ public S3FileIO(SerializableSupplier<S3Client> s3, AwsProperties awsProperties)
...
   }
 
+  /**
+   * Constructor with custom s3 supplier and s3 tags
+   * <p>
+   * Calling {@link S3FileIO#initialize(Map)} will overwrite information set in this constructor.
+   * param s3 s3 supplier
+   * @param properties catalog properties
+   */
+  public S3FileIO(SerializableSupplier<S3Client> s3, Map<String, String> properties) {
```
jackye1995, line 96: the logic should be moved to `initialize`

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -50,6 +54,7 @@
   private transient S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
+  private List<Tag> tags;
```
jackye1995, line 57: should this be a set instead? Could you check if the object hashcode is implemented?

jackye1995: also similar comment, should be named `writeTags`

jackye1995 (commented):
logic mostly looks correct to me, added some comments

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
rajarshisarkar: Thanks, I have made the changes.

rajarshisarkar: Thanks, I have made the changes.

aws/src/main/java/org/apache/iceberg/aws/s3/BaseS3File.java
rajarshisarkar: I have renamed to `writeTags`.

rajarshisarkar: I have renamed to `writeTags`.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
rajarshisarkar: Used a `Set`. Yes, object hashcode is implemented. Also, I have renamed to `writeTags`.

rajarshisarkar: Used a `Set`. Yes, object hashcode is implemented. Also, I have renamed to `writeTags`.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -162,7 +173,28 @@ private String cleanWarehousePath(String path) {
 
   @Override
   protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
-    return new GlueTableOperations(glue, lockManager, catalogName, awsProperties, fileIO, tableIdentifier);
+    if (isS3FileIOConfigured()) {
+      // Create different S3FileIO to use different credentials in future
```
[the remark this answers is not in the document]
rajarshisarkar: We have moved away from this approach in the current implementation.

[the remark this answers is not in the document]
rajarshisarkar: We have moved away from this approach in the current implementation.

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3OutputStream.java
```diff
@@ -85,6 +87,8 @@
   private final Random random = new Random(1);
   private final Path tmpDir = Files.createTempDirectory("s3fileio-test-");
   private final String newTmpDirectory = "/tmp/newStagingDirectory";
+  private final List<Tag> tags = ImmutableList.of(
+      Tag.builder().key("abc").value("123").build());
```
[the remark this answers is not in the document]
rajarshisarkar: Request will be rejected in case we have duplicate keys: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html

Though, we are not supposed to have duplicate keys as the catalog `properties` map maintains uniqueness. So, if the user passes: 
```
--conf spark.sql.catalog.my_catalog.s3.write.tags.my_key=my_val2
```

then the `properties` map would store only `s3.write.tags.my_key=my_val2` which is later parsed to extract `my_key=my_val2` and then converted to `writeTags`.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
rajarshisarkar: Thanks, I have made the changes.

singhpk234 (approved):
LGTM, Thanks @rajarshisarkar !!! for all the work.

core/src/main/java/org/apache/iceberg/util/PropertyUtil.java
```diff
@@ -70,4 +73,26 @@ public static String propertyAsString(Map<String, String> properties,
...
+   * @param prefix prefix to choose keys from input map
+   * @return subset of input map with keys starting with provided prefix and prefix trimmed out
+   */
+  public static Map<String, String> propertiesWithPrefix(Map<String, String> properties,
+      String prefix) {
+    if (properties == null || properties.isEmpty()) {
+      return ImmutableMap.of();
+    }
+
+    Preconditions.checkState(prefix != null, "prefix can't be null.");
```
jackye1995, line 91: nit: prefer to use `checkArgument`

core/src/main/java/org/apache/iceberg/util/PropertyUtil.java
```diff
@@ -70,4 +73,26 @@ public static String propertyAsString(Map<String, String> properties,
...
+
+  /**
+   * Returns subset of provided map with keys matching the provided prefix. Matching is case-sensitive and the matching
+   * prefix is removed from the keys in returned map.
+   *
+   * @param properties input map
+   * @param prefix prefix to choose keys from input map
+   * @return subset of input map with keys starting with provided prefix and prefix trimmed out
+   */
+  public static Map<String, String> propertiesWithPrefix(Map<String, String> properties,
```
jackye1995, line 85: prefer method arguments all on next lines of it has to span multiple lines, e.g. 

```java
propertiesWithPrefix(
  Map<String, String> properties,
  String prefix) {
```

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3OutputStream.java
```diff
@@ -145,7 +153,8 @@ public void testMultipleClose() throws IOException {
   public void testStagingDirectoryCreation() throws IOException {
     AwsProperties newStagingDirectoryAwsProperties = new AwsProperties(ImmutableMap.of(
         AwsProperties.S3FILEIO_STAGING_DIRECTORY, newTmpDirectory));
-    S3OutputStream stream = new S3OutputStream(s3, randomURI(), newStagingDirectoryAwsProperties, nullMetrics());
+    S3OutputStream stream = new S3OutputStream(s3, randomURI(), newStagingDirectoryAwsProperties,
```
jackye1995, line 156: nit: prefer method arguments all on next lines of it has to span multiple lines

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3OutputStream.java
```diff
@@ -116,7 +122,8 @@ public void testWrite() {
   public void testAbortAfterFailedPartUpload() {
     doThrow(new RuntimeException()).when(s3mock).uploadPart((UploadPartRequest) any(), (RequestBody) any());
 
-    try (S3OutputStream stream = new S3OutputStream(s3mock, randomURI(), properties, nullMetrics())) {
+    try (S3OutputStream stream = new S3OutputStream(s3mock, randomURI(), properties, nullMetrics(),
```
jackye1995, line 125: nit: prefer method arguments all on next lines of it has to span multiple lines

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3OutputStream.java
```diff
@@ -127,7 +134,8 @@ public void testAbortAfterFailedPartUpload() {
   public void testAbortMultipart() {
     doThrow(new RuntimeException()).when(s3mock).completeMultipartUpload((CompleteMultipartUploadRequest) any());
 
-    try (S3OutputStream stream = new S3OutputStream(s3mock, randomURI(), properties, nullMetrics())) {
+    try (S3OutputStream stream = new S3OutputStream(s3mock, randomURI(), properties, nullMetrics(),
```
jackye1995, line 137: nit: prefer method arguments all on next lines of it has to span multiple lines

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -110,6 +115,8 @@ private S3Client client() {
   @Override
   public void initialize(Map<String, String> properties) {
     this.awsProperties = new AwsProperties(properties);
+    this.writeTags = toTags(
```
jackye1995, line 118: nit: I think we can combine this line into the helper method, that makes the purpose of the helper method clearer, like `this.writeTags = toTags(properties)`, and make the tag map extraction a part of the helper.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -215,6 +216,14 @@
    */
   public static final String CLIENT_ASSUME_ROLE_REGION = "client.assume-role.region";
 
+  /**
+   * Used by {@link S3FileIO} to tag objects when writing. To set, we can pass a catalog property.
+   * Example in Spark: --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key=my_val
```
jackye1995, line 221: nit: example should also be below <p> not above

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
```diff
@@ -21,22 +21,24 @@
...
 import org.apache.iceberg.io.PositionOutputStream;
 import org.apache.iceberg.metrics.MetricsContext;
 import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.Tag;
 
 public class S3OutputFile extends BaseS3File implements OutputFile {
   public static S3OutputFile fromLocation(String location, S3Client client, AwsProperties awsProperties,
-      MetricsContext metrics) {
-    return new S3OutputFile(client, new S3URI(location), awsProperties, metrics);
+      MetricsContext metrics, Set<Tag> writeTags) {
```
jackye1995, line 36: public method should be be changed, we need to add a separated method

rajarshisarkar: Separated the new method.

jackye1995 (approved):
looks good to me!

amogh-jahagirdar (approved):
Looks good to me!

PR #4260, opened 2022-03-03
Core: Implement table refs in TableMetadataParser

rdblue:
This updates TableMetadataParser to read and write the [refs map](https://iceberg.apache.org/spec/#snapshot-reference).

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
@@ -301,7 +301,7 @@ public class TableMetadata implements Serializable {
     this.schemasById = indexSchemas();
     this.specsById = indexSpecs(specs);
     this.sortOrdersById = indexSortOrders(sortOrders);
-    this.refs = validateRefs(refs);
+    this.refs = validateRefs(currentSnapshotId, refs, snapshotsById);
 
     HistoryEntry last = null;
     for (HistoryEntry logEntry : snapshotLog) {
@@ -465,6 +465,14 @@ public class TableMetadata implements Serializable {
 
   public List<Snapshot> snapshots() {
     return snapshots;
+  }
+
+  public SnapshotRef ref(String name) {
+    return refs.get(name);
+  }
+
+  public Map<String, SnapshotRef> refs() {
+    return refs;
   }
 
   public List<HistoryEntry> snapshotLog() {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -720,15 +728,24 @@ private static SortOrder freshSortOrder(int orderId, Schema schema, SortOrder so
...
     }
+
+    SnapshotRef main = inputRefs.get(SnapshotRef.MAIN_BRANCH);
+    if (currentSnapshotId != -1) {
+      Preconditions.checkArgument(main == null || currentSnapshotId == main.snapshotId(),
+          "Current snapshot ID does not match main branch (%s != %s)",
+          currentSnapshotId, main != null ? main.snapshotId() : null);
+    } else {
+      Preconditions.checkArgument(main == null,
+          "Current snapshot is not set, but main branch exists: %s", main);
```
nastra, line 746: nit: should this mention `...snapshot ID is not set...` to be consistent with the error message above?

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -720,15 +728,24 @@ private static SortOrder freshSortOrder(int orderId, Schema schema, SortOrder so
...
+                                                       Map<Long, Snapshot> snapshotsById) {
+    for (SnapshotRef ref : inputRefs.values()) {
+      Preconditions.checkArgument(snapshotsById.containsKey(ref.snapshotId()),
+          "Snapshot for reference %s does not exist in the existing snapshots list", ref);
     }
+
+    SnapshotRef main = inputRefs.get(SnapshotRef.MAIN_BRANCH);
+    if (currentSnapshotId != -1) {
+      Preconditions.checkArgument(main == null || currentSnapshotId == main.snapshotId(),
+          "Current snapshot ID does not match main branch (%s != %s)",
```
nastra, line 742: nit: I wonder if the message would be slightly clearer like this: `Current snapshot ID '%s' does not match the snapshot ID 
 '%s' of the main branch`

nastra (approved):
changes LGTM once CI passes

kbendick (approved):
+1. This looks good to me once tests are passing.

danielcweeks (approved):
LGTM

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: I strongly prefer keeping the error message itself readable and adding context after. While it is more clear which ID is which in your version, I think it makes the actual problem harder to read because there are longs embedded in the text.

rdblue:
Thanks for the reviews, everyone!

PR #3717, opened 2021-12-11
Core: Commit state should be unknown even when new location is not in history

jfz:
This PR fixes an issue where a commit succeeds but its metadata file is deleted and table is corrupted.

Currently when there's an exception during commit, we refresh table and check commit state by looking at if the new metadata location matches current or in history, and commit state is Failure if not, and the current metadata file is deleted.
There's a possibility that when the check happens, the metadata remote server (e.g. Hive metastore) is still processing the commit and eventually succeed later, in this case, the commit(atomic location swap) succeeds but metadata file is deleted, and the table is in bad state.
The commit state should be considered as unknown to avoid metadata being deleted, also the error propagated to user is more accurate.

PR #4261, opened 2022-03-03
Hive: Known exception should not become CommitStateUnknownException

szehon-ho:
After #3717 all HIve exceptions become CommitStateUnknownException.  That is because checkCommitStatus stops returning FAILURE mode and always either SUCCESS or UNKNOWN, and subsequent specific error-handling blocks [specific catch blocks](https://github.com/apache/iceberg/blob/master/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L287) never get run.

For instance, doing a create table statement with invalid character in name:  
```CREATE TABLE `tbl` (id INT) USING ICEBERG```

Gets this kind of confusing and alarming error message for user:

```
org.apache.iceberg.exceptions.CommitStateUnknownException: `tbl` is not a valid object name
Cannot determine whether the commit was successful or not, the underlying data files may or may not be needed. Manual intervention via the Remove Orphan Files Action can remove these files when a connection to the Catalog can be re-established if the commit was actually unsuccessful.
Please check to see whether or not your commit was successful before retrying this commit. Retrying an already successful operation will result in duplicate records or unintentional modifications.
At this time no files will be deleted including possibly unused manifest lists.
```

Also, another side effect is there is a NPE in checkCommitStatus() code, which was not designed for when table is first being created.

```
java.lang.NullPointerException
	at org.apache.iceberg.BaseMetastoreTableOperations.lambda$checkCommitStatus$4(BaseMetastoreTableOperations.java:302)
	at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:404)
	at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:214)
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:198)
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:190)
	at org.apache.iceberg.BaseMetastoreTableOperations.checkCommitStatus(BaseMetastoreTableOperations.java:300)
	at org.apache.iceberg.hive.HiveTableOperations.doCommit(HiveTableOperations.java:285)
	at org.apache.iceberg.BaseMetastoreTableOperations.commit(BaseMetastoreTableOperations.java:127)
	at org.apache.iceberg.BaseMetastoreCatalog$BaseMetastoreCatalogTableBuilder.create(BaseMetastoreCatalog.java:165)
	at org.apache.iceberg.catalog.Catalog.createTable(Catalog.java:78)
	at org.apache.iceberg.catalog.Catalog.createTable(Catalog.java:112)
```

This change just pushes down the original Hive known error handling for user-errors, and skips the CommitStateUnknown handling for those.  A new handling InvalidObjectException is added to catch the common user-error of invalid table names.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -38,6 +38,7 @@ import org.apache.hadoop.hive.common.StatsSetupConst;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.TableType;
 import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
+import org.apache.hadoop.hive.metastore.api.InvalidObjectException;
 import org.apache.hadoop.hive.metastore.api.LockComponent;
 import org.apache.hadoop.hive.metastore.api.LockLevel;
 import org.apache.hadoop.hive.metastore.api.LockRequest;
@@ -61,6 +62,7 @@ import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.CommitStateUnknownException;
 import org.apache.iceberg.exceptions.NoSuchIcebergTableException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
+import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.hadoop.ConfigProperties;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
@@ -271,29 +273,32 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
       try {
         persistTable(tbl, updateHiveTable);
         commitStatus = CommitStatus.SUCCESS;
-      } catch (Throwable persistFailure) {
+      } catch (org.apache.hadoop.hive.metastore.api.AlreadyExistsException e) {
+        throw new AlreadyExistsException(e, "Table already exists: %s.%s", database, tableName);
+
+      } catch (InvalidObjectException e) {
+        throw new ValidationException(e, "Invalid table name for %s.%s", database, tableName);
+
+      } catch (Throwable e) {
+        if (e.getMessage() != null && e.getMessage().contains("Table/View 'HIVE_LOCKS' does not exist")) {
+          throw new RuntimeException("Failed to acquire locks from metastore because 'HIVE_LOCKS' doesn't " +
+              "exist, this probably happened when using embedded metastore or doesn't create a " +
+              "transactional meta table. To fix this, use an alternative metastore", e);
+        }
+
         LOG.error("Cannot tell if commit to {}.{} succeeded, attempting to reconnect and check.",
-            database, tableName, persistFailure);
+            database, tableName, e);
         commitStatus = checkCommitStatus(newMetadataLocation, metadata);
         switch (commitStatus) {
           case SUCCESS:
             break;
           case FAILURE:
-            throw persistFailure;
+            throw e;
           case UNKNOWN:
-            throw new CommitStateUnknownException(persistFailure);
+            throw new CommitStateUnknownException(e);
         }
       }
-    } catch (org.apache.hadoop.hive.metastore.api.AlreadyExistsException e) {
-      throw new AlreadyExistsException("Table already exists: %s.%s", database, tableName);
-
     } catch (TException | UnknownHostException e) {
-      if (e.getMessage() != null && e.getMessage().contains("Table/View 'HIVE_LOCKS' does not exist")) {
-        throw new RuntimeException("Failed to acquire locks from metastore because 'HIVE_LOCKS' doesn't " +
-            "exist, this probably happened when using embedded metastore or doesn't create a " +
-            "transactional meta table. To fix this, use an alternative metastore", e);
-      }
-
       throw new RuntimeException(String.format("Metastore operation failed for %s.%s", database, tableName), e);
 
     } catch (InterruptedException e) {
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -277,13 +277,13 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
         throw new AlreadyExistsException(e, "Table already exists: %s.%s", database, tableName);
 
       } catch (InvalidObjectException e) {
-        throw new ValidationException(e, "Invalid table name for %s.%s", database, tableName);
+        throw new ValidationException(e, "Invalid Hive object for %s.%s", database, tableName);
 
       } catch (Throwable e) {
         if (e.getMessage() != null && e.getMessage().contains("Table/View 'HIVE_LOCKS' does not exist")) {
-          throw new RuntimeException("Failed to acquire locks from metastore because 'HIVE_LOCKS' doesn't " +
-              "exist, this probably happened when using embedded metastore or doesn't create a " +
-              "transactional meta table. To fix this, use an alternative metastore", e);
+          throw new RuntimeException("Failed to acquire locks from metastore because the underlying metastore " +
+              "table 'HIVE_LOCKS' does not exist. This can occur when using an embedded metastore which does not " +
+              "support transactions. To fix this use an alternative metastore.", e);
         }
 
         LOG.error("Cannot tell if commit to {}.{} succeeded, attempting to reconnect and check.",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by szehon-ho** [comment]
fyi @RussellSpitzer @rdblue @pvary  when you have a chance, can you see if this makes sense?

File: hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java [hunk 1]
```diff
@@ -271,29 +273,32 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       try {
         persistTable(tbl, updateHiveTable);
         commitStatus = CommitStatus.SUCCESS;
-      } catch (Throwable persistFailure) {
+      } catch (org.apache.hadoop.hive.metastore.api.AlreadyExistsException e) {
+        throw new AlreadyExistsException(e, "Table already exists: %s.%s", database, tableName);
```
RussellSpitzer, line 277: This I feel safe about

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -271,29 +273,32 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+
+      } catch (InvalidObjectException e) {
+        throw new ValidationException(e, "Invalid table name for %s.%s", database, tableName);
```
RussellSpitzer, line 280: This scares me a bit more, is InvalidObjectException only thrown for this sort of naming issue?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -271,29 +273,32 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+
+      } catch (Throwable e) {
+        if (e.getMessage() != null && e.getMessage().contains("Table/View 'HIVE_LOCKS' does not exist")) {
+          throw new RuntimeException("Failed to acquire locks from metastore because 'HIVE_LOCKS' doesn't " +
```
RussellSpitzer, line 284: Maybe ... "Failed to acquire locks from metastore because the underlying metastore table 'HIVE_LOCKS' does not exist. This can occur when using an embedded metastore which does not support transactions. To fix this use an alternative metastore."

RussellSpitzer:
I'm only worried about the Invalid Object Exception, everything else looks good to me

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: This is from the original code, I agree yours is better and can change it

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: Yea so I grepped through the instances it throws in Hive code base, it is always like Precondition check in Iceberg for user-error, and never proceeds further.

This is a github search (which unfortunately returns only a few top results):

https://github.com/search?q=%22throw+new+InvalidObjectException*%22+repo%3Aapache%2Fhive&type=code

My grep finds it thrown only in these classes which you can also double-check inside:
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/StatObjectConverter.java
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java


It should not be thrown when Hive commits successfully, or disconnect issue, if that is the fear?  Maybe @pvary can verify?

RussellSpitzer: It's the unknown unknowns I worry about. I just don't have enough familiarity with the codebase to even know where I should be looking or If I can guarantee behaviors in the future.

szehon-ho: So I'll try to explain the HMS flow (using the latest Hive release tag github links to make it persistent):

Every call goes through HMSHandler.  For these two calls (CreateTable or AlterTable), it goes like:

[HMSHandler::create_table_core](https://github.com/apache/hive/blob/8190d2be7b7165effa62bd21b7d60ef81fb0e4af/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java#L1789) -> [ObjectStore::createTable](https://github.com/apache/hive/blob/8190d2be7b7165effa62bd21b7d60ef81fb0e4af/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L1298), the call being at this line : ms.createTable()

[HMSHandler::alter_table_core](https://github.com/apache/hive/blob/8190d2be7b7165effa62bd21b7d60ef81fb0e4af/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java#L4998) -> [ObjectStore::alterTable](https://github.com/apache/hive/blob/8190d2be7b7165effa62bd21b7d60ef81fb0e4af/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L4005) , the call being at the line ms.alterTable()

After this it uses JDO persistence library which does not throw any InvalidObjectException.

So we check the four places.  InvalidObjectException is only ever thrown in the first one (HMSHandler::create_table_core).  

Note that ObjectStore methods do call convertToMTable() that does throw the InvalidObjectException if database is null, but in Iceberg case this is actually checked way before it gets here (can share the stack trace if you want).

Not sure if this helps?

szehon-ho: The alternative, if we are very paranoid, can check the exact exception message to make sure its the exception we want.  Would be a bit hacky but it should be the same purpose as this block is to expect only that exception.

szehon-ho: Makes sense, maybe @pvary could take a look ?

aokolnychyi: I do share Russell's concern as `InvalidObjectException` sounds like something that can be returned not only when the table is not there. However, I also don't know the Hive codebase. I'd ask @pvary.

@szehon-ho, how did you hit this issue, btw? Do you have an example?

pvary: Hive error handling is not something I would be proud of. There are plenty of inconsistencies around the code, and the last time I have tried to clean this up, I have faced serious resistance. The reasoning behind it was that the users of HMS are plenty and they might build above this haphazard error handling.

That said:
- InvalidObjectException is generally used for any error in the provided objects:
  - `You must specify a path for the catalog`
  - `BLABLA is not a valid catalog name`
  - `Invalid type name`
  - `Invalid column ...`
- The error could signal missing objects as well - typically when storing related objects:
  - `No such catalog BLABLA`
  - `Unable to add partition because table or database do not exist`
  - `Database BLABLA doesn't exist`
  
So I would definitely opt for a more general error message, maybe mentioning that the most common error is the wrong table name. All that said, I think the `ValidationException` seems ok, since there is some validation error happened when we tried to store the table.

szehon-ho: Thanks guys for the comment, yea I will definitely change the message.  @aokolnychyi does the description in the issue explain it?

RussellSpitzer: Sounds like we have a consensus that it is ok to catch and rethrow InvalidObjectException these and not change them to commit state unknown. The only thing that remains is the error message.

szehon-ho:
Changed the messages, thanks all for diligence and discussion

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: Done

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: Done

pvary (approved):

RussellSpitzer (approved):

RussellSpitzer:
Thanks @szehon-ho for the pr and @pvary for the review!

PR #4262, opened 2022-03-03
Python support on Literals

dramaticlly:
First step of closing https://github.com/apache/iceberg/issues/4250

This is my first iceberg python contribution, so there might be some naive mistake and please let me know so I can jump to fix, most of the code is refactoring from https://github.com/apache/iceberg/blob/master/python_legacy/iceberg/api/expressions/literals.py to our latest python module

Change highlights
- Refactor all static method to module level methods to be more pythonic
- Use [singledispath](https://docs.python.org/3/library/functools.html#functools.singledispatch) to dynamic overload the Literal factory from python primitive type and conversion between iceberg types
- Adopt the new TimestampType from types, 
   - replace `TimestampType.with_timezone` with `TimestampTzType` 
   - replace `TimestampType.without_timezone` with `TimestampType` 
- Combined 3 literal unit tests together with parametrized pytest style python unit tests

PR #4263, opened 2022-03-04
Python: Serialization/Deserialization for types and partition values (port of conversions.py from python_legacy)

samredai:
This adds functionality for converting a value to bytes and vice-versa augmented by the `PrimitiveType` subclass provided.

Functionally this is a direct port of [conversion.py](https://github.com/apache/iceberg/blob/master/python_legacy/iceberg/api/types/conversions.py) in the python legacy library. The main changes are:
1. I replaced the large dictionaries containing `<type>: <lambda function>` pairs with [@singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch)
2. I removed the [TypeID](https://github.com/apache/iceberg/blob/master/python_legacy/iceberg/api/types/type.py#L24) enum and instead just used the types directly.

Overall this is more lines of source code but I'll summarize my reasonings for the above two changes:

1. I believe the intention behind the lambda maps was to allow for generic `from_...` and `to_...` functions that contain dynamic logic based on the argument type, similar to constructor overloading in Java. I feel that using `@singledispatch` achieves that more directly by creating a generic function and then registering functions based on the first argument's type. This also allows stacking the register decorators for types that share the same logic (such as `IntegerType` and `DateType`) instead of defining that logic multiple times. Also it's a stable feature added as early as python 3.4 so I believe it's future-proof.
2. The `TypeID` enum in the legacy python client carries some information in the enum value, for example `BOOLEAN = {"java_class": "Boolean.class", "python_class": bool, "id": 1}`. These dictionary values weren't used in `conversions.py` but it looks like they're used in other places such as the partition spec. I removed it because this also felt like it may be adding some unnecessary indirection. If anyone feels this should be put back I'm ok with that but I would suggest that we see where the relationships in these dictionaries are required and localize these mappings to that area of the source code (maybe with a simple `PY_TO_JAVA_CLASS = {...}` object and a `get_class_id(...)` function).

### DIFF
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -44,7 +44,6 @@ import struct
 import uuid
 from decimal import ROUND_HALF_DOWN, Decimal, localcontext
 from functools import singledispatch
-from typing import Any
 
 from iceberg.types import (
     BinaryType,
@@ -124,24 +123,25 @@ def _(primitive_type, partition_value_as_str: str):
 
 
 @singledispatch
-def to_bytes(primitive_type, value: Any) -> bytes:
+def to_bytes(primitive_type, value) -> bytes:
     """A generic function which converts a built-in python value to bytes
 
     Args:
         primitive_type(PrimitiveType): An implementation of the PrimitiveType base class
-        value(Any): The value to convert to bytes
+        value: The value to convert to bytes (The type of this value depends on which dispatched function is
+            used--check dispatchable functions for typehints)
     """
     raise TypeError(f"Cannot serialize value, type {primitive_type} not supported: '{value}'")
 
 
 @to_bytes.register(BooleanType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: bool) -> bytes:
     return struct.pack("<?", 1 if value else 0)
 
 
 @to_bytes.register(IntegerType)
 @to_bytes.register(DateType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: int) -> bytes:
     return struct.pack("<i", value)
 
 
@@ -149,38 +149,38 @@ def _(primitive_type, value: Any) -> bytes:
 @to_bytes.register(TimeType)
 @to_bytes.register(TimestampType)
 @to_bytes.register(TimestamptzType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: int) -> bytes:
     return struct.pack("<q", value)
 
 
 @to_bytes.register(FloatType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: float) -> bytes:
     return struct.pack("<f", value)
 
 
 @to_bytes.register(DoubleType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: float) -> bytes:
     return struct.pack("<d", value)
 
 
 @to_bytes.register(StringType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: str) -> bytes:
     return value.encode("UTF-8")
 
 
 @to_bytes.register(UUIDType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: uuid.UUID) -> bytes:
     return struct.pack(">QQ", (value.int >> 64) & 0xFFFFFFFFFFFFFFFF, value.int & 0xFFFFFFFFFFFFFFFF)
 
 
 @to_bytes.register(BinaryType)
 @to_bytes.register(FixedType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: bytes) -> bytes:
     return value
 
 
 @to_bytes.register(DecimalType)
-def _(primitive_type, value: Any) -> bytes:
+def _(primitive_type, value: Decimal) -> bytes:
     value_as_tuple = value.as_tuple()
     value_precision = len(value_as_tuple.digits)
     value_scale = -value_as_tuple.exponent
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -78,7 +78,8 @@ def decimal_to_unscaled(value: Decimal) -> int:
     Returns:
         int: The unscaled value
     """
-    return int(value.to_integral_value())
+    sign, digits, _ = value.as_tuple()
+    return int(Decimal((sign, digits, 0)).to_integral_value())
 
 
 def unscaled_to_decimal(unscaled: int, scale: int) -> Decimal:
@@ -92,7 +93,7 @@ def unscaled_to_decimal(unscaled: int, scale: int) -> Decimal:
         Decimal: A scaled Decimal instance
     """
     sign, digits, _ = Decimal(unscaled).as_tuple()
-    return Decimal((sign, digits, scale))
+    return Decimal((sign, digits, -scale))
 
 
 @singledispatch
@@ -315,4 +316,4 @@ def _(primitive_type, b: bytes) -> bytes:
 @from_bytes.register(DecimalType)
 def _(primitive_type, buf: bytes) -> Decimal:
     unscaled = int.from_bytes(buf, "big", signed=True)
-    return unscaled_to_decimal(unscaled, -primitive_type.scale)
+    return unscaled_to_decimal(unscaled, primitive_type.scale)
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -29,7 +29,7 @@ Note:
 """
 import struct
 import uuid
-from decimal import Decimal, localcontext
+from decimal import Decimal
 from functools import singledispatch
 from typing import Union
 
@@ -52,110 +52,114 @@ from iceberg.types import (
 )
 
 
-def convert_decimal_to_unscaled_value(decimal_type_var: DecimalType, value: Decimal) -> int:
-    """Get an unscaled value given a Decimal value and a DecimalType instance with defined precision and scale
+def handle_none(func):
+    """A decorator function to handle cases where partition values are `None` or "__HIVE_DEFAULT_PARTITION__"
 
     Args:
-        decimal_type_var (DecimalType): A DecimalType instance with precision and scale
-        value (Decimal): A Decimal instance
+        func (Callable): A function registered to the singledispatch function `partition_to_py`
+    """
 
-    Raises:
-        ValueError: If either the scale or precision of `value` does not match that defined in the DecimalType instance provided as `decimal_type_var`
+    def wrapper(primitive_type, value_str):
+        if value_str is None:
+            return None
+        elif value_str == "__HIVE_DEFAULT_PARTITION__":
+            return None
+        return func(primitive_type, value_str)
+
+    return wrapper
+
+
+def decimal_to_unscaled(value: Decimal) -> int:
+    """Get an unscaled value given a Decimal value
+
+    Args:
+        value (Decimal): A Decimal instance
 
     Returns:
         int: The unscaled value
     """
-    value_as_tuple = value.as_tuple()
-    value_precision = len(value_as_tuple.digits)
-    value_scale = -value_as_tuple.exponent
+    return int(value.to_integral_value())
 
-    if value_scale != decimal_type_var.scale:
-        raise ValueError(f"Cannot serialize value, scale of value does not match type {decimal_type_var}: {value_scale}")
-    elif value_precision > decimal_type_var.precision:
-        raise ValueError(
-            f"Cannot serialize value, precision of value is greater than precision of type {decimal_type_var}: {value_precision}"
-        )
 
-    with localcontext() as ctx:
-        ctx.prec = decimal_type_var.precision
-        sign, digits, exponent = value.as_tuple()
-        value_w_adjusted_scale = Decimal((sign, digits, exponent + decimal_type_var.scale))
-        return int(value_w_adjusted_scale.to_integral_value())
+def unscaled_to_decimal(unscaled: int, scale: int) -> Decimal:
+    """Get a scaled Decimal value given an unscaled value and a scale
+
+    Args:
+        unscaled (int): An unscaled value
+        scale (int): A scale to set for the returned Decimal instance
+
+    Returns:
+        Decimal: A scaled Decimal instance
+    """
+    sign, digits, _ = Decimal(unscaled).as_tuple()
+    return Decimal((sign, digits, scale))
 
 
 @singledispatch
-def from_partition_value_to_py(primitive_type, partition_value_as_str: str):
-    """A generic function function which converts a partition string to a python built-in
+def partition_to_py(primitive_type, value_str: str):
+    """A generic function which converts a partition string to a python built-in
 
     Args:
         primitive_type(PrimitiveType): An implementation of the PrimitiveType base class
-        partition_value_as_str(str): A string representation of a partition value
+        value_str(str): A string representation of a partition value
     """
-    if partition_value_as_str is None:
-        return None
-    raise TypeError(
-        f"Cannot convert partition string to python built-in, type {primitive_type} not supported: '{partition_value_as_str}'"
-    )
+    raise TypeError(f"Cannot convert '{value_str}' to unsupported type: {primitive_type}")
 
 
-@from_partition_value_to_py.register(BooleanType)
-def _(primitive_type, partition_value_as_str: str):
-    if partition_value_as_str is None:
-        return None
-    return partition_value_as_str.lower() == "true"
+@partition_to_py.register(BooleanType)
+@handle_none
+def _(primitive_type, value_str: str) -> Union[int, float, str, uuid.UUID]:
+    return value_str.lower() == "true"
 
 
-@from_partition_value_to_py.register(IntegerType)
-@from_partition_value_to_py.register(LongType)
-def _(primitive_type, partition_value_as_str: str):
+@partition_to_py.register(IntegerType)
+@partition_to_py.register(LongType)
+@partition_to_py.register(DateType)
+@partition_to_py.register(TimeType)
+@partition_to_py.register(TimestampType)
+@partition_to_py.register(TimestamptzType)
+@handle_none
+def _(primitive_type, value_str: str) -> int:
     """
-    Note: Before attempting to cast the value to string, a validation happens to ensure that there are no fractional digits. For
-    example, an invalid value such as "123.45" will be converted to a decimal with digits (1, 2, 3, 4, 5) with an exponent of -2 and the last
-    2 digits will be inspected for non-zero values. An example of a valid value is "123.00".
+    Raises:
+        ValueError: If the scale/exponent is not 0
     """
-    if partition_value_as_str is None:
-        return None
-    _, digits, exponent = Decimal(partition_value_as_str).as_tuple()
-    if exponent != 0 and any(digits[exponent:]):  # If there are digits to the right of the exponent, raise if any are not 0
+    _, digits, exponent = Decimal(value_str).as_tuple()
+    if exponent != 0:  # Raise if there are digits to the right of the decimal
         raise ValueError(f"Cannot convert partition value, value cannot have fractional digits for {primitive_type} partition")
-    return int(float(partition_value_as_str))
+    return int(float(value_str))
 
 
-@from_partition_value_to_py.register(FloatType)
-@from_partition_value_to_py.register(DoubleType)
-def _(primitive_type, partition_value_as_str: str):
-    if partition_value_as_str is None:
-        return None
-    return float(partition_value_as_str)
+@partition_to_py.register(FloatType)
+@partition_to_py.register(DoubleType)
+@handle_none
+def _(primitive_type, value_str: str) -> float:
+    return float(value_str)
 
 
-@from_partition_value_to_py.register(StringType)
-def _(primitive_type, partition_value_as_str: str):
-    if partition_value_as_str is None:
-        return None
-    return partition_value_as_str
+@partition_to_py.register(StringType)
+@handle_none
+def _(primitive_type, value_str: str) -> str:
+    return value_str
 
 
-@from_partition_value_to_py.register(UUIDType)
-def _(primitive_type, partition_value_as_str: str):
-    if partition_value_as_str is None:
-        return None
-    return uuid.UUID(partition_value_as_str)
+@partition_to_py.register(UUIDType)
+@handle_none
+def _(primitive_type, value_str: str) -> uuid.UUID:
+    return uuid.UUID(value_str)
 
 
-@from_partition_value_to_py.register(FixedType)
-@from_partition_value_to_py.register(BinaryType)
-def _(primitive_type, partition_value_as_str: str):
-    if partition_value_as_str is None:
-        return None
-    return bytes(partition_value_as_str, "UTF-8")
+@partition_to_py.register(FixedType)
+@partition_to_py.register(BinaryType)
+@handle_none
+def _(primitive_type, value_str: str) -> bytes:
+    return bytes(value_str, "UTF-8")
 
 
-@from_partition_value_to_py.register(DecimalType)
-def _(primitive_type, partition_value_as_str: str):
-    if partition_value_as_str is None:
-        return None
-    return Decimal(partition_value_as_str)
+@partition_to_py.register(DecimalType)
+@handle_none
+def _(primitive_type, value_str: str) -> Decimal:
+    return Decimal(value_str)
 
 
 @singledispatch
@@ -170,7 +174,7 @@ def to_bytes(primitive_type: PrimitiveType, value: Union[bool, bytes, Decimal, f
         value: The value to convert to bytes (The type of this value depends on which dispatched function is
             used--check dispatchable functions for typehints)
     """
-    raise TypeError(f"Cannot serialize value, type {primitive_type} not supported: '{str(value)}'")
+    raise TypeError(f"scale does not match {primitive_type}")
 
 
 @to_bytes.register(BooleanType)
@@ -224,7 +228,29 @@ def _(primitive_type, value: bytes) -> bytes:
 
 @to_bytes.register(DecimalType)
 def _(primitive_type, value: Decimal) -> bytes:
-    unscaled_value = convert_decimal_to_unscaled_value(decimal_type_var=primitive_type, value=value)
+    """Convert a Decimal value to bytes given a DecimalType instance with defined precision and scale
+
+    Args:
+        primitive_type (Decimal): A DecimalType instance with precision and scale
+        value (Decimal): A Decimal instance
+
+    Raises:
+        ValueError: If either the precision or scale of `value` does not match that defined in the DecimalType instance
+
+
+    Returns:
+        bytes: The byte representation of `value`
+    """
+    sign, digits, exponent = value.as_tuple()
+
+    if -exponent != primitive_type.scale:
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {primitive_type}: {-exponent}")
+    elif len(digits) > primitive_type.precision:
+        raise ValueError(
+            f"Cannot serialize value, precision of value is greater than precision of type {primitive_type}: {len(digits)}"
+        )
+
+    unscaled_value = decimal_to_unscaled(value=Decimal((sign, digits, 0)))
     min_num_bytes = ((unscaled_value).bit_length() + 7) // 8
     return unscaled_value.to_bytes(min_num_bytes, "big", signed=True)
 
@@ -287,8 +313,6 @@ def _(primitive_type, b: bytes) -> bytes:
 
 
 @from_bytes.register(DecimalType)
-def _(primitive_type, b: bytes) -> Decimal:
-    integer_representation = int.from_bytes(b, "big", signed=True)
-    with localcontext() as ctx:
-        ctx.prec = primitive_type.precision
-        return Decimal(integer_representation) * Decimal((0, (1,), -primitive_type.scale))
+def _(primitive_type, buf: bytes) -> Decimal:
+    unscaled = int.from_bytes(buf, "big", signed=True)
+    return unscaled_to_decimal(unscaled, -primitive_type.scale)
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -155,6 +155,10 @@ def _(primitive_type, value: int) -> bytes:
 
 @to_bytes.register(FloatType)
 def _(primitive_type, value: float) -> bytes:
+    """
+    Note: float in python is implemented using a double in C. Therefore this involves a conversion of a 32-bit (single precision)
+    float to a 64-bit (double precision) float which introduces some imprecision.
+    """
     return struct.pack("<f", value)
 
 
@@ -231,10 +235,6 @@ def _(primitive_type, b: bytes):
 
 @from_bytes.register(FloatType)
 def _(primitive_type, b: bytes):
-    """
-    Note: float in python is implemented using a double in C. Therefore this involves a conversion of a 32-bit (single precision)
-    float to a 64-bit (double precision) float which introduces some imprecision.
-    """
     return struct.unpack("<f", b)[0]
 
 
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -231,6 +231,10 @@ def _(primitive_type, b: bytes):
 
 @from_bytes.register(FloatType)
 def _(primitive_type, b: bytes):
+    """
+    Note: float in python is implemented using a double in C. Therefore this involves a conversion of a 32-bit (single precision)
+    float to a 64-bit (double precision) float which introduces some imprecision.
+    """
     return struct.unpack("<f", b)[0]
 
 
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -258,4 +258,4 @@ def _(primitive_type, b: bytes):
     integer_representation = int.from_bytes(b, "big", signed=True)
     with localcontext() as ctx:
         ctx.prec = primitive_type.precision
-        return Decimal(integer_representation) * Decimal(10) ** Decimal(-primitive_type.scale)
+        return Decimal(integer_representation) * (Decimal(10) ** Decimal(-primitive_type.scale))
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -42,7 +42,7 @@ Example (stacking registrations):
 """
 import struct
 import uuid
-from decimal import ROUND_HALF_UP, Decimal
+from decimal import ROUND_HALF_DOWN, Decimal, localcontext
 from functools import singledispatch
 from typing import Any
 
@@ -181,11 +181,20 @@ def _(type_, value: Any) -> bytes:
 
 @to_bytes.register(DecimalType)
 def _(type_, value: Any) -> bytes:
-    scale = abs(value.as_tuple().exponent)
-    quantized_value = value.quantize(Decimal("10") ** -scale)
-    unscaled_value = int((quantized_value * 10**scale).to_integral_value())
-    min_num_bytes = (unscaled_value.bit_length() + 7) // 8
-    return unscaled_value.to_bytes(min_num_bytes, "big", signed=True)
+    value_as_tuple = value.as_tuple()
+    value_precision = len(value_as_tuple.digits)
+    value_scale = -value_as_tuple.exponent
+
+    if value_scale != type_.scale:
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {type_}: {value_scale}")
+    elif value_precision != type_.precision:
+        raise ValueError(f"Cannot serialize value, precision of value does not match type {type_}: {value_precision}")
+
+    with localcontext() as ctx:
+        ctx.prec = type_.precision
+        unscaled_value = int((value * 10**type_.scale).to_integral_value(rounding=ROUND_HALF_DOWN))
+        min_num_bytes = ((unscaled_value).bit_length() + 7) // 8
+        return unscaled_value.to_bytes(min_num_bytes, "big", signed=True)
 
 
 @singledispatch
@@ -246,6 +255,7 @@ def _(type_, b: bytes):
 
 @from_bytes.register(DecimalType)
 def _(type_, b: bytes):
-    return Decimal(int.from_bytes(b, "big", signed=True) * 10**-type_.scale).quantize(
-        Decimal("." + "".join(["0" for i in range(1, type_.scale)]) + "1"), rounding=ROUND_HALF_UP
-    )
+    integer_representation = int.from_bytes(b, "big", signed=True)
+    with localcontext() as ctx:
+        ctx.prec = type_.precision
+        return Decimal(integer_representation) * Decimal(10) ** Decimal(-type_.scale)
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -63,8 +63,6 @@ from iceberg.types import (
     UUIDType,
 )
 
-HIVE_NULL = "__HIVE_DEFAULT_PARTITION__"
-
 
 @singledispatch
 def from_partition_value_to_py(primitive_type, partition_value_as_str: str):
@@ -74,7 +72,7 @@ def from_partition_value_to_py(primitive_type, partition_value_as_str: str):
         primitive_type(PrimitiveType): An implementation of the PrimitiveType base class
         partition_value_as_str(str): A string representation of a partition value
     """
-    if partition_value_as_str is None or partition_value_as_str == HIVE_NULL:
+    if partition_value_as_str is None:
         return None
     raise TypeError(
         f"Cannot convert partition string to python built-in, type {primitive_type} not supported: '{partition_value_as_str}'"
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -187,8 +187,10 @@ def _(primitive_type, value: Decimal) -> bytes:
 
     if value_scale != primitive_type.scale:
         raise ValueError(f"Cannot serialize value, scale of value does not match type {primitive_type}: {value_scale}")
-    elif value_precision != primitive_type.precision:
-        raise ValueError(f"Cannot serialize value, precision of value does not match type {primitive_type}: {value_precision}")
+    elif value_precision > primitive_type.precision:
+        raise ValueError(
+            f"Cannot serialize value, precision of value is greater than precision of type {primitive_type}: {value_precision}"
+        )
 
     with localcontext() as ctx:
         ctx.prec = primitive_type.precision
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -26,19 +26,6 @@ Note:
     are defined here as generic functions using the @singledispatch decorator. For each PrimitiveType
     implementation, a concrete function is registered for each generic conversion function. For PrimitiveType
     implementations that share the same conversion logic, registrations can be stacked.
-
-Example (registering `from_partition_to_py` logic for BooleanType):
-    >>> @from_partition_value_to_py.register(BooleanType)
-    >>> def _(primitive_type, partition_value_as_str: str):
-            if partition_value_as_str is None:
-                return None
-            return partition_value_as_str.lower() == "true"
-
-Example (stacking registrations):
-   >>> @from_partition_value_to_py.register(FloatType)
-   >>> @from_partition_value_to_py.register(DoubleType)
-   >>> def _(primitive_type, partition_value_as_str: str):
-           return float(partition_value_as_str)
 """
 import struct
 import uuid
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -1,0 +1,251 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Utility module for various conversions around PrimitiveType implementations
+
+This module enables:
+    - Converting partition strings to built-in python objects
+    - Converting a value to a byte buffer
+    - Converting a byte buffer to a value
+
+Note:
+    Conversion logic varies based on the PrimitiveType implementation. Therefore conversion functions
+    are defined here as generic functions using the @singledispatch decorator. For each PrimitiveType
+    implementation, a concrete function is registered for each generic conversion function. For PrimitiveType
+    implementations that share the same conversion logic, registrations can be stacked.
+
+Example (registering `from_partition_to_py` logic for BooleanType):
+    >>> @from_partition_value_to_py.register(BooleanType)
+    >>> def _(type_, partition_value_as_str: str):
+            if partition_value_as_str is None:
+                return False
+            return partition_value_as_str.lower() == "true"
+
+Example (stacking registrations):
+   >>> @from_partition_value_to_py.register(FloatType)
+   >>> @from_partition_value_to_py.register(DoubleType)
+   >>> def _(type_, partition_value_as_str: str):
+           return float(partition_value_as_str)
+"""
+import struct
+import uuid
+from decimal import ROUND_HALF_UP, Decimal
+from functools import singledispatch
+from typing import Any
+
+from iceberg.types import (
+    BinaryType,
+    BooleanType,
+    DateType,
+    DecimalType,
+    DoubleType,
+    FixedType,
+    FloatType,
+    IntegerType,
+    LongType,
+    StringType,
+    TimestampType,
+    TimestamptzType,
+    TimeType,
+    UUIDType,
+)
+
+HIVE_NULL = "__HIVE_DEFAULT_PARTITION__"
+
+
+@singledispatch
+def from_partition_value_to_py(type_, partition_value_as_str: str):
+    """A generic function function which converts a partition string to a python built-in
+
+    Args:
+        type_(PrimitiveType): An implementation of the PrimitiveType base class
+        partition_value_as_str(str): A string representation of a partition value
+    """
+    if partition_value_as_str is None or partition_value_as_str == HIVE_NULL:
+        return None
+    raise TypeError(f"Cannot convert partition string to python built-in, type {type_} not supported: '{partition_value_as_str}'")
+
+
+@from_partition_value_to_py.register(BooleanType)
+def _(type_, partition_value_as_str: str):
+    if partition_value_as_str is None:
+        return False
+    return partition_value_as_str.lower() == "true"
+
+
+@from_partition_value_to_py.register(IntegerType)
+@from_partition_value_to_py.register(LongType)
+def _(type_, partition_value_as_str: str):
+    return int(float(partition_value_as_str))
+
+
+@from_partition_value_to_py.register(FloatType)
+@from_partition_value_to_py.register(DoubleType)
+def _(type_, partition_value_as_str: str):
+    return float(partition_value_as_str)
+
+
+@from_partition_value_to_py.register(StringType)
+def _(type_, partition_value_as_str: str):
+    return partition_value_as_str
+
+
+@from_partition_value_to_py.register(UUIDType)
+def _(type_, partition_value_as_str: str):
+    return uuid.UUID(partition_value_as_str)
+
+
+@from_partition_value_to_py.register(FixedType)
+def _(type_, partition_value_as_str: str):
+    return bytearray(bytes(partition_value_as_str, "UTF-8"))
+
+
+@from_partition_value_to_py.register(BinaryType)
+def _(type_, partition_value_as_str: str):
+    return bytes(partition_value_as_str, "UTF-8")
+
+
+@from_partition_value_to_py.register(DecimalType)
+def _(type_, partition_value_as_str: str):
+    return Decimal(partition_value_as_str)
+
+
+@singledispatch
+def to_bytes(type_, value: Any) -> bytes:
+    """A generic function which converts a built-in python value to bytes
+
+    Args:
+        type_(PrimitiveType): An implementation of the PrimitiveType base class
+        value(Any): The value to convert to bytes
+    """
+    raise TypeError(f"Cannot serialize value, type {type_} not supported: '{value}'")
+
+
+@to_bytes.register(BooleanType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack("<?", 1 if value else 0)
+
+
+@to_bytes.register(IntegerType)
+@to_bytes.register(DateType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack("<i", value)
+
+
+@to_bytes.register(LongType)
+@to_bytes.register(TimeType)
+@to_bytes.register(TimestampType)
+@to_bytes.register(TimestamptzType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack("<q", value)
+
+
+@to_bytes.register(FloatType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack("<f", value)
+
+
+@to_bytes.register(DoubleType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack("<d", value)
+
+
+@to_bytes.register(StringType)
+def _(type_, value: Any) -> bytes:
+    return value.encode("UTF-8")
+
+
+@to_bytes.register(UUIDType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack(">QQ", (value.int >> 64) & 0xFFFFFFFFFFFFFFFF, value.int & 0xFFFFFFFFFFFFFFFF)
+
+
+@to_bytes.register(BinaryType)
+@to_bytes.register(FixedType)
+def _(type_, value: Any) -> bytes:
+    return value
+
+
+@to_bytes.register(DecimalType)
+def _(type_, value: Any) -> bytes:
+    scale = abs(value.as_tuple().exponent)
+    quantized_value = value.quantize(Decimal("10") ** -scale)
+    unscaled_value = int((quantized_value * 10**scale).to_integral_value())
+    min_num_bytes = (unscaled_value.bit_length() + 7) // 8
+    return unscaled_value.to_bytes(min_num_bytes, "big", signed=True)
+
+
+@singledispatch
+def from_bytes(type_, b: bytes):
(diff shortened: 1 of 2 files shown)


samredai:
cc: @cccs-eric @dramaticlly @CircArgs @rdblue @jun-he @cabhishek @TGooch44

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+        type_(PrimitiveType): An implementation of the PrimitiveType base class
+        partition_value_as_str(str): A string representation of a partition value
+    """
+    if partition_value_as_str is None or partition_value_as_str == HIVE_NULL:
+        return None
+    raise TypeError(f"Cannot convert partition string to python built-in, type {type_} not supported: '{partition_value_as_str}'")
+
+
+@from_partition_value_to_py.register(BooleanType)
+def _(type_, partition_value_as_str: str):
```
dramaticlly, line 83: hey @samredai , I appreciate your introduction of this dynamic overloading. I tried to adopt it in #4262 and looks like mypy is not happy about it as the method `_` now is being redefined multiple times with error code `no-redef`, do you know if we need to configure anything in mypy.ini or tox.ini to work around this issue?

dramaticlly: Basically this is what I saw

```
(venv) ➜  python git:(master) ✗ mypy --no-implicit-optional --show-error-codes  --config tox.ini src
tox.ini: No [mypy] section in config file
src/iceberg/expression/literals.py:27: error: Library stubs not installed for "pytz" (or incompatible with Python 3.8)  [import]
src/iceberg/expression/literals.py:27: note: Hint: "python3 -m pip install types-pytz"
src/iceberg/expression/literals.py:27: note: (or run "mypy --install-types" to install all missing stub packages)
src/iceberg/expression/literals.py:27: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
src/iceberg/expression/literals.py:109: error: Name "_of" already defined on line 104  [no-redef]
src/iceberg/expression/literals.py:119: error: Name "_of" already defined on line 104  [no-redef]
src/iceberg/expression/literals.py:129: error: Name "_of" already defined on line 104  [no-redef]
src/iceberg/expression/literals.py:134: error: Name "_of" already defined on line 104  [no-redef]
src/iceberg/expression/literals.py:139: error: Name "_of" already defined on line 104  [no-redef]
src/iceberg/expression/literals.py:144: error: Name "_of" already defined on line 104  [no-redef]
src/iceberg/expression/literals.py:149: error: Name "_of" already defined on line 104  [no-redef]
```

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+@from_bytes.register(BinaryType)
+@from_bytes.register(FixedType)
+def _(type_, b: bytes):
+    return b
+
+
+@from_bytes.register(DecimalType)
+def _(type_, b: bytes):
+    return Decimal(int.from_bytes(b, "big", signed=True) * 10**-type_.scale).quantize(
+        Decimal("." + "".join(["0" for i in range(1, type_.scale)]) + "1"), rounding=ROUND_HALF_UP
```
kbendick, line 250: Question: I saw in another PR some assertions on the python type major version (checking for 3).

Are we supporting Python 2.7? If so, do we need to be concerned about differences between `range` semantics between the two?

python/src/iceberg/conversions.py
samredai: It looks like you're using your locally installed version of mypy which I'm guessing is earlier than version `0.920` which is when singledispatch was added to the whitelist. The `tox` suite shouldn't run into this since it uses the latest version of mypy (currently `0.931`).

Also although `@singledispatch` was added in python 3.4, `@singledispatchmethod` wasn't added until python 3.8. Fortunately it was a very small addition and it was back-ported to python 3.7, so you'll just have to add `singledispatch` to the `deps` in the tox.ini file and the pyproject.toml file.

python/src/iceberg/conversions.py
samredai: No we're definitely not supporting Python 2 since it's been officially deprecated for quite some time. Currently we're aiming to support 3.7+.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+
+@from_partition_value_to_py.register(FloatType)
+@from_partition_value_to_py.register(DoubleType)
+def _(type_, partition_value_as_str: str):
+    return float(partition_value_as_str)
+
+
+@from_partition_value_to_py.register(StringType)
+def _(type_, partition_value_as_str: str):
+    return partition_value_as_str
```
emkornfield, line 103: does this need to check for `partition_value_as_str == HIVE_NULL`?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+
+@to_bytes.register(IntegerType)
+@to_bytes.register(DateType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack("<i", value)
+
+
+@to_bytes.register(LongType)
+@to_bytes.register(TimeType)
+@to_bytes.register(TimestampType)
```
emkornfield, line 150: Is accepting values as `datetime.time` or `datetime.datetime` something that is handled at a higher level?  Or considered out of scope for the library, or something that wouldn't apply here?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+@to_bytes.register(TimestamptzType)
+def _(type_, value: Any) -> bytes:
```
emkornfield, line 152: I'm not sure if it works with `@singledispatch` but does it make sense to narrow the type of assumed values?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+
+@to_bytes.register(UUIDType)
+def _(type_, value: Any) -> bytes:
+    return struct.pack(">QQ", (value.int >> 64) & 0xFFFFFFFFFFFFFFFF, value.int & 0xFFFFFFFFFFFFFFFF)
+
+
+@to_bytes.register(BinaryType)
+@to_bytes.register(FixedType)
+def _(type_, value: Any) -> bytes:
+    return value
```
emkornfield, line 179: my understanding is type checking is only compile time checks.  It might make sense to validate the type of value here is actually bytes to provide a clearer error message to users.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+@to_bytes.register(DecimalType)
+def _(type_, value: Any) -> bytes:
+    scale = abs(value.as_tuple().exponent)
```
emkornfield, line 184: does there need to be conversion from the implied scale of value and the scale specified in type?

python/tests/test_conversions.py
```diff
@@ -0,0 +1,181 @@
...
+
+@pytest.mark.parametrize(
+    "type_, partition_value_as_str, expected_result",
+    [
+        (DecimalType(5, 2), "123.45", Decimal(123.45)),
+    ],
+)
+def test_from_partition_value_to_py_approximated(type_, partition_value_as_str, expected_result):
+    """Test approximate equality when converting a partition value to a python built-in"""
+    assert conversions.from_partition_value_to_py(type_, partition_value_as_str) == pytest.approx(expected_result)
```
emkornfield, line 76: could this be made non-approx if Decimal("123.45") was used?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+
+@from_bytes.register(BinaryType)
+@from_bytes.register(FixedType)
+def _(type_, b: bytes):
+    return b
+
+
+@from_bytes.register(DecimalType)
+def _(type_, b: bytes):
+    return Decimal(int.from_bytes(b, "big", signed=True) * 10**-type_.scale).quantize(
```
emkornfield, line 249: Does this work for something like Decimal(38, 37)?  It seems that using the [tuple constructor](https://docs.python.org/3/library/decimal.html#decimal-objects) would be more robust?  Or `Decimal(int.from_bytes(...)) * Decimal(10 ** -type.scale)` (a context might need to be passed in for this option)?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,251 @@
...
+    TimestamptzType,
+    TimeType,
+    UUIDType,
+)
+
+HIVE_NULL = "__HIVE_DEFAULT_PARTITION__"
+
+
+@singledispatch
+def from_partition_value_to_py(type_, partition_value_as_str: str):
```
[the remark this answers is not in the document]
jun-he: BTW, also is it feasible to put type annotation for `type_`?

python/src/iceberg/conversions.py
samredai: Yes I think you're right. I believe when a hive catalog is used, `"__HIVE_DEFAULT_PARTITION__"` can be the value for records with nulls in the partition column, regardless of the partition column type. I'm going to move this check out to it's own function and perform this check regardless of the partition type.

emkornfield: I think this function might be the only other place that can actually be dispatched based on type.

python/src/iceberg/conversions.py
samredai: Yes that should work, `@singledispatch` is only concerned with the first args type. I'll narrow the type hints! 👍

python/src/iceberg/conversions.py
samredai: If we narrow the `value` argument's type hint and include type hints for the higher level objects, that should implicitly provide robust mypy validation that `value` is of type bytes if `type_` is `BinaryType` or `FixedType`. If we do need to validate the value types I believe that would be better in the higher level objects (since `foo.type=BinaryType` and `foo.value=<not bytes>` would indicate incorrect logic somewhere).

python/tests/test_conversions.py
samredai: You're absolutely right, I'll change that and remove this approximation test.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,181 @@
...
+        (DateType(), b"\xd2\x04\x00\x00", 1234),
+        (TimeType(), b"\x00\xe8vH\x17\x00\x00\x00", 100000000000),
+        (TimestamptzType(), b"\x00\xe8vH\x17\x00\x00\x00", 100000000000),
+        (TimestampType(), b"\x00\xe8vH\x17\x00\x00\x00", 100000000000),
+        (StringType(), b"foo", "foo"),
+        (UUIDType(), b"\xf7\x9c>\tg|K\xbd\xa4y?4\x9c\xb7\x85\xe7", uuid.UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7")),
+        (FixedType(3), b"foo", b"foo"),
+        (BinaryType(), b"foo", b"foo"),
+        (DecimalType(5, 2), b"\x30\x39", Decimal(123.45).quantize(Decimal(".01"))),
+        (DecimalType(5, 4), b"\x00\x12\xd6\x87", Decimal(123.4567).quantize(Decimal(".0001"))),
```
emkornfield, line 95: similar comment here, I think it might be clearer to use string constructor on Decimal.

python/src/iceberg/conversions.py
samredai: I'm not 100% sure on the requirement here. If a value is provided with a higher scale than the defined type, it feels like just converting it might have implications when lower/upper bound values are deserialized later. I'll need to dig in a bit more here.

emkornfield: I think given that fact that we expect this to be used from literals, it seems like the normalize should probably happen at that level, and here there should be an assert that the scales are equal.

python/src/iceberg/conversions.py
samredai: Ah right, this would run into an issue there. I like the idea to use the tuple constructor and provide a context. Something like:
```py
@from_bytes.register(DecimalType)
def _(type_, b: bytes):
    integer_representation = int.from_bytes(b, "big", signed=True)
    sign = 1 if integer_representation < 0 else 0
    integer_tuple = tuple(int(i) for i in str(abs(integer_representation)))
    return Decimal(value=(sign, integer_tuple, -type_.scale), context=Context(prec=type_.precision, rounding=ROUND_HALF_UP))
```

python/src/iceberg/conversions.py
emkornfield: Also, to be clear, I'm not super familiar with Hive partitioning, so this really was a question which I don't know the answer to.

python/src/iceberg/conversions.py
emkornfield: I think the separate Decimal option might be more slightly more efficient because it doesn't require the string coversion, if it works.

python/src/iceberg/conversions.py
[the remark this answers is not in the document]
samredai: Agreed, I'll rename `type_` to `primitive_type`, that's much clearer. For adding the type annotation, this should be feasible in that I don't think it will break the single dispatch decorator. Are you thinking we hint using the base primitive type class like `def from_partition_value_to_py(type_: PrimitiveType, partition_value_as_str: str):`?

python/src/iceberg/conversions.py
samredai: I see, ok let me try that!

python/src/iceberg/conversions.py
samredai: It looks like in the java client this is really just for tests. I'm going to exclude this here but if it ends up being required it makes more sense to add it as part of the hive catalog implementation.

python/tests/test_conversions.py
samredai: Converted this everywhere in the tests so quantize isn't used anymore.

python/src/iceberg/conversions.py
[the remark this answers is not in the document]
samredai: Updated! Thanks @jun-he, this looks much cleaner.

python/src/iceberg/conversions.py
samredai: Updated the `to_bytes(...)` and `from_bytes(...)` methods dispatched for `DecimalType`. I use the decimal option for from_bytes and I added the scale and precision checks in to_bytes.

python/tests/test_conversions.py
samredai: Updated!

python/src/iceberg/conversions.py
samredai: Added a check and raise for mismatched precision and/or scale.

python/src/iceberg/conversions.py
samredai: Updated the type hints!

python/src/iceberg/conversions.py
samredai: Removed this.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,261 @@
...
+
+@to_bytes.register(DecimalType)
+def _(primitive_type, value: Decimal) -> bytes:
+    value_as_tuple = value.as_tuple()
+    value_precision = len(value_as_tuple.digits)
+    value_scale = -value_as_tuple.exponent
+
+    if value_scale != primitive_type.scale:
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {primitive_type}: {value_scale}")
+    elif value_precision != primitive_type.precision:
```
emkornfield, line 190: Isn't it valid for value to have less digits then precision.

e.g. how is something like "1234.0" handled for Decimal(6, 1)?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,261 @@
...
+def _(primitive_type, b: bytes):
+    return b
+
+
+@from_bytes.register(DecimalType)
+def _(primitive_type, b: bytes):
+    integer_representation = int.from_bytes(b, "big", signed=True)
+    with localcontext() as ctx:
+        ctx.prec = primitive_type.precision
+        return Decimal(integer_representation) * Decimal(10) ** Decimal(-primitive_type.scale)
```
emkornfield, line 261:
```suggestion
        return Decimal(integer_representation) * (Decimal(10) ** Decimal(-primitive_type.scale))
```
Lets try to avoid relying on order of operations by python.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,288 @@
...
+        ),
+        (
+            DecimalType(36, 34),
+            Decimal("1.23456789123456789123456789123456789"),
+            "Cannot serialize value, scale of value does not match type decimal(36, 34): 35",
+        ),
+        (
+            DecimalType(7, 2),
+            Decimal("1234.56"),
+            "Cannot serialize value, precision of value does not match type decimal(7, 2): 6",
```
emkornfield, line 270: To follow-up on my question on the implementation. How can a user specifying this value?  It seems like it should only be an error if precision is exceeded (as you test for below).

python/tests/test_conversions.py
```diff
@@ -0,0 +1,288 @@
...
+    assert value_from_bytes == result
+
+    bytes_from_value = conversions.to_bytes(primitive_type, value_from_bytes)
+    assert bytes_from_value == b
+
+
+@pytest.mark.parametrize(
+    "primitive_type, b, result",
+    [
+        (FloatType(), b"\x19\x04\x9e?", 1.2345),
```
emkornfield, line 218: do these fail if they aren't approximated.  I would think that serializion to and from bytes should be deterministic and not subject to numeric inaccuracies.

python/src/iceberg/conversions.py
samredai: Yes that's totally valid and the spec explicitly says that the precision for a decimal column can be widened. I'll change this to just a check for precision greater than specified by the type. Thanks for catching that!

python/tests/test_conversions.py
samredai: I kept this consistent with the python legacy client which uses the built-in float and double types and just [tests](https://github.com/apache/iceberg/blob/master/python_legacy/tests/api/test_conversions.py#L48#L51) for approximation. According to the Iceberg spec though the float and double primitive types are supposed to be 32-bit and 64-bit IEEE floating points, so the easiest way I think is to use numpy here which was proposed by @CircArgs at some point.

@dramaticlly this would also require that numpy is used in PR #4262 for `FloatLiteral` and `DoubleLiteral`.

emkornfield: Could you elaborate on why is numpy needed here?  

I expect python equality to work, since this is a deterministic round-trip of the data (is it something in the testing framework):
```
>>> 1.2345 == 1.2345
True
```

python/src/iceberg/conversions.py
samredai: Fixed!

python/tests/test_conversions.py
samredai: Replied [here](https://github.com/apache/iceberg/pull/4263#discussion_r822026954) and fixed!

python/tests/test_conversions.py
```diff
@@ -0,0 +1,288 @@
...
+        (DecimalType(23, 18), b"\x02\x9d;\x00\xf8\xae\xb2\x14_\x15", Decimal("12345.123456789123456789")),
+        (DecimalType(24, 19), b"\x1a$N\t\xb6\xd2\xf4\xcb\xb6\xd3", Decimal("12345.1234567891234567891")),
+        (DecimalType(25, 20), b"\x01\x05k\x0ca$=\x8f\xf5$@", Decimal("12345.12345678912345678912")),
+        (DecimalType(26, 21), b"\n6.{\xcbjg\x9f\x93j\x83", Decimal("12345.123456789123456789123")),
+        (DecimalType(27, 22), b'f\x1d\xd0\xd5\xf2(\x0c;\xc2)"', Decimal("12345.1234567891234567891234")),
+        (DecimalType(28, 23), b"\x03\xfd*([u\x90zU\x95\x9bY", Decimal("12345.12345678912345678912345")),
+        (DecimalType(29, 24), b"'\xe3\xa5\x93\x92\x97\xa4\xc7W\xd8\x11\x80", Decimal("12345.123456789123456789123456")),
+        (DecimalType(30, 25), b"\x01\x8e\xe4w\xc3\xb9\xeco\xc9np\xaf\x07", Decimal("12345.1234567891234567891234567")),
+        (DecimalType(31, 26), b"\x0f\x94\xec\xad\xa5C<]\xdePf\xd6N", Decimal("12345.12345678912345678912345678")),
+    ],
```
emkornfield, line 204: It would be nice to test Min/Max values that a particular precision/scale can hold.  i.e. "99999.99999", "-99999.99999"

python/tests/test_conversions.py
samredai: Yes equality works but the imprecision surfaces when converting the float to a c struct and back:
```py
import struct

b = struct.pack("<f", 1.2345)
value = struct.unpack("<f", b)  # 1.2345000505447388
```
and `1.2345 == 1.2345000505447388` returns `False`

python/tests/test_conversions.py
samredai: Great idea, I'll add that in.

python/tests/test_conversions.py
emkornfield: Interesting, I dug bit further and turned up [this old bug report](https://bugs.python.org/issue4114) which has a couple of solutions to round-trip deterministically (but involves truncation.  I think this is only an issue for Floats and Doubles should round-trip deterministically. I assume the byte values are taken from Java?

samredai: There will eventually be a write path from python so the byte values stored in the metadata could potentially be from python or Java. But I'd expect today it's almost entirely byte values generated in Java. As the python library matures we'll eventually have some cross-language tests to make sure the behaviors are synced (to_bytes from Java -> from_bytes from python, for example).

That bug report was very helpful, I'm going to try out some of the examples and generally look into this some more. Hopefully there's a simple workaround but I would also be ok with putting a pin in this here and addressing it when we have some of the higher level objects and can measure the impact.

emkornfield: Yeah, I don't know that there is a great solution for float, so it is ok to bypass it.  I think double should exactly round-trip.  If it doesn't I need to further refine my understanding.

samredai: Oh right, yes double does exactly round trip. I moved the double test out of the approximation test function to the equality test function.

python/tests/test_conversions.py
samredai: Added `test_max_value_round_trip_conversion` and `test_min_value_round_trip_conversion`.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,324 @@
...
+def test_max_value_round_trip_conversion(primitive_type, expected_max_value):
+    """Test round trip conversions of maximum DecimalType values"""
+    b = conversions.to_bytes(primitive_type, expected_max_value)
+    value_from_bytes = conversions.from_bytes(primitive_type, b)
+
+    assert value_from_bytes == expected_max_value
+
+
+@pytest.mark.parametrize(
+    "primitive_type, expected_max_value",
```
emkornfield, line 250: nit: min_value here

emkornfield: and for parameter name.

emkornfield (approved):

python/tests/test_conversions.py
```diff
@@ -0,0 +1,288 @@
...
+)
+def test_from_bytes(primitive_type, b, result):
+    """Test converting from bytes"""
+    assert conversions.from_bytes(primitive_type, b) == result
+
+
+@pytest.mark.parametrize(
+    "primitive_type, b, approximate_result, approximation",
+    [
+        (FloatType(), b"\x19\x04\x9e?", 1.2345, 5),
```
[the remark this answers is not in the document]
samredai: Good idea, I added a comment here and also in the `from_bytes(...)` function that's dispatched for `FloatType`. Let me know if it looks accurate and clear enough for if we have to revisit this in the future.

python/tests/test_conversions.py
samredai: Ah thanks for catching that, fixed!

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+@from_bytes.register(TimestampType)
+@from_bytes.register(TimestamptzType)
+def _(primitive_type, b: bytes):
+    return struct.unpack("<q", b)[0]
+
+
+@from_bytes.register(FloatType)
+def _(primitive_type, b: bytes):
+    """
+    Note: float in python is implemented using a double in C. Therefore this involves a conversion of a 32-bit (single precision)
```
emkornfield, line 235: I think this comment belongs on to_bytes, since the actual precision loss is going from double->float.  Float can be converted without loss to double.

samredai: Ah right, that makes sense. Ok I'll move this there!

samredai: Thinking about this some more, should this be described in both `to_bytes` and `from_bytes`? My understanding is that although the precision is lost when going `python float(64-bit) -> c float(32-bit)`, there's some imprecision that happens when going `c float(32-bit) -> python float(64-bit)`.
```c
int main(void) {
  int i;
  union {
    float f;
    unsigned char bytes[4];
  } packed;

  packed.f = 1.234;
  for (i=0; i<4; i++)
    printf ("\\x%02x", packed.bytes[i]);
  printf("\n");
}
```
output:
```
\xb6\xf3\x9d\x3f
```
```py
import struct

print(struct.unpack("<f", b"\xb6\xf3\x9d\x3f")[0])
```
output:
```
1.2339999675750732
```

emkornfield: I guess it depends on your definition of precision. This arises because inability to actually represent 1.234 in ieee floats, so the  approximation differs between types.  This [stack-overflow post](https://stackoverflow.com/questions/17504833/why-converting-from-float-to-double-changes-the-value) I think gives a decent explanation.

```
>>> f"{1.2339999675750732:f}"
'1.234000'
>>> float(f"{1.2339999675750732:f}")
1.234
```

The only actual differences that I'm aware of between float and double is the bit-width for mantissa and exponent, both of which are strictly greater for double so the conversion should always be lossless, so in this case it would be precise.  How humans interpret that lossless conversion is interesting, and probably relevant for how filters are handled.

Printing doubles has some [research behind it](https://cseweb.ucsd.edu/~lerner/papers/fp-printing-popl16.pdf).  This is also why when people want to track currency, they use Decimal and not floats.

samredai: Great, thanks for the explanation. That makes sense!

samredai: Moved the note to `to_bytes`.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,329 @@
...
+        (DateType(), b"\xd2\x04\x00\x00", 1234),
+        (TimeType(), b"\x00\xe8vH\x17\x00\x00\x00", 100000000000),
+        (TimestamptzType(), b"\x00\xe8vH\x17\x00\x00\x00", 100000000000),
+        (TimestampType(), b"\x00\xe8vH\x17\x00\x00\x00", 100000000000),
+        (StringType(), b"foo", "foo"),
+        (UUIDType(), b"\xf7\x9c>\tg|K\xbd\xa4y?4\x9c\xb7\x85\xe7", uuid.UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7")),
+        (FixedType(3), b"foo", b"foo"),
+        (BinaryType(), b"foo", b"foo"),
+        (DecimalType(5, 2), b"\x30\x39", Decimal("123.45")),
+        (DecimalType(7, 4), b"\x00\x12\xd6\x87", Decimal("123.4567")),
```
cccs-eric, line 85: Suggestion: when I migrated the Java tests to python_legacy, I really appreciated the test comments as it helped me understand what I was actually testing.  For some tests, it might not always be obvious.  For example, the test with `Decimal("123.4567")` is not a random value, but a value that stored on 3 bytes, which is not a power of 2 (which is usually used for int/long, etc...).  The Iceberg spec states to use the minimum number of bytes to store the value.  

It doesn't add value to the code, but it can be very helpful for a new dev when things go sideways...  Again, just a suggestion since I found it useful when I had to play in that code.

cccs-eric (approved):
LGTM 👍🏻

python/tests/test_conversions.py
samredai: This is a great catch, I'll update this PR by bringing over those test comments and maybe add some additional ones. Thanks @cccs-eric !

samredai: I brought over all of the comments and organized them in a module-level docstring in `test_conversions.py`. I also brought over additional tests from the legacy library.

cccs-eric: Nicely done.  Looks pro! 😛

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+def _(primitive_type, partition_value_as_str: str):
+    if partition_value_as_str is None:
+        return False
+    return partition_value_as_str.lower() == "true"
+
+
+@from_partition_value_to_py.register(IntegerType)
+@from_partition_value_to_py.register(LongType)
+def _(primitive_type, partition_value_as_str: str):
+    return int(float(partition_value_as_str))
```
rdblue, line 91: Do these also need checks for `None`?

rdblue: Converting to float and then to int allows values through that would fail in Java.

I think we want to be strict here as well. If a value has a decimal point and isn't an integer, I think it is better to fail. This is only used in narrow cases anyway. In fact, I thought about saying we should remove these methods entirely. Let's leave them in, but be strict about the conversions.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@from_partition_value_to_py.register(UUIDType)
+def _(primitive_type, partition_value_as_str: str):
+    return uuid.UUID(partition_value_as_str)
+
+
+@from_partition_value_to_py.register(FixedType)
+def _(primitive_type, partition_value_as_str: str):
+    return bytearray(bytes(partition_value_as_str, "UTF-8"))
```
rdblue, line 112: I think we will use `bytes` instead of `bytearray`. So we should combine this with `BinaryType` like `IntType` and `LongType`.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@from_partition_value_to_py.register(DecimalType)
+def _(primitive_type, partition_value_as_str: str):
+    return Decimal(partition_value_as_str)
+
+
+@singledispatch
+def to_bytes(primitive_type, value) -> bytes:
+    """A generic function which converts a built-in python value to bytes
```
rdblue, line 127: I think it would be good to refer to the spec here, since these implement single-value serialization: https://iceberg.apache.org/spec/#appendix-d-single-value-serialization

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+    StringType,
+    TimestampType,
+    TimestamptzType,
+    TimeType,
+    UUIDType,
+)
+
+
+@singledispatch
+def from_partition_value_to_py(primitive_type, partition_value_as_str: str):
```
rdblue, line 67: Should this use `PrimitiveType` for the type argument? Or does that not matter because it uses `@singledispatch`?

python/src/iceberg/conversions.py
rdblue: The `to_bytes` conversion below also uses bytes, so that will be more consistent as well.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+
+    if value_scale != primitive_type.scale:
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {primitive_type}: {value_scale}")
+    elif value_precision > primitive_type.precision:
+        raise ValueError(
+            f"Cannot serialize value, precision of value is greater than precision of type {primitive_type}: {value_precision}"
+        )
+
+    with localcontext() as ctx:
+        ctx.prec = primitive_type.precision
```
rdblue, line 200: I think this needs to be `primitive_type.precision + primitive_type.scale`.

Say I have a decimal(4,2) value, 12.34. The unscaled value is 12.34 with precision 4. After multiplying by 10**2, the intermediate decimal value becomes 1234.00, which precision 6. So if the context's precision is limited then that could be a problem because the multiplication temporarily creates a value outside the precision.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@to_bytes.register(BinaryType)
+@to_bytes.register(FixedType)
+def _(primitive_type, value: bytes) -> bytes:
+    return value
+
+
+@to_bytes.register(DecimalType)
+def _(primitive_type, value: Decimal) -> bytes:
```
rdblue, line 187: Other than the precision issue below, I think that this is a correct conversion. We may want to break this up into other decimal functions for use elsewhere. For example, getting the unscaled value might be useful for the truncate transform. Also, the hash transform should call this method! (FYI @jun-he)

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+
+    with localcontext() as ctx:
+        ctx.prec = primitive_type.precision
+        unscaled_value = int((value * 10**primitive_type.scale).to_integral_value(rounding=ROUND_HALF_DOWN))
+        min_num_bytes = ((unscaled_value).bit_length() + 7) // 8
+        return unscaled_value.to_bytes(min_num_bytes, "big", signed=True)
+
+
+@singledispatch
+def from_bytes(primitive_type, b: bytes):
```
rdblue, line 207: Minor: Does Python encourage short variable names like `b`? I think most languages discourage single-character names.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@from_bytes.register(StringType)
+def _(primitive_type, b: bytes):
+    return bytes(b).decode("utf-8")
+
+
+@from_bytes.register(UUIDType)
+def _(primitive_type, b: bytes):
+    return uuid.UUID(int=struct.unpack(">QQ", b)[0] << 64 | struct.unpack(">QQ", b)[1])
```
rdblue, line 253: This runs `unpack` on the bytes twice. It would be better to use a separate variable to avoid doing the work twice.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+def _(primitive_type, b: bytes):
+    return b
+
+
+@from_bytes.register(DecimalType)
+def _(primitive_type, b: bytes):
+    integer_representation = int.from_bytes(b, "big", signed=True)
+    with localcontext() as ctx:
+        ctx.prec = primitive_type.precision
+        return Decimal(integer_representation) * (Decimal(10) ** Decimal(-primitive_type.scale))
```
rdblue, line 267: I think this should construct the multiplier like the [handling in the `literals.py` PR](https://github.com/apache/iceberg/pull/4262/files#r832771258). That's a more direct way to create the `0.001` value:

```python
Decimal(integer_representation) * Decimal((0, (1,), -primitive_type.scale))
```

python/tests/test_conversions.py
```diff
@@ -0,0 +1,422 @@
...
+        (IntegerType(), "9999", 9999),
+        (LongType(), "123456789", 123456789),
+        (FloatType(), "1.1", 1.1),
+        (DoubleType(), "99999.9", 99999.9),
+        (DecimalType(5, 2), "123.45", Decimal("123.45")),
+        (StringType(), "foo", "foo"),
+        (UUIDType(), "f79c3e09-677c-4bbd-a479-3f349cb785e7", uuid.UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7")),
+        (FixedType(3), "foo", bytearray(b"foo")),
+        (BinaryType(), "foo", b"foo"),
+        (None, None, None),
```
rdblue, line 116: You may want a separate test that validates for all primitive types, None produces None.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,422 @@
...
+        (BooleanType(), None, False),
+        (IntegerType(), "1", 1),
+        (IntegerType(), "9999", 9999),
+        (LongType(), "123456789", 123456789),
+        (FloatType(), "1.1", 1.1),
+        (DoubleType(), "99999.9", 99999.9),
+        (DecimalType(5, 2), "123.45", Decimal("123.45")),
+        (StringType(), "foo", "foo"),
+        (UUIDType(), "f79c3e09-677c-4bbd-a479-3f349cb785e7", uuid.UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7")),
+        (FixedType(3), "foo", bytearray(b"foo")),
```
rdblue, line 114: This should be bytes

python/src/iceberg/conversions.py
rdblue: Actually, I think the min_num_bytes calculation is incorrect.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,422 @@
...
+            bytes([247, 156, 62, 9, 103, 124, 75, 189, 164, 121, 63, 52, 156, 183, 133, 231]),
+            uuid.UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7"),
+        ),
+        (UUIDType(), b"\xf7\x9c>\tg|K\xbd\xa4y?4\x9c\xb7\x85\xe7", uuid.UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7")),
+        (FixedType(3), bytes([97, 98]), bytes("ab", "utf8")),
+        (FixedType(3), b"foo", b"foo"),
+        (BinaryType(), bytearray("Z", "utf8"), bytes([90])),
+        (BinaryType(), b"foo", b"foo"),
+        (DecimalType(5, 2), b"\x30\x39", Decimal("123.45")),
+        (DecimalType(7, 4), b"\x00\x12\xd6\x87", Decimal("123.4567")),
```
rdblue, line 153: I don't think this is correct. The number of bytes should be 3, not 4, because null bytes are omitted.

The calculation for min bytes is:
```python
min_num_bytes = ((unscaled_value).bit_length() + 7) // 8
```

That translates to:
```python
((1234567).bit_length() + 7) // 8
=> 3
```

So the calculation looks correct. I'm wondering why the byte array ends up with the extra null byte at the start.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,422 @@
...
+    ],
+)
+def test_from_bytes_approximately(primitive_type, b, approximate_result, approximation):
+    """Test approximate equality when converting from bytes
+
+    Note: This tests approximate equality because floating point numbers in python are implemented
+    using a double in C. Therefore a 32-bit (single precision) float is unpacked to 64-bit (double precision)
+    float in python which introduces some imprecision.
+    """
+    assert conversions.from_bytes(primitive_type, b) == pytest.approx(approximate_result, approximation)
```
rdblue, line 175: I don't think this should be an approximate result. Instead, can you send the expected result through pack/unpack to produce the same 32-bit value?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+        return None
+    raise TypeError(
+        f"Cannot convert partition string to python built-in, type {primitive_type} not supported: '{partition_value_as_str}'"
+    )
+
+
+@from_partition_value_to_py.register(BooleanType)
+def _(primitive_type, partition_value_as_str: str):
+    if partition_value_as_str is None:
+        return False
```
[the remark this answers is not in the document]
emkornfield: In this case since it is very constrained values it could be but the general recommendation is to use [direct comparisons with None](https://peps.python.org/pep-0008/#programming-recommendations) if that is actually your intent.

python/tests/test_conversions.py
emkornfield: This is just a test issue.  This test is from_bytes (it isn't verifying round-trip) so the extra bytes don't impact the tests because they the extra padding can still be converted.  The to_bytes tests look like they have the correct lengths.

python/src/iceberg/conversions.py
emkornfield: I don't think this is the case.  I think precision here refers to significant digits sense, not in the DB sense (probably worth a comment):
```
>>> a = Decimal("123.45")
>>> with localcontext() as ctx:
...   ctx.prec = 1
...   a * 10 ** 2
... 
Decimal('1E+4')

>>> with localcontext() as ctx:
...   ctx.prec = 5
...   a * 10 ** 2
... 
Decimal('12345')
```

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,267 @@
...
+    if value_scale != primitive_type.scale:
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {primitive_type}: {value_scale}")
+    elif value_precision > primitive_type.precision:
+        raise ValueError(
+            f"Cannot serialize value, precision of value is greater than precision of type {primitive_type}: {value_precision}"
+        )
+
+    with localcontext() as ctx:
+        ctx.prec = primitive_type.precision
+        unscaled_value = int((value * 10**primitive_type.scale).to_integral_value(rounding=ROUND_HALF_DOWN))
```
emkornfield, line 201: nit, I don't think rounding should matter here since there shouldn't be anything after the decimal point?

python/src/iceberg/conversions.py
emkornfield: probably adding a comment to avoid confusion makes sense.

python/src/iceberg/conversions.py
samredai: Thanks for catching this! I added some logic to raise if there are any non-zero fractional digits. ([here](https://github.com/apache/iceberg/pull/4263/files#diff-23863ae95c4894ac0025efabd97dd925503dc5a908e386d7c345952e36b9743cR134)):
```py
@from_partition_value_to_py.register(IntegerType)
@from_partition_value_to_py.register(LongType)
def _(primitive_type, partition_value_as_str: str):
    """
    Note: Before attempting to cast the value to string, a validation happens to ensure that there are no fractional digits. For
    example, an invalid value such as "123.45" will be converted to a decimal with digits (1, 2, 3, 4, 5) with an exponent of -2 and the last
    2 digits will be inspected for non-zero values. An example of a valid value is "123.00".
    """
    if partition_value_as_str is None:
        return None
    _, digits, exponent = Decimal(partition_value_as_str).as_tuple()
    if exponent != 0 and any(digits[exponent:]):  # If there are digits to the right of the exponent, raise if any are not 0
        raise ValueError(f"Cannot convert partition value, value cannot have fractional digits for {primitive_type} partition")
    return int(float(partition_value_as_str))
```

python/src/iceberg/conversions.py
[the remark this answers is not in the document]
samredai: I made an assumption here that `None` is all we need to explicitly check for. Thinking about this more for the other functions, how do we want to handle a `StringType()` or `IntegerType()` partition that has a partition value of an empty string? Should that also return `None`?

python/src/iceberg/conversions.py
samredai: Fixed [here](https://github.com/apache/iceberg/pull/4263/files#diff-23863ae95c4894ac0025efabd97dd925503dc5a908e386d7c345952e36b9743cR164), thanks!

python/src/iceberg/conversions.py
samredai: Great idea, added this to the docstring:
```
This conversion follows the serialization scheme for storing single values as individual binary values defined in the Iceberg specification that
can be found at https://iceberg.apache.org/spec/#appendix-d-single-value-serialization
```

python/src/iceberg/conversions.py
samredai: Hmm, I think in some cases it may not matter but in this case since all of the registered functions inherit from the same superclass I think it's a very useful type-hint. Added it [here](https://github.com/apache/iceberg/pull/4263/files#diff-23863ae95c4894ac0025efabd97dd925503dc5a908e386d7c345952e36b9743cR175).

I also added a type-hint for the `value` argument that describes an aggregate of all of the types possible for the registered functions. Another thing is I noticed I was missing a lot of return type-hints so I added those as well. 👍

python/src/iceberg/conversions.py
samredai: I agree with @emkornfield in this particular case no precision is lost since the significant figures aren't changing. However I think adjusting the scale in a more direct way avoids this problem entirely and shouldn't require any additional notes. This method recreates the decimal using the same sign and digits and just adjusts the exponent/scale to the new value.
```py
    with localcontext() as ctx:
        ctx.prec = decimal_type_var.precision
        sign, digits, exponent = value.as_tuple()
        value_w_adjusted_scale = Decimal((sign, digits, exponent + decimal_type_var.scale))
        return int(value_w_adjusted_scale.to_integral_value())
```
Let me know what ya'll think.

python/src/iceberg/conversions.py
samredai: Broke this out into a `convert_decimal_to_unscaled_value(...)` method at the top of the module.

python/src/iceberg/conversions.py
samredai: In general no but I think `b` for bytes and `f` for file are the two exceptions. Also `bytes` is a reserved keyword, I'm not opposed to changing this though, is there a better name here?

python/src/iceberg/conversions.py
samredai: Ah I see, updated it to avoid the repeated unpack. Thanks! ([here](https://github.com/apache/iceberg/pull/4263/files#diff-23863ae95c4894ac0025efabd97dd925503dc5a908e386d7c345952e36b9743cR292))

python/src/iceberg/conversions.py
samredai: Updated [here](https://github.com/apache/iceberg/pull/4263/files#diff-23863ae95c4894ac0025efabd97dd925503dc5a908e386d7c345952e36b9743cR307).

python/tests/test_conversions.py
samredai: Add a [test_none_partition_values](https://github.com/apache/iceberg/pull/4263/files#diff-704ab9ed1661567f3b0eb10979903d6a9eb69efa2a76d099ed517ca19e5889beR143) test and it highlighted that a few weren't returning None for None so I updated those functions as well.

python/tests/test_conversions.py
samredai: Fixed!

python/tests/test_conversions.py
samredai: Ah right, added the pack/unpack and it works without any approximation.

python/src/iceberg/conversions.py
samredai: Good point, removed the rounding. Thanks!

python/tests/test_conversions.py
samredai: Thanks everyone, I removed the extra byte and also changed the round trip to use bytes instead of bytearray just for consistency.

samredai:
@rdblue @emkornfield I appreciate the reviews! I think I addressed everything and I also rebased with the changes from the recently merged doctests PR.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,294 @@
...
+    if partition_value_as_str is None:
+        return None
+    return Decimal(partition_value_as_str)
+
+
+@singledispatch
+def to_bytes(primitive_type: PrimitiveType, value: Union[bool, bytes, Decimal, float, int, str, uuid.UUID]) -> bytes:
+    """A generic function which converts a built-in python value to bytes
+
+    This conversion follows the serialization scheme for storing single values as individual binary values defined in the Iceberg specification that
```
rdblue, line 165: Do we need to wrap lines in docstrings like this? It seems long?

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,294 @@
...
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {decimal_type_var}: {value_scale}")
+    elif value_precision > decimal_type_var.precision:
+        raise ValueError(
+            f"Cannot serialize value, precision of value is greater than precision of type {decimal_type_var}: {value_precision}"
+        )
+
+    with localcontext() as ctx:
+        ctx.prec = decimal_type_var.precision
+        sign, digits, exponent = value.as_tuple()
+        value_w_adjusted_scale = Decimal((sign, digits, exponent + decimal_type_var.scale))
```
rdblue, line 82: When converting using the direct tuple construction, do we need to adjust the local context? According to [this intro to Decimal page](https://www.pythontutorial.net/advanced-python/python-decimal/), "Notice that the decimal context precision only affects the arithmetic operation, not the Decimal constructor".

That means we should just be able to return the adjusted value with scale 0.

Also, since we don't need to set the context precision, we don't need the decimal type at all. In fact, it doesn't make much sense to check the value against the type here either. I think this method could be focused on just getting the unscaled value and the checks can happen in the caller, if needed.

python/src/iceberg/conversions.py
rdblue: I think just directly casting to int is a simpler way to do this. When I remove the inner `float` cast and test it, I get `ValueError: invalid literal for int() with base 10: '12.34'`. That seems correct to me. This method is also _not_ recommended for production use, it is more of an internal convenience.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,294 @@
...
+    PrimitiveType,
+    StringType,
+    TimestampType,
+    TimestamptzType,
+    TimeType,
+    UUIDType,
+)
+
+
+def convert_decimal_to_unscaled_value(decimal_type_var: DecimalType, value: Decimal) -> int:
```
rdblue, line 55: What about a bit simpler name? This could be `decimal_to_unscaled`. The "convert" is implied by being in `conversions` and "value" is unnecessary because this clearly returns a value.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,294 @@
...
+def _(primitive_type, b: bytes) -> bytes:
+    return b
+
+
+@from_bytes.register(DecimalType)
+def _(primitive_type, b: bytes) -> Decimal:
+    integer_representation = int.from_bytes(b, "big", signed=True)
+    with localcontext() as ctx:
+        ctx.prec = primitive_type.precision
+        return Decimal(integer_representation) * Decimal((0, (1,), -primitive_type.scale))
```
rdblue, line 294: One more idea here: `Decimal(integer_representation)` creates a decimal with the right sign and digits. The right scale is `-primitive_type`. So we can avoid doing any decimal arithmetic (and hitting context problems) by using this:

```python
def unscaled_to_decimal(unscaled: int, scale: int) -> Decimal:
    sign, digits, _ = Decimal(unscaled).to_tuple()
    return Decimal((sign, digits, scale))

@from_bytes.register(DecimalType)
def _(decimal_type, buf: bytes) -> Decimal:
    unscaled = int.from_bytes(buf, "big", signed=True)
    return unscaled_to_decimal(unscaled, -primitive_type.exponent)
```

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,294 @@
...
+        return None
+    raise TypeError(
+        f"Cannot convert partition string to python built-in, type {primitive_type} not supported: '{partition_value_as_str}'"
+    )
+
+
+@from_partition_value_to_py.register(BooleanType)
+def _(primitive_type, partition_value_as_str: str):
+    if partition_value_as_str is None:
+        return None
```
rdblue, line 104: Instead of adding this check in every method, could we use an outer function that calls the `@singledispatch` method?

```python
def partition_to_py(primitive_type: PrimitiveType, value_str: str):
    if value_str is None:
        return None
    elif value_str == '__HIVE_DEFAULT_PARTITION__':
        return None
   return _partition_to_py(primitive_type, value_str)

@singledispatch
def _partition_to_py(primitive_type: PrimitiveType, value_str: str):
    raise TypeError("Cannot convert '{value_str}' to unsupported type: {primitive_type}")
```

python/tests/test_conversions.py
```diff
@@ -0,0 +1,442 @@
...
+@pytest.mark.parametrize(
+    "primitive_type, b, result",
+    [
+        (BooleanType(), b"\x00", False),
+        (BooleanType(), b"\x01", True),
+        (IntegerType(), b"\xd2\x04\x00\x00", 1234),
+        (LongType(), b"\xd2\x04\x00\x00\x00\x00\x00\x00", 1234),
+        (DoubleType(), b"\x8d\x97\x6e\x12\x83\xc0\xf3\x3f", 1.2345),
+        (DateType(), bytes([232, 3, 0, 0]), 1000),
+        (DateType(), b"\xd2\x04\x00\x00", 1234),
```
rdblue, line 157: What is the purpose of testing both `bytes` representations? Is this to match existing tests in other places?

python/tests/test_conversions.py
```diff
@@ -0,0 +1,442 @@
...
+def test_max_value_round_trip_conversion(primitive_type, expected_max_value):
+    """Test round trip conversions of maximum DecimalType values"""
+    b = conversions.to_bytes(primitive_type, expected_max_value)
+    value_from_bytes = conversions.from_bytes(primitive_type, b)
+
+    assert value_from_bytes == expected_max_value
+
+
+@pytest.mark.parametrize(
+    "primitive_type, expected_min_value",
```
rdblue, line 343: These don't look like expected min values? They match the expected max cases.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,442 @@
...
+
+    with pytest.raises(TypeError) as exc_info:
+        conversions.from_bytes(FooUnknownType(), b"foo")
+    assert ("Cannot deserialize bytes, type FooUnknownType() not supported: b'foo'") in str(exc_info.value)
+
+
+@pytest.mark.parametrize(
+    "primitive_type, value, expected_error_message",
+    [
+        (DecimalType(7, 3), Decimal("123.4567"), "Cannot serialize value, scale of value does not match type decimal(7, 3): 4"),
```
rdblue, line 385: We can probably omit a few words from the error message here: "scale does not match {decimal_type}"

python/tests/test_conversions.py
```diff
@@ -0,0 +1,442 @@
...
+    "primitive_type, value, should_raise",
+    [
+        (IntegerType(), "123.45", True),
+        (IntegerType(), "1234567.89", True),
+        (IntegerType(), "123.00", False),
+        (IntegerType(), "1234567.00", False),
+        (LongType(), "123.45", True),
+        (LongType(), "1234567.89", True),
+        (LongType(), "123.00", False),
+        (LongType(), "1234567.00", False),
```
rdblue, line 430: I think these should throw exceptions as well. If there is a decimal point, then the type does not match the original type and it should fail.

python/tests/test_conversions.py
```diff
@@ -0,0 +1,442 @@
...
+    ],
+)
+def test_from_partition_value_to_py_raise_on_incorrect_precision_or_scale(primitive_type, value, should_raise):
```
rdblue, line 433: Can you relocate the partition tests so they're grouped together?

python/src/iceberg/conversions.py
samredai: Good catch I would have expected the auto-formatter to chop this line. Fixed!

python/src/iceberg/conversions.py
samredai: Oh wow you're absolutely right. I made those changes and it's much cleaner now.

python/src/iceberg/conversions.py
samredai: Haha yeah that was a pretty ugly function name 😅 . Updated!

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,294 @@
...
+
+@from_bytes.register(BinaryType)
+@from_bytes.register(FixedType)
+def _(primitive_type, b: bytes) -> bytes:
+    return b
+
+
+@from_bytes.register(DecimalType)
+def _(primitive_type, b: bytes) -> Decimal:
+    integer_representation = int.from_bytes(b, "big", signed=True)
```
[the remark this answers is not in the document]
samredai: Sounds good! Updated

python/src/iceberg/conversions.py
samredai: Much cleaner! Switched to this and removed `localcontext` use entirely.

python/src/iceberg/conversions.py
samredai: Another option is to create a simple `@handle_none` decorator and decorate the relevant functions. I think this may be a bit better when navigating the code in the future. When using a "Go to Definition" somehwere `partition_to_py` is used, it will take you straight to the definition without the double hop and I think most people would ignore the `@handle_none` decorator unless it sounds relevant to what they're looking for. (see here and let me know what you think)
```py
def handle_none(func):
    """A decorator function to handle cases where partition values are `None` or "__HIVE_DEFAULT_PARTITION__"

    Args:
        func (Callable): A function registered to the singledispatch function `partition_to_py`
    """

    def wrapper(primitive_type, partition_value_as_str):
        if partition_value_as_str is None:
            return None
        elif partition_value_as_str == "__HIVE_DEFAULT_PARTITION__":
            return None
        return func(primitive_type, partition_value_as_str)

    return wrapper


@partition_to_py.register(BooleanType)
@handle_none
def _(primitive_type, value_str: str) -> Union[int, float, str, uuid.UUID]:
    return value_str.lower() == "true"
```

python/tests/test_conversions.py
samredai: Replaced these with byte strings. I don't think there's any purpose for having both representations or even the tests that validate that the resulting strings are utf-8, these were brought over from the legacy library. I think they may have been relevant at the time because python 2 was still being supported which had some different encoding behaviors compared to python 3+.

python/tests/test_conversions.py
samredai: Fixed!

python/tests/test_conversions.py
samredai: Ah got it, I misunderstood that but it makes sense. Fixed to always raise if there's a decimal point and added some sets that make sure it does not raise when there is no decimal point (as a sanity-check).

python/tests/test_conversions.py
samredai: Done! 👍

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,318 @@
...
+def decimal_to_unscaled(value: Decimal) -> int:
+    """Get an unscaled value given a Decimal value
+
+    Args:
+        value (Decimal): A Decimal instance
+
+    Returns:
+        int: The unscaled value
+    """
+    return int(value.to_integral_value())
```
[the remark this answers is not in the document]
samredai: Thanks for catching that. Pushed a fix and also added `test_decimal_to_unscaled` and `test_unscaled_to_decimal` to test calling those functions directly.

python/src/iceberg/conversions.py
```diff
@@ -0,0 +1,319 @@
...
+    sign, digits, exponent = value.as_tuple()
+
+    if -exponent != primitive_type.scale:
+        raise ValueError(f"Cannot serialize value, scale of value does not match type {primitive_type}: {-exponent}")
+    elif len(digits) > primitive_type.precision:
+        raise ValueError(
+            f"Cannot serialize value, precision of value is greater than precision of type {primitive_type}: {len(digits)}"
+        )
+
+    unscaled_value = decimal_to_unscaled(value=Decimal((sign, digits, 0)))
```
rdblue, line 254: Minor: this could pass `value` now that `decimal_to_unscaled` is fixed.

rdblue (approved):

rdblue:
Thanks, @samredai!

PR #4132, opened 2022-02-15
[Core][Flink][Spark]: Refactor `TaskWriter` implementations

dungdm93:
This PR aim to:
1. replace `BaseTaskWriter` inner classes (`BaseRollingWriter`, `RollingFileWriter`, `RollingEqDeleteWriter`)
by the respective implementations of `RollingFileWriter` interface.
2. provide single implementation of `TaskWriter` that can handle both partition & un-partition data (by delegate to the `PartitioningWriter`)

Here is my approach, from top down:
```txt
TaskWriter
    |
    V
PartitioningWriter
    |
    V
RollingFileWriter
    |
    V
FileWriter
    |
    V
FileAppender
```

1. `TaskWriter` is used to handle diffent kinds of record.
    * With Insert-only data, `TaskWriter` basically call `PartitioningWriter.write`. See [`DirectTaskWriter`](https://github.com/apache/iceberg/blob/18bc27070661c71d5590b3bed39b92a51d6b4c2d/core/src/main/java/org/apache/iceberg/io/DirectTaskWriter.java) for more details.
    * With delta data, `TaskWriter` can have 3 `PartitioningWriter`s for `insertWriter`, `equalityDeleteWriter` and `positionDeleteWriter`. So, for each incomming record, base on its type (insert, update or delete), TaskWriter will call corresponding writer to write data. See [`FlinkTaskWriter`](https://github.com/apache/iceberg/blob/18bc27070661c71d5590b3bed39b92a51d6b4c2d/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkTaskWriter.java) for more details.
2. `PartitioningWriter` is used to write to multiple specs and partitions.
Note that for unpartitioned tables, `partition = null` is passed to `PartitioningWriter.write`
3. Internally, `PartitioningWriter` already use `RollingFileWriter` for rolling to new file if current file is large. `RollingFileWriter` is a just wrapper of other `FileWriter`.
4. `FileWriter` is used to write to single file.

# Why we need that
* The scope that inner classes of bounded to outer object (`BaseTaskWriter` instance) that make the code more complex and hard to read and understand.
* Currently, the approach of separated class for partition & un-partition data is still OK. But it's better if we can unify it into a single class, this mean:
    * Less effort is needed for new engine
[...]

PR #4264, opened 2022-03-04
Flink: Refactor to use the BaseEqualityDeltaWriter.

openinx:
This PR is trying to refactor the flink write path to remove those underlying old writers to the partition specified writers ( Introduced from @aokolnychyi ).

I was reviewing the PR https://github.com/apache/iceberg/pull/4132 from @[dungdm93](https://github.com/dungdm93), which is the similar thing that this one is trying to accomplish.  I tried to rewrite it because I wanted to try out the details of the design for myself and get a better understanding of those new writers. In this way, I think I can provide more background & details design from my perspective.  After high level design are okay, then I think we can make the whole PR  to be smaller ones and let's collaborate together to get this work done @dungdm93 .


core/src/main/java/org/apache/iceberg/deletes/EqualityDeleteWriter.java
```diff
@@ -43,6 +43,7 @@
   private final int[] equalityFieldIds;
   private final SortOrder sortOrder;
   private DeleteFile deleteFile = null;
+  private long rowOffset = 0;
```
dungdm93, line 46: How about name `recordCount` or `rowCount`?
`rowOffset` is specific for tracking record write to file, `recordCount` is more general.

openinx:
FYI @rdblue , @aokolnychyi , @dungdm93 .  I think this draft PR is ready to review now. Mind to take a look when you have a chance ?

core/src/main/java/org/apache/iceberg/io/PartitioningWriter.java
```diff
@@ -46,7 +46,7 @@
    * @param spec a partition spec
    * @param partition a partition or null if the spec is unpartitioned
    */
-  void write(T row, PartitionSpec spec, StructLike partition);
+  PathOffset write(T row, PartitionSpec spec, StructLike partition);
```
openinx, line 49: As the `PartitioningWriter` will have multiple partition writers in its internal implementation,  so we can not just introduce the two methods in this interface: 

```java
  /**
   * Returns the file path that are currently opened.
   *
   * @return the current file path.
   */
  CharSequence location();

  /**
   * Returns the row offset that are currently writing, starting from 0.
   *
   * @return the current row offset.
   */
  long rowOffset();
```

Because those `location()` and `rowOffset()` are binded to specific writers. In theory, different writers should have different `location()` and `rowOffset()`.  So here we add a return value `PathOffset` here to get the latest wrote path & offset for the wrote row.

core/src/main/java/org/apache/iceberg/io/BaseEqualityDeltaWriter.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.StructLike;
+import org.apache.iceberg.deletes.PositionDelete;
+import org.apache.iceberg.relocated.com.google.common.base.Function;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.util.StructLikeMap;
+import org.apache.iceberg.util.StructProjection;
+
+public class BaseEqualityDeltaWriter<T> implements EqualityDeltaWriter<T> {
```
emkornfield, line 33: Are there any style guidelines for documentation?  Typically its nice to have javadoc on al public classes/members that aren't overwritten.

core/src/main/java/org/apache/iceberg/io/BaseEqualityDeltaWriter.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private final ThreadLocal<PositionDelete<T>> posDelete = ThreadLocal.withInitial(PositionDelete::create);
```
emkornfield, line 35: maybe document why thread local is used here?

core/src/main/java/org/apache/iceberg/io/FileWriter.java
```diff
@@ -54,6 +54,20 @@ default void write(Iterable<T> rows) {
    */
   void write(T row);
 
+  /**
+   * Returns the file path that are currently opened.
```
emkornfield, line 58:
```suggestion
   * Returns the file path that is currently opened.
```
Is "opened" important here?  what happens if the writer is closed?

core/src/main/java/org/apache/iceberg/io/FileWriter.java
```diff
@@ -54,6 +54,20 @@ default void write(Iterable<T> rows) {
...
+   *
+   * @return the current file path.
+   */
+  CharSequence location();
+
+  /**
+   * Returns the row offset that are currently writing, starting from 0.
```
emkornfield, line 65:
```suggestion
   * Returns the row offset that will be written to next, starting from 0.
```
Is this interface intended to be seekable at any point?  If not maybe call this rowsWritten which might be clearer?

core/src/main/java/org/apache/iceberg/deletes/EqualityDeleteWriter.java
openinx: I see the `BaseFile` is using the `recordCount`,  I think it's better to keep this consistent with the `BaseFile`'s `recordCount` :-)

core/src/main/java/org/apache/iceberg/io/FileWriter.java
openinx: As we've discussed in https://github.com/apache/iceberg/pull/4264/files#r819238913, I agree `recordCount` is a good & consistent name.

openinx: As we've discussed in https://github.com/apache/iceberg/pull/4264/files#r819238913, I agree `recordCount` is a good & consistent name.

core/src/main/java/org/apache/iceberg/io/BaseEqualityDeltaWriter.java
```diff
@@ -0,0 +1,135 @@
...
+    if (!retireOldKey(keyRefFunc.apply(row), spec, partition)) {
+      equalityWriter.write(row, spec, partition);
+    }
+  }
+
+  @Override
+  public void deleteKey(T key, PartitionSpec spec, StructLike partition) {
+    if (!retireOldKey(asStructLike.apply(key), spec, partition)) {
+      equalityWriter.write(key, spec, partition);
+    }
```
openinx, line 106: I think this will be needed to reconsider because of this PR: https://github.com/apache/iceberg/pull/4364

openinx:
CC @rdblue @aokolnychyi ,  Would you mind to take a look when you have a chance ?

PR #4266, opened 2022-03-04
Core: Convert REST request classes to Immutables

nastra:
@rdblue @kbendick this simplifies/reduces the code in the REST request classes (`CreateNamespaceRequest` / `UpdateNamespacePropertiesRequest`). I figured we first start with those classes before applying similar changes to the [response classes](https://github.com/apache/iceberg/tree/master/core/src/main/java/org/apache/iceberg/rest/responses).

In terms of reviewing I think it makes the most sense to look at the final content of those classes (since the diff is rather noisy because a bunch of stuff is being removed)

### DIFF
--- a/baseline.gradle
+++ b/baseline.gradle
@@ -62,8 +62,8 @@ subprojects {
   pluginManager.withPlugin('com.palantir.baseline-error-prone') {
     tasks.withType(JavaCompile).configureEach {
       options.errorprone.errorproneArgs.addAll (
-          // error-prone is slow, don't run on tests and generated src
-          '-XepExcludedPaths:.*/(test|generated-src)/.*',
+          // error-prone is slow, don't run on tests/generated-src/generated
+          '-XepExcludedPaths:.*/(test|generated-src|generated)/.*',
           // specific to Palantir
           '-Xep:ConsistentLoggerName:OFF',  // Uses name `log` but we use name `LOG`
           '-Xep:FinalClass:OFF',
--- a/core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
@@ -19,81 +19,47 @@
 
 package org.apache.iceberg.rest.requests;
 
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.Map;
-import java.util.Objects;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
-import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
-import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.immutables.value.Value;
 
 /**
  * A REST request to create a namespace, with an optional set of properties.
+ *
+ * Note that Immutable classes by definition don't have a default no-arg constructor (which is required for Jackson),
+ * therefore the @{@link JsonSerialize}/@{@link JsonDeserialize} annotations on the class will generate what's
+ * necessary for Jackson-binding to properly work with this class.
  */
-public class CreateNamespaceRequest {
+@Value.Immutable
+@JsonSerialize(as = ImmutableCreateNamespaceRequest.class)
+@JsonDeserialize(as = ImmutableCreateNamespaceRequest.class)
+public abstract class CreateNamespaceRequest {
 
-  private Namespace namespace;
-  private Map<String, String> properties;
+  public abstract Namespace namespace();
 
-  public CreateNamespaceRequest() {
-    // Needed for Jackson Deserialization.
+  public abstract Map<String, String> properties();
+
+  public static CreateNamespaceRequest of(Namespace namespace) {
+    return ImmutableCreateNamespaceRequest.builder().namespace(namespace).build();
   }
 
-  private CreateNamespaceRequest(Namespace namespace, Map<String, String> properties) {
-    this.namespace = namespace;
-    this.properties = properties;
-    validate();
-  }
-
-  CreateNamespaceRequest validate() {
-    Preconditions.checkArgument(namespace != null, "Invalid namespace: null");
-    return this;
-  }
-
-  public Namespace namespace() {
-    return namespace;
-  }
-
-  public Map<String, String> properties() {
-    return properties != null ? properties : ImmutableMap.of();
+  public static CreateNamespaceRequest of(Namespace namespace, Map<String, String> properties) {
+    Map<String, String> props = properties;
+    if (null == props) {
+      props = ImmutableMap.of();
+    }
+    return ImmutableCreateNamespaceRequest.builder().namespace(namespace).properties(props).build();
   }
 
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
-        .add("namespace", namespace)
-        .add("properties", properties)
+        .add("namespace", namespace())
+        .add("properties", properties())
         .toString();
-  }
-
-  public static Builder builder() {
-    return new Builder();
-  }
-
-  public static class Builder {
-    private Namespace namespace;
-    private final ImmutableMap.Builder<String, String> properties = ImmutableMap.builder();
-
-    private Builder() {
-    }
-
-    public Builder withNamespace(Namespace ns) {
-      Preconditions.checkNotNull(ns, "Invalid namespace: null");
-      this.namespace = ns;
-      return this;
-    }
-
-    public Builder setProperties(Map<String, String> props) {
-      Preconditions.checkNotNull(props, "Invalid collection of properties: null");
-      Preconditions.checkArgument(!props.containsKey(null), "Invalid property: null");
-      Preconditions.checkArgument(!props.containsValue(null),
-          "Invalid value for properties %s: null", Maps.filterValues(props, Objects::isNull).keySet());
-      properties.putAll(props);
-      return this;
-    }
-
-    public CreateNamespaceRequest build() {
-      return new CreateNamespaceRequest(namespace, properties.build());
-    }
   }
 }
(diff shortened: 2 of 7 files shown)


baseline.gradle
```diff
@@ -62,8 +62,8 @@ subprojects {
   pluginManager.withPlugin('com.palantir.baseline-error-prone') {
     tasks.withType(JavaCompile).configureEach {
       options.errorprone.errorproneArgs.addAll (
-          // error-prone is slow, don't run on tests and generated src
-          '-XepExcludedPaths:.*/(test|generated-src)/.*',
+          // error-prone is slow, don't run on tests/generated-src/generated
+          '-XepExcludedPaths:.*/(test|generated-src|generated)/.*',
```
nastra, line 66: this is so that error-prone doesn't complain about the generated Immutable classes

core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
```diff
@@ -19,81 +19,47 @@
...
-      Preconditions.checkNotNull(ns, "Invalid namespace: null");
-      this.namespace = ns;
-      return this;
-    }
-
-    public Builder setProperties(Map<String, String> props) {
-      Preconditions.checkNotNull(props, "Invalid collection of properties: null");
-      Preconditions.checkArgument(!props.containsKey(null), "Invalid property: null");
-      Preconditions.checkArgument(!props.containsValue(null),
-          "Invalid value for properties %s: null", Maps.filterValues(props, Objects::isNull).keySet());
```
nastra, line 90: note that validation code like this (null checks) is implicit in the generated Immutable class)

core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
```diff
@@ -19,81 +19,42 @@
...
+    return ImmutableCreateNamespaceRequest.builder().namespace(namespace).build();
   }
 
-  public Namespace namespace() {
-    return namespace;
-  }
-
-  public Map<String, String> properties() {
-    return properties != null ? properties : ImmutableMap.of();
+  public static CreateNamespaceRequest of(Namespace namespace, Map<String, String> properties) {
```
nastra, line 49: the purpose of the `of()` methods is just to simplify things on the caller's side

core/src/main/java/org/apache/iceberg/rest/requests/UpdateNamespacePropertiesRequest.java
```diff
@@ -19,109 +19,35 @@
...
-          "Invalid property to remove: null");
-      removalsBuilder.addAll(removals);
-      return this;
-    }
-
-    public Builder update(String key, String value) {
-      Preconditions.checkNotNull(key,
-          "Invalid property to update: null");
-      Preconditions.checkNotNull(value,
-          "Invalid value to update for key [%s]: null. Use remove instead", key);
```
nastra, line 102: same here in terms of validation: These are all implicitly checked. The only thing that could be improved in Immutables when this fails would be a similar error message like this one here, where the actual `key` with the null value is shown.
I'll open a PR on the Immutables side to address this

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateNamespaceRequest.java
```diff
@@ -166,6 +139,6 @@ public void assertEquals(CreateNamespaceRequest actual, CreateNamespaceRequest e
 
   @Override
   public CreateNamespaceRequest deserialize(String json) throws JsonProcessingException {
-    return mapper().readValue(json, CreateNamespaceRequest.class).validate();
```
nastra, line 169: note that we don't have to explicitly validate anymore as objects created via Immutables are by definition always in a valid state.

core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateNamespacePropertiesRequest.java
```diff
@@ -45,35 +45,38 @@ public void testRoundTripSerDe() throws JsonProcessingException {
     // Full request
     String fullJson = "{\"removals\":[\"foo\",\"bar\"],\"updates\":{\"owner\":\"Hank\"}}";
     assertRoundTripSerializesEquallyFrom(
-        fullJson, UpdateNamespacePropertiesRequest.builder().updateAll(UPDATES).removeAll(REMOVALS).build());
+        fullJson, ImmutableUpdateNamespacePropertiesRequest.builder().updates(UPDATES).removals(REMOVALS).build());
```
nastra, line 48: given that we're directly working with the `updates` / `removals` collection, it's actually easier to use directly the builder from `ImmutableUpdateNamespacePropertiesRequest` and not have simplified `of(..)` methods like in the other class

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateNamespaceRequest.java
```diff
@@ -73,76 +69,56 @@ public void testCanDeserializeWithoutDefaultValues() throws JsonProcessingExcept
   @Test
   public void testDeserializeInvalidRequest() {
     String jsonIncorrectTypeForNamespace = "{\"namespace\":\"accounting%00tax\",\"properties\":null}";
-    AssertHelpers.assertThrows(
```
nastra, line 76: the reason I changed this to `Assertions.assertThatThrownBy` is because that I think the message in `AssertHelpers.assertThrows(..)` is rather noisy. Usually it's clear from looking at the code what's being expected and when the assertion itself fails, it will be descriptive enough to understand why the code fails.

core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
```diff
@@ -19,81 +19,42 @@
...
-import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
-import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.immutables.value.Value;
 
 /**
  * A REST request to create a namespace, with an optional set of properties.
+ *
+ * Note that Immutable classes by definition don't have a default no-arg constructor (which is required for Jackson),
+ * therefore the @{@link JsonSerialize}/@{@link JsonDeserialize} annotations on the class will generate what's
+ * necessary for Jackson-binding to properly work with this class.
```
rdblue, line 34: What does this produce? Does Immutables use Jackson annotations in the generated code?

core/src/main/java/org/apache/iceberg/rest/requests/UpdateNamespacePropertiesRequest.java
kbendick: Is it possible to have a value of `null` in one of the map collections with immutables (like a configuration or annotation to put on the class maybe)?

core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
nastra: It does 2 things.
1. It generates a JSON Helper class with `@JsonProperty` annotations that looks liks this (note that this is directly copied from the generated class):
```
/**
   * Utility type used to correctly read immutable object from JSON representation.
   * @deprecated Do not use this type directly, it exists only for the <em>Jackson</em>-binding infrastructure
   */
  @Generated(from = "CreateNamespaceRequest", generator = "Immutables")
  @Deprecated
  @SuppressWarnings("Immutable")
  @JsonDeserialize
  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.NONE)
  static final class Json extends CreateNamespaceRequest {
    @Nullable Namespace namespace;
    @Nullable Map<String, String> properties = Collections.emptyMap();
    @JsonProperty("namespace")
    public void setNamespace(Namespace namespace) {
      this.namespace = namespace;
    }
    @JsonProperty("properties")
    public void setProperties(Map<String, String> properties) {
      this.properties = properties;
    }
    @Override
    public Namespace namespace() { throw new UnsupportedOperationException(); }
    @Override
    public Map<String, String> properties() { throw new UnsupportedOperationException(); }
  }
```
2. It generates a utility method with the `@JsonCreator` annotation, that looks exactly like this:
```
/**
   * @param json A JSON-bindable data structure
   * @return An immutable value type
   * @deprecated Do not use this method directly, it exists only for the <em>Jackson</em>-binding infrastructure
   */
  @Deprecated
  @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
  static ImmutableCreateNamespaceRequest fromJson(Json json) {
    ImmutableCreateNamespaceRequest.Builder builder = ImmutableCreateNamespaceRequest.builder();
    if (json.namespace != null) {
      builder.namespace(json.namespace);
    }
    if (json.properties != null) {
      builder.putAllProperties(json.properties);
    }
    return builder.build();
  }
```

Note that this only happens when we add `@JsonSerialize` / `@JsonDeserialize` to the class

core/src/main/java/org/apache/iceberg/rest/requests/UpdateNamespacePropertiesRequest.java
nastra: It looks like this isn't encouraged, but it's possible according to http://immutables.github.io/immutable.html#nulls-in-collection

nastra: here's also an example of its usage: https://github.com/immutables/immutables/blob/399488fe9bfe058cdd0eb3fe82cb069e08656e6f/value-fixture/src-java-11/org/immutables/fixture/nullable/NullablyElements.java#L28

nastra: I opened https://github.com/immutables/immutables/pull/1371 to improve the error messaging for Map null values

core/src/main/java/org/apache/iceberg/rest/requests/CreateNamespaceRequest.java
nastra: attaching https://immutables.github.io/json.html for completeness, since it also mentions alternative ways to Jackson integration

rdblue: Ah, we don't want to use Jackson annotations.

[Annotations aren't required to be present](https://stackoverflow.com/questions/3567413/why-doesnt-a-missing-annotation-cause-a-classnotfoundexception-at-runtime), so they don't work well in libraries. Relying on annotations that may not be there adds a significant source of bugs: the runtime classpath. That's already one of the hardest places to track down issues.

By not using annotations and instead relying on the basic behavior, we can eliminate classpath issues causing serialization problems. I think that's well worth having some builder code that doesn't change much.

PR #4267, opened 2022-03-04
GCP: Add metrics tracking for GCSFileIO

rajarshisarkar:
This PR adds metrics tracking for `GCSFileIO`. This would help reporting IO metrics via the Hadoop `FileSystem.Statistics`.

### DIFF
--- a/gcp/src/main/java/org/apache/iceberg/gcp/gcs/GCSInputFile.java
+++ b/gcp/src/main/java/org/apache/iceberg/gcp/gcs/GCSInputFile.java
@@ -24,14 +24,17 @@ import com.google.cloud.storage.Storage;
 import org.apache.iceberg.gcp.GCPProperties;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.SeekableInputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
 public class GCSInputFile extends BaseGCSFile implements InputFile {
-  public static GCSInputFile fromLocation(String location, Storage storage, GCPProperties gcpProperties) {
-    return new GCSInputFile(storage, BlobId.fromGsUtilUri(location), gcpProperties);
+
+  public static GCSInputFile fromLocation(String location, Storage storage,
+      GCPProperties gcpProperties, MetricsContext metrics) {
+    return new GCSInputFile(storage, BlobId.fromGsUtilUri(location), gcpProperties, metrics);
   }
 
-  GCSInputFile(Storage storage, BlobId blobId, GCPProperties gcpProperties) {
-    super(storage, blobId, gcpProperties);
+  GCSInputFile(Storage storage, BlobId blobId, GCPProperties gcpProperties, MetricsContext metrics) {
+    super(storage, blobId, gcpProperties, metrics);
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


rajarshisarkar:
@danielcweeks Can you please review when you get time. If the implementation looks okay then I can go ahead with the other FileIOs.

gcp/src/test/java/org/apache/iceberg/gcp/gcs/GCSOutputStreamTest.java
```diff
@@ -57,14 +58,16 @@ public void testWrite() {
 
   @Test
   public void testMultipleClose() throws IOException {
-    GCSOutputStream stream = new GCSOutputStream(storage, randomBlobId(), properties);
+    GCSOutputStream stream = new GCSOutputStream(storage, randomBlobId(), properties, MetricsContext
+        .nullMetrics());
```
kbendick, line 62: Nit: Avoid breaking up a chained call as an argument across two lines. In this case I would move all arguments to the next line.

But having `MetricsContext` and then `.nullMetrics()` on the next line is confusing.

danielcweeks:
Hey @rajarshisarkar This is great.  I'd actually point out there is some parallel work to make the initialization dynamic in #4254 and that looks like it's about to go in.  You might want to take a quick peek and see if we can align with that approach.

gcp/src/test/java/org/apache/iceberg/gcp/gcs/GCSOutputStreamTest.java
rajarshisarkar: Thanks, I have made the changes.

rajarshisarkar:
Thanks for having a look, @danielcweeks. Sure, I shall rebase once https://github.com/apache/iceberg/pull/4254 is merged to master.

gcp/src/main/java/org/apache/iceberg/gcp/gcs/GCSInputFile.java
```diff
@@ -24,14 +24,17 @@
 import org.apache.iceberg.gcp.GCPProperties;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.SeekableInputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
 public class GCSInputFile extends BaseGCSFile implements InputFile {
-  public static GCSInputFile fromLocation(String location, Storage storage, GCPProperties gcpProperties) {
-    return new GCSInputFile(storage, BlobId.fromGsUtilUri(location), gcpProperties);
+
+  public static GCSInputFile fromLocation(String location, Storage storage,
```
rdblue, line 31: Should this be public? Since it is a public constructor in a public class, we should deprecate the old constructor instead of directly removing it. If we want to avoid needing to deprecate in the future, then I think we should consider making this class and constructor package-private.

rdblue:
Most of this looks good. Thanks, @rajarshisarkar!

gcp/src/main/java/org/apache/iceberg/gcp/gcs/GCSInputFile.java
rajarshisarkar: Thanks, for the feedback. I have made the class and method package-private.

rajarshisarkar:
Thanks, for the review @kbendick @danielcweeks @rdblue! 

Should we consider merging this PR (or) wait for https://github.com/apache/iceberg/pull/4254 to be merged to master?

rdblue:
I'm going to merge this since there's still ongoing discussion about #4254. Thanks @rajarshisarkar!

PR #2105, opened 2021-01-18
Flink: Initial implementation of  Flink source with the new FLIP-27 source interface

stevenzwu:
Scope in the first version

- simple split assigner (no ordering or locality aware)
- support both batch and streaming read

This is the uber PR for the reference of complete context. Will submit smaller PRs for the code review

1. [MERGED] refactor Flink tests so that new source implementation can reuse](#2047)
2. [MERGED] Upgrade Flink version to 1.12.1(#1956)
3. [PENDING] FLIP-27 Iceberg source split (#3501 )
4. SimpleSplitAssigner (TBD). Note that other assigners will be added after this work is completed.
5. Split enumerator (TBD)
6. IcebergSource where everything put together (TBD)

The new `IcebergSource` will be marked as `@Experimental` as FLIP-27 source is maturing and we are making it production ready.

Here is the [design doc](https://docs.google.com/document/d/1q6xaBxUPFwYsW9aXWxYUh7die6O7rDeAPFQcTAMQ0GM/edit#heading=h.z2xhib39g0nm) that my colleague (@sundargates) and I created, as mentioned in #1626.

PR #4269, opened 2022-03-04
Flink: FLIP-27 source reader

stevenzwu:


stevenzwu:
This PR has no unit test as it mainly contains the plumbing classes for Iceberg source. We will have full test with Iceberg source.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
```diff
@@ -0,0 +1,60 @@
...
+public class IcebergSourceEvents {
+
+  /**
+   * We can remove this class once FLINK-21364 is resolved.
+   */
+  public static final class SplitRequestEvent implements SourceEvent {
+    private static final long serialVersionUID = 1L;
+
+    private final Collection<String> finishedSplitIds;
+    @Nullable
```
rdblue, line 36: Nit: need a newline between the field and method.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Collections;
+import javax.annotation.Nullable;
+import org.apache.flink.api.connector.source.SourceEvent;
+
+public class IcebergSourceEvents {
+
+  /**
+   * We can remove this class once FLINK-21364 is resolved.
+   */
+  public static final class SplitRequestEvent implements SourceEvent {
```
rdblue, line 32: Does this need to be nested? Also, does it need to be public if it is a nested class?

rdblue: Also, review `final`. We don't typically use `final` for classes.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private final Collection<String> finishedSplitIds;
+    @Nullable
+    private final String requesterHostname;
+
+    public SplitRequestEvent() {
+      this(Collections.emptyList());
```
rdblue, line 40: Do we need this constructor? Below, the initial split is requested using `requestSplit(Collections.emptyList())`.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceReaderMetrics.java
```diff
@@ -0,0 +1,68 @@
...
+
+    this.assignedSplits = new AtomicLong();
+    this.assignedBytes = new AtomicLong();
+    this.finishedSplits = new AtomicLong();
+    this.finishedBytes = new AtomicLong();
+    readerMetricGroup.gauge("assignedSplits", assignedSplits::get);
+    readerMetricGroup.gauge("assignedBytes", assignedBytes::get);
+    readerMetricGroup.gauge("finishedSplits", finishedSplits::get);
+    readerMetricGroup.gauge("finishedBytes", finishedBytes::get);
+    this.splitReaderFetchCalls = readerMetricGroup.counter("splitReaderFetchCalls");
```
kbendick, line 46: Can the gauges be made into counters? They don't seem to be accessible outside of this class anyway.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceReader.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  protected IcebergSourceSplit initializedState(IcebergSourceSplit split) {
+    return split;
+  }
+
+  @Override
+  protected IcebergSourceSplit toSplitType(
+      String splitId,
+      IcebergSourceSplit splitState) {
```
rdblue, line 65: Can these args fit on one line?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceReader.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public void start() {
+    requestSplit(Collections.emptyList());
+  }
+
+  @Override
+  protected void onSplitFinished(Map<String, IcebergSourceSplit> finishedSplitIds) {
+    if (!finishedSplitIds.isEmpty()) {
```
rdblue, line 52: When would this be empty? And should this still call `requestSplit` even if there is no finished split ID?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceReaderMetrics.java
```diff
@@ -0,0 +1,68 @@
...
+
+package org.apache.iceberg.flink.source.reader;
+
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.metrics.Counter;
+import org.apache.flink.metrics.MetricGroup;
+
+@Internal
+public class IcebergSourceReaderMetrics {
```
rdblue, line 28: Can we implement `MetricsContext` here? That way we're using the same API for metrics in multiple places rather than introducing a new API.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
```diff
@@ -0,0 +1,134 @@
...
+      } catch (UncheckedIOException e) {
+        throw e.getCause();
+      }
+    } else {
+      return finishSplit();
+    }
+  }
+
+  @Override
+  public void handleSplitsChanges(SplitsChange<IcebergSourceSplit> splitsChanges) {
```
rdblue, line 75: Does this need to validate that SplitsChange is an addition? If that is possible, I'd recommend adding the logic now and throwing an exception for unsupported changes

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  private void checkSplitOrStartNext() throws IOException {
+    if (currentReader != null) {
+      return;
+    }
+
+    IcebergSourceSplit nextSplit = splits.poll();
+    if (nextSplit == null) {
+      throw new IOException("No split remaining");
```
rdblue, line 113: Maybe add a comment to explain why this is the right behavior. It seems odd to me that it should throw an exception when it hasn't received new splits. Maybe that is the expectation from how `fetch` is called, but it is definitely concerning to see.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
```diff
@@ -0,0 +1,134 @@
...
+  private final ReaderFunction<T> readerFunction;
+  private final int indexOfSubtask;
+  private final IcebergSourceReaderMetrics metrics;
+  private final Queue<IcebergSourceSplit> splits;
+
+  private CloseableIterator<RecordsWithSplitIds<RecordAndPosition<T>>> currentReader;
+  private IcebergSourceSplit currentSplit;
+  private String currentSplitId;
+
+  IcebergSourceSplitReader(ReaderFunction<T> readerFunction,
```
rdblue, line 47: It may be better to call this `openSplitFunction`?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
stevenzwu: I will keep it for now as a convenience. Some other test code (not included in this PR) calls this constructor.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
stevenzwu: will move it out and remove the outer class. will remove `final`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
stevenzwu: seems like an invalid comment

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceReader.java
stevenzwu: you are right. checked the base class code. it won't be empty. will remove the unnecessary if check

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: good catch. will add the check and throw an exception.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
kbendick: I _think_ he means that if you use `@Nullable`, you should place an empty line between it and the field before it.

so like
```java
private Collection<String> finishedSplits;

@Nullable
private String …
```

I think that’s the style nit here.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: This is a great question. I looked into Flink code again. Throwing an IOException will cause the fetcher to be closed so that we don't have to poll the fetcher again. that is not a problem, because new split later will create new fetcher.

This is the `FetchTask` that calls `splitReader.fetch`. We can see that the return value can't be null. I agree that this is a little weird behavior. cc @tweise

```
    public boolean run() throws IOException {
        try {
            if (!isWakenUp() && lastRecords == null) {
                lastRecords = splitReader.fetch();
            }

            if (!isWakenUp()) {
                // The order matters here. We must first put the last records into the queue.
                // This ensures the handling of the fetched records is atomic to wakeup.
                if (elementsQueue.put(fetcherIndex, lastRecords)) {
                    if (!lastRecords.finishedSplits().isEmpty()) {
                        // The callback does not throw InterruptedException.
                        splitFinishedCallback.accept(lastRecords.finishedSplits());
                    }
                    lastRecords = null;
                }
            }
        }
```

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: I still feel `ReaderFunction` is more accurate. here is the interface from previously merged PR. It basically read a split and return an iterator of record batch.
```
public interface ReaderFunction<T> extends Serializable,
    Function<IcebergSourceSplit, CloseableIterator<RecordsWithSplitIds<RecordAndPosition<T>>>> {
}
```

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSourceEvents.java
stevenzwu: got it. I will remove the `@Nullable` annotation. so we won't need the empty line anymore

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceReaderMetrics.java
stevenzwu: will switch to `MetricsContext`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: aded a method comment section to explain the IOException behavior

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/RecordAndPosition.java
```diff
@@ -67,12 +67,6 @@ public void set(T newRecord, int newFileOffset, long newRecordOffset) {
     this.recordOffset = newRecordOffset;
   }
 
-  /** Sets the position without setting a record. */
-  public void position(int newFileOffset, long newRecordOffset) {
```
stevenzwu, line 71: removing an unused method

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+      return finishSplit();
+    }
+  }
+
+  @Override
+  public void handleSplitsChanges(SplitsChange<IcebergSourceSplit> splitsChange) {
+    if (!(splitsChange instanceof SplitsAddition)) {
+      throw new UnsupportedOperationException(String.format(
+          "The SplitChange type of %s is not supported.", splitsChange.getClass()));
```
rdblue, line 92: Unsupported split change: %s

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
rdblue: This is probably better as an inline comment.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    LOG.info("Add splits to reader: {}", splitsChange.splits());
```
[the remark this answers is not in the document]
stevenzwu: done

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: thx. done

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: renamed the variable to `openSplitFunction`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/reader/IcebergSourceSplitReader.java
stevenzwu: changed to inline comment

yittg: How about making it a blocking poll that can be waken up by `#wakeUp` per 
https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-base/src/main/java/org/apache/flink/connector/base/source/reader/splitreader/SplitReader.java#L37-L48

It's ok to keep it simple at first though.

stevenzwu: @yittg that is a good question. In theory, I agree it is a good behavior. I haven't figured a good way to implement the wakeUp behavior

Say there is no split available from the queue. if fetch() method doesn't return, it would need to poll the `splits` queue to check for new split in some loop with sleep. if the wakerUp is called and set the wakeUp flag for the fetch method to break out of the loop and return an empty `RecordsWithSplitIds` with no records and no finished splits.

We can't use wait-notify to signal btw wakeUp and fetch. Otherwise, fetch method won't do a splits queue polling and discover new splits during wait. That is also undesirable.

Maybe there is other way I haven't thought about.

stevenzwu: My colleague suggested a good way. We can just return an empty result in this case. it will cause fetcher to be idle and SplitFetcherManager will closes idle fetcher. At least, it is like a normal flow (than IOException). I pushed a commit for the change.

yittg: SGTM roughly, thanks

stevenzwu: Forgot to mention. For the change of return an empty result for no split case, it has been validated by the unit test in the uber draft PR, which contains MiniCluster unit tests for both batch and streaming executions. https://github.com/apache/iceberg/pull/2105

rdblue (approved):

PR #4270, opened 2022-03-04
Flink: FLIP-27 split assigner

stevenzwu:


stevenzwu:
Current interface design incorporated the requirements from the event time /watermark aligned assigner (for a single source or across sources).

We may revisit the watermark alignment assigner. If that changed, we can come back and revisit the assigner interface. For now, I think we can move ahead because we aren't going to public advertise the FLIP-27 source yet.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
```diff
@@ -0,0 +1,80 @@
...
+  private final IcebergSourceSplit split;
+
+  public GetSplitResult(Status status) {
+    this(status, null);
+  }
+
+  public GetSplitResult(Status status, @Nullable IcebergSourceSplit split) {
+    if (null == split && status == Status.AVAILABLE) {
+      throw new IllegalArgumentException("Available status must have a non-null split");
+    }
```
rdblue, line 56: This would normally be a precondition. That helps add more context to the exception message.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
```diff
@@ -0,0 +1,80 @@
...
+     * Assigner doesn't have pending splits.
+     */
+    UNAVAILABLE
+  }
+
+  private final Status status;
+  @Nullable
+  private final IcebergSourceSplit split;
+
+  public GetSplitResult(Status status) {
```
rdblue, line 49: Looks like this should be private since unavailable and constrained are singletons.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
```diff
@@ -0,0 +1,80 @@
...
+    this(status, null);
+  }
+
+  public GetSplitResult(Status status, @Nullable IcebergSourceSplit split) {
```
rdblue, line 53: You may want to make this private and use a factory method like the other event types:

```java
  public static GetSplitResult forSplit(IcebergSourceSplit split) {
    return new GetSplitResult(AVAILABLE, split);
  }
```

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Assigner doesn't have pending splits.
+     */
+    UNAVAILABLE
+  }
+
+  private final Status status;
+  @Nullable
+  private final IcebergSourceSplit split;
```
kbendick, line 47: Nit: If using `@Nullable`, I think there's supposed to be a new line between the two.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/SimpleSplitAssigner.java
```diff
@@ -0,0 +1,98 @@
...
+  private final Deque<IcebergSourceSplit> pendingSplits;
+  private CompletableFuture<Void> availableFuture;
+
+  public SimpleSplitAssigner() {
+    this.pendingSplits = new ArrayDeque<>();
+  }
+
+  public SimpleSplitAssigner(Collection<IcebergSourceSplitState> assignerState) {
+    this.pendingSplits = new ArrayDeque<>(assignerState.size());
+    assignerState.stream().forEach(splitState -> pendingSplits.add(splitState.split()));
```
rdblue, line 49: Is this assuming that all of the splits are unassigned?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/SplitAssigner.java
```diff
@@ -0,0 +1,110 @@
...
+ * <li> Snapshot aware assigner that assign splits based on the order they are committed.</li>
+ * <li> Event time alignment assigner that assign splits satisfying certain time ordering
+ *      within a single source or across sources.</li>
+ * </ul>
+ *
+ * <p>
+ * Enumerator should call the assigner APIs from the coordinator thread.
+ * This is to simplify the thread safety for assigner implementation.
+ */
+public interface SplitAssigner extends AutoCloseable {
```
rdblue, line 44: Most of the time, we use `Closeable` rather than `AutoCloseable` because many libraries check for `Closeable`. `Closeable` implements `AutoCloseable`, but not the other way around so it is usually safer to use `Closeable` even if it isn't as accurate.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
stevenzwu: will remove the @Nullable

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
stevenzwu: will make all constructors private. only construct via factory methods

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/GetSplitResult.java
stevenzwu: will change to precondition

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/assigner/SimpleSplitAssigner.java
stevenzwu: that is correct. will add inline comments to explain

rdblue (approved):

rdblue:
Thanks, @stevenzwu!

PR #4272, opened 2022-03-05
Docs: Clarify contract for file uniqueness

emkornfield:
This is based on a [mailing list thread](https://lists.apache.org/thread/n63h22y1zpg85wsxjf48o4mmjvc5c378) about file uniqueness

### DIFF
--- a/format/spec.md
+++ b/format/spec.md
@@ -444,6 +444,8 @@ Notes:
 
 1. Technically, data files can be deleted when the last snapshot that contains the file as “live” data is garbage collected. But this is harder to detect and requires finding the diff of multiple snapshots. It is easier to track what files are deleted in a snapshot and delete them when that snapshot expires.
 2. Manifest list files are required in v2, so that the `sequence_number` and `snapshot_id` to inherit are always available.
+3. When a data file or delete file is marked as deleted in a manifest the writer of the new snapshot may not include a prexisting manifest that references the data file as EXISTING or ADDED. This implies that the manifest from the prior snapshot that including the newly deleted data file must not be in the new snapshot and that any files referenced by the manifest that are still part of the table must be written to a new
+manifest.
 
 #### Sequence Number Inheritance
 
@@ -549,7 +551,10 @@ For example, an `events` table with a timestamp column named `ts` that is partit
 
 Scan predicates are also used to filter data and delete files using column bounds and counts that are stored by field id in manifests. The same filter logic can be used for both data and delete files because both store metrics of the rows either inserted or deleted. If metrics show that a delete file has no rows that match a scan predicate, it may be ignored just as a data file would be ignored [2].
 
-Data files that match the query filter must be read by the scan.
+Data files that match the query filter must be read by the scan. Note that for any snapshot, all file paths
+may appear at most once across all manifest files in the snapshot. If a file path appears more then once, the results of the scan are undefined. Reader implementations may raise an error in this case, but are
+not required to do so.
+
 
 Delete files that match the query filter must be applied to data files at read time, limited by the scope of the delete file using the following rules.
 
--- a/format/spec.md
+++ b/format/spec.md
@@ -389,7 +389,7 @@ The schema of a manifest file is a struct called `manifest_entry` with the follo
 
 | v1         | v2         | Field id, name           | Type                                                      | Description                                                                           |
 | ---------- | ---------- |--------------------------|-----------------------------------------------------------|---------------------------------------------------------------------------------------|
-| _required_ | _required_ | **`0  status`**          | `int` with meaning: `0: EXISTING` `1: ADDED` `2: DELETED` | Used to track additions and deletions                                                 |
+| _required_ | _required_ | **`0  status`**          | `int` with meaning: `0: EXISTING` `1: ADDED` `2: DELETED` | Used to track additions and deletions. Delitions are informational only and not used in scans.                                                 |
 | _required_ | _optional_ | **`1  snapshot_id`**     | `long`                                                    | Snapshot id where the file was added, or deleted if status is 2. Inherited when null. |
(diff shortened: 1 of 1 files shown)


docs/common/format/spec.md
```diff
@@ -549,7 +551,10 @@ For example, an `events` table with a timestamp column named `ts` that is partit
 
 Scan predicates are also used to filter data and delete files using column bounds and counts that are stored by field id in manifests. The same filter logic can be used for both data and delete files because both store metrics of the rows either inserted or deleted. If metrics show that a delete file has no rows that match a scan predicate, it may be ignored just as a data file would be ignored [2].
 
-Data files that match the query filter must be read by the scan.
+Data files that match the query filter must be read by the scan. Note that for any snapshot, all file paths
+may appear at most once across all manifest files in the snapshot. If a file path appears more then once, the results of the scan are undefined. Reader implementations may raise an error in this case, but are
+not required to do so.
```
rdblue, line 556: Nit: the line breaks are strange here. Can you remove them?

docs/common/format/spec.md
```diff
@@ -549,7 +551,10 @@ For example, an `events` table with a timestamp column named `ts` that is partit
 
 Scan predicates are also used to filter data and delete files using column bounds and counts that are stored by field id in manifests. The same filter logic can be used for both data and delete files because both store metrics of the rows either inserted or deleted. If metrics show that a delete file has no rows that match a scan predicate, it may be ignored just as a data file would be ignored [2].
 
-Data files that match the query filter must be read by the scan.
+Data files that match the query filter must be read by the scan. Note that for any snapshot, all file paths
+may appear at most once across all manifest files in the snapshot. If a file path appears more then once, the results of the scan are undefined. Reader implementations may raise an error in this case, but are
```
rdblue, line 555: I think that the readers may raise an error or may read the file twice. I'm not sure that I'd consider the result undefined, although I see what you're saying. If the scan doesn't check that data files are unique, then the result should be the file is read twice. If it does, then it can fail with an error.

docs/common/format/spec.md
```diff
@@ -444,6 +444,8 @@ Notes:
 
 1. Technically, data files can be deleted when the last snapshot that contains the file as “live” data is garbage collected. But this is harder to detect and requires finding the diff of multiple snapshots. It is easier to track what files are deleted in a snapshot and delete them when that snapshot expires.
 2. Manifest list files are required in v2, so that the `sequence_number` and `snapshot_id` to inherit are always available.
+3. When a data file or delete file is marked as deleted in a manifest the writer of the new snapshot may not include a prexisting manifest that references the data file as EXISTING or ADDED. This implies that the manifest from the prior snapshot that including the newly deleted data file must not be in the new snapshot and that any files referenced by the manifest that are still part of the table must be written to a new
```
rdblue, line 447: I think we can have a stronger and clearer statement about this.

Delete entries are mainly informational. Those entries are ignored when planning scans and are primarily for tracking files for incremental delete. Meaning that when you remove a snapshot, you can scan through the manifests that were written for that snapshot to find what was deleted in it. There are some issues with this approach, so we tend to recommend comparing reachable file sets to actually drop files. For example, we are reluctant to state in the spec that adding a deleted file back to the table after it has been deleted is now allowed. But that could cause incremental deletes to break valid snapshots.

What I would state in the spec is that any given snapshot should not contain more than one ADDED or EXISTING entry for a file. That's what you're getting at in the update below, I think. We can also note that whether a file can be added to the table after it has been deleted is not recommended.

docs/common/format/spec.md
emkornfield: > I'm not sure that I'd consider the result undefined, although I see what you're saying.

I was using undefined behavior in the [PL sense](https://en.wikipedia.org/wiki/Undefined_behavior#:~:text=In%20computer%20programming%2C%20undefined%20behavior,which%20the%20computer%20code%20adheres.).  To be pedantic, I don't think here is statement in the spec yet, about the DELETE being informational only, so that adds to the possibilities, but I didn't list them.

emkornfield: I stand corrected the first paragraph of this section mentions deleted files, I think I maybe missed it because I was searching for "DELETE" status.

docs/common/format/spec.md
emkornfield: I removed them.

docs/common/format/spec.md
emkornfield: I removed this note and added the comment the not recommended comment to point 1

emkornfield:
@rdblue did you have a chance to look again?

docs/common/format/spec.md
liuml07: I'm wondering if it's worth to add some checks in the core (e.g. introducing `FastAppend::validate` and `MergeAppend::validate`) to check if data files were appended more than once.

format/spec.md
```diff
@@ -549,7 +549,10 @@ For example, an `events` table with a timestamp column named `ts` that is partit
 
 Scan predicates are also used to filter data and delete files using column bounds and counts that are stored by field id in manifests. The same filter logic can be used for both data and delete files because both store metrics of the rows either inserted or deleted. If metrics show that a delete file has no rows that match a scan predicate, it may be ignored just as a data file would be ignored [2].
 
-Data files that match the query filter must be read by the scan.
+Data files that match the query filter must be read by the scan. 
+
+Note that for any snapshot, all file paths marked with "ADDED" or "EXISTING" may appear at most once across all manifest files in the snapshot. If a file path appears more then once, the results of the scan are undefined. Reader implementations may raise an error in this case, but are not required to do so.
+
```
rdblue, line 555: Nit: unnecessary newline.

rdblue (approved):

PR #3810, opened 2021-12-27
ORC: Add configurable write properties

hililiwei:
closes https://github.com/apache/iceberg/issues/3272

> To be consistent with Parquet, it would be good to introduce write table properties for ORC too.
Also make sure that we have a configuration available for ORC separately.
Discussed here: #2935 (comment)

PR #4273, opened 2022-03-05
ORC: Add compression properties

zhongyujiang:
This PR adds compression-codec and compression-strategy of ORC in `TableProperties`, @openinx @hililiwei would you help review this? Thanks!

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -153,7 +153,7 @@ public class TableProperties {
 
   public static final String ORC_COMPRESSION_STRATEGY = "write.orc.compression-strategy";
   public static final String DELETE_ORC_COMPRESSION_STRATEGY = "write.delete.orc.compression-strategy";
-  public static final String ORC_COMPRESSION_STRATEGY_DEFAULT = "compression";
+  public static final String ORC_COMPRESSION_STRATEGY_DEFAULT = "speed";
 
   public static final String SPLIT_SIZE = "read.split.target-size";
   public static final long SPLIT_SIZE_DEFAULT = 128 * 1024 * 1024; // 128 MB
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -147,6 +147,14 @@ public class TableProperties {
   public static final String DELETE_ORC_WRITE_BATCH_SIZE = "write.delete.orc.vectorized.batch-size";
   public static final int ORC_WRITE_BATCH_SIZE_DEFAULT = 1024;
 
+  public static final String ORC_COMPRESSION = "write.orc.compression-codec";
+  public static final String DELETE_ORC_COMPRESSION = "write.delete.orc.compression-codec";
+  public static final String ORC_COMPRESSION_DEFAULT = "zlib";
+
+  public static final String ORC_COMPRESSION_STRATEGY = "write.orc.compression-strategy";
+  public static final String DELETE_ORC_COMPRESSION_STRATEGY = "write.delete.orc.compression-strategy";
+  public static final String ORC_COMPRESSION_STRATEGY_DEFAULT = "compression";
+
   public static final String SPLIT_SIZE = "read.split.target-size";
   public static final long SPLIT_SIZE_DEFAULT = 128 * 1024 * 1024; // 128 MB
 
--- a/orc/src/main/java/org/apache/iceberg/orc/ORC.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORC.java
@@ -59,8 +59,10 @@ import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.util.ArrayUtil;
 import org.apache.iceberg.util.PropertyUtil;
+import org.apache.orc.CompressionKind;
 import org.apache.orc.OrcConf;
 import org.apache.orc.OrcFile;
+import org.apache.orc.OrcFile.CompressionStrategy;
 import org.apache.orc.OrcFile.ReaderOptions;
 import org.apache.orc.Reader;
 import org.apache.orc.TypeDescription;
@@ -199,8 +201,8 @@ public class ORC {
 
       OrcConf.STRIPE_SIZE.setLong(conf, context.stripeSize());
       OrcConf.BLOCK_SIZE.setLong(conf, context.blockSize());
-      OrcConf.COMPRESS.setString(conf, context.codecAsString().toUpperCase(Locale.ENGLISH));
-      OrcConf.COMPRESSION_STRATEGY.setString(conf, context.compressionStrategy().toUpperCase(Locale.ENGLISH));
+      OrcConf.COMPRESS.setString(conf, context.compressionKind().name());
+      OrcConf.COMPRESSION_STRATEGY.setString(conf, context.compressionStrategy().name());
       OrcConf.OVERWRITE_OUTPUT_FILE.setBoolean(conf, overwrite);
 
       return new OrcFileAppender<>(schema,
@@ -212,8 +214,8 @@ public class ORC {
       private final long stripeSize;
       private final long blockSize;
       private final int vectorizedRowBatchSize;
-      private final String codecAsString;
-      private final String compressionStrategy;
+      private final CompressionKind compressionKind;
+      private final CompressionStrategy compressionStrategy;
 
       public long stripeSize() {
         return stripeSize;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -143,6 +143,14 @@ private TableProperties() {
   public static final String DELETE_ORC_BLOCK_SIZE_BYTES = "write.delete.orc.block-size-bytes";
   public static final long ORC_BLOCK_SIZE_BYTES_DEFAULT = 256L * 1024 * 1024; // 256 MB
 
+  public static final String ORC_COMPRESSION = "write.orc.compression-codec";
+  public static final String DELETE_ORC_COMPRESSION = "write.delete.orc.compression-codec";
+  public static final String ORC_COMPRESSION_DEFAULT = "zlib";
```
rdblue, line 148: Other formats use "gzip" for this. Is this what ORC requires?

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -186,35 +204,63 @@ public long blockSize() {
...
+        String codecAsString = conf.get(DELETE_ORC_COMPRESSION, dataContext.codecAsString());
+        String compressionStrategy = conf.get(DELETE_ORC_COMPRESSION_STRATEGY, dataContext.compressionStrategy());
 
-        long blockSize =
-            conf.getLong(TableProperties.DELETE_ORC_BLOCK_SIZE_BYTES, dataContext.stripeSize());
+        return new Context(stripeSize, blockSize, codecAsString, compressionStrategy);
+      }
 
-        return new Context(stripeSize, blockSize);
+      static String lookupValue(Configuration conf, OrcConf orcConf, String icebergPropertyKey) {
```
rdblue, line 253: Something is wrong if this is looking up an Iceberg table property like `ORC_STRIPE_SIZE_BYTES` in a Hadoop Configuration. Iceberg settings should be pulled from table properties, not from Hadoop.

rdblue:
I don't think that we can merge this until we fix the ORC configuration PR that recently went in. See https://github.com/apache/iceberg/pull/3810#issuecomment-1061197742 for more details. The configuration context should not use Hadoop Configuration.

core/src/main/java/org/apache/iceberg/TableProperties.java
amogh-jahagirdar: I think ORC as a format does not support GZIP. https://orc.apache.org/api/orc-core/org/apache/orc/CompressionKind.html

Engines like Presto/Trino tend to have ZLIB as a default compression mode for ORC.

zhongyujiang:
Updated based on #4291.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -206,43 +227,64 @@ public int vectorizedRowBatchSize() {
...
+            ORC_WRITE_BATCH_SIZE, ORC_WRITE_BATCH_SIZE_DEFAULT);
         Preconditions.checkArgument(vectorizedRowBatchSize > 0, "VectorizedRow batch size must be > 0");
 
-        return new Context(stripeSize, blockSize, vectorizedRowBatchSize);
+        String codecAsString = config.getOrDefault(OrcConf.COMPRESS.getAttribute(), ORC_COMPRESSION_DEFAULT);
+        codecAsString = config.getOrDefault(ORC_COMPRESSION, codecAsString);
+
+        String compressionStrategy = config.getOrDefault(OrcConf.COMPRESSION_STRATEGY.getAttribute(),
+            ORC_COMPRESSION_STRATEGY_DEFAULT);
+        compressionStrategy = config.getOrDefault(ORC_COMPRESSION_STRATEGY, compressionStrategy);
```
openinx, line 267: Why not just follow the same config parse approach as the above config keys did ?  I'm saying using the generic `PropertyUtil.propertyAsString`.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -206,43 +227,64 @@ public int vectorizedRowBatchSize() {
...
         Preconditions.checkArgument(blockSize > 0, "Block size must be > 0");
 
         int vectorizedRowBatchSize = PropertyUtil.propertyAsInt(config,
-            TableProperties.ORC_WRITE_BATCH_SIZE, TableProperties.ORC_WRITE_BATCH_SIZE_DEFAULT);
+            ORC_WRITE_BATCH_SIZE, ORC_WRITE_BATCH_SIZE_DEFAULT);
         Preconditions.checkArgument(vectorizedRowBatchSize > 0, "VectorizedRow batch size must be > 0");
 
-        return new Context(stripeSize, blockSize, vectorizedRowBatchSize);
+        String codecAsString = config.getOrDefault(OrcConf.COMPRESS.getAttribute(), ORC_COMPRESSION_DEFAULT);
+        codecAsString = config.getOrDefault(ORC_COMPRESSION, codecAsString);
```
openinx, line 263: I will suggest to use the similar approach as `parquet` did:  try to parse the string into a `CompressionKind` so that we can validate whether this value is a valid one in the correct place. 

See: https://github.com/apache/iceberg/blob/master/parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java#L392

orc/src/main/java/org/apache/iceberg/orc/ORC.java
openinx: Besides, please also use the `CompressionStrategy` enum to validate whether this configured value is a valid one ?   ( Just as this comment said: https://github.com/apache/iceberg/pull/4273/files#r836994947 )

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -206,43 +227,64 @@ public int vectorizedRowBatchSize() {
...
+        long blockSize = PropertyUtil.propertyAsLong(config, DELETE_ORC_BLOCK_SIZE_BYTES, dataContext.blockSize());
 
         int vectorizedRowBatchSize = PropertyUtil.propertyAsInt(config,
-            TableProperties.DELETE_ORC_WRITE_BATCH_SIZE, dataContext.vectorizedRowBatchSize());
+            DELETE_ORC_WRITE_BATCH_SIZE, dataContext.vectorizedRowBatchSize());
+
+        String codecAsString = config.getOrDefault(DELETE_ORC_COMPRESSION, dataContext.codecAsString());
+
+        String compressionStrategy = config.getOrDefault(DELETE_ORC_COMPRESSION_STRATEGY,
+            dataContext.compressionStrategy());
```
openinx, line 285: Nit: Please use the same approach to parse the `codec` & `compression-strategy` .

openinx (commented):
I just left serval comments, and  I think we are recommended to add unit tests as this [PR](https://github.com/apache/iceberg/pull/3810/files#diff-1fd7634d23331b314ff21ce1d908c7ce87d467c4aaa20ce061cf66f1ff72b069R55) did to avoid regression .

core/src/main/java/org/apache/iceberg/TableProperties.java
hililiwei: Given `ZSTD`'s good performance, are we considering it as the default?

zhongyujiang:
Updated based on comments.

orc/src/test/java/org/apache/iceberg/orc/TestTableProperties.java
```diff
@@ -82,19 +93,29 @@ public void testOrcTableProperties() throws Exception {
         DynFields.builder().hiddenImpl(writer.getClass(), "conf").build(writer);
 
     Configuration configuration = confField.get();
-    Assert.assertEquals(String.valueOf(blockSizeBytes), configuration.get(OrcConf.BLOCK_SIZE.getAttribute()));
-    Assert.assertEquals(String.valueOf(stripeSizeBytes), configuration.get(OrcConf.STRIPE_SIZE.getAttribute()));
```
zhongyujiang, line 86: Changed this to using OrcConf to get value directly so no need to know property key detail.

orc/src/test/java/org/apache/iceberg/orc/TestTableProperties.java
```diff
@@ -53,12 +56,20 @@
 
   @Test
   public void testOrcTableProperties() throws Exception {
-    Long stripeSizeBytes = 32L * 1024 * 1024;
-    Long blockSizeBytes = 128L * 1024 * 1024;
```
zhongyujiang, line 57: Changed `Long` to `long` because what `OrcConf.XXX#getLong` returns is a `long`, and I got 

> Ambiguous method call. Both assertEquals(Object, Object) in Assert and assertEquals(long, long) in Assert match

when using both `Long` and `long` as parameters for `Assert#assertEquals`.

core/src/main/java/org/apache/iceberg/TableProperties.java
zhongyujiang: Although I believe ZSTD has been used in production by many users already, I personally prefer to align with the ORC default. Besides, I think we'd better start a new thread to discuss this.

openinx: Yeah, I agree with @zhongyujiang, it's good to align with the apache orc default value. 
 According to the document https://orc.apache.org/docs/hive-config.html .

zhongyujiang:
Thanks all for your review!

PR #4276, opened 2022-03-07
Common: Add dynamic loading exceptions as suppressed exceptions

rdblue:
This adds the exceptions thrown while attempting to dynamically load a constructor as suppressed exceptions on the exception that is finally thrown. This should preserve causes that are currently unavailable because only the exception messages are used.


CodingCat (approved):
LGTM, thanks!

openinx (approved):
LGTM, thanks @rdblue for the contribution, and thanks @CodingCat for the reviewing.

PR #4279, opened 2022-03-07
Spark: Allow Bin-Pack Rewrite between Partition Specs

RussellSpitzer:
Previously, a rewrite between partition specs would fail as the rows from files in one spec
may actually belong to a different partition in the current spec. In this PR the partitioning
is updated via a range repartition if the destination spec does not match the origin spec.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DistributionMode;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.actions.BinPackStrategy;
@@ -67,12 +68,19 @@ public class Spark3BinPackStrategy extends BinPackStrategy {
           .option(SparkReadOptions.FILE_OPEN_COST, "0")
           .load(table.name());
 
+      // All files within a file group are written with the same spec, so check the first
+      boolean requiresRepartition = !filesToRewrite.get(0).spec().equals(table.spec());
+
+      // Invoke a shuffle if the partition spec of the incoming partition does not match the table
+      String distributionMode = requiresRepartition ? DistributionMode.RANGE.modeName() :
+          DistributionMode.NONE.modeName();
+
       // write the packed data into new files where each split becomes a new file
       scanDF.write()
           .format("iceberg")
           .option(SparkWriteOptions.REWRITTEN_FILE_SCAN_TASK_SET_ID, groupID)
           .option(SparkWriteOptions.TARGET_FILE_SIZE_BYTES, writeMaxFileSize())
-          .option(SparkWriteOptions.DISTRIBUTION_MODE, "none")
+          .option(SparkWriteOptions.DISTRIBUTION_MODE, distributionMode)
           .mode("append")
           .save(table.name());
 
--- a/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
+++ b/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
@@ -196,6 +196,34 @@ public class TestRewriteDataFilesAction extends SparkTestBase {
   }
 
   @Test
+  public void testBinPackAfterPartitionChange() {
+    Table table = createTable();
+
+    writeRecords(20, SCALE, 20);
+    shouldHaveFiles(table, 20);
+    table.updateSpec().addField(Expressions.ref("c1")).commit();
+
+    List<Object[]> originalData = currentData();
+
+    RewriteDataFiles.Result result =
+        basicRewrite(table)
+            .option(SortStrategy.MIN_INPUT_FILES, "1")
+            .option(SortStrategy.MIN_FILE_SIZE_BYTES, Integer.toString(averageFileSize(table) + 1000))
+            .option(RewriteDataFiles.TARGET_FILE_SIZE_BYTES, Integer.toString(averageFileSize(table) + 1001))
+            .execute();
+
+    Assert.assertEquals("Should have 1 fileGroup because all files were not correctly partitioned",
+        1, result.rewriteResults().size());
+
+    List<Object[]> postRewriteData = currentData();
+    assertEquals("We shouldn't have changed the data", originalData, postRewriteData);
+
+    shouldHaveSnapshots(table, 2);
+    shouldHaveACleanCache(table);
+    shouldHaveFiles(table, 20);
+  }
+
+  @Test
   public void testBinPackWithDeletes() throws Exception {
     Table table = createTablePartitioned(4, 2);
     table.updateProperties().set(TableProperties.FORMAT_VERSION, "2").commit();


RussellSpitzer:
@aokolnychyi this is the PR we were discussing earlier. My big question here is whether I should just skip the test refactor and only test identity transforms. The options are basically

1. Test only identity transforms, very small PR
2. Test bucket transforms, requires refactor so we can run the test in a context where the session extensions exist

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
```diff
@@ -67,12 +68,18 @@ public Table table() {
           .option(SparkReadOptions.FILE_OPEN_COST, "0")
           .load(table.name());
 
+      boolean requiresRepartition = !filesToRewrite.get(0).spec().equals(table.spec());
```
RussellSpitzer, line 71: Actual fix is here, everything else is test refactoring

szehon-ho, line 71: Wondering, is it enough to just check if the first one is not the same as current partition spec, and not the rest?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteExtensions.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.iceberg.spark.actions.TestRewriteDataFilesBase;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.internal.SQLConf;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.apache.hadoop.hive.conf.HiveConf.ConfVars.METASTOREURIS;
+
+public class TestRewriteExtensions extends TestRewriteDataFilesBase {
```
szehon-ho, line 43: Nit: the name sounds a bit weird (that it's rewriting extensions), wonder if "Extensions" is superfluous given the package name or we can clarify it some way?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteExtensions.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @BeforeClass
+  public static void startMetastoreAndSpark() {
```
szehon-ho, line 46: It would be great to re-use SparkExtensionsTestBase somehow

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteExtensions.java
```diff
@@ -0,0 +1,95 @@
...
+
+    List<Object[]> originalData = currentData();
+
+    RewriteDataFiles.Result result =
+        basicRewrite(table)
+            .option(SortStrategy.MIN_INPUT_FILES, "1")
+            .option(SortStrategy.MIN_FILE_SIZE_BYTES, Integer.toString(averageFileSize(table) + 1000))
+            .execute();
+
+    Assert.assertEquals("Should have 1 fileGroup because all files were not correctly partitioned",
```
szehon-ho, line 85: nit: expected and actual are reversed

szehon-ho: Also, I guess we don't need Assert if it's already imported, to be consistent with below assert.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteExtensions.java
szehon-ho: Maybe we can make should extend SparkExtensionsTestBase and make some test utility functions for the rewrite test helpers

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -195,6 +124,33 @@ public void testBinPackWithFilter() {
...
+
+    List<Object[]> originalData = currentData();
+
+    RewriteDataFiles.Result result =
+        basicRewrite(table)
+            .option(SortStrategy.MIN_INPUT_FILES, "1")
+            .option(SortStrategy.MIN_FILE_SIZE_BYTES, Integer.toString(averageFileSize(table) + 1000))
+            .execute();
+
+    Assert.assertEquals("Should have 1 fileGroup because all files were not correctly partitioned",
```
szehon-ho, line 143: Same as above (actual, expected reversed).  And also "Assert" unnecessary.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
```diff
@@ -67,12 +68,18 @@ public Table table() {
...
+
+      // Invoke a shuffle if the partition spec of the incoming partition does not match the table
+      String distributionMode = requiresRepartition ? DistributionMode.RANGE.modeName() :
```
szehon-ho, line 74: And also general question here for learning, what's the reason for range and not hash?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
RussellSpitzer: Yes, we just check the first file because we are guaranteed by the rewriteAction code under the hood to only give us groups of files where every file has the same spec and partition value. They are just not guaranteed to be the same as or compatible with the current spec.

RussellSpitzer: https://github.com/apache/iceberg/blob/c6e68d832113266994c0eb06d342e032fc9cbca3/core/src/main/java/org/apache/iceberg/actions/BaseRewriteDataFilesAction.java#L258

RussellSpitzer: I should also note that this will end up grouping all files without compatible partition info into the same "partition" which will get sub divided randomly based on max file group size for rewrite operations.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
RussellSpitzer: Range is generally safer than hash when we don't know what the underlying distribution really is. For example if we have many small partitions a hash distribution would be a spark shuffle task per partition, with range it will attempt to sample and more evenly distribute the work within the spark shuffle (this also can take advantage of sort columns if they exist).

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteExtensions.java
RussellSpitzer: I forget this all the time

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
szehon-ho: Sounds good, do you think its worthwhile to add a comment, if not already?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteExtensions.java
szehon-ho: Yea I'm ok to not do it, it may not be directly required to reproduce the problem at hand?  But I might not have all the context

RussellSpitzer: Yeah basically we can reproduce this issue whenever we go from one partition spec to another incompatible spec. So going between unpartitioned and partitioned is a fine repo as well.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -453,6 +409,7 @@ public void testBinPackCombineMediumFiles() {
     assertEquals("Rows must match", expectedRecords, actualRecords);
   }
 
+
```
rdblue, line 412: Nit: unnecessary newline.

rdblue (approved):
Looks good to me.

RussellSpitzer:
Ok I think i'm gonna remove the refactor here, just keep only the identity test for now. I think the right thing to do for bucketing (and other functions) is probably to add more tests to extensions itself but no need to mix them up here.

gradle.properties
```diff
@@ -24,4 +24,4 @@ systemProp.knownSparkVersions=2.4,3.0,3.1,3.2
 systemProp.defaultScalaVersion=2.12
 systemProp.knownScalaVersions=2.12,2.13
 org.gradle.parallel=true
-org.gradle.jvmargs=-Xmx768m
+org.gradle.jvmargs=-Xmx1g
```
kbendick, line 27: You might consider changing to the following, so that the number of classes loaded in memory will be capped (unused classes will be unloaded). This has helped CI in both Spark, as well as some things I've used elsewhere).

```
org.gradle.jvmargs=-Xmx1g -XX:MaxMetaspaceSize=512m
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java
RussellSpitzer: After thinking about this for a while I think this is only true for V1 tables and we probably have a bug in V2 tables in our group by partition code in base rewrite action. Consider

```
So for example say I have a V1 table, Columns A, B
I first partition on a
spec1 = Identity<A>
then i drop that and partition on B
spec2 = Void<A>, Identity<B>

In V2 it works like this
spec1 = Identity<A>
spec2 = Identity<B>

In this case while the specs have the same shape, the values do not mean the same thing so our grouping is going to be incorrect.

Comment (kbendick) [reply spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/Spark3BinPackStrategy.java:71]: I personally think it would be, as it was clear to me from jut looking at this class that `filesToRewrite` were already grouped that way. But I'd leave that up to you due to concerns fo comment rot etc.

File: spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java [hunk 9]
```diff
@@ -195,6 +124,33 @@ public void testBinPackWithFilter() {
...
+    Assert.assertEquals("Should have 1 fileGroup because all files were not correctly partitioned",
+        result.rewriteResults().size(), 1);
+
+    List<Object[]> postRewriteData = currentData();
+    assertEquals("We shouldn't have changed the data", originalData, postRewriteData);
+
+    shouldHaveSnapshots(table, 2);
+    shouldHaveACleanCache(table);
+    shouldHaveFiles(table, 4);
+  }
```
kbendick, line 152: Question / thought: Would there be benefit in adding a test that starts out with an identity transform on one column, then switches to an identity transform on another column, and then binpack?

We wouldn't have the complication of having to refactor so much, like if we wanted to use multiple transforms. But we could still check data with. multiple partition specs in it (possibly doing 3 columns worth of changes).

Fully up to you.

kbendick (approved):
A few questions and a suggestion on the gradle jvmargs (which I can also do in a follow up PR if we'd like to keep them separate), but this looks great.

Thanks @RussellSpitzer!

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
RussellSpitzer: I made sure this particular test (just made some mods) will necessarily make 320+ files if the shuffle isn't appropriately applied. I'm going to add some more tests in a follow issue to deal with the partition grouping and V2 tables.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
RussellSpitzer: We are using the Assert assertEquals and not the row comparison SparkTestBase.assertEquals here

RussellSpitzer:
Thanks everyone for the reviews! I'll work on the V2 issue in another PR

PR #4285, opened 2022-03-07
Python: Add doctest to tox

### LINKED ISSUE #4284: Add doctest for Python
I was going through the Python code, and noticed that not all examples are valid. For Python there is Doctest to automatically check the examples in the docstring: https://docs.python.org/3/library/doctest.html This will make sure that our examples are up to date, and that can do simple assertions in the docstring 👍🏻

Fokko:
Closes #4284 

While going through the Python code, I noticed that not all the examples are valid. I've added doctest to check the examples in the CI.

Some observations:
- The s3 tests are currently failing because the bucket is invalid (we could add some credentials to the CI to read/write to a single bucket).
- It looks like the requirements weren't installed.


samredai (commented):
This is great, thanks @Fokko! I left a few comments.

python/setup.py
```diff
@@ -19,7 +19,14 @@
 
 setup(
     name="py-iceberg",
-    install_requires=[],
+    install_requires=[
```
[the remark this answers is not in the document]
samredai: Another thing I just realized is that once you add `pyarrow` to the setup.cfg, you can remove it from the deps list in tox.ini

samredai:
We should also move the `extras_require` in `setup.py` over to `setup.cfg` so that we're not setting package info in two places. This stuff collides easily and causes confusion. We can then just have a barebones `setup.py` and have everything contained in `setup.cfg`:

*setup.py*
```py
from setuptools import setup

setup()
```
Let me know what you think.

python/setup.py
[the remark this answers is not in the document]
Fokko: Hey @samredai thanks for the comments. I just moved the dependencies from the `setup.py` to the `setup.cfg`. I've took them from the `python_legacy` indeed, and added the ones that are required to let the tests pass. I've also removed pyarrow from `tox.ini`.

python/src/iceberg/expressions.py
```diff
@@ -53,7 +54,7 @@ class Operation(Enum):
     AND = auto()
     OR = auto()
 
-    def negate(self):
+    def negate(self) -> Operation:
```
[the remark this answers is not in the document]
Fokko: Personally, I like to omit the quotes and have the import. But I'm fine either way.

python/src/iceberg/io/pyarrow.py
```diff
@@ -43,9 +43,14 @@ class PyArrowFile(InputFile, OutputFile):
     Examples:
         >>> from iceberg.io.pyarrow import PyArrowFile
         >>> input_file = PyArrowFile("s3://foo/bar.txt")
-        >>> file_content = input_file.open().read()  # Read the contents of the PyArrowFile instance
+        >>> # Read the contents of the PyArrowFile instance
+        >>> # Make sure that you have permissions to read/write
+        >>> # file_content = input_file.open().read()
```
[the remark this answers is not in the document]
Fokko: Let me dive into this and see if I can get this working 👍🏻

Fokko:
Hey @samredai thanks for the feedback. I think we can remove the `setup.py` all together. Less is more :)

python/setup.cfg
```diff
@@ -42,6 +42,16 @@ package_dir =
...
 packages = find:
 python_requires = >=3.7
+install_requires =
+    pytz
+    retrying
+    mmh3
+    fastavro>=1.3.2<1.4.0
+    hmsclient==0.1.1
+    boto3
+    pyarrow
```
samredai, line 52: A few of these will be added [here](https://github.com/apache/iceberg/pull/4262/files#diff-1f5419f88832f5bdcc37bc3ef9cfd11a714489c537227ec469756177a2b1578eR45) as part of PR #4262 and some will be part of upcoming PR's (like boto3 being added as part of a `BotoFileIO` PR and hmsclient being added as part of a `HiveCatalog` PR). I think this list can be reduced to just `pyarrow` for now.

python/setup.cfg
```diff
@@ -42,6 +42,16 @@ package_dir =
...
+[options.extras_require]
+dev = tox-travis==0.12; pytest
```
samredai, line 54: Can this be formatted just like the `install_requires`?
```
[options.extras_require]
dev =
    tox-travis==0.12
    pytest
```

samredai (commented):
Looks great but I left 2 small comments.

python/setup.cfg
samredai: Good point then I think it's worth adding this to the `extras_require` block, so something like this:
```
[options.extras_require]
arrow =
    pyarrow
```
This would allow someone to do `pip3 install iceberg[arrow]` which will provide all of the iceberg core dependencies + optional arrow dependencies and as we expand this, something like `pip3 install iceberg[arrow,jdbc]`. This still allows you to just do `pip3 install iceberg` if you already have a specific pyarrow version in your environment that you want to use instead.

python/setup.cfg
Fokko: Yes, much nicer!

python/setup.cfg
Fokko: Hey @samredai @rdblue Great point, and I agree. Pulling in boto might be too much if you don't use it at all, same for Hive.

Ps. Maybe later on we also might want to look into a full dependency management system for the python codebase. Something like [Poetry](https://python-poetry.org/) or [PDM](https://pdm.fming.dev/). We don't really pin versions of packages, which might cause problems when a downstream package is released with a breaking change (which unfortunately [still happens](https://github.com/dbt-labs/dbt-core/issues/4745) quite a lot).

samredai (approved):
LGTM, thanks @Fokko!

rdblue (approved):

rdblue:
Looks great! Thank you for getting this working, @Fokko!

I really like having doctests enforced. This will be useful for the transforms PR.

PR #4287, opened 2022-03-08
Infra - Trigger CI when pushing to a release branch or tag

kbendick:
We'd like to run CI to completion when we push to a release branch.

Release branches are of the form `0.x.y` right now, so matching on `0.**`

For release tags, they are normally `apache-iceberg-**`, so I've updated the CI suites to run on push to a tag of that format.

cc @rdblue @snazy @nastra

### DIFF
--- a/.github/workflows/flink-ci.yml
+++ b/.github/workflows/flink-ci.yml
@@ -22,8 +22,9 @@ on:
   push:
     branches:
     - 'master'
+    - '0.**'
     tags:
-    - '**'
+    - 'apache-iceberg-**'
   pull_request:
     paths-ignore:
     - '.github/workflows/python-ci.yml'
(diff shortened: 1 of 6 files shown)


rdblue:
Sorry, I think I was wrong. Branches are named after the release version, like `0.13.x`. The release tags are named `apache-iceberg-**`, but we already run on tags. So I think it should be `0.**`?

kbendick:
> Sorry, I think I was wrong. Branches are named after the release version, like `0.13.x`. The release tags are named `apache-iceberg-**`, but we already run on tags. So I think it should be `0.**`?

No worries. I updated it to be branches of `0.**` (not sure how to require that it be `0.x.y` formatted and realistically before this we were running CI on push to any branch).

I also updated the tags to be `apache-iceberg-**` since that is the format for release tags.

rdblue (approved):

snazy (approved):
LGTM

nastra (approved):

PR #139, opened 2019-03-20
ORC support integration for Spark 2.4.0

edgarRd:
I noticed that a large portion of the support for ORC in https://github.com/Netflix/iceberg/commit/c59138eadcfb4eb2558a15f73c00b54528d02e12#diff-545c12970ccace1ba019c99192569301 was missing. 
I've adapted that code to be supported with Spark 2.4.0 since the API for `UnsafeWriter` changed.

Test cases are passing for reads and writes.

PR #4291, opened 2022-03-08
ORC:Optimize table properties usage for ORC

hililiwei:
Based on https://github.com/apache/iceberg/pull/3810.
Refer:https://github.com/apache/iceberg/pull/3810#issuecomment-1061197742

> For Parquet the Hadoop configuration is used to pass options into the data file. It is not used as the source of table configuration. Table configuration properties should never come from the Hadoop Configuration.
> 
> The steps should be:
> 
> 1. Get the Hadoop configuration, if it is present in the InputFile or OutputFile. If not, default it
> 2. Get config from the builder and the table properties
> 3. Set configuration from step 2 on the Hadoop config
> 4. Create the reader or writer with the Hadoop config
> 
> The only configuration coming from the Hadoop Configuration itself is whatever was in the environment.
> 
> One possibly confusing thing is that we also set config values directly in the Hadoop configuration. That handles cases where the user wants to pass config properties that are not standardized in Iceberg. So you could use `set("parquet.bloom.filter.enabled#id", "true")` for example. Standardized table settings should override these.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -143,6 +143,10 @@ public class TableProperties {
   public static final String DELETE_ORC_BLOCK_SIZE_BYTES = "write.delete.orc.block-size-bytes";
   public static final long ORC_BLOCK_SIZE_BYTES_DEFAULT = 256L * 1024 * 1024; // 256 MB
 
+  public static final String ORC_VECTOR_ROW_BATCH_SIZE = "write.orc.vectorized-row-batch-size";
+  public static final String DELETE_ORC_VECTOR_ROW_BATCH_SIZE = "write.delete.orc.vectorized-row-batch-size";
+  public static final int ORC_VECTOR_ROW_BATCH_SIZE_DEFAULT = 1024;
+
   public static final String SPLIT_SIZE = "read.split.target-size";
   public static final long SPLIT_SIZE_DEFAULT = 128 * 1024 * 1024; // 128 MB
 
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -143,9 +143,9 @@ public class TableProperties {
   public static final String DELETE_ORC_BLOCK_SIZE_BYTES = "write.delete.orc.block-size-bytes";
   public static final long ORC_BLOCK_SIZE_BYTES_DEFAULT = 256L * 1024 * 1024; // 256 MB
 
-  public static final String ORC_VECTOR_ROW_BATCH_SIZE = "write.orc.vectorized-row-batch-size";
-  public static final String DELETE_ORC_VECTOR_ROW_BATCH_SIZE = "write.delete.orc.vectorized-row-batch-size";
-  public static final int ORC_VECTOR_ROW_BATCH_SIZE_DEFAULT = 1024;
+  public static final String ORC_WRITE_BATCH_SIZE = "write.orc.vectorized.batch-size";
+  public static final String DELETE_ORC_WRITE_BATCH_SIZE = "write.delete.orc.vectorized.batch-size";
+  public static final int ORC_WRITE_BATCH_SIZE_DEFAULT = 1024;
 
   public static final String SPLIT_SIZE = "read.split.target-size";
   public static final long SPLIT_SIZE_DEFAULT = 128 * 1024 * 1024; // 128 MB
--- a/orc/src/main/java/org/apache/iceberg/orc/ORC.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORC.java
@@ -69,7 +69,7 @@ import org.apache.orc.storage.ql.exec.vector.VectorizedRowBatch;
 public class ORC {
 
   /**
-   * @deprecated use {@link TableProperties#ORC_VECTOR_ROW_BATCH_SIZE} instead
+   * @deprecated use {@link TableProperties#ORC_WRITE_BATCH_SIZE} instead
    */
   @Deprecated
   private static final String VECTOR_ROW_BATCH_SIZE = "iceberg.orc.vectorbatch.size";
@@ -174,8 +174,8 @@ public class ORC {
       }
 
       // for compatibility
-      if (conf.get(VECTOR_ROW_BATCH_SIZE) != null && config.get(TableProperties.ORC_VECTOR_ROW_BATCH_SIZE) == null) {
-        config.put(TableProperties.ORC_VECTOR_ROW_BATCH_SIZE, conf.get(VECTOR_ROW_BATCH_SIZE));
+      if (conf.get(VECTOR_ROW_BATCH_SIZE) != null && config.get(TableProperties.ORC_WRITE_BATCH_SIZE) == null) {
+        config.put(TableProperties.ORC_WRITE_BATCH_SIZE, conf.get(VECTOR_ROW_BATCH_SIZE));
       }
 
       // Map Iceberg properties to pass down to the ORC writer
@@ -223,7 +223,7 @@ public class ORC {
         Preconditions.checkArgument(blockSize > 0, "Block size must be > 0");
 
         int vectorizedRowBatchSize = PropertyUtil.propertyAsInt(config,
-            TableProperties.ORC_VECTOR_ROW_BATCH_SIZE, TableProperties.ORC_VECTOR_ROW_BATCH_SIZE_DEFAULT);
+            TableProperties.ORC_WRITE_BATCH_SIZE, TableProperties.ORC_WRITE_BATCH_SIZE_DEFAULT);
         Preconditions.checkArgument(vectorizedRowBatchSize > 0, "VectorizedRow batch size must be > 0");
 
         return new Context(stripeSize, blockSize, vectorizedRowBatchSize);
@@ -239,7 +239,7 @@ public class ORC {
             TableProperties.DELETE_ORC_BLOCK_SIZE_BYTES, dataContext.blockSize());
 
         int vectorizedRowBatchSize = PropertyUtil.propertyAsInt(config,
-            TableProperties.DELETE_ORC_VECTOR_ROW_BATCH_SIZE, dataContext.vectorizedRowBatchSize());
+            TableProperties.DELETE_ORC_WRITE_BATCH_SIZE, dataContext.vectorizedRowBatchSize());
 
         return new Context(stripeSize, blockSize, vectorizedRowBatchSize);
       }
--- a/orc/src/main/java/org/apache/iceberg/orc/ORC.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORC.java
@@ -57,6 +57,7 @@ import org.apache.iceberg.mapping.NameMapping;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.util.ArrayUtil;
+import org.apache.iceberg.util.PropertyUtil;
 import org.apache.orc.OrcConf;
 import org.apache.orc.OrcFile;
 import org.apache.orc.OrcFile.ReaderOptions;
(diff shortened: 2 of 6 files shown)


hililiwei:
cc @rdblue @openinx @zhongyujiang

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -191,28 +198,26 @@ private Context(long stripeSize, long blockSize) {
         this.blockSize = blockSize;
       }
 
-      static Context dataContext(Configuration conf) {
-        long stripeSize = TableProperties.ORC_STRIPE_SIZE_BYTES_DEFAULT;
-        stripeSize = conf.getLong(OrcConf.STRIPE_SIZE.getAttribute(), stripeSize);
-        stripeSize = conf.getLong(OrcConf.STRIPE_SIZE.getHiveConfName(), stripeSize);
-        stripeSize = conf.getLong(TableProperties.ORC_STRIPE_SIZE_BYTES, stripeSize);
+      static Context dataContext(Map<String, String> conf) {
```
openinx, line 201: Could we always use `config`  to indicate that we are referring a configure `Map<String, String>` ?  The `conf` looks like a hadoop configuration in my mind.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcReadMetadataColumns.java
```diff
@@ -123,7 +124,7 @@ public void writeFile() throws IOException {
         // write in such a way that the file contains 10 stripes each with 100 rows
         .set("iceberg.orc.vectorbatch.size", "100")
         .set(OrcConf.ROWS_BETWEEN_CHECKS.getAttribute(), "100")
-        .set(OrcConf.STRIPE_SIZE.getAttribute(), "1")
```
openinx, line 126: So looks like the default value of table property `write.orc.stripe-size-bytes` will override the set value from original ORC config key (`orc.stripe.size` ) ?

hililiwei: Yes, using the original ORC config key directly here is invalid and will be overwritten by the table property or by the table property default value.

hililiwei: If the standard iceberg table properties is not found, do we feedback to try the origin ORC key?

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -143,6 +143,10 @@ private TableProperties() {
   public static final String DELETE_ORC_BLOCK_SIZE_BYTES = "write.delete.orc.block-size-bytes";
   public static final long ORC_BLOCK_SIZE_BYTES_DEFAULT = 256L * 1024 * 1024; // 256 MB
 
+  public static final String ORC_VECTOR_ROW_BATCH_SIZE = "write.orc.vectorized-row-batch-size";
```
openinx, line 146: We are replacing the old `iceberg.orc.vectorbatch.size` with the newly introduced `writer.orc.vectorized-row-batch-size` here , right ?   In this way,  the old table using the old config key won't take any effect in the next release.  I think we'd better to keep the original approach.

hililiwei: `iceberg.orc.vectorbatch.size` doesn't fit our naming conventions. Previously, it was a private variable and was not exposed as a public table configuration. In my humble opinion, if we formally expose it as a table configuration,
, would it be better to use a standard name?

> Config naming
> 1.Use - to link words in one concept
  >     For example, preferred convection access-key-id rather than access.key.id
> 2.Use . to create a hierarchy of config groups
  >     For example, s3 in s3.access-key-id, s3.secret-access-key

openinx: It's true that the original name `iceberg.orc.vectorbatch.size` is not an iceberg name style. This config key was introduced in this PR: https://github.com/apache/iceberg/pull/139/files.  In theory, people could add this config key-value in their enviroment  hadoop configuration file.  But after we merged this PR, the key won't take any effect even if we use the old key in hadoop environment. 

I will suggest to make it deprecated and then introduce the correct table property. But I think we still need to be able to read the value from the old key.

openinx: It's true that the original name `iceberg.orc.vectorbatch.size` is not an iceberg name style. This config key was introduced in this PR: https://github.com/apache/iceberg/pull/139/files.  In theory, people could add this config key-value in their enviroment  hadoop configuration file.  But after we merged this PR, the key won't take any effect even if we use the old key in hadoop environment. 

I will suggest to make it deprecated and then introduce the correct table property. But I think we still need to be able to read the value from the old key.

hililiwei: Okay, I'll restore it, and let's do it separately later if we think it's necessary.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -186,35 +198,47 @@ public long blockSize() {
...
-        blockSize = conf.getLong(TableProperties.ORC_BLOCK_SIZE_BYTES, blockSize);
+        long blockSize = PropertyUtil.propertyAsLong(config,
+            TableProperties.ORC_BLOCK_SIZE_BYTES, TableProperties.ORC_BLOCK_SIZE_BYTES_DEFAULT);
+        Preconditions.checkArgument(blockSize > 0, "Block size must be > 0");
 
-        return new Context(stripeSize, blockSize);
+        int vectorizedRowBatchSize = PropertyUtil.propertyAsInt(config,
+            TableProperties.ORC_VECTOR_ROW_BATCH_SIZE, TableProperties.ORC_VECTOR_ROW_BATCH_SIZE_DEFAULT);
+        vectorizedRowBatchSize = PropertyUtil.propertyAsInt(config,
+            VECTOR_ROW_BATCH_SIZE, vectorizedRowBatchSize);
```
hililiwei, line 223: For compatibility.

core/src/main/java/org/apache/iceberg/TableProperties.java
hililiwei: OK, I added the standardized table configuration and tried to get values from the old key for compatibility, please take a look, thx.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
openinx: The old key which was configured in environment hadoop configuration still won't be parsed in the current design ?  Because we just parse the table properties and never consider the hadoop configuration file.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -153,30 +154,46 @@ public WriteBuilder metricsConfig(MetricsConfig newMetricsConfig) {
...
+      }
+
+      for (Map.Entry<String, String> entry : config.entrySet()) {
+        conf.set(entry.getKey(), entry.getValue());
+      }
+
+      // for compatibility
+      if (conf.get(VECTOR_ROW_BATCH_SIZE) != null && config.get(TableProperties.ORC_VECTOR_ROW_BATCH_SIZE) == null) {
+        config.put(TableProperties.ORC_VECTOR_ROW_BATCH_SIZE, conf.get(VECTOR_ROW_BATCH_SIZE));
+      }
```
hililiwei, line 179: Get the value from the Hadoop conf for compatibility. @openinx

orc/src/main/java/org/apache/iceberg/orc/ORC.java
hililiwei: updated.

openinx (approved):
Looks good to me now. Thanks @hililiwei for the contribution !

zhongyujiang:
Sorry for the delay.

About backwards compatibility, there is another issue I worried about. Users used to be able to set ORC props in Iceberg table properties like "orc.stripe.size" = '1', which could be passed directly to ORC to control stripe size, now we introduced standardized Iceberg ORC_STRIPE_SIZE_BYTES, but I think still we should not ignore those ORC properties which already had been set in table properties by users. I think we can use priorities like #3810, just pull the props from `config` instead of `conf`. What do you think? 

The rest looks good to me.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
rdblue: Compatibility with what? This was unreleased. I don't think that Iceberg should source configuration from the Hadoop Configuration.

core/src/main/java/org/apache/iceberg/TableProperties.java
rdblue: I agree that we should deprecate and keep supporting `iceberg.orc.vectorbatch.size` from the Hadoop conf since that slipped through review. But we can still clean this up with a reasonable table property.

The read property is `read.orc.vectorization.batch-size` so I think this should match: `write.orc.vectorization.batch-size`

orc/src/main/java/org/apache/iceberg/orc/ORC.java
rdblue: Okay, I see. This looks fine to me.

rdblue: This looks good to me.

rdblue: It looks like the old config could also be used in `set` on this builder. Should this support that case as well?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcReadMetadataColumns.java
rdblue: I think the table property should override the Hadoop configuration property. We should probably also support the old `iceberg.orc.vectorbatch.size` property passed in like this. I think the table property should override it.

openinx:
@zhongyujiang As we will still copy the configure key&values to the hadoop configuration here https://github.com/apache/iceberg/pull/4291/files#diff-34d7fce4c1d9417fa9342247cf5ace0636cd86591efbed0555ae80197f79303dR172-R174 .  So I think the current patch could still meet your requirement about `Users used to be able to set ORC props in Iceberg table properties like "orc.stripe.size" = '1', which could be passed directly to ORC to control stripe size`, right ?

hililiwei:
@openinx One thing to note. Although the k-v of config is copied to hadoop conf, if a standardized iceberg key, such as "orc.stripe.size", is set in conf using only the origin key, instead of the standardized key "write.orc.stripe-size-bytes", it will eventually be overwritten by iceberg's default value.
@zhongyujiang seems to be talking about this.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
hililiwei: Hi Ryan, what does 'old config' mean?

openinx:
@hililiwei  Would you mind to fix those travis CI issues ?

hililiwei:
> @hililiwei  Would you mind to fix those travis CI issues ? 

Done.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
openinx: I think @rdblue means:  we can also set the `iceberg.orc.vectorbatch.size` key in the iceberg table properties (Because we just `set` this key value into the hadoop configuration in the original version and then read this key from the hadoop configuration.)

We should also be able to read this config from `config` table properties if people just set it as the old approach. 

In summary,  we shouldn't need to change those unit tests config setting ([1], [2], [3]). if we keep the old usage.

[1]. https://github.com/apache/iceberg/pull/4291/files#diff-37a827d65fdb4bbf9ce2247d6839a846e27c371eb0bad58a47a4bc924170e2f7L124-L126
[2]. https://github.com/apache/iceberg/pull/4291/files#diff-8878d128d226b30d2941cd46b0e8c9aa193d752355288a6dc928af00d35bf811L124-L126
[3], https://github.com/apache/iceberg/pull/4291/files#diff-86cc5e259c85e8e36c90347a4623fd49e2e7a79d81ef4d8d3c8527111bfc8335L124-L126

hililiwei: thx. I tried to reverted unit test changes. and read it from `config`.

PR #4294, opened 2022-03-09
Core: Use existing SnapshotRefParser when serializing ref information in TableMetadata Parser

amogh-jahagirdar:
This change is a follow on to @rdblue change in https://github.com/apache/iceberg/pull/4260 . Since we have an existing SnapshotRefParser I think we can leverage that and reduce some of the implementation in TableMetadataParser.

Let me know your thoughts! @rdblue  @jackye1995

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableMetadataParser.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadataParser.java
@@ -101,10 +101,6 @@ public class TableMetadataParser {
   static final String PROPERTIES = "properties";
   static final String CURRENT_SNAPSHOT_ID = "current-snapshot-id";
   static final String REFS = "refs";
-  static final String TYPE = "type";
-  static final String MIN_SNAPSHOTS_TO_KEEP = "min-snapshots-to-keep";
-  static final String MAX_SNAPSHOT_AGE_MS = "max-snapshot-age-ms";
-  static final String MAX_REF_AGE_MS = "max-ref-age-ms";
   static final String SNAPSHOTS = "snapshots";
   static final String SNAPSHOT_ID = "snapshot-id";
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/iceberg/TableMetadataParser.java
```diff
@@ -252,22 +248,7 @@ private static void toJson(TableMetadata metadata, JsonGenerator generator) thro
...
-      }
-      if (ref.maxSnapshotAgeMs() != null) {
-        generator.writeNumberField(MAX_SNAPSHOT_AGE_MS, ref.maxSnapshotAgeMs());
-      }
-      if (ref.minSnapshotsToKeep() != null) {
-        generator.writeNumberField(MIN_SNAPSHOTS_TO_KEEP, ref.minSnapshotsToKeep());
-      }
-
-      generator.writeEndObject();
+      SnapshotRefParser.toJson(refEntry.getKey(), refEntry.getValue(), generator);
```
rdblue, line 251: I think that the `writeObjectFieldStart` and `writeEndObject` should be here in this function. Moving it to `toJson` in the other class creates an implicit requirement on the state of the `JsonGenerator`. It _must_ be writing an object. We avoid depending on the state of objects passed in where possible, **especially** for methods that are public.

Instead, the state of the generator should always be that it is ready to write any JSON value.

amogh-jahagirdar: Updated. We can write the name field and leverage the existing toJson in SnapshotRefParser. I think this approach works because the ref can be serialized independently without any reliance on state. In TableMetadataParser we can serialize the name + ref entity together

rdblue (approved):

PR #4300, opened 2022-03-09
Docs: Add Drill to Documentation

cgivre:
This PR adds a link to the documentation for Apache Drill's connector for Iceberg tables.

### DIFF
--- a/docs/drill/_index.md
+++ b/docs/drill/_index.md
@@ -1,0 +1,23 @@
+---
+title: "Drill"
+bookIconImage: ../img/drill-logo.png
+bookFlatSection: true
+weight: 430
+bookExternalUrlNewWindow: https://drill.apache.org/docs/iceberg-format-plugin/
+---
+<!--
+ - Licensed to the Apache Software Foundation (ASF) under one or more
+ - contributor license agreements.  See the NOTICE file distributed with
+ - this work for additional information regarding copyright ownership.
+ - The ASF licenses this file to You under the Apache License, Version 2.0
+ - (the "License"); you may not use this file except in compliance with
+ - the License.  You may obtain a copy of the License at
+ -
+ -   http://www.apache.org/licenses/LICENSE-2.0
+ -
+ - Unless required by applicable law or agreed to in writing, software
+ - distributed under the License is distributed on an "AS IS" BASIS,
+ - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ - See the License for the specific language governing permissions and
+ - limitations under the License.
+ -->


docs/drill/_index.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "Drill"
+bookIconImage: ../img/drill-logo.png
```
rdblue, line 3: Does this image need to be added as well?

docs/drill/_index.md
```diff
@@ -0,0 +1,23 @@
...
+bookFlatSection: true
+weight: 430
+bookExternalUrlNewWindow: https://drill.apache.org/docs/iceberg-format-plugin/
```
rdblue, line 6: Side question: does Drill use Iceberg's Catalog interface? From the documentation, it looks like it tries to find table metadata from the file system, which is dangerous.

docs/drill/_index.md
cgivre: @rdblue I wasn't sure where exactly to add the image.  I submitted a PR for the Iceberg doc site [1] with the image.  Is that the wrong place?

[1]: https://github.com/apache/iceberg-docs/pull/62

rdblue: Looks good to me. Thanks!

docs/drill/_index.md
cgivre: @rdblue I'm not sure.  @vvysotskyi is the author of the Drill/iceberg plugin.  We're always looking to make Drill better, so if you have suggestions, we'd love to hear them.  With that said, Drill tries to be as schema free as possible.  This may be what you're seeing.

kbendick: The concern isn't with schema-free necessarily, but with having the source of truth come from a location on the file store.

Iceberg requires an atomic compare and swap operation to be able to correctly determine the correct current metadata files at all times. Where this becomes dangerous is with multiple concurrent writers when using file based approach.

However, it seems like there's maybe a zookeeper component involved in the Iceberg config? https://github.com/apache/drill/blob/00d0cb18e63cfd010038ed590c725fbfba86f817/distribution/src/main/resources/drill-metastore-override-example.conf#L46-L64

cgivre: @kbendick Thanks for looking into this.  Drill config info is all stored in Zookeeper.  Drill also uses various internal persistent stores for storing things that well... need to be stored.  I'd be curious for you to try it out to see what you think of the integration.   I'm not an Iceberg expert myself so feedback is greatly appreciated.

vvysotskyi: @rdblue, Drill uses the `Tables` interface and passes table location only. If I understand docs correctly, it shouldn't cause any issues, the single case is moving the table from one location to another one, but in this case, it should be enough just to specify a new location within the Drill query.

vvysotskyi: To be able to determine whether the table is an Iceberg table, Drill checks that folder has a `metadata` directory. I think it is a single place where we rely on the implementation.

rdblue: The tables interface is only a way to work with Hadoop tables, which are primarily for testing purposes.

Most tables are tracked using the Catalog interface, which supports a variety of metastore backends. I recommend using that because Hadoop tables aren't safe most data stores, including S3 and local disk.

docs/drill/_index.md
cgivre: @rdblue Are we waiting for something to merge this?  Thanks!

PR #2496, opened 2021-04-20
[#2039] Support default value semantics - API changes

shenodaguirguis:
# Summary

Iceberg schema currently does not support default values which imposes a challenge on reading Hive tables written in avro format and have default values (see issue 2039). Specifically, if a field has a non-null default value, it is mapped to a required field - with no default value - in iceberg. Thus, upon reading rows where this field is not manifested, an IllegalArgumentException (Missing required field) will be thrown. Further, default values of nullable fields are lost silently. That is because nullable fields with default values are mapped to optional fields with no default values, and thus null is returned when the field is absent, instead of the default value. This document describes how to support the default values semantics in Iceberg schema to resolve these issues.

# Problem
## Default values are lost
Default values are specified using the Avro schema keyword “default”. For example, the following is an example of an Avro string field with default value “unknown”:

`{"name": "filedName", "type": "string", "default": “unknown”}`

Also, a nullable (optional) avro field can define default value as follows:

`{"name": "fieldName", "type": ["null", "string"], "default": null}`

Please note that nullability is specified via UNION type (i.e., the [“null’, “string”]) and the default value’s type must match the first type in the union. In other words, the following are invalid types:

```
{"name": "fieldName", "type": ["null", "string"], "default": “unnown”}
{"name": "fieldName", "type": ["string", “null”], "default": null}
```

That is, if the default value is of type null, the first type of the field must be the “null”, o.w., if the default value is of type string, the first type of the field must be string.

When converting an avro schema to Iceberg Types, we have 2 cases. If the field is nullable, it maps to an optional NestedField Iceberg type. While if the field is non-nullable, it is mapped to a required NestedField Iceberg type, with no default value, since the NestedFiled does not support default values. In both cases, the default value is lost which leads to a wrong handling of the data when read. In the case of non-nullable fields, error is thrown if the field is not present, whereas in case of null, it goes by as an optional field.
## Where in code this breaks
[...]

PR #4301, opened 2022-03-09
Docs: Default value support feature specification

rzhang10:
This is to continue the work of #2496 , taking into account the discussions that happened there.

This spec change adds in the definition and semantics of `default value`, it specifies what default value could be defined for what data types. It also details how default value is serialized inside/along with table schema using JSON encoding.


wmoustafa:
Could you please update the PR description to give a high level description of the changes introduced in this PR (e.g., the fact that it updates the documentation to describe the default value spec, and which aspects of the spec are covered)?
@rdblue @RussellSpitzer

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+
+Currently, when a default value for a column or nested field is set, it is considered an incompatible change to change it to another value. However, changing default values is allowed when calling the `allowIncompatibleChanges` API explicitly. Changing default values is discouraged since the occurrence of a rewrite may determine whether the new or old default value is returned during the read.
+
+Default values are encoded in JSON format. The representation depends on the type of the corresponding field. The mapping of types and their corresponding default value JSON representation is described in the following table.
+
+| type               | json type     | example                                | note                                                                                                                                                                                                                                         |
```
RussellSpitzer, line 203: This probably would go in the Serialization Json Format appendix down at the bottom

format/spec.md
```diff
@@ -949,25 +977,25 @@ Schemas are serialized as a JSON object with the same fields as a struct in the
 
 Types are serialized according to this table:
 
-|Type|JSON representation|Example|
```
RussellSpitzer, line 952: Probably we should keep this with the same whitespace

format/spec.md
RussellSpitzer: I think we should also match the other tables which don't use vertical alignment

format/spec.md
```diff
@@ -158,7 +158,7 @@ For the representations of these types in Avro, ORC, and Parquet file formats, s
 
 #### Nested Types
 
-A **`struct`** is a tuple of typed values. Each field in the tuple is named and has an integer id that is unique in the table schema. Each field can be either optional or required, meaning that values can (or cannot) be null. Fields may be any type. Fields may have an optional comment or doc string.
+A **`struct`** is a tuple of typed values. Each field in the tuple is named and has an integer id that is unique in the table schema. Each field can be either optional or required, meaning that values can (or cannot) be null. Fields may be any type. Fields may have an optional comment or doc string. Fields can also have a default value (see below).
```
RussellSpitzer, line 161: Do we need to specify that these default values are only retroactive? Ie we still do not allow skipping a required column just because the said column has a default? I haven't read the SQL spec on this or other system's usage, but I assume we want to align on that.

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
 
 For details on how to serialize a schema to JSON, see Appendix C.
 
+#### Default value
+Default values can be assigned to top-level columns or nested fields. Default values are used during schema evolution when adding a new column. The default value is used to read rows belonging to the files that lack the column or nested field prior to the schema evolution.
```
RussellSpitzer, line 197: I think we have to hit a few points very clearly here.

1. Default values can be assigned to any newly created column in an Iceberg Table
2. Default values are only used for files which are missing the new column (Do we explicitly require the files to be written prior to the default value being set? If I import a new file directly into the iceberg table does it get the default? I think it's probably easiest to say yes)
3. New writes to the table do not use Default Values
   a. Required Columns are still required at write time
   b. Optional Columns are still persisted as null if they are missing at write time
4. Changing a default value is basically an undefined behavior
   a. Rows which have been rewritten will permanently keep the default value which was set when the row is rewritten
   b. Rows which have not been rewritten will adjust to the new default value
   c. Because of this we lock changing defaults behind "allowIncompatible changes"
5. Name Mappings are always used in place of defaults. If a file is added with Iceberg Column Ids (via import or what not) we will always use name mapping before attempting to use a column default value.
   
 I think the current text is good but can probably be tightened up.
 
 For example
 
 "Default values can be assigned to any new top-level column added to an Iceberg table." instead of the first two sentences
 "Default Values are only used when reading a data file which is missing the column with the default.

format/spec.md
rdblue: @RussellSpitzer are you saying that all columns are required at write time? It sounds like that's the case. Just because there's a default value doesn't mean you can skip writing a column to a Parquet file. They must be present.

I think that's a good thing to call out!

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
 
 For details on how to serialize a schema to JSON, see Appendix C.
 
+#### Default value
```
rdblue, line 196: Nit: please add a newline to separate the heading from the section text.

format/spec.md
wmoustafa: This sounds like the `INSERT INTO` use case in [this comment](https://github.com/apache/iceberg/pull/2496#issuecomment-1048352760). The support should come from the engine first. Regardless, yes, we cannot support that today.

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+| **`float`**        | **`number`**  | 1.1                                    |                                                                                                                                                                                                                                              |
+| **`double`**       | **`number`**  | 1.1                                    |                                                                                                                                                                                                                                              |
+| **`decimal(P,S)`** | **`string`**  | "0x3162"                               | we use hexadecimal byte literals to encode bytes, with prefix `0x`, the byte array contain the two's-complement representation of the `unscaled` integer value in big-endian byte order, the actual value will be `unscaled * 10 ^ (-scale)` |
```
rdblue, line 210: In Iceberg, we avoid using personal pronouns ("we" or "I") in specs and docs. It is unclear who "we" is, and avoiding it ends up producing more direct documentation that is easily understood.

For example, here you should state that the values is "Stored as the two's-complement big-endian binary using the minimum number of bytes, converted to a hexadecimal string prefixed by 0x".

You should use wording mostly taken from Appendix D and adapted for JSON.

@RussellSpitzer, what do you think about base64 encoding or other representations? Do we care about the size here?

format/spec.md
wmoustafa: > * New writes to the table do not use Default Values
>   a. Required Columns are still required at write time
>   b. Optional Columns are still persisted as null if they are missing at write time

I would skip "New" here. I think default values are not used on the write path at all.

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+| **`string`**       | **`string`**  | "foo"                                  |                                                                                                                                                                                                                                              |
+| **`uuid`**         | **`string`**  | "eb26bdb1-a1d8-4aa6-990e-da940875492c" |                                                                                                                                                                                                                                              |
+| **`fixed(L)`**     | **`string`**  | "0x3162"                               | we use hexadecimal byte literals to encode bytes, with prefix `0x`                                                                                                                                                                           |
```
rdblue, line 217: All text, including notes or column headings, should use sentence case. That is, the first letter should be capitalized. There is no need for complete sentences; be as direct and concise as possible while being clear.

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+| **`timestamptz`**  | **`number`**  | 1646277378000000                       |                                                                                                                                                                                                                                              |
+| **`string`**       | **`string`**  | "foo"                                  |                                                                                                                                                                                                                                              |
+| **`uuid`**         | **`string`**  | "eb26bdb1-a1d8-4aa6-990e-da940875492c" |                                                                                                                                                                                                                                              |
```
rdblue, line 216: Should note that this is a UUID as a string, including `-`. Should this be lower or upper case?

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+|--------------------|---------------|----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **`boolean`**      | **`boolean`** | true                                   |                                                                                                                                                                                                                                              |
+| **`int`**          | **`number`**  | 1                                      |                                                                                                                                                                                                                                              |
```
rdblue, line 206: In most of the appendix, we are more specific about the type even though JSON doesn't distinguish. That is so that people know what to convert to and from. Prefer using "JSON int" or "JSON long" instead of simply "number".

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+| **`struct`**       | **`object`**  | {"a": 1, "foo": "bar"}                 | the keys are the nested fields' names in the struct schema, and the values should be corresponding value literals of the field's type                                                                                                        |
+| **`list`**         | **`array`**   | \[1, 2, 3\]                            | each value should be value literals of the corresponding element type of the list                                                                                                                                                            |
+| **`map`**          | **`object`**  | {"a": 1, "b": 2}                       | the key and value type correspond with the schema of the map, the key (json string) are parsed into the resulting key type of the map schema, an exception will be thrown if parsing fails.                                                  |
```
rdblue, line 221: This encoding doesn't work for non-string keys. I recommend using an array of arrays that is interpreted as a list of key/value pairs.

format/spec.md
```diff
@@ -193,10 +193,38 @@ Notes:
...
+
+Default value is treated as an attribute of a nested field in the table schema when the schema is serialized, see JSON encoding in [Appendix C](#Appendix-C:-JSON-serialization)).
 
 #### Schema Evolution
 
-Schemas may be evolved by type promotion or adding, deleting, renaming, or reordering fields in structs (both nested structs and the top-level schema’s struct).
+Schemas may be evolved by type promotion or adding, deleting, renaming, or reordering fields in structs (both nested structs and the top-level schema’s struct). Setting a default value for a field is also considered a schema evolution.
```
rdblue, line 227: I don't think that this is needed.

Setting a default value can only be done when adding a field, which is already an evolution.

format/spec.md
rdblue: Yeah, no need to reformat this.

format/spec.md
rdblue: 1. Agreed
2. Yes, probably use the default for imported files. But we should state that fields must be written even if there is a default value.
3. I would state this slightly differently:
    * All columns must be written in new data files at write time
    * Optional columns may contain null, required columns must not
    * Optional columns with no default value have an implied default value, `null`
    * Writers are allowed to write the default value if there is no data for a column
4. Changing the default value should not be allowed by the spec, similar to how you cannot add required columns to the spec currently. I don't think that documenting "allow incompatible changes" is a good idea -- that's just an implementation detail if you know what you're doing. Also, we will need to clarify that you may now add an optional column, an optional column with a default value, or a required column with a default value. But you cannot add a required column with no default value.
5. Agreed, but I would phrase it differently again. Name mappings are used to apply IDs for data files. Default values are added after that point, when the data file has its IDs.

rdblue: This just occurred to me and I want to note it somewhere: right now we assume that missing columns are null in a lot of places. We'll need to be careful about the implementation and make sure we catch things like Parquet file filters that assume a missing column is null.

wmoustafa: > * Writers are allowed to write the default value if there is no data for a column

@rdblue could you clarify what you meant by that?

rdblue: Say I have a table with schema `(id bigint NOT NULL, data string DEFAULT "--")`. Then I can write to that table and omit the `data` column:

```
INSERT INTO table VALUES (1), (2), (3);
```

The writer must actually insert `(1, "--"), (2, "--"), (3, "--)`.

This is the row with `id=3` in the behavior test:

```sql
create table default_test (id int);
insert into default_test values (1), (2);
alter table default_test add column data int default 0;
alter table default_test alter column data set default 1000;
insert into default_test (id) values (3);
insert into default_test values (4, null);
alter table default_test alter column data set default 2000;
```
```sql
select * from default_test;
```
```
id | data
-- | --
 1 | 0
 2 | 0
 3 | 1000
 4 | (null)
```

rdblue: > I would skip "New" here. I think default values are not used on the write path at all.

New is doing some useful work. Streaming engines tend to plan writes once and keep running them. And it's okay to do that. If Flink is writing with schema id=2 that is missing a column that was added in schema id=3, it can continue doing that.

format/spec.md
rzhang10: Regarding the alignment, I think my intellij automatically did this reformatting. But this shouldn't affect the rendered md file view, and it seems it alignment also makes the raw file more human-readable too, so I suggest we can keep this change? WDYT?

format/spec.md
rzhang10: Same explanation as above.

format/spec.md
rzhang10: Yes, a json object's key can only be string type, that's why I mentioned the string key will be parsed into corresponding type. But it seems using array of arrays can preserve/enforce the key type information better.

format/spec.md
rdblue: I find that alignment is helpful for editing, but hard to maintain. If IntelliJ does this automatically, then that's good to know.

We already have a mix, so for a new table this is okay. But we should definitely remove the reformatting of an existing table.

format/spec.md
rdblue: It's fine for a new table, but for existing tables we don't want to make the changes any larger than necessary.

format/spec.md
rdblue: @wmoustafa and @RussellSpitzer, thinking about this a bit more, I think we may want to have a second default in the format (but not exposed to users). The first default is immutable and is set when a field is created. That tells Iceberg how to fill in the default value for all files that don't contain it (`id=1` above). The second default would be the one that is used when writing if the column isn't supplied (`id=3` above). The default for existing data files never needs to change, but the default for future writes can change at any time.

What do you think? This only needs to be visible at the format layer. Users would set a default when creating a column and could change it later to affect writes.

RussellSpitzer: @rdblue for the Writer issue. I feel like this is a little ambiguous, are writers required to use the default value with optional columns or can different writers behave differently? For me this is a little odd because we are changing the underlying table schema so I feel like the behavior shouldn't be implementation dependent.

I think we should say that using the default value for unspecified columns is the spec behavior, or say that writers should not use the value for any new values and just use null. I think giving writers the option to choose between these two behaviors is a bit tricky.

format/spec.md
rdblue: Yeah, JSON within JSON within JSON (etc.) is going to be painful. Let's use a different format.

format/spec.md
RussellSpitzer: For decimals I feel like we probably aren't saving that much space here, even with dozens of decimal defaults we would still probably be dominated by other entries in metadata.json and we have to string parse the whole thing before we get this value anyway. I think if we actually compressed metadata.json we probably would do better than smaller tuning like this if we were worried about space.

I think storing decimal values as a string is probably still ok and easier to read for humans.

format/spec.md
RussellSpitzer: I have no strong feelings about the case here, the OpenJDK UUID.fromString method accepts any case and I'd be surprised if any other language parser had a preference on reading. On toString it always uses lower case though. So lower case only?

format/spec.md
wmoustafa: This situation emerges when engines want to support `INSERT INTO` with a subset of columns. Note that the columns must already exist. Now, the `schema evolution` default value may or may not exist (depending on whether the column missing in `INSERT INTO` was added after the table was created (i.e., schema evolution) or existed since the table was first created (i.e., no schema evolution), respectively).

If the `schema evolution` default value exists: I think it is fair to  _reuse the value_ to fill in the missing column values in `INSERT INTO`.

If the `schema evolution` default value does not exists: I think it is fair to  _reuse the current place in the metadata_ to define this default value (upon table creation this time, and not upon schema evolution). This will never conflict with the `schema evolution` default value of the same column because the column _already exists_.

Going by the above, I think it is fine to use the same concept for the default value to serve both use cases.

For the current spec, we may state that we do not support that anyways, and hence default value cannot be leveraged on the write path. It can be a future extension though.

format/spec.md
RussellSpitzer: I agree this would make a very difficult situation, especially for a default struct with a map inside of it. I think we should keep everything as default parsable json as possible.

format/spec.md
wmoustafa: Regarding the change: I believe that both can change. The only reason we cannot support change of `schema evolution` default value right now is that it is not materialized right away. Imagine we issue a complete rewrite upon `ALTER TABLE ADD col  DEFAULT 1`. The default value can be changed without any consequences (even considering `1` was the `schema evolution` one). The fact that we do not issue a full rewrite upon the DDL is an implementation detail in my opinion.

rdblue: > The default value can be changed without any consequences (even considering 1 was the schema evolution one)

@wmoustafa, this is why changing the schema evolution default is allowed as an incompatible change through the API. Mistakes happen and we can allow correcting them (for example, a typo of `DEFAULT 12` instead of `DEFAULT 1`). We just can't allow the default from when the column is added to change through a `SET DEFAULT` command.

wmoustafa: Sorry I should have been clearer. When I said "The default value can be changed without any consequences", I meant in an ideal situation, when there is an immediate rewrite after the `ALTER TABLE t ADD col DEFAULT 1` which will materialize the `1` in the existing rows (similar to how DBMSs do). Assuming we are in this world (`1` is materialized immediately), if later `ALTER TABLE t ALTER col SET DEFAULT 3` takes place, then this can only affect future rows when `INSERT INTO` does not specify that column (since for past rows, `1` is already in the file).  So yes, I think we are saying the same thing which `SET DEFAULT` does not change the values of existing rows, but I was saying it would be the natural behavior even if we combine both concepts to one but assuming a rewrite takes place immediately.

The main problem that I see here is that at the Iceberg API level, I expect both concepts of default values have to be exposed, at least because sometimes we say we can change it freely (for future rows) and sometimes we say we can only change it behind an `allowIncompatibleChanges` API (for existing rows). If we expose two ways or two concepts of default values, it might be unnecessarily complex.

So if I were to choose between options, both of these sounds clean:
1- Default value for `schema evolution` is the same as the one for `INSERT INTO`. None of them can change unless `allowIncompatibleChanges()` is called.
2- Default value for `schema evolution` is the same as the one for `INSERT INTO`. `schema evolution` instance is materialized right away through full rewrite upon the DDL, allowing all other instances (i.e., `INSERT INTO`) to change. From the user's point of view Default value (regardless of which instance it is) just changes without problems.

In both cases, we do not support the `INSERT INTO` use case yet.

rdblue: Replying to @RussellSpitzer:

> For me this is a little odd because we are changing the underlying table schema so I feel like the behavior shouldn't be implementation dependent.

100% agree that behavior cannot be implementation dependent!

> I feel like this is a little ambiguous, are writers required to use the default value with optional columns or can different writers behave differently?

I think that optional vs required should be a separate concept and not related to default values. Optional allows null values in the column, required means that null cannot be in the column. Then I think we have an initial default, which is the default value set when the column is added that is applied to all existing rows. Then there is also a writer default, which can change and is the value that a writer _must_ write for the column if it is not supplied by the user. This write default can change because it affects future writes and not existing data.

The place where optional and default values overlap is that we currently use `null` as the default value for optional fields. That's why I would say that optional fields have an implicit initial default of `null`. To handle this, I think we would assume that the initial default value for an existing optional field is `null` and only allow changing the write default value. When you create an optional field you can set the initial default.

Here are some more concrete evolution rules with defaults:
* When creating a required field, an initial default value must be set. Neither field values nor the default may be `null`.
* When creating an optional field, an initial default value can be set. If the initial default value is not set, it is `null`.
* The initial default value may only be set when adding a field (or through an incompatible change in the API)
* The write default value for a field starts as the initial default value, and is considered set if the field has an initial default
* The write default value may be changed
* When writing, the write default must be written into data files if a value for the column is not supplied
* When writing a field with no write default, the column must be supplied or the write must fail

Does that make it clear?

rdblue: > I was saying it would be the natural behavior even if we combine both concepts to one but assuming a rewrite takes place immediately.

This is basically what I did in the SQL snippet above, so I don't think that you can distinguish between setting the initial default and then setting the write default vs setting the initial default and then immediately changing it. But yeah, I think we're in agreement mostly.

> I expect both concepts have to be exposed somehow, at least because sometimes we say we can change it freely (for future rows) and sometimes we say we can only change it behind an allowIncompatibleChanges API (for existing rows). If we expose two ways or two concepts of default values, it might be unnecessarily complex.

In that case, I'd skip allowing users to change the initial default. If you've really just created a new column and need to update the default, you can always drop it and re-create it with the right default!

As long as we don't have that, then we can hide everything behind the API. The initial default is set when creating a field, and this also sets the write default. Setting the default value for a field only changes the write default.

rdblue: > In both cases, we do not support the INSERT INTO use case yet.

I wouldn't be surprised if we started supporting this in INSERT INTO soon. It seems like a good next step.

wmoustafa: > In that case, I'd skip allowing users to change the initial default. If you've really just created a new column and need to update the default, you can always drop it and re-create it with the right default!
> 
> As long as we don't have that, then we can hide everything behind the API. The initial default is set when creating a field, and this also sets the write default. Setting the default value for a field only changes the write default.

That is very close to option 1. The only difference is that option 1 allowed for changes behind the `allowIncompatibleChanges` API. Did you mean to remove that fallback? (maybe you meant we can keep it but the above is how the spec should read, which I agree with).

format/spec.md
rdblue: Sounds good to me! I was on the fence about this.

Do we need the `0x` prefix, or is it just nice to have that?

format/spec.md
rdblue: It isn't close to option 1 because the write default can still freely change. The only thing we need to put behind `alloweIncompatibleChanges` is modifying the initial default.

This is fine:

```java
table.updateSchema()
    .addRequiredColumn("test", IntegerType.get(), 34) // sets the initial and write defaults to 34
    .setColumnDefault("test", 35) // sets the write default to 35
    .commit();
```

This is what would expose the initial default and would be a bit confusing:

```java
table.updateSchema()
    .addRequiredColumn("test", IntegerType.get(), 34) // sets the initial and write defaults to 34
    .commit();
table.updateSchema()
    .allowIncompatibleChanges()
    .setInitialColumnDefault("test", 35) // sets the initial default to 35
    .commit();
```

I think we could get away without adding `setInitialColumnDefault` because you can always go drop the column and re-add it if it really is a safe change. This is equivalent to the above:

```java
table.updateSchema()
    .addRequiredColumn("test", IntegerType.get(), 34) // sets the initial and write defaults to 34
    .commit();
table.updateSchema()
    .dropColumn("test")
    .commit();
table.updateSchema()
    .addRequiredColumn("test", IntegerType.get(), 35) // sets the initial and write defaults to 35
    .commit();
```

wmoustafa: I see. So the concern is exposing both `setColumnDefault()` and `setInitialColumnDefault()` only because we do not issue a complete rewrite upon DDL. Otherwise, we could use just one.

wmoustafa: Ok now that I see the code snippets, the idea is much clearer. I will refer to code snippets as (1), (2), (3), respectively.

Code snippet (1) is really nice. I think it is closer to what I was expressing in option (2) in my last comment. That is:
* It introduces the column and its schema evolution default value atomically. That would be equivalent to a rewrite, except we are storing this value in a separate placeholder instead of physically storing it with each record (i.e., rewrite).
* It reserves all future calls of `setColumnDefault()` to the future records (i.e., write side default value), and it does not override the value dedicated for schema evolution. `setColumnDefault()` can be used to change the write side default value without affecting the schema evolution default value.

I also agree (2) is confusing and it is the reason I initially thought two values can lead to such APIs.

For (3), I think it is still subject to the rules in (1), in addition to the general rules around dropping and re-introducing a column. Another way to simulate (3)'s effect without dropping and recreating the column is to issue an `UPDATE` statement from `34` to `35`. I think this has the side effect of not distinguishing between organic `34` and `34` from schema evolution, but that is how an RDBMS would behave anyways. In fact, dropping and recreating the column, along with changing default values opens many spec questions which might be beyond the scope of this PR. (For example, if you drop an existing column do you expect to wipe out the all the values (including ones on rows with non-default values) and start over if you re-introduce the column? If not what if a rewrite happens in between? Which rows does the new default value override? etc.) Also, I think this would get even more complex if we attempt to apply it on inner fields in order to update the default value of an inner field. 

So overall, I think what we have in (1) looks good. To update the past rows default value, the user can move within whatever room the rest of the spec allows them to (dropping the column if they can bear the general column dropping consequences), or running an `UPDATE` if they are okay with updating rows coincidentally carrying the default value (but RDBMSs would do the same). Finally, if really needed I think it is okay to introduce (2) to avoid jumping multiple hoops, but it should not be part of the spec. That can be done in the future as well, so we do not preemptively loosen the spec before users ask for it.

rdblue: Code snippet (3) is an alternative to code snippet (2) that doesn't expose `setInitialColumnDefault` and is demonstrating the same behavior as code snippet (1). I'm not sure what you mean that this "has the side effect of not distinguishing between organic 34 and 34 from schema evolution". This is how you would change the initial default for a column: by dropping and re-creating the column with a different default. That works because the use case is that you've made a mistake and need to correct it _before writing any data_. The point here is that we don't necessarily need to add `setInitialColumnDefault`.

The behavior of `dropColumn` is the standard behavior: the column is no longer present in the table. Rewrites are not required to keep column values for dropped columns.

wmoustafa: > I'm not sure what you mean that this "has the side effect of not distinguishing between organic 34 and 34 from schema evolution"

That was a caveat if the user uses `UPDATE` statement to update he default value (which is an alternative way if dropping and recreating the column is not feasible anymore due to writing more values). So after using code snippet (1), and inserting more data, users can also change the default value by issuing an UPDATE DML, but it will not only update the rows that were present before schema evolution, but also other rows after schema evolution that happened to store the same value as the default value.

> That works because the use case is that you've made a mistake and need to correct it _before writing any data_.

Agreed. It was not initially clear if the suggestion also applied to the case when more data could have been written. I guess this is where the `UPDATE` route above can be helpful.

rdblue: Sounds like we agree! I think you're right that you can use `UPDATE` to make changes as well.

rzhang10: > Here are some more concrete evolution rules with defaults:
> 
> * When creating a required field, an initial default value must be set. Neither field values nor the default may be `null`.
> * When creating an optional field, an initial default value can be set. If the initial default value is not set, it is `null`.
> * The initial default value may only be set when adding a field (or through an incompatible change in the API)
> * The write default value for a field starts as the initial default value, and is considered set if the field has an initial default
> * The write default value may be changed
> * When writing, the write default must be written into data files if a value for the column is not supplied
> * When writing a field with no write default, the column must be supplied or the write must fail
> 
> Does that make it clear?

@rdblue 
I think I just want to make a clarification question so that I can make sure I understand the above discussions correctly.

So I assume to summarize this, you mean there should be 2 underlying `default value` concepts: one is `initial default value` and one is `write default value`, the `initial default value` affects old rows already written without the column and the `write default value` affects future rows.
And we only expose `one default value APIs` to users, but that API actually handles the underlying `two default value concepts`, it's just hidden to users.

Say I have a following scenario, there is a dataset with some rows written, call them `R`, a user later adds a new column with default value `d1`, and then changes the default value to `d2` (I assume that will be the `write default` value?) , is it true that according to this spec, when the user reads the dataset at this time, the `R` rows **must** read `d1` for that column, regardless of whether there is an actual full rewriting happening between the change from `d1` to `d2`? Because I think implementation-wise, we do want to lazily do the rewriting, not eagerly rewrite we set a default value?

Is my above understanding correct?

rdblue: Yes, the API only exposes one "default value". But we need to track two different values internally. One for existing data files (that cannot change) and one for future writes (than can be updated).

wmoustafa: > This supports lazy rewriting.

Nicely put 👍

format/spec.md
rdblue: @rzhang10, looks like you overlooked this fix.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
...N string: "boolean"`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    | `"boolean"`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
```
rdblue, line 961: Please revert the changes to this table that are not required.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
+
+| Type               | Json type         | Example                          | Note                                                                                                                                                                                                                                                                                                                      |
+|--------------------|-------------------|----------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **`boolean`**      | **`boolean`**     | `true`                           |                                                                                                                                                                                                                                                                                                                           |
```
rdblue, line 983: Can you make this consistent? This uses `boolean` and `json int`.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
+
+For default values, the serialization depends on the type of the corresponding column or nested field. The mapping of types and their corresponding default value JSON serialization is described in the following table:
+
+| Type               | Json type         | Example                          | Note                                                                                                                                                                                                                                                                                                                      |
```
rdblue, line 981: I think it would be better to match the conventions used by the table above, including the column headings and type references, like `JSON int`.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
+| **`string`**       | **`string`**      | `"foo"`                          |                                                                                                                                                                                                                                                                                                                           |
+| **`uuid`**         | **`string`**      | `"eb26bdb1-a1d8-4aa6-990e-da940875492c"` | Stores the lowercase uuid string                                                                                                                                                                                                                                                                                          |
+| **`fixed(L)`**     | **`string`**      | `"0x3162"`                       | Stored as a hexadecimal byte literal string, prefixex by `0x`                                                                                                                                                                                                                                                             |
```
rdblue, line 995: Typo: prefixex.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
+| **`binary`**       | **`string`**      | `"0x3162"`                       | Stored as a hexadecimal byte literal string, prefixex by `0x`                                                                                                                                                                                                                                                             |
+| **`struct`**       | **`object`**      | `{"a": 1, "foo": "bar"}`         | Use a JSON map to represent struct data, the keys are the nested fields' names in the struct schema, and the values are value literals of corresponding fields' type                                                                                                                                                      |
```
rdblue, line 997: This requires a fairly complicated process to rewrite the default value when the schema changes because it uses field names instead of field IDs. Have you considered other representations, like a list of fields: `[ {"field-id": 5, "value": "s"}, ... ]`?

format/spec.md
```diff
@@ -193,6 +193,13 @@ Notes:
 
 For details on how to serialize a schema to JSON, see Appendix C.
 
+#### Default value
+
+Default value can be assigned to top-level columns or nested fields when they are added to an Iceberg table as part of the schema evolution. The first time a default value is added, this default value is used to read rows belonging to the files that lack the column (i.e. the files were written before the column is added), this default value will also be the default value for the column if user later inserts new rows without specifying the column.
```
rdblue, line 198: This section needs more detail. A spec needs to state how something is tracked and requirements for working with the metadata. See the next section for a good example: "Schemas may be evolved by ... Valid type promotions are ...".

To start, this should have a short overview that describes exactly what is stored and how those are used. This is where you will introduce the `initial-default` and `write-default` properties of a struct field and explain what they are used for.

The spec also needs to clearly state all of the behavior requirements that have been discussed in this PR.

Currently, this is confusing because it refers to a single default, doesn't state how that default is stored, and allows the default to change. From a user's perspective it is okay to think of the two default values as a single default that changes, but that is too unclear for a specification. Instead, this should end with a description of how the two values are used to maintain SQL behavior.

format/spec.md
rdblue: After thinking about this for a couple days, my preference is to use the default values from the struct fields.

If an implementation doesn't update the default struct value then the most reasonable choice is to use the default that was required to create the field. That also means that we don't need to worry about traversing up the schema tree and updating default values when a nested struct changes.

My preference is to encode the default value that the user supplied and add the defaults of child fields to that value when fields are not present. For initial default values, we would use initial defaults. And for write defaults we would use write defaults.

This is too long to go in the JSON encoding section. This should be covered in the default values section above.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
+| **`float`**        | **`json float`**  | `1.1`                            |                                                                                                                                                                                                                                                                                                                           |
+| **`double`**       | **`json double`** | `1.1`                            |                                                                                                                                                                                                                                                                                                                           |
+| **`decimal(P,S)`** | **`string`**      | `"0x3162"`                       | Stores the unscaled value, as the two's-complement big-endian binary using the minimum number of bytes, converted to a hexadecimal string prefixed by `0x`                                                                                                                                                                |
```
RussellSpitzer, line 988: I still think this could be fine as just a decimal string, seems like it would be clearer for a metadata.json reader.

format/spec.md
```diff
@@ -951,24 +958,45 @@ Types are serialized according to this table:
...
+| **`date`**         | **`json int`**    | `19054`                          | Stores days from the 1970-01-01                                                                                                                                                                                                                                                                                           |
+| **`time`**         | **`json long`**   | `36000000000`                    | Stores microseconds from midnight                                                                                                                                                                                                                                                                                         |
+| **`timestamp`**    | **`json long`**   | `1646277378000000`               | Stores microseconds from 1970-01-01 00:00:00.000000                                                                                                                                                                                                                                                                       |
```
RussellSpitzer, line 991: I think we need to specify UTC here as well, we won't be storing the default with timezone information even if the timestamp type has it.

format/spec.md
rzhang10: > This requires a fairly complicated process to rewrite the default value when the schema changes because it uses field names instead of field IDs. Have you considered other representations, like a list of fields: `[ {"field-id": 5, "value": "s"}, ... ]`?

I think you are right about using field id to better track fields in terms of schema evolution. do you think array and map also need to be encoded this way (using field-id instead of name)? My feeling is they are unnecessary because array and map can't have schema evolution themselves anyway?

rzhang10: > My preference is to encode the default value that the user supplied and add the defaults of child fields to that value when fields are not present. For initial default values, we would use initial defaults. And for write defaults we would use write defaults.

Are you saying if the a default value is missing for a outer struct, its child default value should be automatically propagated to its own level ?

I feel adding this propagation logic (either traverse up the tree, or down the tree) will be too complicated. I have a different idea, how about we treat all the default value as independent?

Let's say there is a struct called `foo` and its one child called `foo.bar`.

If the user is to query and project `foo`, i.e. `select foo`, we use the default value at the `foo` level.
If the user is to query and project `foo.bar`, i.e. `select foo.bar`, we use the default value at the `foo.bar` level.
Basically, I'm saying the default value of `foo` and `foo.bar` are independent, and users also set those 2 values independently.

format/spec.md
rzhang10: Yeah, I specifically left out some details because I originally feel this is user-facing documentation. I can update it to expose the details.

format/spec.md
rzhang10: I'm fine with either hex string or decimal string.

format/spec.md
RussellSpitzer: I think you can think of this doc as if you were trying to implement Iceberg yourself. Whatever we write here should be all of the information required to write a brand new implementation.

format/spec.md
rdblue: Agreed. The decimal string would be better and easier to work with. We should specify that it _should_ have the correct number of digits after the decimal, even if that requires padding with 0s, like `34.00` for `decimal(9,2)`.

format/spec.md
rdblue: Yeah, the spec is definitely not user-facing. It is a clear definition of what metadata we are tracking and how it must be used in tables.

format/spec.md
rdblue: > @rdblue In the case of default values on files which have been name-mapped, would we still do name-mapping first and then use the field-id mapping correct? If so I think that sounds pretty reasonable although it would make things a bit more difficult to read in the JSON file.

I don't think that name mapping would have any effect here. Name mapping is a way to produce an Iceberg schema, but Iceberg schemas require field IDs. So everything should have an ID by this point.

If we were to use names, then the ID/name pairings would come from the schema itself -- hence the requirement that we update the defaults as the names change. But that seems needlessly difficult to implement so I think it makes much more sense to track default values by field ID.

rdblue: > do you think array and map also need to be encoded this way (using field-id instead of name)?

No, I think that map and array defaults should be stored as JSON lists. The element or key/value IDs are fairly easy to match up.

> Are you saying if the a default value is missing for a outer struct, its child default value should be automatically propagated to its own level ?

Yes. I think we have to do something because we need to define not only what to write, but how to read. It is likely that some implementation will fail to update a default struct when a child field is added. We can do a two things in that case: fail because we don't know how to handle the child field, or look at the child field and use its default value.

I doubt that we will be able to simply "require" keeping the default updated, and clearly refusing to read the table because of a default value issue is not a good idea. So that leaves using the child's default value.

If we choose to do this now, then we can simplify the requirements because you don't _have_ to update parent default values.

rdblue: > Basically, I'm saying the default value of foo and foo.bar are independent, and users also set and change those 2 values independently.

How is this possible? If I have `foo struct<required bap string> default struct("zap")` and I add `bar` to get `foo struct<required bap string, required bar int default 0>` then the defaults cannot be completely independent because `bar` is required. You either have to update the `foo` default to `struct("zap", 0)` when adding `bar` (which is not reliable), use the default for `bar`, or fail because `bar` is required but has not value in the struct's default value.

format/spec.md
rzhang10: Reverted.

format/spec.md
rzhang10: Updated the details.

rzhang10:
One minor issue: I noticed that although this pr is making changes to the spec.md file, github doesn't give this pr a `docs` tag, is the `format/spec.md` file tracked incorrectly?

format/spec.md
rdblue, line 161: Instead of noting "see below", please make the "default value" text a link to the default value section.

format/spec.md
```diff
@@ -193,6 +193,17 @@ Notes:
 
 For details on how to serialize a schema to JSON, see Appendix C.
 
+#### Default value
+
+Default value can be assigned to a column when the column is added to an Iceberg table as part of the schema evolution. They are tracked at the level of a nested field inside a struct, thus it can be used for both top-level columns and nested columns. Iceberg tracks two default values internally: `initial-default` and `write-default`. The `initial-default` is used to read rows belonging to files that lack the column (i.e. the files were written before the column is added); the `write-default` value will be used for the automatically populating the column if user later inserts new rows without specifying the column.
```
rdblue, line 198: Beginning when "assigned ... when ..." is not giving context about what field defaults are. This should begin with a high-level explanation and then go into detail. Also, this is the Iceberg spec so it doesn't make sense to refer to what "Iceberg tracks". Instead, this should state what is required.

> Default values can be tracked for struct fields (both nested structs and the top-level schema's struct). There are two defaults for a field:
> * `initial-default` is a value that must be projected when reading a data file that was written before the field was added to the schema
> * `write-default` is a value that must be written for all rows when the field is missing from input data while writing a new data file
>
> Note that all schema fields are required when writing data into an Iceberg table. Omitting a known field from a data file is not allowed. The write default for a field should be written when a field is not supplied to a write. If the write default is not set, the writer must fail.

wmoustafa: A couple of suggestions for the names: `file-to-record default` and `record-to-file default`, or `read-time default` and `write-time default`. I feel `file-to-record default` and `record-to-file default` are expressive, symmetric, and accurately capture the function of each default. We might explain that the former cannot be changed, while the latter can be changed throughout the lifecycle of a table. The former cannot be changed because it is used to read existing files while the latter can be changed because it is used to write new files.

We might also start by explaining the semantics of default values from a _contract_ point of view:

> A default value associated with a field is used to:
> (1) populate the field's value for all records that were written before the field is introduced
> (2) populate the field's value for any records that will be written after the field is introduced, when such records do not supply that field's value.
> All fields introduced at table creation time (i.e., not part of schema evolution) only leverage the second use case. Fields introduced as part of a schema evolution can leverage both use cases.
> 
> Changes of default values apply to future records only (i.e., records written after the default value has changed). To prevent retroactive change of record values that are already written, each field is associated with two default values ... (then we can continue with the section above).

format/spec.md
```diff
@@ -965,10 +976,31 @@ Types are serialized according to this table:
...
+| **`double`**       | **`json double`**  | `1.1`                                    |                                                                                                                                                                                                                                                                                                                           |
+| **`decimal(P,S)`** | **`string`**       | `"0x3162"`                               | Stores the unscaled value, as the two's-complement big-endian binary using the minimum number of bytes, converted to a hexadecimal string prefixed by `0x`                                                                                                                                                                |
```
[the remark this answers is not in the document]
rzhang10: Sorry I missed this one, let me modify it!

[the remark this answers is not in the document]
rzhang10: Actually it seems `json number` can also encode arbitrary length decimal numbers? do you think we should use string or number ?

format/spec.md
rzhang10: fixed.

format/spec.md
rdblue: @wmoustafa, I think that these should _not_ be symmetric because they have very different behaviors. The initial default can only be set when the column is created. The write default can change any time. Also, the initial default is explicitly not a read-time default. While it is applied at read time, I think that `file-to-record` or `read-time` imply that the default can be "used" by not writing columns into data files, which is explicitly disallowed.

I'd keep the `initial-default` and `write-default` names for clarity.

wmoustafa: @rdblue I am fine with that. Could you comment on the intro (the second paragraph in my comment)?

rzhang10: @rdblue Updated the spec according to the above discussions, could you take a look again?

sgcowell:
Will default values be part of format version 2, or will this be part of a new format version 3?   If it's format version 2, what happens when existing Iceberg versions read table metadata with default values in it?

format/spec.md
```diff
@@ -965,10 +982,31 @@ Types are serialized according to this table:
...
+| **`timestamp`**    | **`string`**       | `"2007-12-03T10:15:30"`                  | Stores ISO-8601 standard date-time without time-zone                                                                                                                                                                                                                                                                      |
+| **`timestamptz`**  | **`string`**       | `"2007-12-03T10:15:30"`                  | Stores ISO-8601 standard date-time with time-zone                                                                                                                                                                                                                                                                         |
```
rdblue, line 1002: This is an invalid `timestamptz` value. Timestamp with time zone literals are required to have a zone offset.

In addition, the examples should always include precision down to microseconds for `time`, `timestamp`, and `timestamptz`. Even if the parser is more lax and allows values with different precision, the spec should always represent values correctly.

rdblue:
@rzhang10, @wmoustafa, @RussellSpitzer: I've updated this and I think it's about ready to commit. Can you take a quick look to make sure it looks alright?

rdblue:
@sgcowell, version 2 is released and adopted, so these changes must go in v3. We will support these with a compatibility flag until then so you can opt into using them in v2 with known risks. I've documented the incompatibilities Appendix E that describes version changes.

format/spec.md
```diff
@@ -194,6 +194,19 @@ Notes:
...
+- `initial-default` is used to populate the field's value for all records that were written before the field was added to the schema
+- `write-default` is used to populate the field's value for any records written after the field was added to the schema, if the writer does not supply the field's value
+
+The `initial-default` is set only when a field is added to an existing schema. The `write-default` is initially set to the same value as `initial-default` and can be changed through schema evolution. If either default is not set for an optional field, then the default value is null for compatibility with older spec versions.
+
+Together, the `initial-default` and `write-default` produce SQL default value behavior without rewriting data files. That is, default value changes may only affect future records and all known fields are written into data files. To produce this behavior, omitting a known field when writing a data file is not allowed. The write default for a field must be written if a field is not supplied to a write. If the write default for a required field is not set, the writer must fail.
```
RussellSpitzer, line 205: I think this is a bit confusing because of the ordering of sentences. Perhaps something like

"The ANSI(?) SQL default value behavior treats a new column with a default value as if all previous rows missing that field now have the default value. The default is allowed to be changed for new writes but changing the default does not effect earlier writes. To achieve this behavior in Iceberg, omitting a known field when writing a new data file is never allowed. The write default must be used when writing any new files if a value for the default field is not provided. If the field is required and it is not supplied and there are is no default available, the write must fail."

I'm a little confused on allowing the default for required fields and then allowing writers not to supply them. Isn't this supposed to be behavior for an Optional field which has not been set? Maybe an example on the difference between an Optional field with a write default and a required field with a write default? Sorry if I missed this discussion but I'm a little confused on the difference between Optional w/Default and Required w/Default

RussellSpitzer: Based on the evolution rules below I think I understand 

```An optional field may be skipped by a writer. When an optional field is skipped the write-default should be used in place of the missing value and this default may be null. A required field may only be skipped by a writer if a write-default exists for that field and this default must not be null.```

File: format/spec.md [hunk 24]
```diff
@@ -1091,9 +1117,39 @@ This serialization scheme is for storing single values as individual binary valu
...
+| **`map`**          | **`JSON object of key and value arrays`** | `{ "keys": ["a", "b"], "values": [1, 2] }` | Stores arrays of keys and values; individual keys and values are serialized using this JSON single-value format |
+
 
 ## Appendix E: Format version changes
 
+### Version 3
+
+Default values are added to struct fields in v3.
+* The `write-default` is a forward-compatible change because it is only used at write time. Old writers will fail because the field is missing.
+* Tables with `initial-default` will be read correctly by older readers if `initial-default` is always null for optional fields. Otherwise, old readers will default optional columns with null. Old readers will fail to read required fields that have an `initial-default` because the default is not supported, when reading data files that are missing the required fields.
```
RussellSpitzer, line 1151: The last sentence I think can be reordered a bit
"Old readers will fail to read required fields which are populated by `initial-default` because that default is not supported." ?

RussellSpitzer (approved):
I'm on board with the semantics, I have a few suggestions on the wording but I think we are in agreement on the behaviors here.

format/spec.md
```diff
@@ -1128,15 +1128,15 @@ This serialization scheme is for storing single values as individual binary valu
...
-| **`string`**       | **`JSON string`**                         | `"foo"`                                    | |
-| **`uuid`**         | **`JSON string`**                         | `"eb26bdb1-a1d8-4aa6-990e-da940875492c"`   | Stores the lowercase uuid string |
-| **`fixed(L)`**     | **`JSON string`**                         | `"0x3162"`                                 | Stored as a hexadecimal string, prefixed by `0x` |
-| **`binary`**       | **`JSON string`**                         | `"0x3162"`                                 | Stored as a hexadecimal string, prefixed by `0x` |
+| **`string`**       | **`JSON string`**                         | `"iceberg"`                                | |
+| **`uuid`**         | **`JSON string`**                         | `"f79c3e09-677c-4bbd-a479-3f349cb785e7"`   | Stores the lowercase uuid string |
+| **`fixed(L)`**     | **`JSON string`**                         | `"0x00010203"`                             | Stored as a hexadecimal string, prefixed by `0x` |
```
rzhang10, line 1138: Should we also note that the hex string length should be less than or equal to 2*L

format/spec.md
```diff
@@ -227,6 +223,15 @@ Any struct, including a top-level schema, can evolve through deleting fields, ad
 
 Grouping a subset of a struct’s fields into a nested struct is **not** allowed, nor is moving fields from a nested struct into its immediate parent struct (`struct<a, b, c> ↔ struct<a, struct<b, c>>`). Evolving primitive types to structs is **not** allowed, nor is evolving a single-field struct to a primitive (`map<string, int> ↔ map<string, struct<int>>`).
 
+Struct evolution requires the following rules for default values:
+* The `initial-default` must be set when a field is added and cannot change
+* The `write-default` must be set when a field is added and may change
+* When a required field is added, both defaults must be set to a non-null value
+* When an optional field is added, the defaults may be null and should be explicitly set
+* When a new field is added to a struct with a default value, the default should be updated to include the new field's default
```
rzhang10, line 231: Wouldn't this cause a cascading effect? if the struct itself is nested inside another outer struct, should the outer struct's default regarding the inner struct also be updated, so on and so forth?

I thought we tried to use the downward traversing approach at runtime to resolve the default value consistency.

format/spec.md
rdblue: Shouldn't it be exactly `2L+2`?

format/spec.md
rdblue: This was meant as a suggestion. It would be nice if defaults were updated. Or do you think we should not update a default and always recover defaults by traversing child types?

format/spec.md
rzhang10: Oh I was talking about the part after the `0x`, I was previously thinking if less than `2L + 2` we prepend `0`s to it. But I also think making it strictly `2L+2` works too.

format/spec.md
rzhang10: I think from an implementation standpoint, propagating the default update to the whole schema tree could be a bit tricky to implement. I feel the approach of only updating the default at its own level, and delaying the recovery of defaults to access time should be fine? As long as we always traverse/honor the default value defined at deeper levels?

Or do you think updating the defaults of an outer struct should also overwrite the default of an inner child field if it has its own default defined?

format/spec.md
rdblue: I don't think there's a need to specify the length. I just wanted to clarify that it will never be `less than or equal to 2L`. It will always be `2L+2`.

format/spec.md
rdblue: > An optional field may be skipped by a writer.

A write may not specify the value for a field with a default, but the write must only produce data files that contain the column set from the default value. I wouldn't say that you can ever "skip" a field.

rdblue: I updated this based on your suggestions. Thank you!

format/spec.md
rdblue: Thanks, I updated to your wording.

rdblue (approved):

format/spec.md
rdblue: I thought about this a bit more and changed the wording to this:
> When a new field is added to a struct with a default value, updating the struct's default is optional

It is up to the writer whether to update the default. If the write default for the field changes, then not updating the value means not needing to change the struct's default as well. Not requiring this will probably give more expected behavior, but still allow people to set the default explicitly.

rdblue:
Thanks for all the reviews and work, @rzhang10, @wmoustafa, and @RussellSpitzer! I merged this.

wmoustafa:
Great work @rzhang10! Thank you @rdblue and @RussellSpitzer for the insightful discussions!

PR #4303, opened 2022-03-09
Core: Fix deleted data files validation in OverwriteFiles

aokolnychyi:
This PR fixes the deleted data files validation for overwrites by filter in `OverwriteFiles`. Right now, we allow removal of data files in partitions we overwrite, which, I think violates the assumption.

Found by @szehon-ho in #4293.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java
+++ b/core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java
@@ -35,7 +35,7 @@ public class BaseOverwriteFiles extends MergingSnapshotProducer<OverwriteFiles> 
   private Long startingSnapshotId = null;
   private Expression conflictDetectionFilter = null;
   private boolean validateNewDataFiles = false;
-  private boolean validateNewDeleteFiles = false;
+  private boolean validateNewDeletes = false;
 
   protected BaseOverwriteFiles(String tableName, TableOperations ops) {
     super(tableName, ops);
@@ -98,7 +98,7 @@ public class BaseOverwriteFiles extends MergingSnapshotProducer<OverwriteFiles> 
 
   @Override
   public OverwriteFiles validateNoConflictingDeletes() {
-    this.validateNewDeleteFiles = true;
+    this.validateNewDeletes = true;
     failMissingDeletePaths();
     return this;
   }
@@ -134,10 +134,11 @@ public class BaseOverwriteFiles extends MergingSnapshotProducer<OverwriteFiles> 
       validateAddedDataFiles(base, startingSnapshotId, dataConflictDetectionFilter());
     }
 
-    if (validateNewDeleteFiles) {
+    if (validateNewDeletes) {
       if (rowFilter() != Expressions.alwaysFalse()) {
         Expression filter = conflictDetectionFilter != null ? conflictDetectionFilter : rowFilter();
         validateNoNewDeleteFiles(base, startingSnapshotId, filter);
+        validateDeletedDataFiles(base, startingSnapshotId, filter);
       }
 
       if (deletedDataFiles.size() > 0) {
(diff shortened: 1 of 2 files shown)


aokolnychyi:
@rdblue @szehon-ho @RussellSpitzer, could you take a look?

core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java
```diff
@@ -35,7 +35,7 @@
   private Long startingSnapshotId = null;
   private Expression conflictDetectionFilter = null;
   private boolean validateNewDataFiles = false;
-  private boolean validateNewDeleteFiles = false;
+  private boolean validateNewDeletes = false;
```
RussellSpitzer, line 38: Why the rename here?

core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java
```diff
@@ -134,10 +134,11 @@ protected void validate(TableMetadata base) {
       validateAddedDataFiles(base, startingSnapshotId, dataConflictDetectionFilter());
     }
 
-    if (validateNewDeleteFiles) {
+    if (validateNewDeletes) {
       if (rowFilter() != Expressions.alwaysFalse()) {
         Expression filter = conflictDetectionFilter != null ? conflictDetectionFilter : rowFilter();
         validateNoNewDeleteFiles(base, startingSnapshotId, filter);
+        validateDeletedDataFiles(base, startingSnapshotId, filter);
```
RussellSpitzer, line 141: 👍

core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java
aokolnychyi: Cause we not only validate added delete files but also deleted data files. We have two types of deletes to cover.

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
```diff
@@ -747,6 +747,60 @@ public void testConcurrentConflictingPositionDeletesOverwriteByFilter() {
         overwrite::commit);
   }
 
+  @Test
+  public void testConcurrentConflictingDataFileDeleteOverwriteByFilter() {
```
aokolnychyi, line 751: This test would previously fail as the commit would succeed.

RussellSpitzer: The error here is because file 2 is both modified and deleted correct?

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
```diff
@@ -747,6 +747,60 @@ public void testConcurrentConflictingPositionDeletesOverwriteByFilter() {
...
+    Assert.assertNull("Should be empty table", table.currentSnapshot());
+
+    table.newAppend()
+        .appendFile(FILE_DAY_1)
+        .appendFile(FILE_DAY_2)
+        .commit();
+
+    Snapshot firstSnapshot = table.currentSnapshot();
+
+    OverwriteFiles overwrite = table.newOverwrite()
```
RussellSpitzer, line 788: While this one is alright because file 2 is modified, but file 1 is deleted.

RussellSpitzer (approved):
Looks good to me

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
aokolnychyi: Yeah, we concurrently removed data from a partition what we are trying to overwrite.

szehon-ho (approved):

szehon-ho:
+1 from me too, glad we found and fixed it!

aokolnychyi:
Thanks, @RussellSpitzer @szehon-ho! Merged to unblock Szehon's PR.

PR #4306, opened 2022-03-10
Build: Upgrade to Gradle 7.4.1

XN137:
https://docs.gradle.org/7.4.1/release-notes.html

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -83,7 +83,7 @@ done
 APP_HOME=$( cd "${APP_HOME:-./}" && pwd -P ) || exit
 
 if [ ! -e $APP_HOME/gradle/wrapper/gradle-wrapper.jar ]; then
-    curl -o $APP_HOME/gradle/wrapper/gradle-wrapper.jar https://raw.githubusercontent.com/gradle/gradle/v7.4.0/gradle/wrapper/gradle-wrapper.jar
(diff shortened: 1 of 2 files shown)


gradlew
```diff
@@ -83,7 +83,7 @@ done
 APP_HOME=$( cd "${APP_HOME:-./}" && pwd -P ) || exit
 
 if [ ! -e $APP_HOME/gradle/wrapper/gradle-wrapper.jar ]; then
-    curl -o $APP_HOME/gradle/wrapper/gradle-wrapper.jar https://raw.githubusercontent.com/gradle/gradle/v7.4.0/gradle/wrapper/gradle-wrapper.jar
+    curl -o $APP_HOME/gradle/wrapper/gradle-wrapper.jar https://raw.githubusercontent.com/gradle/gradle/v7.4.1/gradle/wrapper/gradle-wrapper.jar
```
XN137, line 86: i would like to note that this does not get the wrapper jar of 7.4.1 but just the wrapper jar the gradle project itself happened to use at the time of the tag:
https://github.com/gradle/gradle/tree/v7.4.1/gradle/wrapper

yet there doesnt seem to be any official wrapper jar download url.

snazy: Yes - the wrapper jar's only in the full gradle distribution.
The recommended way is to put the wrapper jar into the git (but we won't have that here).

rdblue:
Do we need to continuously upgrade gradle minor versions? Is there a benefit to doing this?

snazy:
It’s not a big deal to keep Gradle up to date and get bug fixes before we run into potential build issues.

PR #4307, opened 2022-03-10
Spark 3.2: Allow disabling HiddenPathFilter in RemoveOrphansFiles

### LINKED ISSUE #4249: Allow disabling `HiddenPathFilter` from `RemoveOrphanFiles`
There's a `PathFilter`, `HiddenPathFilter`, that is used during the `RemoveOrphanFiles` Spark actions (amongst other places).

Hadoop filtering will remove anything with an underscore or a dot in a path, so if a partition filter is created that has a leading underscore, then remove orphan files will fail to remove these files.

We should update `RemoveOrphanFiles` to handle cases where directories are erroneously marked as hidden paths.

ulmako:
Closes #4249


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -107,10 +110,12 @@ public void accept(String file) {
...
 
-    this.hadoopConf = new SerializableConfiguration(spark.sessionState().newHadoopConf());
-    this.partitionDiscoveryParallelism = spark.sessionState().conf().parallelPartitionDiscoveryParallelism();
+    hadoopConf = new SerializableConfiguration(spark.sessionState().newHadoopConf());
+    partitionDiscoveryParallelism = spark.sessionState().conf().parallelPartitionDiscoveryParallelism();
     this.table = table;
-    this.location = table.location();
+    includeHiddenPaths = PropertyUtil.propertyAsBoolean(table.properties(), INCLUDE_HIDDEN_PATHS,
+        INCLUDE_HIDDEN_PATHS_DEFAULT);
+    location = table.location();
```
rdblue, line 118: We use the prefix `this.` so that it is easy to see when a statement sets an instance field rather than a local variable.

Also, even if we want to make a change, we tend to prefer leaving code that doesn't need to change alone so that we avoid commit conflicts.

Can you revert all of the changes to remove `this.`?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -227,7 +235,10 @@ private static void listDirRecursively(
 
       List<String> subDirs = Lists.newArrayList();
 
-      for (FileStatus file : fs.listStatus(path, HiddenPathFilter.get())) {
+      FileStatus[] fileStatus =
+          (includeHiddenPaths) ? fs.listStatus(path) : fs.listStatus(path, HiddenPathFilter.get());
```
rdblue, line 239: Style: No need for parens around the boolean variable here.

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -167,6 +167,9 @@ private TableProperties() {
   public static final String ORC_BATCH_SIZE = "read.orc.vectorization.batch-size";
   public static final int ORC_BATCH_SIZE_DEFAULT = 5000;
 
+  public static final String INCLUDE_HIDDEN_PATHS = "read.fs.include-hidden-paths";
+  public static final boolean INCLUDE_HIDDEN_PATHS_DEFAULT = false;
```
rdblue, line 171: I don't think that this is a good candidate for a table property. Instead, I think it should be an option on the `DeleteOrphanFiles` action:

```java
SparkActions.get(spark)
    .deleteOrphanFiles(table)
    .option("ignore-hidden-paths", "false")
    .execute();
```

Table properties are used for more generally useful options. But this one is fairly specific to the action itself.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -107,10 +110,12 @@ public void accept(String file) {
...
     super(spark);
 
-    this.hadoopConf = new SerializableConfiguration(spark.sessionState().newHadoopConf());
-    this.partitionDiscoveryParallelism = spark.sessionState().conf().parallelPartitionDiscoveryParallelism();
+    hadoopConf = new SerializableConfiguration(spark.sessionState().newHadoopConf());
+    partitionDiscoveryParallelism = spark.sessionState().conf().parallelPartitionDiscoveryParallelism();
     this.table = table;
-    this.location = table.location();
+    includeHiddenPaths = PropertyUtil.propertyAsBoolean(table.properties(), INCLUDE_HIDDEN_PATHS,
+        INCLUDE_HIDDEN_PATHS_DEFAULT);
```
rdblue, line 117: I think it is confusing to include paths that are hidden, which is by default false. Instead I think it makes sense to disable ignoring hidden paths. That's why I suggested "ignore-hidden-paths" for the property above.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -263,7 +276,15 @@ private static void listDirRecursively(
...
-        listDirRecursively(dir, predicate, conf.value().value(), maxDepth, maxDirectSubDirs, subDirs, files);
+        listDirRecursively(
+            dir,
+            predicate,
+            conf.value().value(),
+            maxDepth,
+            maxDirectSubDirs,
+            subDirs,
+            files,
+            includeHiddenPaths);
```
rdblue, line 287: Can you revert the formatting change here? We just need to add `includeHiddenPaths`.

rdblue (changes requested):
Thanks for submitting this, @ulmako! It looks really close to being ready but I had a few comments to fix.

Also, do you think that this should automatically detect whether hidden paths should be ignored? We should be able to do that by checking for partition field names starting with `.` or `_` in the table's partition specs. If there is a partition field like `_name` then we know that we should not ignore hidden paths.

ulmako:
Hey @rdblue,

thanks for all your feedback.
The formatting stuff got mixed up before I set up the code-style scheme and I did not check it. I will obviously revert those changes. 

Your suggestion to use the `options` in `BaseSparkAction` is definitely the better choice.

Regarding the automatic detection, by looking into the partition field names, I have one question: would you still ignore all other hidden paths, that are not related to the partitions or disable the HiddenPathFilter completely. For illustration purposes:
Given the partition name `_part` and the following paths:
```
-- /data/
      | --  _part=AA
      | --  _part=BB
      | --  _part=CC
      | --  _some-folder
```

Would you only include the paths `_part=*` and ignore the path `_some-folder` or include all paths?

Also, where should I document the new option?

kbendick (commented):
Hi @ulmako! I left some review comments, but I forgot to click submit 🙈  So sorry about that!

I largely agree with @rdblue that this is likely not a good candidate for a table property.

I also looked into the source code for the hadoop file system list path filtering, and it looks like all of the filtering is done on the client side. Meaning that the data is still queried from HDFS and then filtered within the program as far as I can tell.

We should likely just make this an option or not for the action, but I thought I'd leave that finding as a comment in case it makes a difference to others.]

Thanks again for your work on this and sorry for my slow response! Please feel free to reach out on Slack (or here) if you need anything 😄

kbendick:
> Regarding the automatic detection, by looking into the partition field names, I have one question: would you still ignore all other hidden paths, that are not related to the partitions or disable the HiddenPathFilter completely. For illustration purposes: Given the partition name `_part` and the following paths:
> 
> ```
> -- /data/
>       | --  _part=AA
>       | --  _part=BB
>       | --  _part=CC
>       | --  _some-folder
> ```
> Would you only include the paths `_part=*` and ignore the path `_some-folder` or include all paths?

I originally had the same comment as Ryan, about checking for partitions specifically as that was the source of the original problem.

However, thinking on this question, I would think we would still want to allow for removing directories / paths that start with a leading `_` if the option were enabled, even if they aren't part of partitions. Particularly, some file output committers use `_` before moving their data over via rename operation, which might be something that needs to be cleaned up using `RemoveOrphanFiles` in case the job fails at a specific time.

Please correct me if I'm mistaken about the file output committers as I tend to personally use S3 (and then either S3FileIO or S3A) and the magic output committers if using s3a. 

I have also not historically used HDFS much in production scenarios, so possibly I'm mistaken about the benefits / utility of having hidden paths or how they'd end up in the directory structure of the table (outside of the previously two mentioned situations).

> Also, where should I document the new option?

For documenting the new option, let's worry about that once we get it confirmed into the right place. But definitely let's be sure to document it.

I would start with adding it as a JavaDoc comment, assuming that there is a JavaDoc comment for the class already that documents the options for the action (I believe there is). Once we're closer to the final solution, we can also find a place in the docs themselves to put this option if necessary.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -191,9 +196,10 @@ private String jobDesc() {
     List<String> matchingFiles = Lists.newArrayList();
 
     Predicate<FileStatus> predicate = file -> file.getModificationTime() < olderThanTimestamp;
+    PathFilter filter = pathFilter(table.spec());
```
kbendick, line 199: It's a bit confusing to me to have both `predicate` and `filter`, especially as variables passed to the same function.

I know `pathFilter` is the name of the function, but would it make sense to name this variable as `pathFilter` and the function as something like `buildPathFilter` or `inferHiddenPathFilter`?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -205,15 +211,27 @@ private String jobDesc() {
...
+    }
+
+    return (partitionNames.isEmpty()) ? HiddenPathFilter.get() : new PartitionAwareHiddenPathFilter(partitionNames);
+  }
+
   private static void listDirRecursively(
-      String dir, Predicate<FileStatus> predicate, Configuration conf, int maxDepth,
-      int maxDirectSubDirs, List<String> remainingSubDirs, List<String> matchingFiles) {
+          String dir, Predicate<FileStatus> predicate, Configuration conf, int maxDepth,
+          int maxDirectSubDirs, List<String> remainingSubDirs, PathFilter filter, List<String> matchingFiles) {
```
kbendick, line 234: Nit: This is over-indented. The original indention is correct - 2 spaces for the start of the function declaration, and then 4 spaces for the start of any code that continues into the next line (so the argument list here).

If you use IntelliJ (and some other IDEs), there is a code formatter that will handle things like indents etc: https://iceberg.apache.org/community/#setting-up-ide-and-code-style

I find personally sometimes I have to set it per-subproject (so for Flink, and for Spark, etc). But that could be because of the way I set it up.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -273,4 +293,20 @@ private static void listDirRecursively(
       return files.iterator();
     };
   }
+
+  @VisibleForTesting
+  static class PartitionAwareHiddenPathFilter implements PathFilter, Serializable {
```
kbendick, line 298: Could we add a javadoc to this? Some of the variables are named in a way that doesn't immediately make sense to me on first look, but I can't think of better alternatives so I think a JavaDoc is possibly the best idea.

The JavaDoc for `HiddenPathFilter` is a good place to look for inspiration. So, using that, maybe something like
```
/**
 * A {@link PathFilter} that filters out hidden path, but does not filter out paths that would be marked
 * as hidden by {@link HiddenPathFilter} due to a partition field that starts with one of the characters that
 * indicate a hidden path.
 */
```

That said, you can hold off on adding the JavaDoc until we are closer to merging this 🙂

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -205,15 +211,27 @@ private String jobDesc() {
...
     return spark().createDataset(completeMatchingFileRDD.rdd(), Encoders.STRING()).toDF("file_path");
   }
 
+  private PathFilter pathFilter(PartitionSpec spec) {
+    List<String> partitionNames = Lists.newArrayList();
+    for (PartitionField field : spec.fields()) {
+      if (field.name().startsWith("_") || field.name().startsWith(".")) {
+        partitionNames.add(field.name());
+      }
+    }
```
kbendick, line 227: Would it make sense to move this function into `PartitionAwareHiddenPathFilter` as a static function?

There might be serialization reasons that it can't, but I think it makes more sense for this code to live nearby that.

kbendick, line 227: Would it make sense to move this function into `PartitionAwareHiddenPathFilter` as a static function?

There might be serialization reasons that it can't, but I think it makes more sense for this code to live nearby that.

kbendick, line 227: This function is correct for a table with only one partition spec, but for tables that have evolved their partitioning over time, it won't work.

Tables now track all of their past partition specs using the `specs` field from the table. See https://github.com/apache/iceberg/blob/c1adf2034280884d05afad4e67f09f2872c47d78/api/src/main/java/org/apache/iceberg/Table.java#L77-L82

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -273,4 +284,35 @@ private static void listDirRecursively(
...
+
+    @Override
+    public boolean accept(Path path) {
+      boolean isHiddenPartitionPath = hiddenPathPartitionNames.stream().anyMatch(path.getName()::startsWith);
+      return isHiddenPartitionPath || HiddenPathFilter.get().accept(path);
+    }
+
+    static PathFilter build(Map<Integer, PartitionSpec> specs) {
+      Set<String> partitionNames = new HashSet<>();
```
kbendick, line 304: Nit: We don't use `new HashSet<>()` directly, as generally speaking that will usually lead to having to grow the underlying collection multiple times.

We use `Sets.newHashSet()`, or preferably the constructor that directly takes in its elements or the expected size if possible. Otherwise, we use `Sets.newHashSet()`.

We also shade and relocate Guava so that we don't conflict with the user's existing Guava version on their classpath. The correct import would be for `org.apache.iceberg.relocated.com.google.common.collect.Sets`.

All that said, I think that you might be able to apply a `filter` and then use `Collectors.toSet` below instead.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -273,4 +284,35 @@ private static void listDirRecursively(
...
+      this.hiddenPathPartitionNames = hiddenPathPartitionNames;
+    }
+
+    @Override
+    public boolean accept(Path path) {
+      boolean isHiddenPartitionPath = hiddenPathPartitionNames.stream().anyMatch(path.getName()::startsWith);
+      return isHiddenPartitionPath || HiddenPathFilter.get().accept(path);
+    }
+
+    static PathFilter build(Map<Integer, PartitionSpec> specs) {
```
kbendick, line 303: Nit: You might want to add a null check here and return `HiddenPathFilter` immediately.

I don't think that the `.specs()` method should return `null` ever, but given the way this code is organized, I'd say that's not a bad idea.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -551,6 +554,48 @@ public void testManyLeafPartitions() throws InterruptedException {
...
+    Thread.sleep(1000);
+
+    SparkActions actions = SparkActions.get();
+
+    DeleteOrphanFiles.Result result = actions.deleteOrphanFiles(table)
+        .olderThan(System.currentTimeMillis())
+        .execute();
+
+    Assert.assertEquals("Should delete 2 files", 2, Iterables.size(result.orphanFileLocations()));
+  }
```
kbendick, line 597: Can we add a test that uses a partition spec, writes out data, then updates the partition spec, and write data again, in order to test that we're catching all partition specs propery?

That way we guard against a regression.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -551,6 +554,48 @@ public void testManyLeafPartitions() throws InterruptedException {
     Assert.assertEquals("Rows must match", records, actualRecords);
   }
 
+  @Test
+  public void testHiddenPartitionPaths() throws InterruptedException {
+    Schema schema = new Schema(
+            optional(1, "c1", Types.IntegerType.get()),
+            optional(2, "_c2", Types.StringType.get()),
+            optional(3, "c3", Types.StringType.get())
```
kbendick, line 562: Nit: These are all over-indented. Code that is continued on the next line in the same expression should be. 4 spaces over.

There's an autoformatter that plugs into IntelliJ, and I believe eclipse and some other IDEs too:

Here's a link https://iceberg.apache.org/contribute/#setting-up-ide-and-code-style

But otherwise, these 3 lines should be 4 spaces over from the start of `Schema`. Like so
```java
    Schema schema = new Schema(
        optional(1, "c1", Types.IntegerType.get()),
        optional(2, "_c2", Types.StringType.get()),
        optional(3, "c3", Types.StringType.get()));
```

kbendick: There are a few other places that are 8 spaces over when they should be 4.

kbendick (commented):
Left some more feedback. It seems like it's getting pretty close!

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
ulmako: I already tried to add this test, but couldn't figure out how to update the partition specs. Can you point me to an example on how to do it.

ulmako: Nevermind, I think I found it.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -273,4 +285,35 @@ private static void listDirRecursively(
...
```
rdblue, line 304: We would normally name this method differently. `build` is usually associated with a builder, but this is a simple factory method. For that, I would recommend naming it `forSpecs`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -273,4 +285,35 @@ private static void listDirRecursively(
...
+      if (specs == null) {
+        return HiddenPathFilter.get();
+      }
+
+      Set<String> partitionNames = specs.values().stream()
+          .map(PartitionSpec::fields)
+          .flatMap(List::stream)
+          .filter(partitionField -> partitionField.name().startsWith("_") || partitionField.name().startsWith("."))
+          .map(PartitionField::name)
+          .collect(Collectors.toSet());
```
rdblue, line 314: Looks like this will match prefixes and doesn't care about order. So a partition spec with names `_a` and `_b` would result in accepting paths like `_ahiddenfile` and `_b/_b` (out of order and/or duplicated).

I'm not sure how much complexity is correct here, but I find this a bit too permissive. The problem is that these are the files that are going to be candidates for orphan cleanup, so allowing those cases could result in files getting deleted.

I think a good compromise is to add `=` to each partition field name to use as a prefix: `_a` becomes `_a=`. That way, we at least check that the name is exactly `_a` and the folder name looks like a partition.

Ideally, this would also be able to check for `_a` and `_b` a the correct level in the path, but I don't think that's possible without a lot of work. So let's leave it out.

rdblue (changes requested):
Thanks, @ulmako! This looks close to being ready to go to me. I think we just want to add `=` to the set of partition names and add a test for it. There's more detail in my comment.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
ulmako: This is a good point. I thought about considering the partition field hierarchy for the filter in the beginning, but I was a little spooked by the complexity or better the time it would take to implement it. The `=` is a good compromise. Maybe I can find some time in the future to take another look at it and make a second PR.

ulmako:
Thanks @rdblue, @kbendick for your feedback. I really appreciate it. There's a lot to learn from you guys.
I added your suggestions to the code. Sorry it took some days, but I was pretty occupied with other tasks. Employers tend to get mad, when you ignore the task they give to you ;)

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -551,6 +554,139 @@ public void testManyLeafPartitions() throws InterruptedException {
...
+    df.select("c1", "_c2", "c3")
+        .write()
+        .format("iceberg")
+        .mode("append")
+        .save(tableLocation);
+
+    df.write().mode("append").parquet(tableLocation + "/data/_c2_trunc=AA/c3=AAAA");
+    df.write().mode("append").parquet(tableLocation + "/data/_c2_trunc=AA/c3=AAAA");
+
+    Thread.sleep(1000);
```
rdblue, line 588: Is this needed? We try to avoid sleeping in tests. Does the test fail without this?

If you need to wait for some period of time, then I suggest using `waitUntilAfter(long timestampMs)`:

```java
  protected long waitUntilAfter(long timestampMillis) {
    long current = System.currentTimeMillis();
    while (current <= timestampMillis) {
      current = System.currentTimeMillis();
    }
    return current;
  }
```

rdblue (changes requested):
This looks good to me other than the use of `Thread.sleep` in tests. Could you fix that and remove the `Draft` marker, @ulmako? Thank you!

ulmako:
Actually the sleep was not needed.

There are other test in the same class also using Thread.sleep(). Should I create another PR after this one is merged were I replace the usages with the `waitUntilAfter` method? 

Also do I need to make a PR for documentation?

rdblue:
Yeah, if you don't mind it would be great to get rid of those Thread.sleep calls! And it would still be a good idea to add `waitUntilAfter` here. Since this is using `System.currentTimeMillis` in the next line, I suspect that the tests may be flaky without waiting until after the current millisecond before the expiration call.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -585,7 +585,7 @@ public void testHiddenPartitionPaths() throws InterruptedException {
     df.write().mode("append").parquet(tableLocation + "/data/_c2_trunc=AA/c3=AAAA");
     df.write().mode("append").parquet(tableLocation + "/data/_c2_trunc=AA/c3=AAAA");
 
-    // Thread.sleep(1000);
+    waitUntilAfter(System.currentTimeMillis() + 1000);
```
rdblue, line 588: It should be sufficient to just wait until after `System.currentTimeMillis()`, without the extra delay.

rdblue:
@ulmako, sorry to ask for another change, but we actually don't need the `+1000` when using `waitUntilAfter`. That method should allow us to proceed as quickly as possible to the rest of the test. Could you remove those and then I'll test/merge?

Thanks!

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -205,15 +213,16 @@ private String jobDesc() {
     JavaRDD<String> subDirRDD = sparkContext().parallelize(subDirs, parallelism);
 
     Broadcast<SerializableConfiguration> conf = sparkContext().broadcast(hadoopConf);
-    JavaRDD<String> matchingLeafFileRDD = subDirRDD.mapPartitions(listDirsRecursively(conf, olderThanTimestamp));
+    JavaRDD<String> matchingLeafFileRDD =
```
aokolnychyi, line 216: nit: an alternative way of writing this (with 4 spaces for continued indentation)

```
JavaRDD<String> matchingLeafFileRDD = subDirRDD.mapPartitions(
    listDirsRecursively(conf, olderThanTimestamp, pathFilter)
);
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -242,7 +251,8 @@ private static void listDirRecursively(
       }
 
       for (String subDir : subDirs) {
-        listDirRecursively(subDir, predicate, conf, maxDepth - 1, maxDirectSubDirs, remainingSubDirs, matchingFiles);
+        listDirRecursively(subDir, predicate, conf, maxDepth - 1, maxDirectSubDirs, remainingSubDirs, pathFilter,
```
aokolnychyi, line 254: nit: would it be more readable to put all args on a new line? (with 4 spaces for continued indentation) 

```
listDirRecursively(
    subDir, predicate, conf, maxDepth - 1, maxDirectSubDirs, remainingSubDirs, pathFilter, matchingFiles);
```

aokolnychyi (approved):
Looks reasonable to me (apart from other comments). I had two optional minor nits on styling.

ulmako:
@rdblue, sorry it took me so long to remove the `+1000`. I somehow missed your last comment.

rdblue (approved):

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -551,6 +554,139 @@ public void testManyLeafPartitions() throws InterruptedException {
...
+    List<Row> records = Lists.newArrayList(
+            RowFactory.create(1, "AAAAAAAAAA", "AAAA")
+    );
+    Dataset<Row> df = spark.createDataFrame(records, structType).coalesce(1);
+
+    df.select("c1", "_c2", "c3")
+            .write()
+            .format("iceberg")
+            .mode("append")
+            .save(tableLocation);
```
kbendick, line 671: Nit: These are overindented.

I'm not sure how the style checks passed though. But statements that continue onto the next line should be indented by 4 spaces (from the beginning of the starting line).

So here it would be:

```java
    df.select("c1", "_c2", "c3")
        .write()
        .format("iceberg")
        .mode("append")
        .save(tableLocation);
```

If you run `./gradlew -DsparkVersions=3.2 clean check`, the process should complain about that.

kbendick (approved):
Outside of some style issues with indentation in the tests, this looks good to me.

Thank you @ulmako for working on this!

kbendick:
Oh I see. The checkstyle is failing. If you click on ther failing check, `Java CI / build-checks`, you can see this error.

```
Error: eckstyle] [ERROR] /home/runner/work/iceberg/iceberg/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java:276: Line is longer than 120 characters (found 121). [LineLength]
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
kbendick: Looks like the `=` was added so this can be resolved.

rdblue:
I merged this in #4655. Thanks for fixing this, @ulmako!

ulmako:
> I merged this in #4655. Thanks for fixing this, @ulmako!

Closed because it was merged with other PR

rdblue:
Sorry, I thought I'd closed this one. Thanks for fixing this!

PR #4308, opened 2022-03-10
Core: add parsers for events and unbounded expressions

kunal0829:
In this PR, the Iceberg events are now serializable into JSON by introducing an `EventParser` along with the addition of `ExpressionParser`.


jackye1995:
Can we limit this PR to only event parsers? The purpose is to breakdown the draft, not republish it

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+              "Serialization of Predicate type BoundSetPredicate is not currently supported.");
+    } else if (BOUNDED_UNARY_PREDICATE.equals(predicateType)) {
+      throw new UnsupportedOperationException(
+              "Serialization of Predicate type BoundUnaryPredicate is not currently supported.");
+    } else {
+      throw new IllegalArgumentException("Invalid Predicate Type");
+    }
+  }
+
+  @SuppressWarnings("CyclomaticComplexity")
```
jackye1995, line 287: this is not needed?

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+    } else if (FALSE.equals(expressionType)) {
+      return False.INSTANCE;
+    } else if (PREDICATE_TYPES.contains(expressionType)) {
+      return fromJsonToPredicate(json, expressionType);
+    } else {
+      throw new IllegalArgumentException("Invalid Operation Type");
+    }
+  }
+
+  public static Predicate<?, ?> fromJsonToPredicate(JsonNode json, String predicateType) {
```
jackye1995, line 270: please keep all methods private except for the main one for Expression

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+    if (UNBOUNDED_PREDICATE.equals(predicateType)) {
+      return fromJsonUnboundPredicate(json);
+    } else if (BOUNDED_LITERAL_PREDICATE.equals(predicateType)) {
```
jackye1995, line 273: since we are only targeting unbounded, we can just use a generic else block saying we don't support bounded expression yet

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+    } else if (expression instanceof Predicate) {
+      toJson((Predicate<?, ?>) expression, generator);
+    } else {
+      throw new IllegalArgumentException("Invalid Operation Type");
+    }
+  }
+
+  public static void toJson(And expression, JsonGenerator generator) throws IOException {
+    generator.writeStartObject();
+    generator.writeStringField(TYPE, AND);
```
jackye1995, line 115: here can we reuse the word "operation"? Because phrases like "and" and "or" are operations

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+    }
+  }
+
+  public static void toJson(UnboundPredicate<?> predicate, JsonGenerator generator) throws IOException {
+    generator.writeStartObject();
+    generator.writeStringField(TYPE, UNBOUNDED_PREDICATE);
+    generator.writeStringField(OPERATION, predicate.op().name().toLowerCase());
+    generator.writeFieldName(TERM);
+    toJson(predicate.term(), generator);
+    if (!(ONE_INPUTS.contains(predicate.op()))) {
```
jackye1995, line 167: redundant brackets after !

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  public static void toJson(NamedReference<?> term, JsonGenerator generator) throws IOException {
+    generator.writeStartObject();
+    generator.writeStringField(TYPE, NAMED_REFERENCE);
+    generator.writeStringField(VALUE, term.name());
+    generator.writeEndObject();
+  }
+
+  public static void toJson(List<Literal<?>> literals, JsonGenerator generator) throws IOException {
```
jackye1995, line 194: this method seems unnecessary, can be combined to parent method, also can use enhanced for loop

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+public class ExpressionParser {
+  private static final String TYPE = "type";
+  private static final String VALUE = "value";
+  private static final String OPERATION = "operation";
+  private static final String LITERALS = "literals";
+  private static final String TERM = "term";
+  private static final String LEFT_OPERAND = "left-operand";
+  private static final String RIGHT_OPERAND = "right-operand";
+  private static final String OPERAND = "operand";
+  private static final String AND = "and";
```
jackye1995, line 53: these can just use operation.name().toLowerCase(), no need to define extra variables

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+      return new And(fromJson(json.get(LEFT_OPERAND)), fromJson(json.get(RIGHT_OPERAND)));
+    } else if (OR.equals(expressionType)) {
+      return new Or(fromJson(json.get(LEFT_OPERAND)), fromJson(json.get(RIGHT_OPERAND)));
+    } else if (NOT.equals(expressionType)) {
+      return new Not(fromJson(json.get(OPERAND)));
+    } else if (TRUE.equals(expressionType)) {
+      return True.INSTANCE;
+    } else if (FALSE.equals(expressionType)) {
+      return False.INSTANCE;
+    } else if (PREDICATE_TYPES.contains(expressionType)) {
```
jackye1995, line 263: I think we don't need a set of predicate types, we just need to have a single else block, delegating all the rest of the conditions to `fromJsonToPredicate(json, expressionType);`

core/src/main/java/org/apache/iceberg/util/EventParser.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.iceberg.events.CreateSnapshotEvent;
+import org.apache.iceberg.events.IncrementalScanEvent;
+import org.apache.iceberg.events.ScanEvent;
+import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.ExpressionParser;
+
+public class EventParser {
+  private static final String EVENT_TYPE = "event-type";
+
```
jackye1995, line 40: nit: extra newline

core/src/main/java/org/apache/iceberg/util/EventParser.java
```diff
@@ -0,0 +1,180 @@
...
+    } else if (eventType.equals(CreateSnapshotEvent.class.getName())) {
+      return fromJsonToCreateSnapshotEvent(json);
+    } else if (eventType.equals(IncrementalScanEvent.class.getName())) {
+      return fromJsonToIncrementalScanEvent(json);
+    } else {
+      throw new IllegalArgumentException("Invalid Event Type");
+    }
+  }
+
+  public static ScanEvent fromJsonToScanEvent(JsonNode json) {
```
jackye1995, line 155: keep methods below private

core/src/test/java/org/apache/iceberg/expressions/TestExpressionParser.java
```diff
@@ -0,0 +1,340 @@
...
+            new NamedReference("Column3-Name"));
+
+    And andExpression = new And(inPredicate, eqPredicate);
+    Or orNestedExpression = new Or(andExpression, isNanPredicate);
+    String actual = ExpressionParser.toJson(orNestedExpression, true);
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void testParserBothWays() {
```
jackye1995, line 232: nit: "roundTrip" instead of "bothWays"

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+  private static final String TERM = "term";
+  private static final String LEFT_OPERAND = "left-operand";
+  private static final String RIGHT_OPERAND = "right-operand";
+  private static final String OPERAND = "operand";
+  private static final String AND = "and";
+  private static final String OR = "or";
+  private static final String NOT = "not";
+  private static final String TRUE = "true";
+  private static final String FALSE = "false";
+  private static final String UNBOUNDED_PREDICATE = "unbounded-predicate";
```
rdblue, line 58: The term is "unbound", meaning that the expression has not been bound to a particular incoming struct/row type, not "unbounded".

core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java
```diff
@@ -0,0 +1,334 @@
...
+      toJson(expression, generator);
+      generator.flush();
+      return writer.toString();
+
+    } catch (IOException e) {
+      throw new UncheckedIOException("Failed to write json", e);
+    }
+  }
+
+  public static void toJson(Expression expression, JsonGenerator generator) throws IOException {
```
rdblue, line 95: Iceberg separates the recursive logic to process a tree structure from the logic to do something by using visitors. I think that this should use an expression visitor to produce a JSON representation.

core/src/main/java/org/apache/iceberg/util/EventParser.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  public static String toJson(Object event, boolean pretty) {
+    try {
+      StringWriter writer = new StringWriter();
+      JsonGenerator generator = JsonUtil.factory().createGenerator(writer);
+      if (pretty) {
+        generator.useDefaultPrettyPrinter();
+      }
+      if (event instanceof ScanEvent) {
```
rdblue, line 65: We add newlines after control flow blocks and the following statement.

rdblue:
@kunal0829, can you write up a doc that summarizes the proposed serialization formats?

kunal0829:
@rdblue Here is an example write up that explains the proposed format: https://docs.google.com/document/d/1CvOfsVAMezqQyyAwIn0qfew9EvlX-5PIvKGHPzlHldA/edit?usp=sharing

Please let me know if you have any questions about the proposed format

kunal0829:
@rdblue  Here is an example write up that explains the proposed format: https://docs.google.com/document/d/1CvOfsVAMezqQyyAwIn0qfew9EvlX-5PIvKGHPzlHldA/edit?usp=sharing

Please let me know if you have any questions about the proposed format

rdblue:
I've not been able to comment on the doc since it isn't opened up for discussion. I'll add a few of the comments here instead.

My main concern about the proposed JSON format is that it is needlessly verbose and relies heavily on the JVM classes rather than representing the expressions more directly. For example, it is scoped to only unbound expressions, but still uses "unbound-predicate" and "named-reference". It also uses "left-operand" and "right-operand". I think it would be better to use fewer words and make better use of "type".

Each node in the expression tree should be represented as a JSON object. Those are distinguished by the `type` field: that can have the following values: `and`, `or`, `not`, `eq`, `not-eq`, `in`, `not-in`, `lt`, `lt-eq`, `gt`, `gt-eq`, `starts-with`, `not-starts-with`, `is-null`, `not-null`, `is-nan`, `not-nan`, `reference`, `transform`, and `literal`.

Then there is some specialization by `type`:

* `and` and `or` must have child expressions named  `left` and `right`
* `not` must have a child expression named `child`
* All predicates must have a `term` expression that is either a `transform` or a `reference`
* `transform` must have a `term` expression that is either a `transform` or a `reference` and a `transform` field that is a transform string ("identity", "bucket[16]", etc.)
* Set predicates (`in` and `not-in`) must have a JSON array called `values` of literals
* Comparison predicates (`eq`, etc.) must have a `value` that is a literal
* `reference` must have a `name` that is a JSON string
* `literal` must have a `value` field
* Literal values must use the JSON value serialization defined in #4301 (structs, lists, and maps are not allowed)

I would also allow the reference to be replaced with a string field name, and a literal to be replaced with the value when nested.

I think this format is more direct:

```
-- example 1: event_ts = 2022-04-24T11:42:12.578391 OR bucket(16, id) IN (0, 1)
{
  "type": "or",
  "left": {
    "type": "eq",
    "term": "event_ts", -- equivalent to { "type": "reference", "name": "event_ts" }
    "value": "2022-04-24T11:42:12.578391" -- equivalent to { "type": "literal", "value": "2022-04-24T11:42:12.578391" }
  }
  "right": {
    "type": "in"
    "term": {
      "type": "transform",
      "transform": "bucket[16]",
      "term": "id"
    },
    "values": [
      { "type": "literal", "value": 0 },
      1
    ]
  }
}
-- example 2: true
{ "type": "literal", "value": true }
```

rdblue:
@kunal0829 and @jackye1995, an expression JSON parser was just merged in #5602.

PR #4309, opened 2022-03-10
Core : Fix NPE post HadoopMetricContext Deserialization

### LINKED ISSUE #4288: NullPointerException in spark rewrite_manifests procedure
Using snapshot version `0.13.0-20220307.001124-2`. Spark version 3.2 

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 4.0 failed 4 times, most recent failure: Lost task 0.3 in stage 4.0 (TID 308) (10.103.53.2 executor 1): java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:221)
	at org.apache.iceberg.hadoop.HadoopMetricsContext.counter(HadoopMetricsContext.java:80)
	at org.apache.iceberg.aws.s3.S3OutputStream.<init>(S3OutputStream.java:135)
	at org.apache.iceberg.aws.s3.S3OutputFile.createOrOverwrite(S3OutputFile.java:60)
	at org.apache.iceberg.avro.AvroFileAppender.<init>(AvroFileAppender.java:51)
	at org.apache.iceberg.avro.Avro$WriteBuilder.build(Avro.java:191)
	at org.apache.iceberg.ManifestWriter$V1Writer.newAppender(ManifestWriter.java:281)
	at org.apache.iceberg.ManifestWriter.<init>(ManifestWriter.java:58)
	at org.apache.iceberg.ManifestWriter.<init>(ManifestWriter.java:34)
	at org.apache.iceberg.ManifestWriter$V1Writer.<init>(ManifestWriter.java:260)
	at org.apache.iceberg.ManifestFiles.write(ManifestFiles.java:117)
	at org.apache.iceberg.spark.actions.BaseRewriteManifestsSparkAction.writeManifest(BaseRewriteManifestsSparkAction.java:324)
	at org.apache.iceberg.spark.actions.BaseRewriteManifestsSparkAction.lambda$toManifests$afb7bc39$1(BaseRewriteManifestsSparkAction.java:354)
	at org.apache.spark.sql.Dataset.$anonfun$mapPartitions$1(Dataset.scala:2826)
	at org.apache.spark.sql.execution.MapPartitionsExec.$anonfun$doExecute$3(objects.scala:201)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(RDD.scala:898)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(RDD.scala:898)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:373)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:337)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:506)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1462)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:509)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)

```
I think this may have been introduced in https://github.com/apache/iceberg/pull/4050 From a quick glance at the code it seems that the root cause is that `HadoopMetricsContext#statistics` is a transient variable. If I understood thingas correctly, `HadoopMetricsContext` is initialized in `S3FileIo`:

```java
  @Override
    this.awsProperties = new AwsProperties(properties);

    // Do not override s3 client if it was provided
    if (s3 == null) {
      this.s3 = AwsClientFactories.from(properties)::s3;
    }

    // Report Hadoop metrics if Hadoop is available
    try {
      this.metrics = ctor.newInstance("s3");

    } catch (NoSuchMethodException | ClassCastException e) {
    }
  }
``` 
Since `FileIO` is a broadcast variable in `org.apache.iceberg.spark.actions.BaseRewriteManifestsSparkAction` 

```java
  private static ManifestFile writeManifest(
      List<Row> rows, int startIndex, int endIndex, Broadcast<FileIO> io,
      String location, int format, PartitionSpec spec, StructType sparkType) throws IOException {
...
```
It seems that `HadoopMetricsContext` is serialized along with it and when its deserialized at the executor, `statistcs` variable will be null because it was marked as transient. I'm not really sure if that's exactly whats happening but the only way I could reproduce this outside of spark was calling `HadoopMetricsContext#counter` with statistics variable set to null and the most obvious way that this could happen is after a deserialization

singhpk234:
- Currently when we use any serializer other than JavaSerializer (for ex : KryoSerializer), we face an NPE due to statistics being null and not be re-initialized on de-serialization (since we broadcast FileIO and hence part of this we serialize it at driver & de serialize it in executor for spark) .

This PR attempts to address the issue reported below : 

Closes #4288

### DIFF
--- a/core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
+++ b/core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
@@ -27,6 +27,7 @@ import java.util.function.Consumer;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.FileIOMetricsContext;
+import org.apache.iceberg.util.SerializableSupplier;
 
 /**
  * FileIO Metrics implementation that delegates to Hadoop FileSystem
@@ -36,6 +37,7 @@ public class HadoopMetricsContext implements FileIOMetricsContext {
   public static final String SCHEME = "io.metrics-scheme";
 
   private String scheme;
+  private SerializableSupplier<FileSystem.Statistics> statisticsSupplier;
   private transient FileSystem.Statistics statistics;
 
   public HadoopMetricsContext(String scheme) {
@@ -43,6 +45,7 @@ public class HadoopMetricsContext implements FileIOMetricsContext {
         "Scheme is required for Hadoop FileSystem metrics reporting");
 
     this.scheme = scheme;
+    this.statisticsSupplier = () -> FileSystem.getStatistics(scheme, null);
   }
 
   @Override
@@ -50,7 +53,8 @@ public class HadoopMetricsContext implements FileIOMetricsContext {
     // FileIO has no specific implementation class, but Hadoop will
     // still track and report for the provided scheme.
     this.scheme = properties.getOrDefault(SCHEME, scheme);
-    this.statistics = FileSystem.getStatistics(scheme, null);
+    this.statisticsSupplier = () -> FileSystem.getStatistics(scheme, null);
+    this.statistics = statisticsSupplier.get();
   }
 
   /**
@@ -71,16 +75,16 @@ public class HadoopMetricsContext implements FileIOMetricsContext {
     switch (name) {
       case READ_BYTES:
         ValidationException.check(type == Long.class, "'%s' requires Long type", READ_BYTES);
-        return (Counter<T>) longCounter(statistics::incrementBytesRead);
+        return (Counter<T>) longCounter(statistics()::incrementBytesRead);
       case READ_OPERATIONS:
         ValidationException.check(type == Integer.class, "'%s' requires Integer type", READ_OPERATIONS);
-        return (Counter<T>) integerCounter(statistics::incrementReadOps);
+        return (Counter<T>) integerCounter(statistics()::incrementReadOps);
       case WRITE_BYTES:
         ValidationException.check(type == Long.class, "'%s' requires Long type", WRITE_BYTES);
-        return (Counter<T>) longCounter(statistics::incrementBytesWritten);
+        return (Counter<T>) longCounter(statistics()::incrementBytesWritten);
       case WRITE_OPERATIONS:
         ValidationException.check(type == Integer.class, "'%s' requires Integer type", WRITE_OPERATIONS);
-        return (Counter<T>) integerCounter(statistics::incrementWriteOps);
+        return (Counter<T>) integerCounter(statistics()::incrementWriteOps);
       default:
         throw new IllegalArgumentException(String.format("Unsupported counter: '%s'", name));
     }
@@ -120,6 +124,14 @@ public class HadoopMetricsContext implements FileIOMetricsContext {
 
   private void readObject(ObjectInputStream in) throws ClassNotFoundException, IOException {
     in.defaultReadObject();
-    statistics = FileSystem.getStatistics(scheme, null);
+    statistics = statisticsSupplier.get();
+  }
+
+  private FileSystem.Statistics statistics() {
+    if (statistics == null) {
+      statistics = statisticsSupplier.get();
+    }
+
+    return statistics;
   }
 }
--- a/core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
+++ b/core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
@@ -19,15 +19,11 @@
 
 package org.apache.iceberg.hadoop;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
 import java.util.Map;
 import java.util.function.Consumer;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.FileIOMetricsContext;
-import org.apache.iceberg.util.SerializableSupplier;
 
 /**
  * FileIO Metrics implementation that delegates to Hadoop FileSystem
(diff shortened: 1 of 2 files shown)


szehon-ho (commented):
I think its reasonable, but we may need to make this method/singleton MT-safe?  This class might be called by different threads if I understand correctly (eg, if multiple file output streams, etc)

kbendick:
I agree with @szehon-ho that it would probably be best to guard this in terms of multithreaded calling etc.  Especially as this is in `core`, so it could be running in so many various frameworks and contexts.

singhpk234:
Thanks @szehon-ho @kbendick , added the handling for the same.

kbendick:
> Thanks @szehon-ho @kbendick , added the handling for the same.

Thanks @singhpk234. The other thing I would say this would benefit from is a test.

Can we possibly add one using Spark for example to start? We don’t need to do all file formats etc, but kryo and java serde would be good to use such that we broadcast the table and it undergoes ser/de.

WDYT?

singhpk234:
+1 @kbendick, I totally agree on adding test cases for the same. 

IIUC are you talking of creating an E2E test which actually does broadcasts and hence becomes a part of serDe as within spark itself we have such test's present with both serDe, but it doesn't broadcast rather does serilaization and deserilization on it's own and checks if it's reliable. 

I am referring to this : 

https://github.com/apache/iceberg/blob/1b774c2c3f054309024e0b0e62688ffa9e25fe20/spark/v3.2/spark/src/test/java/org/apache/iceberg/TestTableSerialization.java#L71-L101

P.S have added a test which would have failed without this change. please let me know your thoughts on this approach

core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
```diff
@@ -114,12 +111,18 @@ public void increment(Integer amount) {
     };
   }
 
-  private void writeObject(ObjectOutputStream out) throws IOException {
-    out.defaultWriteObject();
-  }
+  private FileSystem.Statistics statistics() {
+    FileSystem.Statistics localStatisticsRef = statistics;
```
jackye1995, line 115: why do we need a `localStatisticsRef`?

If we already have `private transient FileSystem.Statistics statistics`, I think we can just check against that and see:

```java
    if (statistics == null) {
      synchronized (this) {
        if (statistics == null) {
          this.statistics = FileSystem.getStatistics(scheme, null);
        }
      }
    }

return statistics;
```

I think this is also used in places like https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SerializableTable.java#L117-L132

singhpk234: agree with you in this context it seems to be un-necessary, have made the changes for the same.

generally we use this format in singleton pattern where there is a volatile obj holding the reference of the singleton. (https://www.javacodemonk.com/threadsafe-singleton-design-pattern-java-806ad7e6).


  | Using localRef, we are reducing the access of volatile  variable to just one for positive usecase. If we do not use localRef,  then we would have to access volatile variable twice - once for checking  null and then at method return time. Accessing volatile memory is quite  an expensive affair because it involves reaching out to main memory.
-- | --

jackye1995: >  then we would have to access volatile variable twice - once for checking null and then at method return time

I agree, but the second access is only done for one time for initialization, so I think we can take the cost in this case

core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
```diff
@@ -114,12 +111,15 @@ public void increment(Integer amount) {
     };
   }
 
-  private void writeObject(ObjectOutputStream out) throws IOException {
-    out.defaultWriteObject();
-  }
+  private FileSystem.Statistics statistics() {
+    if (statistics == null) {
```
szehon-ho, line 115: I think this field needs to be volatile by double-check locking pattern, because otherwise it may return partially initialized statistics object in the not null case?

singhpk234: Thanks, missed that corrected it.

core/src/main/java/org/apache/iceberg/hadoop/HadoopMetricsContext.java
singhpk234: came across this blog : https://www.journaldev.com/171/thread-safety-in-java-singleton-classes
This states that having this kind of localRef's can increase perf of this method as much as 25% 

"The local variable result seems unnecessary. But, it’s there to improve the performance of our code. In cases where the instance is already initialized (most of the time), the volatile field is only accessed once (due to “return result;” instead of “return instance;”). This can improve the method’s overall performance by as much as 25 percent."

I think we will access it always 2 times first to check if it;s initialized and the last time to actually return it.

rdblue (approved):

rdblue:
Thanks, @singhpk234! Looks good. And thanks to @kbendick, @szehon-ho, and @jackye1995 for reviewing.

PR #2925, opened 2021-08-03
Core: Support serializable and snapshot isolation for ReplacePartitions

szehon-ho:
* From https://mail-archives.apache.org/mod_mbox/iceberg-dev/202107.mbox/%3CCAMwmD1_c4u%2BwDSOLtScCyid%2BfHicBnKdmyTuKqFTG-uaE9eq6A%40mail.gmail.com%3E and offline discussion with Ryan
* This change will make "insert overwrite" with mode=dynamic support serializable isolation (a ValidationException will be thrown if a conflicting change is made) in the case where it is not selecting from the same table.
* This will catch cases where two insert overwrite happen at the same time, both can potentially succeed and both datas land.
* "insert overwrite" remains snapshot isolation in the case where the select is from the same table. For that to support serializable, we need to plug the snapshotId from initial Spark compile, which could be looked at on spark-extension side.

PR #4310, opened 2022-03-10
Spark 3.2: Cleanup DataFrame overwritePartitions isolation level handling

szehon-ho:
This makes some minor cleanup and add one more test for #2925 

This is from comment https://github.com/apache/iceberg/pull/4293#discussion_r823035860 of splitting these changes from that pr into a separate one.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
@@ -277,17 +277,15 @@ abstract class SparkWrite implements Write, RequiresDistributionAndOrdering {
       IsolationLevel isolationLevel = writeConf.isolationLevel();
       Long validateFromSnapshotId = writeConf.validateFromSnapshotId();
 
+      if (isolationLevel != null && validateFromSnapshotId != null) {
+        dynamicOverwrite.validateFromSnapshot(validateFromSnapshotId);
+      }
+
       if (isolationLevel == SERIALIZABLE) {
-        if (validateFromSnapshotId != null) {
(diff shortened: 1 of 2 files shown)


szehon-ho:
@aokolnychyi fyi

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -277,17 +277,15 @@ public void commit(WriterCommitMessage[] messages) {
...
 
+      if (isolationLevel != null && validateFromSnapshotId != null) {
+        dynamicOverwrite.validateFromSnapshot(validateFromSnapshotId);
+      }
+
       if (isolationLevel == SERIALIZABLE) {
-        if (validateFromSnapshotId != null) {
-          dynamicOverwrite.validateFromSnapshot(validateFromSnapshotId);
-        }
         dynamicOverwrite.validateNoConflictingData();
```
kbendick, line 285: Nit: It seems like this one is repeated in both cases. Since there's no existing `else` clause, it's probably not worth reducing the duplication as we might need more `if` statements but thought I'd call it out.

kbendick (approved):
This looks good to me. Thanks for following up @szehon-ho!

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
szehon-ho: Thanks yea good catch, I was thinking to leave it here to make it more clear in each validation mode what we are validating, and maybe there are future modes where we do not do this as you said.  As opposed to the validationSnapshotId which is the basis of all validations.

szehon-ho:
Thanks @kbendick for the review!

PR #4311, opened 2022-03-11
Core: Fix delete file index with non equality column filter lost equlity delete files

xloya:
When we through Flink to write enabled the `upsert` option, and then we specify non equality columns conditions in the query, some of the equality delete files will be lost due to the wrong usage of the filter.

### DIFF
--- a/core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
+++ b/core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
@@ -128,19 +128,19 @@ public class TestOverwriteWithValidation extends TableTestBase {
       .build();
 
   private static final DataFile FILE_DAY_2_ANOTHER_RANGE = DataFiles
-          .builder(PARTITION_SPEC)
-          .withPath("/path/to/data-3.parquet")
-          .withFileSizeInBytes(0)
-          .withPartitionPath("date=2018-06-09")
-          .withMetrics(new Metrics(5L,
-                  null, // no column sizes
-                  ImmutableMap.of(1, 5L, 2, 3L), // value count
-                  ImmutableMap.of(1, 0L, 2, 2L), // null count
-                  null,
-                  ImmutableMap.of(1, longToBuffer(10L)), // lower bounds
-                  ImmutableMap.of(1, longToBuffer(14L)) // upper bounds
-          ))
-          .build();
+      .builder(PARTITION_SPEC)
+      .withPath("/path/to/data-3.parquet")
+      .withFileSizeInBytes(0)
+      .withPartitionPath("date=2018-06-09")
+      .withMetrics(new Metrics(5L,
+          null, // no column sizes
+          ImmutableMap.of(1, 5L, 2, 3L), // value count
+          ImmutableMap.of(1, 0L, 2, 2L), // null count
+          null,
+          ImmutableMap.of(1, longToBuffer(10L)), // lower bounds
+          ImmutableMap.of(1, longToBuffer(14L)) // upper bounds
+      ))
+      .build();
 
   private static final DeleteFile FILE_DAY_2_ANOTHER_RANGE_EQ_DELETES = FileMetadata.deleteFileBuilder(PARTITION_SPEC)
       .ofEqualityDeletes()
--- a/core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
+++ b/core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
@@ -96,6 +96,23 @@ public class TestOverwriteWithValidation extends TableTestBase {
       ))
       .build();
 
+
+  private static final DataFile FILE_DAY_2_ANOTHER_RANGE = DataFiles
+          .builder(PARTITION_SPEC)
+          .withPath("/path/to/data-3.parquet")
+          .withFileSizeInBytes(0)
+          .withPartitionPath("date=2018-06-09")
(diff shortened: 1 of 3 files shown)


kbendick (commented):
Thanks for providing a patch @xloya.

It would help if. you could provide more context. Is this something that needs to be special cased for flink `upsert` mode in some way? Modifying `core` and removing tests causes me concern at first sight.

It's possible I missed something on the mailing list or some other discussion, forgive me if so.  Is it possible there's an example scenario that can be given to help me understand?

xloya:
> Thanks for providing a patch @xloya.
> 
> It would help if. you could provide more context. Is this something that needs to be special cased for flink `upsert` mode in some way? Modifying `core` and removing tests causes me concern at first sight.
> 
> It's possible I missed something on the mailing list or some other discussion, forgive me if so. Is it possible there's an example scenario that can be given to help me understand?

Of course, we have a scenario to write data to iceberg's v2 table through Flink CDC. They have non-primary key query scenarios. The current implementation in core will add filters, which may lose the latest seq num equality delete files for Flink streaming writing.  
E.g:  
When seq num=1, Flink writes a record with `id=1, date='2021-01-01'`, will insert a data record with `id=1, date='2021-01-01'`, and a equality delete data record with an `id=1, date='2021-01-01'`;  
When seq num=2, writes a record with `id=1, date='2022-01-01'` to update, will insert a data record with `id=1, date='2022-01-01'`, and a equality delete record with `id=1 ,date='2022-01-01' `;  
At this time, when using `select * from xxx where date < '2022-01-01'` to query, due to the addition of the filter, the equality delete file written when seq num=2 will be filtered out.  

This is currently the easiest way to fix the problem. If we want to optimize for Flink upsert, then I think may need to read the latest record with the primary key that already exists in the table and write it to the equality delete file when writing, while instead of writing the inserted data to the equality delete file

kbendick (commented):
Thanks for adding the test and explaining the situation more @xloya.

However, I don't think this is necessarily the appropriate solution. There's several places that call to the delete file index.

Additionally, this removes the ability to have non-conflicting delete transactions go through.

I think we should reconsider how the filter is built in the case of Flink `upsert` mode, and likely check for any other places where this issue exists.

I can confirm though that the test provided will fail if `filterRows` is added back in.

cc @RussellSpitzer @flyrain @aokolnychyi who have all worked on the delete file index or helped review it.

szehon-ho:
I agree with @kbendick , changing core will break some other engine like Spark that depend on this for things like validation check for serializable isolation mode.  Also agree with his suggestion on Flink side using the filter, and if we can add some other option there to deleteFileIndex for its use case.

hililiwei:
We seem to have the same issues.
And it only happens on our Parquet table (doesn't happen on our Avro table). After analysis, we found that the problem occurred in the metric （such as upper_bounds\lower_bounds）filtering process of the MANIFEST file (avro tables did not generate these metric data).
Our solution is different from this PR. We try to  trim the row filter fileds of the metrics. For deleted manifest file, only the metric filed id in the equality_ids will be filtered, please refer to https://github.com/apache/iceberg/pull/4316 for details.
I'm not sure which way is better, or that there is another better solution.
I'm sorry if your PR is to address a different issues.
Thx. 😄

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSink.java
```diff
@@ -253,6 +253,34 @@ public void testInsertIntoPartition() throws Exception {
...
+            TableProperties.FORMAT_VERSION, "2",
+            TableProperties.UPSERT_ENABLED, "true"
+    );
+
+    sql("CREATE TABLE %s(id INT NOT NULL, dt DATE, PRIMARY KEY (id) NOT ENFORCED) PARTITIONED BY (id) WITH %s",
+            tableName, toWithClause(tableProps));
+
+    // insert data set
+    sql("INSERT INTO %s VALUES " +
+                    "(1, to_date('2021-01-01','yyyy-MM-dd'))", tableName);
```
rdblue, line 272: Can you fix formatting? Continuation indents are 4 spaces.

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
```diff
@@ -893,37 +893,6 @@ public void testConcurrentConflictingEqualityDeletes() {
         overwrite::commit);
   }
 
-  @Test
-  public void testConcurrentNonConflictingEqualityDeletes() {
```
rdblue, line 897: What is invalid about this test? It is suspicious that it is removed.

xloya:
> We seem to have the same issues.
> 
> And it only happens on our Parquet table (doesn't happen on our Avro table). After analysis, we found that the problem occurred in the metric （such as `upper_bounds` \ `lower_bounds` ）filtering process of the MANIFEST file (avro tables did not generate these metric data).
> 
> Our solution is different from this PR. We try to trim the row filter fileds used for metric filtering. For deleted manifest file, only the metric filed id in the `equality_ids` will be filtered, please refer to #4316 for details.
> 
> I'm not sure which way is better, or that there is another better solution.
> 
> I'm sorry if your PR is to address a different issues.
> 
> Thx. smile

Thank you for your attention to the same issue! I will take a look at the patch you committed later. In fact, I think my patch is only a temporary solution. I just want to get more opinions from out community to find the most appropriate way to fix the problem

Issue #3228, opened 2021-10-06
[Python] support partition spec in iceberg python library


jun-he:

samredai:
Now that schema has been merged I think we need a couple more PRs in order to close this:
1. PartitionField
2. PartitionSpec
3. PartitionSpecBuilder

For 3. I know we discussed whether we should go with the builder pattern or just lean into python's ability for named optional arguments. There might be a middle ground in there where we use the builder pattern (to maintain consistency with the java codebase) and simply use that builder in the `PartitionSpec` dunder init.

dramaticlly:
I can help by starting with PartitionSpec!

rdblue:
This was added in the PRs above, so I'll close this.

PR #4318, opened 2022-03-13
Python: Add Schema class

samredai:
This adds the Schema class by carving it out of PR #3228 by @nssalian and building upon it.

A `Schema` object is created by providing a list of `NestedField` instances, a schema ID, and optionally a map of aliases to field IDs. The `get_field_id(...)` method allows you to retrieve a field ID by providing the field name or an alias, and the `get_field(...)` method allows you to retrieve the field object by providing the field ID. There's also a `get_type` method that lets you retrieve the type of field by providing the field ID.

```py
from iceberg.table.schema import Schema
from iceberg.types import BooleanType, IntegerType, NestedField, StringType

fields = [
    NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False),
    NestedField(field_id=2, name="bar", field_type=IntegerType(), is_optional=True),
    NestedField(field_id=3, name="baz", field_type=BooleanType(), is_optional=False),
]
table_schema = Schema(fields=fields, schema_id=1, aliases={"qux": 3})
print(table_schema)
```
output
```
1: name=foo, type=string, required=True
2: name=bar, type=int, required=False
3: name=baz, type=boolean, required=True
```
```py
table_schema.find_field_id_by_name("foo")  # 1
table_schema.find_field_by_id(1)  # NestedField(field_id=1, name='foo', field_type=StringType(), is_optional=False)
table_schema.find_field_type(1). # StringType()
```


samredai:
@cabhishek

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,179 @@
...
+            int: The field ID for the given field identifier
+        """
+        if field_identifier_type == FIELD_IDENTIFIER_TYPES.NAME:
+            field_id = self._get_field_identifier_by_name(field_identifier=field_identifier, case_sensitive=case_sensitive)
+        elif field_identifier_type == FIELD_IDENTIFIER_TYPES.ALIAS:
+            field_id = self._get_field_identifier_by_alias(field_identifier=field_identifier, case_sensitive=case_sensitive)
+
+        return field_id
+
+    def get_field(self, field_id: int) -> NestedField:
```
rdblue, line 147: These should use maps that are constructed by indexing with a SchemaVisitor. All of the methods on schema are intended to return a field by full name or by ID from anywhere in the schema. That's why schema is not just a regular struct. Structs will return fields by name or ID, but are limited to just that struct.

That's also why we use names like `find_field` rather than `get_field` (in addition, we avoid `get` in methods names generally).

samredai: Thanks for the explanation, I saw that in the legacy implementation but now I understand the purpose. I'll update this today.

samredai: @rdblue I updated the PR to use the visitor pattern and added the `IndexById` and `IndexByName` schema visitors.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+
+        Raise:
+            ValueError: If the field ID does not exist
+        """
+        for field in self.fields:
+            if field.field_id == field_id:
+                return field
+        raise ValueError(f"Cannot get field, ID does not exist: {field_id}")
+
+    def find_field_type(self, field_id: int) -> IcebergType:
```
cabhishek, line 175: should this be re-named to `def find_field_type_by_id(...)` to be more explicit? 
This seems a little ambiguous 
```
table_schema.find_field_type(1). # StringType()
```

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+
+if sys.version_info >= (3, 8):  # pragma: no cover
+    from typing import Protocol
+else:  # pragma: no cover
+    from typing_extensions import Protocol  # type: ignore
+
+if sys.version_info >= (3, 8):  # pragma: no cover
+    from functools import singledispatchmethod
+else:  # pragma: no cover
+    from singledispatch import singledispatchmethod  # type: ignore
```
kbendick, line 29: Would it make sense to combine these two import sections into one, as they share the same condition?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+
+from iceberg.types import IcebergType, ListType, MapType, NestedField, StructType
+
+
+class SchemaVisitor(Protocol):
+    def visit(self, node) -> None:  # pragma: no cover
```
kbendick, line 35: Question: is this returning `None` and then using the `result` field instead because the returned dictionaries are somewhat different?

In either case, would it make sense to add a doc string to this Protocol, indicating that the `result` is used? Not sure if a good way to phrase it but having `None` as the return type makes me feel like it’s going to mutate `node` or some internal state (which I don’t believe they do).

samredai: That's right with this current design the visitor doesn't manipulate `node` but it accumulates state as it visits different parts of the schema. I'm wondering if it would be too strict to require concrete visitors to store the state in an attribute specifically called `result` but if that's ok then I can actually make this an ABC class which would enforce it. (A child class would still be allowed to have arguments in addition to `node` if needed):
```py

class SchemaVisitor(ABC):

    @abstractmethod
    def visit(self, node) -> None:
        ...

    @property
    @abstractmethod
    def result(self):
        ...
```
Let me know what you think.

python/src/iceberg/table/schema.py
samredai: Sounds good, updated 👍

python/src/iceberg/table/schema.py
samredai: Definitely, not sure how I missed that, thanks! (Updated the PR)

samredai:
Rebased to add the changes from the docstest PR that was merged in.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+    from functools import singledispatchmethod
+    from typing import Protocol
+else:  # pragma: no cover
+    from typing_extensions import Protocol  # type: ignore
+    from singledispatch import singledispatchmethod  # type: ignore
+
+from iceberg.types import IcebergType, ListType, MapType, NestedField, StructType
+
+
+class SchemaVisitor(Protocol):
```
rdblue, line 31: I think this should be as close to Java as possible so that visitor implementations can be translated directly. A lot of thinking and bug fixing has gone into those and it's a big risk if the translation over to python isn't straightforward. Rewriting logic in the visitors to handle a different structure could easily cause translation errors and at a minimum will probably require more effort than we want to review.

While I like how simple this is, we're probably going to run into problems. The Java visitor has before/after callbacks that aren't supported with this structure using `@singledispatch`. This structure also requires each visitor to recursively call `visit`, which mixes the logic to traverse a schema with the logic to produce some result. Last, using a `result` produced by the visitor instead of returning some value is an unnecessary change for translation.

I think we should go with the structure that I originally posted on the [partition spec PR](https://github.com/apache/iceberg/pull/3407/commits/e09d2616b8dae311a25fb13ec7bcebff96ada548).

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+        for field in self.fields:
+            schema_str += f"{field.field_id}: name={field.name}, type={field.type}, required={field.is_required}\n"
+        return schema_str.rstrip()
+
+    def __repr__(self):
+        return f"Schema(fields={repr(self.fields)}, schema_id={self.schema_id})"
+
+    @property
+    def fields(self) -> Tuple:
+        return self._struct.fields
```
rdblue, line 81: Why does this deviate from the Java API? Renaming `columns` to `fields` doesn't seem like a necessary change to me. I would opt to use mostly the same names for methods where possible so that it is easy to go between the two APIs.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+
+    @property
+    def schema_id(self) -> int:
+        return self._schema_id
+
+    @property
+    def struct(self) -> StructType:
```
rdblue, line 88: `as_struct`?

Also, should this be a property or a method?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+        self._index_by_name = index_by_name_visitor.result
+
+    def __len__(self):
+        return len(self.struct.fields)
+
+    def __str__(self):
+        schema_str = ""
+        for field in self.fields:
+            schema_str += f"{field.field_id}: name={field.name}, type={field.type}, required={field.is_required}\n"
+        return schema_str.rstrip()
```
rdblue, line 74: I think it is generally best practice to use join rather than strip to produce a string like this.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+        index_by_name_visitor.visit(self._struct)
+        self._index_by_name = index_by_name_visitor.result
+
+    def __len__(self):
+        return len(self.struct.fields)
+
+    def __str__(self):
+        schema_str = ""
+        for field in self.fields:
+            schema_str += f"{field.field_id}: name={field.name}, type={field.type}, required={field.is_required}\n"
```
rdblue, line 73: Should we match the same format the other implementation uses? Why change it?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+
+    @property
+    def schema_id(self) -> int:
+        return self._schema_id
+
+    @property
+    def struct(self) -> StructType:
+        return self._struct
+
+    def find_field_id_by_name(self, field_name: str, case_sensitive: bool = True) -> int:
```
rdblue, line 91: There isn't an equivalent method on the Java Schema class.

I think this should implement largely the same API:
* `def columns() -> List[NestedField]`
* `def find_field(self, name_or_id: str | int, case_sensitive: bool = True) -> NestedField`
* `def find_type(self, name_or_id: str | int, case_sensitive: bool = True) -> IcebergType`
* `def find_column_name(self, id: int) -> str`
* `def select(self, names: List[str], case_sensitive: bool = True) -> Schema`
* `def accessor_for_field(self, id: int) -> Accessor` -- probably not in this initial PR!

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,346 @@
...
+            NestedField(field_id=3, name="baz", field_type=BooleanType(), is_optional=False), \
+        ]
+        >>> table_schema = Schema(fields=fields, schema_id=1, aliases={"qux": 3})
+        >>> print(table_schema)
+        1: name=foo, type=string, required=True
+        2: name=bar, type=int, required=False
+        3: name=baz, type=boolean, required=True
+    """
+
+    def __init__(self, fields: Iterable[NestedField], schema_id: int, aliases: dict = {}):
```
rdblue, line 54: Instead of adding aliases right now, I recommend adding `identifier_field_ids`. Iceberg tables now track the fields that identify rows. Since this is a core part of the schema, I'd add it now. (This was probably missing from the python legacy codebase)

The aliases map is primarily for reading Parquet files because we need to keep track of the name that was used in a file for projection. I would implement that later when we get to reading Parquet.

samredai:
@rdblue thanks for the review! I brought over the the schema and visitors from the partition spec PR which also addresses the other comments.

samredai:
The latest [click](https://pypi.org/project/click/#history) release has broken black. I'm sure it'll be resolved soon enough so we can ignore these test failures for now.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,215 @@
...
+    def as_struct(self):
+        return self._struct
+
+    def _find_field_by_name(self, index: dict, name: str) -> NestedField:
+        matched_fields = [field for field_id, field in index.items() if field.name == name]
+        if not matched_fields:
+            raise ValueError("Cannot find field: {name_or_id}")
+        return matched_fields[0]
+
+    def find_field(self, name_or_id: Union[str, int], case_sensitive: bool = True) -> NestedField:
```
rdblue, line 58: Do you know when we can use [PEP 604](https://peps.python.org/pep-0604/)?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,215 @@
...
+        index = index_by_id(self)
+        if isinstance(name_or_id, int):
+            return index[name_or_id]
+        return self._find_field_by_name(index, name_or_id)
```
rdblue, line 62: This should use a name to ID map and then look up fields in the ID to field map, like Java does.

The name to ID map is important because building that requires that names are unique. It also provides better access than checking each field name. The names that are valid for `find_field` are actually full names, like `a.b.c`. That is a flattened namespace so `a.b.c` could be `["a.b", "c"]` or `["a", "b", "c"]` and we know that if indexing succeeded that there is only one column.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def find_type(self, name_or_id: Union[str, int], case_sensitive: bool = True) -> IcebergType:
+        index = index_by_id(self)
+        if isinstance(name_or_id, int):
+            return index[name_or_id].type
+        return self._find_field_by_name(index, name_or_id).type
+
+    def find_column_name(self, id: int) -> str:
+        index = index_by_id(self)
+        return index[id].name
```
rdblue, line 72: This should also be the full field name, not `NestedField.name`.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def select(self, names: List[str], case_sensitive: bool = True) -> "Schema":
+        return (
+            self._case_sensitive_select(schema=self, names=names)
+            if case_sensitive
+            else self._case_insensitive_select(schema=self, names=names)
+        )
```
rdblue, line 79: I would probably opt to avoid the if expression for cases like this. This is now 5 lines of code and probably harder to read than an if/else:

```python
    def select(self, names: List[str], case_sensitive: bool = True) -> "Schema":
        if case_sensitive:
            return self._case_sensitive_select(schema=self, names=names)
        else:
            return self._case_insensitive_select(schema=self, names=names)
```

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @classmethod
+    def _case_sensitive_select(cls, schema: "Schema", names: List[str]):
+        return cls(*[column for column in schema.columns if column.name.lower() in names])
```
rdblue, line 83: `select` needs to be implemented using a schema visitor, like we do in Java: https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/types/TypeUtil.java#L104-L116

We don't need `PruneColumns` in this PR, but we shouldn't implement this using only top-level names. It's okay to throw `NotImplementedError`.

python/tests/table/test_schema.py
```diff
@@ -0,0 +1,246 @@
...
+def test_schema_str():
+    """Test casting a schema to a string"""
+    columns = [
+        NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False),
+        NestedField(field_id=2, name="bar", field_type=IntegerType(), is_optional=True),
+        NestedField(field_id=3, name="baz", field_type=BooleanType(), is_optional=False),
+    ]
+    table_schema = schema.Schema(*columns)
+    assert """table { 
+ 1: foo: required string
```
rdblue, line 41: Shouldn't these be indented?

python/src/iceberg/table/schema.py
samredai: It's been back-ported the earliest version we support but not enabled by default. We'd have to add `from __future__ import annotations` to the top of every file where we use it. I do like the newer syntax, let me know if the visual cost of the import feels worth it and I'll go ahead and update these.

samredai:
Added an `IndexByName` visitor and did a comparison with the java client which produces the same result.

```java
import org.apache.iceberg.Schema;
import org.apache.iceberg.types.TypeUtil;

Schema schema = new Schema(
      Types.NestedField.required(1, "foo", Types.StringType.get()),
      Types.NestedField.optional(2, "bar", Types.IntegerType.get()),
      Types.NestedField.required(3, "baz", Types.BooleanType.get()),
      Types.NestedField.required(4, "qux", Types.ListType.ofOptional(5, Types.StringType.get())),
      Types.NestedField.required(6, "quux", Types.MapType.ofOptional(7, 8, Types.StringType.get(), Types.MapType.ofOptional(9, 10, Types.StringType.get(), Types.IntegerType.get())))
);
Map<String, Integer> index = TypeUtil.indexByName(schema.asStruct());
System.out.println(index);
```
output:
```
{foo=1, bar=2, baz=3, qux=4, qux.element=5, quux=6, quux.key=7, quux.value=8, quux.value.key=9, quux.value.value=10}
```


```py
from iceberg.types import (
    BooleanType,
    IntegerType,
    ListType,
    MapType,
    NestedField,
    StringType,
    StructType,
)
from iceberg.table.schema import Schema, index_by_id

schema = Schema(        NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False),
        NestedField(field_id=2, name="bar", field_type=IntegerType(), is_optional=True),
        NestedField(field_id=3, name="baz", field_type=BooleanType(), is_optional=False),
        NestedField(field_id=4, name="qux", field_type=ListType(element_id=5, element_type=StringType(), element_is_optional=True), is_optional=True),
        NestedField(field_id=6, name="quux", field_type=MapType(key_id=7, key_type=StringType(), value_id=8, value_type=MapType(key_id=9, key_type=StringType(), value_id=10, value_type=IntegerType(), value_is_optional=True), value_is_optional=True), is_optional=True))
    
index = index_by_name(table_schema)
print(index)
```
output:
```
{'foo': 1, 'bar': 2, 'baz': 3, 'qux': 4, 'qux.element': 5, 'quux': 6, 'quux.key': 7, 'quux.value': 8, 'quux.value.key': 9, 'quux.value.value': 10}
```

samredai:
Also, here's a comparison to `TypeUtil.indexById` in Java:
### Java
```java
import org.apache.iceberg.Schema;
import org.apache.iceberg.types.TypeUtil;

Schema schema = new Schema(
      Types.NestedField.required(1, "foo", Types.StringType.get()),
      Types.NestedField.optional(2, "bar", Types.IntegerType.get()),
      Types.NestedField.required(3, "baz", Types.BooleanType.get()),
      Types.NestedField.required(4, "qux", Types.ListType.ofOptional(5, Types.StringType.get())),
      Types.NestedField.required(6, "quux", Types.MapType.ofOptional(7, 8, Types.StringType.get(), Types.MapType.ofOptional(9, 10, Types.StringType.get(), Types.IntegerType.get())))
);
Map<Integer, Types.NestedField> index = TypeUtil.indexById(schema.asStruct());
System.out.println(index);
```
output:
```
{1=1: foo: required string, 2=2: bar: optional int, 3=3: baz: required boolean, 4=4: qux: required list<string>, 5=5: element: optional string, 6=6: quux: required map<string, map<string, int>>, 7=7: key: required string, 8=8: value: optional map<string, int>, 9=9: key: required string, 10=10: value: optional int}
```

### Python
```py
from iceberg.types import (
    BooleanType,
    IntegerType,
    ListType,
    MapType,
    NestedField,
    StringType,
    StructType,
)
from iceberg.table.schema import Schema, index_by_id

schema = Schema(
    NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False),
    NestedField(field_id=2, name="bar", field_type=IntegerType(), is_optional=True),
    NestedField(field_id=3, name="baz", field_type=BooleanType(), is_optional=False),
    NestedField(field_id=4, name="qux", field_type=ListType(element_id=5, element_type=StringType(), element_is_optional=True), is_optional=True),
    NestedField(field_id=6, name="quux", field_type=MapType(key_id=7, key_type=StringType(), value_id=8, value_type=MapType(key_id=9, key_type=StringType(), value_id=10, value_type=IntegerType(), value_is_optional=True), value_is_optional=True), is_optional=True)
)
    
index = index_by_id(schema)
print(index)
```
output:
```
{
 1: NestedField(field_id=1, name='foo', field_type=StringType(), is_optional=False),
 2: NestedField(field_id=2, name='bar', field_type=IntegerType(), is_optional=True),
 3: NestedField(field_id=3, name='baz', field_type=BooleanType(), is_optional=False),
 4: NestedField(field_id=4, name='qux', field_type=ListType(element_id=5, element_type=StringType(), element_is_optional=True), is_optional=True),
 5: NestedField(field_id=5, name='element', field_type=StringType(), is_optional=True),
 6: NestedField(field_id=6, name='quux', field_type=MapType(key_id=7, key_type=StringType(), value_id=8, value_type=MapType(key_id=9, key_type=StringType(), value_id=10, value_type=IntegerType(), value_is_optional=True), value_is_optional=True), is_optional=True),
 7: NestedField(field_id=7, name='key', field_type=StringType(), is_optional=False),
 8: NestedField(field_id=8, name='value', field_type=MapType(key_id=9, key_type=StringType(), value_id=10, value_type=IntegerType(), value_is_optional=True), is_optional=True),
 9: NestedField(field_id=9, name='key', field_type=StringType(), is_optional=False),
 10: NestedField(field_id=10, name='value', field_type=IntegerType(), is_optional=True),
}
```

python/src/iceberg/table/schema.py
samredai: Good point, it's actually another line shaved off if we just use an implied else. Updated and looks much cleaner 👍

python/src/iceberg/table/schema.py
samredai: Updated the `IndexById` visitor and compared it to the java client, see [here](https://github.com/apache/iceberg/pull/4318#issuecomment-1082621634).

python/src/iceberg/table/schema.py
samredai: Added an `IndexByName` visitor, see the comparison to java [here](https://github.com/apache/iceberg/pull/4318#issuecomment-1082617851).

python/src/iceberg/table/schema.py
samredai: Switched these to raise not implemented and left a TODO comment to add a PruneColumn schema visitor.

python/src/iceberg/table/schema.py
samredai: Removed aliases and added `identifier_field_ids` which I'm assuming is a `List[int]`, right?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+    @property
+    def id(self) -> int:
+        return self._schema_id
+
+    @property
+    def identifier_field_ids(self) -> List[int]:
+        return self._identifier_field_ids
+
+    @property
+    def id_index(self) -> Dict[int, NestedField]:
```
rdblue, line 64: Do we need to expose the indexes? Why not only support the `find` methods?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class Schema:
+    """A table Schema"""
+
+    def __init__(self, *columns: Iterable[NestedField], schema_id: int, identifier_field_ids: List[int] = []):
+        self._struct = StructType(*columns)  # type: ignore
+        self._schema_id = schema_id
+        self._identifier_field_ids = identifier_field_ids
+        self._name_index: Dict[str, int] = index_by_name(self)
```
rdblue, line 42: I think it is more clear to name indexes like `x_to_y` so that it is easier to read. To me, `name_index` sounds like I should be able to use it to get names, not that the index is by name. `name_to_id` makes more sense to me.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+        self._id_index: Dict[int, NestedField] = {}  # Will be lazily set when self.id_index property method is called
+
+    def __str__(self):
+        return "table { \n" + "\n".join([" " + str(field) for field in self.columns]) + "\n }"
```
rdblue, line 46: I think Java uses 2 spaces to indent.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def __repr__(self):
+        return f"Schema(fields={repr(self.columns)})"
```
rdblue, line 49: Should this also contain the schema ID and identifier field IDs?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+
+    @property
+    def columns(self) -> Iterable[NestedField]:
+        return self._struct.fields
+
+    @property
+    def id(self) -> int:
```
rdblue, line 56: I think this should be `schema_id` to be clear.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+    @property
+    def name_index(self) -> Dict[str, int]:
+        return self._name_index
+
+    def as_struct(self) -> StructType:
+        return self._struct
+
+    def find_field(self, name_or_id: str | int, case_sensitive: bool = True) -> NestedField:
+        if isinstance(name_or_id, int):
+            return self.id_index[name_or_id]
```
rdblue, line 78: I think these should call `get` to avoid throwing `KeyError`. If the field is not found, this should return `None`.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+        if case_sensitive:
+            field_id = self.name_index[name_or_id]
+        else:
+            name_index_lower = {name.lower(): field_id for name, field_id in self.name_index.items()}
```
rdblue, line 82: I don't think this class should ever iterate through a dictionary, except for just one time to produce the lower case name to id index. I think this should create the index and store it as an instance variable.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+        return self.id_index[field_id]
+
+    def find_type(self, name_or_id: str | int, case_sensitive: bool = True) -> IcebergType:
+        if isinstance(name_or_id, int):
+            return self.id_index[name_or_id].type
+        if case_sensitive:
+            field_id = self.name_index[name_or_id]
+        else:
+            name_index_lower = {name.lower(): field_id for name, field_id in self.name_index.items()}
+            field_id = name_index_lower[name_or_id.lower()]
```
rdblue, line 93: The name to id lookup appears to be duplicated. Can you write a method to do this instead?

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+        return self.id_index[field_id].type
+
+    def find_column_name(self, column_id: int) -> str:
+        matched_column = [name for name, field_id in self.name_index.items() if field_id == column_id]
```
rdblue, line 97: This should use an index, not iterate over something.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+        if not matched_column:
+            raise ValueError(f"Cannot find column name: {column_id}")
```
rdblue, line 99: This should return `None` if there is column with the given ID.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+    @abstractmethod
+    def map(self, map_type: MapType, key_result: T, value_result: T) -> T:
+        ...
+
+    @abstractmethod
+    def primitive(self, primitive: PrimitiveType) -> T:
+        ...
+
+
+def visit(obj, visitor: SchemaVisitor[T]) -> T:
```
rdblue, line 168: Minor: I think we could refactor the visit method to use `@singledispatch`:

```python
@singledispatch
def visit(obj, visitor: SchemaVisitor[T]) -> T:
    raise NotImplementedError("Cannot visit unsupported type: %s" % obj)

@visit.register(Schema)
def visit(schema: Schema, visitor: SchemaVisitor[T]) -> T:
    return visitor.schema(schema, visit(schema.as_struct(), visitor))

...
```

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+
+    elif isinstance(obj, PrimitiveType):
+        return visitor.primitive(obj)
+
+    else:
+        raise NotImplementedError("Cannot visit non-type: %s" % obj)
+
+
+def index_by_id(schema_or_type) -> Dict[int, NestedField]:
+    class IndexById(SchemaVisitor[Dict[int, NestedField]]):
```
rdblue, line 217: Does nesting this class redefine it each time? We could always move it out. I don't see a big problem with exposing it under `schema`.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+            return self._index
+
+        def primitive(self, primitive):
+            return self._index
+
+    return visit(schema_or_type, IndexById())
+
+
+def index_by_name(schema_or_type) -> Dict[str, int]:
+    class IndexByName(SchemaVisitor[Dict[str, int]]):
```
rdblue, line 247: It looks like this doesn't implement short names like the Java class does. Is it possible to add short names? It's nice to be able to refer to `locations.latitude` rather than `locations.element.latitude`.

python/src/iceberg/table/schema.py
```diff
@@ -0,0 +1,286 @@
...
+    MapType,
+    NestedField,
+    PrimitiveType,
+    StructType,
+)
+
+T = TypeVar("T")
+
+
+class Schema:
```
rdblue, line 35: I just noticed that this is under `table`. Is that needed? Why not just put this in the `iceberg` module directly?

python/src/iceberg/table/schema.py
samredai: Updated to use `self._name_to_id`, `self._name_to_id_lower`, and `self._id_to_field`

python/src/iceberg/table/schema.py
samredai: Ah right, fixed!

python/src/iceberg/table/schema.py
samredai: Updated

python/src/iceberg/table/schema.py
samredai: Updated to use `get` and return None if the field isn't found.

python/src/iceberg/table/schema.py
samredai: Great idea, updated

python/src/iceberg/table/schema.py
samredai: Updated to use the `self._lazy_id_to_field()` index.

python/src/iceberg/table/schema.py
samredai: Fixed

python/src/iceberg/table/schema.py
samredai: Great idea, done! Looks much better

python/src/iceberg/table/schema.py
samredai: Moved this out to the module level and added a `_` prefix to mark it as internal to the module.

python/src/iceberg/table/schema.py
samredai: Added the logic for short-names. Thanks for the help with this!

python/src/iceberg/table/schema.py
samredai: Moved this out to `iceberg/schema.py`

python/src/iceberg/table/schema.py
samredai: Switched this to internal methods `_lazy_id_to_field()` and `_lazy_name_to_id_lower()` which set `self._id_to_field` and `self._name_to_id_lower` the first time they're called, respectively. Subsequent calls will use the cached value. Updated the find methods to use these lazy methods.

python/src/iceberg/schema.py
```diff
@@ -0,0 +1,448 @@
...
+        """
+        if not self._name_to_id_lower:
+            self._name_to_id_lower = {name.lower(): field_id for name, field_id in self._name_to_id.items()}
+        return self._name_to_id_lower
+
+    def as_struct(self) -> StructType:
+        """Returns the underlying struct"""
+        return self._struct
+
+    def find_field(self, name_or_id: str | int, case_sensitive: bool = True) -> NestedField:
```
rdblue, line 101: I know I suggested this syntax, but I had IntelliJ complain that python 3.7 doesn't support it when I tried to use it. But tests are passing so it's okay?

python/src/iceberg/schema.py
```diff
@@ -0,0 +1,448 @@
...
+            column_id (int): The ID of the column
+
+        Raises:
+            ValueError: If no column name can be found for the given column ID
+
+        Returns:
+            str: The column name
+        """
+        column = self._lazy_id_to_field().get(column_id)
+        return None if column is None else column.name  # type: ignore
```
rdblue, line 146: This actually needs to return the full name, not the field name. In Java, this uses the same index visitor as the name to ID, but it produces the `byId` map.

rdblue (commented):
Looks great, other than the `find_column_name` method doesn't return the full column name.

rdblue:
@samredai, if you want, you can throw `NotImplementedError` in `find_column_name` and we can add it in a follow-up.

rdblue (approved):

python/src/iceberg/schema.py
samredai: It must not be recognizing the `from __future__ import annotations` at the top which is required to backport it to 3.7, but yeah if the tests are passing we're ok.

rdblue: I found this when I tried to use it in my recent changes and didn't know to import from __future__. I'll update that eventually.

PR #4241, opened 2022-02-28
Core: Add REST catalog adapter and tests

rdblue:
This adds an implementation of `RESTClient` that translates REST protocol requests to the `Catalog` API. This makes it possible to test the `RESTCatalog` implementation like any other catalog. These tests do not exercise serialization; request and response objects are passed between methods directly.

* `RESTCatalogAdaptor` is a simple translation from `RESTClient` routes to catalog handlers
* `CatalogHandlers` accepts request objects, makes the appropriate calls to a catalog, and produces response objects
* `RESTCatalog` is a `Catalog` implementation that translates to calls to `RESTClient`
* `TestRESTCatalog` runs `CatalogTests` against the `RESTCatalog`

This also implements load table and update table routes, with new tests in `CatalogTests` that are passing for REST and JDBC.

To get tests working, this PR includes a few other changes and fixes:

* Adds `equals` to `CharSequenceSet` for assertions
* Updates `TableMetadataParser` to handle `refs`
* Supports `-1` to signal "last added" in schema, partition spec, and sort order changes
* Fixes `TableMetadata` history timestamps by detecting whether a snapshot was added
* Adds `applyTo(TableMetadata.Builder)` to `MetadataUpdate` to apply changes from the REST protocol on the server side
* Adds requirements classes and validations that are used server side to check the loaded metadata

PR #4319, opened 2022-03-13
Core: Add transaction catalog tests

rdblue:
This extends the recently added `CatalogTests` class with more tests that cover transactions. There are also minor updates to other classes to make tests pass by standardizing error messages and exceptions.

This implements `equals`, `hashCode`, and `toString` in `CharSequenceSet` to make testing easier.

This is in support of #4241.


api/src/main/java/org/apache/iceberg/util/CharSequenceSet.java
```diff
@@ -152,4 +156,28 @@ public boolean removeAll(Collection<?> objects) {
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(wrapperSet);
+  }
+
+  @Override
+  public String toString() {
+    return Streams.stream(iterator()).collect(Collectors.joining("CharSequenceSet({", ", ", "})"));
```
rdblue, line 181: @szehon-ho, I addressed [your comment](https://github.com/apache/iceberg/pull/4241/files#r816330118).

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -319,6 +402,1345 @@ public void testNamespaceWithDot() {
...
+        CommitFailedException.class, expectedMessage, update::commit);
+
+    Table loaded = catalog.loadTable(TABLE);
+
+    // the spec ID may not match, so check equality of the fields
+    Assert.assertEquals("Loaded table should have expected spec", expected.fields(), loaded.spec().fields());
+  }
+
+  @Test
+  public void testUpdateTableSortOrder() {
```
nastra, line 812: `testUpdateTableSortOrder` and `testUpdateTableSortOrderServerSideRetry` for example only differ in 1-2 lines, so I woder if we can improve the tests a bit and try and reduce code duplication (in a follow up PR)

nastra (commented):
the new tests themselves LGTM, added just a few suggestions

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
```diff
@@ -115,7 +116,11 @@ protected void doRefresh() {
   public void commit(TableMetadata base, TableMetadata metadata) {
     // if the metadata is already out of date, reject it
     if (base != current()) {
-      throw new CommitFailedException("Cannot commit: stale table metadata");
+      if (base != null) {
+        throw new CommitFailedException("Cannot commit: stale table metadata");
+      } else {
+        throw new AlreadyExistsException("Table already exists: %s", tableName());
```
kbendick, line 122: This exception message reads a little funny to me, but I think in context it will make sense.

For clarification, what's happening here is that we entered some sort of create table transaction, and then at the end of the transaction, the table was created by some other process. Is that correct?

A comment might help make that more clear but I can see how the stack trace will make sense in context.

szehon-ho, line 122: I didnt look deeply, but wondering why if caller passes base=null and current is not null , we can tell the table already exists?   It seems it would be a new assumption to all the calling code?

rdblue: When `base` is null, that signals that the commit is trying to create the table and it should not exist. When `current` is not null, that signals that the table already exists. So in order to throw `AlreadyExistsException` from the create transaction, this needs to catch that case.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -319,6 +402,1345 @@ public void testNamespaceWithDot() {
...
+        TABLE_SCHEMA.asStruct(), table.schema().asStruct());
+  }
+
+  @Test
+  public void testCompleteCreateTable() {
+    C catalog = catalog();
+
+    TableIdentifier ident = TableIdentifier.of("ns", "table");
+
+    Assert.assertFalse("Table should not exist", catalog.tableExists(ident));
```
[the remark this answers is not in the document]
kbendick: I think the only thing that could be extracted out would be the create table step. However, some of these tests only create a namespace and not the table. Others create a table in a different namespace, using the added `TABLE` constant.

We could make a class level constant out of `ident`, but it might confuse things with `TABLE`.

Is it possible to use that instead? I'm assuming there's a reason for the change in namespace.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -319,6 +402,1345 @@ public void testNamespaceWithDot() {
...
+
+    // validate table settings
+    Assert.assertEquals("Table name should report its full name", catalog.name() + "." + ident, table.name());
+    Assert.assertEquals("Schema should match expected ID assignment",
+        TABLE_SCHEMA.asStruct(), table.schema().asStruct());
+    Assert.assertNotNull("Should have a location", table.location());
+    Assert.assertTrue("Should be unpartitioned", table.spec().isUnpartitioned());
+    Assert.assertTrue("Should be unsorted", table.sortOrder().isUnsorted());
+    Assert.assertNotNull("Should have table properties", table.properties());
+  }
```
kbendick, line 424: Do we need to drop the tables that are being created?

rdblue: No, the catalog is re-initialized every time and data on disk is in a temporary folder.

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
szehon-ho: yea makes sense now, thanks.  I guess its same concern as @kbendick :)  Maybe we should add this to @base argument javadoc in the TableOperations could clarify it, but itd be unrelated.

core/src/main/java/org/apache/iceberg/jdbc/JdbcTableOperations.java
```diff
@@ -123,6 +123,9 @@ public void doCommit(TableMetadata base, TableMetadata metadata) {
     } catch (SQLWarning e) {
       throw new UncheckedSQLException(e, "Database warning");
     } catch (SQLException e) {
+      if (e.getMessage().contains("constraint failed")) {
```
danielcweeks, line 126: This feels like a pretty non-portable way of handling constraint failed.  There may be a better way to identify this either using the `SQLState` string or with the nested exceptions in the provided `iterator()`.

rdblue: `SQLIntegrityConstraintViolationException` is handled above. The problem is that SQLite doesn't set anything correctly. It doesn't throw `SQLIntegrityConstraintViolationException`, or set `SQLState=23`, or include other exceptions. It uses its own exception class with this error message:

> [SQLITE_CONSTRAINT_PRIMARYKEY]  A PRIMARY KEY constraint failed (UNIQUE constraint failed: iceberg_tables.catalog_name, iceberg_tables.table_namespace, iceberg_tables.table_name)

It does set `e.vendorCode=19`, but that would require checking that the exception class contains `SQLite` so we're doing string matching either way.

I'll add a comment to make it clear what's happening here.

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
rdblue: I added a comment to make this clear.

rdblue: I added a comment to make this clear.

danielcweeks (approved):

rdblue:
Thanks for reviewing, @nastra, @szehon-ho, @kbendick, and @danielcweeks!

PR #4320, opened 2022-03-13
Core: Add MetadataUpdate.applyTo to re-apply changes

rdblue:
This updates `TableMetadata` and `MetadataUpdate` so that changes can be reconstructed in `TableMetadata.Builder` from a list of `MetadataUpdate` instances that are produced by the builder.

This also fixes a few minor issues with the builder:
* `lastAssignedPartitionId` was not public
* Supports set changes that use `-1` to refer to the last added schema, spec, or sort order
* Updates snapshot history timestamp based on whether the snapshot was added, rather than the `setCurrentSnapshot` call

This is part of #4241.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdate.java
@@ -22,11 +22,14 @@ package org.apache.iceberg;
 import java.io.Serializable;
 import java.util.Map;
 import java.util.Set;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 
 /**
  * Represents a change to table metadata.
  */
 public interface MetadataUpdate extends Serializable {
+  void applyTo(TableMetadata.Builder metadataBuilder);
+
   class AssignUUID implements MetadataUpdate {
     private final String uuid;
 
@@ -36,6 +39,11 @@ public interface MetadataUpdate extends Serializable {
 
     public String uuid() {
       return uuid;
+    }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      throw new UnsupportedOperationException("Not implemented");
     }
   }
 
@@ -48,6 +56,11 @@ public interface MetadataUpdate extends Serializable {
 
     public int formatVersion() {
       return formatVersion;
+    }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      metadataBuilder.upgradeFormatVersion(formatVersion);
     }
   }
 
@@ -67,6 +80,11 @@ public interface MetadataUpdate extends Serializable {
     public int lastColumnId() {
       return lastColumnId;
     }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      metadataBuilder.addSchema(schema, lastColumnId);
+    }
   }
 
   class SetCurrentSchema implements MetadataUpdate {
@@ -78,6 +96,11 @@ public interface MetadataUpdate extends Serializable {
 
     public int schemaId() {
       return schemaId;
+    }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      metadataBuilder.setCurrentSchema(schemaId);
     }
   }
 
@@ -91,6 +114,11 @@ public interface MetadataUpdate extends Serializable {
     public PartitionSpec spec() {
       return spec;
     }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      metadataBuilder.addPartitionSpec(spec);
+    }
   }
 
   class SetDefaultPartitionSpec implements MetadataUpdate {
@@ -103,6 +131,11 @@ public interface MetadataUpdate extends Serializable {
     public int specId() {
       return specId;
     }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      metadataBuilder.setDefaultPartitionSpec(specId);
+    }
   }
 
   class AddSortOrder implements MetadataUpdate {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/MetadataUpdate.java
```diff
@@ -37,6 +40,11 @@ public AssignUUID(String uuid) {
     public String uuid() {
       return uuid;
     }
+
+    @Override
+    public void applyTo(TableMetadata.Builder metadataBuilder) {
+      throw new UnsupportedOperationException("Not implemented");
```
rdblue, line 46: This is not part of the REST API, so I left it unimplemented. I'm not sure whether we will need this because tables have had UUIDs for a long time. It seems unlikely that someone will update from 0.6.0 or earlier directly to 0.14.0 and want to use the REST catalog.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -873,12 +891,18 @@ public Builder setCurrentSchema(int schemaId) {
 
       this.currentSchemaId = schemaId;
 
-      changes.add(new MetadataUpdate.SetCurrentSchema(schemaId));
+      if (lastAddedSchemaId != null && lastAddedSchemaId == schemaId) {
+        // use -1 to signal that current was set to the last added schema
+        changes.add(new MetadataUpdate.SetCurrentSchema(-1));
```
kbendick, line 896: Would this be something to make as a constant? Perhaps `USE_LAST_ADDED_SCHEMA = -1`?

EDIT: Looking at further code, we use the same constant for partition spec and many other things. So maybe this comment would be better placed above the builder class or something? Or using a more generic constant like `private static final int KEEP_EXISTING_ID = -1`?

I'm overall ok with this though and like the idea of using `-1` this way.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -1328,5 +1387,16 @@ private void setBranchSnapshot(Snapshot snapshot, String branch, Long currentTim
...
+
+    private boolean isAddedSnapshot(long snapshotId) {
+      return changes(MetadataUpdate.AddSnapshot.class)
+          .anyMatch(add -> add.snapshot().snapshotId() == snapshotId);
+    }
+
+    private <U extends MetadataUpdate> Stream<U> changes(Class<U> updateClass) {
+      return changes.stream()
+          .filter(updateClass::isInstance)
+          .map(updateClass::cast);
```
kbendick, line 1399: If this is the only place this is used, is it really needed? If so, maybe `changesFor(Class<U> updateClass)` would be better. Specifically as ther eis already a class field called `changes`.

kbendick (approved):
A few comments, but overall this LGTM.

Thank you @rdblue!

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -771,12 +774,15 @@ public static Builder buildFrom(TableMetadata base) {
...
-    private Map<String, SnapshotRef> refs;
+    private final Map<String, SnapshotRef> refs;
 
     // change tracking
     private final List<MetadataUpdate> changes;
     private final int startingChangeCount;
     private boolean discardChanges = false;
+    private Integer lastAddedSchemaId = null;
+    private Integer lastAddedSpecId = null;
+    private Integer lastAddedOrderId = null;
```
danielcweeks, line 785: nit: none of the other private varibles are explicitly set in definition.  Do we really need to set these to `null`?

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -821,6 +827,12 @@ private Builder(TableMetadata base) {
       this.sortOrdersById = Maps.newHashMap(base.sortOrdersById);
     }
 
+    public Builder withMetadataLocation(String newMetadataLocation) {
+      this.metadataLocation = newMetadataLocation;
+      this.discardChanges = true;
```
danielcweeks, line 832: What's the motivation for discarding changes automatically in this case?  Is it that we don't want to apply on top of unknown metadata?  A comment here would be helpful since we're performing a hidden action on this builder call.

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: We explicitly set variables that are not `final`. Using `final` guarantees that they are set in a constructor, but there is no guarantee for mutable fields.

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: I'll fix this. I think I want the caller to specifically call `discardChanges()` or fail validation if there are any changes.

The contract for `TableMetadata` is that the contents of the metadata location matches the table metadata. That's because the only way to get a `TableMetadata` with a non-null metadata location was to read it from the location using `TableMetadataParser`.

Now, table metadata is coming from a REST endpoint that also sends `metadataLocation`, so we need to be able to read the table metadata and re-attach the metadata location.

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: Added `LAST_ADDED`. Good suggestion, it's easier to read.

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: There are a few places where we go through the changes, so I think it's a good idea to keep this for future changes rather than implementing it slightly differently in multiple places.

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: Added a precondition to the build method and removed this.

danielcweeks (approved):

rdblue:
Thanks for the reviews, @danielcweeks and @kbendick!

PR #4322, opened 2022-03-13
Core: Add REST catalog table requests and responses

rdblue:
This adds REST request and response classes for table operations.

* `CreateTableRequest` is used to create tables
* `UpdateTableRequest` is used to update tables by [applying a list of `MetadataUpdate` changes](https://github.com/apache/iceberg/pull/4320)
* `LoadTableResponse` is used for create, load, and update
* `DropTableResponse` is used for drop

`UpdateTableRequest` includes a set of assertions from the REST spec, with some additions that were needed to get [tests](https://github.com/apache/iceberg/pull/4319) working:
* `AssertTableDoesNotExist` is used for create table transactions
* `AssertTableUUID` checks that the table UUID has not changed from the base metadata (prevents errors from concurrent drop/create)
* `AssertLastAssignedFieldId` is used to check that no concurrent operation has assigned schema field IDs that may conflict
* `AssertLastAssignedPartitionId` is used to check that no concurrent operation has assigned partition field IDs that may conflict

`UpdateTableRequest` will automatically add assertions for the requested changes.

This is part of #4241.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java
@@ -1,0 +1,399 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest.requests;
+
+import java.util.List;
+import java.util.Set;
+import org.apache.iceberg.MetadataUpdate;
+import org.apache.iceberg.SnapshotRef;
+import org.apache.iceberg.TableMetadata;
+import org.apache.iceberg.exceptions.CommitFailedException;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


rdblue:
This is blocked by #4320. Will rebase when that one is committed.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java
```diff
@@ -0,0 +1,399 @@
...
+        update((MetadataUpdate.AddSchema) update);
+      } else if (update instanceof MetadataUpdate.SetCurrentSchema) {
+        update((MetadataUpdate.SetCurrentSchema) update);
+      } else if (update instanceof MetadataUpdate.AddPartitionSpec) {
+        update((MetadataUpdate.AddPartitionSpec) update);
+      } else if (update instanceof MetadataUpdate.SetDefaultPartitionSpec) {
+        update((MetadataUpdate.SetDefaultPartitionSpec) update);
+      } else if (update instanceof MetadataUpdate.SetDefaultSortOrder) {
+        update((MetadataUpdate.SetDefaultSortOrder) update);
+      }
```
danielcweeks, line 151: you might want an `else` to throw at the end otherwise future updates (if there are any) may fall through.

rdblue: Other updates are okay. These methods just add requirements, so things like `AddSnapshot` are supposed to fall through without changing anything.

core/src/main/java/org/apache/iceberg/rest/responses/DropTableResponse.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/**
+ * Represents a REST response to drop a table.
+ */
+public class DropTableResponse {
+
+  // For Jackson to properly fail when deserializing, this needs to be boxed.
+  // Otherwise, the boolean is parsed according to "loose" javascript JSON rules.
+  private Boolean dropped;
```
danielcweeks, line 32: I'm a little confused about the conditions under which `dropped` would be `false`.  According to the spec, `404` should be returned if the table is not found.   `401/403` for auth.  So how do we get a `200/202` response with a `false` flag?

rdblue: Ah, we need to remove 404 from the spec. The catalog API requires that the catalog [should return false if the table does not exist](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/catalog/Catalog.java#L307) and does not allow throwing `NoSuchTableException`. I'll open a PR to clarify the REST spec.

danielcweeks (approved):

PR #4324, opened 2022-03-14
Spark: Simplify validation in BaseRewriteDataFilesSparkAction

rajarshisarkar:
This change simplifies the validation in `BaseRewriteDataFilesSparkAction`. 

The earlier validation was `!partialProgressEnabled || partialProgressEnabled && maxCommits > 0`. The right side of the OR condition is always evaluated when `partialProgressEnabled` is true. Hence, we need not to mention `partialProgressEnabled` explicitly on the right side of the OR condition.


rajarshisarkar:
@aokolnychyi @RussellSpitzer Can you please have a look once you get time?

RussellSpitzer (approved):

Issue #4051, opened 2022-02-06
How do I make Iceberg ignore corrupt (parquet, ORC, avro) file?


dubeme:
In our datalake, it is posible to corrupt files to exists in the same directory of a hive table. We've been using the Iceber `add_files` procedure to migrate those tables to Iceberg. Unfortunately, Iceberg will fail when it encounters a corrupt (_a file that doesn't match the file type Iceberg is looking for_) file. Below is a sample exception. **So, is it possible to tell Iceberg to ignore those files?** This is possible in Spark, setting `spark.sql.files.ignoreCorruptFiles=true`.

```
Error: Job aborted due to stage failure: Task 34 in stage 1.0 failed 4 times, most recent failure: Lost task 34.3 in stage 1.0 (TID 222) (vm-f1668438 executor 19): java.lang.RuntimeException: <REDACTED PATH> is not a Parquet file (length is too low: 0)
	at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:539)
	at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:776)
	at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileReader.open(ParquetFileReader.java:645)
	at org.apache.iceberg.parquet.ParquetUtil.fileMetrics(ParquetUtil.java:78)
	at org.apache.iceberg.data.TableMigrationUtil.getParquetMerics(TableMigrationUtil.java:145)
	at org.apache.iceberg.data.TableMigrationUtil.lambda$listPartition$3(TableMigrationUtil.java:113)
	at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:404)
	at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:214)
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:198)
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:190)
	at org.apache.iceberg.data.TableMigrationUtil.listPartition(TableMigrationUtil.java:112)
	at org.apache.iceberg.data.TableMigrationUtil.listPartition(TableMigrationUtil.java:80)
	at org.apache.iceberg.spark.SparkTableUtil.listPartition(SparkTableUtil.java:304)
[...]

PR #4325, opened 2022-03-14
Spark: Skip corrupt files in Spark Procedures and Actions

hililiwei:
Closes: https://github.com/apache/iceberg/issues/4051

 Skip corrupt files in Spark Procedures and Action

## Brief change log
The `'skip_corrupt_files'`  option is added to Spark `migrate/add_files` Procedures, which can be used to skip corrupted files.
```SQL
CALL catalog.system.migrate(
table => 'table',
skip_on_error => true)
```
```SQL
CALL catalog.system.add_files(
table => 'table',
source_table => 'source_table',
skip_on_error => true)
```
## Verifying this change

1. After generating two available data files, delete the second and create an empty file to verify that it works as expected.
TestAddFilesProcedure.java\TestMigrateTableProcedure.java

### DIFF
--- a/data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
+++ b/data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
@@ -69,26 +69,26 @@ public class TableMigrationUtil {
    * not file footers, will not be populated.
    *
    * @param partition map of column names to column values for the partition
-   * @param uri partition location URI
+   * @param partitionUri partition location URI
    * @param format partition format, avro, parquet or orc
    * @param spec a partition spec
    * @param conf a Hadoop conf
-   * @param metricsConfig a metrics conf
+   * @param metricsSpec a metrics conf
    * @param mapping a name mapping
    * @param skipOnError if true, skip files which cannot be imported into Iceberg
    * @return a List of DataFile
    */
   public static List<DataFile> listPartition(
-      Map<String, String> partition,
-      String uri,
+      Map<String, String> partitionPath,
+      String partitionUri,
       String format,
       PartitionSpec spec,
       Configuration conf,
-      MetricsConfig metricsConfig,
+      MetricsConfig metricsSpec,
       NameMapping mapping,
       boolean skipOnError) {
     return listPartition(
-        partition, uri, format, spec, conf, metricsConfig, mapping, 1, skipOnError);
+        partitionPath, partitionUri, format, spec, conf, metricsSpec, mapping, 1, skipOnError);
   }
 
   /**
@@ -100,33 +100,33 @@ public class TableMigrationUtil {
    * <p>Note: certain metrics, like NaN counts, that are only supported by iceberg file writers but
    * not file footers, will not be populated.
    *
-   * @param partition partition key, e.g., "a=1/b=2"
-   * @param uri partition location URI
+   * @param partitionPath partition key, e.g., "a=1/b=2"
+   * @param partitionUri partition location URI
    * @param format partition format, avro, parquet or orc
    * @param spec a partition spec
    * @param conf a Hadoop conf
-   * @param metricsConfig a metrics conf
+   * @param metricsSpec a metrics conf
    * @param mapping a name mapping
    * @return a List of DataFile
    */
   public static List<DataFile> listPartition(
-      Map<String, String> partition,
-      String uri,
+      Map<String, String> partitionPath,
+      String partitionUri,
       String format,
       PartitionSpec spec,
       Configuration conf,
-      MetricsConfig metricsConfig,
+      MetricsConfig metricsSpec,
       NameMapping mapping) {
     return listPartition(
-        partition,
-        uri,
+        partitionPath,
+        partitionUri,
         format,
         spec,
         conf,
-        metricsConfig,
+        metricsSpec,
         mapping,
         1,
-        false /* skipOnErrorOrNot */);
+        false /* skip on error or not */);
   }
 
   /**
@@ -168,9 +168,29 @@ public class TableMigrationUtil {
         metricsSpec,
         mapping,
         parallelism,
-        false /* skipOnErrorOrNot */);
+        false /* skip on error or not */);
   }
 
+  /**
+   * Returns the data files in a partition by listing the partition location.
+   *
+   * <p>For Parquet and ORC partitions, this will read metrics from the file footer. For Avro
+   * partitions, metrics are set to null.
+   *
+   * <p>Note: certain metrics, like NaN counts, that are only supported by iceberg file writers but
+   * not file footers, will not be populated.
+   *
+   * @param partitionPath partition key, e.g., "a=1/b=2"
+   * @param partitionUri partition location URI
+   * @param format partition format, avro, parquet or orc
+   * @param spec a partition spec
+   * @param conf a Hadoop conf
+   * @param metricsSpec a metrics conf
+   * @param mapping a name mapping
+   * @param parallelism size of the thread pool
+   * @param skipOnError if true, skip files which cannot be imported into Iceberg
+   * @return a List of DataFile
+   */
   public static List<DataFile> listPartition(
       Map<String, String> partitionPath,
       String partitionUri,
@@ -199,7 +219,10 @@ public class TableMigrationUtil {
       Tasks.Builder<Integer> task = Tasks.range(fileStatus.size());
 
       if (skipOnError) {
-        task.suppressFailureWhenFinished();
+        task.onFailure(
+                (index, exc) ->
+                    LOG.warn("Failed to build data file from: {}", fileStatus.get(index), exc))
+            .suppressFailureWhenFinished();
       } else {
         task.stopOnFailure().throwFailureWhenFinished();
       }
@@ -237,9 +260,7 @@ public class TableMigrationUtil {
           LOG.warn("Skipping unknown partition format: {} - {}", format, partitionUri);
         } else {
           throw new UnsupportedOperationException(
-              String.format(
-                  "Unknown partition format: %s - %s, set skip_on_error=true to skip it.",
-                  format, partitionUri));
+              String.format("Unknown partition format: %s - %s", format, partitionUri));
         }
       }
       return Arrays.stream(datafiles).filter(Objects::nonNull).collect(Collectors.toList());
--- a/data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
+++ b/data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
@@ -23,6 +23,7 @@ import java.io.UncheckedIOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadPoolExecutor;
@@ -47,8 +48,12 @@ import org.apache.iceberg.parquet.ParquetUtil;
 import org.apache.iceberg.relocated.com.google.common.util.concurrent.MoreExecutors;
 import org.apache.iceberg.relocated.com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.iceberg.util.Tasks;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class TableMigrationUtil {
+  private static final Logger LOG = LoggerFactory.getLogger(TableMigrationUtil.class);
+
   private static final PathFilter HIDDEN_PATH_FILTER =
       p -> !p.getName().startsWith("_") && !p.getName().startsWith(".");
 
@@ -70,6 +75,38 @@ public class TableMigrationUtil {
    * @param conf a Hadoop conf
    * @param metricsConfig a metrics conf
    * @param mapping a name mapping
+   * @param skipOnError if true, skip files which cannot be imported into Iceberg
+   * @return a List of DataFile
+   */
+  public static List<DataFile> listPartition(
+      Map<String, String> partition,
+      String uri,
+      String format,
+      PartitionSpec spec,
+      Configuration conf,
+      MetricsConfig metricsConfig,
+      NameMapping mapping,
+      boolean skipOnError) {
+    return listPartition(
+        partition, uri, format, spec, conf, metricsConfig, mapping, 1, skipOnError);
+  }
+
+  /**
+   * Returns the data files in a partition by listing the partition location.
+   *
+   * <p>For Parquet and ORC partitions, this will read metrics from the file footer. For Avro
+   * partitions, metrics are set to null.
+   *
+   * <p>Note: certain metrics, like NaN counts, that are only supported by iceberg file writers but
+   * not file footers, will not be populated.
+   *
+   * @param partition partition key, e.g., "a=1/b=2"
+   * @param uri partition location URI
+   * @param format partition format, avro, parquet or orc
+   * @param spec a partition spec
+   * @param conf a Hadoop conf
+   * @param metricsConfig a metrics conf
+   * @param mapping a name mapping
    * @return a List of DataFile
    */
   public static List<DataFile> listPartition(
@@ -80,7 +117,16 @@ public class TableMigrationUtil {
       Configuration conf,
       MetricsConfig metricsConfig,
       NameMapping mapping) {
-    return listPartition(partition, uri, format, spec, conf, metricsConfig, mapping, 1);
+    return listPartition(
+        partition,
+        uri,
+        format,
+        spec,
+        conf,
+        metricsConfig,
+        mapping,
+        1,
+        false /* skipOnErrorOrNot */);
   }
 
   /**
@@ -112,6 +158,29 @@ public class TableMigrationUtil {
       MetricsConfig metricsSpec,
       NameMapping mapping,
       int parallelism) {
+
+    return listPartition(
+        partitionPath,
+        partitionUri,
+        format,
+        spec,
+        conf,
+        metricsSpec,
+        mapping,
+        parallelism,
+        false /* skipOnErrorOrNot */);
+  }
+
+  public static List<DataFile> listPartition(
+      Map<String, String> partitionPath,
+      String partitionUri,
+      String format,
+      PartitionSpec spec,
+      Configuration conf,
+      MetricsConfig metricsSpec,
+      NameMapping mapping,
+      int parallelism,
+      boolean skipOnError) {
     ExecutorService service = null;
     try {
       List<String> partitionValues =
@@ -127,8 +196,13 @@ public class TableMigrationUtil {
               .filter(FileStatus::isFile)
               .collect(Collectors.toList());
       DataFile[] datafiles = new DataFile[fileStatus.size()];
-      Tasks.Builder<Integer> task =
-          Tasks.range(fileStatus.size()).stopOnFailure().throwFailureWhenFinished();
+      Tasks.Builder<Integer> task = Tasks.range(fileStatus.size());
+
+      if (skipOnError) {
+        task.suppressFailureWhenFinished();
+      } else {
+        task.stopOnFailure().throwFailureWhenFinished();
+      }
 
       if (parallelism > 1) {
         service = migrationService(parallelism);
@@ -159,9 +233,16 @@ public class TableMigrationUtil {
                   buildDataFile(fileStatus.get(index), partitionValues, spec, metrics, "orc");
             });
       } else {
-        throw new UnsupportedOperationException("Unknown partition format: " + format);
+        if (skipOnError) {
+          LOG.warn("Skipping unknown partition format: {} - {}", format, partitionUri);
+        } else {
+          throw new UnsupportedOperationException(
+              String.format(
+                  "Unknown partition format: %s - %s, set skip_on_error=true to skip it.",
+                  format, partitionUri));
+        }
       }
-      return Arrays.asList(datafiles);
+      return Arrays.stream(datafiles).filter(Objects::nonNull).collect(Collectors.toList());
     } catch (IOException e) {
       throw new RuntimeException("Unable to list files in partition: " + partitionUri, e);
     } finally {
--- a/docs/spark-procedures.md
+++ b/docs/spark-procedures.md
@@ -479,11 +479,12 @@ By default, the original table is retained with the name `table_BACKUP_`.
 
 #### Usage
 
-| Argument Name | Required? | Type | Description |
-|---------------|-----------|------|-------------|
-| `table`       | ✔️  | string | Name of the table to migrate |
-| `properties`  | ️   | map<string, string> | Properties for the new Iceberg table |
-| `drop_backup` |   | boolean | When true, the original table will not be retained as backup (defaults to false) |
+| Argument Name     | Required? | Type                | Description                                                                        |
+|-------------------|-----------|---------------------|------------------------------------------------------------------------------------|
+| `table`           | ✔️        | string              | Name of the table to migrate                                                       |
+| `properties`      | ️         | map<string, string> | Properties for the new Iceberg table                                               |
+| `drop_backup`     | ️         | boolean             | When true, the original table will not be retained as backup (defaults to false)   |
+| `skip_on_error`   | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (false by default)       |
 
 #### Output
 
@@ -522,6 +523,7 @@ will then treat these files as if they are part of the set of files  owned by Ic
 | `source_table`          | ✔️        | string              | Table where files should come from, paths are also possible in the form of \`file_format\`.\`path\` |
 | `partition_filter`      | ️         | map<string, string> | A map of partitions in the source table to import from                                              |
 | `check_duplicate_files` | ️         | boolean             | Whether to prevent files existing in the table from being added (defaults to true)                  |
+| `skip_on_error`         | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (false by default)                        |
 
 Warning : Schema is not validated, adding files with different schema to the Iceberg table will cause issues.
 
--- a/docs/spark-procedures.md
+++ b/docs/spark-procedures.md
@@ -484,7 +484,7 @@ By default, the original table is retained with the name `table_BACKUP_`.
 | `table`           | ✔️        | string              | Name of the table to migrate                                                       |
 | `properties`      | ️         | map<string, string> | Properties for the new Iceberg table                                               |
 | `drop_backup`     | ️         | boolean             | When true, the original table will not be retained as backup (defaults to false)   |
-| `skip_on_error`   | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (false by default)       |
+| `skip_on_error`   | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (defaults to false)      |
 
 #### Output
 
@@ -523,7 +523,7 @@ will then treat these files as if they are part of the set of files  owned by Ic
 | `source_table`          | ✔️        | string              | Table where files should come from, paths are also possible in the form of \`file_format\`.\`path\` |
 | `partition_filter`      | ️         | map<string, string> | A map of partitions in the source table to import from                                              |
 | `check_duplicate_files` | ️         | boolean             | Whether to prevent files existing in the table from being added (defaults to true)                  |
-| `skip_on_error`         | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (false by default)                        |
+| `skip_on_error`         | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (defaults to false)                        |
 
 Warning : Schema is not validated, adding files with different schema to the Iceberg table will cause issues.
 
--- a/spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
+++ b/spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
@@ -974,6 +974,55 @@ public class TestAddFilesProcedure extends SparkExtensionsTestBase {
         sql("SELECT * FROM %s ORDER BY id", tableName));
   }
 
+  @Test
+  public void testSkipOnError() throws IOException {
+    createUnpartitionedFileTable("parquet");
+
+    String createIceberg =
+        "CREATE TABLE %s (id Integer, name String, dept String, subdept String) USING iceberg";
+
+    sql(createIceberg, tableName);
+
+    // Create an empty(considered corrupted) file.
+    Assert.assertTrue(new File(fileTableDir + File.separator + "corrupt.parquet").createNewFile());
+
+    AssertHelpers.assertThrows(
+        "Throws an exception when a corrupted file is encountered",
+        RuntimeException.class,
+        "not a Parquet file (length is too low: 0)",
+        () ->
+            sql(
+                "CALL %s.system.add_files(" + "table => '%s', " + "source_table => '%s')",
+                catalogName, tableName, sourceTableName));
+
+    AssertHelpers.assertThrows(
+        "Throws an exception when a corrupted file is encountered",
+        RuntimeException.class,
+        "not a Parquet file (length is too low: 0)",
+        () ->
+            sql(
+                "CALL %s.system.add_files("
+                    + "table => '%s', "
+                    + "source_table => '%s',"
+                    + "skip_on_error => false)",
+                catalogName, tableName, sourceTableName));
+
+    List<Object[]> result =
+        sql(
+            "CALL %s.system.add_files("
+                + "table => '%s',"
+                + "source_table => '%s',"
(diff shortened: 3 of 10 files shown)


data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -97,8 +102,10 @@ private TableMigrationUtil() {
               .collect(Collectors.toList());
       DataFile[] datafiles = new DataFile[fileStatus.size()];
       Tasks.Builder<Integer> task = Tasks.range(fileStatus.size())
-              .stopOnFailure()
-              .throwFailureWhenFinished();
+          .throwFailureWhenFinished(!skipCorruptFiles);
+      if (!skipCorruptFiles) {
+        task.stopOnFailure();
```
kbendick, line 107: Nit: Instead of updating the `throwFailureWhenFinished` and then having this `if` statement, it might be cleaner to simply place the `throwFailureWhenFinished` inside of the if statement as well.

kbendick (commented):
Thanks @hililiwei!

Still reviewing this, but can you please add to the PR description a summary of the issue?

hililiwei:
cc @kbendick @RussellSpitzer , PTAL, Thx. 😸

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -96,9 +126,14 @@ private TableMigrationUtil() {
...
-      Tasks.Builder<Integer> task = Tasks.range(fileStatus.size())
-              .stopOnFailure()
-              .throwFailureWhenFinished();
+      Tasks.Builder<Integer> task = Tasks.range(fileStatus.size());
+
+      if (skipCorruptFiles) {
+        task.suppressFailureWhenFinished();
+      } else {
+        task.stopOnFailure()
+            .throwFailureWhenFinished();
```
RussellSpitzer, line 135: nit: one line

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -121,9 +156,13 @@ private TableMigrationUtil() {
           datafiles[index] = buildDataFile(fileStatus.get(index), partitionKey, spec, metrics, "orc");
         });
       } else {
-        throw new UnsupportedOperationException("Unknown partition format: " + format);
+        if (skipCorruptFiles) {
+          LOG.warn("Unknown partition format: {}", format);
```
RussellSpitzer, line 160: Aren't we fine with just throwing an error here because we suppress them?

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -140,7 +179,7 @@ private static Metrics getAvroMerics(Path path,  Configuration conf) {
       return new Metrics(rowCount, null, null, null, null);
     } catch (UncheckedIOException e) {
       throw new RuntimeException("Unable to read Avro file: " +
-              path, e);
```
RussellSpitzer, line 143: Good change in style, but this is probably fine as a single line too

docs/spark/spark-procedures.md
```diff
@@ -417,10 +417,11 @@ To leave the original table intact while testing, use [`snapshot`](#snapshot) to
...
 
-| Argument Name | Required? | Type | Description |
-|---------------|-----------|------|-------------|
-| `table`       | ✔️  | string | Name of the table to migrate |
-| `properties`  | ️   | map<string, string> | Properties for the new Iceberg table |
+| Argument Name        | Required? | Type                | Description                                    |
+|----------------------|-----------|---------------------|------------------------------------------------|
+| `table`              | ✔️        | string              | Name of the table to migrate                   |
+| `properties`         | ️         | map<string, string> | Properties for the new Iceberg table           |
+| `skip_corrupt_files` | ️         | boolean             | If true, skip corrupt files (false by default) |
```
RussellSpitzer, line 424: I think we need a better term than "corrupt" here. Maybe "skip on error"? Then the description is "If true, skips files which cannot be imported into Iceberg"?

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
hililiwei: It is not within the `Task`,  suppression exception does not take effect here.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
hililiwei: updated

docs/spark/spark-procedures.md
hililiwei: Changed in the latest commite.

hililiwei:
hi @RussellSpitzer , could you please task a took when you are available?

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
RussellSpitzer: In the warn can you just add 
`"Skipping unknown partition format: {} - {}", format, partitionUri)`
Or whatever makes sense so a user has the full context here?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
```diff
@@ -759,6 +759,51 @@ public void testPartitionedImportFromEmptyPartitionDoesNotThrow() {
         sql("SELECT * FROM %s ORDER BY id", tableName));
   }
 
+  @Test
+  public void testSkipOnError() throws IOException {
```
RussellSpitzer, line 763: I would probably split this into two tests one for it skipping on error, and one fore it not skipping on error. We should also add one for an invalid partition format if possible although i'm not sure we can manage that easily.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
```diff
@@ -759,6 +759,51 @@ public void testPartitionedImportFromEmptyPartitionDoesNotThrow() {
...
+    createUnpartitionedFileTable("parquet");
+
+    List<Object[]> source = sql("SELECT * FROM %s ORDER BY id", sourceTableName);
+    Assert.assertEquals(String.format("Rows in source table did not match\nExpected :%s rows \nFound    :%s",
```
RussellSpitzer, line 767: Usually we just have our expected stored as a List<Object[]> and compare against that. In this case I don't think you really need to check that the createUnpartitioned method worked correctly.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
RussellSpitzer: Actually, looking at this again. I think this test with both the Skip off and ON is fine. We really don't have a better way of checking for invalid files

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
```diff
@@ -759,6 +759,51 @@ public void testPartitionedImportFromEmptyPartitionDoesNotThrow() {
...
+        "CREATE TABLE %s (id Integer, name String, dept String, subdept String) USING iceberg";
+
+    sql(createIceberg, tableName);
+
+    File[] expectedFiles = fileTableDir.listFiles((dir, name) -> !name.endsWith("crc") && !name.contains("_SUCCESS"));
+
+    Assert.assertEquals("Expected number of source files", 2, expectedFiles.length);
+
+    // Corrupt the second file
+    Assume.assumeTrue("Delete source file!", expectedFiles[1].delete());
```
RussellSpitzer, line 780: These are not the correct usages of Assume. This will cause the test to be skipped if the operation returns false rather than fail.

In general I think you should only assert on conditions that prove the test has failed or succeeded, here we have a lot of asserts that check whether test prep methods work which I think makes the test a bit harder to follow. For example we make very sure there are 2 source files but this isn't really relevant except that the test chooses two delete the second file.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
```diff
@@ -759,6 +759,51 @@ public void testPartitionedImportFromEmptyPartitionDoesNotThrow() {
...
+            "table => '%s'," +
+            "source_table => '%s'," +
+            "skip_on_error => true)",
+        catalogName, tableName, sourceTableName);
+
+    Assert.assertEquals(1L, result);
+
+    List<Object[]> actual = sql("SELECT * FROM %s ORDER BY id", tableName);
+
+    Assert.assertEquals(String.format("Rows in table did not match\nExpected :%s rows \nFound    :%s",
```
RussellSpitzer, line 803: Rather than this, you can just check against ```unpartitionedDF ``` which has all of the records that should be in the table

RussellSpitzer (commented):
One suggestion for the logging,

A couple fixes I think may be required for the tests

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
hililiwei: removed.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
hililiwei: Thanks for the guidance. On reflection, apart from what you mentioned, this part is indeed a bit redundant, and we can achieve the same effect by creating an empty file with the `parquet` suffix.

hililiwei:
> One suggestion for the logging,
> 
> A couple fixes I think may be required for the tests

Updated according to the latest feedback. Thx.

hililiwei:
hi @RussellSpitzer , could you please take a look when you are available?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -271,12 +271,12 @@ public static List<SparkPartition> getPartitionsByFilter(SparkSession spark, Tab
...
    * @return a List of DataFile
    * @deprecated use {@link TableMigrationUtil#listPartition(Map, String, String, PartitionSpec, Configuration,
-   * MetricsConfig, NameMapping)}
+   * MetricsConfig, NameMapping, boolean)}
    */
   @Deprecated
   public static List<DataFile> listPartition(SparkPartition partition, PartitionSpec spec,
                                              SerializableConfiguration conf, MetricsConfig metricsConfig) {
-    return listPartition(partition, spec, conf, metricsConfig, null);
+    return listPartition(partition, spec, conf, metricsConfig, null, false);
```
RussellSpitzer, line 279: we can't change deprecated methods

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -375,12 +376,67 @@ private static Iterator<ManifestFile> buildManifest(SerializableConfiguration co
...
+   * @param sourceTableIdent an identifier of the source Spark table
+   * @param targetTable an Iceberg table where to import the data
+   * @param stagingDir a staging directory to store temporary manifest files
    * @param checkDuplicateFiles if true, throw exception if import results in a duplicate data file
+   * @param skipOnError if true, skip files which cannot be imported into Iceberg
    */
   public static void importSparkTable(SparkSession spark, TableIdentifier sourceTableIdent, Table targetTable,
                                       String stagingDir, Map<String, String> partitionFilter,
-                                      boolean checkDuplicateFiles) {
+                                      boolean checkDuplicateFiles, boolean skipOnError) {
```
RussellSpitzer, line 439: This breaks a public api, we would need to keep the older version without the skipOnError arg, although we can mark the older version as deprecated

RussellSpitzer: Oh nvm, i see it below

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -289,16 +289,17 @@ public static List<DataFile> listPartition(SparkPartition partition, PartitionSp
...
    * @return a List of DataFile
    * @deprecated use {@link TableMigrationUtil#listPartition(Map, String, String, PartitionSpec, Configuration,
-   * MetricsConfig, NameMapping)}
+   * MetricsConfig, NameMapping, boolean)}
    */
   @Deprecated
   public static List<DataFile> listPartition(SparkPartition partition, PartitionSpec spec,
                                              SerializableConfiguration conf, MetricsConfig metricsConfig,
-                                             NameMapping mapping) {
+                                             NameMapping mapping, boolean skipOnError) {
```
RussellSpitzer, line 300: Cannot change the signature of a deprecated method

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkActionOptions.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.apache.iceberg.spark.actions;
+
+public class SparkActionOptions {
+
+  private SparkActionOptions() {
+  }
+
+  public static final String SKIP_ON_ERROR = "skip_on_error";
```
RussellSpitzer, line 27: While I think it makes sense to have these final constants, they only apply to Snapshot and Migrate, and I think this may not be the right place for them then? Or they need a java doc saying that this only applies to those particular actions.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -375,12 +376,67 @@ private static Iterator<ManifestFile> buildManifest(SerializableConfiguration co
...
    * @param targetTable an Iceberg table where to import the data
    * @param stagingDir a staging directory to store temporary manifest files
-   * @param partitionFilter only import partitions whose values match those in the map, can be partially defined
+   */
+  public static ImportSparkTableBuilder importSparkTableBuilder(SparkSession spark, TableIdentifier sourceTableIdent,
+                                                                Table targetTable, String stagingDir) {
+    return new ImportSparkTableBuilder(spark, sourceTableIdent, targetTable, stagingDir);
+  }
+
+  public static class ImportSparkTableBuilder {
```
RussellSpitzer, line 385: Do we need the builder here? I'm not sure this method is going to change much more in the future and we tend to want to limit Builders to those apis we think will evolve a lot in the future.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -496,7 +580,8 @@ private static void importUnpartitionedSparkTable(SparkSession spark, TableIdent
    * @param checkDuplicateFiles if true, throw exception if import results in a duplicate data file
    */
   public static void importSparkPartitions(SparkSession spark, List<SparkPartition> partitions, Table targetTable,
-                                           PartitionSpec spec, String stagingDir, boolean checkDuplicateFiles) {
+                                           PartitionSpec spec, String stagingDir, boolean checkDuplicateFiles,
```
RussellSpitzer, line 583: Change to a public api needs a fallback, 

``` public static void importSparkPartitions(SparkSession spark, List<SparkPartition> partitions, Table targetTable, PartitionSpec spec, String stagingDir, boolean checkDuplicateFiles ) ```

File: spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkActionOptions.java (hunk 13 above)
Comment (hililiwei) [reply spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkActionOptions.java:27]: Have been added the javadoc

I introduce it here to migrate all Spark action options, such as the following:
https://github.com/apache/iceberg/blob/8155e5e8129365ed8d33b8cced2e851734a302c5/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/DeleteReachableFilesSparkAction.java#L56-L57

https://github.com/apache/iceberg/blob/8155e5e8129365ed8d33b8cced2e851734a302c5/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/ExpireSnapshotsSparkAction.java#L71-L72

https://github.com/apache/iceberg/blob/8155e5e8129365ed8d33b8cced2e851734a302c5/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteManifestsSparkAction.java#L86-L87

If we don't think it makes sense, I can remove it

File: spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java [hunk 16]
```diff
@@ -431,7 +511,8 @@ public static void importSparkTable(SparkSession spark, TableIdentifier sourceTa
    */
   public static void importSparkTable(SparkSession spark, TableIdentifier sourceTableIdent, Table targetTable,
                                       String stagingDir, boolean checkDuplicateFiles) {
-    importSparkTable(spark, sourceTableIdent, targetTable, stagingDir, Collections.emptyMap(), checkDuplicateFiles);
+    importSparkTable(spark, sourceTableIdent, targetTable, stagingDir, Collections.emptyMap(), checkDuplicateFiles,
+        false);
```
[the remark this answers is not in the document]
hililiwei: added, thx.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
hililiwei: It's been removed.

Because this method takes a lot of parameters and can not set default parameter values like scala, so i added a Builder. :sweat_smile:

hililiwei: Also, I had to create multiple methods to distinguish checkDuplicateFiles from skipOnError.  It can not specify parameter name when values are passed in, as Scala does. 😂

hililiwei:
> This looks really close to me, the only issues I think are remaining are
> 
> 1. We cannot change any deprecated apis
> 2. Any public apis that are changed need to remain, additional apis can be added and the originals can be deprecated
> 3. I'm not sure about where the Action Constants should go. I think the SparkActionOptions is maybe too general but is probably fine if we document it with some Javadocs. As Is the option is there but there is no way of determining which actions it applies too. Ideally I would have had Migrate + Snapshot inherit from something with the option. It seems like putting it API would probably be too much, but maybe we just add an interface that they both can inherit SparkCreateAction, which has the option as a public field?

I added some docs to `SparkActionOptions`, and I originally wanted to move all the Spark action options to it (not in this PR). I found that we already have `BaseSparkAction` , maybe we can put these options in it (i'm not sure)?

rdblue:
@nastra, can you take a look at this PR? It sounds like it's close.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
```diff
@@ -406,6 +408,76 @@ public void testMigrateUnpartitioned() throws Exception {
...
+        .saveAsTable(source);
+
+    List<File> expectedFiles = expectedFiles(source).collect(Collectors.toList());
+
+    Assert.assertEquals("Expected number of source files", 2, expectedFiles.size());
+
+    // Corrupt the second file
+    File file = expectedFiles.get(1);
+    Assume.assumeTrue("Delete source file!", file.delete());
+    Assume.assumeTrue("Create a empty source file!", file.createNewFile());
```
nastra, line 448: I don't think we should be using `assumeTrue` here as that would silently skip the test. I'd suggest to rather add an assertion here

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
```diff
@@ -406,6 +408,76 @@ public void testMigrateUnpartitioned() throws Exception {
...
+    validateTables(source, dest);
+
+    SparkTable destTable = loadTable(dest);
+    Assert.assertEquals(
+        "Provider should be iceberg",
+        "iceberg",
+        destTable.properties().get(TableCatalog.PROP_PROVIDER));
+    List<Row> actual = spark.table(dest).collectAsList();
+
+    Assert.assertEquals(
```
nastra, line 471: I'd rather suggest to use `Assertions.assertThat(actual).hasSize(3);` as that is easier to debug when the assertion ever fails. 
For example, I've modified it locally to `Assertions.assertThat(actual).hasSize(2);` and the output shows what the actual content was, which makes debugging (especially on CI failures) much easier.

```
java.lang.AssertionError: 
Expected size: 2 but was: 3 in:
[[1,a], [2,b], [3,c]]
```

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
```diff
@@ -406,6 +408,76 @@ public void testMigrateUnpartitioned() throws Exception {
...
+        .table(baseTableName)
+        .write()
+        .mode(SaveMode.Append)
+        .format(format)
+        .partitionBy(partitionColumns.toSeq())
+        .saveAsTable(source);
+
+    List<File> expectedFiles = expectedFiles(source).collect(Collectors.toList());
+
+    Assert.assertEquals("Expected number of source files", 2, expectedFiles.size());
```
nastra, line 443:
```suggestion
    Assertions.assertThat(expectedFiles).hasSize(2);
```

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -139,9 +213,18 @@ public static List<DataFile> listPartition(
                   buildDataFile(fileStatus.get(index), partitionKey, spec, metrics, "orc");
             });
       } else {
-        throw new UnsupportedOperationException("Unknown partition format: " + format);
+        if (skipOnError) {
+          LOG.warn("Skipping unknown partition format: {} - {}", format, partitionUri);
+        } else {
+          throw new UnsupportedOperationException(
```
nastra, line 219: it looks like this part is actually never hit in the tests and I wonder how difficult it would be to have a test that would hit this exception

spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
```diff
@@ -974,6 +974,55 @@ public void testPartitionedImportFromEmptyPartitionDoesNotThrow() {
...
+
+    String createIceberg =
+        "CREATE TABLE %s (id Integer, name String, dept String, subdept String) USING iceberg";
+
+    sql(createIceberg, tableName);
+
+    // Create an empty(considered corrupted) file.
+    Assert.assertTrue(new File(fileTableDir + File.separator + "corrupt.parquet").createNewFile());
+
+    AssertHelpers.assertThrows(
```
nastra, line 989: just an FYI that we're trying to move people away from using `AssertHelpers` (https://github.com/apache/iceberg/pull/6977)

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
nastra: @hililiwei looks like most of my comments are marked as `outdated` due to a new push, but would be still good if you could address those

docs/spark-procedures.md
```diff
@@ -425,11 +425,12 @@ By default, the original table is retained with the name `table_BACKUP_`.
...
-| Argument Name | Required? | Type | Description |
-|---------------|-----------|------|-------------|
-| `table`       | ✔️  | string | Name of the table to migrate |
-| `properties`  | ️   | map<string, string> | Properties for the new Iceberg table |
-| `drop_backup` |   | boolean | When true, the original table will not be retained as backup (defaults to false) |
+| Argument Name   | Required? | Type | Description |
+|-----------------|-----------|------|-------------|
+| `table`         | ✔️        | string | Name of the table to migrate |
+| `properties`    | ️         | map<string, string> | Properties for the new Iceberg table |
+| `drop_backup`   |           | boolean | When true, the original table will not be retained as backup (defaults to false) |
```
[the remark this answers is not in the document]
hililiwei: Repaired.

spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestAddFilesProcedure.java
hililiwei: Thanks. I've made a replacement.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
hililiwei: UT `testMigrateUnknownFormatTableSkipOnError` was added to try to cover this case.

nastra: thanks

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
nastra: this would be also good to apply in all the new tests as it makes understanding test failures much easier

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -80,7 +117,16 @@ public static List<DataFile> listPartition(
...
+    return listPartition(
+        partition,
+        uri,
+        format,
+        spec,
+        conf,
+        metricsConfig,
+        mapping,
+        1,
+        false /* skipOnErrorOrNot */);
```
rdblue, line 129: Comments should be formatted using separate words.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -139,9 +213,16 @@ public static List<DataFile> listPartition(
                   buildDataFile(fileStatus.get(index), partitionKey, spec, metrics, "orc");
             });
       } else {
-        throw new UnsupportedOperationException("Unknown partition format: " + format);
+        if (skipOnError) {
+          LOG.warn("Skipping unknown partition format: {} - {}", format, partitionUri);
+        } else {
+          throw new UnsupportedOperationException(
+              String.format(
+                  "Unknown partition format: %s - %s, set skip_on_error=true to skip it.",
```
rdblue, line 221: This isn't the correct place to instruct the caller about how to avoid the error. This can be used through different paths, but only the procedure has `skip_on_error`.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -92,6 +138,29 @@ public static List<DataFile> listPartition(
...
+        format,
+        spec,
+        conf,
+        metricsSpec,
+        mapping,
+        parallelism,
+        false /* skipOnErrorOrNot */);
+  }
+
+  public static List<DataFile> listPartition(
```
rdblue, line 154: Javadoc is missing.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -107,8 +176,13 @@ public static List<DataFile> listPartition(
               .filter(FileStatus::isFile)
               .collect(Collectors.toList());
       DataFile[] datafiles = new DataFile[fileStatus.size()];
-      Tasks.Builder<Integer> task =
-          Tasks.range(fileStatus.size()).stopOnFailure().throwFailureWhenFinished();
+      Tasks.Builder<Integer> task = Tasks.range(fileStatus.size());
+
+      if (skipOnError) {
+        task.suppressFailureWhenFinished();
```
rdblue, line 182: I think it would be better to skip based on specific failures. As this is written, this would skip files when there is a transient S3 error, like a 503 "slow down". It would be better to handle errors like `ParquetDecodingException` to skip instead.

I don't think that would be an easy thing to implement, so I would say that this should at least log and should include retries for certain exceptions types.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
hililiwei: added.

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
hililiwei: removed

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
hililiwei: done.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -494,14 +562,21 @@ public static void importSparkTable(
...
       SparkSession spark, TableIdentifier sourceTableIdent, Table targetTable, String stagingDir) {
     importSparkTable(
-        spark, sourceTableIdent, targetTable, stagingDir, Collections.emptyMap(), false);
+        spark,
+        sourceTableIdent,
+        targetTable,
+        stagingDir,
+        Collections.emptyMap(),
+        false,
+        false /* skip on error or not */);
```
RussellSpitzer, line 571: not sure the comment here is necessary

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
```diff
@@ -978,6 +1044,11 @@ private void validateTables(String source, String dest)
 
   private long expectedFilesCount(String source)
       throws NoSuchDatabaseException, NoSuchTableException, ParseException {
+    return expectedFiles(source).count();
+  }
+
+  private Stream<File> expectedFiles(String source)
```
RussellSpitzer, line 1050: Why do we need this change?

RussellSpitzer: Seems like we never really use this as a stream? Could we just make it a list?

RussellSpitzer (approved):
I think this is pretty much ready to go. I'm a little reluctant to keep growing the api of listPartitions but this makes sense to me as a good new feature. @rdblue any other comments?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
hililiwei: > Why do we need this change?

Existing cases that get the length of the result are sufficient, but newly added cases need to get the elements.



> Seems like we never really use this as a stream? Could we just make it a list?

Done

hililiwei:
@RussellSpitzer @rdblue @nastra could you please take another look when you have chance?

docs/spark-procedures.md
```diff
@@ -432,11 +432,12 @@ By default, the original table is retained with the name `table_BACKUP_`.
...
-|---------------|-----------|------|-------------|
-| `table`       | ✔️  | string | Name of the table to migrate |
-| `properties`  | ️   | map<string, string> | Properties for the new Iceberg table |
-| `drop_backup` |   | boolean | When true, the original table will not be retained as backup (defaults to false) |
+| Argument Name     | Required? | Type                | Description                                                                        |
+|-------------------|-----------|---------------------|------------------------------------------------------------------------------------|
+| `table`           | ✔️        | string              | Name of the table to migrate                                                       |
+| `properties`      | ️         | map<string, string> | Properties for the new Iceberg table                                               |
+| `drop_backup`     | ️         | boolean             | When true, the original table will not be retained as backup (defaults to false)   |
+| `skip_on_error`   | ️         | boolean             | If true, skip files which cannot be imported into Iceberg (false by default)       |
```
RussellSpitzer, line 440: line 439 has "defaults to false", I would pick just one wording and that seems to be more common in this doc

RussellSpitzer (approved):
Small inconsistency in the docs, other than that I think this is good to go.

hililiwei:
@RussellSpitzer Thank you for your review. It has been revised.

PR #4329, opened 2022-03-15
Flink: FLIP-27 source enumerator help classes

stevenzwu:
It mainly contains classes for streaming read (continuous split discovery) along with config and enumeration position classes.


flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -35,7 +36,8 @@
 /**
  * Context object with optional arguments for a Flink Scan.
  */
-class ScanContext implements Serializable {
+@Internal
+public class ScanContext implements Serializable {
```
stevenzwu, line 40: This is made public because it is accessed by `ContinuousSplitPlannerImpl` class in this PR.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -200,12 +202,13 @@ ScanContext copyWithAppendsBetween(long newStartSnapshotId, long newEndSnapshotI
         .project(schema)
         .filters(filters)
         .limit(limit)
-        .exposeLocality(exposeLocality)
         .includeColumnStats(includeColumnStats)
+        .exposeLocality(exposeLocality)
+        .planParallelism(planParallelism)
```
yittg, line 207: Thanks for catching this, i think we can move this to another separated hot fix commit?

yittg: i created a hot fix for it #4341

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -23,6 +23,7 @@
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import org.apache.flink.annotation.Internal;
```
yittg, line 26: Is it proper to use the flink internal annotation?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,184 @@
...
+
+  private void validate() {
+    Preconditions.checkArgument(scanContext.snapshotId() == null,
+        "Can't set snapshotId in ScanContext for continuous enumerator");
+    Preconditions.checkArgument(scanContext.asOfTimestamp() == null,
+        "Can't set asOfTimestamp in ScanContext for continuous enumerator");
+    Preconditions.checkArgument(scanContext.startSnapshotId() == null,
+        "Can't set startSnapshotId in ScanContext for continuous enumerator");
+    Preconditions.checkArgument(scanContext.endSnapshotId() == null,
+        "Can't set endSnapshotId in ScanContext for continuous enumerator");
```
yittg, line 66: Does that mean scanContext#monitorInterval should be null, because it is configured in IcebergEnumeratorConfig?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/IcebergEnumeratorConfig.java
```diff
@@ -0,0 +1,160 @@
...
+     */
+    public Builder splitDiscoveryInterval(Duration interval) {
+      this.splitDiscoveryInterval = interval;
+      return this;
+    }
+
+    public Builder startingStrategy(StartingStrategy strategy) {
+      this.startingStrategy = strategy;
+      return this;
+    }
```
yittg, line 125: Would it be better to use named strategy method accepting required options, like

```suggestion
  public Builder startingWithSpecificSnapshot(long startId) {
      this.startingStrategy = StartingStrategy.SPECIFIC_START_SNAPSHOT_ID;
      this.startSnapshotId = startId;
      return this;
  }
```

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,184 @@
...
+
+    IcebergEnumeratorPosition position = IcebergEnumeratorPosition.builder()
+        .endSnapshotId(startSnapshotEntry.snapshotId())
+        .endSnapshotTimestampMs(startSnapshotEntry.timestampMillis())
+        .build();
+    return new ContinuousEnumerationResult(splits, position);
+  }
+
+  @VisibleForTesting
+  static HistoryEntry getStartSnapshot(Table table, IcebergEnumeratorConfig enumeratorConfig) {
```
yittg, line 131: it's a little confused whether the start snapshot is counted or not? it is exclusive based on `TableScan#appendsBetween`. However, it does not match the intuition for strategy `EARLIEST_SNAPSHOT`, even `SPECIFIC_START_SNAPSHOT_ID` and `SPECIFIC_START_SNAPSHOT_TIMESTAMP `

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,184 @@
...
+        startSnapshotEntry.snapshotId(), config.startingStrategy());
+    List<IcebergSourceSplit> splits;
+    if (config.startingStrategy() ==
+        IcebergEnumeratorConfig.StartingStrategy.TABLE_SCAN_THEN_INCREMENTAL) {
+      // do a full table scan first
+      splits = FlinkSplitPlanner.planIcebergSourceSplits(table, scanContext, workerPool);
+      LOG.info("Discovered {} splits from initial full table scan with snapshotId {}",
+          splits.size(), startSnapshotEntry);
+    } else {
+      splits = Collections.emptyList();
```
yittg, line 120: We should also return splits by scanContext#copyWithSnapshotId(startSnapshotEntry.snapshotId()) for strategy `EARLIEST_SNAPSHOT`

yittg (commented):
Thanks @stevenzwu, i have some question about this change, hoping you can help

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: yes, because Iceberg doesn't have such annotation. I have been using it extensively in the FLIP-27 source, which is approved by Ryan Blue.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: that is a good question. I am also pondering maybe we should merge the `IcebergEnumeratorConfig` with `ScanContext`. As you pointed out, there are some overlapped configs, like `monitorInterval`, `startSnapshotId`. Please let me know your preference. Will also get some input from Ryan.

Right now, I am leaning toward merging them.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/IcebergEnumeratorConfig.java
stevenzwu: agree, I like your suggestion.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: start snapshot is exclusive. I am thinking that we can document the behavior better for `StartingStrategy` enum class.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: merged the `IcebergEnumeratorConfig` into `ScanContext`.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
yittg: Thanks, got it.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
yittg: IIUC, these strategies mean semantic with AFTER?

EARLIEST_SNAPSHOT actually means AFTER_ EARLIEST_SNAPSHOT;
SPECIFIC_START_SNAPSHOT_ID actually means START_AFTER_SNAPSHOT_ID;
SPECIFIC_START_SNAPSHOT_TIMESTAMP actually means START_AFTER_SNAPSHOT_TIMESTAMP;

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
yittg: i still have a litter concern, let's take an example:

1. Table T initially has some snapshots S1, S2 with data files F1-3;
2. Then table T append F4, F5 with S3, and F6 with S4;
3. Now S1, S2 are removed based on retention policy, and table T has snapshots S3, S4 with data files F1-6.

Now S3 is the earliest snapshot, right?
IIUC, the current implementation will scan only F6 and the following append files.

I think we should scan F4-6 for `EARLIEST_SNAPSHOT` strategy to correspond to `LATEST_SNAPSHOT` strategy.

WDYT, @stevenzwu

stevenzwu: yes, that is the current behavior for exclusive starting snapshot. Because EARLIEST_SNAPSHOT is S3, the incremental scan of (S3, S4], we will only scan F6 only (committed by S4).

if it is LATEST_SNAPSHOT (S4), there won't be any files from F1-F6. Only the files are committed by the next snapshot (S5) will be discovered.

@yittg you like to make the starting snapshot inclusive (vs exclusive that is currently implemented), right? It is doable, although a little more complexity. If that is what other people favor, we can make the change.

yittg: @stevenzwu i don't mean to make the starting inclusive. It should be consistent with the other places in iceberg.

We can compare the similar strategy in kafka.
I mean , it should include the earliest by `earliest`, and it should include the next coming by `latest`.

stevenzwu: Discussed with @yittg offline
* agree that both earliest and latest should have the same inclusive/exclusive behavior for easier . In the above example from Yi, here are the expectations if we use the exclusive behavior. earliest (starting S3) should discovery only F6 from S4 via the increment scan of (S3, S4]. latest (starting S4) should discover new files committed by future S5 via incremental scan of (s4, s5].
* because the current FlinkSource (`StreamingMonitorFunction`) implements exclusive behavior for `startSnapshotId` (probably matching the Iceberg `appendBetween` behavior), I am favoring keeping the exclusive behavior.

Will get some inputs from other people on the inclusive vs exclusive behavior regarding start snapshot.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -114,77 +128,100 @@ private ScanContext(boolean caseSensitive, Long snapshotId, Long startSnapshotId
...
     this.planParallelism = planParallelism;
+
+    validate();
+  }
+
+  private void validate() {
+    if (isStreaming) {
+      if (startingStrategy == StreamingStartingStrategy.SPECIFIC_START_SNAPSHOT_ID) {
+        Preconditions.checkArgument(startSnapshotId != null,
+            "startSnapshotId cannot be null for SPECIFIC_START_SNAPSHOT_ID starting strategy");
```
rdblue, line 139: We probably want to remove the specific variable names from the error messages. And we also phrase error messages more directly: `(Problem): (context)`

This should probably be: `Invalid starting snapshot for SPECIFIC_START_SNAPSHOT_ID strategy: null`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Start incremental mode from a specific startSnapshotId
+   */
+  SPECIFIC_START_SNAPSHOT_ID,
+
+  /**
+   * Start incremental mode from a specific startTimestamp.
+   * Starting snapshot has a timestamp lower than or equal to the specified timestamp.
+   */
+  SPECIFIC_START_SNAPSHOT_TIMESTAMP
```
rdblue, line 52: What happens if table history doesn't go back that far? It should probably fail because the user's request can't be satisfied.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Start incremental mode from the earliest snapshot
+   */
+  EARLIEST_SNAPSHOT,
+
+  /**
+   * Start incremental mode from a specific startSnapshotId
+   */
+  SPECIFIC_START_SNAPSHOT_ID,
```
rdblue, line 46: Inclusive of the changes in this snapshot?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+  private final Table table;
+  private final ScanContext scanContext;
+  private final ExecutorService workerPool;
+
+  public ContinuousSplitPlannerImpl(Table table, ScanContext scanContext) {
+    this.table = table;
+    this.scanContext = scanContext;
+    // Within a JVM, table name should be unique across sources.
+    // Hence it is used as worker pool thread name prefix.
+    this.workerPool = ThreadPools.newWorkerPool("iceberg-worker-pool-" + table.name(), scanContext.planParallelism());
```
rdblue, line 52: Isn't it possible to process the table twice in the same JVM? I agree operator ID would be better.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  @Override
+  public ContinuousEnumerationResult planSplits(IcebergEnumeratorPosition lastPosition) {
+    table.refresh();
+    if (lastPosition != null) {
+      return discoverDeltaSplits(lastPosition);
```
rdblue, line 59: I would avoid using "Delta" because that has multiple meanings. Incremental is better.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+
+    IcebergEnumeratorPosition position = IcebergEnumeratorPosition.builder()
+        .endSnapshotId(startSnapshotEntry.snapshotId())
+        .endSnapshotTimestampMs(startSnapshotEntry.timestampMillis())
+        .build();
+    return new ContinuousEnumerationResult(splits, position);
+  }
+
+  @VisibleForTesting
+  static HistoryEntry getStartSnapshot(Table table, ScanContext scanContext) {
```
rdblue, line 117: Here you probably want to use `SnapshotUtil`, which handles most of these cases using the current table state's ancestors.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+            .filter(entry -> entry.snapshotId() == scanContext.startSnapshotId())
+            .findFirst();
+        if (matchedEntry.isPresent()) {
+          startEntry = matchedEntry.get();
+        } else {
+          throw new IllegalArgumentException(
+              "Snapshot id not found in history: {}" + scanContext.startSnapshotId());
+        }
+        break;
+      case SPECIFIC_START_SNAPSHOT_TIMESTAMP:
```
rdblue, line 139: Probably use https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L112

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+  @VisibleForTesting
+  static HistoryEntry getStartSnapshot(Table table, ScanContext scanContext) {
+    List<HistoryEntry> historyEntries = table.history();
+    HistoryEntry startEntry;
+    switch (scanContext.startingStrategy()) {
+      case TABLE_SCAN_THEN_INCREMENTAL:
+      case LATEST_SNAPSHOT:
+        startEntry = historyEntries.get(historyEntries.size() - 1);
+        break;
+      case EARLIEST_SNAPSHOT:
```
rdblue, line 125: Use https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L89

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  @VisibleForTesting
+  static HistoryEntry getStartSnapshot(Table table, ScanContext scanContext) {
+    List<HistoryEntry> historyEntries = table.history();
+    HistoryEntry startEntry;
+    switch (scanContext.startingStrategy()) {
+      case TABLE_SCAN_THEN_INCREMENTAL:
+      case LATEST_SNAPSHOT:
+        startEntry = historyEntries.get(historyEntries.size() - 1);
```
rdblue, line 123: `table.currentSnapshot()`?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+        break;
+      case EARLIEST_SNAPSHOT:
+        startEntry = historyEntries.get(0);
+        break;
+      case SPECIFIC_START_SNAPSHOT_ID:
+        Optional<HistoryEntry> matchedEntry = historyEntries.stream()
```
rdblue, line 129: `table.snapshot(snapshotId)`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+      return new ContinuousEnumerationResult(splits, position);
+    }
+  }
+
+  /**
+   * Discovery initial set of splits based on {@link StreamingStartingStrategy}
+   */
+  private ContinuousEnumerationResult discoverInitialSplits() {
+    HistoryEntry startSnapshotEntry = getStartSnapshot(table, scanContext);
+    LOG.info("get startSnapshotId {} based on starting strategy {}",
```
rdblue, line 96: Clean up log message?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+          .build();
+      LOG.info("Discovered {} splits from incremental scan: {}", splits.size(), position);
+      return new ContinuousEnumerationResult(splits, position);
+    }
+  }
+
+  /**
+   * Discovery initial set of splits based on {@link StreamingStartingStrategy}
+   */
+  private ContinuousEnumerationResult discoverInitialSplits() {
```
rdblue, line 94: Can you add to the javadoc that this only returns splits if there is a starting table scan?

rdblue: And that this is going to return a starting position of the starting snapshot, which is exclusive.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
stevenzwu: mentioned the `exclusive` behave at the class-level Javadoc

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: make the thread pool name a constructor argument. 

Because FLIP-27 source interface doesn't expose the operator ID, for now `tableName-UUID` is used to guarantee uniqueness. comments will be updated.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: will update

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will update

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will switch

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will switch

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will switch

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will switch

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will switch

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will update

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will update

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -114,77 +128,100 @@ private ScanContext(boolean caseSensitive, Long snapshotId, Long startSnapshotId
...
+      }
+      if (startingStrategy == StreamingStartingStrategy.SPECIFIC_START_SNAPSHOT_TIMESTAMP) {
+        Preconditions.checkArgument(startSnapshotTimestamp != null,
+            "startSnapshotTimestamp cannot be null for SPECIFIC_START_SNAPSHOT_TIMESTAMP starting strategy");
```
kbendick, line 143: Should we validate that  `startSnapshotId` isn't supplied if the user has set the `startingStrategy` to be by timestamp.

As well as making sure there's not a provided `startSnapshotTimestamp`  when using snapshot id starting strategy?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
stevenzwu: switched to `SnapshotUtil#snapshotIdAsOfTime`, which handles this scenario internally (throwing exception)

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+
+  private final Table table;
+  private final ScanContext scanContext;
+  private final ExecutorService workerPool;
+
+  public ContinuousSplitPlannerImpl(Table table, ScanContext scanContext, String threadPoolName) {
+    this.table = table;
+    this.scanContext = scanContext;
+    this.workerPool = ThreadPools.newWorkerPool(
+        "iceberg-enumerator-pool-" + threadPoolName, scanContext.planParallelism());
```
kbendick, line 50: Nit: Would it be more informative to mention `split-planner-pool-**` or something that correlates more to the class name?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+
+    IcebergEnumeratorPosition position = IcebergEnumeratorPosition.builder()
+        .endSnapshotId(startSnapshot.snapshotId())
+        .endSnapshotTimestampMs(startSnapshot.timestampMillis())
+        .build();
+    return new ContinuousEnumerationResult(splits, position);
+  }
+
+  @VisibleForTesting
+  static Snapshot getStartSnapshot(Table table, ScanContext scanContext) {
```
yittg, line 123: How would you deal with an empty table, i.e. no snapshots are produced?

Guess this method should return an `Optional<Snapshot>` ?

yittg, line 123: How would you deal with an empty table, i.e. no snapshots are produced?

Guess this method should return an `Optional<Snapshot>` ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Discover incremental changes between @{code lastPosition} and current table snapshot
+   */
+  private ContinuousEnumerationResult discoverIncrementalSplits(IcebergEnumeratorPosition lastPosition) {
+    // incremental discovery mode
+    Snapshot currentSnapshot = table.currentSnapshot();
+    if (currentSnapshot.snapshotId() == lastPosition.endSnapshotId()) {
+      LOG.info("Current table snapshot is already enumerated: {}", currentSnapshot.snapshotId());
+      return new ContinuousEnumerationResult(Collections.emptyList(), lastPosition);
+    } else {
```
yittg, line 72: nit: this else is not necessary.

stevenzwu: I actually like the symmetry of if-else. I searched some existing Iceberg code. I saw many usage of if-else block for throwing exception. so will stay put

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: good point. it is a corner case. will switch to `Optional` and also add a unit test

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: sure. will add them

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: if we use the current naming convention in `FlinkInputFormat`, we can probably change it to `iceberg-plan-worker-pool-<threadPoolName>`.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,160 @@
...
+   * for the next incremental split discovery with exclusive behavior. Meaning files committed
+   * by the snapshot from the position in {@code ContinuousEnumerationResult} won't be included
+   * in the next incremental scan.
+   */
+  private ContinuousEnumerationResult discoverInitialSplits() {
+    Optional<Snapshot> startSnapshotOptional = getStartSnapshot(table, scanContext);
+    if (!startSnapshotOptional.isPresent()) {
+      // set IcebergEnumeratorPosition to null in this case
+      return new ContinuousEnumerationResult(Collections.emptyList(), null);
+    }
```
yittg, line 106: guess we can not return null position here, then the next plan will still invoke `#discoverInitialSplits`. If some snapshots produced during this period， they will be seen as start snapshots and excluded.

yittg: Empty table is invalid if the strategies must correspond to a certain snapshot, but it's weird to say empty table can not be applied with these two strategies.

There may be more than one snapshots can be committed before next round of discover, so it's not only the first snapshot.

How about returning a position with end snapshot as -1, which i think is idiomatic in iceberg.

stevenzwu: agree it is not just the first snapshot. I was using it just to simplify the discussion. it doesn't really matter how many snapshots are committed before the first discovery. semantics are the same.

I was saying `earliest` or `latest` snapshot can be interpreted as the earliest or latest valid snapshot of the table. Let's say when the Flink job started, it is an empty table. initial scan return null. 
* case 1: there are 3 snapshots (S1, S2, S3) committed. Now Flink source  ran initial scan again. For earliest strategy, S1 will be used as starting snapshot. For latest strategy, S3 will be used as starting snapshot. 
* case 2: there is 1 snapshot committed. Now both earliest and latest strategy will discover S1 as starting snapshot

regarding -1 snapshot id, is it a reserved special id? I wouldn't find the usage for snapshot id. Also, `IncrementalDataTableScan` seems to validate the start snapshot id. This may not work.
```
    Preconditions.checkArgument(
        snapshotIdsRange.contains(newFromSnapshotId),
        "from snapshot id %s not in existing snapshot ids range (%s, %s]",
        newFromSnapshotId, context().fromSnapshotId(), newToSnapshotId);
```

@yittg I definitely see your perspective, which is a valid interpretation of earliest and latest for empty table. In your interpretation, earliest and latest will behave the same (with some imaginary snapshot as starting point). all future snapshot should be discovered. To implement this model, we need to change `appendsBetween` implementation to support a special starting snapshot id, which is invalid and yet indicates a special starting point.

To me, both interpretations are reasonable. I also think for the practical purpose, how to interpret earliest and latest strategy for an empty table is not that important for streaming job. Most of the streaming job's bootstrap point is not really important, as it is arbitrary anyway.

yittg: sounds reasonable to me, thanks.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+public enum StreamingStartingStrategy {
+  /**
+   * First do a regular table scan. then switch to incremental mode.
+   */
+  TABLE_SCAN_THEN_INCREMENTAL,
+
+  /**
+   * Start incremental mode from the latest snapshot
+   */
+  LATEST_SNAPSHOT,
```
rdblue, line 35: `INCREMENTAL_AFTER_LATEST`?
`INCREMENTAL_AFTER_EARLIEST`?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Start incremental mode from the earliest snapshot
+   */
+  EARLIEST_SNAPSHOT,
+
+  /**
+   * Start incremental mode from a specific startSnapshotId
+   */
+  SPECIFIC_START_SNAPSHOT_ID,
```
rdblue, line 45: `INCREMENTAL_AFTER_SPECIFIC_SNAPSHOT_ID`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: @yittg after discussing with @rdblue, we agreed on the inclusive starting behavior. With inclusive behavior, empty or non-empty tables should behave the same. For an empty table, earliest/latest will include all snapshots afterwards.

yittg: Thanks for letting me know, good job.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -114,77 +128,104 @@ private ScanContext(boolean caseSensitive, Long snapshotId, Long startSnapshotId
     this.includeColumnStats = includeColumnStats;
     this.exposeLocality = exposeLocality;
     this.planParallelism = planParallelism;
+
+    validate();
+  }
+
+  private void validate() {
+    if (isStreaming) {
```
openinx, line 136: Will it be allowed to set any `startingStrategy` value if it's in batch mode ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -114,77 +128,104 @@ private ScanContext(boolean caseSensitive, Long snapshotId, Long startSnapshotId
...
+      if (startingStrategy == StreamingStartingStrategy.SPECIFIC_START_SNAPSHOT_ID) {
```
openinx, line 137: Should we also consider the validation for other `strategy`s ? 

* `TABLE_SCAN_THEN_INCREMENTAL`

Will the `startSnapshotId`  or `startSnapshotTimestamp` specify the first snapshot to do the regular table scan ?  Is there any meaning to set the `startSnapshotTimestamp` ? 

* LATEST_SNAPSHOT

As we always read data from the latest available snapshot, so is there any meaning to set the `startSnapshotId`,  `startSnapshotTimestamp`, `endSnapshotId`, `asOfTimestamp` ? 

* EARLIEST_SNAPSHOT

Ditto


Besides,  do we expect the streaming strategy to be a bounded stream in which case we may need to set the `endSnapshotId`, or a unbounded stream ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
openinx: Agreed.   It's more clear to add the `INCREMENTAL` prefix to align with the `TABLE_SCAN_THEN_INCREMENTAL`.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg.flink.source;
+
+/**
+ * Starting strategy for streaming execution. Note that starting snapshot is exclusive in the incremental
+ * split discovery mode. Files appended by the starting snapshot aren't included in split discovery.
+ */
+public enum StreamingStartingStrategy {
```
YesOrNo828, line 26: Is this just a strategy for unbounded reading ? And will there be a strategy for bounded reading, e.g.  `BatchStartingStragegy` ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -75,7 +85,9 @@
   private final boolean caseSensitive;
   private final boolean exposeLocality;
   private final Long snapshotId;
+  private final StreamingStartingStrategy startingStrategy;
   private final Long startSnapshotId;
+  private final Long startSnapshotTimestamp;
```
YesOrNo828, line 90: Should we add the other Long parameter : `endSnapshotTimestamp`?  just like `startSnapshotId` and `endSnapshotId`.

YesOrNo828 (commented):
hello @stevenzwu , I left some questions about this pr, looking forward to your reply.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: @YesOrNo828 I thought about the `endSnapshotTimestamp`. This is only useful if we want to set an end time in the future. If it is an end time in the past, we don't really need the streaming read with a stop/end timestamp. We can just use a regular filter with batch read mode.

We can add the `endSnapshotTimestamp` feature when there is real need for it. I just didn't include it in the current PR.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: Will the `startSnapshotId` or `startSnapshotTimestamp`, there is no regular table scan. Say they map to some snapshot id X. With the latest agreement on inclusive behavior for the starting snapshot, we will start the incremental scan btw (X's parent snapshot id, Y]. Previous/current implementation has the start snapshot as exclusive, which is the same behavior as the current source impl.

Only `TABLE_SCAN_THEN_INCREMENTAL` does a regular table scan and then start the incremental discovery process.

right now, I didn't add extensive/strict validations. We can't do some of the validations. E.g., `endSnapshotId` will be added in the copied `ScanContext`. Validation on `endSnapshotId` will fail. I guess we can fix the copy method and unset the streaming related fields like `isStreaming`.
```
ScanContext incrementalScan = scanContext
          .copyWithAppendsBetween(lastPosition.endSnapshotId(), currentSnapshot.snapshotId());
```

Regarding the end condition, it might be useful for a future `endSnapshotTimestamp`. A past `endSnapshotTimestamp` probably doesn't make sense, since we can just do a regular batch scan with filters. 

A `endSnapshotId` in the future also doesn't make sense, since we won't know it when the streaming read starts. A  `endSnapshotId` in the past should be just a regular batch scan and hence is not applicable to streaming mode.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
stevenzwu: It is just a streaming strategy. Streaming source can also be bounded, which is also the case for Flink kafka source. As mentioned in another comment, it is reasonable to have the Iceberg source stop at a future timestamp. It is not implemented in this PR. We can discuss the feature in the future if people want it useful.

For batch scan, we won't need any starting strategy. We would just need to specify the filters.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: In a sync meeting with @rdblue , we agreed on two things
1. The starting snapshot should be inclusive. This is consistent with the behavior as Kafka source. It is also consistent with the empty table behavior below. I will rename the enum to `INCREMENTAL_FROM_...`, which hopefully clarify the inclusive behavior.
2. Earliest and latest start snapshot should work with empty table (all future snapshots should be included). This is original point from @yittg in this thread.

I looked into the IncrementalDataTableScan and SnapshotUtil. To achieve the above behavior, we can make fromSnapshotId nullable (not primitive) in TableScan . 
```java
TableScan appendsBetween(long->Long fromSnapshotId, long toSnapshotId);
```

1. We can just pass in starting snapshot's parent snapshotId as the fromSnapshotId  (could be null)
2. The earliest or latest start strategy for an empty table would just set the fromSnapshotId to null.

This method from SnapshotUtil already handles null fromSnapshotId
```java
private static List<Snapshot> snapshotsWithin(Table table, Long fromSnapshotId, long toSnapshotId)
```

**It is source compatible (as boxing happens automatically), but not binary compatible**. Is that acceptable?

cc @rdblue @openinx

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: we can add the validation that `startingStrategy` is null for batch mode.

will also need to change the default value to null for the config option.
```
  private static final ConfigOption<StreamingStartingStrategy> STARTING_STRATEGY =
      ConfigOptions.key("starting-strategy").enumType(StreamingStartingStrategy.class)
          .defaultValue(StreamingStartingStrategy.LATEST_SNAPSHOT);
```

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: created a separate PR to support null fromSnapshotId for incremental scan.
https://github.com/apache/iceberg/pull/4529

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/IcebergEnumeratorPosition.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.base.Objects;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+class IcebergEnumeratorPosition {
+  private final Long snapshotId;
+  private final Long snapshotTimestampMs;
+
+  IcebergEnumeratorPosition(Long snapshotId, Long snapshotTimestampMs) {
```
rdblue, line 30: Position.empty() -> always passes null
Position.create(long snapshotId, long snapshotTimestampMs) -> always passes non-null

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/IcebergEnumeratorPosition.java
```diff
@@ -0,0 +1,74 @@
...
+        "snapshot id and timestamp should be either both null or both not null");
+    this.snapshotId = snapshotId;
+    this.snapshotTimestampMs = snapshotTimestampMs;
+  }
+
+  public Long snapshotId() {
+    return snapshotId;
+  }
+
+  public Long snapshotTimestampMs() {
```
rdblue, line 43: If this is just for logging, we may want to use `toString` so we don't expose this. I'm concerned about people trying to use it later to filter.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlanner.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.flink.annotation.Internal;
+
+/**
+ * This interface is introduced so that we can plug in
+ * different split planner for unit test
+ */
+@Internal
+public interface ContinuousSplitPlanner {
+
+  ContinuousEnumerationResult planSplits(IcebergEnumeratorPosition lastPosition);
```
rdblue, line 31: Can you document the guarantees or contract for this method?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,172 @@
...
+   */
+  @VisibleForTesting
+  static Optional<Snapshot> getStartSnapshot(Table table, ScanContext scanContext) {
+    switch (scanContext.startingStrategy()) {
+      case TABLE_SCAN_THEN_INCREMENTAL:
+      case INCREMENTAL_FROM_LATEST_SNAPSHOT:
+        return Optional.ofNullable(table.currentSnapshot());
+      case INCREMENTAL_FROM_EARLIEST_SNAPSHOT:
+        return Optional.ofNullable(SnapshotUtil.oldestAncestor(table));
+      case INCREMENTAL_FROM_SNAPSHOT_ID:
```
rdblue, line 150: This appears to be always inclusive?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        if (matchedSnapshotById != null) {
+          return Optional.of(matchedSnapshotById);
+        } else {
+          throw new IllegalArgumentException(
+              "Start snapshot id not found in history: " + scanContext.startSnapshotId());
+        }
+      case INCREMENTAL_FROM_SNAPSHOT_TIMESTAMP:
+        // snapshotIdAsOfTime returns valid snapshot id only.
+        // it throws IllegalArgumentException when a matching snapshot not found.
+        long snapshotIdAsOfTime = SnapshotUtil.snapshotIdAsOfTime(table, scanContext.startSnapshotTimestamp());
```
rdblue, line 161: This is going to return the snapshot that was current as of some time:
* t0: s1 is committed
* t1
* t2: s2 is committed

`snapshotAsOfTime(t1)` will return s1. And it looks like this is going to return s1. Is that going to be exclusive or inclusive? I think it should be exclusive unless t0 == t1.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,56 @@
...
+ * Starting strategy for streaming execution.
+ */
+public enum StreamingStartingStrategy {
+  /**
+   * First do a regular table scan. Then switch to incremental mode after the table scan snapshot.
+   */
+  TABLE_SCAN_THEN_INCREMENTAL,
+
+  /**
+   * Start incremental mode from the latest snapshot inclusive
```
rdblue, line 32: I think it makes sense for this to be inclusive because you want to have some changes.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/StreamingStartingStrategy.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package org.apache.iceberg.flink.source;
+
+/**
+ * Starting strategy for streaming execution.
+ */
+public enum StreamingStartingStrategy {
+  /**
+   * First do a regular table scan. Then switch to incremental mode after the table scan snapshot.
```
rdblue, line 27: This will consume changes exclusive because the current snapshot is the one that gets scanned.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
rdblue: I think to make this consistent, you should do the check here. If this snapshot's timestamp matches, return it for inclusive start. If not, then return the next snapshot for the inclusive start.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,172 @@
...
+      toPosition = new IcebergEnumeratorPosition(parentSnapshotId, parentSnapshotTimestampMs);
+      LOG.info("Start incremental scan with start snapshot (inclusive): id = {}, timestamp = {}",
+          startSnapshot.snapshotId(), startSnapshot.timestampMillis());
+    }
+
+    return new ContinuousEnumerationResult(splits, null, toPosition);
+  }
+
+  /**
+   * Optional is used because table may be empty and has no snapshot
```
rdblue, line 140: Probably note here that if the mode is NOT `TABLE_SCAN_THEN_INCREMENTAL` then the snapshot should be consumed inclusive.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,172 @@
...
+          splits.size(), startSnapshot.snapshotId());
+      toPosition = new IcebergEnumeratorPosition(startSnapshot.snapshotId(), startSnapshot.timestampMillis());
+    } else {
+      splits = Collections.emptyList();
+      // Use parentId to achieve the inclusive behavior.
+      // Note that it is fine if parentId is null.
+      Long parentSnapshotId = startSnapshot.parentId();
+      Long parentSnapshotTimestampMs = null;
+      if (parentSnapshotId != null) {
+        parentSnapshotTimestampMs = table.snapshot(parentSnapshotId).timestampMillis();
```
rdblue, line 128: For oldest ancestor, the parent snapshot won't be found and this will result in `NullPointerException`.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    LOG.info("Get starting snapshot id {} based on strategy {}",
+        startSnapshot.snapshotId(), scanContext.startingStrategy());
+    List<IcebergSourceSplit> splits;
+    IcebergEnumeratorPosition toPosition;
+    if (scanContext.startingStrategy() == StreamingStartingStrategy.TABLE_SCAN_THEN_INCREMENTAL) {
+      // do a full table scan first
+      splits = FlinkSplitPlanner.planIcebergSourceSplits(table, scanContext, workerPool);
+      LOG.info("Discovered {} splits from initial full table scan with snapshot Id {}",
+          splits.size(), startSnapshot.snapshotId());
+      toPosition = new IcebergEnumeratorPosition(startSnapshot.snapshotId(), startSnapshot.timestampMillis());
```
rdblue, line 120: Creates a starting position that is exclusive.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * Discover incremental changes between @{code lastPosition} and current table snapshot
+   */
+  private ContinuousEnumerationResult discoverIncrementalSplits(IcebergEnumeratorPosition lastPosition) {
+    Snapshot currentSnapshot = table.currentSnapshot();
+    if (currentSnapshot == null) {
+      // empty table
+      IcebergEnumeratorPosition newPosition = new IcebergEnumeratorPosition(null, null);
```
rdblue, line 71: In this case, we know that `lastPosition` must be empty, right? We can probably just return `lastPosition` as the new position. And there's a problem if `lastPosition` is non-empty.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/IcebergEnumeratorPosition.java
stevenzwu: realized this getter is also used in unit test. Since this class is non public, I will mark the two getters methods as package default scope too.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: Except for `TABLE_SCAN_THEN_INCREMENTAL`, all other starting strategy is inclusive from the starting snapshot.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
stevenzwu: will add a `Preconditions` check and reuse the `lastPosition`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,184 @@
...
+      case INCREMENTAL_FROM_LATEST_SNAPSHOT:
+        return Optional.ofNullable(table.currentSnapshot());
+      case INCREMENTAL_FROM_EARLIEST_SNAPSHOT:
+        return Optional.ofNullable(SnapshotUtil.oldestAncestor(table));
+      case INCREMENTAL_FROM_SNAPSHOT_ID:
+        Snapshot matchedSnapshotById = table.snapshot(scanContext.startSnapshotId());
+        if (matchedSnapshotById != null) {
+          return Optional.of(matchedSnapshotById);
+        } else {
+          throw new IllegalArgumentException(
```
rdblue, line 163: We usually prefer `Preconditions.checkArgument` instead of an extra `if` statement.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,184 @@
...
+  private ContinuousEnumerationResult discoverIncrementalSplits(IcebergEnumeratorPosition lastPosition) {
+    Snapshot currentSnapshot = table.currentSnapshot();
+    if (currentSnapshot == null) {
+      // empty table
+      Preconditions.checkArgument(lastPosition.snapshotId() == null,
+          "Invalid last enumerated position for an empty table: not null");
+      LOG.info("Skip incremental scan because table is empty");
+      return new ContinuousEnumerationResult(Collections.emptyList(), lastPosition, lastPosition);
+    } else {
+      if (lastPosition.snapshotId() != null && currentSnapshot.snapshotId() == lastPosition.snapshotId()) {
```
rdblue, line 77: Nit: this could be an `else if`

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImplStartStrategy.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.iceberg.flink.source.StreamingStartingStrategy;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.TestRule;
+
+public class TestContinuousSplitPlannerImplStartStrategy {
```
rdblue, line 42: Looks good!

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,461 @@
...
+
+    ContinuousEnumerationResult result = splitPlanner.planSplits(lastPosition);
+    Assert.assertEquals(lastPosition.snapshotId(), result.fromPosition().snapshotId());
+    Assert.assertEquals(lastPosition.snapshotTimestampMs(), result.fromPosition().snapshotTimestampMs());
+    Assert.assertEquals(snapshot.snapshotId(), result.toPosition().snapshotId().longValue());
+    Assert.assertEquals(snapshot.timestampMillis(), result.toPosition().snapshotTimestampMs().longValue());
+    Assert.assertEquals(1, result.splits().size());
+    IcebergSourceSplit split = result.splits().iterator().next();
+    Assert.assertEquals(1, split.task().files().size());
+    Assert.assertEquals(dataFile.path().toString(), split.task().files().iterator().next().file().path().toString());
```
rdblue, line 106: Minor: you can use `Iterables.getOnlyElement` to avoid calling `iterator().next()` without validation.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/enumerator/ContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,184 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(ContinuousSplitPlannerImpl.class);
+
+  private final Table table;
+  private final ScanContext scanContext;
+  private final ExecutorService workerPool;
+
+  public ContinuousSplitPlannerImpl(Table table, ScanContext scanContext, String threadPoolName) {
+    this.table = table;
+    this.scanContext = scanContext;
+    this.workerPool = ThreadPools.newWorkerPool(
```
rdblue, line 51: Minor: for testing, could we set this to null so that we don't spawn a worker pool?

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImpl.java
```diff
@@ -0,0 +1,461 @@
...
+    ScanContext scanContext = ScanContext.builder()
+        .startingStrategy(StreamingStartingStrategy.TABLE_SCAN_THEN_INCREMENTAL)
+        .build();
+    ContinuousSplitPlannerImpl splitPlanner = new ContinuousSplitPlannerImpl(
+        tableResource.table(), scanContext, testName.getMethodName());
+
+    ContinuousEnumerationResult emptyTableInitialDiscoveryResult = splitPlanner.planSplits(null);
+    Assert.assertTrue(emptyTableInitialDiscoveryResult.splits().isEmpty());
+    Assert.assertNull(emptyTableInitialDiscoveryResult.fromPosition());
+    Assert.assertNull(emptyTableInitialDiscoveryResult.toPosition().snapshotId());
```
rdblue, line 121: We may want to add an `isEmpty` method to check this rather than checking `snapshotId()` is null.

rdblue (approved):

rdblue (approved):

rdblue:
Thanks, @stevenzwu!

PR #4330, opened 2022-03-15
Nessie: Fix Use reference with hash command

ajantha-bhat:
Use the new configuration `NessieConfigConstants.CONF_NESSIE_REF_HASH` to set/reset the initial hash for the use reference command.

NessieCatalog#loadReference() is called from 
a) NessieCatalog#Initialize()
b) newTableOps()

hash from `newTableOps()` flow should only be used for current operation. When Hash is not specified in `newTableOps()`, it should fallback to Initial Hash (set from Use reference command)

This is part of a fix for https://github.com/projectnessie/nessie/issues/3439

### DIFF
--- a/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
+++ b/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
@@ -87,6 +87,7 @@ public class NessieCatalog extends BaseMetastoreCatalog implements AutoCloseable
   private String name;
   private FileIO fileIO;
   private Map<String, String> catalogOptions;
+  private String initialHash;
 
   public NessieCatalog() {
   }
@@ -127,7 +128,8 @@ public class NessieCatalog extends BaseMetastoreCatalog implements AutoCloseable
       throw new IllegalStateException("Parameter 'warehouse' not set, Nessie can't store data.");
     }
     final String requestedRef = options.get(removePrefix.apply(NessieConfigConstants.CONF_NESSIE_REF));
-    this.reference = loadReference(requestedRef, null);
+    initialHash = options.get(removePrefix.apply(NessieConfigConstants.CONF_NESSIE_REF_HASH));
+    this.reference = loadReference(requestedRef, initialHash);
   }
 
   private static NessieClientBuilder<?> createNessieClientBuilder(String customBuilder) {
@@ -161,7 +163,9 @@ public class NessieCatalog extends BaseMetastoreCatalog implements AutoCloseable
         "timestamp is not supported)");
     UpdateableReference newReference = this.reference;
     if (tr.getReference() != null) {
-      newReference = loadReference(tr.getReference(), tr.getHash());
+      // use the initialHash when hash is not specified.
+      String hash = tr.getHash() == null ? initialHash : tr.getHash();
+      newReference = loadReference(tr.getReference(), hash);
     }
     return new NessieTableOperations(
         ContentKey.of(org.projectnessie.model.Namespace.of(tableIdentifier.namespace().levels()), tr.getName()),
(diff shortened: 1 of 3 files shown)


ajantha-bhat:
cc: @nastra , @snazy

nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
```diff
@@ -101,14 +101,27 @@ public void testCatalogOnReference() {
...
     testCatalogEquality(refCatalog, testCatalog, true, true, () -> { });
 
     // catalog created with hash points to same catalog as above
-    NessieCatalog refHashCatalog = initCatalog("main");
+    NessieCatalog refHashCatalog = initCatalog("main", null);
     testCatalogEquality(refHashCatalog, catalog, true, true, () -> { });
   }
 
+  @Test
+  public void testCatalogInitialHash() {
```
ajantha-bhat, line 113: End to end testcase with the USE reference command will be added in Nessie Integration test.

nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
```diff
@@ -117,17 +117,21 @@ public void beforeEach(@NessieUri URI nessieUri) throws IOException {
     }
 
     hadoopConfig = new Configuration();
-    catalog = initCatalog(branch);
+    catalog = initCatalog(branch, null);
   }
 
-  NessieCatalog initCatalog(String ref) {
+  NessieCatalog initCatalog(String ref, String hash) {
```
nastra, line 123: maybe just add another `initCatalog` method that accepts a `hash` parameter, so that all the places in the tests don't have to pass `null` all the time

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -355,6 +357,10 @@ public String currentHash() {
     return reference.getHash();
   }
 
+  public String initialHash() {
```
nastra, line 360: does this really have to be public? This should rather have a `@VisibleForTesting` annotation and remove the `public`

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -379,14 +385,19 @@ private UpdateableReference loadReference(String requestedRef, String hash) {
     try {
       Reference ref = requestedRef == null ? api.getDefaultBranch()
           : api.getReference().refName(requestedRef).get();
-      if (hash != null) {
-        if (ref instanceof Branch) {
-          ref = Branch.of(ref.getName(), hash);
-        } else {
-          ref = Tag.of(ref.getName(), hash);
-        }
+      String hashOnRef;
```
nastra, line 388: not sure these changes are even needed. wouldn't the code behave the same way without doing any changes and just passing `initialHash` to the method?

nastra (changes requested):

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -379,12 +388,10 @@ private UpdateableReference loadReference(String requestedRef, String hash) {
     try {
       Reference ref = requestedRef == null ? api.getDefaultBranch()
           : api.getReference().refName(requestedRef).get();
-      if (hash != null) {
```
ajantha-bhat, line 382: I think this hash check is needed, I am re-evaluating.

ajantha-bhat:
Use reference at hash works with this change, but the subsequent insert is failing with this error `You can only mutate tables when using a branch without a hash or timestamp`

So, tried calling use reference without hash, but that is setting hash to null. (if I don't set hash to null, it uses previous hash which is wrong). But if I set to null refresh is not called.

So, not sure how to handle this, probably a change at nessie side should send the latest hash instead of null for use reference without hash.

ajantha-bhat:
https://github.com/apache/iceberg/pull/4700 gonna fix this.

PR #4333, opened 2022-03-15
Core: Make CatalogTests run with JUnit5

nastra:
We are planning to add a `TestNessieCatalog` test that extends the
existing `CatalogTests` class. However, Nessie tests require
JUnit5 functionality to set up the Nessie server, therefore we need to
convert these catalog tests to JUnit5.

### DIFF
--- a/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java
+++ b/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java
@@ -59,16 +59,15 @@ import org.apache.iceberg.transforms.Transform;
 import org.apache.iceberg.transforms.Transforms;
 import org.apache.iceberg.types.Types;
 import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.TemporaryFolder;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 import static org.apache.iceberg.NullOrder.NULLS_FIRST;
 import static org.apache.iceberg.SortDirection.ASC;
 import static org.apache.iceberg.types.Types.NestedField.required;
 
-public class TestJdbcCatalog extends CatalogTests<JdbcCatalog>  {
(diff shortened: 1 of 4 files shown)


rdblue:
I'm not clear on why this is needed. Can't Nessie run these tests with the vintage engine?

nastra:
> I'm not clear on why this is needed. Can't Nessie run these tests with the vintage engine?

Unfortunately not. Nessie tests use Junit 5 extensions to bring up the nessie server for tests and so the tests themselves need to be actual Junit 5 tests

kbendick:
If we add this to `core`, what are the implications for existing JUnit 4 based tests? As well as for things like overall runtime or any other concerns in operating in a mixed mode? Possibly classpath issues?

Trying to better understand the implications of adding it.

nastra:
> If we add this to `core`, what are the implications for existing JUnit 4 based tests? As well as for things like overall runtime or any other concerns in operating in a mixed mode? Possibly classpath issues?
> 
> Trying to better understand the implications of adding it.

Existing Junit4 tests will work as usual, so there are no implications and no classpath issues. The only implication is that subclasses of `CatalogTests` will have to be Junit5 tests (which is what's being addressed here as well)

snazy (approved):
LGTM

core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java
```diff
@@ -82,9 +81,8 @@
   private static JdbcCatalog catalog;
   private static String warehouseLocation;
 
-  @Rule
-  public TemporaryFolder temp = new TemporaryFolder();
-  File tableDir = null;
+  @TempDir
+  java.nio.file.Path tableDir;
```
snazy, line 85: Import?

nastra: required as otherwise it would clash with the `org.apache.hadoop.fs.Path` import

snazy: Argh, okay - i looked through the imports but haven't seen that one.

snazy (approved):

rdblue (approved):

PR #4334, opened 2022-03-15
AWS: Add support for s3 access points

singhpk234:
This tries to add support of s3 access points 
ref : https://aws.amazon.com/s3/features/access-points/

```shell
    spark-shell \
        --conf spark.sql.catalog.test=org.apache.iceberg.spark.SparkCatalog \
        --conf spark.sql.catalog.test.warehouse=s3://my-bucket \
        --conf spark.sql.catalog.test.catalog-impl=org.apache.iceberg.aws.glue.GlueCatalog \
        --conf spark.sql.catalog.test.io-impl=org.apache.iceberg.aws.s3.S3FileIO \
        --conf spark.sql.catalog.test.s3.access-points.my-bucket=arn:aws:s3::123456789012:accesspoint:mfzwi23gnjvgw.mrap
```

The s3.access-points config provides a map of bucket to endpoint mapping for paths stored in Iceberg, so that the S3FileIO can use the access point to get the bucket instead when the mapping is configured.

---

### DIFF
--- a/aws/src/integration/java/org/apache/iceberg/aws/AwsIntegTestUtil.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/AwsIntegTestUtil.java
@@ -21,8 +21,11 @@ package org.apache.iceberg.aws;
 
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.glue.GlueClient;
 import software.amazon.awssdk.services.glue.model.DeleteDatabaseRequest;
 import software.amazon.awssdk.services.s3.S3Client;
@@ -31,6 +34,9 @@ import software.amazon.awssdk.services.s3.model.DeleteObjectsRequest;
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Response;
 import software.amazon.awssdk.services.s3.model.ObjectIdentifier;
+import software.amazon.awssdk.services.s3control.S3ControlClient;
+import software.amazon.awssdk.services.s3control.model.CreateAccessPointRequest;
+import software.amazon.awssdk.services.s3control.model.DeleteAccessPointRequest;
 
 public class AwsIntegTestUtil {
 
@@ -40,11 +46,38 @@ public class AwsIntegTestUtil {
   }
 
   /**
+   * Get the environment variable AWS_REGION to use for testing
+   * @return region
+   */
+  public static String testRegion() {
+    return System.getenv("AWS_REGION");
+  }
+
+  /**
+   * Get the environment variable AWS_CROSS_REGION to use for testing
+   * @return region
+   */
+  public static String testCrossRegion() {
+    String crossRegion = System.getenv("AWS_CROSS_REGION");
+    Preconditions.checkArgument(!testRegion().equals(crossRegion), "AWS_REGION should not be equal to " +
+        "AWS_CROSS_REGION");
+    return crossRegion;
+  }
+
+  /**
    * Set the environment variable AWS_TEST_BUCKET for a default bucket to use for testing
    * @return bucket name
    */
   public static String testBucketName() {
     return System.getenv("AWS_TEST_BUCKET");
+  }
+
+  /**
+   * Set the environment variable AWS_TEST_CROSS_REGION_BUCKET for a default bucket to use for testing
+   * @return bucket name
+   */
+  public static String testCrossRegionBucketName() {
+    return System.getenv("AWS_TEST_CROSS_REGION_BUCKET");
   }
 
   /**
@@ -79,6 +112,40 @@ public class AwsIntegTestUtil {
       } catch (Exception e) {
         LOG.error("Cannot delete namespace {}", namespace, e);
       }
+    }
+  }
+
+  public static S3ControlClient createS3ControlClient(String region) {
+    return S3ControlClient.builder()
+        .httpClientBuilder(UrlConnectionHttpClient.builder())
+        .region(Region.of(region))
+        .build();
+  }
+
+  public static void createAccessPoint(S3ControlClient s3ControlClient, String accessPointName, String bucketName) {
+    try {
+      s3ControlClient.createAccessPoint(CreateAccessPointRequest
+          .builder()
+          .name(accessPointName)
+          .bucket(bucketName)
+          .accountId(testAccountId())
+          .build()
+      );
+    } catch (Exception e) {
+      LOG.error("Cannot create access point {}", accessPointName, e);
+    }
+  }
+
+  public static void deleteAccessPoint(S3ControlClient s3ControlClient, String accessPointName) {
+    try {
+      s3ControlClient.deleteAccessPoint(DeleteAccessPointRequest
+          .builder()
+          .name(accessPointName)
+          .accountId(testAccountId())
+          .build()
+      );
+    } catch (Exception e) {
+      LOG.error("Cannot delete access point {}", accessPointName, e);
     }
   }
 }
--- a/aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
@@ -38,12 +38,16 @@ import org.apache.iceberg.aws.AwsProperties;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
 import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.regions.PartitionMetadata;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.kms.KmsClient;
 import software.amazon.awssdk.services.kms.model.ListAliasesRequest;
 import software.amazon.awssdk.services.kms.model.ListAliasesResponse;
@@ -57,14 +61,21 @@ import software.amazon.awssdk.services.s3.model.ObjectCannedACL;
 import software.amazon.awssdk.services.s3.model.Permission;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
 import software.amazon.awssdk.services.s3.model.ServerSideEncryption;
+import software.amazon.awssdk.services.s3control.S3ControlClient;
+import software.amazon.awssdk.utils.ImmutableMap;
 import software.amazon.awssdk.utils.IoUtils;
 
 public class TestS3FileIOIntegration {
 
   private static AwsClientFactory clientFactory;
   private static S3Client s3;
+  private static S3ControlClient s3Control;
+  private static S3ControlClient crossRegionS3Control;
   private static KmsClient kms;
   private static String bucketName;
+  private static String crossRegionBucketName;
+  private static String accessPointName;
+  private static String crossRegionAccessPointName;
   private static String prefix;
   private static byte[] contentBytes;
   private static String content;
@@ -78,17 +89,27 @@ public class TestS3FileIOIntegration {
     clientFactory = AwsClientFactories.defaultFactory();
     s3 = clientFactory.s3();
     kms = clientFactory.kms();
+    s3Control = AwsIntegTestUtil.createS3ControlClient(AwsIntegTestUtil.testRegion());
+    crossRegionS3Control = AwsIntegTestUtil.createS3ControlClient(AwsIntegTestUtil.testCrossRegion());
     bucketName = AwsIntegTestUtil.testBucketName();
+    crossRegionBucketName = AwsIntegTestUtil.testCrossRegionBucketName();
+    accessPointName = UUID.randomUUID().toString();
+    crossRegionAccessPointName = UUID.randomUUID().toString();
     prefix = UUID.randomUUID().toString();
     contentBytes = new byte[1024 * 1024 * 10];
     deletionBatchSize = 3;
     content = new String(contentBytes, StandardCharsets.UTF_8);
     kmsKeyArn = kms.createKey().keyMetadata().arn();
+
+    AwsIntegTestUtil.createAccessPoint(s3Control, accessPointName, bucketName);
+    AwsIntegTestUtil.createAccessPoint(crossRegionS3Control, crossRegionAccessPointName, crossRegionBucketName);
   }
 
   @AfterClass
   public static void afterClass() {
     AwsIntegTestUtil.cleanS3Bucket(s3, bucketName, prefix);
+    AwsIntegTestUtil.deleteAccessPoint(s3Control, accessPointName);
+    AwsIntegTestUtil.deleteAccessPoint(crossRegionS3Control, crossRegionAccessPointName);
     kms.scheduleKeyDeletion(ScheduleKeyDeletionRequest.builder().keyId(kmsKeyArn).pendingWindowInDays(7).build());
   }
 
@@ -96,6 +117,11 @@ public class TestS3FileIOIntegration {
   public void before() {
     objectKey = String.format("%s/%s", prefix, UUID.randomUUID().toString());
     objectUri = String.format("s3://%s/%s", bucketName, objectKey);
+  }
+
+  @BeforeEach
+  public void beforeEach() {
+    clientFactory.initialize(Maps.newHashMap());
   }
 
   @Test
@@ -107,10 +133,65 @@ public class TestS3FileIOIntegration {
   }
 
   @Test
+  public void testNewInputStreamWithAccessPoint() throws Exception {
+    s3.putObject(PutObjectRequest.builder().bucket(bucketName).key(objectKey).build(),
+        RequestBody.fromBytes(contentBytes));
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3);
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testRegion(), accessPointName)));
+    validateRead(s3FileIO);
+  }
+
+  @Test
+  public void testNewInputStreamWithCrossRegionAccessPoint() throws Exception {
+    clientFactory.initialize(ImmutableMap.of(AwsProperties.S3_USE_ARN_REGION_ENABLED, "true"));
+    S3Client s3Client = clientFactory.s3();
+    s3Client.putObject(PutObjectRequest.builder().bucket(bucketName).key(objectKey).build(),
+        RequestBody.fromBytes(contentBytes));
+    // make a copy in cross-region bucket
+    s3Client.putObject(PutObjectRequest.builder()
+        .bucket(testAccessPointARN(AwsIntegTestUtil.testCrossRegion(), crossRegionAccessPointName))
+            .key(objectKey).build(),
+        RequestBody.fromBytes(contentBytes));
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3);
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testCrossRegion(), crossRegionAccessPointName)));
+    validateRead(s3FileIO);
+  }
+
+  @Test
   public void testNewOutputStream() throws Exception {
     S3FileIO s3FileIO = new S3FileIO(clientFactory::s3);
     write(s3FileIO);
     InputStream stream = s3.getObject(GetObjectRequest.builder().bucket(bucketName).key(objectKey).build());
+    String result = IoUtils.toUtf8String(stream);
+    stream.close();
+    Assert.assertEquals(content, result);
+  }
+
+  @Test
+  public void testNewOutputStreamWithAccessPoint() throws Exception {
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3);
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testRegion(), accessPointName)));
+    write(s3FileIO);
+    InputStream stream = s3.getObject(GetObjectRequest.builder().bucket(bucketName).key(objectKey).build());
+    String result = IoUtils.toUtf8String(stream);
+    stream.close();
+    Assert.assertEquals(content, result);
+  }
+
+  @Test
+  public void testNewOutputStreamWithCrossRegionAccessPoint() throws Exception {
+    clientFactory.initialize(ImmutableMap.of(AwsProperties.S3_USE_ARN_REGION_ENABLED, "true"));
+    S3Client s3Client = clientFactory.s3();
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3);
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testCrossRegion(), crossRegionAccessPointName)));
+    write(s3FileIO);
+    InputStream stream = s3Client.getObject(GetObjectRequest.builder()
+        .bucket(testAccessPointARN(AwsIntegTestUtil.testCrossRegion(), crossRegionAccessPointName))
+        .key(objectKey).build());
     String result = IoUtils.toUtf8String(stream);
     stream.close();
     Assert.assertEquals(content, result);
@@ -219,6 +300,23 @@ public class TestS3FileIOIntegration {
   }
 
   @Test
+  public void testDeleteFilesMultipleBatchesWithAccessPoints() throws Exception {
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3, getDeletionTestProperties());
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testRegion(), accessPointName)));
+    testDeleteFiles(deletionBatchSize * 2, s3FileIO);
+  }
+
+  @Test
+  public void testDeleteFilesMultipleBatchesWithCrossRegionAccessPoints() throws Exception {
+    clientFactory.initialize(ImmutableMap.of(AwsProperties.S3_USE_ARN_REGION_ENABLED, "true"));
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3, getDeletionTestProperties());
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testCrossRegion(), crossRegionAccessPointName)));
+    testDeleteFiles(deletionBatchSize * 2, s3FileIO);
+  }
+
+  @Test
   public void testDeleteFilesLessThanBatchSize() throws Exception {
     S3FileIO s3FileIO = new S3FileIO(clientFactory::s3, getDeletionTestProperties());
     testDeleteFiles(deletionBatchSize - 1, s3FileIO);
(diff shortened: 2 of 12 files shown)


aws/src/main/java/org/apache/iceberg/aws/s3/S3URI.java
```diff
@@ -68,6 +69,34 @@
...
+  /**
+   * Creates a new S3URI in the form of scheme://bucket/key?query#fragment
+   * <p>
+   * The URI supports any valid URI schemes to be backwards compatible with s3a and s3n,
+   * and also allows users to use S3FileIO with other S3-compatible object storage services like GCS.
+   *
+   * @param location fully qualified URI
+   * @param bucketToAccessPointMapping contains mapping of bucket to access point
+   */
+  S3URI(String location, Map<String, String> bucketToAccessPointMapping) {
```
jackye1995, line 81: I think the original constructor can just call `S3URI(location, ImmutableMap.of())`

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3URI.java
```diff
@@ -90,4 +92,17 @@ public void testValidSchemes() {
       assertEquals("path/to/file", uri.key());
     }
   }
+
+  @Test
+  public void testS3URIWithBucketToAccessPointMapping() {
```
jackye1995, line 97: might be a bit hard, but is it possible to run with integration test by creating an access point on the fly?

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
```diff
@@ -43,6 +44,11 @@ public static S3OutputFile fromLocation(String location, S3Client client, AwsPro
     return new S3OutputFile(client, new S3URI(location), awsProperties, metrics, writeTags);
   }
 
+  public static S3OutputFile fromLocation(String location, S3Client client, AwsProperties awsProperties,
```
jackye1995, line 47: the `fromLocation` at L42 was just added, not released yet, we can just update that method to also include this new field instead of creating a new static helper.

jackye1995:
FYI @flyrain @anuragmantri, this is based on our discussion in https://apache-iceberg.slack.com/archives/C025PH0G1D4/p1645066803099319

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3URI.java
singhpk234: I am thinking of finding a way to do that, As per my understanding this would require creating a `S3ControlClient` as per https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3control/S3ControlClient.html#createAccessPoint-- which available at sdk version >= `2.17.149` tried searching it my intellij maybe at present version (`2.17.131`) it's available with diff name if present. S3MockRule at present doesn't have a way to test this atleast from what I can say which is what is used in the `Tests3OutputStream`

once the above is sorted I believe it rest of stuff should be easy to follow.

anuragmantri (approved):
Changes look good to me, pending integration tests requested by @jackye1995.

Thanks for adding this @singhpk234

aws/src/main/java/org/apache/iceberg/aws/s3/S3URI.java
```diff
@@ -50,6 +52,19 @@
...
     Preconditions.checkNotNull(location, "Location cannot be null.");
```
anuragmantri, line 68: Nit: I'm not sure if `propertiesWithPrefix()` already ensures this but could you add a null check for the `bucketToAccessPointMapping` as well here. This cannot be null.

singhpk234: As per my understanding it's ensured it will return empty map rather than null in this scenario, (tested it locally as well). 
https://github.com/apache/iceberg/blob/e9a287b7f68d8e7138c47dcb686d3709bdde402e/core/src/main/java/org/apache/iceberg/util/PropertyUtil.java#L85-L97


I agree the on the adding the check, as this being a public method, and can be accessed in future with different code-paths, adding this check.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -225,6 +225,16 @@
    */
   public static final String S3_WRITE_TAGS_PREFIX = "s3.write.tags.";
 
+  /**
+   * Used by {@link S3FileIO} for access point. To set, we can pass a catalog property.
+   * <p>
+   * For more details, see https://aws.amazon.com/s3/features/access-points/
+   * <p>
+   * Example in Spark:
+   *    --conf spark.sql.catalog.test.s3.access-points.my-bucket=arn:aws:s3::123456789012:accesspoint:mfzwi23gnjvgw.mrap
```
rdblue, line 234: I would probably not include a Spark-specific example. Instead, just show configuration from the catalog level, like `s3.access-points.my-bucket=...`.

aws/src/main/java/org/apache/iceberg/aws/s3/S3URI.java
```diff
@@ -50,7 +52,21 @@
    * @param location fully qualified URI
    */
   S3URI(String location) {
+    this(location, ImmutableMap.of());
+  }
+
+  /**
+   * Creates a new S3URI in the form of scheme://bucket/key?query#fragment
```
jackye1995, line 59: the javadoc here does not really explain the additional mapping information. I think  the javadoc can be removed if they are duplicated, or we explain it as "Create a new S3URI with additional information of bucket to access point mapping"

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -68,6 +68,7 @@
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
   private Set<Tag> writeTags = Sets.newHashSet();
+  private Map<String, String> bucketToAccessPointMapping = Maps.newHashMap();
```
jackye1995, line 71: nit: variable here can just be named `accessPoints`

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -101,17 +102,17 @@ public S3FileIO(SerializableSupplier<S3Client> s3, AwsProperties awsProperties)
...
   @Override
   public InputFile newInputFile(String path) {
-    return S3InputFile.fromLocation(path, client(), awsProperties, metrics);
+    return S3InputFile.fromLocation(path, client(), awsProperties, metrics, bucketToAccessPointMapping);
   }
 
   @Override
   public OutputFile newOutputFile(String path) {
-    return S3OutputFile.fromLocation(path, client(), awsProperties, metrics, writeTags);
+    return S3OutputFile.fromLocation(path, client(), awsProperties, metrics, writeTags, bucketToAccessPointMapping);
```
jackye1995, line 110: I am thinking about this a bit, it does not really scale if we keep adding new arguments to the constructor, we will basically have 1 new factory method per release. Traditionally we have been adding all the properties directly in `awsProperties`, and we did not really evaluate that option when adding `writeTags`. We should think about this a bit, I feel it is cleaner if we move all these new fields back to `AwsProperties`, it was designed to be the extension point for new configs without changing constructors in different places? Any thoughts?

aws/src/integration/java/org/apache/iceberg/aws/AwsIntegTestUtil.java
```diff
@@ -47,6 +58,23 @@ public static String testBucketName() {
...
+    return System.getenv("AWS_TEST_ACCESS_POINT");
+  }
+
+  /**
+   * Get AccessPointARN for a default bucket to use for testing
+   * @return access point arn
+   */
+  public static String testAccessPointARN() {
+    // format : arn:aws:s3:region:account-id:accesspoint/resource
+    return String.format("arn:aws:s3:%s:%s:accesspoint/%s", testRegion(), testAccountId(), testAccessPointName());
```
jackye1995, line 75: partition is not always `aws`, should use `PartitionMetadata.of(region)` to get the partition value

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
```diff
@@ -78,17 +83,23 @@ public static void beforeClass() {
...
     bucketName = AwsIntegTestUtil.testBucketName();
+    accountId = AwsIntegTestUtil.testAccountId();
+    accessPointName = AwsIntegTestUtil.testAccessPointName();
     prefix = UUID.randomUUID().toString();
     contentBytes = new byte[1024 * 1024 * 10];
     deletionBatchSize = 3;
     content = new String(contentBytes, StandardCharsets.UTF_8);
     kmsKeyArn = kms.createKey().keyMetadata().arn();
+
+    AwsIntegTestUtil.createAccessPoint(s3Control, accessPointName, bucketName, accountId);
```
jackye1995, line 96: if we are creating the access points on the fly, why do we still need those environment variables? Can't we just define the `accessPointName` in tests?

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
```diff
@@ -116,6 +137,18 @@ public void testNewOutputStream() throws Exception {
     Assert.assertEquals(content, result);
   }
 
+  @Test
+  public void testNewOutputStreamWithAccessPoint() throws Exception {
```
jackye1995, line 141: Probably no need to test those, but I think this should work as well for multi-region access point and bucket alias, right?

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
rajarshisarkar: I agree with this, this should avoid changes to the constructor (or) introducing a new one. I will publish the PR for moving the new fields to `AwsProperties`. Thanks, for bringing this up.

singhpk234: +1, Agree with this, changing constructors definitely doesn't seems to be scalable.

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
singhpk234: yes , we can and these arguments are indeed not required in the present set-up can be directly accessed in the Util class itself.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
rajarshisarkar: I have published the PR: https://github.com/apache/iceberg/pull/4350

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
singhpk234: yes this should work out of box, tested mrap locally it works

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
singhpk234: I agree with this,  but by naming it accessPoints at first glance could mean just a collection of access points where as here what we actually have  is a mapping from s3 bucket to accessPoint, if this rational makes sense I would like to keep it as its.. WDYT ?

jackye1995: yes good point, let's keep as is

aws/src/integration/java/org/apache/iceberg/aws/AwsIntegTestUtil.java
```diff
@@ -47,6 +60,24 @@ public static String testBucketName() {
     return System.getenv("AWS_TEST_BUCKET");
   }
 
+  /**
+   * Get the environment variable AWS_TEST_ACCESS_POINT for a default bucket to use for testing
+   * @return bucket name
+   */
+  public static String testAccessPointName() {
```
jackye1995, line 67: I think this part is not fully updated since the last comment? We don't need to get this info from environment variable, because we can just define any name we like in the test.

aws/src/integration/java/org/apache/iceberg/aws/AwsIntegTestUtil.java
```diff
@@ -47,6 +60,24 @@ public static String testBucketName() {
...
+    return System.getenv("AWS_TEST_ACCESS_POINT");
+  }
+
+  /**
+   * Get AccessPointARN for a default bucket to use for testing
+   * @return access point arn
+   */
+  public static String testAccessPointARN() {
```
jackye1995, line 75: if we remove `testAccessPointName`, it means this can also be removed, or moved into `TestS3FileIOIntegration`

aws/src/main/java/org/apache/iceberg/aws/s3/S3URI.java
```diff
@@ -50,7 +52,24 @@
...
+    this(location, ImmutableMap.of());
+  }
+
+  /**
+   * Creates a new S3URI in the form of scheme://(bucket|accessPoint)/key?query#fragment with additional information
+   * on accessPoints.
+   * <p>
+   * The URI supports any valid URI schemes to be backwards compatible with s3a and s3n,
+   * and also allows users to use S3FileIO with other S3-compatible object storage services like GCS.
+   * If the accessPoints contains a mapping of the given bucket used in location then the corresponding accessPoint
```
jackye1995, line 64: nit: this sentence is a bit convoluted, what about "if the bucket of the location has an access point in the mapping, the access point is used to perform all the S3 operations"

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
```diff
@@ -268,4 +306,11 @@ private void validateRead(S3FileIO s3FileIO) throws Exception {
     stream.close();
     Assert.assertEquals(content, result);
   }
+
+  private String testAccessPointARN() {
```
jackye1995, line 310: I think we also need to test cross-region endpoint, because I see this S3 client config:

```
    /**
     * Returns whether the client is allowed to make cross-region calls when an S3 Access Point ARN has a different
     * region to the one configured on the client.
     * <p>
     * @return True if a different region in the ARN can be used.
     */
    public boolean useArnRegionEnabled() {
        return useArnRegionEnabled.value();
    }
```

Maybe we should turn this on when access point is configured. Could you check?

singhpk234: - by configured you mean there is a s3bucketToAccessPoint is not empty or introducing a flag to enable this feature.
- From the looks of this, it looks like it would be required trying out a quick test case for the same.

singhpk234: I tried this and it failed with "java.lang.IllegalArgumentException: The region field of the ARN being passed as a bucket parameter to an S3 operation does not match the region the client was configured with. To enable this behavior and prevent this exception set 'useArnRegionEnabled' to true in the configuration when building the S3 client. Provided region: 'us-east-2'; client region: 'us-east-1'."

Looks like we need to set this

jackye1995: cool, so to summarize, same-single-region and multi-region don't need this flag, and cross-single-region need this flag, is that right?

If so, I think we should introduce this flag, because there is some performance implications in enabling this feature as the signing region is changed. We should not enable this feature by default for people who use MRAP.

singhpk234: +1, same-single-region / multi-region doesn't need this (verified this locally)

should we make this flag explicit to access points`s3-acess-point.cross-single-region.enabled` or it can be as general as `s3-cross-single-region.enabled` based on that we can set this in our s3 client the required conf.

jackye1995: For boolean config, prefer to use `-enabled` suffix as convention. All S3 features should be under `s3.` config namespace.

I think we can just follow the S3 variable name, `s3.arn-region-enabled` since S3 already gave this feature this name. We can directly link to this feature's github link in AWS SDK v2 repo as a reference. What do you think?

singhpk234: This makes sense, having such a name and linking it to the direct AWS SDK v2 repo as a reference adds more value. Let me make the changes.

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactory.java
```diff
@@ -56,6 +57,12 @@
    */
   DynamoDbClient dynamo();
 
+  /**
+   * Create a Amazon S3Control client
+   * @return s3Control client
+   */
+  S3ControlClient s3Control();
```
jackye1995, line 64: I think we no longer need this in the client factory? It is for test only and we already have a method there to create s3 control client.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -170,6 +171,17 @@
    */
   public static final String S3FILEIO_SESSION_TOKEN = "s3.session-token";
 
+  /**
+   * Enable to make S3FileIO, to make cross-region call to the region specified in the ARN.
+   * <p>
+   * When set, the client can make a cross-region call to the region specified in the ARN,
+   * otherwise an exception will be thrown if cross-region call is tried.
+   *
+   * For more details see : https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java#L186
```
jackye1995, line 180: line number might change, linking to the github code is probably a bad idea, sorry for the bad suggestion. We can link to this: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3Configuration.html#useArnRegionEnabled--

build.gradle
```diff
@@ -313,6 +313,7 @@ project(':iceberg-aws') {
 
     compileOnly 'software.amazon.awssdk:url-connection-client'
     compileOnly 'software.amazon.awssdk:s3'
+    compileOnly 'software.amazon.awssdk:s3control'
```
jackye1995, line 316: this can be marked as a test only dependency

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
```diff
@@ -106,6 +132,33 @@ public void testNewInputStream() throws Exception {
...
+    s3.putObject(PutObjectRequest.builder().bucket(bucketName).key(objectKey).build(),
+        RequestBody.fromBytes(contentBytes));
+    S3FileIO s3FileIO = new S3FileIO(clientFactory::s3);
+    s3FileIO.initialize(ImmutableMap.of(AwsProperties.S3_ACCESS_POINTS_PREFIX + bucketName,
+        testAccessPointARN(AwsIntegTestUtil.testRegion(), accessPointName)));
+    validateRead(s3FileIO);
+  }
+
+  @Test
+  public void testNewInputStreamWithCrossRegionAccessPoint() throws Exception {
```
jackye1995, line 146: Just a side note that in https://github.com/aws/aws-sdk-java-v2/commit/51632cb58491e13f1d88e636353d4ce87b34e81d, a new config `multiRegionEnabled` is added, which defaults to true and controls cross-region access when using MRAP. The default works for us, so we probably don't need to add it for now.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -170,6 +171,17 @@
    */
   public static final String S3FILEIO_SESSION_TOKEN = "s3.session-token";
 
+  /**
+   * Enable to make S3FileIO, to make cross-region call to the region specified in the ARN.
```
jackye1995, line 175: in the ARN of an access point?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -170,6 +171,17 @@
...
+   * <p>
+   * When set, the client can make a cross-region call to the region specified in the ARN,
+   * otherwise an exception will be thrown if cross-region call is tried.
+   *
+   * For more details see : https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java#L186
+   */
+  public static final String S3_ARN_REGION_ENABLED = "s3.arn-region-enabled";
```
jackye1995, line 182: Sorry I just notice the name in S3Configuration is `useArnRegionEnabled`, I think we should map the name exactly to the s3 to call it `s3.use-arn-region-enabled`

jackye1995 (approved):
looks good to me!

aws/src/main/java/org/apache/iceberg/aws/AssumeRoleAwsClientFactory.java
```diff
@@ -46,12 +46,14 @@
   private int timeout;
   private String region;
   private String s3Endpoint;
+  private Boolean s3UseArnRegionEnabled;
```
rdblue, line 49: Why is this boxed?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -170,6 +171,17 @@
    */
   public static final String S3FILEIO_SESSION_TOKEN = "s3.session-token";
 
+  /**
+   * Enable to make S3FileIO, to make cross-region call to the region specified in the ARN of an access point.
+   * <p>
+   * When set, the client can make a cross-region call to the region specified in the ARN of an access point,
+   * otherwise an exception will be thrown if cross-region call is tried.
+   *
+   * For more details see : https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3Configuration.html#useArnRegionEnabled--
```
rdblue, line 180: Typo: `:` should not have a space before it.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -170,6 +171,17 @@
    */
   public static final String S3FILEIO_SESSION_TOKEN = "s3.session-token";
 
+  /**
+   * Enable to make S3FileIO, to make cross-region call to the region specified in the ARN of an access point.
+   * <p>
+   * When set, the client can make a cross-region call to the region specified in the ARN of an access point,
+   * otherwise an exception will be thrown if cross-region call is tried.
```
rdblue, line 178: I think it would be more clear to say "By default, attempting to use an access point in a different region will throw an exception. When enabled, this property is allows using access points in other regions".

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -254,6 +266,15 @@
    */
   public static final String S3_WRITE_TAGS_PREFIX = "s3.write.tags.";
 
+  /**
+   * Used by {@link S3FileIO} for access point. To set, we can pass a catalog property.
```
rdblue, line 270: I'm not sure what this is saying. Do you mean "prefix used for access point configuration"?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -254,6 +266,15 @@
...
+   * <p>
+   * For more details, see https://aws.amazon.com/s3/features/access-points/
+   * <p>
+   * Example : s3.access-points.my-bucket=arn:aws:s3:region:account-id:accesspoint/resource
```
rdblue, line 274: I don't think that this context needs to be repeated in multiple places.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -290,6 +312,7 @@ public AwsProperties() {
     this.s3fileIoStagingDirectory = System.getProperty("java.io.tmpdir");
     this.isS3ChecksumEnabled = S3_CHECKSUM_ENABLED_DEFAULT;
     this.s3WriteTags = Sets.newHashSet();
+    this.s3BucketToAccessPointMapping = Maps.newHashMap();
```
rdblue, line 315: Why not use an immutable map instead?

aws/src/main/java/org/apache/iceberg/aws/s3/S3URI.java
```diff
@@ -50,7 +52,24 @@
...
+   * and also allows users to use S3FileIO with other S3-compatible object storage services like GCS.
+   * If the bucket in the location has an access point in the mapping, the access point is used to perform all the S3
+   * operations
+   *
+   * @param location fully qualified URI
+   * @param bucketToAccessPointMapping contains mapping of bucket to access point
+   */
+  S3URI(String location, Map<String, String> bucketToAccessPointMapping) {
     Preconditions.checkNotNull(location, "Location cannot be null.");
+    Preconditions.checkNotNull(bucketToAccessPointMapping, "BucketToAccessPointMapping cannot be null.");
```
rdblue, line 72: This error message probably doesn't make sense to people that will see it. We avoid using variable or class names in error messages because users aren't familiar with Iceberg code internal details.

In addition, I don't think there's anything the user would be able to do about this because the mapping is created by AWS code. Instead, I think this should just default the mapping to an empty immutable map.

aws/src/main/java/org/apache/iceberg/aws/s3/S3URI.java
```diff
@@ -50,7 +52,24 @@
...
+  }
+
+  /**
+   * Creates a new S3URI in the form of scheme://(bucket|accessPoint)/key?query#fragment with additional information
+   * on accessPoints.
+   * <p>
+   * The URI supports any valid URI schemes to be backwards compatible with s3a and s3n,
+   * and also allows users to use S3FileIO with other S3-compatible object storage services like GCS.
+   * If the bucket in the location has an access point in the mapping, the access point is used to perform all the S3
+   * operations
```
rdblue, line 65: This looks like class documentation, not constructor documentation.

aws/src/test/java/org/apache/iceberg/aws/TestAwsClientFactories.java
```diff
@@ -34,7 +34,6 @@
 import software.amazon.awssdk.services.kms.KmsClient;
 import software.amazon.awssdk.services.s3.S3Client;
 
-
```
rdblue, line 37: Please revert this change. We avoid changes that are unnecessary and could cause merge failures.

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3URI.java
```diff
@@ -90,4 +92,17 @@ public void testValidSchemes() {
...
+    String p1 = "s3://bucket/path/to/file?query=foo#bar";
+    Map<String, String> bucketToAccessPointMapping = ImmutableMap.of(
+        "bucket", "access-point"
+    );
+    S3URI uri1 = new S3URI(p1, bucketToAccessPointMapping);
+
+    assertEquals("access-point", uri1.bucket());
```
rdblue, line 104: Should there be any validation that the access point is in a valid format? Or is that handled by the SDK and we don't want to duplicate logic?

jackye1995: There could be 2 possiblities:
1. an ARN, something like `arn:aws:s3:region:111122223333:accesspoint/my-access-point`
2. an alias, something like `my-access-point-aqfqprnstn7aefdfbarligizwgyfouse1a-s3alias`

because of 2, the string can really be anything that satisfies the bucket regex, so it's a bit hard to validate on service side, given the fact that we do not really check for normal bucket name correctness either. I think it should be fine to delegate to the S3 service to check for name correctness.

aws/src/test/java/org/apache/iceberg/aws/TestAwsClientFactories.java
singhpk234: Thanks, noted and made the changes.

aws/src/main/java/org/apache/iceberg/aws/AssumeRoleAwsClientFactory.java
singhpk234: Thanks, made the changes.

I overlooked it, I think `primitive` should be sufficient.

jackye1995 (approved):
thanks for addressing all the comments, @rdblue I think all the comments are fixed, let us know if there is any more concern, thanks!

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -170,6 +171,17 @@
    */
   public static final String S3FILEIO_SESSION_TOKEN = "s3.session-token";
 
+  /**
+   * Enable to make S3FileIO, to make cross-region call to the region specified in the ARN of an access point.
+   * <p>
+   * By default, attempting to use an access point in a different region will throw an exception.
+   * When enabled, this property allows using access points in other regions.
+   *
```
jackye1995, line 179: nit: <p> missing

jackye1995, line 179: nit: <p> missing

aws/src/integration/java/org/apache/iceberg/aws/s3/TestS3FileIOIntegration.java
```diff
@@ -268,4 +366,10 @@ private void validateRead(S3FileIO s3FileIO) throws Exception {
     stream.close();
     Assert.assertEquals(content, result);
   }
+
+  private String testAccessPointARN(String region, String accessPoint) {
+    // format : arn:aws:s3:region:account-id:accesspoint/resource
```
jackye1995, line 371: nit: no space before `:`

jackye1995, line 371: nit: no space before `:`

jackye1995 (approved):

jackye1995:
This looks good to me, and I think there has been no comment for a few days, I will merge it to unblock other feature contributions, and for people to test out the access point usages. @rdblue @rajarshisarkar let us know if you have any additional concern, we can address before the upcoming release.

Issue #4139, opened 2022-02-16
Add Delete Files To Files metadata table


szehon-ho:
I am thinking current "files" table suffices to add "delete file" information.

Today, "entries" table already shows some delete file information, though in struct form.
Example entries table "row" for positional delete file:
```
|status|snapshot_id        |sequence_number|data_file
|1     |3491280865879215816|2              |{1, file:/my_table_path/data/p=1/00129-2-47a2d786-6eaa-45c8-a7df-6bf1303553ec-00001.parquet, PARQUET, 0, {1}, 1, 1798, {2147483546 -> 186, 2147483545 -> 46}, {2147483546 -> 1, 2147483545 -> 1}, {2147483546 -> 0, 2147483545 -> 0}, {}, {2147483546 -> file:/my_table_path/data/p=1/00000-0-2043c59b-7a64-4c51-9a99-fb95f83ac076-00001.parquet, 2147483545 -> }, {2147483546 -> file:/my_table/data/p=1/00000-0-2043c59b-7a64-4c51-9a99-fb95f83ac076-00001.parquet, 2147483545 -> }, null, null, null, null}|
```

I think the current "files" table has an adequate schema to capture delete files metadata.

```
|content|   file_path|file_format|spec_id|partition|record_count|file_size_in_bytes|      column_sizes|    value_counts|null_value_counts|nan_value_counts|        lower_bounds|        upper_bounds|key_metadata|split_offsets|equality_ids|sort_order_id|
```

szehon-ho:
@jackye1995 @rdblue @chenjunjiedada  (or sorry if missed someone also involved) wanted to check, has there been some  discussion of the whether this is a good way to go (re-use existing data_files table), versus making a new metadata table for delete files?  Thanks

I am thinking that the obstacle would be the naming, ie 'files' indicates data+delete files , and its current counterpart 'all_data_files' indicates just data files.

rdblue:
The `all_data_files` metadata table was added after we were planning to add delete files, which is why we fixed the name. I'd prefer to add a separate `delete_files` metadata table since they're reasonably separate. I'm not sure whether we should add deletes to `files` or not, but I like the ability to read separate metadata tables.

What do you think, @szehon-ho?

szehon-ho:
Got it, thanks for the background/reply, then it seems all_data_files was designed to only have data files and thus we need all_delete_files.  

[...]

PR #4336, opened 2022-03-15
Core: Include delete files in 'files' and add 'data_files' table

szehon-ho:
As discussed in https://github.com/apache/iceberg/issues/4139, this is the second step after adding 'delete_files' table, to synchronize the similar metadata tables to handle the delete file concept.

In this PR:
- 'files' table now returns data_files and delete_files
- 'data_files' table is newly added to return only data_files.

This might be hard to split into two prs, as it's doing a hot-swap migrating the 'files' table class (called DataFilesTable) to support new table 'data_files'.


core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -0,0 +1,255 @@
...
+
+  private final MetadataTableType type;
+
+  @Parameterized.Parameters(name = "table_type = {0}, format = {1}")
+  public static Object[][] parameters() {
+    return new Object[][] {
+        { MetadataTableType.DATA_FILES, 1 },
+        { MetadataTableType.DATA_FILES, 2 },
+        { MetadataTableType.DELETE_FILES, 2 },
+        { MetadataTableType.FILES, 2 }
```
aokolnychyi, line 50: Would it be possible to also test `FILES` for v1? We may skip some tests by using `assume`.

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -0,0 +1,255 @@
...
+      case DATA_FILES:
+        return new DataFilesTable(table.ops(), table);
+      case DELETE_FILES:
+        return new DeleteFilesTable(table.ops(), table);
+      default:
+        throw new IllegalArgumentException("Unsupported metadata table type:" + type);
+    }
+  }
+
+  private int expectedCount(int partitions) {
```
aokolnychyi, line 105: nit: expected count of what?

aokolnychyi (approved):

szehon-ho:
Merged, thanks @aokolnychyi for the review

PR #4338, opened 2022-03-16
Parquet: parquet_writer.java should close the writeStore when flushRowGroup's finished is false

renshangtao:
When parquetWriter.java call close， it called the writeStore.close 

    if (!closed) {
      this.closed = true;
      flushRowGroup(true);
      writeStore.close();                       ------ here close the writeStore
      if (writer != null) {
        writer.end(metadata);
      }
    }
  }

But in checksize function when called the  flushRowGroup(false); 

private void flushRowGroup(boolean finished) {
    try {
      if (recordCount > 0) {
        ensureWriterInitialized();
        writer.startBlock(recordCount);
        writeStore.flush();
        flushPageStoreToWriter.invoke(writer);
        writer.endBlock();
        if (!finished) {
          startRowGroup();                ---- here new another one, but the old not closed
        }
      }
    } catch (IOException e) {
      throw new UncheckedIOException("Failed to flush row group", e);
    }
  }

It doesn't close the writeStore and then new another one.

So i think when the finish is false we should close the writeStore.


rdblue (approved):

rdblue:
Thanks, @renshangtao!

PR #4340, opened 2022-03-16
Enable parallel_delete in spark action to speed up snapshot expire

0xffmeta:
If there are huge amount of files(~80K) to delete for the snapshot expire, the spark job will be running hours to delete those files even we have already used `max_concurrent_deletes` in the spark procedure.
We can leverage the spark `mapPartitions` mechanism to distribute the tasks to executors for executing, which can fully utilize the executor CPU/IO for large snapshot expiry. See some testing result as below:

#### Using `max_concurrent_deletes` = 50
Driver spec: 45G/8 CPU Core
Executor spec: 6G/1 Core
Executor num: 50
Deleted data file count: ~70K
Total running hour: 3.5H

#### Using `parallel_delete`(use spark `mapPartitions` to distribute the tasks to Spark executors)
Driver spec: 10G/1 CPU Core
Executor spec: 6G/1 Core
Executor num: 50
Deleted data file count: ~110K
Total running hour: 1.5H

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.spark.actions;
 
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
@@ -26,6 +27,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import org.apache.iceberg.HasTableOperations;
+import org.apache.iceberg.SerializableTable;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableOperations;
@@ -70,7 +72,7 @@ public class BaseExpireSnapshotsSparkAction
     extends BaseSparkAction<ExpireSnapshots, ExpireSnapshots.Result> implements ExpireSnapshots {
   private static final Logger LOG = LoggerFactory.getLogger(BaseExpireSnapshotsSparkAction.class);
 
-  public static final String STREAM_RESULTS = "stream-results";
+  public static final String PARALLEL_DELETE = "parallel_delete";
 
   private static final String CONTENT_FILE = "Content File";
   private static final String MANIFEST = "Manifest";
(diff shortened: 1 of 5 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -70,7 +72,7 @@
     extends BaseSparkAction<ExpireSnapshots, ExpireSnapshots.Result> implements ExpireSnapshots {
   private static final Logger LOG = LoggerFactory.getLogger(BaseExpireSnapshotsSparkAction.class);
 
-  public static final String STREAM_RESULTS = "stream-results";
+  public static final String PARALLEL_DELETE = "parallel_delete";
```
rdblue, line 75: This should not remove unrelated options.

Also, options should use `-` to separate words, not `_`.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -212,11 +214,21 @@ private String jobDesc() {
...
+      Table serializableTable = SerializableTable.copyOf(table);
+
+      return expire().javaRDD().mapPartitions((Iterator<Row> iter) ->
+        Arrays.asList(
+            ActionUtils.deleteFiles(iter, (file) -> serializableTable.io().deleteFile(file), null)
+        ).iterator()
+      ).reduce((r1, r2) -> new BaseExpireSnapshotsActionResult(
+          r1.deletedDataFilesCount() + r2.deletedDataFilesCount(),
+          r1.deletedManifestsCount() + r2.deletedManifestsCount(),
+          r1.deletedManifestListsCount() + r2.deletedManifestListsCount()));
```
rdblue, line 229: Using Spark to delete files can easily result in too high of a request rate, which causes S3 to throttle and send 503, "Slow down", responses. When that happens, files are probably leaked as orphans. That's why we have a way to parallelize locally, but not to use Spark.

This action also exposes the `expire()` method that returns a `Dataset` you can use if you want to run the expiration like this in your environment.

rdblue, line 229: Using Spark to delete files can easily result in too high of a request rate, which causes S3 to throttle and send 503, "Slow down", responses. When that happens, files are probably leaked as orphans. That's why we have a way to parallelize locally, but not to use Spark.

This action also exposes the `expire()` method that returns a `Dataset` you can use if you want to run the expiration like this in your environment.

rdblue:
I'm not sure that this is a good idea. We purposely don't run deletes in parallel in Spark.

@jackye1995, what do you think?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
0xffmeta: @rdblue Thanks for your comment. In our last few runs, we didn't see too many S3 503 errors, maybe the requests were distributed into different prefixes(with partition dt/hh). I think by using the local parallelize mechanism, it might require more resource on driver side, while all the executors seem to be all in idle. E.g. I need to request more CPUs for the driver to be able to view the Spark UI(50 concurrent delete), which is not cost efficient considering it also needs longer time to run.

For the request with 503, we probably can collect them back into the driver for anther retry?

PR #4341, opened 2022-03-16
Flink: fix missing copy in ScanContext

yittg:
Fix missing copy metioned in #4329 by @stevenzwu .

### DIFF
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
@@ -200,8 +200,9 @@ class ScanContext implements Serializable {
         .project(schema)
         .filters(filters)
         .limit(limit)
-        .exposeLocality(exposeLocality)
         .includeColumnStats(includeColumnStats)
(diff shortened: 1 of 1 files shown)


flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -200,8 +200,9 @@ ScanContext copyWithAppendsBetween(long newStartSnapshotId, long newEndSnapshotI
         .project(schema)
         .filters(filters)
         .limit(limit)
-        .exposeLocality(exposeLocality)
         .includeColumnStats(includeColumnStats)
+        .exposeLocality(exposeLocality)
```
rdblue, line 204: Why did `exposeLocality` move? To be consistent?

rdblue:
Thanks, @yittg!

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: yes. this is to be consistent with the ordering of how variables are defined

PR #4110, opened 2022-02-13
Spark: Exclude netty from spark runtime modules

pan3793:
The unexpected `io.netty.util` comes from `netty-common` exists in spark runtime jars, we should relocate it to avoid classes conflict.

image

PR #4344, opened 2022-03-16
Revert "Spark: Exclude Netty from Spark runtime modules (#4110)"

ajantha-bhat:
This reverts commit eace97307cecf1e6a8cdd96cf78a5d0847ea1cb7.

while using the latest build from master branch with `iceberg-spark-runtime-3.0_2.12-0.14.0-SNAPSHOT.jar`.
**Basic select query from a table failed with below error.**

Caused by: java.lang.NoSuchMethodError: io.netty.buffer.PooledByteBufAllocatorL.allocate(J)Lio/netty/buffer/UnsafeDirectLittleEndian;

Screenshot 2022-03-16 at 6 24 46 PM

Once I revert this Change. Query was success for me.

### DIFF
--- a/spark/v2.4/build.gradle
+++ b/spark/v2.4/build.gradle
@@ -121,7 +121,6 @@ project(':iceberg-spark:iceberg-spark-runtime-2.4') {
       exclude group: 'org.xerial.snappy'
       exclude group: 'javax.xml.bind'
       exclude group: 'javax.annotation'
-      exclude group: 'io.netty'
     }
   }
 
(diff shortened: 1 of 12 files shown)


ajantha-bhat:
@pan3793 , @rdblue and others: **Please try to reproduce locally with latest spark-3.0 run time jar before merging or investigating further on this.** 

I should have debugged more on this. But the revert worked for me and Bit occupied with some other work. Hence I am notifying here. Thanks.

pan3793 (approved):

pan3793:
I see what happens, the `io.netty.buffer.PooledByteBufAllocatorL` comes from `arrow-memory-netty`, and I guess it use the package `io.netty.buffer` rather than `org.apache.arrow` may cause of class access permission, thus we can not relocated those classes, the only option should be shade and relocate all netty classes.

ajantha-bhat:
@pan3793 : Thanks for checking and approving the change. 

@rdblue , @openinx : Can you please help merge this PR ?

ajantha-bhat:
ping... @rdblue

rdblue:
Reverted. Thanks for looking into this more, @pan3793 and @ajantha-bhat!

PR #4345, opened 2022-03-16
Doc: Fix spec doc read-   to  read-optimized

renshangtao:
"Both read-  and write-optimized formats will be available."

This sentence lost a word.


rdblue:
Why remove `ParquetWriter`?

PR #4347, opened 2022-03-16
AWS : Address S3FileIOIntegrationTest failures

singhpk234:
presently TestS3FileIOIntegration fails with NPE because of s3FileIo.Initialize not being called.

This happens because writeTags is never intialized if we don't call it's initialize and it always remains null so when we check if it's empty or not while initializing upload / mutlipart upload it faces NPE.

https://github.com/apache/iceberg/blob/c1adf2034280884d05afad4e67f09f2872c47d78/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java#L70

https://github.com/apache/iceberg/blob/c1adf2034280884d05afad4e67f09f2872c47d78/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java#L263

Ideally we should initialize while declaration like we do metrics to avoid this scenario. 

Post this fix test passes

-----


jackye1995 (approved):
thanks for the fix! Verified it works

PR #4348, opened 2022-03-16
Core: Add RESTCatalog and RESTTableOperations

rdblue:
This adds `RESTCatalog` and `RESTTableOperations` implementations that are passing `CatalogTests` from #4319. This is the remainder of the changes from #4241.

In addition, this adds `CatalogHandlers` that implements the REST server operations that accept request classes, call `Catalog`, and produce response classes. This class can be used to implement REST catalog services that are backed by a `Catalog`, like `JdbcCatalog`. In this PR, the handlers are used to test `RESTCatalog` and `RESTTableOperations` along with a test class that implements `RESTClient` and calls the correct handler.

There are some features of the REST catalog that are missing and will be done in follow-up PRs:
* The config route is not called to configure the catalog
* The purge flag is not passed to the catalog service
* The parent query param for namespace listing is not passed to the catalog service
* Rename is not implemented

To support create transactions in `CatalogHandlers`, this exposes the table operations and starting metadata in `BaseTransaction`.


core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+      CreateTableRequest request = CreateTableRequest.builder()
+          .withName(ident.name())
+          .withSchema(schema)
+          .withPartitionSpec(spec)
+          .withWriteOrder(writeOrder)
+          .withLocation(location)
+          .setProperties(properties)
+          .build();
+
+      // TODO: will this be a specific route or a modified create?
```
rdblue, line 352: Right now, this uses a new route to state a table create. I haven't added this to the REST spec yet because it is an open question whether we want to use the existing create route with a query param or request flag, or use a separate route like this. Any opinions?

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
```diff
@@ -0,0 +1,212 @@
...
+      default:
+        throw new UnsupportedOperationException(String.format("Update type %s is not supported", updateType));
+    }
+
+    baseChanges.forEach(requestBuilder::update);
+    metadata.changes().forEach(requestBuilder::update);
+    UpdateTableRequest request = requestBuilder.build();
+
+    // the error handler will throw necessary exceptions like CommitFailedException and UnknownCommitStateException
+    // TODO: ensure that the HTTP client lib passes HTTP client errors to the error handler
```
rdblue, line 123: @kbendick, this is relevant to the HTTP implementation of `RESTClient`. How does it handle errors that are coming from the HTTP connection and not from the service? In some cases, we need to throw `CommitStateUnknownException`.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+import static org.apache.iceberg.TableProperties.COMMIT_NUM_RETRIES_DEFAULT;
+import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT;
+
+public class CatalogHandlers {
+  private static final Schema EMPTY_SCHEMA = new Schema();
+
+  private CatalogHandlers() {
+  }
+
+  static class UnprocessableEntityException extends RuntimeException {
```
RussellSpitzer, line 71: I think it would help even though they are mostly internal classes, to have some brief javadocs on when they should be thrown. The message here makes it seem like this is only used for Namespace updates?

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+    UnprocessableEntityException(Set<String> commonKeys) {
+      super(String.format("Invalid namespace update, cannot set and remove keys: %s", commonKeys));
+    }
+  }
+
+  private static class ValidationFailureException extends RuntimeException {
```
RussellSpitzer, line 77: Is this just a Runtime wrapped version of CFE?

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+
+public class CatalogHandlers {
+  private static final Schema EMPTY_SCHEMA = new Schema();
+
+  private CatalogHandlers() {
+  }
+
+  static class UnprocessableEntityException extends RuntimeException {
+    UnprocessableEntityException(Set<String> commonKeys) {
+      super(String.format("Invalid namespace update, cannot set and remove keys: %s", commonKeys));
```
RussellSpitzer, line 73: Looking below, I think this should be something more in line with, "cannot simultaneously set and remove keys: "?

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  public static UpdateNamespacePropertiesResponse updateNamespaceProperties(
+      SupportsNamespaces catalog, Namespace namespace, UpdateNamespacePropertiesRequest request) {
+    Set<String> removals = Sets.newHashSet(request.removals());
+    Map<String, String> updates = request.updates();
+
+    Set<String> commonKeys = Sets.intersection(updates.keySet(), removals);
+    if (!commonKeys.isEmpty()) {
+      throw new UnprocessableEntityException(commonKeys);
```
RussellSpitzer, line 132: I wonder if this should be checked in the UpdateNamespacePropertiesRequest constructor? NBD, just wondering because we control the construction of that object as well.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    Map<String, String> startProperties = catalog.loadNamespaceMetadata(namespace);
+    Set<String> missing = Sets.difference(removals, startProperties.keySet());
+
```
RussellSpitzer, line 137: Another potential validation here, we are allowed to request the removal of a property that is not set, is that ok? Seems fine to me since I think a lot of property structures let you remove a non-existing key without error.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+    if (!removals.isEmpty()) {
+      // remove the original set just in case there was an update just after loading properties
+      catalog.removeProperties(namespace, removals);
+    }
+
+    return UpdateNamespacePropertiesResponse.builder()
+            .addMissing(missing)
+            .addUpdated(updates.keySet())
+            .addRemoved(Sets.difference(removals, missing))
```
RussellSpitzer, line 150: This may not be true at this point since (as your comment alludes to above) we may have removed properties that were set in between our determination of missing and our execution of catalog.removeProperties. Not sure if we can make a better response though unless we change catalog.removeProperties to return which properties actually got removed.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+      throw new AlreadyExistsException("Table already exists: %s", ident);
+    }
+
+    Map<String, String> properties = Maps.newHashMap();
+    properties.put("created-at", OffsetDateTime.now().toString());
+    properties.putAll(request.properties());
+
+    TableMetadata metadata = TableMetadata.newTableMetadata(
+        request.schema(),
+        request.spec() != null ? request.spec() : PartitionSpec.unpartitioned(),
```
RussellSpitzer, line 173: I wonder if we should just add these to the "spec() and writeOrder()" methods rather than null checking here.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
RussellSpitzer: Checking below we have this as part of CreateTableRequest.validate() so maybe it should be part of a .validate() method?

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+  }
+
+  public static LoadTableResponse updateTable(Catalog catalog, TableIdentifier ident, UpdateTableRequest request) {
+    TableMetadata finalMetadata;
+    if (isCreate(request)) {
+      // this is a hacky way to get TableOperations for an uncommitted table
+      Transaction transaction = catalog.buildTable(ident, EMPTY_SCHEMA).createOrReplaceTransaction();
```
RussellSpitzer, line 224: Is the plan here that we end up making a replace table transaction here which replaces the newly created table with the updates? I'm not quite sure I follow this branch

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,315 @@
...
+  private static boolean isCreate(UpdateTableRequest request) {
+    boolean isCreate = request.requirements().stream()
+        .anyMatch(UpdateTableRequest.UpdateRequirement.AssertTableDoesNotExist.class::isInstance);
+
+    if (isCreate) {
+      List<UpdateTableRequest.UpdateRequirement> invalidRequirements = request.requirements().stream()
+          .filter(req -> !(req instanceof UpdateTableRequest.UpdateRequirement.AssertTableDoesNotExist))
+          .collect(Collectors.toList());
+      Preconditions.checkArgument(invalidRequirements.isEmpty(),
+          "Invalid create requirements: %s", invalidRequirements);
```
RussellSpitzer, line 256: So either all requirements assert that the table does not exist, or none do? If only some do we have an invalid Create request?

RussellSpitzer (commented):
Went through CatalogHandlers, Will try to get through the rest later

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+import org.apache.iceberg.rest.responses.DropNamespaceResponse;
+import org.apache.iceberg.rest.responses.DropTableResponse;
+import org.apache.iceberg.rest.responses.GetNamespaceResponse;
+import org.apache.iceberg.rest.responses.ListNamespacesResponse;
+import org.apache.iceberg.rest.responses.ListTablesResponse;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
+import org.apache.iceberg.util.Pair;
+
+public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Configuration> {
```
danielcweeks, line 65: Is it possible to support configurable without a direct dependency on Hadoop?  This will cause problems in non-hadoop environments.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+    String tablePath = tablePath(identifier);
+    // TODO: support purge flag
+    DropTableResponse response = client.delete(
+        tablePath, DropTableResponse.class, ErrorHandlers.tableErrorHandler());
+    return response.isDropped();
+  }
+
+  @Override
+  public void renameTable(TableIdentifier from, TableIdentifier to) {
+
```
danielcweeks, line 115: Looks like this implementation may have been overlooked

danielcweeks: nevermind, just saw this wasn't slated for this PR

core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -90,6 +90,14 @@ public Table table() {
     return transactionTable;
   }
 
+  public TableMetadata startMetadata() {
+    return current;
+  }
+
+  public TableOperations underyingOps() {
```
danielcweeks, line 97: nit: typo: `underlying`

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+    return String.format("%s.%s", catalogName, ident);
+  }
+
+  private static String tablePath(TableIdentifier ident) {
+    return "v1/namespaces/" + RESTUtil.urlEncode(ident.namespace()) + "/tables/" + ident.name();
+  }
+
+  private Map<String, String> fullConf(Map<String, String> config) {
+    Map<String, String> fullConf = Maps.newHashMap(properties);
+    properties.putAll(config);
```
danielcweeks, line 400: Maybe I'm reading this wrong, but shouldn't this be `fullConfig.putAll`?  This looks like we're updating the catalog level properties and then just returning the original catalog properties.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
kbendick: Currently, this is HTTP response is presently used only for conflicting namespace property updates.

Though I don't disagree agree that some internal JavaDocs would be very beneficial.

I can grab some links to the spec that we might be able to pull from like [this usage of this exception](https://github.com/apache/iceberg/blob/dcbd5f694908781c07937f495690a4ae16886c61/open-api/rest-catalog-open-api.yaml#L302-L310).

The Mozilla docs would also be helpful, as this is HTTP 422 [`Unprocessable Entity`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422). But the specific occurrence we have at present that is represented here is in the mentioned namespace update scenario.

Per the spec, [this is the example](https://github.com/apache/iceberg/blob/dcbd5f694908781c07937f495690a4ae16886c61/open-api/rest-catalog-open-api.yaml#L1714-L1724) for this specific error (we can update the example's phrasing to match what's here as I prefer the usage in this PR for this one - but they can åbe consolidated generally speaking).

```
        "error": {
          "message": "The request cannot be processed as there is a key present multiple times",
          "type": "UnprocessableEntityException",
          "code": 422
        }
```

core/src/main/java/org/apache/iceberg/BaseTransaction.java
rdblue: Good catch!

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Yeah, that's better.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Yeah, this one is purely internal because we always catch it and unwrap the CFE.

I'll move the other one to a top-level class and add Javadoc to both.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Yeah, good call. I'll put it in `validate`.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+        .build();
+
+    // for now, ignore the response because there is no way to return it
+    client.post("v1/namespaces", request, CreateNamespaceResponse.class, ErrorHandlers.namespaceErrorHandler());
+  }
+
+  @Override
+  public List<Namespace> listNamespaces(Namespace namespace) throws NoSuchNamespaceException {
+    Preconditions.checkArgument(namespace.isEmpty(), "Cannot list namespaces under parent: %s", namespace);
+    // String joined = NULL.join(namespace.levels());
```
danielcweeks, line 147: nit: unnecessary comment

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: The idea is for these to be idempotent, so it is fine to request removing a property that is not set.

If we didn't allow that, then you'd get incorrect errors when a user does something like `SHOW TBLPROPERTIES` followed by concurrent `UNSET` commands. Retries would also cause problems.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: This is something for implementations to improve on. Right now there's no way to do this through the catalog API.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
danielcweeks: Actually, it looks like we're not using `namespaces` here.  This should be either the `/v1/namespaces` route with query params.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: I don't know. I'm reluctant to make the request return defaults instead of `null`. I'd probably leave this as it is here and let the implementation decide how to handle requests without these. That, or update `newTableMetadata` to do it.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Replace transactions are already working.

What's happening here is that the existing `Catalog` implementations don't allow you to get `TableOperations`. For all other commits, we load the table and grab `TableOperations` from `BaseTable` (see line 236). But for a create transaction, there is no table to load. The way I found to create `TableOperations` for a table that doesn't exist yet is to start a create transaction and get `TableOperations` from there.

As the comment says, this is a hacky way to get `TableOperations`, but I don't have a better one. Other options are: (1) don't support create transactions when wrapping `Catalog`, or (2) expose `TableOperations` in `Catalog` and force implementations to add a factory method. I think this is the cleanest way.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Yes, that's correct.

If you're asserting that the table does not exist, then it makes no sense to assert that it has a certain current schema ID, for example.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: This doesn't have a direct dependency on Hadoop. That's why we're using Iceberg's `Configurable`. The Hadoop `Configuration` class here is removed by type erasure.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
rdblue: Fixed.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Fixed.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Done.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Fixed.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Moved to the exceptions package as a subclass of `RESTException`.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
danielcweeks: I personally vote for query param.  It's almost an entire duplication of the create route if we go with a separate endpoint.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: Yes, you're right.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: Yeah, the parent query param isn't supported yet.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: Fixed.

rdblue:
Since this is already large and close to being committed, I plan to update the code for #4366 once this is in.

api/src/main/java/org/apache/iceberg/exceptions/UnprocessableEntityException.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.iceberg.exceptions;
+
+import com.google.errorprone.annotations.FormatMethod;
+
+/**
+ * REST exception thrown when a request is well-formed but cannot be applied.
+ * <p>
+ * For example, this is used when a property update requests that properties are simultaneously set and removed.
+ */
+public class UnprocessableEntityException extends RESTException {
```
kbendick, line 29: It might make sense to mention that this is HTTP 422. Somewhat implied by the name as well as extending RESTException though.

api/src/main/java/org/apache/iceberg/exceptions/UnprocessableEntityException.java
```diff
@@ -0,0 +1,34 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.exceptions;
+
+import com.google.errorprone.annotations.FormatMethod;
```
openinx, line 22: Should we also relocate this common `com.google.errorprone` package prefix ?

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
```diff
@@ -0,0 +1,212 @@
...
+
+      case REPLACE:
+        Preconditions.checkState(base != null, "Invalid base metadata: null");
+        // use the original replace base metadata because the transaction will refresh
+        requestBuilder = UpdateTableRequest.builderForReplace(replaceBase);
+        baseChanges = createChanges;
+        errorHandler = ErrorHandlers.tableCommitHandler();
+        break;
+
+      case SIMPLE:
```
openinx, line 107: What's the usage & meaning of the `SIMPLE` type ?

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+    String ns = RESTUtil.urlEncode(namespace);
+    ListTablesResponse response = client
+        .get("v1/namespaces/" + ns + "/tables", ListTablesResponse.class, ErrorHandlers.namespaceErrorHandler());
+    return response.identifiers();
+  }
+
+  @Override
+  public boolean dropTable(TableIdentifier identifier, boolean purge) {
+    String tablePath = tablePath(identifier);
+    // TODO: support purge flagN
```
RussellSpitzer, line 107: typo, also maybe we should throw an error if "UnsupportedOperationException" if purge = true? I guess technically whatever happens is up to the catalog if we don't pass through option, but still maybe best to just throw an error till we have the flag actually propagated?

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
RussellSpitzer: Unsupported Operation Exception?

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+        .setProperties(metadata)
+        .build();
+
+    // for now, ignore the response because there is no way to return it
+    client.post("v1/namespaces", request, CreateNamespaceResponse.class, ErrorHandlers.namespaceErrorHandler());
+  }
+
+  @Override
+  public List<Namespace> listNamespaces(Namespace namespace) throws NoSuchNamespaceException {
+    Preconditions.checkArgument(namespace.isEmpty(), "Cannot list namespaces under parent: %s", namespace);
```
RussellSpitzer, line 146: "Cannot list namespaces with a parent specified: %s passed" ?

The current message has an implication that perhaps another parent would work.

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
```diff
@@ -0,0 +1,212 @@
...
+    switch (updateType) {
+      case CREATE:
+        Preconditions.checkState(base == null, "Invalid base metadata for create transaction, expected null: %s", base);
+        requestBuilder = UpdateTableRequest.builderForCreate();
+        baseChanges = createChanges;
+        errorHandler = ErrorHandlers.tableErrorHandler(); // throws NoSuchTableException
+        break;
+
+      case REPLACE:
+        Preconditions.checkState(base != null, "Invalid base metadata: null");
```
RussellSpitzer, line 100: "Invalid base metadata for replace transaction, expected non-null"?

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
```diff
@@ -0,0 +1,212 @@
...
+  public void commit(TableMetadata base, TableMetadata metadata) {
+    UpdateTableRequest.Builder requestBuilder;
+    List<MetadataUpdate> baseChanges;
+    Consumer<ErrorResponse> errorHandler;
+    switch (updateType) {
+      case CREATE:
+        Preconditions.checkState(base == null, "Invalid base metadata for create transaction, expected null: %s", base);
+        requestBuilder = UpdateTableRequest.builderForCreate();
+        baseChanges = createChanges;
+        errorHandler = ErrorHandlers.tableErrorHandler(); // throws NoSuchTableException
```
RussellSpitzer, line 96: Not sure I follow the comment here, shouldn't we through NoSuchTableException in the Replace branch below?

RussellSpitzer: I see it has

https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/rest/ErrorHandlers.java#L80

Which we do want

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
```diff
@@ -0,0 +1,212 @@
...
+    // safely ignored. there is no requirement to update config on refresh or commit.
+    if (current == null || !Objects.equals(current.metadataFileLocation(), response.metadataLocation())) {
+      this.current = response.tableMetadata();
+    }
+
+    return current;
+  }
+
+  private static String metadataFileLocation(TableMetadata metadata, String filename) {
+    String metadataLocation = metadata.properties()
```
RussellSpitzer, line 148: nit: maybe rename to customMetadataLocation, otherwise the code below reads a bit like "if metadatalocation is null use metadata.location".

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,314 @@
...
+
+    if (!removals.isEmpty()) {
+      // remove the original set just in case there was an update just after loading properties
+      catalog.removeProperties(namespace, removals);
+    }
+
+    return UpdateNamespacePropertiesResponse.builder()
+            .addMissing(missing)
+            .addUpdated(updates.keySet())
+            .addRemoved(Sets.difference(removals, missing))
```
kbendick, line 149: Nit: the preceding 4 lines are over indented.

RussellSpitzer (approved):
Left some minor comments, Looks good to me as a first step

api/src/main/java/org/apache/iceberg/exceptions/UnprocessableEntityException.java
kbendick: I checked in our other exceptions that use `FormatMethod`, and Russell is right. We import this without relocating.

I don't think it needs to be relocated as it only exists at compile time. In either case, it shouldn't be done in this PR. But it shouldn't be necessary at all.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -0,0 +1,314 @@
...
+            .build();
```
kbendick, line 150: Nit: This line and the ones above it are over indented.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+
+    return changes.build();
+  }
+
+  private String fullTableName(TableIdentifier ident) {
+    return String.format("%s.%s", catalogName, ident);
+  }
+
+  private static String tablePath(TableIdentifier ident) {
+    return "v1/namespaces/" + RESTUtil.urlEncode(ident.namespace()) + "/tables/" + ident.name();
```
kbendick, line 395: What would happen if `ident.name()` contained a slash? I don't believe that there is anything preventing that from happening at the moment.

I checked and a TableIdentifier can be instantiated with a `/` in it. I personally think we should disallow this, as it would likely also cause issues with paths on HDFS.

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
rdblue: You're right. It should be AlreadyExistsException

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
rdblue: `metadata.location()` is the table's location, not the metadata location.

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
RussellSpitzer: I was matching them message used directly above, that's why I made the suggestion

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: Yeah, this is one of the problems I pointed out when we talked about adding a helper method to generate these paths. This is a test case for that extension.

rdblue:
Thanks for reviewing, @kbendick, @RussellSpitzer, @openinx, and @szehon-ho! I'm merging to unblock the next ones, which are what the remaining comments are about.

I'll also fix a couple of the nits in the next PR, like the indentation in `CatalogHandlers` and the comment about the error handler.

RussellSpitzer:
SGTM, There is always time for more commits :)

core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java
rdblue: I included more context above because it isn't clear _what_ was invalid about the metadata. I think the fact that Iceberg expected null is important. Here, the main thing the person reading this needs to know is that metadata was null and that's almost never a good thing.

Issue #3086, opened 2021-09-07
Allow config for write object storage path to be table location (for remove orphans)


szehon-ho:
The current "write.object-storage.path" is a good solution for throttling on object-stores, example S3.

However, we like to still use the built-in 'removeOrphan' functionality for tracking storage costs/privacy.  Remove Orphan relies on listing under table location to find potential orphan files.  Using the doc-recommended 'write.object-storage.path' : 's3://bucket/' breaks removeOrphans, as resultant data files: 's3://bucket/xxxxxxxx/db/table/data/part=a/0000.parquet' are not correctly identified in a listing of the table location: s3://bucket/db/table.

The only solution seems to be to set the write.object-storage.path to be the table location manually.  This does fix  throttling (if object store prefixes are set to each table location plus some characters of the hash)

But it is not so user friendly:

- Path becomes :  's3://bucket/db/table/xxxxxxxx/db/table/p=a/0000.parquet'.  In particular, the 'db/table' are repeated, adding unnecessarily to path length.  It is also a bit awkward in general.
- From UX perspective, user needs to be aware of the table location when setting this property (table location is optional in normal create table statement and many users do not set it).  It complicates efforts to set this en-masse

It will be great to have a new property like 'write.object-storage.table-location', which if enabled to true:

- For table with default table location (s3://bucket/db/table), sets the path to be the more reasonable and efficient 's3://bucket/db/table/xxxxxxxx/p=a/0000.parquet' (removing /data/db/table/ just as /data was removed by the original implementation)
- sets the write object storage path to automatically following the default table-location without user having to specify explicitly any path in either write.object-storage.path or table-location.  Changing table-location will automatically change the write object storage path.

szehon-ho:
Do folks think this will be useful?  @jackye1995 for feedback

szehon-ho:
Closing this issue as the first problem has been addressed in #4349.  The second is still kind of annoying to have to do , maybe we can still add the flag if there is interest to default the write-object-store-path to table's location

PR #4349, opened 2022-03-16
Core: Skip context for object storage in table prefixes

rdblue:
This detects when an object storage location is within a table's prefix and avoids adding the table and database names to file paths as context.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/LocationProviders.java
+++ b/core/src/main/java/org/apache/iceberg/LocationProviders.java
@@ -104,7 +104,12 @@ public class LocationProviders {
 
     ObjectStoreLocationProvider(String tableLocation, Map<String, String> properties) {
       this.storageLocation = stripTrailingSlash(dataLocation(properties, tableLocation));
-      this.context = pathContext(tableLocation);
+      // if the storage location is within the table prefix, don't add table and database name context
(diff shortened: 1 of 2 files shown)


szehon-ho (approved):
I think this is a great change, I had seen this myself and filed an overcomplicated issue https://github.com/apache/iceberg/issues/3086, this is much cleaner.

So, upon upgrade of Iceberg, tables will start writing files to new locations (still under the configured location), should be minimal user impact.

Do you think its worth effort to add a small test in TestLocationProvider ?

anuragmantri (approved):
Thanks for making this change. Maybe it's worth adding 
- a unit test
- a note in the [docs](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout) about this behavior

rdblue:
Thanks for reviewing, @szehon-ho and @anuragmantri!

rdblue:
> a note in the [docs](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout) about this behavior

There's currently no discussion of how object storage paths are laid out and I'm not sure I'd over-complicate the page by describing it in enough detail to cover this. I think it is sufficient to say that we use a hash in the location to distribute files.

PR #4350, opened 2022-03-17
AWS: Move S3 writeTags to AwsProperties

rajarshisarkar:
This change moves S3 `writeTags` to `AwsProperties`. This provides a cleaner solution as `AwsProperties` was designed to be the extension point for new configs without changing constructors.

Users can pass their custom tags as part of the catalog properties.

Spark SQL launch command:
```
    --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key2=my_val2
```

Tags added in S3:

```
aws s3api get-object-tagging --bucket iceberg-warehouse --key s3-tagging/data/00000-0-7eb83d01-ad82-41ca-814b-c7686d9d9211-00001.parquet
{
    "TagSet": [
        {
            "Key": "my_key2",
            "Value": "my_val2"
        },
        {
            "Key": "my_key",
            "Value": "my_val"
        }
    ]
}
```
---

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
@@ -26,6 +26,7 @@ import java.util.stream.Collectors;
 import org.apache.iceberg.aws.dynamodb.DynamoDbCatalog;
 import org.apache.iceberg.aws.s3.S3FileIO;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.util.PropertyUtil;
 import software.amazon.awssdk.services.s3.model.ObjectCannedACL;
 import software.amazon.awssdk.services.s3.model.Tag;
@@ -270,7 +271,7 @@ public class AwsProperties implements Serializable {
 
   private boolean isS3ChecksumEnabled;
 
-  private Set<Tag> s3WriteTags;
+  private final Set<Tag> s3WriteTags;
 
   public AwsProperties() {
     this.s3FileIoSseType = S3FILEIO_SSE_TYPE_NONE;
@@ -289,6 +290,8 @@ public class AwsProperties implements Serializable {
     this.glueCatalogSkipArchive = GLUE_CATALOG_SKIP_ARCHIVE_DEFAULT;
 
     this.dynamoDbTableName = DYNAMODB_TABLE_NAME_DEFAULT;
+
+    this.s3WriteTags = Sets.newHashSet();
   }
 
   public AwsProperties(Map<String, String> properties) {
@@ -456,7 +459,7 @@ public class AwsProperties implements Serializable {
     return s3WriteTags;
   }
 
-  public Set<Tag> toTags(Map<String, String> properties, String prefix) {
+  private Set<Tag> toTags(Map<String, String> properties, String prefix) {
     return PropertyUtil.propertiesWithPrefix(properties, prefix)
         .entrySet().stream()
         .map(e -> Tag.builder().key(e.getKey()).value(e.getValue()).build())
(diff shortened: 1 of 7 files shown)


jackye1995:
Thanks, this was based on the discussion in https://github.com/apache/iceberg/pull/4334/files#r828411959 for some context

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -267,6 +270,8 @@
 
   private boolean isS3ChecksumEnabled;
 
+  private Set<Tag> writeTags;
```
jackye1995, line 273: prefer to prefix with service name, `s3WriteTags`

jackye1995 (approved):
overall looks good to me, just a nitpicking

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -339,6 +344,8 @@ public AwsProperties(Map<String, String> properties) {
 
     this.dynamoDbTableName = PropertyUtil.propertyAsString(properties, DYNAMODB_TABLE_NAME,
         DYNAMODB_TABLE_NAME_DEFAULT);
+
+    this.s3WriteTags = toTags(properties, S3_WRITE_TAGS_PREFIX);
```
singhpk234, line 348: we should also initialize the s3WriteTags value in the constructor above : 
https://github.com/apache/iceberg/pull/4350/files#diff-19af8af4e4204814f89c7e71633aae4b635229472dd20772d7651044d20cc900R291

```java
this.s3WriteTags = Sets.newHashSet();
```

otherwise the people using the constructor will get NPE.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -444,4 +451,15 @@ public boolean isS3ChecksumEnabled() {
...
+  public Set<Tag> getS3WriteTags() {
+    return s3WriteTags;
+  }
+
+  public Set<Tag> toTags(Map<String, String> properties, String prefix) {
+    return PropertyUtil.propertiesWithPrefix(properties, prefix)
+        .entrySet().stream()
+        .map(e -> Tag.builder().key(e.getKey()).value(e.getValue()).build())
+        .collect(Collectors.toSet());
+  }
```
singhpk234, line 464: [question] any reason why we made it public, as earlier it was private

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
singhpk234: Ahh I see you made a check of tags != null below then it might not be required, but I think still empty set as default is good as it avoids us checking the null checks when it's being used. WDYT ?

rajarshisarkar: I agree, empty set would be a cleaner solution. I have made the changes.

singhpk234 (approved):
Thanks @rajarshisarkar for the change :) , LGTM !!!

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
rajarshisarkar: It can be private here (unlike other methods of `AwsProperties` which are public). I don't see a scope of `toTags` method to be used by other classes. We can make it public later on, if required.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
jackye1995: Nice catch for this! I think we still need to check null, because the object is serialized, and it might serialize an old version of the object that does not have the `writeTags` fields, which means the object will be deserialized with the value equal to null.

Also nit: similar to the above comments, could you move the assignment to L347?

jackye1995 (approved):

PR #4355, opened 2022-03-17
AWS: Fix TestGlueCatalogTable failure

xiaoxuandev:


aws/src/integration/java/org/apache/iceberg/aws/glue/GlueTestBase.java
```diff
@@ -69,8 +69,8 @@
 
   // table location properties
   static final Map<String, String> tableLocationProperties = ImmutableMap.of(
-      TableProperties.WRITE_DATA_LOCATION, "s3://" + testBucketName + "/writeDataLoc",
       TableProperties.WRITE_METADATA_LOCATION, "s3://" + testBucketName + "/writeMetaDataLoc",
+      TableProperties.WRITE_DATA_LOCATION, "s3://" + testBucketName + "/writeDataLoc",
```
singhpk234, line 73: if we were to make changes I would recommend changing this : 
https://github.com/apache/iceberg/blob/5b9cd7a68478f9fddd569765bfaec285d92f5907/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogTable.java#L77

sort both the collections as it may happen post picking your changes it failing in my machine
```java
    Assert.assertEquals("additionalLocations should match", tableLocationProperties.values().stream().sorted().collect(
        Collectors.toList()),
        response.table().storageDescriptor().additionalLocations().stream().sorted().collect(Collectors.toList()));
```

xiaoxuandev: Thanks for the suggestion. That makes sense as we are comparing a map's valueSet and a list. I have updated the change. Please have another look.

singhpk234 (approved):
LGTM, Thanks @xiaoxuandev for the fix :) !!

jackye1995 (approved):
also looks good to me!

aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogTable.java
```diff
@@ -74,8 +75,9 @@ public void testCreateTable() {
     Assert.assertTrue(response.table().parameters().containsKey(BaseMetastoreTableOperations.METADATA_LOCATION_PROP));
     Assert.assertEquals(schema.columns().size(), response.table().storageDescriptor().columns().size());
     Assert.assertEquals(partitionSpec.fields().size(), response.table().partitionKeys().size());
-    Assert.assertEquals("additionalLocations should match", tableLocationProperties.values(),
-        response.table().storageDescriptor().additionalLocations());
+    Assert.assertEquals("additionalLocations should match", tableLocationProperties.values().stream().sorted().collect(
```
jackye1995, line 78: can we break newline at a more intuitive place?

```
Assert.assertEquals("additionalLocations should match", 
  tableLocationProperties.values().stream().sorted().collect(Collectors.toList()),
  response.table().storageDescriptor().additionalLocations().stream().sorted().collect(Collectors.toList()));
```

jackye1995: actually, it looks like we should compare it as Set instead of List

xiaoxuandev: fixed line break. list also handles special case like duplicates.

jackye1995: oh yes good point

jackye1995 (approved):
thanks for the fix!

PR #4359, opened 2022-03-18
[Spark][Core] Convert rewrite results after committed

xloya:
Convert and save rewrite results in advance after rewrite groups committed, we don't actually use the rewrite group information in the results except the number of rewrite groups
cc @RussellSpitzer

### DIFF
--- a/core/src/main/java/org/apache/iceberg/actions/RewriteDataFilesCommitManager.java
+++ b/core/src/main/java/org/apache/iceberg/actions/RewriteDataFilesCommitManager.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.RewriteFiles;
 import org.apache.iceberg.Table;
@@ -133,7 +134,7 @@ public class RewriteDataFilesCommitManager {
   public class CommitService implements Closeable {
     private final ExecutorService committerService;
     private final ConcurrentLinkedQueue<RewriteFileGroup> completedRewrites;
-    private final List<RewriteFileGroup> committedRewrites;
+    private final List<RewriteDataFiles.FileGroupRewriteResult> committedRewrites;
     private final int rewritesPerCommit;
     private final AtomicBoolean running = new AtomicBoolean(false);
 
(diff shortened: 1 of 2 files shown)


xloya:
The single tests failed because the `Core`  module was modified and the historical spark versions were incompatible. The backports will be added after reviews

core/src/main/java/org/apache/iceberg/actions/RewriteDataFilesCommitManager.java
```diff
@@ -133,7 +134,7 @@ public CommitService service(int rewritesPerCommit) {
   public class CommitService implements Closeable {
     private final ExecutorService committerService;
     private final ConcurrentLinkedQueue<RewriteFileGroup> completedRewrites;
-    private final List<RewriteFileGroup> committedRewrites;
+    private final List<RewriteDataFiles.FileGroupRewriteResult> committedRewrites;
```
rdblue, line 137: This should not expose the concrete class.

xloya: I'm not sure if this modification is standard, because the `asResults()` method in the current `RewriteFileGroup` class should actually be written like this:  
https://github.com/apache/iceberg/blob/0f3e7c7badf47bd0e2178591082886cd47d94a1b/core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java#L67-L70  
`  public RewriteDataFiles.FileGroupRewriteResult asResult() {  
    Preconditions.checkState(addedFiles != null, "Cannot get result, Group was never rewritten");  
    return new BaseFileGroupRewriteResult(info, addedFiles.size(), fileScanTasks.size());  
  }`

RussellSpitzer:
I'm not sure I see the benefit to this change? Could you elaborate?

xloya:
> I'm not sure I see the benefit to this change? Could you elaborate?

Let me explain why this change was made:  
Recently, I found that the driver oom will occur when the compaction jobs rewrited the v2 table. After analyzing the memory of the driver heap, I found that the `RewriteFileGroup` will occupy a large amount of memory, so I thought maybe we could get rid of the reference to the group where the rewrite was done.  
The heap occupation like this:  
image

We now use `Stream` to save the `RewriteFileGroup`, which will not release the memory until all threads are completed. I'm not sure if anyone has encountered it, but the compaction job with v2 tables on my side encounters a lot of groups that need to be rewritten, resulting in a lot of memory for this class. So I modified the `Stream` to a `List,` removed the reference to it after the rewrite, and converted it to a simple class when storing in the result set to reduce the memory occupation.  

@RussellSpitzer What do you think about this? Please give any comments, thanks!

PR #4360, opened 2022-03-18
API: Add UnboundSortOrder that has no schema

rdblue:
This is to support the REST catalog. Building a sort order currently requires a schema, which is not part of the serialization format. This adds an intermediate class, `UnboundSortOrder` that can be built without a schema and later bound to a schema.


danielcweeks (approved):
LGTM

rdblue:
Thanks, @danielcweeks!

kbendick:
Have been out sick but retroactive +1 as well.

PR #4361, opened 2022-03-18
API: Add UnboundPartitionSpec that has no schema

rdblue:
Similar to #4360, this adds an unbound partition spec that can deserialize without a schema and bind to a schema to produce a normal `PartitionSpec`.


danielcweeks (approved):
LGTM

danielcweeks:
Looks like we have some check failures (not sure if those are related)

PR #4366, opened 2022-03-20
Core: Update DELETE routes to use 204 for success

rdblue:
This is an alternative to #4337. Instead of using 200 for all cases where the table does not exist, it uses 204 if the namespace or table was removed (equivalent to dropped=true) and 404 if the namespace or table did not exist (equivalent to dropped=false).


danielcweeks (approved):

rdblue:
Thanks for reviewing, @danielcweeks! I'll open a PR for this once #4348 is in.

PR #4367, opened 2022-03-20
Spec: Clean up error responses in REST protocol

rdblue:
This cleans up some of the error response documentation in the REST protocol.

* Documents 500 and 504 specifically in the table commit route
* Adds `stack` to error responses for sending server-side stack traces (optional)
* Adds 419 to all routes to signal that authentication has timed out
* Adds 503 to all routes for a back-pressure signal

### DIFF
--- a/open-api/rest-catalog-open-api.yaml
+++ b/open-api/rest-catalog-open-api.yaml
@@ -115,6 +115,12 @@ paths:
           $ref: '#/components/responses/UnauthorizedResponse'
         403:
           $ref: '#/components/responses/ForbiddenResponse'
+        419:
+          $ref: '#/components/responses/AuthenticationTimeoutResponse'
+        503:
+          $ref: '#/components/responses/ServiceUnavailableResponse'
+        5XX:
+          $ref: '#/components/responses/ServerErrorResponse'
 
   /v1/namespaces:
 
@@ -158,6 +164,10 @@ paths:
               examples:
                 NoSuchNamespaceExample:
                   $ref: '#/components/examples/NoSuchNamespaceError'
+        419:
+          $ref: '#/components/responses/AuthenticationTimeoutResponse'
+        503:
+          $ref: '#/components/responses/ServiceUnavailableResponse'
         5XX:
           $ref: '#/components/responses/ServerErrorResponse'
 
@@ -194,6 +204,10 @@ paths:
               examples:
                 NamespaceAlreadyExists:
                   $ref: '#/components/examples/NamespaceAlreadyExistsError'
+        419:
+          $ref: '#/components/responses/AuthenticationTimeoutResponse'
+        503:
+          $ref: '#/components/responses/ServiceUnavailableResponse'
         5XX:
           $ref: '#/components/responses/ServerErrorResponse'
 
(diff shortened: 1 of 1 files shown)


danielcweeks (approved):
LGTM

open-api/rest-catalog-open-api.yaml
```diff
@@ -1537,9 +1622,39 @@ components:
...
+              "message": "Slow down",
+              "type": "SlowDownException",
+              "code": 503
+            }
+          }
+
     ServerErrorResponse:
       description:
-        A server-side problem that might not be addressable from the client side. Used for server 5xx errors.
+        A server-side problem that might not be addressable from the client side. Used for server 5xx errors without more specific documentation in individual routes.
```
kbendick, line 1657: Nit: Can we wrap this description so that it's not all on one line? Seems a little long to me. It is 167 characters (including leading white space).

Maybe placing the second sentence on its own line would help.

open-api/rest-catalog-open-api.yaml
```diff
@@ -1537,9 +1622,39 @@ components:
...
+            "error": {
+              "message": "Credentials have timed out",
+              "type": "AuthenticationTimeoutException",
+              "code": 419
+            }
+          }
+
+    ServiceUnavailableResponse:
+      description:
+        The service is not ready to handle the request. The client should wait and retry.
```
kbendick, line 1642: According to the mozilla documentation, if possible, a `Retry-After` header should be used if possible to indicate when the client should retry after: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

Presumably the HTTP client handles all of this, and it doesn't need to be in the spec, but wanted to call attention to it. Possibly something to add to the description as a note, even if just in a follow up.

open-api/rest-catalog-open-api.yaml
```diff
@@ -469,23 +507,52 @@ paths:
...
-                  "type": "InternalServerError",
+                  "type": "CommitStateUnknownException",
                   "code": 500
                 }
               }
+        503:
+          $ref: '#/components/responses/ServiceUnavailableResponse'
+        504:
+          description:
+            An server-side gateway timeout occurred; the commit state is unknown.
```
kbendick, line 530: Nit: `A server-side` not `An server-side`. Unnecessary `n`.

kbendick (approved):
A few small nits, but overall this looks good to me.

Thank you @rdblue!

open-api/rest-catalog-open-api.yaml
rdblue: It doesn't sound like this is required. Maybe we should mention it as optional?

kbendick: Yeah it's not required. Mentioning it as optional seems beneficial to server implementors, particularly in the rate limiting case.

It can be added as a folow-up though.

rdblue: Added "The service may additionally send a Retry-After header to indicate when to retry".

rdblue:
Thanks for reviewing, @kbendick and @danielcweeks!

Issue #4192, opened 2022-02-22
Remove Orphan Files error...


Initial-neko:
when I call SparkAction 
`SparkActions
                .get()
                .deleteOrphanFiles(table)
                .olderThan(System.currentTimeMillis() - 1000000)
                .execute();`
it throws

`Exception in thread "main" java.lang.IllegalArgumentException: Cannot find the metadata table for xxxx of type ALL_MANIFESTS
	at org.apache.iceberg.spark.actions.BaseSparkAction.loadMetadataTable(BaseSparkAction.java:191)
	at org.apache.iceberg.spark.actions.BaseSparkAction.buildValidDataFileDF(BaseSparkAction.java:121)
	at org.apache.iceberg.spark.actions.BaseDeleteOrphanFilesSparkAction.doExecute(BaseDeleteOrphanFilesSparkAction.java:154)
	at org.apache.iceberg.spark.actions.BaseSparkAction.withJobGroupInfo(BaseSparkAction.java:101)
	at org.apache.iceberg.spark.actions.BaseDeleteOrphanFilesSparkAction.execute(BaseDeleteOrphanFilesSparkAction.java:141)
	at org.apache.iceberg.spark.actions.BaseDeleteOrphanFilesSparkAction.execute(BaseDeleteOrphanFilesSparkAction.java:76)
	`

So what exactly is a metadata table? Why don't I find it in my data file and metadata file
thanks~

RussellSpitzer:
The issue is probably that the table object you are using is not from the configured Spark Catalog.  Try loading the table instance without Spark3Util.loadIcebergTable. 

The error is because Spark needs the catalog name to match the one in the Spark conf. Otherwise it does not know where to find the table. When instantiating a new table using the Java api it will have the default catalog name which is most likely not the same.

Initial-neko:
> The issue is probably that the table object you are using is not from the configured Spark Catalog. Try loading the table instance without Spark3Util.loadIcebergTable.
> 
> The error is because Spark needs the catalog name to match the one in the Spark conf. Otherwise it does not know where to find the table. When instantiating a new table using the Java api it will have the default catalog name which is most likely not the same.
> 
> A metadata table is just a special view of a table which returns information like what partitions are in the table or what files are in the table. All iceberg tables have them available, see the docs for more info.

Our tables are loaded through hivecatalog. There will be no such problems in expiresnapshot and rewrite operations. I will confirm it; Thanks for your reply~

JunjianS:
[...]

Issue #4086, opened 2022-02-10
Compact data files faild : org.apache.spark.sql.AnalysisException: spark_catalog requires a single-part namespace, but got [hive, test]


lordk911:
--------vesion infomation----
spark : 3.2.1
iceberg : org.apache.iceberg:iceberg-spark-runtime-3.2_2.12:0.13.0
java : 1.8

```
val catalog = new HiveCatalog()
catalog.setConf(spark.sparkContext.hadoopConfiguration)
val properties = Map("warehouse" -> "hdfs://xx", "uri" -> "thrift://xx:9083,thrift://yy:9083").asJava
catalog.initialize("hive", properties)
val table = catalog.loadTable(TableIdentifier.of("test","aplog"))
SparkActions.get().rewriteDataFiles(table).filter(Expressions.lessThan("dtime", lasteHour)).option("target-file-size-bytes", String.valueOf(500 * 1024 * 1024)).execute();
```
<img width="1477" alt="image" src="https://user-images.githubusercontent.com/19989300/153427177-6836a58e-0d7a-42fa-8ed7-d061873e6f89.png">

RussellSpitzer:
@lordk911 the issue here is that the "HiveCatalog" you are creating in the code does not have information required for Spark to actually locate the table. Instead get the table reference through Spark by using 

https://github.com/apache/iceberg/blob/4d7837093bd19a42237446a9b656a248e106b789/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/Spark3Util.java#L634-L649

szehon-ho:
Yea I hit this too , maybe we should update the docs.. existing docs seem to indicate you can load the Table from a new HiveCatalog and expect it to work with SparkActions, which seems pretty logical.

https://iceberg.apache.org/docs/latest/maintenance/
https://iceberg.apache.org/docs/latest/java-api-quickstart/#create-a-table

Also just wondering, there's no way to strip the catalog before loading the table in the Spark3BinPackStrategy and have it work, right?

RussellSpitzer:
> Also just wondering, there's no way to strip the catalog before loading the table in the Spark3BinPackStrategy and have it work, right?

Yeah unfortunately the way we have it set up now is that we work directly on the "table" object passed in and if it is missing catalog information there is no way to get it back.

[...]

PR #4369, opened 2022-03-21
Docs: update Spark Write confs for isolation-level

szehon-ho:


docs/spark/spark-configuration.md
```diff
@@ -185,4 +185,5 @@ df.write
 | snapshot-property._custom-key_    | null            | Adds an entry with custom-key and corresponding value in the snapshot summary  |
 | fanout-enabled       | false        | Overrides this table's write.spark.fanout.enabled  |
 | check-ordering       | true        | Checks if input schema and table schema are same  |
-
+| isolation-level | null | Desired isolation level for transaction.  For overwrites, `serializable` checks for concurrent inserts or deletes in destination partitions, `snapshot` checks for concurrent deletes. |
+| validate-from-snapshot-id | null | If isolation level set, identifies base snapshot from which to check concurrent conflicts. If null, check from the very first snapshot |
```
RussellSpitzer, line 189: If null, uses the oldest unexpired snapshot

docs/spark/spark-configuration.md
```diff
@@ -185,4 +185,5 @@ df.write
 | snapshot-property._custom-key_    | null            | Adds an entry with custom-key and corresponding value in the snapshot summary  |
 | fanout-enabled       | false        | Overrides this table's write.spark.fanout.enabled  |
 | check-ordering       | true        | Checks if input schema and table schema are same  |
-
+| isolation-level | null | Desired isolation level for transaction.  For overwrites, `serializable` checks for concurrent inserts or deletes in destination partitions, `snapshot` checks for concurrent deletes. |
```
RussellSpitzer, line 188: I think it's unclear what this applies to in the context here. Does it only apply to merges? All Operations? What does "null" mean?

szehon-ho:
Updated docs

RussellSpitzer (approved):

docs/spark/spark-configuration.md
```diff
@@ -185,4 +185,5 @@ df.write
 | snapshot-property._custom-key_    | null            | Adds an entry with custom-key and corresponding value in the snapshot summary  |
 | fanout-enabled       | false        | Overrides this table's write.spark.fanout.enabled  |
 | check-ordering       | true        | Checks if input schema and table schema are same  |
-
+| isolation-level | null | Desired isolation level for Dataframe overwrite operations.  Null => no checks, `serializable` => check for concurrent inserts or deletes in destination partitions, `snapshot` => checks for concurrent deletes in destination partitions. |
+| validate-from-snapshot-id | null | If isolation level is set, identifies base snapshot from which to check concurrent conflicts. If null, check from the first un-expired snapshot.  Obtain via `Spark3Util.loadIcebergTable().currentSnapshot().snapshotId()` |
```
rdblue, line 189: I don't think that `Spark3Util` is intended for public consumption. Plus, what table does this load?

docs/spark/spark-configuration.md
```diff
@@ -185,4 +185,5 @@ df.write
 | snapshot-property._custom-key_    | null            | Adds an entry with custom-key and corresponding value in the snapshot summary  |
 | fanout-enabled       | false        | Overrides this table's write.spark.fanout.enabled  |
 | check-ordering       | true        | Checks if input schema and table schema are same  |
-
+| isolation-level | null | Desired isolation level for Dataframe overwrite operations.  Null => no checks, `serializable` => check for concurrent inserts or deletes in destination partitions, `snapshot` => checks for concurrent deletes in destination partitions. |
```
rdblue, line 188: I think of the `null` case as useful for when the write is idempotent. For example, if you're consuming changed data from an upstream table hourly and writing a complete agg, then the change is idempotent and it doesn't matter what else happens to the partition (for dynamic overwrite).

Maybe there is a good way to state that use case for the "no checks" case?

docs/spark/spark-configuration.md
RussellSpitzer: I've definitely been recommending that folks use this method to get table references from Spark, do you think we should be making the function private? For example when using the programmatic Action API, Spark3Util.loadIcebergTable is much less error prone than creating the Catalog class (like HiveCatalog) directly and loading the table that way.

docs/spark/spark-configuration.md
szehon-ho: Sure, ```Null => no checks (idempotent writes)```?

docs/spark/spark-configuration.md
szehon-ho: Yea internally we have been recommending this API, due to issues in using tables loaded by other catalogs like:  https://github.com/apache/iceberg/issues/4192, https://github.com/apache/iceberg/issues/4086.

Somehow loading a catalog with CatalogUtil giving an arbitrary name not matching the one configured in Spark, causes issue in Spark table actions like RewriteDataFiles that need the right catalog name.  

Though it should not affect this case.

rdblue: I think the main thing is that we don't guarantee anything about this API.

We should really document what people can rely on with guarantees!

docs/spark/spark-configuration.md
rdblue: Yes? Do you think that's clear? Maybe "writes should be idempotent" or "for idempotent writes"?

docs/spark/spark-configuration.md
RussellSpitzer: That's true, maybe we should get some Annotations? Or separate functions we consider to be for end users into a separate file?

docs/spark/spark-configuration.md
szehon-ho: Sounds good, let me open another small pr for this

docs/spark/spark-configuration.md
rdblue: We were just talking about this leading up to discussion at the community sync. In my experience, annotations tend to be a (1) lot of work, (2) out of date, and (3) ignored because they're almost always inaccurate.

I think we should consider the API module stable and make semver guarantees about it, and not point end users to classes in any of the runtime Jars.

That still leaves a problem: what do we do about the classes in core that are used by engines to integrate? I think we probably want to learn more about what's there and have a reasonable policy, like we will remove things there but will go through a deprecation cycle over a release or two first.

rdblue: For this issue specifically, I think we should add a way to do common things in different environments. If Spark is present, then we should have some `Actions.get()` class that uses the current Spark context and `Catalogs.get(String name)` that gets the right catalog from Spark.

PR #4304, opened 2022-03-09
Core: ValidationException on metadata-only delete of certain delete files

szehon-ho:
The presence of some V2 delete files lead to following error for a subsequent delete that tries to use metadata.

```
Failed to cleanly delete data files matching: ref(name="id") == 1
java.lang.IllegalArgumentException: Failed to cleanly delete data files matching: ref(name="id") == 1
	at org.apache.iceberg.spark.source.SparkTable.deleteWhere(SparkTable.java:282)
	at org.apache.spark.sql.execution.datasources.v2.DeleteFromTableExec.run(DeleteFromTableExec.scala:31)
	at org.apache.spark.sql.execution.datasources.v2.V2CommandExec.result$lzycompute(V2CommandExec.scala:43)
	at org.apache.spark.sql.execution.datasources.v2.V2CommandExec.result(V2CommandExec.scala:43)
	at org.apache.spark.sql.execution.datasources.v2.V2CommandExec.executeCollect(V2CommandExec.scala:49)
	at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.$anonfun$applyOrElse$1(QueryExecution.scala:110)
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$5(SQLExecution.scala:103)
	at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:163)
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$1(SQLExecution.scala:90)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
	at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:64)
	at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.applyOrElse(QueryExecution.scala:110)
	at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.applyOrElse(QueryExecution.scala:106)
	at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$transformDownWithPruning$1(TreeNode.scala:481)
	at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:82)
	at org.apache.spark.sql.catalyst.trees.TreeNode.transformDownWithPruning(TreeNode.scala:481)
[...]

PR #4370, opened 2022-03-21
Core: Make removal of delete files optional in manifest filtering

aokolnychyi:
This PR is a follow-up to #4304. In that PR, we updated the method that checks whether a manifest has an entry to be deleted. However, we did not update the method that actually does the manifest filtering. As a consequence, the test added in this PR would fail with a runtime exception as one delete file should be removed and one kept. In addition, we should not introduce any extra calls to `evaluator` as they are expensive.

Overall, the problem is that whenever we overwrite data by filter, we should not fail operations when it is not possible to remove delete files. Removal of delete files is optional and is an optimization we should perform if possible.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
@@ -369,11 +369,11 @@ abstract class ManifestFilterManager<F extends ContentFile<F>> {
 
         hasDeletedFiles = allRowsMatch;
 
-        if (failAnyDelete) {
-          throw new DeleteException(reader.spec().partitionToPath(file.partition()));
-        }
-
         if (hasDeletedFiles) {
+          if (failAnyDelete) {
+            throw new DeleteException(reader.spec().partitionToPath(file.partition()));
+          }
+
           // as soon as a deleted file is detected, stop scanning
           break;
         }
--- a/core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
@@ -350,10 +350,10 @@ abstract class ManifestFilterManager<F extends ContentFile<F>> {
     return canContainExpressionDeletes || canContainDroppedPartitions || canContainDroppedFiles || canContainDropBySeq;
   }
 
-  @SuppressWarnings({"CollectionUndefinedEquality", "checkstyle:CyclomaticComplexity"})
+  @SuppressWarnings("CollectionUndefinedEquality")
   private boolean manifestHasDeletedFiles(PartitionAndMetricsEvaluator evaluator, ManifestReader<F> reader) {
     boolean isDelete = reader.isDeleteManifestReader();
-    boolean hasDeletedFiles = false;
+
     for (ManifestEntry<F> entry : reader.liveEntries()) {
       F file = entry.file();
       boolean markedForDelete = deletePaths.contains(file.path()) ||
@@ -367,19 +367,18 @@ abstract class ManifestFilterManager<F extends ContentFile<F>> {
             "Cannot delete file where some, but not all, rows match filter %s: %s",
             this.deleteExpression, file.path());
 
-        hasDeletedFiles = allRowsMatch;
-
-        if (hasDeletedFiles) {
+        if (allRowsMatch) {
           if (failAnyDelete) {
             throw new DeleteException(reader.spec().partitionToPath(file.partition()));
           }
 
           // as soon as a deleted file is detected, stop scanning
-          break;
+          return true;
         }
       }
     }
-    return hasDeletedFiles;
+
+    return false;
   }
 
   @SuppressWarnings({"CollectionUndefinedEquality", "checkstyle:CyclomaticComplexity"})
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
```diff
@@ -360,23 +360,23 @@ private boolean manifestHasDeletedFiles(PartitionAndMetricsEvaluator evaluator,
           dropPartitions.contains(file.specId(), file.partition()) ||
           (isDelete && entry.sequenceNumber() > 0 && entry.sequenceNumber() < minSequenceNumber);
 
-      boolean nonMatchingDeleteFile = !file.content().equals(FileContent.DATA) && !evaluator.rowsMustMatch(file);
```
aokolnychyi, line 363: We should avoid extra calls to `evaluator` if possible. Also, there was `isDelete` var already defined.

core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
```diff
@@ -396,29 +396,33 @@ private ManifestFile filterManifestWithDeletedFiles(PartitionAndMetricsEvaluator
       try {
         reader.entries().forEach(entry -> {
           F file = entry.file();
-          boolean fileDelete = deletePaths.contains(file.path()) ||
+          boolean markedForDelete = deletePaths.contains(file.path()) ||
```
aokolnychyi, line 399: Renamed for consistency.

core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
```diff
@@ -396,29 +396,33 @@ private ManifestFile filterManifestWithDeletedFiles(PartitionAndMetricsEvaluator
...
 
-              writer.delete(entry);
-
-              CharSequenceWrapper wrapper = CharSequenceWrapper.wrap(entry.file().path());
-              if (deletedPaths.contains(wrapper)) {
-                LOG.warn("Deleting a duplicate path from manifest {}: {}",
-                    manifest.path(), wrapper.get());
-                duplicateDeleteCount += 1;
+              if (allRowsMatch) {
+                writer.delete(entry);
```
aokolnychyi, line 411: Simply moved the existing logic under `if` now.

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
```diff
@@ -995,4 +997,35 @@ public void testOverwriteCaseSensitivity() {
...
+        .overwriteByRowFilter(EXPRESSION_DAY_2_ANOTHER_ID_RANGE)
+        .addFile(FILE_DAY_2_MODIFIED)
+        .validateFromSnapshot(baseSnapshot.snapshotId())
+        .conflictDetectionFilter(EXPRESSION_DAY_2_ANOTHER_ID_RANGE)
+        .validateNoConflictingData()
+        .validateNoConflictingDeletes()
+        .commit();
+
+    validateTableFiles(table, FILE_DAY_2, FILE_DAY_2_MODIFIED);
+    validateTableDeleteFiles(table, FILE_DAY_2_POS_DELETES);
```
aokolnychyi, line 1029: This would previously fail as we should remove one delete file and one should be kept.

core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
```diff
@@ -360,23 +360,23 @@ private boolean manifestHasDeletedFiles(PartitionAndMetricsEvaluator evaluator,
...
-            markedForDelete || evaluator.rowsMustMatch(file),
+            allRowsMatch || isDelete, // ignore delete files where some records may not match the expression
             "Cannot delete file where some, but not all, rows match filter %s: %s",
             this.deleteExpression, file.path());
 
-        hasDeletedFiles = true;
+        hasDeletedFiles = allRowsMatch;
+
         if (failAnyDelete) {
           throw new DeleteException(reader.spec().partitionToPath(file.partition()));
```
szehon-ho, line 373: In failAnyDelete case, this would  trigger if we have a deleteFile but rows do not match.  Probably this never happens in delete call, but wonder if we need to update the condition (failAnyDelete && allRowsMatch).

szehon-ho (approved):
Thanks for this fix.  It makes sense, you are right on the optimization tip and the case there are multiple delete files in the manifest that have this condition, sorry I did not think of it earlier.

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
```diff
@@ -995,4 +997,35 @@ public void testOverwriteCaseSensitivity() {
...
+        .addDeletes(FILE_DAY_2_POS_DELETES)
+        .addDeletes(FILE_DAY_2_ANOTHER_RANGE_EQ_DELETES)
+        .commit();
+
+    Snapshot baseSnapshot = table.currentSnapshot();
+
+    table.newOverwrite()
+        .overwriteByRowFilter(EXPRESSION_DAY_2_ANOTHER_ID_RANGE)
+        .addFile(FILE_DAY_2_MODIFIED)
+        .validateFromSnapshot(baseSnapshot.snapshotId())
```
szehon-ho, line 1022: Optional: this validate from snapshot seems to be a bit useless in reproducing the problem, especially as there should be no other transactions after this base here, would it be better to remove all the validateXX flags to simplify the test?

core/src/main/java/org/apache/iceberg/ManifestFilterManager.java
aokolnychyi: Good catch. I'll fix.

core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java
aokolnychyi: Good call, will simplify.

szehon-ho (approved):
Yep, new changes look good to me.

aokolnychyi:
Thanks, @szehon-ho!

PR #4371, opened 2022-03-21
AWS: Add support for using specified HTTP client

xiaoxuandev:


aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -208,6 +209,14 @@
    */
   public static final String CLIENT_FACTORY = "client.factory";
 
+  /**
+   * The implementation class of {@link software.amazon.awssdk.http.SdkHttpClient} used by {@link AwsClientFactory}
+   * If set, all AWS clients will use this specified HTTP client.
+   * If not set, CLIENT_HTTP_TYPE_DEFAULT will be used.
+   */
+  public static final String CLIENT_HTTP_TYPE = "client.http.type";
+  public static final String CLIENT_HTTP_TYPE_DEFAULT = UrlConnectionHttpClient.class.getName();
```
jackye1995, line 218: we need to use string here, otherwise it will force loading the UrlConnectionHttpClient class.

jackye1995: also because we are talking about "type" but not "impl" here, prefer to use shorter strings instead of full class name. In this case, 2 types could be "url" and "apache"

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -118,6 +121,18 @@ public void initialize(Map<String, String> properties) {
       ValidationException.check((s3AccessKeyId == null && s3SecretAccessKey == null) ||
           (s3AccessKeyId != null && s3SecretAccessKey != null),
           "S3 client access key ID and secret access key must be set at the same time");
+      this.clientHttpType = PropertyUtil.propertyAsString(properties,
+          AwsProperties.CLIENT_HTTP_TYPE, AwsProperties.CLIENT_HTTP_TYPE_DEFAULT);
+    }
+  }
+
+  static SdkHttpClient.Builder getHttpClientBuilder(String clientHttpType) {
```
jackye1995, line 129: prefer more specific verb for method instead of `get`, maybe `configureHttpClientBuilder`?

aws/src/main/java/org/apache/iceberg/aws/AssumeRoleAwsClientFactory.java
```diff
@@ -97,12 +99,13 @@ public void initialize(Map<String, String> properties) {
 
     clientBuilder.credentialsProvider(
         StsAssumeRoleCredentialsProvider.builder()
-            .stsClient(StsClient.builder().httpClientBuilder(UrlConnectionHttpClient.builder()).build())
+            .stsClient(StsClient.builder().httpClientBuilder(
```
jackye1995, line 102: I think we should have a private method for `sts()` similar to what you have for `lakeformation()`, it makes the code more clear about all the AWS clients

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
xiaoxuandev: fixed. thanks!

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -38,6 +40,8 @@
 
 public class AwsClientFactories {
 
+  public static final String APACHE_HTTP_TYPE = "apache";
```
jackye1995, line 43: this should be in `AwsProperties`, with variables `CLIENT_HTTP_TYPE_APACHE`, `CLIENT_HTTP_TYPE_URL`, `CLIENT_HTTP_TYPE_DEFAULT`

jackye1995 (approved):
looks good to me, thanks for adding this!

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -208,6 +208,27 @@
    */
   public static final String CLIENT_FACTORY = "client.factory";
 
+  /**
+   * The type of {@link software.amazon.awssdk.http.SdkHttpClient} implementation used by {@link AwsClientFactory}
+   * If set, all AWS clients will use this specified HTTP client.
+   * If not set, CLIENT_HTTP_TYPE_DEFAULT will be used.
+   * For specific types supported, see CLIENT_HTTP_TYPE_* defined below.
+   */
+  public static final String CLIENT_HTTP_TYPE = "client.http.type";
```
jackye1995, line 217: I am a bit hesitated about this config name (although I proposed myself...), maybe we can have something better, @xiaoxuandev @amogh-jahagirdar @rajarshisarkar @singhpk234 if there is any better suggestion let us know!

amogh-jahagirdar: What seems more natural to me is http.client.type?

jackye1995: I think the issue I was having with the original name was that it's a "client type", but there are 2 "client"s in the config name. `http.client.type` sounds good to me, it also matches the AWS class path for these clients, for example `software.amazon.awssdk.http.apache`. We don't really need the top level `client.` config, because it is configuring one level deeper.

I also thought about the potential configs in the future we want to add, such as connection and socket timeout, common configs can be after the same prefix, such as `http.client.connection-timeout`, and client specific ones can be for example `http.apache.connection-acquisition-timeout`.

@xiaoxuandev what do you think?

xiaoxuandev: Are we going to have different sublevel other than client under http? If no, maybe we should use http-client.type? If yes, then http.client.type looks fine to me.

jackye1995: yeah `http-client.type` sounds good, we can go with that, the `.client` part is optional for shared HTTP configs, so we can have:
- http-client.type
- http-client.connection-timeout
- http-client.apache.connection-acquisition-timeout

what do you think @amogh-jahagirdar

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -208,6 +208,27 @@
...
+
+  /**
+   * If this is set under CLIENT_HTTP_TYPE, {@link software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient}
+   * will be used as the HTTP Client in {@link AwsClientFactory}
+   */
+  public static final String CLIENT_HTTP_TYPE_URL = "url";
```
jackye1995, line 223: based on the AWSSDK class path, I think we should name this type as `urlconnection` instead of just `url`

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
amogh-jahagirdar: I am good with http-client.*!

rajarshisarkar: I am good with `http-client.*` as well.

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -118,6 +122,22 @@ public void initialize(Map<String, String> properties) {
...
           (s3AccessKeyId != null && s3SecretAccessKey != null),
           "S3 client access key ID and secret access key must be set at the same time");
+      this.clientHttpType = PropertyUtil.propertyAsString(properties,
+          AwsProperties.CLIENT_HTTP_TYPE, AwsProperties.CLIENT_HTTP_TYPE_DEFAULT);
+    }
+  }
+
+  static SdkHttpClient.Builder configureHttpClientBuilder(String clientHttpType) {
+    if (Strings.isNullOrEmpty(clientHttpType)) {
+      return  UrlConnectionHttpClient.builder();
```
singhpk234, line 132: should we return what is present in default `AwsProperties.CLIENT_HTTP_TYPE_DEFAULT` here rather than Url client always ?

xiaoxuandev: good point! fixed. thanks.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -245,6 +245,27 @@
    */
   public static final String CLIENT_ASSUME_ROLE_REGION = "client.assume-role.region";
 
+  /**
+   * The type of {@link software.amazon.awssdk.http.SdkHttpClient} implementation used by {@link AwsClientFactory}
+   * If set, all AWS clients will use this specified HTTP client.
+   * If not set, HTTP_CLIENT_TYPE_DEFAULT will be used.
```
jackye1995, line 251: nit: `{@link #HTTP_CLIENT_TYPE_DEFAULT}`

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -245,6 +245,27 @@
...
+   * For specific types supported, see HTTP_CLIENT_TYPE_* defined below.
+   */
+  public static final String HTTP_CLIENT_TYPE = "http-client.type";
+
+  /**
+   * If this is set under HTTP_CLIENT_TYPE, {@link software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient}
```
jackye1995, line 257: nit: similar comment, need `@link`

jackye1995:
Thanks for adding the support!

PR #4374, opened 2022-03-22
Core：fix unnecessarily qualified

felixYyu:
this PR fix Types should not be qualified if they are also imported.
```
iceberg\core\src\main\java\org\apache\iceberg\avro\AvroSchemaUtil.java:380: [UnnecessarilyQualified] Types should not be qualified if they are also imported
    Preconditions.checkArgument(array.getType() == org.apache.avro.Schema.Type.ARRAY,
                                                                  ^
    (see https://github.com/palantir/gradle-baseline#baseline-error-prone-checks)
  Did you mean 'Preconditions.checkArgument(array.getType() == Schema.Type.ARRAY,'?
```


felixYyu:
can you take a look at it, thanks.  cc @rdblue @RussellSpitzer

rdblue:
Thanks, @felixYyu!

PR #4377, opened 2022-03-23
Spark: Add option to introduce ordering of RewriteFileGroup

rajarshisarkar:
This PR introduces ordering of `RewriteFileGroup`. This should be helpful in cases where we want to force the compaction order according to the size (in bytes) or number of files.

Introduce a new Spark option `rewrite.job-order = {"bytes-asc", "bytes-desc", "files-asc", "files-desc", "none"}`.

- `rewrite.job-order=bytes-asc`: rewrite the smallest job groups first.
- `rewrite.job-order=bytes-desc`: rewrite the largest job groups first.
- `rewrite.job-order=files-asc`: rewrite the job groups with the least files first.
- `rewrite.job-order=files-desc`: rewrite the job groups with the most files first.
- `rewrite.job-order=none`: rewrite job groups in the order they were planned (no specific ordering).

---

cc: @RussellSpitzer

### DIFF
--- a/api/src/main/java/org/apache/iceberg/RewriteJobOrder.java
+++ b/api/src/main/java/org/apache/iceberg/RewriteJobOrder.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.util.Locale;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/**
+ * Enum of supported rewrite job order, it defines the order in which the file groups should be
+ * written.
+ * <p><ul>
+ * <li> bytes-asc: rewrite the smallest job groups first.
+ * <li> bytes-desc: rewrite the largest job groups first.
+ * <li> files-asc: rewrite the job groups with the least files first.
+ * <li> files-desc: rewrite the job groups with the most files first.
+ * <li> none: rewrite job groups in the order they were planned (no specific ordering).
+ * </ul><p>
+ */
+public enum RewriteJobOrder {
+  BYTES_ASC("bytes-asc"), BYTES_DESC("bytes-desc"),
+  FILES_ASC("files-asc"), FILES_DESC("files-desc"), NONE("none");
+
+  private final String orderName;
+
+  RewriteJobOrder(String orderName) {
+    this.orderName = orderName;
+  }
+
+  public String orderName() {
+    return orderName;
+  }
+
+  public static RewriteJobOrder fromName(String orderName) {
+    Preconditions.checkArgument(orderName != null, "Rewrite job order name should not be null");
+    // Replace the hyphen in order name with underscore to map to the enum value. For example: bytes-asc to BYTES_ASC
+    return RewriteJobOrder.valueOf(orderName.replaceFirst("-", "_").toUpperCase(Locale.ENGLISH));
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
+++ b/api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
@@ -20,6 +20,7 @@
 package org.apache.iceberg.actions;
 
 import java.util.List;
+import org.apache.iceberg.RewriteJobOrder;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.expressions.Expression;
@@ -87,6 +88,21 @@ public interface RewriteDataFiles extends SnapshotUpdate<RewriteDataFiles, Rewri
    */
   String USE_STARTING_SEQUENCE_NUMBER = "use-starting-sequence-number";
   boolean USE_STARTING_SEQUENCE_NUMBER_DEFAULT = true;
+
+  /**
+   * Forces the rewrite job order based on the value.
+   * <p><ul>
+   * <li> If rewrite.job-order=bytes-asc, then rewrite the smallest job groups first.
+   * <li> If rewrite.job-order=bytes-desc, then rewrite the largest job groups first.
+   * <li> If rewrite.job-order=files-asc, then rewrite the job groups with the least files first.
+   * <li> If rewrite.job-order=files-desc, then rewrite the job groups with the most files first.
+   * <li> If rewrite.job-order=none, then rewrite job groups in the order they were planned (no
+   * specific ordering).
+   * </ul><p>
+   * Defaults to none.
+   */
+  String REWRITE_JOB_ORDER = "rewrite.job-order";
+  String REWRITE_JOB_ORDER_DEFAULT = RewriteJobOrder.NONE.orderName();
 
   /**
    * Choose BINPACK as a strategy for this rewrite operation
--- a/core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
+++ b/core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
@@ -77,4 +77,12 @@ public class RewriteFileGroup {
         .add("numAddedFiles", addedFiles == null ? "Rewrite Incomplete" : Integer.toString(addedFiles.size()))
         .toString();
   }
+
+  public long sizeInBytes() {
+    return fileScanTasks.stream().mapToLong(FileScanTask::length).sum();
+  }
+
+  public int numFiles() {
+    return fileScanTasks.size();
+  }
 }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
@@ -22,6 +22,7 @@ package org.apache.iceberg.spark.actions;
 import java.io.IOException;
 import java.math.RoundingMode;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -34,6 +35,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.RewriteJobOrder;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
@@ -80,7 +82,8 @@ public class BaseRewriteDataFilesSparkAction
       PARTIAL_PROGRESS_ENABLED,
       PARTIAL_PROGRESS_MAX_COMMITS,
       TARGET_FILE_SIZE_BYTES,
-      USE_STARTING_SEQUENCE_NUMBER
+      USE_STARTING_SEQUENCE_NUMBER,
+      REWRITE_JOB_ORDER
   );
 
   private final Table table;
@@ -90,6 +93,7 @@ public class BaseRewriteDataFilesSparkAction
   private int maxCommits;
   private boolean partialProgressEnabled;
   private boolean useStartingSequenceNumber;
+  private RewriteJobOrder rewriteJobOrder;
   private RewriteStrategy strategy = null;
 
   protected BaseRewriteDataFilesSparkAction(SparkSession spark, Table table) {
@@ -146,7 +150,6 @@ public class BaseRewriteDataFilesSparkAction
     }
 
     validateAndInitOptions();
-    strategy = strategy.options(options());
 
     Map<StructLike, List<List<FileScanTask>>> fileGroupsByPartition = planFileGroups(startingSnapshotId);
     RewriteExecutionContext ctx = new RewriteExecutionContext(fileGroupsByPartition);
@@ -166,7 +169,7 @@ public class BaseRewriteDataFilesSparkAction
     }
   }
 
-  private Map<StructLike, List<List<FileScanTask>>> planFileGroups(long startingSnapshotId) {
+  Map<StructLike, List<List<FileScanTask>>> planFileGroups(long startingSnapshotId) {
     CloseableIterable<FileScanTask> fileScanTasks = table.newScan()
         .useSnapshot(startingSnapshotId)
         .filter(filter)
@@ -328,11 +331,9 @@ public class BaseRewriteDataFilesSparkAction
     return new BaseRewriteDataFilesResult(rewriteResults);
   }
 
-  private Stream<RewriteFileGroup> toGroupStream(RewriteExecutionContext ctx,
-                                                 Map<StructLike, List<List<FileScanTask>>> fileGroupsByPartition) {
-
-    // Todo Add intelligence to the order in which we do rewrites instead of just using partition order
-    return fileGroupsByPartition.entrySet().stream()
+  Stream<RewriteFileGroup> toGroupStream(RewriteExecutionContext ctx,
+      Map<StructLike, List<List<FileScanTask>>> fileGroupsByPartition) {
+    Stream<RewriteFileGroup> rewriteFileGroupStream = fileGroupsByPartition.entrySet().stream()
         .flatMap(e -> {
           StructLike partition = e.getKey();
           List<List<FileScanTask>> fileGroups = e.getValue();
@@ -343,9 +344,26 @@ public class BaseRewriteDataFilesSparkAction
             return new RewriteFileGroup(info, tasks);
           });
         });
+
+    return rewriteFileGroupStream.sorted(rewriteGroupComparator());
   }
 
-  private void validateAndInitOptions() {
+  private Comparator<RewriteFileGroup> rewriteGroupComparator() {
+    switch (rewriteJobOrder) {
+      case BYTES_ASC:
+        return Comparator.comparing(RewriteFileGroup::sizeInBytes);
+      case BYTES_DESC:
+        return Comparator.comparing(RewriteFileGroup::sizeInBytes, Comparator.reverseOrder());
+      case FILES_ASC:
+        return Comparator.comparing(RewriteFileGroup::numFiles);
+      case FILES_DESC:
+        return Comparator.comparing(RewriteFileGroup::numFiles, Comparator.reverseOrder());
+      default:
+        return (fileGroupOne, fileGroupTwo) -> 0;
+    }
+  }
+
+  void validateAndInitOptions() {
     Set<String> validOptions = Sets.newHashSet(strategy.validOptions());
     validOptions.addAll(VALID_OPTIONS);
 
@@ -355,6 +373,8 @@ public class BaseRewriteDataFilesSparkAction
     Preconditions.checkArgument(invalidKeys.isEmpty(),
         "Cannot use options %s, they are not supported by the action or the strategy %s",
         invalidKeys, strategy.name());
+
+    strategy = strategy.options(options());
 
     maxConcurrentFileGroupRewrites = PropertyUtil.propertyAsInt(options(),
         MAX_CONCURRENT_FILE_GROUP_REWRITES,
@@ -371,6 +391,10 @@ public class BaseRewriteDataFilesSparkAction
     useStartingSequenceNumber = PropertyUtil.propertyAsBoolean(options(),
         USE_STARTING_SEQUENCE_NUMBER,
         USE_STARTING_SEQUENCE_NUMBER_DEFAULT);
+
+    rewriteJobOrder = RewriteJobOrder.fromName(PropertyUtil.propertyAsString(options(),
+        REWRITE_JOB_ORDER,
+        REWRITE_JOB_ORDER_DEFAULT));
 
     Preconditions.checkArgument(maxConcurrentFileGroupRewrites >= 1,
         "Cannot set %s to %s, the value must be positive.",
--- a/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
+++ b/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
@@ -41,6 +41,7 @@ import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.MetadataTableType;
 import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.RewriteJobOrder;
 import org.apache.iceberg.RowDelta;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SortOrder;
@@ -76,6 +77,7 @@ import org.apache.iceberg.spark.FileRewriteCoordinator;
 import org.apache.iceberg.spark.FileScanTaskSetManager;
 import org.apache.iceberg.spark.SparkTableUtil;
 import org.apache.iceberg.spark.SparkTestBase;
+import org.apache.iceberg.spark.actions.BaseRewriteDataFilesSparkAction.RewriteExecutionContext;
 import org.apache.iceberg.spark.source.ThreeColumnRecord;
 import org.apache.iceberg.types.Comparators;
 import org.apache.iceberg.types.Conversions;
@@ -807,6 +809,12 @@ public class TestRewriteDataFilesAction extends SparkTestBase {
         () -> basicRewrite(table)
             .option("foobarity", "-5")
             .execute());
+
+    AssertHelpers.assertThrows("Cannot set rewrite.job-order to foo",
+        IllegalArgumentException.class,
+        () -> basicRewrite(table)
+            .option(RewriteDataFiles.REWRITE_JOB_ORDER, "foo")
+            .execute());
   }
 
   @Test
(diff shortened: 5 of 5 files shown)


api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
```diff
@@ -88,6 +88,15 @@
   String USE_STARTING_SEQUENCE_NUMBER = "use-starting-sequence-number";
   boolean USE_STARTING_SEQUENCE_NUMBER_DEFAULT = true;
 
+  /**
+   * Forces the compaction order according to the partition size in decreasing order, instead of
```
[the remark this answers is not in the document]
RussellSpitzer: This is a bit internal, but do we mean per fileGroup here as well? For example if a partition has 2 file groups, 100GB and 2 gb, do they both go before a partition with 1 file group of (50 gb)?

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
```diff
@@ -77,4 +78,9 @@ public String toString() {
         .add("numAddedFiles", addedFiles == null ? "Rewrite Incomplete" : Integer.toString(addedFiles.size()))
         .toString();
   }
+
+  @Override
+  public int compareTo(@Nonnull final RewriteFileGroup o) {
```
RussellSpitzer, line 83: this doesn't seem like it is required and i'm a bit dubious about having all rewriteFileGroups treated as equal. The Comparator approach below seems like the right thing to do

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -339,19 +342,25 @@ private Result doExecuteWithPartialProgress(RewriteExecutionContext ctx, Stream<
 
   private Stream<RewriteFileGroup> toGroupStream(RewriteExecutionContext ctx,
                                                  Map<StructLike, List<List<FileScanTask>>> fileGroupsByPartition) {
+    PriorityQueue<RewriteFileGroup> rewriteFileGroups = new PriorityQueue<>(comparator());
```
RussellSpitzer, line 345: Why this over say,

 Stream.sorted(comparator())?

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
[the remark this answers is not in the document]
rajarshisarkar: I would let the users control the comparator by two options: 
- `rewrite-partitions-by-bytes` when enabled by the user should sort the partitions by bytes (higher to lower).
- `rewrite-partitions-by-num-files` when enabled by the user should sort the partitions by number of files in it (higher to lower).
- If nothing is enabled use the natural partition order.
- User cannot enable both the options.

In the above example, 

If `rewrite-partitions-by-bytes` is enabled then file group 1 (100 GB + 2GB = 102 GB) should go before file group 2 (50 GB). 
If `rewrite-partitions-by-num-files` the file group having more files should go before.

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
rajarshisarkar: This was mainly for the case when the user doesn't pass any option and we end up following the natural partition order. I have handled this in `rewriteGroupComparator()` now.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
rajarshisarkar: Thanks, I have made the changes.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -339,19 +342,25 @@ private Result doExecuteWithPartialProgress(RewriteExecutionContext ctx, Stream<
...
+      int globalIndex = ctx.currentGlobalIndex();
+      int partitionIndex = ctx.currentPartitionIndex(partition);
+      FileGroupInfo info = new BaseRewriteDataFilesFileGroupInfo(globalIndex, partitionIndex, partition);
+      rewriteFileGroups.offer(new RewriteFileGroup(info, tasks));
+    }));
+
+    return rewriteFileGroups.stream();
+  }
+
+  private Comparator<RewriteFileGroup> comparator() {
```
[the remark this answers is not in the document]
rajarshisarkar: Thanks, I have renamed it as `rewriteGroupComparator()`

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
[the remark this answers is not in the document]
RussellSpitzer: Don't you think it should be ascending size of bytes? Since the smaller jobs are more likely to finish quickly and allow for interleaving of more jobs and the larger ones are more likely to be stragglers?

I think also maybe it should be something like

job-order: {bytes | files | none/unset }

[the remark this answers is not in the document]
rajarshisarkar: I see, I have made the changes to introduce `rewrite.job-order = {"bytes", "files", "none"}`.

* If `rewrite.job-order=bytes`, then compact based on ascending order of partition size in bytes.
* If `rewrite.job-order=files`, then compact based on descending order of number of files.
* If `rewrite.job-order=none`, then compact based on partition order.

Do you feel if there is a use case for descending order for partition size in bytes / ascending order of number of files? Or maybe some custom option which lets the user to pass any comparator as per their choice?

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
```diff
@@ -88,6 +90,19 @@
   String USE_STARTING_SEQUENCE_NUMBER = "use-starting-sequence-number";
   boolean USE_STARTING_SEQUENCE_NUMBER_DEFAULT = true;
 
+  /**
+   * Forces the compaction order based on the value.
+   * <p>
+   * If rewrite.job-order=bytes, then compact based on ascending order of partition size in bytes.
```
RussellSpitzer, line 96: order of the **job group**

RussellSpitzer: Just because it isn't necessarily a full partition

RussellSpitzer: Maybe to make this clearer,

bytes : Rewrite the smallest job groups first
files: Rewrite the job groups with the most files first
none: Rewrite job groups in the order they were planned (no specific ordering)

RussellSpitzer: When I commented on the size before I wasn't sure which made more sense, do these sort orders make sense for your use case? I was generally thinking about performance and how usually we want to leave possible stragglers till the end but I could be wrong. I'm interested to hear your reasoning and why we should go one way or the other.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -352,6 +353,70 @@ private Result doExecuteWithPartialProgress(RewriteExecutionContext ctx, Stream<
             return new RewriteFileGroup(info, tasks);
           });
         });
+
+    return rewriteFileGroupStream.sorted(rewriteGroupComparator(fileGroupsByPartition));
+  }
+
+  private Comparator<RewriteFileGroup> rewriteGroupComparator(
```
RussellSpitzer, line 360: I think this may be the wrong approach here. There are a few issues

1. We build up a cache of bytes/sizes for every comparisons even though we only use two elements from it
2. We probably need an enum in there for selecting
3. We can probably run without the cache and make the code a bit cleaner.
4. The code currently sorts partitions as a whole, rather than file groups (is this the intention?)

My suggestion would be to have the code look a bit more like
```java
if (rewriteJobOrder.equals(SmallestSizeFirst)) {
  return Comparator.comparing(RewriteFileGroup::size)) ; // Where you implement size in the rewriteFileGroup class or numFIles or whatnot
}
```

Unless our goal here is to sort by Partitions as chunks, but that doesn't make as much sense for me but let me know what you think?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1038,6 +1038,76 @@ public void testInvalidAPIUsage() {
         "Cannot set strategy", () -> actions().rewriteDataFiles(table).sort(SortOrder.unsorted()).binPack());
   }
 
+  @Test
+  public void testRewriteJobOrderBytes() {
+    Table table = createTablePartitioned(4, 2);
```
RussellSpitzer, line 1043: it would be great if these tests checked to see if the ordering actually worked, I think this may be possible through some package private methods which the test class is allowed to access?

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
rajarshisarkar: I think these can be the use cases (I can be convinced otherwise):
- bytes-ascending (avoid larger ones as they are more likely to be stragglers).
- bytes-descending (compact groups with more bytes to gain performance, customer can enable `partial-progress.enabled` and set a low max commit to compact the bulkiest group greedily. This should be done at a maintenance time window where stragglers are not a problem).
- files-descending (intention would be to reduce number of files to gain performance).

I not sure about the files-ascending use case. I feel we still might be missing on many other customer use cases. 

Thought: Iceberg can have some pre-defined rewrite job orders (initial PR scope). Can we think of introducing a custom order where the user gets to pass the comparator at runtime and we load it via reflection?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -352,6 +353,70 @@ private Result doExecuteWithPartialProgress(RewriteExecutionContext ctx, Stream<
...
+      Map<StructLike, List<List<FileScanTask>>> fileGroupsByPartition) {
+    // Ascending order of partition size in bytes
+    if (rewriteJobOrder.equals("bytes")) {
```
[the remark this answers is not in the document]
rajarshisarkar: I have made the change.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
rajarshisarkar: Thanks, for the feedback. I think sorting file groups makes sense.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
rajarshisarkar: I agree, I will add the test cases in the next commit.

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
RussellSpitzer: I would probably just add files ascending as well just for symmetry. I think we should avoid any reflection based solutions. We should keep the API as tight as we can, if there are active use cases for other patterns later we can address them then.

api/src/main/java/org/apache/iceberg/RewriteJobOrder.java
```diff
@@ -0,0 +1,62 @@
...
+    return orderName;
+  }
+
+  public static RewriteJobOrder fromName(String orderName) {
+    Preconditions.checkNotNull(orderName, "Name of rewrite job order should not be null");
+    return RewriteJobOrder.valueOf(orderName.toUpperCase(Locale.ENGLISH));
+  }
+
+  public static boolean isValid(String value) {
+    for (RewriteJobOrder rewriteJobOrder : RewriteJobOrder.values()) {
```
RussellSpitzer, line 54: I'm not sure you need this, if youd did it would probably be sufficent to see if 
"fromName" throws an exception.

RewriteJobOrder.valueOf() will throw a   ```if the specified enum type has no constant with the specified name, or the specified class object does not represent an enum type.```

In general we just call `fromName` and let the illegal arg exception bubble up.

If you wanted to make an isValid method specifically (I don't think you need one here) you could just call 
```java
try {
  fromName(value)
  return true
} catch {
  return false
}

File: api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java (hunk 5 above)
Comment (rajarshisarkar) [reply api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java]: Thanks, for the feedback. I'll add the following job orders: bytes-asc, bytes-desc, files-asc, files-desc, none.

File: api/src/main/java/org/apache/iceberg/RewriteJobOrder.java (hunk 9 above)
Comment (rajarshisarkar) [reply api/src/main/java/org/apache/iceberg/RewriteJobOrder.java]: Yes, it should be possible using `fromName`. I just want to validate value of `rewriteJobOrder` in `validateAndInitOptions()`.

File: spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java (hunk 7 above)
Comment (rajarshisarkar) [reply spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java:1051]: I have changed `planFileGroups()` and `toGroupStream()` to package private methods and have added the tests which checks if the ordering actually worked.

File: api/src/main/java/org/apache/iceberg/RewriteJobOrder.java [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+    this.orderName = orderName;
+  }
+
+  public String orderName() {
+    return orderName;
+  }
+
+  public static RewriteJobOrder fromName(String orderName) {
+    Preconditions.checkNotNull(orderName, "Name of rewrite job order should not be null");
+    if ("bytes-asc".equalsIgnoreCase(orderName)) {
```
RussellSpitzer, line 54: I think this whole thing could `RewriteJobOrder.valueOf(orderName.toLowerCase(Locale.ROOT))`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -352,6 +354,27 @@ private Result doExecuteWithPartialProgress(RewriteExecutionContext ctx, Stream<
             return new RewriteFileGroup(info, tasks);
           });
         });
+
+    return rewriteFileGroupStream.sorted(rewriteGroupComparator());
+  }
+
+  private Comparator<RewriteFileGroup> rewriteGroupComparator() {
+    if (rewriteJobOrder.equals(RewriteJobOrder.BYTES_ASC.orderName())) {
+      // Rewrite the smallest job groups first
```
RussellSpitzer, line 363: docs probably aren't needed here, I think everything will be clear when we change the RewriteFileGroup method names

RussellSpitzer (changes requested):
Left some notes on Enum usage, we can remove some of the if statements by
using built in functions and switch statements.

Tests can probably be simplified a bit to just check the task ordering.

Other than that I think we are general good to go

api/src/main/java/org/apache/iceberg/RewriteJobOrder.java
rajarshisarkar: Actually there is a minor difference between the string user passes: `bytes-asc` and the enum: `BYTES_ASC`. Do notice the difference in the hyphen and underscore. It's something similar to [NullOrder](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SortOrderParser.java#L186-L195) or [RowLevelOperationMode](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/RowLevelOperationMode.java#L43-L52). 

Please let me know your thoughts on how to handle this scenario.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1038,6 +1046,157 @@ public void testInvalidAPIUsage() {
...
+        rewrite.planFileGroups(table.currentSnapshot().snapshotId());
+    Result result = rewrite.execute();
+
+    Assert.assertEquals("Action should rewrite 8 data files", 8, result.rewrittenDataFilesCount());
+    Assert.assertEquals("Action should add 4 data file", 4, result.addedDataFilesCount());
+    shouldHaveFiles(table, 4);
+    List<Object[]> actualRecords = currentData();
+    assertEquals("Rows must match", expectedRecords, actualRecords);
+
+    Stream<RewriteFileGroup> groupStream = rewrite.toGroupStream(
```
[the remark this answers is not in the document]
rajarshisarkar: Thanks for the feedback, I have dropped the previous part and have compared the order of size in bytes/number of files with and without the option.

api/src/main/java/org/apache/iceberg/RewriteJobOrder.java
```diff
@@ -0,0 +1,70 @@
...
+    this.orderName = orderName;
+  }
+
+  public String orderName() {
+    return orderName;
+  }
+
+  public static RewriteJobOrder fromName(String orderName) {
+    Preconditions.checkArgument(orderName != null, "Rewrite job order name should not be null");
+    switch (orderName.toLowerCase(Locale.ROOT)) {
```
RussellSpitzer, line 55: The case statement was for the fix below, this should use the "valueOf" logic

Please again check 
https://github.com/apache/iceberg/blob/c75ac359c1de6bf9fd4894b40009c5c42d2fee9d/api/src/main/java/org/apache/iceberg/DistributionMode.java#L52

rajarshisarkar: Made changes to use valueOf logic.

rajarshisarkar:
```
  public static RewriteJobOrder fromName(String orderName) {
    Preconditions.checkArgument(orderName != null, "Rewrite job order name should not be null");
    // Replace the hyphen in order name with underscore to map to the enum value. For example: bytes-asc to BYTES_ASC
    return RewriteJobOrder.valueOf(orderName.replaceFirst("-", "_").toUpperCase(Locale.ENGLISH));
  }
```

The logic would throw `IllegalArgumentException` in case the user passes some arbitrary value. The same has been asserted in the `testInvalidOptions()` test.

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
```diff
@@ -88,6 +89,20 @@ public interface RewriteDataFiles extends SnapshotUpdate<RewriteDataFiles, Rewri
...
+   * If rewrite.job-order=bytes-asc, then rewrite the smallest job groups first.
+   * If rewrite.job-order=bytes-desc, then rewrite the largest job groups first.
+   * If rewrite.job-order=files-asc, then rewrite the job groups with the least files first.
+   * If rewrite.job-order=files-desc, then rewrite the job groups with the most files first.
+   * If rewrite.job-order=none, then rewrite job groups in the order they were planned (no specific ordering).
+   * <p>
+   * Defaults to none.
+   */
+  String REWRITE_JOB_ORDER = "rewrite.job-order";
+  String REWRITE_JOB_ORDER_DEFAULT = RewriteJobOrder.NONE.orderName();
```
RussellSpitzer, line 104: nit: Default value can be of the enum type instead of string

RussellSpitzer: actually given how you set this it's probably fine. Let's keep it this way

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1038,6 +1046,128 @@ public void testInvalidAPIUsage() {
         "Cannot set strategy", () -> actions().rewriteDataFiles(table).sort(SortOrder.unsorted()).binPack());
   }
 
+  @Test
+  public void testRewriteJobOrderBytesAsc() {
+    Table table = createTablePartitioned(4, 2);
```
RussellSpitzer, line 1051: Within these tests are we sure that they are actually producing row groups with different sizes of bytes and files?

RussellSpitzer: I believe this would end up just making 2 file groups, one for each partition , and they should be the same size? (at least in number of files?)

rajarshisarkar: I have added uneven files under each partition now. I checked using debugger to ensure bytes in size and number of files are different.

rajarshisarkar: For `BYTES_ASC`:

RussellSpitzer: Sounds good to me, Just to make sure I would just have an assert in the test that sorting in both directions is different :) That way any future changes to parquet protocols would be automatically checked. 

Ie Asstert expected == actual && expected.reversed != actual

rajarshisarkar: Makes sense to me. I have pushed the change.

rajarshisarkar:
@RussellSpitzer PR is ready for another round of review.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -356,6 +374,8 @@ private void validateAndInitOptions() {
         "Cannot use options %s, they are not supported by the action or the strategy %s",
         invalidKeys, strategy.name());
 
+    strategy = strategy.options(options());
```
RussellSpitzer, line 377: What is this doing?

RussellSpitzer: ah I see we are moving this into the validate function. That's ok

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1038,6 +1046,152 @@ public void testInvalidAPIUsage() {
...
+                .binPack();
+    List<Long> actual = toGroupStream(table, jobOrderRewrite)
+        .mapToLong(RewriteFileGroup::sizeInBytes)
+        .boxed()
+        .collect(Collectors.toList());
+
+    expected.sort(Comparator.naturalOrder());
+    Assert.assertEquals("Size in bytes order should be ascending", actual, expected);
+    Collections.reverse(expected);
+    Assert.assertNotEquals("Size in bytes order should be ascending", actual, expected);
```
RussellSpitzer, line 1080: This message should be "should not be descending"

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1038,6 +1046,152 @@ public void testInvalidAPIUsage() {
...
+                .binPack();
+    List<Long> actual = toGroupStream(table, jobOrderRewrite)
+        .mapToLong(RewriteFileGroup::sizeInBytes)
+        .boxed()
+        .collect(Collectors.toList());
+
+    expected.sort(Comparator.reverseOrder());
+    Assert.assertEquals("Size in bytes order should be descending", actual, expected);
+    Collections.reverse(expected);
+    Assert.assertNotEquals("Size in bytes order should be descending", actual, expected);
```
RussellSpitzer, line 1114: Should not be ascending

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
rajarshisarkar: Yeah, while calling the validate function from test I saw the strategy options were not being initialised. Hence, moved it here. This helps in getting the different bytes/files for the groups too.

RussellSpitzer (approved):

RussellSpitzer:
Thanks for the patch @rajarshisarkar , Let's do a quick followup to move the changes into 3.0 and 3.1

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
```diff
@@ -88,6 +89,21 @@ public interface RewriteDataFiles extends SnapshotUpdate<RewriteDataFiles, Rewri
...
+   * <li> If rewrite.job-order=bytes-asc, then rewrite the smallest job groups first.
+   * <li> If rewrite.job-order=bytes-desc, then rewrite the largest job groups first.
+   * <li> If rewrite.job-order=files-asc, then rewrite the job groups with the least files first.
+   * <li> If rewrite.job-order=files-desc, then rewrite the job groups with the most files first.
+   * <li> If rewrite.job-order=none, then rewrite job groups in the order they were planned (no
+   * specific ordering).
+   * </ul><p>
+   * Defaults to none.
+   */
+  String REWRITE_JOB_ORDER = "rewrite.job-order";
```
aokolnychyi, line 104: Why add a prefix and not just use `job-order`? Aren't all options rewrite specific?
cc @RussellSpitzer @rajarshisarkar

RussellSpitzer: we could do rewrite-job-order in the backport pr, I think that would be clearest if we want to avoid the sub prop

rajarshisarkar: Sure, I'll incorporate this change in the backport PR.

PR #4379, opened 2022-03-23
Docs: Clarify timestamp in stored procedures

ajantha-bhat:
- Spark supports [microseconds precision](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/types/TimestampType.scala#L26)  but in code we cast to milliseconds, so better to warn the users.
- Some code comment was wrong about spark precision. 
- Remove generic spark expressions for timestamp in expire_snapshots example (as it is not supported)

### DIFF
--- a/docs/spark/spark-procedures.md
+++ b/docs/spark/spark-procedures.md
@@ -105,7 +105,7 @@ This procedure invalidates all cached Spark plans that reference the affected ta
 
 #### Example
 
-Roll back `db.sample` to one day
+Roll back `db.sample` to a specific day and time.
 ```sql
 CALL catalog_name.system.rollback_to_timestamp('db.sample', TIMESTAMP '2021-06-30 00:00:00.000')
 ```
@@ -215,16 +215,10 @@ If `older_than` and `retain_last` are omitted, the table's [expiration propertie
 
 #### Examples
 
-Remove snapshots older than one day, but retain the last 100 snapshots:
+Remove snapshots older than specific day and time, but retain the last 100 snapshots:
 
 ```sql
 CALL hive_prod.system.expire_snapshots('db.sample', TIMESTAMP '2021-06-30 00:00:00.000', 100)
-```
-
-Erase all snapshots older than the current timestamp but retain the last 5 snapshots:
-
-```sql
-CALL hive_prod.system.expire_snapshots(table => 'db.sample', older_than => now(), retain_last => 5)
 ```
 
 ### `remove_orphan_files`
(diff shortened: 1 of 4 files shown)


docs/spark/spark-procedures.md
```diff
@@ -215,18 +215,12 @@ If `older_than` and `retain_last` are omitted, the table's [expiration propertie
...
-Remove snapshots older than one day, but retain the last 100 snapshots:
+Remove snapshots older than one day (assuming today is TIMESTAMP '2021-06-31 00:00:00.000'), but retain the last 100 snapshots:
 
 ```sql
 CALL hive_prod.system.expire_snapshots('db.sample', TIMESTAMP '2021-06-30 00:00:00.000', 100)
 ```
 
-Erase all snapshots older than the current timestamp but retain the last 5 snapshots:
-
-```sql
```
ajantha-bhat, line 226: Removed generic spark expressions for timestamp as it is not supported

ajantha-bhat:
cc: @aokolnychyi , @RussellSpitzer

docs/spark/spark-procedures.md
```diff
@@ -198,7 +198,7 @@ the `expire_snapshots` procedure will never remove files which are still require
 | Argument Name | Required? | Type | Description |
 |---------------|-----------|------|-------------|
 | `table`       | ✔️  | string | Name of the table to update |
-| `older_than`  | ️   | timestamp | Timestamp before which snapshots will be removed (Default: 5 days ago) |
+| `older_than`  | ️   | timestamp | Timestamp before which snapshots will be removed (Default: 5 days ago). Considered upto millisecond precision. |
```
RussellSpitzer, line 201: "up to" - although in general i'm not sure this clarification is needed, do you think this is a confusion among users?

docs/spark/spark-procedures.md
```diff
@@ -105,7 +105,7 @@ This procedure invalidates all cached Spark plans that reference the affected ta
 
 #### Example
 
-Roll back `db.sample` to one day
+Roll back `db.sample` to one day before (assuming today is TIMESTAMP '2021-06-31 00:00:00.000')
```
RussellSpitzer, line 108: I would just change this to "Roll back to a specific day", just to keep it simple.

ajantha-bhat: yeah, I like this suggestion.

docs/spark/spark-procedures.md
ajantha-bhat: I think people assume it is microsecond precision as spark is microsecond precision, but code we are losing the precision (lossy conversion to milli), so I think better to clarify in the docs.

RussellSpitzer: I'm just not sure how often folks would expect their microsecond values to matter in snapshot expiration.

ajantha-bhat: ok, as snapshots itself is having creation time in milliseconds, we cannot expect to have microsecond precision. 
I will remove it.

ajantha-bhat:
@RussellSpitzer : can you please take another look and help merging this PR ?

openinx (approved):
Looks good to me !

PR #4381, opened 2022-03-23
Core: Make DeleteFilter's constructor parameters more specific

alexjo2144:
Limits the constructor arguments for DeleteFilters to just the parts of the FileScanTask actually used. This came up as a part of the Trino work to support merge-on-read tables and reusing DeleteFilter there.

### DIFF
--- a/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
+++ b/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
@@ -25,10 +25,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
 import org.apache.iceberg.Accessor;
-import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DeleteFile;
 import org.apache.iceberg.FileContent;
-import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.MetadataColumns;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.StructLike;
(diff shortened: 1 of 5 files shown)


alexjo2144:
Mind taking a look? @rdblue @danielcweeks @findepi

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -64,7 +62,7 @@
       MetadataColumns.DELETE_FILE_POS);
 
   private final long setFilterThreshold;
-  private final DataFile dataFile;
+  private final CharSequence filePath;
```
findepi, line 65: `CharSequence` doesn't guarantee immutabilty. Let's make it a `String`.

danielcweeks (approved):
LGTM

Zhangg7723:
I want to filter eq delete rows loaded in delete set by lower-upper bounds in data file, so can I replace filePath to dataFile as parameter in the construction? 
DeleteFilter interface is defined in Trino, I think DeleteFilter in Iceberg is never used in Trino, right? @findepi 
@danielcweeks

findepi:
> DeleteFilter interface is defined in Trino, I think DeleteFilter in Iceberg is never used in Trino, right? @findepi

I guess you mean these changes https://github.com/trinodb/trino/pull/13219
David would be the best person to answer questions.
For Trino Iceberg specifics let's maybe talk on `#iceberg` channel on Trino slack

PR #4382, opened 2022-03-23
Core: Predicate pushdown for all_data_files table

szehon-ho:
This adopts all_data_files table on new base class added in #4139 , which will allow it predicate-push-down for partition predicates.

Change the base class  manifests() method slightly from List to Iterable, to support the original ParallelIterable of the AllDataFilesTable.

Inherit some WAP (staging snapshot) handling from BaseAllMetadataTableScan.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
@@ -81,14 +81,13 @@ public class AllDataFilesTable extends BaseFilesTable {
 
     @Override
     public CloseableIterable<FileScanTask> planFiles() {
-      return super.planAllFiles();
+      return super.planFilesAllSnapshots();
     }
 
     @Override
     protected CloseableIterable<ManifestFile> manifests() {
       try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
-          Iterables.transform(table().snapshots(),
-              snapshot -> (Iterable<ManifestFile>) () -> snapshot.dataManifests().iterator()),
+          Iterables.transform(table().snapshots(), snapshot -> () -> snapshot.dataManifests().iterator()),
           context().planExecutor())) {
         return CloseableIterable.withNoopClose(Sets.newHashSet(iterable));
       } catch (IOException e) {
--- a/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
@@ -65,11 +65,6 @@ public class AllDataFilesTable extends BaseFilesTable {
     }
 
     @Override
-    public boolean allScan() {
-      return true;
-    }
-
-    @Override
     protected TableScan newRefinedScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
       return new AllDataFilesTableScan(ops, table, schema, fileSchema(), context);
     }
@@ -82,6 +77,11 @@ public class AllDataFilesTable extends BaseFilesTable {
     @Override
     public TableScan asOfTime(long timestampMillis) {
       throw new UnsupportedOperationException("Cannot select snapshot: all_data_files is for all snapshots");
+    }
+
+    @Override
+    public CloseableIterable<FileScanTask> planFiles() {
+      return super.planAllFiles();
     }
 
     @Override
--- a/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
@@ -80,6 +80,11 @@ public class AllDataFilesTable extends BaseFilesTable {
     }
 
     @Override
+    public CloseableIterable<FileScanTask> planFiles() {
+      return super.planAllFiles();
+    }
+
+    @Override
     protected CloseableIterable<ManifestFile> manifests() {
       try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
           Iterables.transform(table().snapshots(),
--- a/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
@@ -20,18 +20,10 @@
 package org.apache.iceberg;
 
 import java.io.IOException;
-import java.util.List;
-import java.util.concurrent.ExecutorService;
-import org.apache.iceberg.BaseFilesTable.ManifestReadTask;
 import org.apache.iceberg.exceptions.RuntimeIOException;
-import org.apache.iceberg.expressions.Expression;
-import org.apache.iceberg.expressions.Expressions;
-import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
-import org.apache.iceberg.types.TypeUtil;
-import org.apache.iceberg.types.Types.StructType;
 import org.apache.iceberg.util.ParallelIterable;
 
 /**
@@ -41,7 +33,7 @@ import org.apache.iceberg.util.ParallelIterable;
  * <p>
  * This table may return duplicate rows.
  */
-public class AllDataFilesTable extends BaseMetadataTable {
+public class AllDataFilesTable extends BaseFilesTable {
 
   AllDataFilesTable(TableOperations ops, Table table) {
     this(ops, table, table.name() + ".all_data_files");
@@ -57,44 +49,24 @@ public class AllDataFilesTable extends BaseMetadataTable {
   }
 
   @Override
-  public Schema schema() {
-    StructType partitionType = Partitioning.partitionType(table());
-    Schema schema = new Schema(DataFile.getType(partitionType).fields());
-    if (partitionType.fields().size() < 1) {
-      // avoid returning an empty struct, which is not always supported. instead, drop the partition field (id 102)
-      return TypeUtil.selectNot(schema, Sets.newHashSet(102));
-    } else {
-      return schema;
-    }
-  }
-
-  @Override
   MetadataTableType metadataTableType() {
     return MetadataTableType.ALL_DATA_FILES;
   }
 
-  public static class AllDataFilesTableScan extends BaseAllMetadataTableScan {
-    private final Schema fileSchema;
+  public static class AllDataFilesTableScan extends BaseFilesTableScan {
 
     AllDataFilesTableScan(TableOperations ops, Table table, Schema fileSchema) {
-      super(ops, table, fileSchema);
-      this.fileSchema = fileSchema;
+      super(ops, table, fileSchema, MetadataTableType.ALL_DATA_FILES);
     }
 
     private AllDataFilesTableScan(TableOperations ops, Table table, Schema schema, Schema fileSchema,
                                   TableScanContext context) {
-      super(ops, table, schema, context);
-      this.fileSchema = fileSchema;
-    }
-
-    @Override
-    protected String tableType() {
-      return MetadataTableType.ALL_DATA_FILES.name();
+      super(ops, table, schema, fileSchema, context, MetadataTableType.ALL_DATA_FILES);
     }
 
     @Override
     protected TableScan newRefinedScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
-      return new AllDataFilesTableScan(ops, table, schema, fileSchema, context);
+      return new AllDataFilesTableScan(ops, table, schema, fileSchema(), context);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -88,6 +93,18 @@ public TableScan appendsAfter(long fromSnapshotId) {
           String.format("Cannot incrementally scan table of type %s", type.name()));
     }
 
+    @Override
+    public CloseableIterable<FileScanTask> planFiles() {
+      if (type.equals(MetadataTableType.ALL_DATA_FILES)) {
```
szehon-ho, line 98: I had to move this code from BaseAllMetadataTableScan.  

I didn't realize that there was an alternate base class for "All*" metadata tables with special handling (it seems for WAP case).  As there's no multiple inheritance for Java, I can't inherit from both BaseFileTable and BaseAllMetadataTable scans, so I decided to copy the logic.

An alternative is to use composition, let me know if that's a better way, as I thought it's a bit overkill for a few lines.  As a consequence now though, the BaseAllMetadataTableScan supports only 'all_entries' table.

szehon-ho:
The failure may not be related, filed: https://github.com/apache/iceberg/issues/4383, re-triggering to see if it goes away.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Refactored to make it slightly cleaner.

szehon-ho:
Looks like build is broken in general by https://github.com/apache/iceberg/commit/0f6398f1ae6f5fbde5329858e0cd64064ceb165e, will retrigger after.

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
```diff
@@ -38,4 +44,14 @@ public long targetSplitSize() {
         TableProperties.METADATA_SPLIT_SIZE_DEFAULT);
     return PropertyUtil.propertyAsLong(options(), TableProperties.SPLIT_SIZE, tableValue);
   }
+
+  /**
+   * Alternative to {@link #planFiles()}, allows exploring old snapshots even for empty table.
+   */
+  protected CloseableIterable<FileScanTask> planAllFiles() {
```
RussellSpitzer, line 51: Could this be in BaseFilesTableScan? Just wondering if it would be relevant for say "Manifests" tables?

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -110,21 +110,19 @@ public TableScan appendsAfter(long fromSnapshotId) {
     /**
      * @return list of manifest files to explore for this files metadata table scan
```
RussellSpitzer, line 111: Doc is slightly wrong now

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -108,30 +80,20 @@ public TableScan asOfTime(long timestampMillis) {
...
-  private static CloseableIterable<ManifestFile> allDataManifestFiles(
-      List<Snapshot> snapshots, ExecutorService workerPool) {
-    try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
-        Iterables.transform(snapshots, snapshot -> (Iterable<ManifestFile>) () -> snapshot.dataManifests().iterator()),
-        workerPool)) {
-      return CloseableIterable.withNoopClose(Sets.newHashSet(iterable));
-    } catch (IOException e) {
-      throw new RuntimeIOException(e, "Failed to close parallel iterable");
+    @Override
+    protected CloseableIterable<ManifestFile> manifests() {
```
RussellSpitzer, line 88: We change the return type of manifests to CloseableIterable for this table? I just notice all the other implementations just wrap the list. I think this makes sense

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
szehon-ho: Yea this was tricky and I went back and forth with various solutions.  

BaseAllMetadataTableScan used to override planFiles() , and was inherited by AllDataFiles, AllManifests, AllEntries.  

Now I moved AllDataFiles to BaseFilesTableScan and so I don't get this override anymore.

So it has to be called from two different hierarchies (BaseAllMetadataTableScans and BaseFilesTableScan), hence moving to their only common parent to continue sharing the code.

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
szehon-ho: Yea as the AllDataFilesTable used to use a parallel iterable.  It's more flexible I think in case the other Files tables want to use that later.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Will fix

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
szehon-ho: Actually I tried to improve a little bit and made the override at the BaseMetadataTableScan parent class, based on a boolean flag the subclasses configure, let me know if that is better.

RussellSpitzer (approved):
Small note on the "allScan" naming, other than that I think this is good to go

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMetadataTables.java
```diff
@@ -181,6 +181,94 @@ public void testPartitionedTable() throws Exception {
...
+    );
+    spark.createDataset(records, Encoders.bean(SimpleRecord.class))
+        .coalesce(1)
+        .writeTo(tableName)
+        .append();
+
+
+    Table table = Spark3Util.loadIcebergTable(spark, tableName);
+    List<ManifestFile> expectedDataManifests = TestHelpers.dataManifests(table);
+    Assert.assertEquals("Should have 2 data files", 1, expectedDataManifests.size());
```
RussellSpitzer, line 203: "should have 1 data manifest"

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -108,30 +85,15 @@ public TableScan asOfTime(long timestampMillis) {
...
-    try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
-        Iterables.transform(snapshots, snapshot -> (Iterable<ManifestFile>) () -> snapshot.dataManifests().iterator()),
-        workerPool)) {
-      return CloseableIterable.withNoopClose(Sets.newHashSet(iterable));
-    } catch (IOException e) {
-      throw new RuntimeIOException(e, "Failed to close parallel iterable");
+    protected CloseableIterable<ManifestFile> manifests() {
+      try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
+          Iterables.transform(table().snapshots(),
+              snapshot -> (Iterable<ManifestFile>) () -> snapshot.dataManifests().iterator()),
```
aokolnychyi, line 91: nit: I think you can drop the cast (which seems redundant) and move the entire call to `Iterables` on one line.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -108,23 +108,21 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
 
     /**
-     * @return list of manifest files to explore for this files metadata table scan
+     * @return iterable of manifest files to explore for this files metadata table scan
      */
-    protected abstract List<ManifestFile> manifests();
+    protected abstract CloseableIterable<ManifestFile> manifests();
 
-    private CloseableIterable<ManifestFile> filterManifests(List<ManifestFile> manifests,
+    private CloseableIterable<ManifestFile> filterManifests(CloseableIterable<ManifestFile> manifests,
```
aokolnychyi, line 115: nit: not necessarily related but could you fix the arg alignment in `planFiles` above?

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
```diff
@@ -31,11 +37,30 @@ protected BaseMetadataTableScan(TableOperations ops, Table table, Schema schema,
     super(ops, table, schema, context);
   }
 
+  /**
+   * @return if metadata table scan is for all snapshots, ie 'all_x' metadata tables
+   */
+  protected boolean allScan() {
```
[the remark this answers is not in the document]
aokolnychyi: +1, I think `allScan` is not very descriptive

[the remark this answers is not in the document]
aokolnychyi: In fact, what about removing this flag completely and overriding `planFiles` in `BaseAllMetadataTableScan`?

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
```diff
@@ -31,11 +37,30 @@ protected BaseMetadataTableScan(TableOperations ops, Table table, Schema schema,
...
   @Override
   public long targetSplitSize() {
     long tableValue = tableOps().current().propertyAsLong(
         TableProperties.METADATA_SPLIT_SIZE,
         TableProperties.METADATA_SPLIT_SIZE_DEFAULT);
     return PropertyUtil.propertyAsLong(options(), TableProperties.SPLIT_SIZE, tableValue);
   }
+
+  @Override
+  public CloseableIterable<FileScanTask> planFiles() {
```
aokolnychyi, line 56: I think we should override `planFiles` in `BaseAllMetadataTableScan` instead.

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
aokolnychyi: Seems reasonable.

aokolnychyi:
We should fix the predicate pushdown in a separate PR as it only uses the current table spec instead of the spec of a manifest that we filter. The logic I am talking about is in `filterManifests`.

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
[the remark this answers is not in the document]
aokolnychyi: The current implementation of `planFiles` seems a little bit weird with this flag.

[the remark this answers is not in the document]
szehon-ho: @aokolnychyi yea i had that version in the previous iteration, having a method like:

BaseMetadataTableScan.scanAllManifests(); //like planFiles but across all manifests.

and then having:

@Override
BaseAllMetadataTableScan.planFiles() {
   super.scanAllManifests();
}

@Override
AllDataFileTableScan.planFiles() {
  super.scanAllManifests();
}

Is that better than this?  I can revert back to that version.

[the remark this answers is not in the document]
szehon-ho: The reason I changed to this approach , it's slightly more clear in forcing implementators to have to override this flag, rather than have to find this method just by reading the code , but it's not a big deal either way as it's fairly internal.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMetadataTables.java
szehon-ho: Good catch

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
szehon-ho: Yep, just copied from original code, removed.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Done

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
szehon-ho: Actually, this does not work, somehow the compiler used in the build cannot infer the type without this cast (though intellij can), ref:  https://github.com/apache/iceberg/runs/5824575297

szehon-ho:
@aokolnychyi addressed as many comments as I could, can you check when you have time, if its any clearer?

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
aokolnychyi: Oh, then forget about it.

aokolnychyi (approved):
Well, I did not realize we have a little bit weird hierarchy in our metadata tables and ideally want to extend two classes, which is obviously not possible. This isn't a problem of this PR and I don't have an obvious solution. Let's get this in and check if we can maybe use composition rather than inheritance in a few places to simplify the hierarchy of metadata tables and scans.

Issue #4383, opened 2022-03-24
Spark: Random failure in TestCopyOnWriteMerge::testMergeWithSnapshotIsolation


szehon-ho:
On Spark 3, hit this error:

```
org.apache.iceberg.spark.extensions.TestCopyOnWriteMerge > testMergeWithSnapshotIsolation[catalogName = testhive, implementation = org.apache.iceberg.spark.SparkCatalog, config = {type=hive, default-namespace=default}, format = orc, vectorized = true, distributionMode = none] FAILED
    java.util.concurrent.ExecutionException: java.util.ConcurrentModificationException
        at java.util.concurrent.FutureTask.report(FutureTask.java:122)
        at java.util.concurrent.FutureTask.get(FutureTask.java:192)
        at org.apache.iceberg.spark.extensions.TestMerge.testMergeWithSnapshotIsolation(TestMerge.java:720)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
[...]

PR #4384, opened 2022-03-24
ORC : Avoid modifying existing conf of HadoopOutputFile rather create new one

singhpk234:
A probable fix to 

Close https://github.com/apache/iceberg/issues/4383

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/ORC.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORC.java
@@ -162,11 +162,9 @@ public class ORC {
     public <D> FileAppender<D> build() {
       Preconditions.checkNotNull(schema, "Schema is required");
 
-      Configuration conf;
+      Configuration conf = new Configuration();
       if (file instanceof HadoopOutputFile) {
-        conf = ((HadoopOutputFile) file).getConf();
-      } else {
-        conf = new Configuration();
+        ((HadoopOutputFile) file).getConf().forEach(entry -> conf.set(entry.getKey(), entry.getValue()));
       }
 
       for (Map.Entry<String, String> entry : config.entrySet()) {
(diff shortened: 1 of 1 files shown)


orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -162,11 +162,9 @@ private WriteBuilder createContextFunc(Function<Map<String, String>, Context> ne
     public <D> FileAppender<D> build() {
       Preconditions.checkNotNull(schema, "Schema is required");
 
-      Configuration conf;
+      Configuration conf = new Configuration();
       if (file instanceof HadoopOutputFile) {
-        conf = ((HadoopOutputFile) file).getConf();
-      } else {
-        conf = new Configuration();
+        ((HadoopOutputFile) file).getConf().forEach(entry -> conf.set(entry.getKey(), entry.getValue()));
```
openinx, line 167: Seems we will merge the hadoop file's configuration with the default environment configuration by default if it's a `HadoopOutputFile`, right ? 

Should we just use the original code piece ( according to [here](https://github.com/apache/iceberg/pull/4291/files#diff-34d7fce4c1d9417fa9342247cf5ace0636cd86591efbed0555ae80197f79303dL90-L94)):

```java
      if (file instanceof HadoopOutputFile) {
        this.conf = new Configuration(((HadoopOutputFile) file).getConf());
      } else {
        this.conf = new Configuration();
      }
```

singhpk234: makes sense, made the changes

openinx (approved):
Looks good to me, thanks @singhpk234 for the contribution !

singhpk234:
Thanks @openinx , for prompt review :) !!!!

PR #4388, opened 2022-03-24
Core: InvalidateTable in CachingCatalog should invoke InvalidateTable on the wrapped catalog too.

samarthjain:
When there is a chain of catalogs with one catalog wrapped inside the other, then the missing catalog.invalidateTable
call can prevent a table from being invalidated. In case the wrapped catalog is a CachingCatalog, then this prevents the table to be removed from the cache. We need to call catalog.invalidateTable() so that the call can recurse to invoke invalidateTable on all the chained catalogs.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/CachingCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/CachingCatalog.java
@@ -164,22 +164,19 @@ public class CachingCatalog implements Catalog {
   @Override
   public boolean dropTable(TableIdentifier ident, boolean purge) {
     boolean dropped = catalog.dropTable(ident, purge);
-    invalidate(ident);
+    invalidateTable(ident);
     return dropped;
   }
 
   @Override
   public void renameTable(TableIdentifier from, TableIdentifier to) {
     catalog.renameTable(from, to);
-    invalidate(from);
+    invalidateTable(from);
   }
 
   @Override
   public void invalidateTable(TableIdentifier ident) {
-    invalidate(ident);
-  }
-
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/CachingCatalog.java
```diff
@@ -164,22 +164,19 @@ public Table loadTable(TableIdentifier ident) {
...
-    invalidate(from);
+    invalidateTable(from);
   }
 
   @Override
   public void invalidateTable(TableIdentifier ident) {
-    invalidate(ident);
-  }
-
-  private void invalidate(TableIdentifier ident) {
```
samarthjain, line 182: The private `invalidate(ident)` method wasn't doing anything other than calling the public `invalidateTable(ident)` method. I got rid of it.

core/src/test/java/org/apache/iceberg/hadoop/TestCachingCatalog.java
```diff
@@ -324,6 +324,18 @@ public void testCacheExpirationIsDisabledByANegativeValue() throws IOException {
         catalog.isCacheExpirationEnabled());
   }
 
+  @Test
```
samarthjain, line 327: Added a test to verify that all the chained  CachingCatalogs don't have the table present in their cache.

rdblue (approved):

rdblue:
Looks good to me, but isn't it a bit strange to have caching at multiple levels?

rdblue:
Thanks, @samarthjain!

PR #4385, opened 2022-03-24
Nessie: Bump Nessie and add Namespace support

nastra:

PR #4392, opened 2022-03-24
Nessie: Add Catalog tests for NessieCatalog

nastra:
currently depends on #4385 and #4333

### DIFF
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieCatalog.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieCatalog.java
@@ -1,0 +1,162 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.nessie;
+
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.Path;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.catalog.CatalogTests;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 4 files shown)


snazy (commented):
The last commit LGTM, but you will to rebase, so saving my "Approve" for that time ;)

core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java
```diff
@@ -82,9 +81,8 @@
   private static JdbcCatalog catalog;
   private static String warehouseLocation;
 
-  @Rule
-  public TemporaryFolder temp = new TemporaryFolder();
-  File tableDir = null;
+  @TempDir
+  java.nio.file.Path tableDir;
```
snazy, line 85: Import?

rdblue:
@nastra can you rebase this so we can make sure it still works on the current master?

core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java
nastra: required as otherwise it would clash with the org.apache.hadoop.fs.Path import

nastra:
@rdblue rebased the PR. Will rebase again once https://github.com/apache/iceberg/pull/4385 and https://github.com/apache/iceberg/pull/4333 are in

rymurr (approved):

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieCatalog.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Override
+  protected boolean supportsNamespaceProperties() {
+    return false;
+  }
+
+  @Override
+  protected boolean supportsServerSideRetry() {
+    // TODO: we do support retries, but a bunch of tests are currently failing
```
rdblue, line 144: I think you'd have to update to the diff-based commits to make this work. If you end up using the REST catalog API that would be a nice feature to pull in at the same time.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieCatalog.java
```diff
@@ -0,0 +1,162 @@
...
+    return false;
+  }
+
+  @Test
+  @Override
+  @Disabled(
+      "Nessie does not differentiate between table creates & updates, thus a concurrent transaction does not fail")
```
rdblue, line 151: Not even in the same branch?

rdblue:
Thanks for reviewing, @rymurr! Sorry I didn't get to this.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieCatalog.java
nastra: currently no, not even on the same branch. However, it's likely that we add this in an upcoming release of Nessie

rdblue: I'm glad that we're testing to find the differences between catalogs! Can we document this limitation somewhere? If we're disabling a test for transactional behavior then it seems like something we should note for users.

nastra: done in https://github.com/apache/iceberg/pull/4501

PR #4393, opened 2022-03-24
Aliyun: Add metrics tracking for OSSFileIO

rajarshisarkar:
This PR adds metrics tracking for `OSSFileIO`. This would help reporting IO metrics via the Hadoop `FileSystem.Statistics`.

---
cc: @danielcweeks @rdblue @kbendick

### DIFF
--- a/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputFile.java
+++ b/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputFile.java
@@ -24,17 +24,22 @@ import org.apache.iceberg.aliyun.AliyunProperties;
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.SeekableInputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
+/**
+ * @deprecated moving to package-private in 0.15.0; use OSSFileIO to create InputFile instances
+ */
+@Deprecated
 public class OSSInputFile extends BaseOSSFile implements InputFile {
 
   private Long length = null;
 
-  OSSInputFile(OSS client, OSSURI uri, AliyunProperties aliyunProperties) {
-    super(client, uri, aliyunProperties);
+  OSSInputFile(OSS client, OSSURI uri, AliyunProperties aliyunProperties, MetricsContext metrics) {
+    super(client, uri, aliyunProperties, metrics);
   }
 
-  OSSInputFile(OSS client, OSSURI uri, AliyunProperties aliyunProperties, long length) {
-    super(client, uri, aliyunProperties);
+  OSSInputFile(OSS client, OSSURI uri, AliyunProperties aliyunProperties, long length, MetricsContext metrics) {
+    super(client, uri, aliyunProperties, metrics);
     ValidationException.check(length >= 0, "Invalid file length: %s", length);
     this.length = length;
   }
@@ -49,6 +54,6 @@ public class OSSInputFile extends BaseOSSFile implements InputFile {
 
   @Override
   public SeekableInputStream newStream() {
-    return new OSSInputStream(client(), uri());
+    return new OSSInputStream(client(), uri(), metrics());
   }
 }
--- a/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
+++ b/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
@@ -24,7 +24,11 @@ import com.aliyun.oss.model.GetObjectRequest;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
+import org.apache.iceberg.io.FileIOMetricsContext;
 import org.apache.iceberg.io.SeekableInputStream;
+import org.apache.iceberg.metrics.MetricsContext;
+import org.apache.iceberg.metrics.MetricsContext.Counter;
+import org.apache.iceberg.metrics.MetricsContext.Unit;
 import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.io.ByteStreams;
@@ -44,10 +48,20 @@ public class OSSInputStream extends SeekableInputStream {
   private long next = 0;
   private boolean closed = false;
 
+  private final Counter<Long> readBytes;
+  private final Counter<Integer> readOperations;
+
   public OSSInputStream(OSS client, OSSURI uri) {
+    this(client, uri, MetricsContext.nullMetrics());
+  }
+
+  OSSInputStream(OSS client, OSSURI uri, MetricsContext metrics) {
     this.client = client;
     this.uri = uri;
     this.createStack = Thread.currentThread().getStackTrace();
+
+    this.readBytes = metrics.counter(FileIOMetricsContext.READ_BYTES, Long.class, Unit.BYTES);
+    this.readOperations = metrics.counter(FileIOMetricsContext.READ_OPERATIONS, Integer.class, Unit.COUNT);
   }
 
   @Override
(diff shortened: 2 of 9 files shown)


aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputFile.java
```diff
@@ -24,17 +24,18 @@
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.SeekableInputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
-public class OSSInputFile extends BaseOSSFile implements InputFile {
+class OSSInputFile extends BaseOSSFile implements InputFile {
```
rdblue, line 29: This is technically a breaking change. I think we should roll it back in case anyone references this file. We can deprecate the class so that we can make this change in a release or two.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
```diff
@@ -44,10 +48,16 @@
   private long next = 0;
   private boolean closed = false;
 
-  public OSSInputStream(OSS client, OSSURI uri) {
+  private final Counter<Long> readBytes;
+  private final Counter<Integer> readOperations;
+
+  public OSSInputStream(OSS client, OSSURI uri, MetricsContext metrics) {
```
rdblue, line 54: Since the old constructor was public, should we keep it and pass `nullMetrics`?

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSOutputFile.java
```diff
@@ -25,15 +25,17 @@
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.io.PositionOutputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
 class OSSOutputFile extends BaseOSSFile implements OutputFile {
 
-  OSSOutputFile(OSS client, OSSURI uri, AliyunProperties aliyunProperties) {
-    super(client, uri, aliyunProperties);
+  OSSOutputFile(OSS client, OSSURI uri, AliyunProperties aliyunProperties, MetricsContext metrics) {
```
rdblue, line 32: Similar here. Should we keep the old constructor and pass `nullMetrics`?

rdblue (changes requested):
Mostly looks good. I just have a couple questions about compatibility. If we kept the existing constructors, that would also cut down on a the changes to tests.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
rajarshisarkar: Thanks, for the feedback. I was retained the old constructor.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputFile.java
rajarshisarkar: I have rolled back the change.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
rajarshisarkar: Thanks, for the feedback. I have retained the old constructor.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputFile.java
```diff
@@ -24,17 +24,19 @@
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.SeekableInputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
+@Deprecated
```
rdblue, line 29: Can you add some context for this? Usually, we add javadoc to explain the deprecation:

```java
  /**
   * @deprecated moving to package-private in 0.15.0; use OSSFileIO to create InputFile instances.
   */
```

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
```diff
@@ -44,10 +48,27 @@
...
+
   public OSSInputStream(OSS client, OSSURI uri) {
     this.client = client;
     this.uri = uri;
     this.createStack = Thread.currentThread().getStackTrace();
+
+    this.readBytes = MetricsContext.nullMetrics().counter(
+        FileIOMetricsContext.READ_BYTES, Long.class, Unit.BYTES);
+    this.readOperations = MetricsContext.nullMetrics().counter(
+        FileIOMetricsContext.READ_OPERATIONS, Integer.class, Unit.COUNT);
```
rdblue, line 62: Can this constructor call `this(client, uri, MetricsContext.nullMetrics());` instead?

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
rdblue: Can you make this one package-private? I don't think we need to expose it.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSOutputFile.java
```diff
@@ -25,15 +25,17 @@
...
+    super(client, uri, aliyunProperties, metrics);
   }
 
-  static OSSOutputFile fromLocation(OSS client, String location, AliyunProperties aliyunProperties) {
-    return new OSSOutputFile(client, new OSSURI(location), aliyunProperties);
+  static OSSOutputFile fromLocation(OSS client, String location, AliyunProperties aliyunProperties,
```
rdblue, line 36: This is only used in tests as far as I can tell. How about not updating it and passing `nullMetrics` here? That will cut down on unnecessary test changes.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputFile.java
rajarshisarkar: Added the javadoc.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
rajarshisarkar: Thanks, I have made the changes.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSOutputFile.java
rajarshisarkar: Nice catch, I have passed `MetricsContext.nullMetrics()` to reduce the test changes.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
```diff
@@ -44,10 +48,20 @@
   private long next = 0;
   private boolean closed = false;
 
+  private final Counter<Long> readBytes;
+  private final Counter<Integer> readOperations;
+
   public OSSInputStream(OSS client, OSSURI uri) {
```
rdblue, line 54: Should we deprecate this, too?

rdblue (approved):

rdblue:
Thanks, @rajarshisarkar!

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
rajarshisarkar: Yes, I agree. Deprecating this would maintain the parity with the other InputStream classes (which have a package private constructor).

PR #4395, opened 2022-03-24
Spark: Add custom metric for number of file splits read by a SparkScan

wypoon:
Spark 3.2 added support for custom `SQLMetric`s which are shown in the Spark UI.
Here we add a custom metric for the number of files read by a SparkScan.

Attached are screenshots of the Spark UI for a query after this change.
Screen Shot 2022-03-24 at 2 36 44 PMScreen Shot 2022-03-24 at 2 37 11 PM

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/NumFiles.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/NumFiles.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.source;
+
+import java.text.NumberFormat;
+import org.apache.spark.sql.connector.metric.CustomMetric;
+
+public class NumFiles implements CustomMetric {
+
+  @Override
+  public String name() {
+    return "numFiles";
+  }
+
+  @Override
+  public String description() {
+    return "number of files read";
+  }
+
+  @Override
+  public String aggregateTaskMetrics(long[] taskMetrics) {
+    long sum = initialValue;
+    for (int i = 0; i < taskMetrics.length; i++) {
+      sum += taskMetrics[i];
+    }
+    return NumberFormat.getIntegerInstance().format(sum);
+  }
+}
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
@@ -93,11 +93,6 @@ class SparkBatchQueryScan extends SparkScan implements SupportsRuntimeFiltering 
     }
   }
 
-  @Override
-  public String description() {
-    return "IcebergBatchScan " + super.description();
-  }
-
   Long snapshotId() {
     return snapshotId;
   }
@@ -274,7 +269,7 @@ class SparkBatchQueryScan extends SparkScan implements SupportsRuntimeFiltering 
   @Override
   public String toString() {
     return String.format(
-        "IcebergBatchScan(table=%s, type=%s, filters=%s, runtimeFilters=%s, caseSensitive=%s)",
+        "IcebergScan(table=%s, type=%s, filters=%s, runtimeFilters=%s, caseSensitive=%s)",
         table(), expectedSchema().asStruct(), filterExpressions(), runtimeFilterExpressions, caseSensitive());
   }
 }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
@@ -93,6 +93,11 @@ class SparkBatchQueryScan extends SparkScan implements SupportsRuntimeFiltering 
     }
   }
 
+  @Override
+  public String description() {
+    return "IcebergScan " + super.description();
+  }
+
   Long snapshotId() {
     return snapshotId;
   }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
@@ -95,7 +95,7 @@ class SparkBatchQueryScan extends SparkScan implements SupportsRuntimeFiltering 
 
   @Override
   public String description() {
-    return "IcebergScan " + super.description();
+    return "IcebergBatchScan " + super.description();
   }
 
   Long snapshotId() {
@@ -274,7 +274,7 @@ class SparkBatchQueryScan extends SparkScan implements SupportsRuntimeFiltering 
   @Override
   public String toString() {
     return String.format(
-        "IcebergScan(table=%s, type=%s, filters=%s, runtimeFilters=%s, caseSensitive=%s)",
+        "IcebergBatchScan(table=%s, type=%s, filters=%s, runtimeFilters=%s, caseSensitive=%s)",
         table(), expectedSchema().asStruct(), filterExpressions(), runtimeFilterExpressions, caseSensitive());
   }
 }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java
@@ -77,6 +77,11 @@ class SparkCopyOnWriteScan extends SparkScan implements SupportsRuntimeFiltering
     }
   }
 
+  @Override
+  public String description() {
+    return "IcebergCopyOnWriteScan " + super.description();
+  }
+
   Long snapshotId() {
     return snapshot != null ? snapshot.snapshotId() : null;
   }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java
@@ -77,11 +77,6 @@ class SparkCopyOnWriteScan extends SparkScan implements SupportsRuntimeFiltering
     }
   }
 
-  @Override
-  public String description() {
-    return "IcebergCopyOnWriteScan " + super.description();
-  }
-
   Long snapshotId() {
     return snapshot != null ? snapshot.snapshotId() : null;
   }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkFilesScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkFilesScan.java
@@ -51,6 +51,11 @@ class SparkFilesScan extends SparkScan {
   }
 
   @Override
+  public String description() {
+    return "IcebergFilesScan " + super.description();
+  }
+
+  @Override
   protected List<CombinedScanTask> tasks() {
     if (tasks == null) {
       FileScanTaskSetManager taskSetManager = FileScanTaskSetManager.get();
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkFilesScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkFilesScan.java
@@ -51,11 +51,6 @@ class SparkFilesScan extends SparkScan {
   }
 
   @Override
-  public String description() {
-    return "IcebergFilesScan " + super.description();
-  }
-
-  @Override
   protected List<CombinedScanTask> tasks() {
     if (tasks == null) {
       FileScanTaskSetManager taskSetManager = FileScanTaskSetManager.get();
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
@@ -214,6 +214,7 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
       numFilesToRead = numFilesToScan(task.task);
+      LOG.debug("reading {} files for table {}", numFilesToRead, task.table().name());
     }
 
     @Override
@@ -228,6 +229,7 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
     BatchReader(ReadTask task, int batchSize) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive(), batchSize);
       numFilesToRead = numFilesToScan(task.task);
+      LOG.debug("reading {} files for table {}", numFilesToRead, task.table().name());
     }
 
     @Override
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
@@ -202,21 +202,13 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
     }
   }
 
-  static long numFilesToScan(CombinedScanTask scanTask) {
-    long fileCount = 0L;
-    for (FileScanTask file : scanTask.files()) {
-      fileCount += 1L;
-    }
-    return fileCount;
-  }
-
   private static class RowReader extends RowDataReader implements PartitionReader<InternalRow> {
     private long numSplits;
 
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
-      numSplits = numFilesToScan(task.task);
-      LOG.debug("reading {} file splits for table {} using RowReader", numSplits, task.table().name());
+      numSplits = (long) task.task.files().size();
+      LOG.debug("Reading {} file split(s) for table {} using RowReader", numSplits, task.table().name());
     }
 
     @Override
@@ -230,8 +222,8 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
 
     BatchReader(ReadTask task, int batchSize) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive(), batchSize);
-      numSplits = numFilesToScan(task.task);
-      LOG.debug("reading {} file splits for table {} using BatchReader", numSplits, task.table().name());
+      numSplits = (long) task.task.files().size();
+      LOG.debug("Reading {} file split(s) for table {} using BatchReader", numSplits, task.table().name());
     }
 
     @Override
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
@@ -39,6 +39,8 @@ import org.apache.iceberg.spark.Spark3Util;
 import org.apache.iceberg.spark.SparkReadConf;
 import org.apache.iceberg.spark.SparkSchemaUtil;
 import org.apache.iceberg.spark.SparkUtil;
+import org.apache.iceberg.spark.source.metrics.NumSplits;
+import org.apache.iceberg.spark.source.metrics.TaskNumSplits;
 import org.apache.iceberg.util.PropertyUtil;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.apache.spark.broadcast.Broadcast;
@@ -214,7 +216,7 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
       numSplits = numFilesToScan(task.task);
-      LOG.debug("reading {} file splits for table {}", numSplits, task.table().name());
+      LOG.debug("reading {} file splits for table {} using RowReader", numSplits, task.table().name());
     }
 
     @Override
@@ -229,7 +231,7 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
     BatchReader(ReadTask task, int batchSize) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive(), batchSize);
       numSplits = numFilesToScan(task.task);
-      LOG.debug("reading {} file splits for table {}", numSplits, task.table().name());
+      LOG.debug("reading {} file splits for table {} using BatchReader", numSplits, task.table().name());
     }
 
     @Override
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
@@ -207,7 +207,7 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
 
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
-      numSplits = (long) task.task.files().size();
+      numSplits = task.task.files().size();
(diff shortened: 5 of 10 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -193,15 +200,39 @@ public boolean supportColumnarReads(InputPartition partition) {
...
+
     BatchReader(ReadTask task, int batchSize) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive(), batchSize);
+      numFilesToRead = numFilesToScan(task.task);
+    }
+
+    @Override
+    public CustomTaskMetric[] currentMetricsValues() {
+      return new CustomTaskMetric[] { new TaskNumFiles(numFilesToRead) };
     }
```
wypoon, line 236: Note: Spark calls this every 100 rows for each `PartitionReader`. The `numFilesToRead` can be computed up front and stored, and is. I wonder if it is worth caching the `CustomTaskMetric[]` as well.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -93,6 +93,11 @@
     }
   }
 
+  @Override
+  public String description() {
+    return "IcebergScan " + super.description();
+  }
```
wypoon, line 99: Unfortunately the label for the node in the SQL DAG in the Spark UI just says "BatchScan" (that is the `nodeName` for a `BatchScanExec` node, and we can't affect it from Iceberg), but the tool tip for the node contains this `description`.
Also, the detailed Physical Plan shows this `description`. The `toString` method also uses "IcebergScan" (I followed its lead), but I don't see the `toString` being used in the Spark UI.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -193,15 +200,39 @@ public boolean supportColumnarReads(InputPartition partition) {
...
+
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
+      numFilesToRead = numFilesToScan(task.task);
+    }
+
+    @Override
+    public CustomTaskMetric[] currentMetricsValues() {
+      return new CustomTaskMetric[] { new TaskNumFiles(numFilesToRead) };
     }
```
wypoon, line 222: Note: Spark calls this every 100 rows for each `PartitionReader`. The `numFilesToRead` can be computed up front and stored, and is. I wonder if it is worth caching the `CustomTaskMetric[]` as well.

wypoon:
Note that in order for the Spark UI to show the custom metric, the custom metric class must be available in the classpath of the Spark History Server. The simplest way to do this is to put the Iceberg Spark runtime JAR in the SHS classpath.

RussellSpitzer:
> Note that in order for the Spark UI to show the value of the custom metric, the custom metric class must be available in the classpath of the Spark History Server. The simplest way to do this is to put the Iceberg Spark runtime JAR in the SHS classpath. Spark's `SQLAppStatusListener` loads the custom metric class by name, instantiates it, and calls its `aggregateTaskMetrics` method to get the value of the metric. If it is not able to load the custom metric class, it shows "N/A".

Yeah I was very frustrated with this when i was messing with it. Makes it hard if your history server is not running the same Spark build as your apps :(

wypoon:
@RussellSpitzer thank you for the interest. I just updated the PR to add debug logging on the `PartitionReader`s to show how many files each is reading (and the table being read). If you're playing with this, that helps you see the numbers that are being reported for each `InputPartition`, and the sum should be reported in the Spark UI.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/NumFiles.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public String aggregateTaskMetrics(long[] taskMetrics) {
+    long sum = initialValue;
+    for (int i = 0; i < taskMetrics.length; i++) {
+      sum += taskMetrics[i];
+    }
+    return NumberFormat.getIntegerInstance().format(sum);
+  }
```
singhpk234, line 44: [question] This is already awesome :) ... 
just thinking out loud here what if we display this metrics in this format 
> "total (min, med, max) "

we would be even have better understanding of the what was files distribution at the task level

singhpk234, line 44: [question] should we make this return "total(min, max, med)" we already have all the metrics required, this should give us even deeper insights on files distribution at task level. WDYT ?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -193,15 +200,41 @@ public boolean supportColumnarReads(InputPartition partition) {
...
   }
 
+  static long numFilesToScan(CombinedScanTask scanTask) {
+    long fileCount = 0L;
+    for (FileScanTask file : scanTask.files()) {
+      fileCount += 1L;
```
singhpk234, line 206: should we incorporate delete's as well assosiated with the FileScanTask
```suggestion
      fileCount += 1L + file.deletes().size();
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
singhpk234: can it be a property of this RowReader then and we could just return that when ever asked.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/NumFiles.java
wypoon: @singhpk234 thank you for your interesting suggesions!
I am open to presenting  the min, median and max as well, but I wonder how useful those statistics really are. My impression is that Iceberg scan planning works well enough that the number of files to scan are usually very evenly distributed (except for cases like where there are two `InputPartition`s). Usually skew occurs with times (which can of course show great variation), or with number of records after a shuffle (rather than a `BatchScan`).
@rdblue, @aokolnychyi, @RussellSpitzer what do you think?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
wypoon: We could have a separate metric for number of delete files to read. Or rather, a sum of the number of delete files to apply, over the data files to be read; we won't actually track what the delete files are, if there is any overlap among them (can there be overlap?).
@rdblue @aokolnychyi what do you think?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
wypoon: Yes, this is what I meant, is it worth caching the `CustomTaskMetric[]`, so it can be constructed just once.
I'm undecided.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/NumFiles.java
RussellSpitzer: I though the way these are shown in the UI are per task as well? Or am I remembering that wrong. IE we should have all these values (but not the calculated sum?) If not I think that would be good to report, I would suggest probably

min, 25percentile, median, 75percentile, max?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
RussellSpitzer: This sounds good to me even outside the scope of this PR, although maybe we sould right Iceberg Batch Scan, since we will probably be generating different descriptions for SparkFileScan and any other Scans we build

RussellSpitzer: Ah I see we have the other descriptions below, well at least let's match that pattern :) "IcebergBatchScan"?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
RussellSpitzer: I think it's fair to count all the delete files, but definitely in another metric. 

There will be duplicates in two cases
1. It's an equality delete
2. There is a position delete file for a data file which is split into being read in two different tasks (this is also a problem for duplicate files being read)

RussellSpitzer:
So one big issue I have here, and why I kind of gave up on this for the moment, is that we actually know all of this information on the driver before any executors run. Not only do we have it there but we can also eliminate duplicate counts for both delete files and for data files. 

In the current code we would get incorrect results for number of data files if our split size is smaller than our file size. Imagine we have a 512 mb file and 128 mb split size (and row group) this will generate 4 separate tasks for the same Data file. Counting the file 4 times.

On the driver we have already materialized every single FileScanTask and know this information and can do the dedupe, our only problem is currently spark doesn't allow us to populate custom metrics at the Source itself on the driver. I think we should push for that capability in spark rather than doing this at the executor level. (Or do it in conjunction with the executor metrics)

RussellSpitzer:
Something we could do with custom metrics that we don't know at planning time would be tracking for example, how many records are eliminated via delete files. That's a good metric to know and we wouldn't know it until a task completes.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/NumFiles.java
wypoon: @RussellSpitzer I think you may be thinking of task metrics that Spark itself collects (by default, Duration, GC Time,  Input Size / Records, and Shuffle Size /Records are shown), and are shown with the min, 25th percentile, median, 75th percentile, and max.
For custom `SQLMetric`s, one implements the `aggregateTaskMetrics` and we'd need to do the calculation ourselves if we want such statistics. Spark doesn't do it for us. Spark sends all the values it collected from all the `PartitionReader`s for a given custom metric in a `long[]` to this `aggregateTaskMetrics`; that's all. It also doesn't separately calculate these statistics either.

wypoon:
> So one big issue I have here, and why I kind of gave up on this for the moment, is that we actually know all of this information on the driver before any executors run. Not only do we have it there but we can also eliminate duplicate counts for both delete files and for data files.
> 
Yes, I realize that we know the number of files before the tasks are run. It would certainly be better if Spark had driver-side custom metrics. But this is something we can implement with what is Spark 3.2 now. It's not the best way, but I think it's still useful to provide this metric, and it can be implemented in a better way when we get driver-side custom metrics in Spark.

> In the current code we would get incorrect results for number of data files if our split size is smaller than our file size. Imagine we have a 512 mb file and 128 mb split size (and row group) this will generate 4 separate tasks for the same Data file. Counting the file 4 times.
> 
That is a good point. Even so, such a count is still a useful metric, in that it is a measure of amount of work to do, although "number of files" is not the correct name for it. Of course, the actual number of files is a metric we're interested in as well.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
wypoon: I was matching the pattern in the `toString` method for each of the subclasses of `SparkScan`. If you prefer, we can use "IcebergBatchScan" in both `description` and `toString` for `SparkBatchQueryScan`.

kbendick: I am +1 on `BatchScan` -> `IcebergBatchScan` for UI purposes.

I think that updating the `toString` method should be fine as well if need be, provided it passes tests etc. `IcebergBatchScan` is closer to the name of the class anyway than plain `IcebergScan`.

kbendick:
> > Note that in order for the Spark UI to show the value of the custom metric, the custom metric class must be available in the classpath of the Spark History Server. The simplest way to do this is to put the Iceberg Spark runtime JAR in the SHS classpath. Spark's `SQLAppStatusListener` loads the custom metric class by name, instantiates it, and calls its `aggregateTaskMetrics` method to get the value of the metric. If it is not able to load the custom metric class, it shows "N/A".
> 
> Yeah I was very frustrated with this when i was messing with it. Makes it hard if your history server is not running the same Spark build as your apps :(

This qualification for the history server would be a really good thing to call out in the docs somewhere if we merge this.

>> In the current code we would get incorrect results for number of data files if our split size is smaller than our file size. Imagine we have a 512 mb file and 128 mb split size (and row group) this will generate 4 separate tasks for the same Data file. Counting the file 4 times.

> That is a good point. Even so, such a count is still a useful metric, in that it is a measure of amount of work to do, although "number of files" is not the correct name for it. Of course, the actual number of files is a metric we're interested in as well.

I also think there's value in this. Would it just be better to call it "number of splits read" or something? Provided we can give it a name that clearly delineates what is being measured, I think this has value.

Though ideally metrics could be exported from the driver as well (which would need to be added to Spark).

wypoon:
@kbendick thank you for the perspective. As @RussellSpitzer has pointed out, custom task metrics do not appear to be the right mechanism for getting a correct count for number of data files read, or number of delete files applied while reading data files. I think we need to wait for Spark support for driver-side custom metrics to implement those properly. In the meantime, I was thinking of renaming the custom metric that I have here "number of file splits read". I'll update the PR very soon.
For other metrics that are actually computed at task time, such as the one Russell suggested, let me look into that.

wypoon:
I have updated the PR. I renamed the metric "numSplits" and its description is now "number of file splits read" (this is what appears in the Spark UI). I also changed "IcebergScan" to "IcebergBatchScan" in the `description` and `toString` methods of `SparkBatchQueryScan`.
For now, I am not adding any other metrics in this PR. Task metrics do not appear to be the right mechanism for computing how many delete files are applied, for example. I will look into other metrics for which task metrics are the right mechanism, but that can be a separate PR.
I am open to computing min, median and max for the numSplits metric if that is desired.

wypoon:
@kbendick do you have any further feedback on this?
@RussellSpitzer I have implemented your suggestion for a custom metric for how many delete rows are applied in a scan. I decided to open a new PR for that: #4588.
I think both the number of file splits read as well as number of delete rows applied are useful metrics to show. More custom metrics can be added in future but they can be in new PRs.

RussellSpitzer:
I left a bunch of comments here that would also apply to the other PR. I'll do another pass on the custom metric pr for just file splits in a bit, but please take a look at my comments here in the mean time.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -93,6 +93,11 @@ class SparkBatchQueryScan extends SparkScan implements SupportsRuntimeFiltering
     }
   }
 
+  @Override
```
RussellSpitzer, line 96: I would separate these changes out as they are not related to this PR

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java
```diff
@@ -77,6 +77,11 @@ class SparkCopyOnWriteScan extends SparkScan implements SupportsRuntimeFiltering
     }
   }
 
+  @Override
+  public String description() {
+    return "IcebergCopyOnWriteScan " + super.description();
```
RussellSpitzer, line 82: Also unrelated to this pr

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -193,15 +202,41 @@ public boolean supportColumnarReads(InputPartition partition) {
...
+    }
+    return fileCount;
```
RussellSpitzer, line 210: return scanTask.files().size() ?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -193,15 +202,41 @@ public boolean supportColumnarReads(InputPartition partition) {
...
+  }
+
   private static class RowReader extends RowDataReader implements PartitionReader<InternalRow> {
+    private long numSplits;
+
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
+      numSplits = numFilesToScan(task.task);
+      LOG.debug("reading {} file splits for table {} using RowReader", numSplits, task.table().name());
```
RussellSpitzer, line 219: Capital R "Reading

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -193,15 +202,41 @@ public boolean supportColumnarReads(InputPartition partition) {
     }
   }
 
+  static long numFilesToScan(CombinedScanTask scanTask) {
```
RussellSpitzer, line 205: numSplitsToScan?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/metrics/NumSplits.java
```diff
@@ -0,0 +1,45 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.source.metrics;
+
+import java.text.NumberFormat;
+import org.apache.spark.sql.connector.metric.CustomMetric;
+
+public class NumSplits implements CustomMetric {
```
RussellSpitzer, line 25: CustomSumMetric fits this a bit better and you will not need to reimplement the Aggregate

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/metrics/NumSplits.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public String name() {
+    return "numSplits";
+  }
+
+  @Override
+  public String description() {
+    return "number of file splits read";
```
RussellSpitzer, line 34: Start with a Capital leter

RussellSpitzer (changes requested):
A few comments, Let's try to remove the changes to the descriptions, although I think those would be good as another PR.

The Metric could extend CustomSumMetric 
There are some remaining "file" methods that should read as "split"

And some other minor changes, Looks good though!

RussellSpitzer:
One other thing, how are we going to test this? I'm not sure of any ways to get to this new metric programmatically but we should definitely investigate that.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/metrics/NumSplits.java
wypoon: I explained in #4588 why there is no benefit to extend `CustomSumMetric`.

wypoon:
@RussellSpitzer thank you for reviewing again!
The changes to the descriptions were discussed in an earlier review comment, and my understanding was that reviewers were in favor of the changes. They affect the Spark UI. That said, if you prefer, I will remove them and put them in a separate PR.
I have tested this manually. My take is this: In Spark, there are unit tests for testing consumption of custom metrics by the SparkAppListener. Here we are just plugging into the API. On the Iceberg side, how is the metric we are adding computed? It is simply scanTask.files().size() as you point out. Scan planning is presumably tested, so there is really nothing essentially new that needs to be tested. In #4588, on the other hand, there are extensive changes needed to compute the metric being added, so I test that computation.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
wypoon: Removed.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java
wypoon: Removed.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkFilesScan.java
```diff
@@ -50,6 +50,11 @@ class SparkFilesScan extends SparkScan {
     this.splitOpenFileCost = readConf.splitOpenFileCost();
   }
 
+  @Override
+  public String description() {
```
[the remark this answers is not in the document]
wypoon: Removed.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
wypoon: Adopted.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
wypoon: Since this is now a one-liner, I have removed the function altogether.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/metrics/NumSplits.java
wypoon: I explained in #4588 that I lowercase "number" to be consistent with the standard metrics shown in the Spark UI.

wypoon:
@RussellSpitzer I have addressed your feedback. I think this one should be ready.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -202,21 +202,13 @@ public boolean supportColumnarReads(InputPartition partition) {
...
-  }
-
   private static class RowReader extends RowDataReader implements PartitionReader<InternalRow> {
     private long numSplits;
 
     RowReader(ReadTask task) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive());
-      numSplits = numFilesToScan(task.task);
-      LOG.debug("reading {} file splits for table {} using RowReader", numSplits, task.table().name());
+      numSplits = (long) task.task.files().size();
```
sririshindra, line 210: Do we really need to typecast this to long? Isn't `task.task.files().size();` enough?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -230,8 +222,8 @@ private static class BatchReader extends BatchDataReader implements PartitionRea
 
     BatchReader(ReadTask task, int batchSize) {
       super(task.task, task.table(), task.expectedSchema(), task.isCaseSensitive(), batchSize);
-      numSplits = numFilesToScan(task.task);
-      LOG.debug("reading {} file splits for table {} using BatchReader", numSplits, task.table().name());
+      numSplits = (long) task.task.files().size();
```
sririshindra, line 225: ditto

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
wypoon: `java.util.Collection#size()` returns an `int`. Yes, without the explicit cast, the compiler will do it for us as `numsplits` is declared to be a `long`.
If an explicit cast is worse for performance, I shall remove it. Otherwise, it is a matter of style.

RussellSpitzer: I don't think it really makes a difference, the implicit cast should be just as good so I would probably just remove the cast

wypoon: I'll remove the cast.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/metrics/TaskNumSplits.java
```diff
@@ -0,0 +1,40 @@
...
+public class TaskNumSplits implements CustomTaskMetric {
+  private final long value;
+
+  public TaskNumSplits(long value) {
+    this.value = value;
+  }
+
+  @Override
+  public String name() {
+    return "numSplits";
```
RussellSpitzer, line 33: Should we say numFileSplits? I'm always worried about how overloaded the term "split" is especially in Spark. I believe you used this in the debug statement above.

Like here https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L71 :)

wypoon: This name is internal. The name in the `CustomTaskMetric` (in this case `TaskNumSplits`) needs to match the name in the `CustomMetric` (in this case `NumSplits`).
The user facing part is the description in the `CustomMetric`, in this case "number of file splits read". This appears in the Spark UI.
The debug logging does not use the name. (It uses the value of a `numSplits` variable.) It would log something like
"Reading 1835 file split(s) for table spark_catalog.tpcds_iceberg.catalog_sales using RowReader".

I had previously changed the numFiles/NumFiles in names to numSplits/NumSplits at your suggestion. Do you really think it is necessary to change all those names again?

wypoon:
The changes here were part of #4588 which has been merged. Closing.

Issue #4289, opened 2022-03-08
Procedure do not work with comments


cccs-jc:
this works
```
spark.sql("""
CALL catalog_name.system.remove_orphan_files(table => 'db.sample', dry_run => true)
""")
```

But this fails
```
spark.sql("""
/* comment */
CALL catalog_name.system.remove_orphan_files(table => 'db.sample', dry_run => true)
""")
```

Tools like dbt add comments to the statements that are executed. The comments are useful for auditing.

kbendick:
Hi @cccs-jc! Thanks for reporting this.

Can you please let me know which Iceberg (and Spark) version you have tried this with (just your normal one would be fine, or the latest one you've tried). Also, can you provide me an example (redacted) comment that's more representative of the kind DBT has?

I see some comment handling code in our `IcebergSqlExtensions.g4` file which generates the antlr grammar, but some comments have meaning (SQL hints) and so having a closer example to what's really generated would be helpful while I take a look.

Thank you!

RussellSpitzer:
I think we'd have to modify our grammer, since our parser uses it's own grammar which doesn't have comments. That or finally copy the grammar from the relevant version of spark and start using that instead of our own custom extension. This may be the time now that we separated by version

cccs-jc:
dbt issues comments like this before it renders the procedure calls


/* {"app": "dbt", "dbt_version": "1.0.1", "profile_name": "profile1", "target_name": "dev", "node_id": "model.profile1.stg_users"} */

right now I modified my dbt-spark adapter to strip this out but I would rather keep the comments for auditing

kbendick:
I think I have found a solution that seems to work for `CALL` statements. The issue is similar to this one, for making the iceberg statements more recognizable: https://github.com/apache/iceberg/pull/2729/files

I will open a draft PR shortly.

kbendick:
I matched the admittedly wild regex used there against this test query statement. https://regex101.com/r/9cs1yV/1

Fortunately, this function doesn't modify the query at all, just how the query is determined to be an Iceberg query or not.

We still delegate to the _existing_ antlr parser for processing the query, and only strip comments out when we're checking if this is Iceberg specific DDL. Hence there's no concerns about accidentally dropping SQL hints, etc.

kbendick:
This has been merged in for Spark 3.2. There are open PRs for Spark 3.0 and 3.1 as well. I'm going to close this issue now.

[...]

PR #2729, opened 2021-06-24
make iceberg sql command recognition more whitespace tolerant

jfz:
Currently, iceberg sql command recognition is strict: extra white spaces between some of the keywords are not allowed. 
This fix looses this restriction and make it more user friendly as well as more aligned with general sql syntax.

PR #4396, opened 2022-03-24
Spark - Ignore SQL comments when testing if SQL command is an Iceberg specific command

kbendick:
This closes issue https://github.com/apache/iceberg/issues/4289

Currently, We ignore simple comments (those that start with `--`), as well as bracketed comments in our IcebergSqlExtensions ANTLR grammar.

However, we have a function `isIcebergCommand`, that checks if a command is an Iceberg command by checking if the string starts with specific phrases that match known Iceberg commands.

The sqlText has not yet been passed through the full ANTLR process, so we need to ignore comments manually, akin to ignoring new-lines in https://github.com/apache/iceberg/pull/2729

-------------------------------------------------

There might be a better way to handle this, but this does work for recognizing the Iceberg commands, while still allowing the antlr parser to remove comments, as it has special handling to ensure that SQL hints are _not_ stripped out.

So nothing about how the query is processed is changing, just how we determine whether or not this is Iceberg specific DDL.

### DIFF
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
@@ -177,7 +177,8 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
 
   private def isIcebergCommand(sqlText: String): Boolean = {
     val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
-      // Catch all SQL comments, including simple comments (that start with --), as well as multiline comments
+      // Catch all SQL comments, including simple comments (that start with --), as well as comments
+      // of the form /* ... */, including those that span multiple lines.
       .replaceAll("(?ms)/\\*.*?\\*/|--.*?\\n", " ")
       // Replace any remaining newlines
       .replaceAll("\\s+", " ")
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
@@ -177,15 +177,10 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
 
   private def isIcebergCommand(sqlText: String): Boolean = {
     val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
-//      .replaceAll("\\s+", " ")
-//      .replaceAll("(?m)^--.*$", "") // replace all simple single line SQL comments
-//      .replaceAll("/\\*.*?\\*/", "") // replace bracketed SQL comments
       // Catch all SQL comments, including simple comments (that start with --), as well as multiline comments
       .replaceAll("(?ms)/\\*.*?\\*/|--.*?\\n", " ")
-      // replace new lines - needed for getting multiline comments that have SQL text after them
-      .replaceAll("\\s+", " ") // replace new lines - needed for getting multiline comments
-//      .replaceAll("/\\*.*?\\*/", "") // replace bracketed SQL comments
-      .replaceAll("\\s+", " ") // replace newlines
+      // Replace any remaining newlines
+      .replaceAll("\\s+", " ")
       .trim()
     normalized.startsWith("call") || (
         normalized.startsWith("alter table") && (
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
@@ -177,11 +177,13 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
 
   private def isIcebergCommand(sqlText: String): Boolean = {
     val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
-      // Catch all SQL comments, including simple comments (that start with --), as well as comments
-      // of the form /* ... */, including those that span multiple lines.
-      .replaceAll("(?ms)/\\*.*?\\*/|--.*?\\n", " ")
-      // Replace any remaining newlines
+      // Strip simple SQL comments that terminate a line, e.g. comments starting with `--` .
+      .replaceAll("--.*?\\n", " ")
+      // Strip newlines.
       .replaceAll("\\s+", " ")
+      // Strip comments of the form  /* ... */. This must come after stripping newlines so that
+      // comments that span multiple lines are caught.
+      .replaceAll("/\\*.*?\\*/", " ")
       .trim()
     normalized.startsWith("call") || (
         normalized.startsWith("alter table") && (
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
@@ -176,7 +176,17 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
   }
 
   private def isIcebergCommand(sqlText: String): Boolean = {
-    val normalized = sqlText.toLowerCase(Locale.ROOT).trim().replaceAll("\\s+", " ")
+    val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
+//      .replaceAll("\\s+", " ")
+//      .replaceAll("(?m)^--.*$", "") // replace all simple single line SQL comments
+//      .replaceAll("/\\*.*?\\*/", "") // replace bracketed SQL comments
+      // Catch all SQL comments, including simple comments (that start with --), as well as multiline comments
+      .replaceAll("(?ms)/\\*.*?\\*/|--.*?\\n", " ")
+      // replace new lines - needed for getting multiline comments that have SQL text after them
+      .replaceAll("\\s+", " ") // replace new lines - needed for getting multiline comments
+//      .replaceAll("/\\*.*?\\*/", "") // replace bracketed SQL comments
+      .replaceAll("\\s+", " ") // replace newlines
+      .trim()
     normalized.startsWith("call") || (
         normalized.startsWith("alter table") && (
             normalized.contains("add partition field") ||
--- a/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
+++ b/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
@@ -22,8 +22,10 @@ package org.apache.iceberg.spark.extensions;
 import java.math.BigDecimal;
 import java.sql.Timestamp;
 import java.time.Instant;
+import java.util.List;
 import org.apache.iceberg.AssertHelpers;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.catalyst.expressions.Expression;
 import org.apache.spark.sql.catalyst.expressions.Literal;
@@ -135,6 +137,28 @@ public class TestCallStatementParser {
     AssertHelpers.assertThrows("Should fail with a sensible parse error", IcebergParseException.class,
         "missing '(' at 'radish'",
         () -> parser.parsePlan("CALL cat.system radish kebab"));
+  }
+
+  @Test
+  public void testCallStripsLeadingBracketedComments() throws ParseException {
+    // These comments are meant to look like those that dbt would add to statements.
+    List<String> callStatementsWithComments = Lists.newArrayList(
+        "/* bracketed comment */  CALL cat.system.func('${spark.extra.prop}')",
+        "/**/  CALL cat.system.func('${spark.extra.prop}')",
+        "-- single line comment \n CALL cat.system.func('${spark.extra.prop}')",
+        "-- multiple \n-- single line \n-- comments \n CALL cat.system.func('${spark.extra.prop}')",
+        "/* select * from multiline_comment \n where x like '%sql%'; */ CALL cat.system.func('${spark.extra.prop}')",
+        "/* {\"app\": \"dbt\", \"dbt_version\": \"1.0.1\", \"profile_name\": \"profile1\", \"target_name\": \"dev\", " +
+            "\"node_id\": \"model.profile1.stg_users\"} \n*/ CALL cat.system.func('${spark.extra.prop}')"
+    );
+    for (String sqlText : callStatementsWithComments) {
+      CallStatement call = (CallStatement) parser.parsePlan(sqlText);
+      Assert.assertEquals(ImmutableList.of("cat", "system", "func"), JavaConverters.seqAsJavaList(call.name()));
+
+      Assert.assertEquals(1, call.args().size());
+
+      checkArg(call, 0, "value", DataTypes.StringType);
+    }
   }
 
   private void checkArg(CallStatement call, int index, Object expectedValue, DataType expectedType) {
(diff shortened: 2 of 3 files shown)


kbendick:
I matched the admittedly wild regex used there against this test query statement. https://regex101.com/r/9cs1yV/1

Fortunately, this function doesn't modify the query at all, just how the query is determined to be an Iceberg query or not.

We still delegate to the existing antlr parser for processing the query, and only strip comments out when we're checking if this is Iceberg specific DDL. Hence there's no concerns about accidentally dropping SQL hints, etc.

kbendick:
cc @aokolnychyi @RussellSpitzer @rdblue @jackye1995 @danielcweeks @szehon-ho @flyrain  - can you please look at this? Trying to allow for comments in Iceberg DDL, so that people can use Iceberg with DBT.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
```diff
@@ -176,7 +176,12 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
   }
 
   private def isIcebergCommand(sqlText: String): Boolean = {
-    val normalized = sqlText.toLowerCase(Locale.ROOT).trim().replaceAll("\\s+", " ")
+    val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
+      // Catch all SQL comments, including simple comments (that start with --), as well as multiline comments
+      .replaceAll("(?ms)/\\*.*?\\*/|--.*?\\n", " ")
```
RussellSpitzer, line 181: Did this get copied from somewhere else or did you craft it yourself?

kbendick: Kind of a combo of both? I first found a similar regex in Perl or maybe it was PSQL, then added to it.

I can dig it up if you’d like or if we should credit the original two or so stack overflow pages.

kbendick: I can also explain it if we’d like. It’s operating in multi-line mode as well as single line mode, to be able to catch two scenarios:

1) `/\\*.*?\\*/`  is for catching comments of the form `/* …. */`. We operate in `(m)` multi line mode to be able to catch the case where that comment spans multiple lines. The `m` allows for the `.` to match on a new line character.
2) `--.*?\\n` is for catching simple SQL comments, that start with two dashes and then consume the rest of the line they are on like `-- some line terminating comment`. We operate in `(s)` single line mod here. We don’t risk hitting multi-line mode here really, eg having a comment like that also catch the next line, because we explicitly search for the newline character (and there’s a test for that case).

But admittedly there’s no chance I’d just sit down and craft a regex like this without a lot of process 🙂

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
```diff
@@ -137,6 +139,28 @@ public void testCallParseError() {
         () -> parser.parsePlan("CALL cat.system radish kebab"));
   }
 
+  @Test
+  public void testCallStripsLeadingComments() throws ParseException {
+    // These comments are meant to look like those that systems like DBT would add to statements.
+    List<String> callStatementsWithComments = Lists.newArrayList(
```
RussellSpitzer, line 145: Can you please in in tests for comments interspersed and after valid commands?

kbendick: Can you explain that one more time? You're asking for tests with comments inside of and after valid commands? I can add that.

I'm also going to add a test that actually calls the `CALL` procedure with a comment in it.

kbendick: I added tests for comments interspersed with valid commands.

I also added an end to end test, although it's only testing one command at present. The normal `spark.sql` parser that we fall back to doesn't seem to handle comments very well if there are two or more statements in it.

kbendick: So I think this will solve the user's problem, as only one statement is executed at a time in dbt if I'm not mistaken (and this fix is needed regardless to recognize Iceberg commands with comments in them.

But for multiple statements being processed, we'd have to add some splitting around `;` like the Spark Hive CLI does https://github.com/apache/spark/commit/450418bdfdc46c2569ff22c307de15c4fca76ebc

kbendick: In fact, plain `spark.sql` can't handle multiple commands at once.

I tried this in a test, `sql("INSERT INTO TABLE %s VALUES (1, 'a'); INSERT INTO TABLE %s VALUES(2, 'a');", tableName, tableName)`, and got the following error message:

```
== SQL ==
INSERT INTO TABLE testhive.default.table VALUES (1, 'a'); INSERT INTO TABLE testhive.default.table VALUES(2, 'a');
----------------------------------------------------------^^^

	at app//org.apache.spark.sql.catalyst.parser.ParseException.withCommand(ParseDriver.scala:266)
	at app//org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parse(ParseDriver.scala:127)
	at app//org.apache.spark.sql.execution.SparkSqlParser.parse(SparkSqlParser.scala:51)
	at app//org.apache.spark.sql.catalyst.parser.AbstractSqlParser.parsePlan(ParseDriver.scala:77)
	at app//org.apache.spark.sql.catalyst.parser.extensions.IcebergSparkSqlExtensionsParser.parsePlan(IcebergSparkSqlExtensionsParser.scala:123)
```

So that's falling back to the delegate parser (Spark's own parser). But the case of a single SQL command having comments at the beginning (or in the middle or at the end) is covered by this change. Previously, it would have worked for commands that are not covered by `isIcebergCommend` - e.g. anything starting with `call` or some subset of `alter table` commands.

kbendick: The full error message is this 

```
extraneous input 'INSERT' expecting {<EOF>, ';'}(line 1, pos 58)

== SQL ==
INSERT INTO TABLE testhive.default.table VALUES (1, 'a'); INSERT INTO TABLE testhive.default.table VALUES(2, 'a');
----------------------------------------------------------^^^
```

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
```diff
@@ -176,7 +176,12 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
...
+      // Replace any remaining newlines
+      .replaceAll("\\s+", " ")
+      .trim()
```
flyrain, line 184: Suggest to separate the logic of line 179-182 to a new method `normalizeSqlText()`, it is weird as a part of method `isIcebergCommand`. So that the caller side would be
```
 val sqlTextAfterSubstitution = substitutor.substitute(sqlText)
 val normalizedSqlText = normalizeSqlText(sqlTextAfterSubstitution)
    if (isIcebergCommand(sqlTextAfterSubstitution)) {
```

flyrain (approved):
LGTM. Thanks for the PR @kbendick.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
RussellSpitzer: This sounds reasonable to me too

kbendick: Assuming you mean to pass the result of `normalizeSqlText` to `isIcebergCommand`, my concern is that we want to allow the parser to do the actual removal of comments.

I’m happy to move the normalization into a separate function, but I don’t think we should be manipulating the SQL text beyond whether or not we are checking if this is Iceberg specific DDL.

If the request is to just make it a function that normalizes and the normalized result is _only_ passed to `isIcebergCommand`, I’m not opposed to that. Though I do think it opens up the possibility of accidentally refactoring and then continuing to pass the stripped out SQL text to the parser.

The parser is responsible for removing comments etc, as it knows how to differentiate between comments that are in lined SQL hints etc.

I could make a named function that is then called inside of `isIcebergCommand` though (or whose results are _only_ passed to `isIcebergCommand`).

kbendick: Ok. I updated it to use a named function, but the constraints on that function are pretty large as we want the antlr parser to be responsible for removing SQL comments etc.

I think it's better the original way and that the best solution would be to add additional comments clarifying why we're manipulating the `sqlText` within `isIcebergCommand`. I do concede that overall it is somewhat weird to be normalizing the SQL text directly within that function, but doing so helps prevent people from accidentally calling `normalizeSqlText` on sql text that will be passed to the parser.

So my suggestion would be something more like:
```scala
  private def isIcebergCommand(sqlText: String): Boolean = {
    // The sqlText has not been passed through the parser yet and therefore has not had
    // unimportant comments and unnecessary whitespace and newlines removed.
    // 
    // SQL comments and newlines are removed here so that Iceberg commands can be properly recognized,
    // while still allowing the antlr parser to perform the actual work of differentiating
    // comments that don't have any meaning to the query from comments that behave as sql hints etc.
    val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
      // Catch all SQL comments, including simple comments (that start with --), as well as comments
      // of the form /* ... */, including those that span multiple lines.
      .replaceAll("(?ms)/\\*.*?\\*/|--.*?\\n", " ")
      // Replace any remaining newlines
      .replaceAll("\\s+", " ")
      .trim()
    normalized.startsWith("call") || (
        normalized.startsWith("alter table") && (
            normalized.contains("add partition field") ||
            normalized.contains("drop partition field") ||
            normalized.contains("replace partition field") ||
            normalized.contains("write ordered by") ||
            normalized.contains("write locally ordered by") ||
            normalized.contains("write distributed by") ||
            normalized.contains("write unordered") ||
            normalized.contains("set identifier fields") ||
            normalized.contains("drop identifier fields")))
  }
```

I'm not in love with this either, but I think what's making things a bit weird is the fact that we have to check whether or not something is an Iceberg specific command before using the actual parser.

I'm not prepared to refactor the parser to that level yet, so I think his is the best solution. That said, if we want to use the current `normalizeSqlText` function with the comment / understanding that it should only be used in that one spot, that is a decision I would leave up to the community. I personally think it's unsafe in the long term to have a function, `normalizeSqlText`, that is only meant to be used in one very specific location and which could have dangerous side effects if used elsewhere.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
```diff
@@ -177,11 +189,6 @@ class IcebergSparkSqlExtensionsParser(delegate: ParserInterface) extends ParserI
 
   private def isIcebergCommand(sqlText: String): Boolean = {
     val normalized = sqlText.toLowerCase(Locale.ROOT).trim()
```
kbendick, line 191: Under the new suggested form of first normalizing the sql text in a separate function and then passing it to this function, I chose to leave `toLowerCase` inside of this function because having the input not be lowercased would potentially break the check.

As mentioned, I prefer to keep the logic all within `isIcebergCommand`, given that we don't want to use the normalized text outside of this function / this specific purpose.

That said, if you both agree that the named function is better (despite possibly being unsafe), I can move all of the normalization into the new function and leave just the boolean check in `isIcebergCommand`.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
flyrain: Thanks for the explain. Sorry, I miss that point. But I see another issue here. The method `isIcebergCommand` isn't reliable. For example, you can have a statement like this. 
```
alter table table_name SET TBLPROPERTIES ('key1' = 'add partition field')
```
It isn't an Iceberg command, but `isIcebergCommand` will return true since it just searches for the string `add partition field`.

One solution is to use the parser to get all tokens first, then decide if it is an Iceberg command.

kbendick: That's a fair assessment. However, I think that should be handed in a separate PR. As it's unrelated to this issue / is an existing issue. And admittedly I'm not going to have time to make large changes to the existing parsing code.

So I agree that there's definitely cases it doesn't handle well, but I wouldn't say that this makes it any worse and it does make it better for end users who would like to use procedures with DBT.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
kbendick: Resolving this as we've decided that having the `normalizeSqlText` function exist outside the scope of `isIcebergCommand` could potentially lead to issues as the antlr parser should be responsible for actually processing the sqlText.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCallStatementParser.java
kbendick: Anyway, the tests have been added 🙂

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
kbendick: Resolving this but pulling the explanation up to a top level comment.

kbendick:
Explanation of the regex (which was buried in a comment):

The added regex has two parts

1. /\\*.*?\\*/ is for catching comments of the form /* …. */. We operate in (m) multi line mode to be able to catch the case where that comment spans multiple lines. The m allows for the . to match on a new line character.
2. --.*?\\n is for catching simple SQL comments, that start with two dashes and then consume the rest of the line they are on like -- some line terminating comment. We operate in (s) single line mod here. We don’t risk hitting multi-line mode here really, eg having a comment like that also catch the next line, because we explicitly search for the newline character (and there’s a test for that case).

All of this is separated by `|`, which acts as the OR.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
rdblue: What about structuring this differently?

It seems strange to me to use both multiline mode and single line mode in the same regex. Since we're removing all whitespace eventually anyway, why not remove single-line comments then remove whitespace, and finally remove multiline comments? That doesn't require multiline mode at all because we're removing newlines before processing comments.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
rdblue: Yeah, I agree that the table property issue is unrelated. I think it's fine either way. Given that we would only call normalizeSqlText in one place, I don't see a lot of value in separating it out just yet.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
kbendick: So I thought it seemed strange as well to use multi-line and single line mode as well, but removing single line mode actually made the following test case fail:

```sql
/* some multi-line
 comment */ CALL …
```

So both single line and multi-line mode are needed. And the multiline one like that, where the statement begins on the same line as the closing comment line, is what DBT is adding (that people would like to keep).

It could be done with multiple regexes, but it wasn’t much more readable and is also likely slower. Though the decision to use the regex as is was made based on the fact that it works, it’s been thoroughly tested, and it can be broke down into what’s needed and what’s not / explained relatively simply,

admittedly, the multiline / single line thing is kind of weird. But I think it would be needed with multiple regex too as the same regex needs both multiline mode and single line mode.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/IcebergSparkSqlExtensionsParser.scala
kbendick: Yeah it would be nice to do that as the parser already takes care of this, but I'm not quite sure how either.

I suspect that it's more work than it probably initially seems.

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #4397, opened 2022-03-24
Add support for listing partition recursively during the table migration

puchengy:
Data files that locate at the subdirectory in a Hive table's partition are invisible to Iceberg table migration. This diff solves that.

I have two questions:
1. I am thinking to make the recursive support turned off by default and make it configurable. What will be a good place to hold such config? Should this be a table property? or should this be an engine specific property?
2. should we add a unit test for this? if so, where I should add to. thanks.

### DIFF
--- a/data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
+++ b/data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.data;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
```diff
@@ -133,6 +132,20 @@ private TableMigrationUtil() {
     }
   }
 
+  private static List<FileStatus> listStatus(Configuration conf, String partitionUri) throws IOException {
```
RussellSpitzer, line 135: Since we have the conf here, couldn't we use the same option that Hive uses? I also think it is probably safe to just always be recursive?

RussellSpitzer: I'm suggesting here we can just use the Hadoop Configuration

RussellSpitzer:
I think the change here is good, but we definitely need some tests.

If you like you could put them in

https://github.com/apache/iceberg/blob/4d7837093bd19a42237446a9b656a248e106b789/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java#L412

data/src/main/java/org/apache/iceberg/data/TableMigrationUtil.java
kbendick: I don't love the idea of being recursive by default. We already know that the cost of listing in these situations is relatively high. Making them recursive by default seems like it would only make the relatively expensive list potentially more expensive.

But my biggest opposition is that I'm not sure about keeping recursive directories when importing the table to Iceberg. I guess if the metadata files have the correct file paths in them, then it's fine? I still worry there are spots we would miss by importing recursively though.

So I'm not opposed to adding it, but I would like for it to be opt-in.

For spark, there's also a config for reading file based dataframes with recursive lookup: `recursiveFileLookup`. https://spark.apache.org/docs/latest/sql-data-sources-generic-options.html#recursive-file-lookup

I know this is in `data` and not `spark`, but if we decide not to use some of the Hive based names, that might be a decent configuration option name.

PR #4401, opened 2022-03-25
Spark: add procedure to generate symlink manifests

jackye1995:
Add a Spark procedure to generate symlink manifests, so that systems without Iceberg support can read Iceberg table data using an external table:

```sql
CREATE EXTERNAL TABLE mytable ([(col_name1 col_datatype1, ...)])
[PARTITIONED BY (col_name2 col_datatype2, ...)]
ROW FORMAT SERDE 'org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe'
STORED AS INPUTFORMAT 'org.apache.hadoop.hive.ql.io.SymlinkTextInputFormat'
OUTPUTFORMAT 'org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat'
LOCATION '<symlink-table-root-path>' 
```

I did not add an action for this because this is to just give a gateway for users with any existing query engine that does not natively support Iceberg (in my case it's Redshift Spectrum) to start reading Iceberg, because most engines support Hive with symlink input format to some extent. If we think it deserves an action in core API, I can also add that.

The procedure looks like:

```
CALL catalog.system.generate_symlink_format_manifest(
  table => 'table_name', 
  symlink_root_location => 's3://some/path'
);
```

The `symlink_root_location` is optional. The default is `<table_root>/_symlink_format_manifest/<snapshot_id>`. A snapshot ID suffix is added because if this procedure is executed twice against the same table, we don't want to mix the results if the table is updated. If users want to use a consistent root path for the symlink table, it could be input as an override. 

I thought about adding another option for `snapshot_id` in the input, so we can generate a symlink table for any historical snapshots, but decided to not do that to avoid making the procedure too complicated. We can add it as a follow up if needed.

The procedure currently returns the `snapshot_id` that the procedure is executed against, and `data_file_count` for the number of data files in the symlink manifests.

Regarding partitioning, the generated symlink table exposes all the hidden partitions, and use the union of all historical table partition specs. For example, if the table is partitioned by spec1 `category`, spec 2 `bucket(16, id)`, users are expected to create a symlink table with `PARTITIONED BY (id_bucket int, category string)`.

Regarding merge-on-read, generated symlink table does not consider delete files. This is basically a "snapshot view" of the table. A compaction is needed to generate the most up-to-date view of the table.

### DIFF
--- a/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestGenerateSymlinkFormatManifestsProcedure.java
+++ b/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestGenerateSymlinkFormatManifestsProcedure.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.extensions;
+
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import org.apache.iceberg.AssertHelpers;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.exceptions.ValidationException;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestGenerateSymlinkFormatManifestsProcedure extends SparkExtensionsTestBase {
+
+  public TestGenerateSymlinkFormatManifestsProcedure(
+      String catalogName, String implementation, Map<String, String> config) {
+    super(catalogName, implementation, config);
+  }
+
+  @After
+  public void removeTables() throws Exception {
+    sql("DROP TABLE IF EXISTS %s", tableName);
+  }
+
+  @Test
+  public void testGenerateSymlinkFormatManifestsEmptyTable() {
+    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg", tableName);
+    AssertHelpers.assertThrows("Should not support generate symlink manifest for empty table",
+        ValidationException.class,
+        "Cannot generate symlink manifests for an empty table",
+        () -> sql("CALL %s.system.generate_symlink_format_manifest('%s')", catalogName, tableIdent));
+  }
+
+  @Test
+  public void testGenerateSymlinkFormatManifestsUnpartitioned() {
+    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1, 'a')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'b')", tableName);
+    List<Object[]> result = sql("CALL %s.system.generate_symlink_format_manifest('%s')", catalogName, tableIdent);
+    Table table = validationCatalog.loadTable(tableIdent);
+    List<Object[]> expected = Lists.newArrayList();
+    expected.add(row(table.currentSnapshot().snapshotId(), 2L));
+    assertEquals("Should find 2 files", expected, result);
+    checkDirectoryExists(table.location(), "_symlink_format_manifest");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()));
+  }
+
+  @Test
+  public void testGenerateSymlinkFormatManifestsPartitioned() {
+    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg PARTITIONED BY (data)", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1, 'a'), (1, 'b'), (1, 'c')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'b'), (2, 'c'), (2, 'd')", tableName);
+    List<Object[]> result = sql("CALL %s.system.generate_symlink_format_manifest('%s')", catalogName, tableIdent);
+    Table table = validationCatalog.loadTable(tableIdent);
+    List<Object[]> expected = Lists.newArrayList();
+    expected.add(row(table.currentSnapshot().snapshotId(), 6L));
+    assertEquals("Should find 6 files", expected, result);
+    checkDirectoryExists(table.location(), "_symlink_format_manifest");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()));
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data=a");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data=b");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data=c");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data=d");
+  }
+
+  @Test
+  public void testGenerateSymlinkFormatManifestsHiddenPartitioned() {
+    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg " +
+        "PARTITIONED BY (bucket(16, data))", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1, 'a'), (1, 'b'), (1, 'c')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'b'), (2, 'c'), (2, 'd')", tableName);
+    List<Object[]> result = sql("CALL %s.system.generate_symlink_format_manifest('%s')", catalogName, tableIdent);
+    Table table = validationCatalog.loadTable(tableIdent);
+    String location = table.location();
+    List<Object[]> expected = Lists.newArrayList();
+    expected.add(row(table.currentSnapshot().snapshotId(), 6L));
+    assertEquals("Should find 6 files", expected, result);
+    checkDirectoryExists(table.location(), "_symlink_format_manifest");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()));
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data_bucket=15");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data_bucket=2");
+    checkDirectoryExists(table.location(), "_symlink_format_manifest",
+        Long.toString(table.currentSnapshot().snapshotId()), "data_bucket=3");
+  }
+
+  @Test
+  public void testGenerateSymlinkFormatManifestsCustomLocationPositionArgument() {
+    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1, 'a')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'b')", tableName);
+    Table table = validationCatalog.loadTable(tableIdent);
+    String customLocation = table.location() + "/" + UUID.randomUUID();
+    List<Object[]> result = sql("CALL %s.system.generate_symlink_format_manifest('%s', '%s')",
+        catalogName, tableIdent, customLocation);
+    List<Object[]> expected = Lists.newArrayList();
+    expected.add(row(table.currentSnapshot().snapshotId(), 2L));
+    assertEquals("Should find 2 files", expected, result);
+    checkDirectoryExists(customLocation);
+  }
+
+  @Test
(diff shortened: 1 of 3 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/GenerateSymlinkFormatManifestProcedure.java
```diff
@@ -0,0 +1,142 @@
...
+
+    Types.StructType partitionType = Partitioning.partitionType(icebergTable);
+    Dataset<Row> entries = SparkTableUtil.loadCatalogMetadataTable(spark(), icebergTable, MetadataTableType.ENTRIES)
+        .filter("status < 2 AND data_file.content = 0");
+
+    long rowCount = entries.count();
+    if (partitionType.fields().isEmpty()) {
+      entries.select("data_file.file_path")
+          .write()
+          .format("parquet")
```
singhpk234, line 105: [question] As per this : https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/io/SymlinkTextInputFormat.java#L47-L52

**_Symlink file is a text file which contains a list of filename / dirname_**

does it works when Symlink file is in parquet format as well ? Am i missing something here

singhpk234, line 105: [question] should the format here be "text" instead ? As per this : https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/io/SymlinkTextInputFormat.java#L47-L52

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/GenerateSymlinkFormatManifestProcedure.java
```diff
@@ -0,0 +1,142 @@
...
+    ValidationException.check(icebergTable.currentSnapshot() != null,
+        "Cannot generate symlink manifests for an empty table");
+
+    long snapshotId = icebergTable.currentSnapshot().snapshotId();
+    String symlinkRootLocation = args.isNullAt(1) ?
+        defaultSymlinkManifestRootLocation(icebergTable.location(), snapshotId) : args.getString(1);
+
+    Types.StructType partitionType = Partitioning.partitionType(icebergTable);
+    Dataset<Row> entries = SparkTableUtil.loadCatalogMetadataTable(spark(), icebergTable, MetadataTableType.ENTRIES)
+        .filter("status < 2 AND data_file.content = 0");
```
singhpk234, line 99: we should cache the DF as well, otherwise we will end up scanning the metaDataTable twice. in :
1. count()
2. write()

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestGenerateSymlinkFormatManifestsProcedure.java
```diff
@@ -0,0 +1,153 @@
...
+    List<Object[]> expected = Lists.newArrayList();
+    expected.add(row(table.currentSnapshot().snapshotId(), 2L));
+    assertEquals("Should find 2 files", expected, result);
+    checkDirectoryExists(customLocation);
+  }
+
+  private void checkDirectoryExists(String... paths) {
+    String path = String.join("/", paths);
+    Assert.assertTrue("Directory should exist: " + path, Files.isDirectory(Paths.get(URI.create(path))));
+  }
```
singhpk234, line 152: should we also check the symlink folder when being used as the external table is return's same result as iceberg table ? WDYT

kbendick (commented):
Thanks for this addition @jackye1995!

> Regarding merge-on-read, generated symlink table does not consider delete files. This is basically a "snapshot view" of the table. A compaction is needed to generate the most up-to-date view of the table.

Should we error out if there are delete files present?

We can add an option to still allow for ignoring them to get this "snapshot view", but I think by default a symlink manifest shouldn't be generated that can represent a table at a state it never existed in.

For example, if we're using MOR, and we have one file that gets written as a plain.data file (as it only contains appends), and then some number of delta files, the current proposed symlink manifest based table _will_ contain that new file but won't contain the deltas.

Correct me if I'm wrong, but I think that would be a state the table was never in. I can see how it would avoid unnecessary compute to skip trying to process any delta files, e.g. if the user did compact it ahead of time, but I don't think that should be the default for the procedure. 

The default output should return a symlink that represents the table at least as it is at some point in time.

Ranjith-AR:
> Add a Spark procedure to generate symlink manifests, so that systems without Iceberg support can read Iceberg table data using an external table:
> 
> ```sql
> CREATE EXTERNAL TABLE mytable ([(col_name1 col_datatype1, ...)])
> [PARTITIONED BY (col_name2 col_datatype2, ...)]
> ROW FORMAT SERDE 'org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe'
> STORED AS INPUTFORMAT 'org.apache.hadoop.hive.ql.io.SymlinkTextInputFormat'
> OUTPUTFORMAT 'org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat'
> LOCATION '<symlink-table-root-path>' 
> ```
> 
> I did not add an action for this because this is to just give a gateway for users with any existing query engine that does not natively support Iceberg (in my case it's Redshift Spectrum) to start reading Iceberg, because most engines support Hive with symlink input format to some extent. If we think it deserves an action in core API, I can also add that.
> 
> The procedure looks like:
> 
> ```
> CALL catalog.system.generate_symlink_format_manifest(
>   table => 'table_name', 
>   symlink_root_location => 's3://some/path'
> );
> ```
> 
> The `symlink_root_location` is optional. The default is `<table_root>/_symlink_format_manifest/<snapshot_id>`. A snapshot ID suffix is added because if this procedure is executed twice against the same table, we don't want to mix the results if the table is updated. If users want to use a consistent root path for the symlink table, it could be input as an override.
> 
> I thought about adding another option for `snapshot_id` in the input, so we can generate a symlink table for any historical snapshots, but decided to not do that to avoid making the procedure too complicated. We can add it as a follow up if needed.
> 
> The procedure currently returns the `snapshot_id` that the procedure is executed against, and `data_file_count` for the number of data files in the symlink manifests.
> 
> Regarding partitioning, the generated symlink table exposes all the hidden partitions, and use the union of all historical table partition specs. For example, if the table is partitioned by spec1 `category`, spec 2 `bucket(16, id)`, users are expected to create a symlink table with `PARTITIONED BY (id_bucket int, category string)`.
> 
> Regarding merge-on-read, generated symlink table does not consider delete files. This is basically a "snapshot view" of the table. A compaction is needed to generate the most up-to-date view of the table.

**in the above default location "<table_root>/_symlink_format_manifest/<snapshot_id>", 
what is the name of the manifest file ? <snapshot_id>_manifest ? 

does this mean that when I regenerate the symlink_format_manifest, i have to alter the external table location ?

how to execute the procedure to override / exclude the snapshot_id prefix and generate manifest file like below
"<table_root>/_symlink_format_manifest/manifest" ?**

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/GenerateSymlinkFormatManifestProcedure.java
```diff
@@ -0,0 +1,142 @@
...
+  @Override
+  public String description() {
+    return "GenerateSymlinkFormatManifestProcedure";
+  }
+
+  private String defaultSymlinkManifestRootLocation(String tableLocation, long snapshotId) {
+    int len = tableLocation.length();
+    StringBuilder sb = new StringBuilder();
+    sb.append(tableLocation);
+    if (sb.charAt(len - 1) != '/') {
```
szehon-ho, line 133: Optional: can save one line above to get "len" by just checking tableLocation.endsWith().  (Not sure if there's any other reason to do it this way)

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/GenerateSymlinkFormatManifestProcedure.java
```diff
@@ -0,0 +1,142 @@
...
+  @Override
+  public StructType outputType() {
+    return OUTPUT_TYPE;
+  }
+
+  @Override
+  public InternalRow[] call(InternalRow args) {
+    String tableIdent = args.getString(0);
+    Preconditions.checkArgument(tableIdent != null && !tableIdent.isEmpty(),
+        "Cannot handle an empty identifier for argument table");
```
szehon-ho, line 83: Nit: maybe a slightly more user-friendly error message, "table cannot be null or empty"?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/GenerateSymlinkFormatManifestProcedure.java
```diff
@@ -0,0 +1,142 @@
...
+    CatalogPlugin defaultCatalog = spark().sessionState().catalogManager().currentCatalog();
+    Spark3Util.CatalogAndIdentifier catalogAndIdent = Spark3Util.catalogAndIdentifier(
+        PROCEDURE_CONTEXT, spark(), tableIdent, defaultCatalog);
+    SparkTable sparkTable = loadSparkTable(catalogAndIdent.identifier());
+    org.apache.iceberg.Table icebergTable = sparkTable.table();
+    ValidationException.check(icebergTable.currentSnapshot() != null,
+        "Cannot generate symlink manifests for an empty table");
+
+    long snapshotId = icebergTable.currentSnapshot().snapshotId();
+    String symlinkRootLocation = args.isNullAt(1) ?
```
szehon-ho, line 94: Curious, do we need to add any validation that path ends with /?  (Noticed we put a / at the end of the default path, not sure if that was a hard requirement)

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestGenerateSymlinkFormatManifestsProcedure.java
```diff
@@ -0,0 +1,153 @@
...
+
+  @Test
+  public void testGenerateSymlinkFormatManifestsUnpartitioned() {
+    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1, 'a')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'b')", tableName);
+    List<Object[]> result = sql("CALL %s.system.generate_symlink_format_manifest('%s')", catalogName, tableIdent);
+    Table table = validationCatalog.loadTable(tableIdent);
+    List<Object[]> expected = Lists.newArrayList();
```
szehon-ho, line 64: Nit: can we use Lists.newArrayList(row(table...))? 

(same for other tests)

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestGenerateSymlinkFormatManifestsProcedure.java
```diff
@@ -0,0 +1,153 @@
...
+    expected.add(row(table.currentSnapshot().snapshotId(), 2L));
+    assertEquals("Should find 2 files", expected, result);
+    checkDirectoryExists(customLocation);
+  }
+
+  private void checkDirectoryExists(String... paths) {
+    String path = String.join("/", paths);
+    Assert.assertTrue("Directory should exist: " + path, Files.isDirectory(Paths.get(URI.create(path))));
+  }
+}
```
szehon-ho, line 153: If not too hard, can we add one more test about evolving partition spec?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestGenerateSymlinkFormatManifestsProcedure.java
```diff
@@ -0,0 +1,153 @@
...
+    sql("INSERT INTO TABLE %s VALUES (1, 'a')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'b')", tableName);
+    Table table = validationCatalog.loadTable(tableIdent);
+    String customLocation = table.location() + "/" + UUID.randomUUID();
+    List<Object[]> result = sql("CALL %s.system.generate_symlink_format_manifest('%s', '%s')",
+        catalogName, tableIdent, customLocation);
+    List<Object[]> expected = Lists.newArrayList();
+    expected.add(row(table.currentSnapshot().snapshotId(), 2L));
+    assertEquals("Should find 2 files", expected, result);
+    checkDirectoryExists(customLocation);
```
szehon-ho, line 130: Can we add check (maybe in this test or another), of directory exists under customLocation for each partition?  To make sure that custom location works with partitioned table.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/GenerateSymlinkFormatManifestProcedure.java
```diff
@@ -0,0 +1,142 @@
...
+    return PARAMETERS;
+  }
+
+  @Override
+  public StructType outputType() {
+    return OUTPUT_TYPE;
+  }
+
+  @Override
+  public InternalRow[] call(InternalRow args) {
```
szehon-ho, line 80: Do you think it's beneficial to add an Action and have Procedure call it, like the other ones?  I am not sure.  cc @RussellSpitzer @aokolnychyi if any thoughts.

szehon-ho (commented):
Thanks, this is great to see this simple change.  Left a round of comments. 

I agree with @kbendick to add check if the table lacks delete file (or maybe, just support table v1 table initially).

amogh-jahagirdar:
> @jackye1995 Is there a way to do the same thing using Java API? How can we read the location for all of data files given tableId and snapshotId?

@prashantgohel1 You may have already gotten it, but  If you're looking for getting the data files for a table at a given snapshot, going through the Table APIs table.snapshot to get the snapshot, and then using snapshot's data manifest API to get all the data file manifests should work.https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/Snapshot.java#L101  Then it's a matter of using the Java library to read the manifest file.  https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/ManifestFiles.java#L71

Another way is to query the data files metadata table.

amogh-jahagirdar:
Discussed offline with @jackye1995 I'll be carrying this PR forward. So my thinking is the following:

1.) I do think we need to have an explicit flag for ignoring deletes, it makes sense to me that the operation fails if there are deletes. This is because as a user I expect the actual state of the table to be represented in the symlink file by default. If the operation fails, it does force a user to do a compaction prior to running this procedure but I think that still makes sense. And then if they really don't want this behavior they can pass in the flag to ignore deletes @kbendick @jackye1995 let me know your thinking or we can discuss on the PRs I plan on raising. 

2.) Looks like there's interest in having this be an actual Spark Action, I will add that. 

@jackye1995 Feel free to close this, I will add you as a co-author on the PRs I raise. Thanks!

PR #4402, opened 2022-03-25
AWS: Add table and namespace S3 tags

rajarshisarkar:
This change adds table level [S3 Tags](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html) to the objects while writing using `S3FileIO`. Users can pass these opt-in catalog properties to tag the objects in S3:

Spark SQL launch command:
```
    --conf spark.sql.catalog.my_catalog.warehouse=s3://<bucket>/s3-tagging \
    --conf spark.sql.catalog.my_catalog.s3.write.table-name-tag-enabled=true \
    --conf spark.sql.catalog.my_catalog.s3.write.namespace-name-tag-enabled=true
```

Tags added in S3:

```
aws s3api get-object-tagging --bucket <bucket> --key s3-tagging/data/00000-0-a81d170b-9d71-4b45-886a-390088b2513b-00001.parquet
{
    "TagSet": [
        {
            "Key": "iceberg.table",
            "Value": "test_table_tag"
        },
        {
            "Key": "iceberg.namespace",
            "Value": "test_db"
        }
    ]
}
```

---

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.iceberg.aws.dynamodb.DynamoDbCatalog;
+import org.apache.iceberg.aws.glue.GlueCatalog;
 import org.apache.iceberg.aws.lakeformation.LakeFormationAwsClientFactory;
 import org.apache.iceberg.aws.s3.S3FileIO;
 import org.apache.iceberg.exceptions.ValidationException;
@@ -395,6 +396,48 @@ public class AwsProperties implements Serializable {
   public static final String S3_WRITE_TAGS_PREFIX = "s3.write.tags.";
 
   /**
+   * Used by {@link GlueCatalog} to tag objects when writing. To set, we can pass a catalog
+   * property.
+   *
+   * <p>For more details, see
+   * https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   *
+   * <p>Example: s3.write.table-tag-enabled=true
+   */
+  public static final String S3_WRITE_TABLE_TAG_ENABLED = "s3.write.table-tag-enabled";
+
+  public static final boolean S3_WRITE_TABLE_TAG_ENABLED_DEFAULT = false;
+
+  /**
+   * Used by {@link GlueCatalog} to tag objects when writing. To set, we can pass a catalog
+   * property.
+   *
+   * <p>For more details, see
+   * https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   *
+   * <p>Example: s3.write.namespace-tag-enabled=true
+   */
+  public static final String S3_WRITE_NAMESPACE_TAG_ENABLED = "s3.write.namespace-tag-enabled";
+
+  public static final boolean S3_WRITE_NAMESPACE_TAG_ENABLED_DEFAULT = false;
+
+  /**
+   * Tag name that will be used by {@link #S3_WRITE_TAGS_PREFIX} when {@link
+   * #S3_WRITE_TABLE_TAG_ENABLED} is enabled
+   *
+   * <p>Example: iceberg.table=tableName
+   */
+  public static final String S3_TAG_ICEBERG_TABLE = "iceberg.table";
+
+  /**
+   * Tag name that will be used by {@link #S3_WRITE_TAGS_PREFIX} when {@link
+   * #S3_WRITE_NAMESPACE_TAG_ENABLED} is enabled
+   *
+   * <p>Example: iceberg.namespace=namespaceName
+   */
+  public static final String S3_TAG_ICEBERG_NAMESPACE = "iceberg.namespace";
+
+  /**
    * Used by {@link S3FileIO} to tag objects when deleting. When this config is set, objects are
    * tagged with the configured key-value pairs before deletion. This is considered a soft-delete,
    * because users are able to configure tag-based object lifecycle policy at bucket level to
@@ -498,6 +541,8 @@ public class AwsProperties implements Serializable {
   private ObjectCannedACL s3FileIoAcl;
   private boolean isS3ChecksumEnabled;
   private final Set<Tag> s3WriteTags;
+  private boolean s3WriteTableTagEnabled;
+  private boolean s3WriteNamespaceTagEnabled;
   private final Set<Tag> s3DeleteTags;
   private int s3FileIoDeleteThreads;
   private boolean isS3DeleteEnabled;
@@ -546,6 +591,8 @@ public class AwsProperties implements Serializable {
     this.s3fileIoStagingDirectory = System.getProperty("java.io.tmpdir");
     this.isS3ChecksumEnabled = S3_CHECKSUM_ENABLED_DEFAULT;
     this.s3WriteTags = Sets.newHashSet();
+    this.s3WriteTableTagEnabled = S3_WRITE_TABLE_TAG_ENABLED_DEFAULT;
+    this.s3WriteNamespaceTagEnabled = S3_WRITE_NAMESPACE_TAG_ENABLED_DEFAULT;
     this.s3DeleteTags = Sets.newHashSet();
     this.s3FileIoDeleteThreads = Runtime.getRuntime().availableProcessors();
     this.isS3DeleteEnabled = S3_DELETE_ENABLED_DEFAULT;
@@ -679,6 +726,12 @@ public class AwsProperties implements Serializable {
             "Deletion batch size must be between 1 and %s", S3FILEIO_DELETE_BATCH_SIZE_MAX));
 
     this.s3WriteTags = toS3Tags(properties, S3_WRITE_TAGS_PREFIX);
+    this.s3WriteTableTagEnabled =
+        PropertyUtil.propertyAsBoolean(
+            properties, S3_WRITE_TABLE_TAG_ENABLED, S3_WRITE_TABLE_TAG_ENABLED_DEFAULT);
+    this.s3WriteNamespaceTagEnabled =
+        PropertyUtil.propertyAsBoolean(
+            properties, S3_WRITE_NAMESPACE_TAG_ENABLED, S3_WRITE_NAMESPACE_TAG_ENABLED_DEFAULT);
     this.s3DeleteTags = toS3Tags(properties, S3_DELETE_TAGS_PREFIX);
     this.s3FileIoDeleteThreads =
         PropertyUtil.propertyAsInt(
@@ -854,6 +907,22 @@ public class AwsProperties implements Serializable {
 
   public Set<Tag> s3WriteTags() {
     return s3WriteTags;
+  }
+
+  public boolean s3WriteTableTagEnabled() {
+    return s3WriteTableTagEnabled;
+  }
+
+  public void setS3WriteTableTagEnabled(boolean s3WriteTableNameTagEnabled) {
+    this.s3WriteTableTagEnabled = s3WriteTableNameTagEnabled;
+  }
+
+  public boolean s3WriteNamespaceTagEnabled() {
+    return s3WriteNamespaceTagEnabled;
+  }
+
+  public void setS3WriteNamespaceTagEnabled(boolean s3WriteNamespaceTagEnabled) {
+    this.s3WriteNamespaceTagEnabled = s3WriteNamespaceTagEnabled;
   }
 
   public Set<Tag> s3DeleteTags() {
--- a/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
@@ -199,20 +199,30 @@ public class GlueCatalog extends BaseMetastoreCatalog
   protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
     if (catalogProperties != null) {
       ImmutableMap.Builder<String, String> tableSpecificCatalogPropertiesBuilder =
-          ImmutableMap.<String, String>builder()
-              .putAll(catalogProperties)
-              .put(
-                  AwsProperties.LAKE_FORMATION_DB_NAME,
-                  IcebergToGlueConverter.getDatabaseName(
-                      tableIdentifier, awsProperties.glueCatalogSkipNameValidation()))
-              .put(
-                  AwsProperties.LAKE_FORMATION_TABLE_NAME,
-                  IcebergToGlueConverter.getTableName(
-                      tableIdentifier, awsProperties.glueCatalogSkipNameValidation()));
+          ImmutableMap.<String, String>builder().putAll(catalogProperties);
+      boolean skipNameValidation = awsProperties.glueCatalogSkipNameValidation();
+
+      if (awsProperties.s3WriteTableTagEnabled()) {
+        tableSpecificCatalogPropertiesBuilder.put(
+            AwsProperties.S3_WRITE_TAGS_PREFIX.concat(AwsProperties.S3_TAG_ICEBERG_TABLE),
+            IcebergToGlueConverter.getTableName(tableIdentifier, skipNameValidation));
+      }
+
+      if (awsProperties.s3WriteNamespaceTagEnabled()) {
+        tableSpecificCatalogPropertiesBuilder.put(
+            AwsProperties.S3_WRITE_TAGS_PREFIX.concat(AwsProperties.S3_TAG_ICEBERG_NAMESPACE),
+            IcebergToGlueConverter.getDatabaseName(tableIdentifier, skipNameValidation));
+      }
 
       if (awsProperties.glueLakeFormationEnabled()) {
-        tableSpecificCatalogPropertiesBuilder.put(
-            AwsProperties.S3_PRELOAD_CLIENT_ENABLED, String.valueOf(true));
+        tableSpecificCatalogPropertiesBuilder
+            .put(
+                AwsProperties.LAKE_FORMATION_DB_NAME,
+                IcebergToGlueConverter.getDatabaseName(tableIdentifier, skipNameValidation))
+            .put(
+                AwsProperties.LAKE_FORMATION_TABLE_NAME,
+                IcebergToGlueConverter.getTableName(tableIdentifier, skipNameValidation))
+            .put(AwsProperties.S3_PRELOAD_CLIENT_ENABLED, String.valueOf(true));
       }
 
       // FileIO initialization depends on tableSpecificCatalogProperties, so a new FileIO is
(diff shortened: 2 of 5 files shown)


danielcweeks:
I think it would be good to discuss what the use case is for this. I can understand trying to annotate some additional information but I'm not sure that this tracks very well with the evolution of tables. If you rename a table now all of the objects are out of sync.  This also isn't particularly helpful for recovering table data because you actually don't know based on the file what manifests they belong to and that of course can change as well.  

Because iceberg metadata stems from the metadata file all the way down to the physical file, I'm not sure there's any reason to do this that I can think of.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -342,11 +372,17 @@
...
+  private boolean s3WriteTableNameTagEnabled;
+  private boolean s3WriteNamespaceTagEnabled;
   private final Map<String, String> s3BucketToAccessPointMapping;
 
   private String glueCatalogId;
   private boolean glueCatalogSkipArchive;
   private boolean glueLakeFormationEnabled;
+  private String tableName;
+  private String namespace;
+
```
jackye1995, line 385: nit: extra newline

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -388,6 +429,8 @@ public AwsProperties(Map<String, String> properties) {
     this.glueLakeFormationEnabled = PropertyUtil.propertyAsBoolean(properties,
         GLUE_LAKEFORMATION_ENABLED,
         GLUE_LAKEFORMATION_ENABLED_DEFAULT);
+    this.tableName = properties.get(LAKE_FORMATION_TABLE_NAME);
```
jackye1995, line 432: we should not get the table and namespace name here, this should be retrieved when creating `newTableOps`, otherwise it's always going to be null

rajarshisarkar: Few doubts:
1. For table level S3 tags, do we need to create new FileIOs for each `newTableOps` call (when LF is not enabled)? The current condition is:
```
protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
    if (catalogProperties != null) {
    ...
    }
```
2. The `tableIdentifier` doesn't have table UUID right now. Should I add the same in the table identifier object?

rajarshisarkar:
> I think it would be good to discuss what the use case is for this. I can understand trying to annotate some additional information but I'm not sure that this tracks very well with the evolution of tables. If you rename a table now all of the objects are out of sync.

Yes, only the table uuid tag would in sync for this scenario. @jackye1995 on the use case.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -357,6 +388,18 @@ public class AwsProperties implements Serializable {
    */
   public static final String LAKE_FORMATION_DB_NAME = "lakeformation.db-name";
 
+  /**
+   * Used by {@link GlueCatalog}.
+   * The table name is used as part of table specific catalog properties.
```
jackye1995, line 393: nit: as a part of

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -357,6 +388,18 @@ public class AwsProperties implements Serializable {
...
+   */
+  public static final String TABLE_NAME = "table-name";
+
+  /**
+   * Used by {@link GlueCatalog}.
+   * The database name is used as part of table specific catalog properties.
```
jackye1995, line 399: nit: as a part of

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -329,6 +330,36 @@ public class AwsProperties implements Serializable {
...
+  public static final boolean S3_WRITE_TABLE_NAME_TAG_ENABLED_DEFAULT = false;
+
+  /**
+   * Used by {@link S3FileIO} to tag objects when writing. To set, we can pass a catalog property.
+   * <p>
+   * For more details, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   * <p>
+   * Example: s3.write.namespace-tag-enabled=true
+   */
+  public static final String S3_WRITE_NAMESPACE_TAG_ENABLED = "s3.write.namespace-tag-enabled";
```
jackye1995, line 360: should this be `namespace-name`?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -371,11 +414,16 @@ public class AwsProperties implements Serializable {
...
+  private boolean s3WriteTableUuidTagEnabled;
+  private boolean s3WriteTableNameTagEnabled;
+  private boolean s3WriteNamespaceTagEnabled;
   private final Map<String, String> s3BucketToAccessPointMapping;
 
   private String glueCatalogId;
   private boolean glueCatalogSkipArchive;
   private boolean glueLakeFormationEnabled;
+  private String tableName;
+  private String namespace;
```
jackye1995, line 426: similarly, namespaceName

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -357,6 +388,18 @@ public class AwsProperties implements Serializable {
    */
   public static final String LAKE_FORMATION_DB_NAME = "lakeformation.db-name";
 
+  /**
+   * Used by {@link GlueCatalog}.
+   * The table name is used as part of table specific catalog properties.
+   */
+  public static final String TABLE_NAME = "table-name";
```
jackye1995, line 395: 1. The `table-name` and `db-name` properties are quite hacky... why not just dynamically add the tags based on catalog tag prefix `s3.write.tags.`? For example, if table name tag enabled, add property `s3.write.tags.table-name` in Glue catalog.
2. the tag names themselves, I think we should probably prefix them by `iceberg.`, for example `iceberg.table-name`, `iceberg.namespace-name`.
3. the variable name should really indicate this is a special tag name, for example `S3_TABLE_SPECIFIC_WRITE_TAGS_TABLE_NAME`.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -329,6 +330,36 @@ public class AwsProperties implements Serializable {
...
   public static final boolean S3_DELETE_ENABLED_DEFAULT = true;
 
+  /**
+   * Used by {@link S3FileIO} to tag objects when writing. To set, we can pass a catalog property.
+   * <p>
+   * For more details, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   * <p>
+   * Example: s3.write.table-uuid-tag-enabled=true
+   */
+  public static final String S3_WRITE_TABLE_UUID_TAG_ENABLED = "s3.write.table-uuid-tag-enabled";
```
jackye1995, line 340: Using UUID as tag might be a bit tricker, because when creating table you don't know the table UUID yet when you create the table operation. So I would suggest we add table and namespace name tags first, and add UUID in a separated PR.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -110,13 +110,15 @@ public GlueCatalog() {
   @Override
   public void initialize(String name, Map<String, String> properties) {
     AwsClientFactory awsClientFactory;
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.<String, String>builder().putAll(properties);
```
jackye1995, line 113: I think we can make this a `tableSpecificPropertiesBuilder`, and at the end if this builds an empty map it means there is no custom table specific catalog properties, and we should use the else clause.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -197,6 +199,10 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
               IcebergToGlueConverter.getDatabaseName(tableIdentifier))
           .put(AwsProperties.LAKE_FORMATION_TABLE_NAME,
               IcebergToGlueConverter.getTableName(tableIdentifier))
+          .put(AwsProperties.DB_NAME,
```
jackye1995, line 202: this is what I meant by we can reuse `s3.write.tags.` property, if we just add tags in that way we directly leverage existing logic.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
rajarshisarkar: This is not needed anymore as we are directly leveraging existing logic.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
rajarshisarkar: Makes sense to me.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
rajarshisarkar: I have made the changes.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: I have made the changes to leverage the existing logic.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: [Doubt] If the user passes only the table level tag properties, then we will build the map and according to this logic we would skip the else block. Thereby not initialising `awsClientFactory` and `catalogFileIO` which should not be the intended behaviour.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -357,6 +380,18 @@ public class AwsProperties implements Serializable {
    */
   public static final String LAKE_FORMATION_DB_NAME = "lakeformation.db-name";
 
+  /**
+   * Used by {@link GlueCatalog}.
+   * The table name is used as a part of table specific catalog properties.
+   */
+  public static final String S3_TABLE_SPECIFIC_WRITE_TAGS_TABLE_NAME = "iceberg.table-name";
```
jackye1995, line 387: looks like this should be moved to under `S3_WRITE_TAGS_PREFIX`? And we can have a simpler name like `S3_TAG_TABLE_NAME`

rajarshisarkar: Thanks, I have made the change.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -370,6 +371,44 @@ public class AwsProperties implements Serializable {
...
+   * https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   *
+   * <p>Example: s3.write.namespace-name-tag-enabled=true
+   */
+  public static final String S3_WRITE_NAMESPACE_TAG_ENABLED = "s3.write.namespace-name-tag-enabled";
+
+  public static final boolean S3_WRITE_NAMESPACE_TAG_ENABLED_DEFAULT = false;
+
+  /**
+   * Used by {@link GlueCatalog}. The table name is used as a part of table specific catalog
```
jackye1995, line 401: we should note this is the tag name that will be used by `{@link S3_WRITE_TAGS_PREFIX}`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -355,4 +355,8 @@ void cleanupMetadataAndUnlock(CommitStatus commitStatus, String metadataLocation
       }
     }
   }
+
+  public Map<String, String> tableCatalogProperties() {
```
jackye1995, line 359: looks like this is only for testing purpose? can we use spy for that instead of exposing it?

aws/src/test/java/org/apache/iceberg/aws/glue/TestGlueCatalog.java
```diff
@@ -585,4 +585,32 @@ public void testValidateIdentifierSkipNameValidation() {
         CATALOG_NAME, WAREHOUSE_PATH, props, glue, LockManagers.defaultLockManager(), null);
     Assert.assertEquals(glueCatalog.isValidIdentifier(TableIdentifier.parse("db-1.a-1")), true);
   }
+
+  @Test
+  public void testTableLevelS3TagProperties() {
+    Map<String, String> properties =
+        ImmutableMap.of(
+            "s3.write.table-name-tag-enabled",
```
jackye1995, line 593: nit: can import from `AwsProperties`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -199,20 +200,31 @@ void initialize(
...
-                  AwsProperties.LAKE_FORMATION_DB_NAME,
-                  IcebergToGlueConverter.getDatabaseName(
-                      tableIdentifier, awsProperties.glueCatalogSkipNameValidation()))
-              .put(
-                  AwsProperties.LAKE_FORMATION_TABLE_NAME,
-                  IcebergToGlueConverter.getTableName(
-                      tableIdentifier, awsProperties.glueCatalogSkipNameValidation()));
+          ImmutableMap.<String, String>builder().putAll(catalogProperties);
+
+      if (isTableLevelTagEnabled(catalogProperties)) {
```
jackye1995, line 205: we already have `awsProperties` here, no need to parse property map again

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -113,10 +113,11 @@ public void initialize(String name, Map<String, String> properties) {
     this.catalogProperties = ImmutableMap.copyOf(properties);
     AwsClientFactory awsClientFactory;
     FileIO catalogFileIO;
-    if (PropertyUtil.propertyAsBoolean(
-        properties,
-        AwsProperties.GLUE_LAKEFORMATION_ENABLED,
-        AwsProperties.GLUE_LAKEFORMATION_ENABLED_DEFAULT)) {
+    if (isTableLevelTagEnabled(properties)) {
```
jackye1995, line 116: why is this needed?

jackye1995:
Thanks for updating this, could you also add an integration test to verify files written to a table are properly tagged in s3?

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
rajarshisarkar: Shouldn't be needed, I missed that `catalogProperties` was initialised as `ImmutableMap.copyOf(properties);`

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -394,6 +395,48 @@ public class AwsProperties implements Serializable {
...
+  /**
+   * Used by {@link GlueCatalog} to tag objects when writing. To set, we can pass a catalog
+   * property.
+   *
+   * <p>For more details, see
+   * https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
+   *
+   * <p>Example: s3.write.table-name-tag-enabled=true
+   */
+  public static final String S3_WRITE_TABLE_NAME_TAG_ENABLED = "s3.write.table-name-tag-enabled";
```
jackye1995, line 407: I am wondering if we should simplify these config names to `s3.write.table-tag-enabled`, `s3.write.namespace-tag-enabled`, `iceberg.table`, `iceberg.namespace`, what do you think?

rajarshisarkar: Yes, seems clean to me. I have made the changes.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -394,6 +395,48 @@ public class AwsProperties implements Serializable {
...
+
+  public static final boolean S3_WRITE_NAMESPACE_TAG_ENABLED_DEFAULT = false;
+
+  /**
+   * Tag name that will be used by {@link #S3_WRITE_TAGS_PREFIX} when {@link
+   * #S3_WRITE_TABLE_TAG_ENABLED} is enabled
+   *
+   * <p>Example: iceberg.table=tableName
+   */
+  public static final String S3_TAG_TABLE = "iceberg.table";
```
jackye1995, line 430: what about `S3_TAG_ICEBERG_TABLE`? just to match the value

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -498,6 +541,8 @@ public class AwsProperties implements Serializable {
   private ObjectCannedACL s3FileIoAcl;
   private boolean isS3ChecksumEnabled;
   private final Set<Tag> s3WriteTags;
+  private boolean s3WriteTableNameTagEnabled;
```
jackye1995, line 544: can we match the config name and name this `s3WriteTableTagEnabled`?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -856,6 +909,22 @@ public Set<Tag> s3WriteTags() {
     return s3WriteTags;
   }
 
+  public boolean s3WriteTableNameTagEnabled() {
```
jackye1995, line 912: same here for the method name matching config name

jackye1995 (approved):
looks good to me!

jackye1995:
cc @singhpk234 @amogh-jahagirdar for any comment

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -199,20 +199,33 @@ void initialize(
...
-                  IcebergToGlueConverter.getTableName(
-                      tableIdentifier, awsProperties.glueCatalogSkipNameValidation()));
+          ImmutableMap.<String, String>builder().putAll(catalogProperties);
 
-      if (awsProperties.glueLakeFormationEnabled()) {
+      if (awsProperties.s3WriteTableTagEnabled()) {
+        tableSpecificCatalogPropertiesBuilder.put(
+            AwsProperties.S3_WRITE_TAGS_PREFIX.concat(AwsProperties.S3_TAG_ICEBERG_TABLE),
+            IcebergToGlueConverter.getTableName(
+                tableIdentifier, awsProperties.glueCatalogSkipNameValidation()));
```
singhpk234, line 208: [minor] can extract this into local vars to be reused in both places.

singhpk234 (approved):
LGTM as well !!

amogh-jahagirdar (approved):
Overall LGTM, thanks!

jackye1995:
Thanks for finishing the work @rajarshisarkar ! and thanks @amogh-jahagirdar and @singhpk234 for the reviews!

PR #4404, opened 2022-03-25
[Spec] Clarify whether floating point lower and upper bounds can contain NaNs

emkornfield:
This is somewhat pedantic.  Based on current wording:
`Each value must be less than or equal to all non-null, non-NaN values in the column for the file [2]`

Based on defined sorting statistics
`Sorting floating-point numbers should produce the following behavior: -NaN < -Infinity < -value < -0 < 0 < value < Infinity < NaN` -NaN could be a lower bound and NaN could be an upper bound.

This PR tries to clarify whether statistics can or can't used for lower and upper bounds (I think the first sentence quoted might have been intended to rule out NaN but it isn't clear, so I'm happy to reverse this language to state that +/- NaN are valid values for bounds if that is the case).

### DIFF
--- a/format/spec.md
+++ b/format/spec.md
@@ -423,7 +423,7 @@ Notes:
 
 1. Single-value serialization for lower and upper bounds is detailed in Appendix D.
 2. For `float` and `double`, the value `-0.0` must precede `+0.0`, as in the IEEE 754 `totalOrder` predicate. NaNs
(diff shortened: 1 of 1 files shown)


rdblue:
You're correct. NaN values are not allowed in lower or upper bounds. Thanks!

rdblue (approved):

PR #4245, opened 2022-03-01
Core - Add in an http implementation of the RESTClient interface

kbendick:
This adds an HTTP implementation of the `RESTClient` interface.

### Testing
We use `mock-server` to test each method that we use: `post`, `head`, `get`, and `delete` (there are all the verbs that are used in the spec so far).

This tests that the HTTP client can parse and return the expected response type, as well as whether or not it calls the error handlers for any error response codes (400+). It uses a real HTTP connection by setting a set of expectations on a mock server.

The mock-server also allows for taking in an OpenAPI document, so in the future we can also add API specification based testing directly in the code, but for now, we're just interested in testing the HTTP client as catalog specific tests will come in another PR.

### Notes
At present, it's relatively simple and more of an MVP. Things like connection pooling, client timeouts, etc can be handled in follow ups, but I didn't want to make this PR too large to begin with (and the RESTClient isn't exactly high throughput as it only sends the request / response objects at present).

I also didn't relocate the client, but we might want to do that in resulting runtime / uber-jars.

**Some things to consider for the future**:
1) Max payload size / pagination - Some people have tables with 100+ columns, and fetching metadata for those tables in bulk might result in a large payload. Right now, the user can pass in their own `HttpClient`, and configure whatever they want on that. Otherwise the default apache http5 client is used.
2) Retries - right now, there's no additional configuration for retries. It's possible we'll want to handle retries at the catalog level in some cases, but I figured this could be added as an option after the first pass of reviews is done / this MVP is completed.

PR #4405, opened 2022-03-25
Core - Add marker interfaces for both RESTRequests and RESTResponses

kbendick:
Adds three empty interfaces, that are used for marking classes that correspond to REST requests / responses.
1) `RESTMessage` - This marks either a RESTRequest or a RESTResponse. The name is chosen to match many libraries, like Netty, where the `HTTPRequest` extends `HTTPMessage`.
2) `RESTRequest` - Specific empty subinterface for classes that correlate to REST catalog requests.
3) `RESTResponse` - Specific empty subinterface for classes that correlate to REST catalog resposnes.

Right now, these are empty but in the future we could add things like the `validate` method, or anything else that is needed. Eventually we could possibly even move these into the api module.

For now, this is in response to [a comment here](https://github.com/apache/iceberg/pull/4245#discussion_r832570043) so that we can specify that the generic type going into a method is in fact a REST catalog request and that what's returned is a REST catalog response.

cc @rdblue

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
+++ b/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
@@ -201,17 +201,20 @@ public class HTTPClient implements RESTClient {
   }
 
   @Override
-  public <T> T get(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+  public <T extends RESTResponse> T get(String path, Class<T> responseType,
+                                        Consumer<ErrorResponse> errorHandler) {
     return execute(Method.GET, path, null, responseType, errorHandler);
   }
 
   @Override
-  public <T> T post(String path, Object body, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+  public <T extends RESTResponse> T post(String path, RESTRequest body, Class<T> responseType,
+                                         Consumer<ErrorResponse> errorHandler) {
     return execute(Method.POST, path, body, responseType, errorHandler);
   }
 
   @Override
-  public <T> T delete(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+  public <T extends RESTResponse> T delete(String path, Class<T> responseType,
+                                           Consumer<ErrorResponse> errorHandler) {
     return execute(Method.DELETE, path, null, responseType, errorHandler);
   }
 
--- a/core/src/main/java/org/apache/iceberg/rest/RESTMessage.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTMessage.java
@@ -19,22 +19,16 @@
 
 package org.apache.iceberg.rest;
 
-import java.io.Closeable;
-import java.util.function.Consumer;
-import org.apache.iceberg.rest.responses.ErrorResponse;
-
 /**
- * Interface for a basic HTTP Client for interfacing with the REST catalog.
+ * Interface to mark both REST requests and responses.
  */
-public interface RESTClient extends Closeable {
+public interface RESTMessage {
 
-  void head(String path, Consumer<ErrorResponse> errorHandler);
-
-  <T extends RESTResponse> T delete(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler);
-
-  <T extends RESTResponse> T get(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler);
-
-  <T extends RESTResponse> T post(String path, RESTRequest body, Class<T> responseType,
-                                  Consumer<ErrorResponse> errorHandler);
+  /**
+   * Ensures that a constructed instance of a REST message is valid according to the REST spec.
+   * <p>
+   * This is needed when parsing data that comes from external sources and the object might have
+   * been constructed without all the required fields present.
+   */
+  void validate();
 }
-
--- a/core/src/main/java/org/apache/iceberg/rest/RESTResponse.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTResponse.java
@@ -19,22 +19,8 @@
 
 package org.apache.iceberg.rest;
 
-import java.io.Closeable;
-import java.util.function.Consumer;
-import org.apache.iceberg.rest.responses.ErrorResponse;
-
 /**
- * Interface for a basic HTTP Client for interfacing with the REST catalog.
+ * Interface to mark a REST response
  */
-public interface RESTClient extends Closeable {
-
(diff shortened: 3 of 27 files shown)


core/src/main/java/org/apache/iceberg/rest/RESTMessage.java
```diff
@@ -0,0 +1,26 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
+
+/**
+ * Interface to mark both REST requests and responses.
+ */
+public interface RESTMessage {
+}
```
kbendick, line 26: If we want to simplify, we can remove this and then have `RESTRequest` and `RESTResponse` be entirely independent interfaces for the moment.

But I think they should be tied together, as they share the same structure overall.

rdblue: I'd probably add the `validate` method now. Otherwise, I'm not sure what the value of this interface is.

rdblue (changes requested):
Should this also update the `RESTClient` interface?

core/src/main/java/org/apache/iceberg/rest/RESTMessage.java
kbendick: Well, the interface would still have value for type restraints even if it were empty.

But I agree that `validate` is good to include now.

kbendick:
> Should this also update the `RESTClient` interface?

Yes. Also `HTTPClient`. Plus associated tests classes such as `RESTCatalogAdapter`.

core/src/main/java/org/apache/iceberg/rest/RESTClient.java
```diff
@@ -27,9 +27,14 @@
...
-  void  head(String path, Consumer<ErrorResponse> errorHandler);
+
+  void head(String path, Consumer<ErrorResponse> errorHandler);
+
+  <T extends RESTResponse<T>> T delete(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler);
+
+  <T extends RESTResponse<T>> T get(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler);
+
+  <T extends RESTResponse<T>> T post(
+      String path, Object body, Class<T> responseType, Consumer<ErrorResponse> errorHandler);
```
kbendick, line 38: Style wise, I wasn't really sure the best way to deal with these. Specifically because `post` would require multiple lines for the arguments.

I moved `head` up because it returns `void` now. 

Please feel free to make suggestions @rdblue

core/src/main/java/org/apache/iceberg/rest/RESTResponse.java
```diff
@@ -0,0 +1,26 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
+
+/**
+ * Interface to mark a REST response
+ */
+public interface RESTResponse extends RESTMessage<RESTResponse> {
+}
```
kbendick, line 26: Initially, I had this as `RESTResponse<T> extends RESTMessage<T>`.

Arguably that's more type safe, but it required _every single_ request / response object to redeclare itself twice.

I think it's sufficient to just have type checking on the fact that it's a request or a response, as it would be unlikely to define a class with a `validate` method that returns some other `RESTRequest` or `RESTResponse`.

If we think this trade off isn't worth it, I can revert the change.

core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
```diff
@@ -201,17 +201,20 @@ public void head(String path, Consumer<ErrorResponse> errorHandler) {
...
   @Override
-  public <T> T get(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+  public <T extends RESTResponse> T get(String path, Class<T> responseType,
+                                        Consumer<ErrorResponse> errorHandler) {
     return execute(Method.GET, path, null, responseType, errorHandler);
   }
 
   @Override
-  public <T> T post(String path, Object body, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+  public <T extends RESTResponse> T post(String path, Object body, Class<T> responseType,
```
rdblue, line 210: Should body be something that extends `RESTRequest`?

core/src/main/java/org/apache/iceberg/rest/RESTMessage.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.apache.iceberg.rest;
+
+/**
+ * Interface to mark both REST requests and responses.
+ */
+public interface RESTMessage<T> {
+
+  /**
+   * Ensures that a constructed instance of a REST message meets all of its requirements.
```
rdblue, line 28: Minor: "all of its requirements" is a bit vague. You may want to be more specific, like "is valid according to the REST spec"?

core/src/main/java/org/apache/iceberg/rest/RESTResponse.java
rdblue: Yeah, I see what you mean.

I think I'd probably change `validate` to return `void` instead of `T` so that we don't need a type. As you pointed out, it's annoying to add every class as a parameter to the superclass. And the benefit isn't that large. I like being able to use `return msg.validate()` but you can also use separate statements quite easily.

The drawback to the current approach is that `msg.validate()` actually returns `RESTResponse` instead of the actual type, so you still can't return the results in most cases. I would opt for the simple version. We can safely add the type paramters and return value later.

rdblue (changes requested):
Mostly looks good. I would just remove the type param from RESTMessage.

core/src/main/java/org/apache/iceberg/rest/RESTResponse.java
kbendick: Agreed. With the current approach, you can return `msg.validate()` only if the return type is explicitly stated as the correct subtype, and it's up to the user to specify it correctly. So it's not providing much value.

Will update to return `void` there and remove the type parameter.

core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
kbendick: Yeah. I used just `RESTRequest` here, but not certain if it needs to be another type variable here for post of like `<Req extends RESTRequst>`.

I think just using `RESTRequest` is fine.

rdblue (approved):

rdblue:
Thanks, @kbendick!

Issue #4406, opened 2022-03-25
HiveCatalog: renameTable recursively lists file status


RussellSpitzer:
While #734 fixes the recursive file status listing for HiveTableOperations, we still trigger this in HiveCatalog.renameTable https://github.com/apache/iceberg/blob/d9296ca2d1b1143e6bbcdd1004962e653400489d/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L211

singhpk234:
based on this discussion https://github.com/apache/iceberg/pull/734#issuecomment-574062407
Looks like passing this env context to the alter table should be sufficient to avoid recursive listing here as well : 
```java
          EnvironmentContext envContext = new EnvironmentContext(
              ImmutableMap.of(StatsSetupConst.DO_NOT_UPDATE_STATS, StatsSetupConst.TRUE)
          );
```

can I pick this up ?

RussellSpitzer:
Sure I am way behind on reviews anyway :)

PR #734, opened 2020-01-12
HIVE: On alter table prevent hive from recursively scanning the complete table dir

alexey-fin:

PR #4407, opened 2022-03-25
Hive: Avoid recursive listing in HiveCatalog#renameTable

singhpk234:
Attempts to avoid recursive listing in HiveCatalog#renameTable

ref : https://github.com/apache/iceberg/pull/734#issuecomment-574062407

--- 

Closes https://github.com/apache/iceberg/issues/4406

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
@@ -26,10 +26,12 @@ import java.util.stream.Collectors;
 import org.apache.hadoop.conf.Configurable;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.common.StatsSetupConst;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.AlreadyExistsException;
 import org.apache.hadoop.hive.metastore.api.Database;
+import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
 import org.apache.hadoop.hive.metastore.api.InvalidOperationException;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Table;
@@ -46,6 +48,7 @@ import org.apache.iceberg.TableOperations;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.SupportsNamespaces;
 import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.common.DynMethods;
 import org.apache.iceberg.exceptions.NamespaceNotEmptyException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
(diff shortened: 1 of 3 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -64,6 +68,14 @@
 public class HiveCatalog extends BaseMetastoreCatalog implements SupportsNamespaces, Configurable {
   public static final String LIST_ALL_TABLES = "list-all-tables";
   public static final String LIST_ALL_TABLES_DEFAULT = "false";
+  private static final DynMethods.UnboundMethod ALTER_TABLE = DynMethods.builder("alter_table")
```
RussellSpitzer, line 71: I did kinda wonder if we should extract this to a package utility method ? But this seems fine with me if you want to go this route. It's only used in 2 places so I think this is fine

RussellSpitzer (approved):

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
singhpk234: +1, Agree with you on this, I think we can move this to the existing [MetaStoreUtil](https://github.com/apache/iceberg/blob/master/hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java) in the package WDYT ?

hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
```diff
@@ -37,6 +53,17 @@ public static boolean hive3PresentOnClasspath() {
     return HIVE3_PRESENT_ON_CLASSPATH;
   }
 
+  /**
+   * Invokes alterTable via metaStoreClient without triggering recursive listing.
+   * ref: https://github.com/apache/iceberg/pull/734#issuecomment-574062407
```
rdblue, line 58: Rather than referencing the issue, it would be more clear to just state the resolution: if possible, call the alter_table method with an environment context that turns off stats updates.

rdblue (changes requested):
I think this is good, but should have a minor update to docs so that the change is clear without going to an issue.

hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
```diff
@@ -54,8 +54,8 @@ public static boolean hive3PresentOnClasspath() {
   }
 
   /**
-   * Invokes alterTable via metaStoreClient without triggering recursive listing.
-   * ref: https://github.com/apache/iceberg/pull/734#issuecomment-574062407
+   * If possible, call the alter_table method via metaStoreClient with an environment context that turns off stats
```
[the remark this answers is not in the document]
singhpk234: > Nits: Why "if possible"? The final comma is also not needed so

It's due to this signature which is added as final fallback 
```java
      .impl(IMetaStoreClient.class, "alter_table",
          String.class, String.class, Table.class)
```

> "The final comma is also not needed so" 

makes sense.

should I rephrase it to : 

" If possible, calls alter_table method using the metastore client with an environmental context that turns off stats update to avoid recursive file listing"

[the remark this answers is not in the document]
RussellSpitzer: The "if possible" here applies to the whole sentence which makes it sound like this may call alter_table or it may not. So maybe instead
"Calls alter_table method using the metastore client. If possible an environmental context will be used that ...."

RussellSpitzer (approved):

rdblue (approved):

RussellSpitzer:
Thanks @singhpk234 for the PR @rdblue for helping out with review! Merged

PR #4328, opened 2022-03-15
Spark - Update commands for running JMH tests with Spark 3.2 JMH

kbendick:
The javadoc for the JMH tests references the old syntax, from prior to Iceberg 0.13.0 when we separated the code base by Spark versions.

Updating this as it came up in another PR.

PR #4409, opened 2022-03-26
Spark: Update Spark 3.2 JMH Benchmark javadoc instructions

kbendick:
As a followup to https://github.com/apache/iceberg/pull/4328, the Spark JMH benchmarks previously had invalid commands in their JavaDoc comments since we migrated to version specific Spark builds.

This updates all of the Spark 3.2 tests to include the `sparkVersions` system property, so that the tests are correct even when we move to Spark 3.3 or change the default spark version.

This also updates 4 tests that were missed in https://github.com/apache/iceberg/pull/4328


rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #4412, opened 2022-03-27
Core: Update REST catalog to use 204/404 for DELETE

rdblue:
This updates the REST catalog implementation for #4366 by removing the delete response objects. Whether the object was deleted or not found is now communicated by response code. 204 indicates that the object was successfully deleted. 404 indicates that the object did not exist.

Purge is still not implemented, but will use 202 to return a callback for checking on a purge operation that has not yet completed.


danielcweeks (approved):
LGTM.

rdblue:
Thanks for reviewing, @danielcweeks!

PR #4413, opened 2022-03-27
Core: Add stack trace to ErrorResponse

rdblue:
#4367 added a way to send stack traces in REST error responses. This updates the `ErrorResponse` implementation and test code to send stack traces.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
+++ b/core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.UncheckedIOException;
+import java.util.List;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.util.JsonUtil;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
```diff
@@ -64,6 +66,13 @@ public static void toJson(ErrorResponse errorResponse, JsonGenerator generator)
     generator.writeStringField(MESSAGE, errorResponse.message());
     generator.writeStringField(TYPE, errorResponse.type());
     generator.writeNumberField(CODE, errorResponse.code());
+    if (errorResponse.stack() != null) {
+      generator.writeArrayFieldStart(STACK);
+      for (String line : errorResponse.stack()) {
+        generator.writeString(line);
+      }
+      generator.writeEndArray();
```
kbendick, line 74: Nit: Should we consider writing the `stack` field even if it's `null`?

kbendick (commented):
Overall this looks good to me.

I left a small nit about writing out the `stack` field to the JSON even if it is `null`, but I'm leaning more towards not adding it for now.

We should update the OpenAPI document though.

danielcweeks (approved):

core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
rdblue: I think it is good this way. Why send extra information that is already clear?

rdblue:
Thanks for reviewing, @danielcweeks and @kbendick!

PR #4414, opened 2022-03-27
Core - Mark the object mapper in HTTPClient#Builder as visible for testing

kbendick:
As a follow up to some comments on this PR here https://github.com/apache/iceberg/pull/4245#discussion_r835964263, the RESTCatalog's HTTPClient's builder should only expose the object mapper for testing.

Otherwise, the correct object mapper that knows how to handle all of the necessary request / response types should be used.

Also removing a method that is not used, `Builder#httpClient`.

cc @rdblue


rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #4364, opened 2022-03-18
Flink - Fix incorrect / old row being written into delta files when using upsert mode

kbendick:
Presently, we are writing incorrect data into equality delete file manifests when using upsert mode with Flink.

The rows come in as `INSERT` row kind, but presently we are calling `writer.delete` on the entire row, which is the new row.

This leads to incorrect data being written into the equality delete manifests.

Instead, we need to be writing data using only the equality field id keys in that case, so that the delete will apply to the old row and not the new entire row.

Co-authored-by: Li Lliwei <hililiwei@gmail.com>

Opening this PR, as it addresses some of the feedback from https://github.com/apache/iceberg/pull/4316, as well as only applies to Flink 1.14. As @hililiwei and I worked on this together all week, I've marked us as co-authors. We can close that PR as this one is only oriented towards Flink 1.14 (and has the correct co-authorship in it so it can be merged at anyone's discretion).

As some code in `core` has changed, some changes need to be applied in earlier versions of Flink. To keep the PR to the minimum required, I'm leaving that out where possible and we can add it in after.

The original reporting issue is thanks to @xloya in https://github.com/apache/iceberg/pull/4311

PR #4415, opened 2022-03-28
Flink - Demonstrate (position-based) bug with multiple upserts to same row

kbendick:
As a follow up to https://github.com/apache/iceberg/pull/4364, it was discovered that there is an existing bug that is likely related to the positional deletes generated during a single snapshot.

Opening this draft so that others can inspect it. Will also be opening an issue for it as well.

### DIFF
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
@@ -55,7 +55,10 @@ public class TestFlinkMultipleUpsertsOnNonPKField extends FlinkCatalogTestBase {
   private final Map<String, String> tableUpsertProps = Maps.newHashMap();
   private TableEnvironment tEnv;
 
-  public TestFlinkMultipleUpsertsOnNonPKField(String catalogName, Namespace baseNamespace, FileFormat format, Boolean isStreamingJob) {
+  public TestFlinkMultipleUpsertsOnNonPKField(String catalogName,
+                                              Namespace baseNamespace,
+                                              FileFormat format,
+                                              Boolean isStreamingJob) {
     super(catalogName, baseNamespace);
     this.isStreamingJob = isStreamingJob;
     tableUpsertProps.put(TableProperties.FORMAT_VERSION, "2");
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
@@ -118,8 +118,8 @@ public class TestFlinkMultipleUpsertsOnNonPKField extends FlinkCatalogTestBase {
   }
 
   @Test
-  public void  testMultipleUpsertsToOneRowWithNonPKFieldChanging() {
-    String tableName = "test_multiple_upserts_on_one_row";
+  public void  testMultipleUpsertsToOneRowWithNonPKFieldChanging_fails() {
+    String tableName = "test_multiple_upserts_on_one_row_fails";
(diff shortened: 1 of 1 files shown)


kbendick:
cc @openinx @stevenzwu @hililiwei you might be interested in this additional bug I found when using Flink in upsert mode.

Of these two tests, the first one _does not_ pass, and the second one does (as noted).

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
```diff
@@ -0,0 +1,207 @@
...
+      sql("INSERT INTO %s VALUES " +
+          "('aaa', TO_DATE('2022-03-01'), 6, false)," +
+          "('aaa', TO_DATE('2022-03-01'), 6, true)," +
+          "('aaa', TO_DATE('2022-03-01'), 6, false)," +
+          "('aaa', TO_DATE('2022-03-01'), 6, false)",
+          tableName);
+
+      TestHelpers.assertRows(
+          sql("SELECT * FROM %s", tableName),
+          Lists.newArrayList(Row.of("aaa", dt, 6, false), Row.of("bbb", dt, 3, false)));
```
kbendick, line 155: This fails with the actual row of `Row.of("aaa", dt, 6, true)` every time.

hililiwei:
> cc @openinx @stevenzwu @hililiwei you might be interested in this additional bug I found when using Flink in upsert mode.
> 
> Of these two tests, the first one _does not_ pass, and the second one does (as noted).

Looks interesting. Let me try. Thanks for Ping me. 😃

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
```diff
@@ -0,0 +1,207 @@
...
+      // rows 2 back will be used instead.
+      //
+      // Indicating possibly an issue with insertedRowMap checking and/or the positional delete
+      // writer.
+      sql("INSERT INTO %s VALUES " +
+          "('aaa', TO_DATE('2022-03-01'), 1, false)," +
+          "('aaa', TO_DATE('2022-03-01'), 2, false)," +
+          "('aaa', TO_DATE('2022-03-01'), 3, false)," +
+          "('aaa', TO_DATE('2022-03-01'), 6, false)",
+          tableName);
```
kbendick, line 198: This works correctly.

kbendick:
The usage of `TO_DATE` causes calcite to generate a plan which adds a union across all of the `VALUES` entries, causing them to potentially be omitted in parallel which makes the plan not what is expected.

This is discussed in length here: https://github.com/apache/iceberg/issues/4515

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkMultipleUpsertsOnNonPKField.java
hililiwei: In my local environment, it seems to be random and occasionally successful. I'm digging deeper to find out why.

rdblue:
I just spent some time debugging this and I think the problem is with Flink, not with Iceberg.

When I run the case that fails, the issue is that the upserted rows are out of order. In both the final insert data file and the records that I see in the debugger, the column with `bool=true` is passed in last. Because this is an upsert, the last row is the version that you end up with. I'm not sure why Flink is passing the rows in the wrong order, but it could be that the parser doesn't keep track of the order of rows in a `VALUES` clause.

I also did a little exploration into what is different with the "succeeds" case. The significant change between the two in my testing was the use of `TO_DATE('2022-03-01')` instead of `DATE '2022-03-01'`. `TO_DATE` appears to trigger the rows getting out of order.

I'm going to close this issue. Feel free to reopen if you think it is a problem with Iceberg somehow reordering the rows, but that seems unlikely given that these rows are coming from Flink and are identical when they come through the Iceberg API (the `TO_DATE` / `DATE` distinction is internal to Flink).

PR #4422, opened 2022-03-28
Core: use primitive type in error code in ErrorResponse

singhpk234:
This change attempts to use primitive type in code property of Error Response, As per my understanding since we use our own parser i.e ErrorResponseParser we should be fine. 

Don't have strong reason to support why this is required, would say something that caught my eye and in general we should use primitive type over boxed if possible.

--- 
cc @kbendick

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
+++ b/core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
@@ -100,7 +100,7 @@ public class ErrorResponseParser {
     JsonNode error = jsonNode.get(ERROR);
     String message = JsonUtil.getStringOrNull(MESSAGE, error);
     String type = JsonUtil.getStringOrNull(TYPE, error);
-    Integer code = JsonUtil.getIntOrNull(CODE, error);
+    int code = JsonUtil.getInt(CODE, error);
     List<String> stack = JsonUtil.getStringListOrNull(STACK, error);
     return ErrorResponse.builder()
         .withMessage(message)
--- a/core/src/test/java/org/apache/iceberg/rest/responses/TestErrorResponseParser.java
+++ b/core/src/test/java/org/apache/iceberg/rest/responses/TestErrorResponseParser.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.rest.responses;
 
 import java.util.Arrays;
 import java.util.List;
+import org.apache.iceberg.AssertHelpers;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -30,7 +31,7 @@ public class TestErrorResponseParser {
   public void testErrorResponseToJson() {
     String message = "The given namespace does not exist";
     String type = "NoSuchNamespaceException";
-    Integer code = 404;
+    int code = 404;
     String errorModelJson = String.format("{\"message\":\"%s\",\"type\":\"%s\",\"code\":%d}", message, type, code);
     String json = "{\"error\":" + errorModelJson + "}";
     ErrorResponse response = ErrorResponse.builder().withMessage(message).withType(type).responseCode(code).build();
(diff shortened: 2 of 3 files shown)


core/src/test/java/org/apache/iceberg/rest/responses/TestErrorResponseParser.java
```diff
@@ -40,14 +41,25 @@ public void testErrorResponseToJson() {
...
+  @Test
+  public void testErrorResponseFromJsonWithoutCode() {
+    String message = "The given namespace does not exist";
+    String type = "NoSuchNamespaceException";
+    String errorModelJson = String.format("{\"message\":\"%s\",\"type\":\"%s\"}", message, type);
+    String json = "{\"error\":" + errorModelJson + "}";
+
+    AssertHelpers.assertThrows("ErrorResponse should contain code", IllegalArgumentException.class,
+        "Cannot parse missing int code", () -> ErrorResponseParser.fromJson(json));
+  }
```
kbendick, line 61: My concern here is that the error code can still be parsed from other places (like the response itself). I agree it _should_ be there, but I'm not sure if adding this additional constraint provides that much benefit.

The number of boxed `Integer`s is still likely to be relatively small, as most operations will still go through normal `FileIO`.

I think this optimization is possibly a bit premature, but I'll let others weigh in as well first.

core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
```diff
@@ -91,7 +91,7 @@ public static ErrorResponse fromJson(JsonNode jsonNode) {
     JsonNode error = jsonNode.get(ERROR);
     String message = JsonUtil.getStringOrNull(MESSAGE, error);
     String type = JsonUtil.getStringOrNull(TYPE, error);
-    Integer code = JsonUtil.getIntOrNull(CODE, error);
+    int code = JsonUtil.getInt(CODE, error);
```
kbendick, line 94: Question: if `CODE` isn't present, what exception will this throw?

kbendick (commented):
Thanks for the contribution @singhpk234.

Right now, if the `ErrorResponse` fails to parse, we build an `ErrorResponse` using values from the response. And then we rely on that information in the response error handlers.

So I'm not fully opposed to this, given the importance of this field. But I do think that this optimization is a bit premature, given the other places the response code can come from.

I'll let others weigh in.

core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
singhpk234: It would throw an IllegalArgumentException, the UT added with this PR shows the same.

core/src/test/java/org/apache/iceberg/rest/responses/TestErrorResponseParser.java
singhpk234: > I'm not sure if adding this additional constraint provides that much benefit.

+1, I also expected that code should be present anyways, absence of it during build() time we check code is not null for creating Response obj, and throw [IllegalArgs exception](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponse.java#L132).

> I think this optimization is possibly a bit premature, but I'll let others weigh in as well first.

Agree with you, I don't have a strong opinion of must having this change either. Happy to close this as well :) !!!

core/src/main/java/org/apache/iceberg/rest/responses/ErrorResponseParser.java
kbendick: Oh my bad. The unit test completely shows that. I've been somewhat sick the past few days so forgive me 😅

singhpk234: Not at all a problem :) !!

jackye1995:
Closing old PR as this is not active

PR #4423, opened 2022-03-28
AWS: Add LakeFormation Integration tests

xiaoxuandev:


aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -267,16 +267,16 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
       glue.deleteTable(DeleteTableRequest.builder()
           .catalogId(awsProperties.glueCatalogId())
           .databaseName(IcebergToGlueConverter.getDatabaseName(identifier))
           .name(identifier.name())
           .build());
       LOG.info("Successfully dropped table {} from Glue", identifier);
-      if (purge && lastMetadata != null) {
-        CatalogUtil.dropTableData(ops.io(), lastMetadata);
-        LOG.info("Glue table {} data purged", identifier);
-      }
```
jackye1995, line 279: ah interesting, I overlooked this aspect. Does that mean when user has LF enabled, we must purge the table before dropping the table in Glue, otherwise it cannot get temporary S3 credential?

If that's the case, we should make delete table in the finally block, so that purge failure does not impact actual table drop.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -111,6 +113,17 @@ protected void doRefresh() {
...
+    if (awsProperties.glueLakeFormationEnabled() && base == null) {
+      glue.createTable(CreateTableRequest.builder()
+          .databaseName(databaseName)
+          .tableInput(TableInput.builder()
+              .parameters(ImmutableMap.of(TABLE_TYPE_PROP, ICEBERG_TABLE_TYPE_VALUE))
+              .name(tableName)
+              .storageDescriptor(StorageDescriptor.builder().location(metadata.location()).build())
+              .build()).build());
+    }
     String newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
```
jackye1995, line 127: I guess under the LF model this dummy table creation is the best we can do...

This means we should move the whole block to the try catch block. When the execution fails, and if we created this dummy table, we should also delete it.

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/LakeFormationTestBase.java
```diff
@@ -0,0 +1,505 @@
...
+    String arn = getArnForS3Location(s3Location);
+    try {
+      lf.deregisterResource(DeregisterResourceRequest.builder().resourceArn(arn).build());
+    } catch (EntityNotFoundException e) {
+      LOG.info("Resource {} not found. Error: {}", arn, e);
+    }
+  }
+
+  private static String createPolicyArn(String policyName) {
+    return TEST_POLICY_ARN
```
jackye1995, line 366: I don't think we need to do string replace, we can just do `String.format("arn:%s:iam::%s:policy/%s", ...)` here

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/TestLakeFormationDML.java
```diff
@@ -0,0 +1,159 @@
...
+          .withRecordCount(1)
+          .build();
+      AssertHelpers.assertThrows("attempt to insert to a table without INSERT permission should fail",
+          S3Exception.class,
+          "Access Denied",
+          () -> table.newAppend().appendFile(dataFile).commit());
+    } finally {
+      lfRegisterPathRoleDeleteTable(testDbName, testTableName);
+      lfRegisterPathRoleDeleteDb(testDbName);
+    }
```
singhpk234, line 89: [suggestion] try structuring the code in BeforeEach, AfterEach 

      lfRegisterPathRoleDeleteTable(testDbName, testTableName);
      lfRegisterPathRoleDeleteDb(testDbName);

seems to be good candidate for AfterEach, 

    lfRegisterPathRoleCreateDb(testDbName);
    lfRegisterPathRoleCreateTable(testDbName, testTableName);

seems to good candidate for BeforeEach

can make testDbName / testTableName as class property which you set before each ut execution and similarly you can reference it AfterEach to get rid of it.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
xiaoxuandev: done. thanks!

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -264,15 +265,25 @@ private boolean isGlueIcebergTable(Table table) {
...
     try {
-      TableOperations ops = newTableOps(identifier);
-      TableMetadata lastMetadata = ops.current();
       glue.deleteTable(DeleteTableRequest.builder()
           .catalogId(awsProperties.glueCatalogId())
           .databaseName(IcebergToGlueConverter.getDatabaseName(identifier))
           .name(identifier.name())
           .build());
       LOG.info("Successfully dropped table {} from Glue", identifier);
+      if (purge && awsProperties.glueLakeFormationEnabled() && lastMetadata != null) {
```
jackye1995, line 283: We can directly do the check upfront right? If `purge && awsProperties.glueLakeFormationEnabled()` we should directly throw error without loading last metadata.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -264,15 +265,25 @@ private boolean isGlueIcebergTable(Table table) {
...
+        throw new RuntimeException("Can not delete table S3 files, should call deleteTable first " +
```
jackye1995, line 284: can use ValidationException.check

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -111,10 +114,25 @@ protected void doRefresh() {
 
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
-    String newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
-    CommitStatus commitStatus = CommitStatus.FAILURE;
-
+    CommitStatus commitStatus = null;
+    String newMetadataLocation = null;
+    boolean glueDummyTableCreated = false;
```
jackye1995, line 119: prefer to not use words like dummy. Maybe `glueTempTableCreated`?

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -111,10 +114,25 @@ protected void doRefresh() {
 
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
-    String newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
-    CommitStatus commitStatus = CommitStatus.FAILURE;
-
+    CommitStatus commitStatus = null;
```
jackye1995, line 117: why change this line?

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
xiaoxuandev: In that case we won't be able to drop table if `purge && awsProperties.glueLakeFormationEnabled()` even S3 data is already deleted.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
xiaoxuandev: oh yes, this should be FAILURE

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/LakeFormationTestBase.java
```diff
@@ -0,0 +1,511 @@
...
+  static String lfPrivilegedRolePolicyName;
+  static DataLakePrincipal principalUnderTest;
+  static String testBucketPath = "s3://" + AwsIntegTestUtil.testBucketName() + "/" + TEST_PATH_PREFIX;
+  static Schema schema = new Schema(Types.NestedField.required(1, "c1", Types.StringType.get(), "c1"));
+  static PartitionSpec partitionSpec = PartitionSpec.builderFor(schema).build();
+
+  static GlueCatalog glueCatalogRegisterPathRole;
+  static GlueCatalog glueCatalogPrivilegedRole;
+  static IamClient iam;
+  static LakeFormationClient lf;
```
jackye1995, line 119: nit: prefer to use full name `lakeformation`

jackye1995, line 119: nit: prefer to use full name `lakeformation`

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/LakeFormationTestBase.java
```diff
@@ -0,0 +1,511 @@
...
+            "\"Statement\":[{" +
+            "\"Effect\":\"Allow\"," +
+            "\"Principal\":{" +
+            "\"AWS\":\"arn:aws:iam::" + AwsIntegTestUtil.testAccountId() + ":root\"}," +
+            "\"Action\": [\"sts:AssumeRole\"," +
+            "\"sts:TagSession\"]}]}")
+        .maxSessionDuration(ASSUME_ROLE_SESSION_DURATION)
+        .build());
+    lfPrivilegedRoleArn = response.role().arn();
+    LOG.info("lf privileged role is {} :", lfPrivilegedRoleArn);
```
jackye1995, line 176: nit: no need for logging in tests

jackye1995, line 176: nit: no need for logging in tests

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/LakeFormationTestBase.java
```diff
@@ -0,0 +1,511 @@
...
+            "\"Effect\":\"Allow\"," +
+            "\"Principal\":{" +
+            "\"Service\":[\"glue.amazonaws.com\"," +
+            "\"lakeformation.amazonaws.com\"]," +
+            "\"AWS\":\"arn:aws:iam::" + AwsIntegTestUtil.testAccountId() + ":root\"}," +
+            "\"Action\": [\"sts:AssumeRole\"]}]}")
+        .maxSessionDuration(ASSUME_ROLE_SESSION_DURATION)
+        .build());
+    lfRegisterPathRoleArn = response.role().arn();
+    LOG.info("register path role is : {}", lfRegisterPathRoleArn);
```
jackye1995, line 150: nit: no need for logging in tests

jackye1995, line 150: nit: no need for logging in tests

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/TestLakeFormationDDL.java
```diff
@@ -0,0 +1,291 @@
...
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.exceptions.CommitStateUnknownException;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.junit.Assert;
+import org.junit.Test;
+import software.amazon.awssdk.services.glue.model.AccessDeniedException;
+import software.amazon.awssdk.services.lakeformation.model.Permission;
+
+public class TestLakeFormationDDL extends LakeFormationTestBase {
```
jackye1995, line 37: the test class names are a bit awkward, DDL and DML are for SQL languages, maybe something like `TestLakeFormationMetadataOperations` and `TestLakeFormationDataOperations`?

jackye1995, line 37: the test class names are a bit awkward, DDL and DML are for SQL languages, maybe something like `TestLakeFormationMetadataOperations` and `TestLakeFormationDataOperations`?

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/TestLakeFormationDML.java
```diff
@@ -0,0 +1,125 @@
...
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+public class TestLakeFormationDML
+    extends LakeFormationTestBase {
+
+  private static String testDbName;
+  private static String testTableName;
+
+  @Before
+  public void beforeEach() {
```
jackye1995, line 43: nit: prefer name matching annotation, `before()`

jackye1995, line 43: nit: prefer name matching annotation, `before()`

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/TestLakeFormationDML.java
```diff
@@ -0,0 +1,125 @@
...
+    testDbName = getRandomDbName();
+    testTableName = getRandomTableName();
+    lfRegisterPathRoleCreateDb(testDbName);
+    lfRegisterPathRoleCreateTable(testDbName, testTableName);
+  }
+
+  @After
+  public void afterEach() {
```
jackye1995, line 51: nit: prefer name matching annotation, `after()`

jackye1995, line 51: nit: prefer name matching annotation, `after()`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -264,15 +266,24 @@ private boolean isGlueIcebergTable(Table table) {
...
   public boolean dropTable(TableIdentifier identifier, boolean purge) {
+    TableOperations ops = newTableOps(identifier);
+    TableMetadata lastMetadata;
+    try {
+      lastMetadata = ops.current();
+    } catch (NoSuchKeyException ex) {
+      // S3 files already deleted, can not get S3 object through metadata location
+      lastMetadata = null;
+    }
     try {
```
jackye1995, line 277: nit: newline after try catch block

jackye1995, line 277: nit: newline after try catch block

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -264,15 +266,24 @@ private boolean isGlueIcebergTable(Table table) {
...
-      TableOperations ops = newTableOps(identifier);
-      TableMetadata lastMetadata = ops.current();
       glue.deleteTable(DeleteTableRequest.builder()
           .catalogId(awsProperties.glueCatalogId())
           .databaseName(IcebergToGlueConverter.getDatabaseName(identifier))
           .name(identifier.name())
           .build());
       LOG.info("Successfully dropped table {} from Glue", identifier);
+      ValidationException.check(!purge || !awsProperties.glueLakeFormationEnabled() || lastMetadata == null,
+          "Can not delete table S3 files, should call deleteTable first " +
```
jackye1995, line 285: The error message is a bit unclear from user perspective, my understanding is that "Cannot purge table with LakeFormation enabled because S3 access is lost after table is dropped. Please purge table data first before dropping the table"

jackye1995, line 285: The error message is a bit unclear from user perspective, my understanding is that "Cannot purge table with LakeFormation enabled because S3 access is lost after table is dropped. Please purge table data first before dropping the table"

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -264,15 +266,24 @@ private boolean isGlueIcebergTable(Table table) {
...
     try {
-      TableOperations ops = newTableOps(identifier);
-      TableMetadata lastMetadata = ops.current();
       glue.deleteTable(DeleteTableRequest.builder()
           .catalogId(awsProperties.glueCatalogId())
           .databaseName(IcebergToGlueConverter.getDatabaseName(identifier))
           .name(identifier.name())
           .build());
       LOG.info("Successfully dropped table {} from Glue", identifier);
+      ValidationException.check(!purge || !awsProperties.glueLakeFormationEnabled() || lastMetadata == null,
```
jackye1995, line 284: why do we need `lastMetadata = null` check in this validation? Seems unnecessary to me

jackye1995, line 284: why do we need `lastMetadata = null` check in this validation? Seems unnecessary to me

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -111,10 +114,24 @@ protected void doRefresh() {
...
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
-    String newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
     CommitStatus commitStatus = CommitStatus.FAILURE;
-
+    String newMetadataLocation = null;
+    boolean glueTempTableCreated = false;
     try {
+      if (awsProperties.glueLakeFormationEnabled() && base == null) {
+        // LakeFormation credential require TableArn as input， so creating a dummy table
```
jackye1995, line 122: nit: extra space before "so"

jackye1995, line 122: nit: extra space before "so"

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -111,10 +114,24 @@ protected void doRefresh() {
...
+        glue.createTable(CreateTableRequest.builder()
+            .databaseName(databaseName)
+            .tableInput(TableInput.builder()
+                .parameters(ImmutableMap.of(TABLE_TYPE_PROP, ICEBERG_TABLE_TYPE_VALUE))
+                .name(tableName)
+                .storageDescriptor(StorageDescriptor.builder().location(metadata.location()).build())
+                .build()).build());
+        glueTempTableCreated = true;
+      }
+      newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
```
jackye1995, line 133: nit: newline after if block

jackye1995, line 133: nit: newline after if block

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -111,10 +114,24 @@ protected void doRefresh() {
...
+      if (awsProperties.glueLakeFormationEnabled() && base == null) {
+        // LakeFormation credential require TableArn as input， so creating a dummy table
+        // beforehand for create table scenario
+        glue.createTable(CreateTableRequest.builder()
+            .databaseName(databaseName)
+            .tableInput(TableInput.builder()
+                .parameters(ImmutableMap.of(TABLE_TYPE_PROP, ICEBERG_TABLE_TYPE_VALUE))
+                .name(tableName)
+                .storageDescriptor(StorageDescriptor.builder().location(metadata.location()).build())
+                .build()).build());
```
jackye1995, line 130: newline needed between 2 `build()`

jackye1995, line 130: newline needed between 2 `build()`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
jackye1995, line 277: try should be at the top including `TableOperations ops = newTableOps(identifier);`

jackye1995, line 277: try should be at the top including `TableOperations ops = newTableOps(identifier);`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
xiaoxuandev: we need this to know if the S3 data file has already been deleted

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
xiaoxuandev: done. thanks!

xiaoxuandev: done. thanks!

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -259,13 +261,23 @@ private boolean isGlueIcebergTable(Table table) {
   public boolean dropTable(TableIdentifier identifier, boolean purge) {
     try {
       TableOperations ops = newTableOps(identifier);
-      TableMetadata lastMetadata = ops.current();
+      TableMetadata lastMetadata;
```
jackye1995, line 264: discussed offline, loading table metadata still feels hacky to me. We can get Glue table and check the metadata location to determine if the table is already purged or not.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -259,13 +261,23 @@ private boolean isGlueIcebergTable(Table table) {
...
+      try {
+        lastMetadata = ops.current();
```
jackye1995, line 266: this will keep retrying because it will only fail with `NoSuchTableException`, so `ops.current()` should not be called.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -514,4 +530,22 @@ public void setConf(Configuration conf) {
...
+  private boolean isTablePurged(TableIdentifier identifier, S3Client s3Client) {
+    Table table = glue.getTable(GetTableRequest.builder()
+        .catalogId(awsProperties.glueCatalogId())
+        .databaseName(identifier.namespace().toString())
+        .name(identifier.name())
+        .build()).table();
+    String metadataLocation = table.parameters().get(BaseMetastoreTableOperations.METADATA_LOCATION_PROP);
+
+    try {
+      S3URI s3URI = new S3URI(metadataLocation);
```
jackye1995, line 543: we don't need to make `S3URI` and `S3FileIO.client()` public, we can just do `fileIO.newInputFile(...).exists()`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -158,7 +165,31 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+            .tableInput(TableInput.builder()
+                .parameters(ImmutableMap.of(TABLE_TYPE_PROP, ICEBERG_TABLE_TYPE_VALUE))
+                .name(tableName)
+                .storageDescriptor(StorageDescriptor.builder().location(metadataLocation).build())
+                .build())
+            .build());
+        return true;
+      }
+    } catch (GlueException e) {
+      return false;
```
jackye1995, line 190: I think it should not mute the exception, the operation should fail if temp table creation fails

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -158,7 +165,31 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
     }
+    return false;
```
jackye1995, line 192: nit: newline after control statement

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -158,7 +165,31 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       }
     } finally {
       cleanupMetadataAndUnlock(commitStatus, newMetadataLocation);
+      if (glueTempTableCreated && commitStatus != CommitStatus.SUCCESS) {
+        glue.deleteTable(DeleteTableRequest.builder().databaseName(databaseName).name(tableName).build());
+      }
+    }
+  }
+
+  private boolean createGlueTempTable(TableMetadata base, String metadataLocation) {
```
jackye1995, line 174: the name seems to be `createGlueTempTableIfNecessary`?

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -158,7 +165,31 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       }
     } finally {
       cleanupMetadataAndUnlock(commitStatus, newMetadataLocation);
+      if (glueTempTableCreated && commitStatus != CommitStatus.SUCCESS) {
```
jackye1995, line 168: we can move this to a helper `cleanupGlueTempTableIfNecessary`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -273,13 +278,24 @@ private boolean isGlueIcebergTable(Table table) {
...
+        lastMetadata = ops.current();
+      }
+
       glue.deleteTable(DeleteTableRequest.builder()
           .catalogId(awsProperties.glueCatalogId())
           .databaseName(IcebergToGlueConverter.getDatabaseName(identifier))
           .name(identifier.name())
           .build());
       LOG.info("Successfully dropped table {} from Glue", identifier);
+      ValidationException.check(!purge || !awsProperties.glueLakeFormationEnabled() || isTablePurged,
```
jackye1995, line 296: prefer to make check logic consistent with the error message, in this case `!(purge && awsProperties.glueLakeFormationEnabled() && !isTablePurged)`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -273,13 +278,24 @@ private boolean isGlueIcebergTable(Table table) {
   public boolean dropTable(TableIdentifier identifier, boolean purge) {
     try {
       TableOperations ops = newTableOps(identifier);
-      TableMetadata lastMetadata = ops.current();
+
+      GlueTableOperations glueOps = (GlueTableOperations) ops;
```
jackye1995, line 282: I don't think we need to cast it to `GlueTableOperations` to get `io()`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -273,13 +278,24 @@ private boolean isGlueIcebergTable(Table table) {
...
+      S3FileIO s3FileIO = (S3FileIO) glueOps.io();
```
jackye1995, line 283: this might not always be true, people can use other FileIO implementations.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -273,13 +278,24 @@ private boolean isGlueIcebergTable(Table table) {
...
+      }
+
       glue.deleteTable(DeleteTableRequest.builder()
           .catalogId(awsProperties.glueCatalogId())
           .databaseName(IcebergToGlueConverter.getDatabaseName(identifier))
           .name(identifier.name())
           .build());
       LOG.info("Successfully dropped table {} from Glue", identifier);
+      ValidationException.check(!purge || !awsProperties.glueLakeFormationEnabled() || isTablePurged,
+          "Cannot purge table with LakeFormation enabled because S3 access is lost after table is dropped. " +
```
jackye1995, line 297: Overall for this section of logic, my opinion is that we should not do glue.deleteTable if the table is LF enabled and purge is requested. So the validation should happen before we do the delete.

xiaoxuandev: I think we still want the glue table to be dropped even LF enabled and purge is set to true? And the original logic is to ensure the glue table is dropped even purge fails.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -242,7 +273,9 @@ void cleanupMetadataAndUnlock(CommitStatus commitStatus, String metadataLocation
     try {
       if (commitStatus == CommitStatus.FAILURE) {
         // if anything went wrong, clean up the uncommitted metadata file
-        io().deleteFile(metadataLocation);
+        if (metadataLocation != null && !metadataLocation.isEmpty()) {
```
jackye1995, line 276: this statement can be combined with L274

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -114,10 +117,13 @@ protected void doRefresh() {
 
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
-    String newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
     CommitStatus commitStatus = CommitStatus.FAILURE;
-
```
jackye1995, line 119: nit: should not remove newlines, this causes merge conflict in other people's PRs

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -514,4 +524,17 @@ public void setConf(Configuration conf) {
...
   }
+
+  private boolean isTablePurged(TableIdentifier identifier, FileIO tableFileIO) {
+    Table table = glue.getTable(GetTableRequest.builder()
+        .catalogId(awsProperties.glueCatalogId())
+        .databaseName(identifier.namespace().toString())
+        .name(identifier.name())
+        .build()).table();
+    String metadataLocation = table.parameters().get(BaseMetastoreTableOperations.METADATA_LOCATION_PROP);
+    if (metadataLocation == null || metadataLocation.isEmpty()) {
```
jackye1995, line 535: L535-538 can be combined in a single return

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -514,4 +524,17 @@ public void setConf(Configuration conf) {
   protected Map<String, String> properties() {
     return catalogProperties == null ? ImmutableMap.of() : catalogProperties;
   }
+
+  private boolean isTablePurged(TableIdentifier identifier, FileIO tableFileIO) {
```
jackye1995, line 528: the method name is not really accurate, we are just checking if metadata location is still there, nothing else. What about `tableMetadataExists`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -514,4 +524,17 @@ public void setConf(Configuration conf) {
...
+    Table table = glue.getTable(GetTableRequest.builder()
+        .catalogId(awsProperties.glueCatalogId())
+        .databaseName(identifier.namespace().toString())
```
jackye1995, line 531: use `getDatabaseName` util, same as below

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
jackye1995: > I think we still want the glue table to be dropped even LF enabled and purge is set to true?

Let's think about this again...

What we know is that if table is LF enabled and purge is true, there is no way we can make this happen safely, because:
1. if we drop table first, we can pre-fetch s3 credential but that might not be enough time to remove all the table data files
2. if we remove all the table data files first, we might fail in the middle, this causes orphan files. 

Originally I suggested that because there is no perfect way, let's just ask customer to purge by themselves and we validate and not allow them to purge in Iceberg if LF is enabled. But now I think again, this is known behavior even for other normal tables, and it looks like we should go with 2, so that we first remove table data files, all the failures are already suppressed in the operation, and then we drop the table.

+ @amogh-jahagirdar any thoughts?

amogh-jahagirdar: Sorry for the late reply, but yeah Pre-fetching credentials and deleting the table is too brittle because those credentials would be invalidated. However, the case I am thinking of is if we do a best effort deletion of the files first, if there's concurrent operations which don't conflict (like append), those could succeed. Then if dropping the table after fails, then the table is in a bad state. To address this there can be a DROP_IN_PROGRESS STATE introduced which is set first and then used when committing to prevent commits against a table being deleted. I think any attempt of deletion of files should only happen when it's impossible to modify the table further.

Let me know if I misunderstood the problem though, I'll think more on it but I think 1 certainly can be ruled out.

xiaoxuandev: Discussed offline, we decide to go with approach 1, a few reasons:
1. we don't want the data file to be deleted if the users don't have the drop permission, so this rules out the option 2
2. we can use the pre-fetched credentials because LF always gives us the credentials for each individual service, and it is always our own responsibility to use them in a disciplined way. Even after the Glue table is dropped, we know those S3 files belong to this table, so using the pre-fetched S3 credentials to delete them is OK. The credential can expire, but the file deletion is best effort anyway.

jackye1995 (approved):
thanks for adding the tests and make the fix!

amogh-jahagirdar (approved):
Looks good to me, thanks @xiaoxuandev !

PR #4427, opened 2022-03-30
[Spark][Core]: Release memory when rewriting RewriteFileGroup finished

xloya:
Please see this for the initial patch: #4359, this patch makes some optimizations.  
When rewriting, `Tasks` do not need to retain successful tasks for example `RewriteFileGroup` to revert. However, the references to these RewriteFileGroups are currently retained, and the memory will not be released after the rewriting finished. When rewriting the v2 table, a `RewriteFileGroup` may contain a large number of data files and delete files. After the `partial commit` is enabled, the memory overhead of the driver will also increase as the rewrite proceeds. So it makes sense to release memory here.  
@RussellSpitzer @rdblue Could you help to review again when you have time? Thanks!


core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -212,7 +212,9 @@ public boolean run(Task<I, RuntimeException> task) {
           I item = iterator.next();
           try {
             runTaskWithRetry(task, item);
-            succeeded.add(item);
+            if (revertTask != null) {
+              succeeded.add(item);
+            }
```
rdblue, line 217: This is very suspicious to me. Can you explain why `Tasks` needs to change? It seems like this may be trying to handle the case where a revert happens later and so the success didn't necessarily go through. But this check doesn't look like the right way to accomplish that.

core/src/main/java/org/apache/iceberg/actions/RewriteDataFilesCommitManager.java
```diff
@@ -198,9 +199,9 @@ public void offer(RewriteFileGroup group) {
     }
 
     /**
-     * Returns all File groups which have been committed
+     * Returns all file groups results which have been committed
      */
-    public List<RewriteFileGroup> results() {
+    public List<RewriteDataFiles.FileGroupRewriteResult> results() {
```
rdblue, line 204: Can you import `FileGroupRewriteResult` rather than using the long name here?

rdblue (changes requested):
This seems reasonable except for changes to `Tasks`.

xloya:
@rdblue Sorry for the late reply as I just finished my vacation, I'll look at the comments later, thanks again for your help reviewing!

core/src/main/java/org/apache/iceberg/actions/RewriteDataFilesCommitManager.java
xloya: Of course, I think the `FileGroupRewriteResult` interface can be taken out separately to solve this problem

core/src/main/java/org/apache/iceberg/util/Tasks.java
xloya: Let me explain why the change is needed here:  
At present, when rewriting, it will be rewritten with the `RewriteFileGroup` as the rewrite unit through the `Tasks` tool class without `revertWith` method:  
    // Start rewrite tasks
    Tasks.foreach(groupStream)
        .suppressFailureWhenFinished()
        .executeWith(rewriteService)
        .noRetry()
        .onFailure((fileGroup, exception) -> LOG.error("Failure during rewrite group {}", fileGroup.info(), exception))
        .run(fileGroup -> commitService.offer(rewriteFiles(ctx, fileGroup)));  
But at this time, after the rewrite is finished, there is no need to revert the successful task. As you can see from the following code, unless we call and set `revertTask`, the logic here will not be executed anyway:  
    public Builder<I> revertWith(Task<I, ?> task) {
      this.revertTask = task;
      return this;
    }  
    ...  
    if (taskFailed.get() && revertTask != null) {
      // at least one task failed, revert any that succeeded
      for (final I item : succeeded) {
        futures.add(service.submit(new Runnable() {
          @Override
          public void run() {
            if (stopRevertsOnFailure && revertFailed.get()) {
              return;
            }


            boolean failed = true;
            try {
              revertTask.run(item);
              failed = false;
            } catch (Exception e) {
              LOG.error("Failed to revert task", e);
              // swallow the exception
            } finally {
              if (failed) {
                revertFailed.set(true);
              }
            }
          }
        }));
      }  
But the current logic is that no matter what, the reference of the successful task will be saved to the concurrent queue of succeeded, and the `RewriteFileGroup` that has been rewritten will also be saved. If we want to release the memory of `RewriteFileGroup`, then we should also remove unnecessary references to it here

Issue #3899, opened 2022-01-13
Support Spark time travel against a branch


jackye1995:
Time-based time travel can be performed against a specific branch instead of the main branch

hililiwei:
I'd like to try it. Could you please assign it to me?

PR #4428, opened 2022-03-30
Core, API: Add getting refs and snapshot by ref to the Table API

amogh-jahagirdar:
This PR adds the ability to perform a table scan from a given branch or tag. Spark integration will be in a separate PR. Related to https://github.com/apache/iceberg/issues/3899


amogh-jahagirdar:
@rdblue @jackye1995  Let me know what you think!

amogh-jahagirdar:
Ah. Core tests are failing from my change, when we test serializing metadata tables.

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -49,6 +49,24 @@
...
   TableScan useSnapshot(long snapshotId);
 
+  /**
+   * Create a new {@link TableScan} from this scan's configuration that will use the given branch
+   *
+   * @param branch the branch to scan from
+   * @return a new scan based on this with the given branch
+   * @throws IllegalArgumentException if the branch cannot be found
+   */
+  TableScan branch(String branch);
```
nastra, line 59: does the distinction between branch/tag really matter here or could we just have a `TableScan ref(String ref)` method?

amogh-jahagirdar: It's a good point. My thought was that from an API point of view people would be more familiar with the terms branch and tag, so they would leverage that. I also think if is using this, they know before hand if it's a branch or a tag, so it's not cumbersome. But yeah currently, the logic is effectively the same which is to use the snapshot for the reference.

Thoughts?

jackye1995: Would `useRef` be better? Just to match `useSnapshot`.

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -286,4 +286,18 @@ default AppendFiles newFastAppend() {
...
+   * @return the current refs for the table
+   */
+  Map<String, SnapshotRef> refs();
+
+  /**
+   * Returns the snapshot ref with the given name or null if it does not exist
+   *
+   * @return the ref with the given name or null if it does not exist.
+   */
+  SnapshotRef ref(String ref);
```
jackye1995, line 302: do we really need this method, if it's always `refs().get(ref)`

jackye1995: oh nvm we also have the method in `TableMetadata`, so probably better to have it here

api/src/main/java/org/apache/iceberg/TableScan.java
jackye1995: Actually I think there is a need to distinguish `useBranch` and `useTag`, because `useBranch` can be used in combiantion with `asOfTime` to perform time travel against a specific branch.

jackye1995:
Can we separate this a bit to 3 parts:
1. we need to definitely add `refs()` related APIs to `Table` related places
2. we want to support `useTag` that is basically the same as `useSnapshot`
3. we want to support `useBranch` that can be used in combination with `asOfTime`

api/src/main/java/org/apache/iceberg/Table.java
amogh-jahagirdar: Right, my thinking was just to keep it the same

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -156,6 +156,16 @@ public List<HistoryEntry> history() {
     return table().history();
   }
 
+  @Override
+  public SnapshotRef ref(String ref) {
```
jackye1995, line 160: feels a bit awkward for metadata table to have refs... but given the fact that we are just mapping this to methods of the source `table()` for all the other methods, this seems to be the only way to do it

jackye1995 (approved):
looks good to me

xiaoxuandev (approved):

jackye1995:
@nastra any additional comments? If not I think this is a straight-forward change and will merge later.

singhpk234 (approved):

api/src/main/java/org/apache/iceberg/Table.java
rdblue: Yeah, I think it's nice to provide the convenience method.

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -297,4 +297,18 @@ default AppendFiles newFastAppend() {
    * Returns a {@link LocationProvider} to provide locations for new data files.
    */
   LocationProvider locationProvider();
+
+  /**
+   * Returns the current refs for the table
+   *
+   * @return the current refs for the table
+   */
+  Map<String, SnapshotRef> refs();
```
rdblue, line 306: Do we want to expose `SnapshotRef` in the `Table` API? Why not return Snapshot for ref names?

amogh-jahagirdar: I was thinking it made sense to expose Refs in the Table API because refs are maintained per table and any operation which leverages the table API can easily access them.

 I think refs() is helpful mostly as a convenience method to list all the refs in a table (similar being able to list the table properties, or the schemas/partition specs etc).

For the cases where a user of the table API knows what ref they are looking for then the ref(ref) becomes helpful. If we don't return Snapshot for a given ref name, then a caller has to do 

`table.snapshot(table.refs(name).snashotId())`

and they have to also take care of the null check in case a ref with name does not exist.

So my conclusion is the following:

1.) Expose the SnapshotRef in the Table API should be fine, because refs are maintained at the table level and an API expose to them either as a collection or a convenience method to look them up by name fits the model, and could be used.

2.) It will be common to want the snapshot for a given ref, and it also makes sense to have an API for returning a Snapshot for a given ref name.

Alternative:

The tradeoff of the above is the API is less minimal. For the purpose of the table scan we will ultimately need the snapshot for a given ref. So if we want to start minimal, what we can do is add the 

`Snapshot snapshot(String ref)` signature and only when the refs(), ref(String name) are truly needed we add those. 

 @rdblue @jackye1995 @singhpk234 Let me know your thoughts!

rdblue: For now, I'd probably add the `snapshot(String name)` method and put off adding methods to `Table` until we know that we will definitely need them.

amogh-jahagirdar: Actually, one thing I forgot was for SerializableTable, we'll need to pass the refs through https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SerializableTable.java#L83 so we'll need Map<String, SnapshotRef> refs() to implement the contract of snapshot(String name) in SerializableTable.

So then we need both refs() and snapshot(String name). We can leave off the ref(String name) for now, until we know that convenience method is really useful.

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -297,4 +297,25 @@ default AppendFiles newFastAppend() {
...
+
+  /**
+   * Returns the snapshot referenced by the given name or null if no such reference exists.
+   *
+   * @return the snapshot which is referenced by the given name or null if no such reference exists.
+   */
+  default Snapshot snapshot(String name) {
```
amogh-jahagirdar, line 313: Implemented snapshot(String name) a default method on the interface because most of the table implementations would have this same logic.

nastra (approved):
LGTM

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -297,4 +297,25 @@ default AppendFiles newFastAppend() {
    * Returns a {@link LocationProvider} to provide locations for new data files.
    */
   LocationProvider locationProvider();
+
+  /**
+   * Returns the current refs for the table
+   *
+   * @return the current refs for the table
+   */
+  Map<String, SnapshotRef> refs();
```
rdblue, line 306: Shouldn't this also have a default that creates a map of "main" to the current snapshot ID?

rdblue: Yeah, that sounds reasonable as well.

namratha2403:
@amogh-jahagirdar Is there any pending commit to this PR? I would like to start contributing for spark integration for this.

amogh-jahagirdar:
> > @amogh-jahagirdar Is there any pending commit to this PR? I would like to start contributing for spark integration for this.
> 
> @namratha2403 I raised #4922 for addressing [#4428 (comment)](https://github.com/apache/iceberg/pull/4428#discussion_r885895391) separately. I'll be updating that PR later today. At that point this PR would be unblocked or if we conclude not to set the main ref to the current snapshot when parsing, we would have a default implementation for refs() here.
> 
> cc: @rdblue

@rdblue Since we now are guaranteed to have a main ref when parsing metadata https://github.com/apache/iceberg/blob/dec56790be30180b24c468e88989fbd3b7bf9450/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L426 I think the code in this change can operate with that assumption. Let me know what you think!

hililiwei (approved):

jackye1995:
Looks like we have a consensus for this, @amogh-jahagirdar could you do a rebase to make sure spotless check passes?

jackye1995 (approved):

jackye1995:
Thanks for the work! Given the conversation history I think @rdblue is also good with adding these APIs, and other people have all approved. Thanks for all the reviews!

PR #4432, opened 2022-03-30
Flink: fix parquet readers for array data

yittg:
In flink, only three predefined types of `ArrayData` are accepted by
`ArrayDataSerializer#copy`. It's not allowed to return customized type
of `ArrayData`, otherwise a `ClassCastException` will be thrown.

### DIFF
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
@@ -24,10 +24,12 @@ import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.time.ZoneOffset;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import org.apache.flink.table.data.ArrayData;
 import org.apache.flink.table.data.DecimalData;
+import org.apache.flink.table.data.GenericArrayData;
 import org.apache.flink.table.data.GenericRowData;
 import org.apache.flink.table.data.MapData;
 import org.apache.flink.table.data.RawValueData;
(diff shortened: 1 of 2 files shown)


zhongyujiang:
`Map` type used to have similar issue, but got fixed in Flink, see [here](https://issues.apache.org/jira/browse/FLINK-21247). Unforunately the similar [issue](https://issues.apache.org/jira/browse/FLINK-25238?jql=project%20%3D%20FLINK%20AND%20text%20~%20%22iceberg%20array%22) for `Array` type  have been closed.
I think we should continue to use custom `ReusableArrayData` for memory saving and solve the serialization problem in a similar way, instead of creating a new `GenericArrayData`.

zhongyujiang:
`Map` type used to have similar issue, but got fixed in Flink, see [here](https://issues.apache.org/jira/browse/FLINK-21247). Unforunately the similar [issue](https://issues.apache.org/jira/browse/FLINK-25238?jql=project%20%3D%20FLINK%20AND%20text%20~%20%22iceberg%20array%22) for `Array` type  have been closed.
I think we should continue to use custom `ReusableArrayData` for memory saving and solve the serialization problem in a similar way, instead of creating a new `GenericArrayData`.

openinx:
Yes, I generally agree with @zhongyujiang . The correct direction is improving the apache flink project to support customized `ArrayData` instead of adjusting the apache iceberg to support `GenericArrayData` only. 

Btw,  since we've got the https://issues.apache.org/jira/browse/FLINK-21247 resolved in the current flink release,   I think we'd better to add the unit tests also in the apache iceberg project to address the `ReusableMapData` issue.  The `ReusableArrayData` ditto, but can be marked as `Ignore` since we still get a correct flink release the fix this `ArrayData` issue.

yittg:
Thanks @zhongyujiang , @openinx for your comments, totally agree with you.
Actually this PR is only a trivial fix, it is basically for adding a test case for the issue and starting a discussion.

We can reuse the ReusableArrayData, and maybe copy it to GenericArrayData on a higher level without fixing it in Flink.
Of course, it should be better to fix it in Flink.

So, what's your opinion @zhongyujiang, @openinx

openinx:
Reconsidered about this thing again,  as we usually cannot align the iceberg release timeline with apache flink,  it's good to make this feature work fine rather than keeping this existing bug in the future iceberg release for waiting the flink bug fix release to accomplish the best performance.  I think we'd better to file an issue in apache flink firstly, and then add a TODO comment to refer it ( if we choose the option#1 temporarily).

openinx:
@rdblue Yes, I agreed.   The RP looks good to me now,  let's wait the travis CI results.

openinx (approved):

yittg:
Thanks @openinx , @rdblue , and @zhongyujiang.

We should try to push it be fixed in Flink.

And @openinx , should we port this walk around fix back to Flink 1.13, and the release branch 0.13.x?

Issue #4426, opened 2022-03-29
Support `rewrite-all` rewrite flag in Binpack


RussellSpitzer:
Currently only the Sort RewriteDataFiles strategy supports triggering a rewrite on all data files [regardless of size](https://github.com/apache/iceberg/blob/61fe9b1ce3ecff4354bb7624c9b5e3f76df2221b/core/src/main/java/org/apache/iceberg/actions/SortStrategy.java#L51-L56). Every once in a while we see a use case where a user wants to trigger a rewrite on every file, regardless of size or other criteria. I think it makes sense to allow for the `rewrite-all` flag to work in Binpack as well.

My suggestion would be that we take the property and move it to BinPackStrategy and allow both BinPackStrategy and SortStrategy to use the option. An additional test would be required as well to make sure that BinPack Strategy can use the new option.

rajarshisarkar:
Can I pick this up?

rajarshisarkar:
Thanks, for bringing this up. I'm working on this.

anuragmantri:
Thanks for creating this issue Russell. 

@rajarshisarkar - Feel free to add me to the PR. I will be happy to review.

rajarshisarkar:
Sure, @anuragmantri.

PR #4433, opened 2022-03-30
Core: Support `rewrite-all` rewrite flag in BinPackStrategy

rajarshisarkar:
Closes https://github.com/apache/iceberg/issues/4426

cc: @RussellSpitzer @anuragmantri


core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -48,14 +48,6 @@
...
-   * Rewrites all files, regardless of their size. Defaults to false, rewriting only mis-sized
-   * files;
-   */
-  public static final String REWRITE_ALL = "rewrite-all";
-  public static final boolean REWRITE_ALL_DEFAULT = false;
-
-
   private static final Set<String> validOptions = ImmutableSet.of(
       REWRITE_ALL
   );
```
anuragmantri, line 53: This is not needed anymore.

anuragmantri (commented):
Thanks for the change @rajarshisarkar. I have added few comments on refactoring.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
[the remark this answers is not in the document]
RussellSpitzer: I agree we should probably just expose rewrite all as a protected method and let the sub class use the value set in the parent (binpackstrat)

core/src/main/java/org/apache/iceberg/actions/BinPackStrategy.java
```diff
@@ -140,28 +149,39 @@ public RewriteStrategy options(Map<String, String> options) {
...
     return this;
   }
 
   @Override
   public Iterable<FileScanTask> selectFilesToRewrite(Iterable<FileScanTask> dataFiles) {
-    return FluentIterable.from(dataFiles)
-        .filter(scanTask -> scanTask.length() < minFileSize || scanTask.length() > maxFileSize ||
-                taskHasTooManyDeletes(scanTask));
+    if (rewriteAll) {
+      return dataFiles;
```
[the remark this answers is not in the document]
rajarshisarkar: I have added the logger and have used `super. selectFilesToRewrite()` in `SortStrategy`.

core/src/main/java/org/apache/iceberg/actions/BinPackStrategy.java
```diff
@@ -140,28 +149,39 @@ public RewriteStrategy options(Map<String, String> options) {
...
     List<List<FileScanTask>> potentialGroups = packer.pack(dataFiles, FileScanTask::length);
-
-    return potentialGroups.stream().filter(group ->
-            (group.size() >= minInputFiles && group.size() > 1) ||
-                (sizeOfInputFiles(group) > targetFileSize && group.size() > 1) ||
-                sizeOfInputFiles(group) > maxFileSize ||
-                group.stream().anyMatch(this::taskHasTooManyDeletes)
-    ).collect(Collectors.toList());
+    if (rewriteAll) {
+      return potentialGroups;
```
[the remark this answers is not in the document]
rajarshisarkar: Thanks, I have made the changes.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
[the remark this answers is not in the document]
rajarshisarkar: I have made the changes.

core/src/main/java/org/apache/iceberg/actions/BinPackStrategy.java
```diff
@@ -140,28 +149,39 @@ public RewriteStrategy options(Map<String, String> options) {
         DELETE_FILE_THRESHOLD,
         DELETE_FILE_THRESHOLD_DEFAULT);
 
+    rewriteAll = PropertyUtil.propertyAsBoolean(options,
+        REWRITE_ALL,
+        REWRITE_ALL_DEFAULT);
+
```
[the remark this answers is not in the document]
rajarshisarkar: Yes, I have made the changes.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -108,22 +84,12 @@ public RewriteStrategy options(Map<String, String> options) {
 
   @Override
```
RussellSpitzer, line 85: This override is not needed if we are only calling the super method

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -108,22 +84,12 @@ public RewriteStrategy options(Map<String, String> options) {
...
-      LOG.info("Sort Strategy for table {} set to rewrite all data files", table().name());
-      return dataFiles;
-    } else {
-      return super.selectFilesToRewrite(dataFiles);
-    }
+    return super.selectFilesToRewrite(dataFiles);
   }
 
   @Override
   public Iterable<List<FileScanTask>> planFileGroups(Iterable<FileScanTask> dataFiles) {
```
RussellSpitzer, line 91: This is also not needed if we are only calling the super method

core/src/main/java/org/apache/iceberg/actions/BinPackStrategy.java
```diff
@@ -240,6 +265,10 @@ protected long writeMaxFileSize() {
     return (long) (targetFileSize + ((maxFileSize - targetFileSize) * 0.5));
   }
 
+  protected boolean rewriteAll() {
```
RussellSpitzer, line 268: Not used

hameizi:
Is this chenge necessary? Because if we need rewrite all file in V2 table we can set `delete-file-threshold=0` , and in V1 table rewrite all file is default. If file size has reach targetFileSize it is unnecessary to rewrite.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
rajarshisarkar: Yeah, not required really. I have removed it.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
rajarshisarkar: Yeah, not required really. I have removed it.

core/src/main/java/org/apache/iceberg/actions/BinPackStrategy.java
rajarshisarkar: I have removed it.

rajarshisarkar:
>  Because if we need rewrite all file in V2 table we can set delete-file-threshold=0 

There are many other options that we need to override to ensure no filtering happens:

https://github.com/apache/iceberg/blob/4269536e5c48adba39f093aefcdd382c69f41fc7/core/src/main/java/org/apache/iceberg/actions/BinPackStrategy.java#L159-L164

`rewrite-all` is a single option which does that seamlessly. Also, there have been use cases where a user wants to trigger a rewrite on every file, regardless of size or other criteria: https://github.com/apache/iceberg/issues/4426#issue-1185049517

hameizi:
> Also, there have been use cases where a user wants to trigger a rewrite on every file, regardless of size or other criteria: [#4426 (comment)](https://github.com/apache/iceberg/issues/4426#issue-1185049517)

Get it, just i don't think it's a useful scene.

RussellSpitzer:
Thanks for the pr @rajarshisarkar  and thanks to @anuragmantri  and @hameizi for your review!

PR #4436, opened 2022-03-30
Python: Updates to literals

rdblue:
This includes a few updates to Python literals:
* Move imports to the top, import UUID
* Move constants to the top
* Use `Literal[...]` return types rather than specific classes
* Remove unnecessary `IntegerLiteral`
* Fix conversions (FloatLiteral and DoubleLiteral require float)
* Fix decimal conversions
* Extend test to validate None cannot be used as a literal value


python/src/iceberg/expression/literals.py
```diff
@@ -19,10 +19,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
-import datetime
+from datetime import datetime, date, time
```
dramaticlly, line 22: I think tox is complaining about the import order, we can probably reformat it with `tox -e format` to reorder it for you

python/src/iceberg/expression/literals.py
```diff
@@ -51,7 +52,12 @@
     UUIDType,
 )
 
-EPOCH = datetime.datetime.utcfromtimestamp(0)
+
+EPOCH_DATE = date.fromisoformat("1970-01-01")
+EPOCH_TIMESTAMP = datetime.fromisoformat("1970-01-01T00:00:00.000000")
+ISO_TIMESTAMP = re.compile(r"\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d(.\d{1,6})?")
+EPOCH_TIMESTAMPTZ = datetime.fromisoformat("1970-01-01T00:00:00.000000+00:00")
```
dramaticlly, line 59: nit, maybe worth renaming it to `EPOCH_TIMESTAMP_TZ` and `ISO_TIMESTAMP_TZ`

python/src/iceberg/expression/literals.py
```diff
@@ -200,78 +205,52 @@ def _(self, type_var):
         return self
 
 
-class IntegerLiteral(Literal[int]):
```
dramaticlly, line 203: lovely, did not realized we can actually ditch this entirely. But looking at the rest of the change altogether, it make more sense as LongLiteral is probably works fine behind the scene and when explicitly convert to IntegerType, we enforce the java like min and max but we are still return LongLiteral in the end but provide a better bound

python/src/iceberg/expression/literals.py
```diff
@@ -200,78 +205,52 @@ def _(self, type_var):
...
     @to.register(FloatType)
-    def _(self, type_var: FloatType) -> "FloatLiteral":
+    def _(self, type_var: FloatType) -> Literal[float]:
         return FloatLiteral(float(self.value))
 
     @to.register(DoubleType)
-    def _(self, type_var: DoubleType) -> "DoubleLiteral":
-        return DoubleLiteral(self.value)
+    def _(self, type_var: DoubleType) -> Literal[float]:
+        return DoubleLiteral(float(self.value))
```
dramaticlly, line 231: oh gotcha, this is about what you suggested [earlier on what's missing](https://github.com/apache/iceberg/pull/4262#discussion_r837642495), right? Thank you!

python/src/iceberg/expression/literals.py
rdblue: I think it's fine since these are usually one word.

dramaticlly (approved):
LGTM!

python/src/iceberg/expression/literals.py
rdblue: Yes, that's what I was referring to. Looks like Python allows float and int comparison as long as the value has an exact representation in floating point. That's why the tests still works, but this change is still needed to make sure the inner type is correct.

danielcweeks (approved):
LGTM (though lint failed)

samredai:
LGTM!

The lint failure is because mypy is confused by the `isinstance(...)` in the constructor for `Literal`. Once it sees isinstance it sets the type as a general `object`. I think the best thing to do is to add a mypy ignore comment to the value property for `Literal`:
```py
    @property
    def value(self) -> T:
        return self._value  # type: ignore
```

rdblue:
Thanks, @samredai! That fixed it.

PR #4376, opened 2022-03-22
ALL：fixing static analysis warnings

felixYyu:
this PR fixed these grouping parentheses are unnecessary. cc @rdblue @RussellSpitzer 
```
iceberg\core\src\main\java\org\apache\iceberg\BaseFileScanTask.java:224: [UnnecessaryParentheses] These grouping parentheses are unnecessary; it is unlikely the code will be misinterpreted without them
          (this.file().equals(other.file())) &&
          ^
    (see https://errorprone.info/bugpattern/UnnecessaryParentheses)
  Did you mean 'this.file().equals(other.file()) &&'?
```

PR #4437, opened 2022-03-31
Static Analysis: Supressions we know are false positive

felixYyu:
fix #4376 cc @RussellSpitzer


felixYyu:
please take a look again, thanks.  Static Analysis has 8 tasks.

core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -300,49 +301,46 @@ private void tryRunOnFailure(I item, Exception failure) {
...
-              // run the task with retries
-              boolean threw = true;
-              try {
-                runTaskWithRetry(task, item);
+        futures.add(service.submit(() -> {
+          if (!(stopOnFailure && taskFailed.get())) {
+            // run the task with retries
+            boolean threw = true;
+            try {
+              runTaskWithRetry(task, item);
```
rdblue, line 309: I don't think that suppressing warnings should change this class at all.

felixYyu: revert

PR #4438, opened 2022-03-31
Static Analysis: Rewrite  empty and inappropriate Javadocs

felixYyu:
fix #4376 cc @RussellSpitzer


core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -108,7 +108,8 @@ public TableScan appendsAfter(long fromSnapshotId) {
     }
 
     /**
-     * @return list of manifest files to explore for this files metadata table scan
+     * Returns list of {@link ManifestFile} files to explore for this file's metadata table scan
+     * @return list of manifest files
```
rdblue, line 112: Why is this change needed?

hive3/src/main/java/org/apache/hadoop/hive/ql/io/orc/OrcSplit.java
```diff
@@ -191,12 +191,12 @@ public boolean hasFooter() {
   }
 
   /**
-   * @return {@code true} if file schema doesn't have Acid metadata columns
    * Such file may be in a delta_x_y/ or base_x due to being added via
    * "load data" command.  It could be at partition|table root due to table having
    * been converted from non-acid to acid table.  It could even be something like
    * "warehouse/t/HIVE_UNION_SUBDIR_15/000000_0" if it was written by an
    * "insert into t select ... from A union all select ... from B"
+   * @return {@code true} if file schema doesn't have Acid metadata columns
```
rdblue, line 199: I think this can be improved, but moving this line to the end doesn't help readability. Can you rewrite this doc instead?

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
felixYyu: the static check warning log ，so I think should need change:

```
[MissingSummary] A summary fragment is required; consider using the value of the @return block as a summary fragment instead.
     * @return list of manifest files to explore for this files metadata table scan
       ^
    (see https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment)
  Did you mean '*Returns list of manifest files to explore for this files metadata table scan.'?
```

hive3/src/main/java/org/apache/hadoop/hive/ql/io/orc/OrcSplit.java
felixYyu: removing redundant e.g.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -27,9 +27,6 @@
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.rest.RESTResponse;
 
-/**
- *
- */
```
rdblue, line 32: I think the correct fix is to add javadoc, not remove the placeholder.

hive3/src/main/java/org/apache/hadoop/hive/ql/io/orc/OrcSplit.java
rdblue: I think the right fix is to document the method, not to just move the `@return`. Can you fix the documentation here instead? Javadoc should start with a short (one sentence) description and then go into more detail if needed after a paragraph break.

felixYyu: thanks, I got it and tried to description it.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
felixYyu: added javadoc

felixYyu:
thanks very much @RussellSpitzer @rdblue suggests, Help me to impove my skills.

felixYyu:
Can you take a look if have time? @rdblue

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -115,7 +115,7 @@ protected BaseAllFilesTableScan(TableOperations ops, Table table, Schema schema,
     }
 
     /**
-     * Returns an iterable of manifest files to explore for this all files metadata table scan
+     * Returns list of {@link ManifestFile} files to explore for this file's metadata table scan
```
rdblue, line 118: This isn't correct. Can you revert these changes?

hive3/src/main/java/org/apache/hadoop/hive/ql/io/orc/OrcSplit.java
```diff
@@ -191,12 +191,15 @@ public boolean hasFooter() {
...
-   * been converted from non-acid to acid table.  It could even be something like
+   * A check for whether or not this file was originally created in a Hive ACID table.
+   *
+   * A file may be in a delta_x_y/ or base_x due to being added via
+   * "load data" command.  It could be located in a partition|table root due to table having
+   * been converted from non-acid to acid table.  It could also even have paths containing
    * "warehouse/t/HIVE_UNION_SUBDIR_15/000000_0" if it was written by an
    * "insert into t select ... from A union all select ... from B"
+   *
+   * @return true if the file was written originally by a Hive ACID table
```
rdblue, line 202: There is no need to change this `@return`. The new version is more confusing because it is up to the Javadoc reader to know what it means to be originally created in a Hive ACID table. Instead, just use the original, "true if the file schema does not have ACID metadata columns". That should also be the first sentence description.

In addition, Javadoc paragraphs should be separated by `<p>` on its own line.

felixYyu: thanks, fixed

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
felixYyu: reverted

PR #4439, opened 2022-03-31
Static Analysis: Missing private constructor, method need return single mutable, and other few static warning

felixYyu:
fix #4376 cc @RussellSpitzer


mr/src/test/java/org/apache/iceberg/mr/TestIcebergInputFormats.java
```diff
@@ -337,7 +339,7 @@ public void testLocality() throws Exception {
     helper.appendToTable(null, expectedRecords);
 
     for (InputSplit split : testInputFormat.create(builder.conf()).getSplits()) {
-      Assert.assertArrayEquals(IcebergSplit.ANYWHERE, split.getLocations());
+      Assert.assertArrayEquals(ANYWHERE, split.getLocations());
```
rdblue, line 342: What was the problem here?

rdblue:
This PR goes well beyond "Changing parens". Can you update the summary and description with the motivation and what you're actually changing?

mr/src/test/java/org/apache/iceberg/mr/TestIcebergInputFormats.java
felixYyu: because the IcebergSplit check warning log, so make public 'ANYWHERE' var to private. the TestIcebergInputFormats need define a private internal variable.

```
mr\src\main\java\org\apache\iceberg\mr\mapreduce\IcebergSplit.java:38: [MutablePublicArray] Non-empty arrays are mutable, so this `public static final` array is not a constant and can be modified by clients of this class.  Prefer an ImmutableList, or provide an accessor method that returns a defensive copy.
  public static final String[] ANYWHERE = new String[]{"*"};
                               ^
    (see https://errorprone.info/bugpattern/MutablePublicArray)
```

felixYyu:
- DynClasses DynFields SnapshotSummary : A Builder with a static factory method on the encapsulating class must have a private constructor
- BaseFileScanTask : This method returns both mutable and immutable collections or maps from different paths,This method returns both mutable and immutable collections or maps from different paths.
- FlinkCatalog : This method returns both mutable and immutable collections or maps from different paths
- ParquetMetricsRowGroupFilter : Integer id = ref.fieldId(); id type maybe privimitive

felixYyu:
Do you have any suggestions about summary？@rdblue

rdblue:
I think this should be split into different PRs for specific fixes.

felixYyu:
- [ ] #4481

felixYyu:
split into different PRs for specific fixes. and close thi PR.

PR #4443, opened 2022-03-31
Docs: Add S3 delete tagging docs

rajarshisarkar:
This PR add docs for https://github.com/apache/iceberg/pull/4342

cc: @rdblue @danielcweeks @jackye1995 @arminnajafi @singhpk234 @amogh-jahagirdar @xiaoxuandev @yyanyy


docs/integrations/aws.md
```diff
@@ -433,6 +433,25 @@ spark-sql --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCata
...
+    --conf spark.sql.catalog.my_catalog.io-impl=org.apache.iceberg.aws.s3.S3FileIO \
+    --conf spark.sql.catalog.my_catalog.s3.write.tags.my_ke1y=my_val1 \
+    --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key2=my_val2 \
+    --conf spark.sql.catalog.my_catalog.s3.delete.tags.my_key3=my_val3 \
+    --conf spark.sql.catalog.my_catalog.s3.delete-enabled=false
+```
+
+When `s3.delete-enabled` is disabled, users are expected to set the delete tags with `s3.delete.tags` and manage the deleted files through S3 lifecycle policy.
+With the `s3.delete.tags` config, objects are tagged with the configured key-value pairs before deletion. This is considered a soft-delete, because users are able to configure tag-based object lifecycle policy at bucket level to transition objects to different tiers.
+For more details, see [managing your storage lifecycle documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html).
```
singhpk234, line 451: should we add this is note section, also I believe we should document `s3.delete-enabled` in a seperate section as if this is disabled we will skip deleting & consider output file as soft-deleted with / without delete tagging in place.

rajarshisarkar: I don't see note sections being added in the AWS docs. I have added a new section for `s3.delete-enabled`.

docs/integrations/aws.md
```diff
@@ -419,9 +419,22 @@ If for any reason you have to use S3A, here are the instructions:
 To ensure integrity of uploaded objects, checksum validations for S3 writes can be turned on by setting catalog property `s3.checksum-enabled` to `true`. 
 This is turned off by default.
 
+### S3 Delete 
+
+When `s3.delete-enabled` is disabled, the objects are not deleted from S3.
+For example, to disable deletion with Spark 3.0, you can start the Spark SQL shell with:
```
singhpk234, line 425: [minor] can add what is the default value (though it might be obvious) , referring the checksum verification example above

docs/integrations/aws.md
```diff
@@ -433,6 +446,25 @@ spark-sql --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCata
...
+    --conf spark.sql.catalog.my_catalog.s3.write.tags.my_key2=my_val2 \
+    --conf spark.sql.catalog.my_catalog.s3.delete.tags.my_key3=my_val3 \
+    --conf spark.sql.catalog.my_catalog.s3.delete-enabled=false
+```
+
+When `s3.delete-enabled` is disabled, users are expected to set the delete tags with `s3.delete.tags` and manage the deleted files through S3 lifecycle policy.
+With the `s3.delete.tags` config, objects are tagged with the configured key-value pairs before deletion. This is considered a soft-delete, because users can configure tag-based object lifecycle policy at bucket level to transition objects to different tiers.
+For more details, see [managing your storage lifecycle documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html).
+
+For the above example, the objects in S3 will be saved with tags: `my_key1=my_val1`, `my_key2=my_val2` and `my_key3=my_val3` before deletion.
```
singhpk234, line 466: since in [L#459](https://github.com/apache/iceberg/pull/4443/files#diff-45ac9a4ad0d5c90123d0537d61bdcd243a7c06b684b39a3914b831ee35c11052R459) we have disabled delete should we call it soft-delete here ? Or can we remove the s3.delete-enabled conf from the example as per my understanding it adds less value to the context of the delete tags now (since we moved this in a separate section).

docs/integrations/aws.md
```diff
@@ -433,6 +446,25 @@ spark-sql --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCata
 ```
 For the above example, the objects in S3 will be saved with tags: `my_key1=my_val1` and `my_key2=my_val2`.
```
singhpk234, line 447: I think we might need to call out here that these tags specified will be used only in object creation like we called out that delete tags will be used in tagging in delete operation

singhpk234 (commented):
Thanks @rajarshisarkar for this change !!!

docs/integrations/aws.md
rajarshisarkar: I have ddded the default value.

rajarshisarkar: I have added the default value.

docs/integrations/aws.md
rajarshisarkar: I have added the details.

docs/integrations/aws.md
rajarshisarkar: Called it out as soft-delete and removed the `s3.delete-enabled` details from this section.

docs/integrations/aws.md
```diff
@@ -431,7 +445,23 @@ spark-sql --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCata
...
+    --conf spark.sql.catalog.my_catalog.catalog-impl=org.apache.iceberg.aws.glue.GlueCatalog \
+    --conf spark.sql.catalog.my_catalog.io-impl=org.apache.iceberg.aws.s3.S3FileIO \
+    --conf spark.sql.catalog.my_catalog.s3.delete.tags.my_key3=my_val3
+```
+
+Users are expected to set the delete tags with `s3.delete.tags` and manage the deleted files through S3 lifecycle policy.
+With the `s3.delete.tags` config, objects are tagged with the configured key-value pairs before deletion. This is considered a soft-delete, because users can configure tag-based object lifecycle policy at bucket level to transition objects to different tiers.
+For more details, see [managing your storage lifecycle documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html).
+
+For the above example, the objects in S3 are considered as a soft delete and will be saved with tags: `my_key3=my_val3` before deletion.
```
singhpk234, line 464: [question] why are we considering it as soft-delete as we have not set `s3.delete-enabled` as false. The object will be actually deleted right ?

singhpk234: should we just say here object will be tagged with `my_key3=my_val3` before deletion. WDYT

singhpk234 (approved):
LGTM @rajarshisarkar !!!, just a question on soft-delete

docs/integrations/aws.md
```diff
@@ -431,7 +445,24 @@ spark-sql --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCata
...
+```
+sh spark-sql --conf spark.sql.catalog.my_catalog=org.apache.iceberg.spark.SparkCatalog \
+    --conf spark.sql.catalog.my_catalog.warehouse=s3://iceberg-warehouse/s3-tagging \
+    --conf spark.sql.catalog.my_catalog.catalog-impl=org.apache.iceberg.aws.glue.GlueCatalog \
+    --conf spark.sql.catalog.my_catalog.io-impl=org.apache.iceberg.aws.s3.S3FileIO \
+    --conf spark.sql.catalog.my_catalog.s3.delete.tags.my_key3=my_val3
+```
+
+Users are expected to set the delete tags with `s3.delete.tags` and manage the deleted files through S3 lifecycle policy.
+With the `s3.delete.tags` config, objects are tagged with the configured key-value pairs before deletion. This is considered a soft-delete, because users can configure tag-based object lifecycle policy at bucket level to transition objects to different tiers.
```
amogh-jahagirdar, line 461: I would leave out "before deletion." and replace it with "instead of deleting". I would also take out "This is considered a soft-delete". My thought is we can just say "Users can configure tag-based object lifecycle policy at bucket level to transition objects to different tiers."

rajarshisarkar: > I would leave out "before deletion." and replace it with "instead of deleting".

I am carrying the term forward from https://github.com/apache/iceberg/pull/4342#discussion_r831485921 as for "tag before delete" use-case it might make sense: https://github.com/apache/iceberg/pull/4342#issuecomment-1073347947

> I would also take out "This is considered a soft-delete". 

Makes sense.

docs/integrations/aws.md
```diff
@@ -419,9 +419,23 @@ If for any reason you have to use S3A, here are the instructions:
 To ensure integrity of uploaded objects, checksum validations for S3 writes can be turned on by setting catalog property `s3.checksum-enabled` to `true`. 
 This is turned off by default.
 
+### S3 Delete 
+
+When the catalog property `s3.delete-enabled` is set to `false`, the objects are not deleted from S3.
```
jackye1995, line 424: we should highlight "hard-deleted" instead of just "deleted"

docs/integrations/aws.md
```diff
@@ -419,9 +419,23 @@ If for any reason you have to use S3A, here are the instructions:
 To ensure integrity of uploaded objects, checksum validations for S3 writes can be turned on by setting catalog property `s3.checksum-enabled` to `true`. 
 This is turned off by default.
 
+### S3 Delete 
+
+When the catalog property `s3.delete-enabled` is set to `false`, the objects are not deleted from S3. 
+This is set to `true` by default.
```
jackye1995, line 425: we should link to the next section with something in the line of "This is expected to be used in combination with s3 delete tagging", so objects are tagged and removed using S3 lifecycle policy. See xxx for more details".

jackye1995: Actually, given that the delete-enabled is only relevant under the context of tagging, would it make more sense to just mention this in the tagging section below, instead of having its own section?

samredai:
This isn't a comment about this PR specifically but the content in the aws.md file feels like a collection of many unrelated things under the AWS umbrella. Maybe we should start thinking about how to break this up, such as moving the S3 docs to a FileIO section, move the Spark material somewhere in the Spark section, etc.

rajarshisarkar:
Thanks, for the inputs @samredai !

@jackye1995 Please let me know your thoughts on the scope of improvement for `aws.md`. We can create tasks accordingly.

jackye1995 (approved):
Thanks for the work!

samredai:
I have a PR for consolidating configurations, it might be good to add these properties there like `spark.sql.catalog.<catalog>.s3.delete.tags.<key>`: https://github.com/apache/iceberg/pull/4801

rajarshisarkar:
@samredai Yeah, makes sense to me.

samredai:
Added to the configurations page PR: https://github.com/apache/iceberg/pull/4801/files#diff-947e999f40ea405c702496fc987ab2c322222d2488680be54ddfa3447c3569faR154-R158

PR #4444, opened 2022-03-31
Changes in date handling classes

felixYyu:
fix #4376 cc @RussellSpitzer


rdblue:
@felixYyu can you update the description with why these changes are needed and what the result is? What is the difference between `Date.toString` and `Instant.toString`?

felixYyu:
> What is the difference between Date.toString and Instant.toString?
because Date has a bad API that leads to bugs, the static check error log:
```
iceberg\core\src\main\java\org\apache\iceberg\RemoveSnapshots.java:121: [JavaUtilDate] Date has a bad API that leads to bugs; prefer java.time.Instant or LocalDate.
    LOG.info("Expiring snapshots older than: {} ({})", new Date(timestampMillis), timestampMillis);
                                                       ^
    (see https://errorprone.info/bugpattern/JavaUtilDate)
```

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -118,7 +118,7 @@ public ExpireSnapshots expireSnapshotId(long expireSnapshotId) {
 
   @Override
   public ExpireSnapshots expireOlderThan(long timestampMillis) {
-    LOG.info("Expiring snapshots older than: {} ({})", new Date(timestampMillis), timestampMillis);
+    LOG.info("Expiring snapshots older than: {} ({})", Instant.ofEpochMilli(timestampMillis), timestampMillis);
```
rdblue, line 121: How about using `DateTimeUtil.formatTimestampMillis` instead?

felixYyu: ScanSummary also fixed using DateTimeUtil.formatTimestampMillis instead.

PR #4446, opened 2022-03-31
Spark [3.2] : Improve stats estimation for spark scan

singhpk234:
As per my understanding, presently we don't take `SplitScanTask`(which also implements FileScanTask),into consideration where we would _**not**_ be reading the complete file rather a subpart of the file. 

So in this case it's better to find number of records actually scanned by the task, rather than going to datafile this splitscanTask is based on and get's its complete record count. 

At present it seems like we are over-estimating the size by taking complete record count of the file associated with the SplitScanTask, where as we are just scanning the subset of data from this file.

---

cc @rdblue @aokolnychyi @RussellSpitzer @jackye1995  @wypoon

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
@@ -145,10 +145,12 @@ abstract class SparkScan implements Scan, SupportsReportStatistics {
     }
 
     long numRows = 0L;
+    long defaultRowSizeInBytes = readSchema().defaultSize();
 
     for (CombinedScanTask task : tasks()) {
       for (FileScanTask file : task.files()) {
-        numRows += file.file().recordCount();
+        // TODO: if possible, take deletes also into consideration.
+        numRows += Math.min(file.length() / defaultRowSizeInBytes, file.file().recordCount());
       }
     }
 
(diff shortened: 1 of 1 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -145,10 +145,12 @@ protected Statistics estimateStatistics(Snapshot snapshot) {
     }
 
     long numRows = 0L;
+    long defaultRowSizeInBytes = readSchema().defaultSize();
 
     for (CombinedScanTask task : tasks()) {
       for (FileScanTask file : task.files()) {
-        numRows += file.file().recordCount();
+        // TODO: if possible, take deletes also into consideration.
+        numRows += Math.min(file.length() / defaultRowSizeInBytes, file.file().recordCount());
```
rdblue, line 153: We always want to base the estimate on the number of records in the file and not on the file length. Calculations based on the file length are quite bad in practice because Parquet compression varies.

If you want to find out how many records from the file will actually be scanned by this task, then you could do something different, like finding the proportion of the file that is covered by the task and multiplying by the record count.

rdblue (approved):

rdblue:
Thanks, @singhpk234!

PR #4448, opened 2022-03-31
Core: Add in RenameTableRequest for the RESTCatalog

kbendick:
Adds `RenameTableRequest` as a request for the `RESTCatalog` and adds the call for `renameTable` to the `RESTCatalog`.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java
@@ -194,7 +194,9 @@ public class JdbcCatalog extends BaseMetastoreCatalog
   public void renameTable(TableIdentifier from, TableIdentifier to) {
     int updatedRecords = execute(
         err -> {
-          if (err instanceof SQLIntegrityConstraintViolationException) {
+          // SQLite doesn't set SQLState or throw SQLIntegrityConstraintViolationException
+          if (err instanceof SQLIntegrityConstraintViolationException ||
+              err.getMessage() != null && err.getMessage().contains("constraint failed")) {
             throw new AlreadyExistsException("Table already exists: %s", to);
           }
         },
--- a/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
+++ b/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
@@ -47,6 +47,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.rest.requests.CreateNamespaceRequest;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
+import org.apache.iceberg.rest.requests.RenameTableRequest;
 import org.apache.iceberg.rest.requests.UpdateNamespacePropertiesRequest;
 import org.apache.iceberg.rest.requests.UpdateTableRequest;
 import org.apache.iceberg.rest.responses.CreateNamespaceResponse;
@@ -243,6 +244,10 @@ public class CatalogHandlers {
     return LoadTableResponse.builder()
         .withTableMetadata(finalMetadata)
         .build();
+  }
+
+  public static void renameTable(Catalog catalog, RenameTableRequest request) {
+    catalog.renameTable(request.source(), request.destination());
   }
 
   private static boolean isCreate(UpdateTableRequest request) {
--- a/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
@@ -53,6 +53,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.rest.requests.CreateNamespaceRequest;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
+import org.apache.iceberg.rest.requests.RenameTableRequest;
 import org.apache.iceberg.rest.requests.UpdateNamespacePropertiesRequest;
 import org.apache.iceberg.rest.responses.ConfigResponse;
 import org.apache.iceberg.rest.responses.CreateNamespaceResponse;
@@ -129,7 +130,13 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
 
   @Override
   public void renameTable(TableIdentifier from, TableIdentifier to) {
+    RenameTableRequest request = RenameTableRequest.builder()
+        .withSource(from)
+        .withDestination(to)
+        .build();
 
+    // for now, ignore the response because there is no way to return it.
+    client.post("v1/tables/rename", request, null, ErrorHandlers.tableErrorHandler());
   }
 
   private LoadTableResponse loadInternal(TableIdentifier identifier) {
--- a/core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest.requests;
+
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.rest.RESTRequest;
+
+/**
+ * A REST request to rename a table.
+ */
+public class RenameTableRequest implements RESTRequest {
+
+  private TableIdentifier source;
+  private TableIdentifier destination;
+
+  @SuppressWarnings("unused")
+  public RenameTableRequest() {
+    // Needed for Jackson Deserialization.
+  }
+
+  private RenameTableRequest(TableIdentifier source, TableIdentifier destination) {
+    this.source = source;
+    this.destination = destination;
+    validate();
+  }
+
+  public void validate() {
+    Preconditions.checkArgument(source != null, "Invalid source table: null");
+    Preconditions.checkArgument(destination != null, "Invalid destination table: null");
+  }
+
+  public TableIdentifier source() {
+    return source;
+  }
+
+  public TableIdentifier destination() {
+    return destination;
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("source", source)
+        .add("destination", destination)
+        .toString();
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+    private TableIdentifier source;
+    private TableIdentifier destination;
+
+    private Builder() {
+    }
+
+    public Builder withSource(TableIdentifier sourceTable) {
+      Preconditions.checkNotNull(sourceTable, "Invalid source table identifier: null");
+      this.source = sourceTable;
+      return this;
+    }
+
+    public Builder withDestination(TableIdentifier destinationTable) {
+      Preconditions.checkNotNull(destinationTable, "Invalid destination table identifier: null");
+      this.destination = destinationTable;
+      return this;
+    }
+
+    public RenameTableRequest build() {
+      return new RenameTableRequest(source, destination);
+    }
+  }
+}
--- a/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
+++ b/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
@@ -31,6 +31,7 @@ import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DataFiles;
 import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.HasTableOperations;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.ReplaceSortOrder;
 import org.apache.iceberg.Schema;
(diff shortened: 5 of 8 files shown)


core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
```diff
@@ -0,0 +1,99 @@
...
+    this.source = source;
+    this.destination = destination;
+    validate();
+  }
+
+  public RenameTableRequest validate() {
+    Preconditions.checkArgument(source != null, "Invalid source table: null");
+    Preconditions.checkArgument(destination != null, "Invalid destination table: null");
+    Preconditions.checkArgument(source.hasNamespace(), "Invalid source table: missing namespace");
+    Preconditions.checkArgument(destination.hasNamespace(), "Invalid destination table: missing namespace");
```
kbendick, line 50: Do we want to allow for the empty namespace to be used? Its not allowed in some other catalogs, but it seems like that might be beneficial for e.g. catalogs that don't support namespaces.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -516,6 +517,29 @@ public void testLoadMissingTable() {
...
+
+    Assert.assertFalse("Destination table should not exist before rename", catalog.tableExists(RENAMED_TABLE));
+
+    catalog.renameTable(TABLE, RENAMED_TABLE);
+    Assert.assertTrue("Table should exist with new name", catalog.tableExists(RENAMED_TABLE));
+    Assert.assertFalse("Original table should no longer exist", catalog.tableExists(TABLE));
+
+    catalog.dropTable(RENAMED_TABLE);
+    assertEmpty("Should not contain table after drop", catalog, NS);
+  }
```
kbendick, line 541: This test is pretty basic as I didn't want to make the PR too large.

Let me know if I should add more `CatalogTests` for the `renameTable` operation in this PR 🙂

core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
rdblue: Yes, I don't see a reason not to allow requesting the empty namespace. That's valid in many catalogs.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
rdblue: I'd prefer adding tests for missing source table, existing destination table, and others. It's a good time to add them.

core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
kbendick: Ok will remove that.

kbendick: Removed.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
kbendick: I added test for missing source and existing destination.

However, the JdbcCatalog does not throw the correct error for existing destination table, which can be verified by looking at the current JdbcCatalog tests: https://github.com/apache/iceberg/blob/62f3ca334c4bf377b1aaf9b7386bec1bfaddb5cd/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java#L385-L390

I have put togeher a hacky fix to it in this PR, but I think we should just throw an `@Ignore` on this particular test and then fix the jdbc catalog in a separate PR. All of the inforrmation from the error was very SQLIte specific and I'm not sure of a general way to handle this. Open to suggestions.

core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java
```diff
@@ -194,7 +194,8 @@ public List<TableIdentifier> listTables(Namespace namespace) {
   public void renameTable(TableIdentifier from, TableIdentifier to) {
     int updatedRecords = execute(
         err -> {
-          if (err instanceof SQLIntegrityConstraintViolationException) {
+          if (err instanceof SQLIntegrityConstraintViolationException ||
+              err.getMessage() != null && err.getMessage().contains("UNIQUE constraint failed")) {
```
kbendick, line 198: I don't think this test is going to hold up for most JDBC databases.

There's a SQLCode on the error, but it was SQLIte specific in the test runs.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
rdblue: I hit the same problem when testing create table atomicity and [fixed it similarly](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/jdbc/JdbcTableOperations.java#L126-L129). There's more background in [this thread](https://github.com/apache/iceberg/pull/4319#pullrequestreview-909504774).

I'd probably use the same logic as in that fix, which checks the message for "constraint failed".

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -245,6 +246,11 @@ public static LoadTableResponse updateTable(Catalog catalog, TableIdentifier ide
         .build();
   }
 
+  public static void renameTable(Catalog catalog, RenameTableRequest request) {
+    request.validate();
```
rdblue, line 250: I don't think other methods in `CatalogHandlers` run `validate`. I think that should be done before passing it in here.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -562,6 +563,87 @@ public void testLoadMissingTable() {
...
+
+  @Test
+  public void testRenameTableMissingSourceTable() {
+    C catalog = catalog();
+
+    if (requiresNamespaceCreate()) {
+      catalog.createNamespace(NS);
+    }
+
+    Assert.assertFalse("Source table should not exist before create", catalog.tableExists(TABLE));
```
rdblue, line 597: `create` -> `rename`.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -562,6 +563,87 @@ public void testLoadMissingTable() {
...
+     *
+     *   Updating JDBC catalog seems to have fixed it, but it's pretty hacky and probably pretty SQLite specific.
+     */
+    AssertHelpers.assertThrows("Should reject renaming a table if the new name already exists",
+        AlreadyExistsException.class,
+        "Table already exists",
+        () -> catalog.renameTable(TABLE, RENAMED_TABLE));
+
+    Assert.assertTrue("Source table should still exist after failed rename", catalog.tableExists(TABLE));
+    Assert.assertTrue("Destination table should still exist after failed rename", catalog.tableExists(RENAMED_TABLE));
```
rdblue, line 640: It would be nice to have a validation that the destination table is distinct from the source table. Maybe checking UUID or making it a partitioned table and ensuring it is still partitioned.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -562,6 +563,87 @@ public void testLoadMissingTable() {
...
+
+    catalog.dropTable(TABLE);
+    catalog.dropTable(RENAMED_TABLE);
+    assertEmpty("Should not contain table after drop", catalog, NS);
```
rdblue, line 644: These get cleaned up with the catalog, so no need to drop and assert.

core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java
rdblue: We looked into this and there isn't a good way to do it. In `JdbcTableOperations`, we do the same thing but check for just "constraint failed". I pasted some links for background in my other comment.

rdblue (changes requested):
@kbendick, this looks ready to go to me, except for the unfinished changes from the JDBC catalog issue. If you remove the comments in tests and update the `contains` check, I'll merge this.

kbendick:
> @kbendick, this looks ready to go to me, except for the unfinished changes from the JDBC catalog issue. If you remove the comments in tests and update the `contains` check, I'll merge this.

Cool. I figured that the JDBC catalog issue was something there wasn't a clean way to resolve.

I will make the changes so we can merge this.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
kbendick: Oh good point they don't. Since we're building them via the builder, it's not really necessary. Plus validate is called in the constructor anyway.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
kbendick: I'll check the UUIDs. That seems simplest.

core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    this.source = source;
+    this.destination = destination;
+    validate();
+  }
+
+  public void validate() {
+    Preconditions.checkArgument(source != null, "Invalid source table: null");
+    Preconditions.checkArgument(destination != null, "Invalid destination table: null");
+    Preconditions.checkArgument(!source.equals(destination),
+        "Invalid rename table request: source must not be the same as destination");
```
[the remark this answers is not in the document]
kbendick: I suppose it can be a no-op. I looked at the Glue catalog, and it doesn't specifically check for this case.

In terms of making it a no-op, should we just let the client send the request and allow the server to deal with it?

core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java
kbendick: Updated to use the referenced code (with the same comment).

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
kbendick: Updated the check.

core/src/main/java/org/apache/iceberg/rest/requests/RenameTableRequest.java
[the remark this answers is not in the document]
kbendick: Allowing it to go through and the server can deal with it.

This is in line with some of the other catalogs.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -562,6 +564,79 @@ public void testLoadMissingTable() {
...
+
+    AssertHelpers.assertThrows("Should reject renaming a table that does not exist",
+        NoSuchTableException.class,
+        "Table does not exist",
+        () -> catalog.renameTable(TABLE, RENAMED_TABLE));
+
+    Assert.assertFalse("Destination table should not exist after failed rename", catalog.tableExists(RENAMED_TABLE));
+  }
+
+  // WIP - This needs to throw AlreadyExistsException but its giving UncheckedSQLException
```
rdblue, line 609: Unnecessary comment?

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -113,7 +114,14 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
   @Override
   public void renameTable(TableIdentifier from, TableIdentifier to) {
+    RenameTableRequest request = RenameTableRequest.builder()
+        .withSource(from)
+        .withDestination(to)
+        .build();
 
+    // for now, ignore the response because there is no way to return it.
+    // also, the spec does not define a response beyond simply `ok`.
+    client.post("v1/tables/rename", request, null, ErrorHandlers.tableErrorHandler());
```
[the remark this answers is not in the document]
kbendick: Given it's a POST request, the 200 won't be cached so I think either is fine.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -562,6 +564,79 @@ public void testLoadMissingTable() {
...
+        "Table already exists",
+        () -> catalog.renameTable(TABLE, RENAMED_TABLE));
+
+    Assert.assertTrue("Source table should still exist after failed rename", catalog.tableExists(TABLE));
+    Assert.assertTrue("Destination table should still exist after failed rename", catalog.tableExists(RENAMED_TABLE));
+
+    String sourceTableUUID = ((HasTableOperations) catalog.loadTable(TABLE)).operations().current().uuid();
+    String destinationTableUUID = ((HasTableOperations) catalog.loadTable(RENAMED_TABLE)).operations().current().uuid();
+    Assert.assertNotEquals("Source and destination table should remain distinct after failed rename",
+        sourceTableUUID, destinationTableUUID);
```
rdblue, line 637: Looks good.

rdblue (approved):
Looks great. I think there was a `WIP` comment accidentally left in, but overall looks good to merge.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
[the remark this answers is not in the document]
kbendick: Looking at the spec, I think that 204 might not be the best, as that indicates that the page should not be refreshed.

If a web UI were to receive a 204 response, it would possibly not refresh the page, but if the rename is successful and the page lists the current tables, then the UI should refresh.

https://www.rfc-editor.org/rfc/rfc2616#section-10.2.5

```
[10.2.5](https://www.rfc-editor.org/rfc/rfc2616#section-10.2.5) 204 No Content

   The server has fulfilled the request but does not need to return an
   entity-body, and might want to return updated metainformation. The
   response MAY include new or updated metainformation in the form of
   entity-headers, which if present SHOULD be associated with the
   requested variant.

   If the client is a user agent, it SHOULD NOT change its document view
   from that which caused the request to be sent. This response is
   primarily intended to allow input for actions to take place without
   causing a change to the user agent's active document view, although
   any new or updated metainformation SHOULD be applied to the document
   currently in the user agent's active view.

   The 204 response MUST NOT include a message-body, and thus is always
   terminated by the first empty line after the header fields.
```

That said, given that there's no response body, I am open either way.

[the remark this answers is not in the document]
kbendick: The 200 content does say the POST body should have the updated information in it though, so I'm ok either way: https://www.rfc-editor.org/rfc/rfc2616#section-10.2.1

```
[10.2.1](https://www.rfc-editor.org/rfc/rfc2616#section-10.2.1) 200 OK

   The request has succeeded. The information returned with the response
   is dependent on the method used in the request, for example:

   GET    ....

   POST   an entity describing or containing the result of the action;
```

kbendick:
> Looks great. I think there was a `WIP` comment accidentally left in, but overall looks good to merge.

Just noticed it. Removed!

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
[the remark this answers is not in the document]
rdblue: I'm good with 200.

kbendick:
This should be good to go when you get a chance. Rebased and removed unnecessary commits.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
kbendick: Removed!

rdblue (approved):

PR #4449, opened 2022-03-31
Fix assert message error

zhangjun0x01:
I think the assert message maybe error.I fixed it.


spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1084,14 +1084,14 @@ protected void shouldHaveACleanCache(Table table) {
     List<Pair<Pair<T, T>, Pair<T, T>>>
         overlappingFiles = checkForOverlappingFiles(table, column);
 
-    Assert.assertEquals("Found overlapping files", Collections.emptyList(), overlappingFiles);
+    Assert.assertEquals("Found no overlapping files", Collections.emptyList(), overlappingFiles);
```
rdblue, line 1087: I think this was correct before.

PR #4450, opened 2022-03-31
Aliyun: Deprecate OSSInputStream public constructor

rajarshisarkar:
This PR is a follow-up of @rdblue's feedback: https://github.com/apache/iceberg/pull/4393#discussion_r838900710

Deprecating the constructor would maintain parity with the other InputStream classes like `GCSInputStream`, `EcsSeekableInputStream`, and `S3InputStream` which have a package private constructor.

cc: @rdblue

### DIFF
--- a/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
+++ b/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
@@ -35,6 +35,10 @@ import org.apache.iceberg.relocated.com.google.common.io.ByteStreams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * @deprecated moving to package-private in 0.15.0
+ */
+@Deprecated
 public class OSSInputStream extends SeekableInputStream {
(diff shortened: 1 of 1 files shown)


aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSInputStream.java
```diff
@@ -51,6 +51,10 @@
   private final Counter<Long> readBytes;
   private final Counter<Integer> readOperations;
 
+  /**
+   * @deprecated moving to package-private in 0.15.0
+   */
+  @Deprecated
```
rdblue, line 57: Should the class itself also be moved to package-private?

rajarshisarkar: Yes, that would maintain parity with `GCSInputStream`, `EcsSeekableInputStream`, and `S3InputStream` as well.

openinx (approved):
Looks good to me !

rajarshisarkar:
Thanks for the quick review, @openinx!

PR #4451, opened 2022-03-31
Spark: support stream-results in DeleteOrphanFiles action

microbearz:


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -172,17 +173,29 @@ private String jobDesc() {
...
 
+  private BaseDeleteOrphanFilesActionResult deleteFiles(Dataset<String> orphanDF) {
+    boolean streamResults = PropertyUtil.propertyAsBoolean(options(), STREAM_RESULTS, false);
+    Iterator<String> itr;
+    if (streamResults) {
+      itr = orphanDF.toLocalIterator();
+    } else {
+      itr = orphanDF.collectAsList().iterator();
+    }
+    List<String> orphanFiles = Lists.newArrayList();
```
singhpk234, line 189: [question] elements in the list is being added concurrently do we need to ensure thread safety ?
```suggestion
    List<String> orphanFiles = Collections.synchronizedList(Lists.newArrayList());
```
refering to this [SO Answer](https://stackoverflow.com/questions/2715983/concurrent-threads-adding-to-arraylist-at-same-time-what-happens)

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRemoveOrphanFilesProcedure.java
```diff
@@ -360,4 +360,29 @@ public void testRemoveOrphanFilesWithDeleteFiles() throws Exception {
...
+  @Test
+  public void testRemoveOrphanFilesWithStreamResultsEnabled() throws Exception {
+    sql("CREATE TABLE %s (id bigint NOT NULL) USING iceberg", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1)", tableName);
+    Dataset<Row> df = spark.sql(String.format("select * from %s", tableName));
+    Table table = Spark3Util.loadIcebergTable(spark, tableName);
+    df.write().parquet(table.location() + "/data/orphan_files");
+    Thread.sleep(1000);
+    Timestamp currentTimestamp = Timestamp.from(Instant.ofEpochMilli(System.currentTimeMillis()));
+    List<Object[]> output1 = sql(
```
felixYyu, line 373: Follow meaningful variable names，change `output1 `variable `dryrunOutput `instead, change `output2 `variable `streamResultsOutput `instead

microbearz: @felixYyu change `output1` -> `dryrunOutput `, `output2` -> `streamResultsOutput ` , also change `result1` -> `dryRunResult`, `result2` -> `streamResultsResult` in 'org.apache.iceberg.spark.actions.TestRemoveOrphanFilesAction#testRemoveOrphanFilesWithStreamResultsEnabled'

singhpk234 (approved):
Thanks @microbearz for the change, LGTM.

It would be really nice if you could update the doc (obviously in seperate PR) with this new param : https://iceberg.apache.org/docs/latest/spark-procedures/#remove_orphan_files

microbearz:
@singhpk234 thanks for your review. I will update the doc(https://iceberg.apache.org/docs/latest/spark-procedures/#remove_orphan_files) with a new PR

microbearz:
cc @aokolnychyi @rdblue

rdblue:
I'll try to look this weekend, but @RussellSpitzer knows this code better than I do and may be able to review more quickly.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRemoveOrphanFilesProcedure.java
```diff
@@ -360,4 +360,29 @@ public void testRemoveOrphanFilesWithDeleteFiles() throws Exception {
...
   }
+
+  @Test
+  public void testRemoveOrphanFilesWithStreamResultsEnabled() throws Exception {
+    sql("CREATE TABLE %s (id bigint NOT NULL) USING iceberg", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1)", tableName);
+    Dataset<Row> df = spark.sql(String.format("select * from %s", tableName));
+    Table table = Spark3Util.loadIcebergTable(spark, tableName);
+    df.write().parquet(table.location() + "/data/orphan_files");
+    Thread.sleep(1000);
```
RussellSpitzer, line 371: Is this necessary? We could always just put our currentTimestamp in the future

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRemoveOrphanFilesProcedure.java
```diff
@@ -360,4 +360,29 @@ public void testRemoveOrphanFilesWithDeleteFiles() throws Exception {
...
+    Timestamp currentTimestamp = Timestamp.from(Instant.ofEpochMilli(System.currentTimeMillis()));
+    List<Object[]> dryRunOutput = sql(
+            "CALL %s.system.remove_orphan_files(" +
+                    "table => '%s'," +
+                    "older_than => TIMESTAMP '%s'," +
+                    "dry_run => true)",
+            catalogName, tableIdent, currentTimestamp);
+    List<Object[]> streamResultsOutput = sql(
```
RussellSpitzer, line 379: There should probably be a test in here that we are actually streaming results.  See

https://github.com/apache/iceberg/blob/3a0cc769a2f5859a0f930a7d9a026c86cd52546a/spark/src/test/java/org/apache/iceberg/actions/TestExpireSnapshotsAction.java#L1058-L1060

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -172,17 +174,29 @@ private String jobDesc() {
...
 
+  private BaseDeleteOrphanFilesActionResult deleteFiles(Dataset<String> orphanDF) {
+    boolean streamResults = PropertyUtil.propertyAsBoolean(options(), STREAM_RESULTS, false);
+    Iterator<String> itr;
+    if (streamResults) {
+      itr = orphanDF.toLocalIterator();
+    } else {
+      itr = orphanDF.collectAsList().iterator();
+    }
+    List<String> orphanFiles = Collections.synchronizedList(Lists.newArrayList());
```
RussellSpitzer, line 190: Is there a benefit to streaming the results from the job but then collecting everything in memory anyway? Seems like we avoid materializing everything in memory only to materialize it all in memory ...

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RemoveOrphanFilesProcedure.java
```diff
@@ -80,12 +81,14 @@ public StructType outputType() {
   }
 
   @Override
+  @SuppressWarnings("CyclomaticComplexity")
   public InternalRow[] call(InternalRow args) {
     Identifier tableIdent = toIdentifier(args.getString(0), PARAMETERS[0].name());
     Long olderThanMillis = args.isNullAt(1) ? null : DateTimeUtil.microsToMillis(args.getLong(1));
     String location = args.isNullAt(2) ? null : args.getString(2);
-    boolean dryRun = args.isNullAt(3) ? false : args.getBoolean(3);
+    boolean dryRun = !args.isNullAt(3) && args.getBoolean(3);
```
RussellSpitzer, line 89: Why the change here?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RemoveOrphanFilesProcedure.java
```diff
@@ -80,12 +81,14 @@ public StructType outputType() {
...
     Integer maxConcurrentDeletes = args.isNullAt(4) ? null : args.getInt(4);
+    boolean streamResults = !args.isNullAt(5) && args.getBoolean(5);
```
RussellSpitzer, line 91: I would just match the pattern of the other settings here

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
microbearz: @RussellSpitzer thanks for your reminder. To avoid materializing job in memory, when `stream-results` enabled, we can streaming the results like:
```
if (streamResults) {
      return new BaseDeleteOrphanFilesActionResult(() -> orphanDF.toLocalIterator());
    } else {
      return new BaseDeleteOrphanFilesActionResult(orphanDF.collectAsList());
    }
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -172,18 +173,34 @@ private String jobDesc() {
...
+            .as(Encoders.STRING());
+    return deleteFiles(orphanDF);
+  }
+
+  private BaseDeleteOrphanFilesActionResult deleteFiles(Dataset<String> orphanDF) {
+    boolean streamResults = PropertyUtil.propertyAsBoolean(options(), STREAM_RESULTS, false);
+    Iterable<String> itr;
+    if (streamResults) {
+      // cache df to avoid recomputing
+      orphanDF.cache();
```
RussellSpitzer, line 186: Does this prevent redoing the shuffle? I thought (I could be wrong about this) that we would basically leave the job in the state where the join had written all shuffle files, then spark issues a new job for each reducer in the iterator. I'm not pro or con this move, although we probably will need to document this.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -172,18 +173,34 @@ private String jobDesc() {
...
+    }
+    Tasks.foreach(itr)
+            .noRetry()
+            .executeWith(deleteExecutorService)
+            .suppressFailureWhenFinished()
+            .onFailure((file, exc) -> LOG.warn("Failed to delete file: {}", file, exc))
+            .run(file -> {
+              deleteFunc.accept(file);
+            });
+    if (streamResults) {
```
RussellSpitzer, line 199: This may end up being very expensive since the code in the Procedure use this iterable a bunch of times. It also ends up putting everything in memory when it build the output rows ... InternalRow[] :/


One thing I think we should still be thinking about is moving all the delete operations to the executors. I know we had a little chat about this on Iceberg Slack and I think it's worth thinking about.

microbearz: I think so. For some tables in our cluster, we have to 'moving all the delete operations to the executors' and use streaming results to avoid oom and accelerate the job.

microbearz: @RussellSpitzer Or we can just return the size of deleted file list instead of the list of files when call the procedure

felixYyu: The RewriteDataFiles and RewriteManifestFiles Action are the stream-result option also required ？@RussellSpitzer

RussellSpitzer: @microbearz changing that api for that may be have to wait for a version change ... Your fix would change the situation for the action version. Although we probably can't cache like we are currently doing, I think this would leave the uncache for garbage collection of the object on the driver (if it gets uncached at all) which could be a very long time from the completion of the action depending on the use case.

aokolnychyi:
Is there a use case when the list of orphan files does not fit into the driver memory?

PR #4454, opened 2022-03-31
AWS: Make S3 client initialization thread safe in S3FileIO.

jfz:
FileIO can be shared by threads in the pool (ManifestGroup.planFiles - ParallelIterable), and those threads likely start at about the same time, race condition could happen.

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -72,7 +72,7 @@ public class S3FileIO implements FileIO, SupportsBulkOperations {
 
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
-  private transient S3Client client;
+  private transient volatile S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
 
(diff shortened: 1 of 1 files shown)


rdblue (approved):

rdblue:
This looks good to me. It may be simpler just to eagerly create the S3 client.

For @samarthjain's concern, I think it is okay. `initialize` is called just after creating the `FileIO` instance before it is shared across threads.

rdblue:
Thanks, @jfz!

jfz:
> 



> I have a few questions about this change: Should a FileIO object be shareable between multiple threads? I don't see any thread safety guarantees documented. If it cannot be shared, then ideally the fix should be in the usage of S3FileIO objects where we make sure a new instance is used within each thread.
> 
> If the S3FileIO is meant to be shareable between threads, then there are more thread safety concerns. For example, the below needs to be made thread safe too.
> 
> ```
> @Override
>   public void initialize(Map<String, String> properties) {
>     this.awsProperties = new AwsProperties(properties);
> 
>     // Do not override s3 client if it was provided
>     if (s3 == null) {
>       this.s3 = AwsClientFactories.from(properties)::s3;
>     }
> ```

I think the intend of shareable is to avoid creating a new S3Client every time and reduce over the wire transportation of the object. I do see some of the other stateful methods(executorService/close) are guarded already. 

You're right about initialize not guarded but I'd assume in most real cases that will be called once early at the lifecycle of this object.

PR #4456, opened 2022-03-31
Hive: Expose Snapshot Stats in HMS.

flyrain:
cc @aokolnychyi @RussellSpitzer @raptond @karuppayya @szehon-ho @kbendick

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -417,15 +417,15 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   }
 
   private void setSnapshotStats(TableMetadata metadata, Map<String, String> parameters) {
+    parameters.remove(TableProperties.CURRENT_SNAPSHOT_ID);
+    parameters.remove(TableProperties.CURRENT_SNAPSHOT_TIMESTAMP);
+    parameters.remove(TableProperties.CURRENT_SNAPSHOT_SUMMARY);
+
     Snapshot currentSnapshot = metadata.currentSnapshot();
     if (currentSnapshot != null) {
       parameters.put(TableProperties.CURRENT_SNAPSHOT_ID, String.valueOf(currentSnapshot.snapshotId()));
       parameters.put(TableProperties.CURRENT_SNAPSHOT_TIMESTAMP, String.valueOf(currentSnapshot.timestampMillis()));
       setSnapshotSummary(parameters, currentSnapshot);
-    } else {
-      parameters.remove(TableProperties.CURRENT_SNAPSHOT_ID);
-      parameters.remove(TableProperties.CURRENT_SNAPSHOT_TIMESTAMP);
-      parameters.remove(TableProperties.CURRENT_SNAPSHOT_SUMMARY);
     }
 
     parameters.put(TableProperties.SNAPSHOT_COUNT, String.valueOf(metadata.snapshots().size()));
@@ -438,12 +438,10 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
       if (summary.length() <= maxHiveTablePropertySize) {
         parameters.put(TableProperties.CURRENT_SNAPSHOT_SUMMARY, summary);
       } else {
-        parameters.remove(TableProperties.CURRENT_SNAPSHOT_SUMMARY);
         LOG.warn("Not exposing the current snapshot({}) summary in HMS since it exceeds {} characters",
             currentSnapshot.snapshotId(), maxHiveTablePropertySize);
       }
     } catch (JsonProcessingException e) {
-      parameters.remove(TableProperties.CURRENT_SNAPSHOT_SUMMARY);
       LOG.warn("Failed to convert current snapshot({}) summary to a json string", currentSnapshot.snapshotId(), e);
     }
   }
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -92,9 +92,8 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
   private static final String HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES = "iceberg.hive.metadata-refresh-max-retries";
   private static final String HIVE_TABLE_LEVEL_LOCK_EVICT_MS = "iceberg.hive.table-level-lock-evict-ms";
-  private static final String HIVE_LEGACY_TABLE_PARAMETER_SIZE = "iceberg.hive.legacy.table.parameter.size";
+  private static final String HIVE_TABLE_PARAMETER_SIZE_MAX = "iceberg.hive.table.parameter.size.max";
   private static final long HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT = 32672;
-  private static final long HIVE_TABLE_PARAMETER_SIZE_MAX_LEGACY = 4000;
   private static final long HIVE_ACQUIRE_LOCK_TIMEOUT_MS_DEFAULT = 3 * 60 * 1000; // 3 minutes
   private static final long HIVE_LOCK_CHECK_MIN_WAIT_MS_DEFAULT = 50; // 50 milliseconds
   private static final long HIVE_LOCK_CHECK_MAX_WAIT_MS_DEFAULT = 5 * 1000; // 5 seconds
@@ -176,8 +175,7 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
         conf.getInt(HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES, HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES_DEFAULT);
     long tableLevelLockCacheEvictionTimeout =
         conf.getLong(HIVE_TABLE_LEVEL_LOCK_EVICT_MS, HIVE_TABLE_LEVEL_LOCK_EVICT_MS_DEFAULT);
-    this.maxHiveTableParameterSize = conf.getBoolean(HIVE_LEGACY_TABLE_PARAMETER_SIZE, false) ?
-        HIVE_TABLE_PARAMETER_SIZE_MAX_LEGACY : HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT;
+    this.maxHiveTableParameterSize = conf.getLong(HIVE_TABLE_PARAMETER_SIZE_MAX, HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT);
     initTableLevelLockCache(tableLevelLockCacheEvictionTimeout);
   }
 
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -92,7 +92,9 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
   private static final String HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES = "iceberg.hive.metadata-refresh-max-retries";
   private static final String HIVE_TABLE_LEVEL_LOCK_EVICT_MS = "iceberg.hive.table-level-lock-evict-ms";
-  private static final long HIVE_TABLE_PROPERTY_VALUE_SIZE_MAX = 4000;
+  private static final String HIVE_LEGACY_TABLE_PARAMETER_SIZE = "iceberg.hive.legacy.table.parameter.size";
+  private static final long HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT = 32672;
+  private static final long HIVE_TABLE_PARAMETER_SIZE_MAX_LEGACY = 4000;
   private static final long HIVE_ACQUIRE_LOCK_TIMEOUT_MS_DEFAULT = 3 * 60 * 1000; // 3 minutes
   private static final long HIVE_LOCK_CHECK_MIN_WAIT_MS_DEFAULT = 50; // 50 milliseconds
   private static final long HIVE_LOCK_CHECK_MAX_WAIT_MS_DEFAULT = 5 * 1000; // 5 seconds
@@ -151,6 +153,7 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private final long lockAcquireTimeout;
   private final long lockCheckMinWaitTime;
   private final long lockCheckMaxWaitTime;
+  private final long maxHiveTableParameterSize;
   private final int metadataRefreshMaxRetries;
   private final FileIO fileIO;
   private final ClientPool<IMetaStoreClient, TException> metaClients;
@@ -173,6 +176,8 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
         conf.getInt(HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES, HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES_DEFAULT);
     long tableLevelLockCacheEvictionTimeout =
         conf.getLong(HIVE_TABLE_LEVEL_LOCK_EVICT_MS, HIVE_TABLE_LEVEL_LOCK_EVICT_MS_DEFAULT);
+    this.maxHiveTableParameterSize = conf.getBoolean(HIVE_LEGACY_TABLE_PARAMETER_SIZE, false) ?
+        HIVE_TABLE_PARAMETER_SIZE_MAX_LEGACY : HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT;
     initTableLevelLockCache(tableLevelLockCacheEvictionTimeout);
   }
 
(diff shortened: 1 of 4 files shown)


mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
```diff
@@ -624,7 +624,7 @@ public void testIcebergAndHmsTableProperties() throws Exception {
     Assert.assertEquals(expectedIcebergProperties, icebergTable.properties());
 
     if (Catalogs.hiveCatalog(shell.getHiveConf(), tableProperties)) {
-      Assert.assertEquals(11, hmsParams.size());
+      Assert.assertEquals(12, hmsParams.size());
```
anuragmantri, line 627: What is the one added parameter here?

anuragmantri (commented):
If I understand correctly, is the idea here to get snapshot information just by querying the catalog (in cases where we only have access to catalog)?

flyrain:
> If I understand correctly, is the idea here to get snapshot information just by querying the catalog (in cases where we only have access to catalog)?

That's correct.

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
flyrain: The snapshot count.

anuragmantri: Oh got it. For new tables, this is the only param that's getting added. Thanks!

anuragmantri (approved):
Thanks for change. It looks harmless to me. I will let others comment.

singhpk234 (commented):
Thanks @flyrain for this feature :) !!!

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -402,9 +404,27 @@ private void setHmsTableParameters(String newMetadataLocation, Table tbl, TableM
...
   }
 
+  private void setSnapshotStats(TableMetadata metadata, Map<String, String> parameters) {
+    Snapshot currentSnapshot = metadata.currentSnapshot();
+    if (currentSnapshot != null) {
+      parameters.put(TableProperties.CURRENT_SNAPSHOT_ID, String.valueOf(currentSnapshot.snapshotId()));
+      parameters.put(TableProperties.CURRENT_SNAPSHOT_TIMESTAMP, String.valueOf(currentSnapshot.timestampMillis()));
+      try {
+        parameters.put(TableProperties.CURRENT_SNAPSHOT_SUMMARY,
+            JsonUtil.mapper().writeValueAsString(currentSnapshot.summary()));
```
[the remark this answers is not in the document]
flyrain: Good question! According to https://issues.apache.org/jira/browse/HIVE-12274 and https://issues.apache.org/jira/browse/HIVE-14145, we can at least store 4000 chars as the values before HMS3.x, and more in HMS3.x depending on the under layer RDMS, at least 32672 bytes. I'm going to cap it at 4000 chars.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -402,9 +404,27 @@ private void setHmsTableParameters(String newMetadataLocation, Table tbl, TableM
...
+      } catch (JsonProcessingException e) {
+        LOG.warn("Failed to convert snapshot summary to a json string", e);
```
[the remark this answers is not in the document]
flyrain: Make sense, will print out snapshot id in the warn message.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -90,6 +92,7 @@
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
   private static final String HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES = "iceberg.hive.metadata-refresh-max-retries";
   private static final String HIVE_TABLE_LEVEL_LOCK_EVICT_MS = "iceberg.hive.table-level-lock-evict-ms";
+  private static final long HIVE_TABLE_PROPERTY_VALUE_SIZE_MAX = 4000;
```
szehon-ho, line 95: Should we should expose this configuration, otherwise it may not match what works for all users (different database and backends)?  I am not sure what others think.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -402,9 +405,32 @@ private void setHmsTableParameters(String newMetadataLocation, Table tbl, TableM
       parameters.put(StatsSetupConst.TOTAL_SIZE, summary.get(SnapshotSummary.TOTAL_FILE_SIZE_PROP));
     }
 
+    setSnapshotStats(metadata, parameters);
+
     tbl.setParameters(parameters);
   }
 
+  private void setSnapshotStats(TableMetadata metadata, Map<String, String> parameters) {
+    Snapshot currentSnapshot = metadata.currentSnapshot();
```
szehon-ho, line 414: I think we need a way to clear the properties if they are not there anymore in current snapshot (like original code setHmsProperties).  For example if the summary becomes suddenly over the length in new snapshot, then it will not override the old summary and would be wrong.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -402,9 +405,32 @@ private void setHmsTableParameters(String newMetadataLocation, Table tbl, TableM
...
+    if (currentSnapshot != null) {
+      parameters.put(TableProperties.CURRENT_SNAPSHOT_ID, String.valueOf(currentSnapshot.snapshotId()));
+      parameters.put(TableProperties.CURRENT_SNAPSHOT_TIMESTAMP, String.valueOf(currentSnapshot.timestampMillis()));
+      try {
+        String summary = JsonUtil.mapper().writeValueAsString(currentSnapshot.summary());
+        if (summary.length() <= HIVE_TABLE_PROPERTY_VALUE_SIZE_MAX) {
+          parameters.put(TableProperties.CURRENT_SNAPSHOT_SUMMARY, summary);
```
szehon-ho, line 421: Also had one concern, there's some repeating information (setHmsTableParameters puts some but not all of the summary).  Did we ever consider just inlining these fields from the summary like they did over there?

szehon-ho (commented):
Hi @flyrain I think it is useful for hive-based tooling, thanks for this, had some comments and questions.

Would be nice to get a look from @RussellSpitzer @aokolnychyi @pvary too.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: If it's a lot of properties, the json makes sense, just thinking its a shame there is a bit of repeat.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Ideally, it should be configurable. Not sure it's worth to do that at this moment. I'm open to any suggestion.

RussellSpitzer (approved):
I had a few minor comments, but @szehon-ho is the Hive expert so when he is +1 I think this is good to merge

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: I guess my comment is not complete, @flyrain's analysis in https://github.com/apache/iceberg/pull/4456#discussion_r840899312 is the exact one.  Older versions of Hive had the 4000 limit (due to need to support some older databases), and newer versions increased to 32672.  

My point was rather, if we don't have some kind of flag, we lose the full potential of current HMS versions just to support the older ones.

szehon-ho: Hm, yea I'm also not entirely sure what the right thing to do here is.  Yea it's hidden a bit deeply in the [package.jdo](https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/resources/package.jdo) file in Hive..

I see we do have some Hive specific configs in this file right (like iceberg.hive.lock-timeout-ms for example), and also ConfigProperties.java?  Could it be with that prefix (ie, iceberg.hive.summary.stats.max.size), with default value  and upper bound 32672?   Or if we don't want to give too much complexity to user, maybe  iceberg.hive.legacy.table.parameter.size which is just a boolean controlling whether its 32672 or 4000.

flyrain: Sounds good. The only downside is we cannot separate them when two hive catalogs are configured, but that's probably fine. `iceberg.hive.lock-timeout-ms` couldn't either.

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -468,4 +473,54 @@ public void testUUIDinTableProperties() throws Exception {
...
+          .withLocation(location)
+          .create();
+
+      String tableName = tableIdentifier.name();
+      org.apache.hadoop.hive.metastore.api.Table hmsTable =
+          metastoreClient.getTable(tableIdentifier.namespace().level(0), tableName);
+
+      // check whether parameters are in expected state
+      Map<String, String> parameters = hmsTable.getParameters();
+      Assert.assertEquals("0", parameters.get(TableProperties.SNAPSHOT_COUNT));
```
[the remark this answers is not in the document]
flyrain: I'm not sure it is necessary. The code here can explain itself quite well.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Fixed in the new commit.

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -468,4 +473,54 @@ public void testUUIDinTableProperties() throws Exception {
       catalog.dropTable(tableIdentifier);
     }
   }
+
+  @Test
+  public void testSnapshotStatsTableProperties() throws Exception {
```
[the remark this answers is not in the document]
flyrain: Add a test case for that

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -90,6 +92,9 @@
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
   private static final String HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES = "iceberg.hive.metadata-refresh-max-retries";
   private static final String HIVE_TABLE_LEVEL_LOCK_EVICT_MS = "iceberg.hive.table-level-lock-evict-ms";
+  private static final String HIVE_LEGACY_TABLE_PARAMETER_SIZE = "iceberg.hive.legacy.table.parameter.size";
```
flyrain, line 95: The current name doesn't look like a boolean type. I'm open to suggestion. `iceberg.hive.is.legacy.table.parameter.size`?

szehon-ho: hive.use.legacy.table.parameter.size?

flyrain:
Talked to @szehon-ho offline. We will give a default value for the HIVE_TABLE_PROPERTY_SIZE, while allowing user to configure it with different number due to a variety of HMS backends.
> Oracle DB's varchar2 supports 32767 bytes now, by setting the configuration parameter MAX_STRING_SIZE to EXTENDED. ([source](http://docs.oracle.com/database/121/SQLRF/sql_elements001.htm#SQLRF55623))
> Postgres supports a max of 1GB for character datatype. ([source](http://www.postgresql.org/docs/8.3/static/datatype-character.html))
> MySQL can support upto 65535 bytes for the entire row. So long as the PARAM_KEY value + PARAM_VALUE is less than 65535, we should be good. ([source](http://dev.mysql.com/doc/refman/5.0/en/char.html))
> SQL Server's varchar max length is 8000 and can go beyond using "varchar(max)" with the same limitation as MySQL being 65535 bytes for the entire row. ([source](http://dev.mysql.com/doc/refman/5.0/en/char.html))
> Derby's varchar can be upto 32672 bytes. ([source](https://db.apache.org/derby/docs/10.7/ref/rrefsqlj41207.html))

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -170,6 +175,7 @@ protected HiveTableOperations(Configuration conf, ClientPool metaClients, FileIO
         conf.getInt(HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES, HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES_DEFAULT);
     long tableLevelLockCacheEvictionTimeout =
         conf.getLong(HIVE_TABLE_LEVEL_LOCK_EVICT_MS, HIVE_TABLE_LEVEL_LOCK_EVICT_MS_DEFAULT);
+    this.maxHiveTableParameterSize = conf.getLong(HIVE_TABLE_PARAMETER_SIZE_MAX, HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT);
```
szehon-ho, line 178: Nit: can you move it up one line so not to break flow of the tableLevelLockCacheEvictionTimeout statement?

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -468,4 +478,83 @@ public void testUUIDinTableProperties() throws Exception {
...
+    Map<String, String> parameter = Maps.newHashMap();
+    spyOps.setSnapshotSummary(parameter, snapshot);
+    Assert.assertEquals("The snapshot summary must be in parameters", 1, parameter.size());
+
+    // create a snapshot summary whose json string size exceeds the limit
+    for (int i = 0; i < 1000; i++) {
+      summary.put(String.valueOf(i), "value");
+    }
+    long summarySize = JsonUtil.mapper().writeValueAsString(summary).length();
+    Assert.assertTrue(summarySize > 4000 && summarySize < 32672);
```
szehon-ho, line 556: Nit: i think no need for the upper limit check, right?  We are not really saving in Hive.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -90,6 +92,8 @@
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
   private static final String HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES = "iceberg.hive.metadata-refresh-max-retries";
   private static final String HIVE_TABLE_LEVEL_LOCK_EVICT_MS = "iceberg.hive.table-level-lock-evict-ms";
+  private static final String HIVE_TABLE_PARAMETER_SIZE_MAX = "iceberg.hive.table.parameter.size.max";
+  private static final long HIVE_TABLE_PARAMETER_SIZE_MAX_DEFAULT = 32672;
```
szehon-ho, line 96: Should we add a comment, giving some reference to when it was changed in Hive?

szehon-ho (commented):
@flyrain thanks for changes, looks mostly good, just a few more nits

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
flyrain: This is to make sure the new limit takes affect.
```
    conf.set("iceberg.hive.table.parameter.size.max", "4000");
```

flyrain: Add a comment in the code

szehon-ho: Hm, I meant, can't we just do 

```Assert.assertTrue("Summary size should be greater than limit", summarySize > 4000).  ```

To me the test is checking whether you can persist something beyond 4000 chars right?  The fact that it's below or above 32627 chars should not trigger now that we changed limit to 4000 right?

File: hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java (hunk 12 above)
Comment (szehon-ho) [reply hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java:98]: Thanks, nit: how about "iceberg.hive.max.table.parameter.size".

Also, regarding the comment, it's kind of based on backend but Hive had a global limit of 4000 regardless of database to allow it to support across different database, so maybe "For Hive versions below 2.3, table parameter values should be capped at 4000 characters, see https://issues.apache.org/jira/browse/HIVE-12274"?

File: hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java (hunk 11 above)
Comment (flyrain) [reply hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java:557]: The main purpose is to test whether we save the summary in HMS parameters when the size exceeds the limit. Besides, I also want to test if the limit has changed from 32627 to 4000. That's why I check both.
```
 Assert.assertTrue(summarySize > 4000 && summarySize < 32672);
```

File: hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java (hunk 12 above)
Comment (flyrain) [reply hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java:99]: Made the change

File: hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java (hunk 11 above)
Comment (szehon-ho) [reply hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java:557]: OK got it.

File: hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java (hunk 12 above)
Comment (szehon-ho) [reply hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java:99]: one more nit: can the variable match the changed string:  HIVE_MAX_TABLE_PARAMETER_SIZE

Comment (szehon-ho) [reply hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java:99]: Actually sorry about this , I just noticed other flags, weird how they do dash after "iceberg.hive".  How about 'iceberg.hive.table-property-max-size"

Comment (flyrain) [reply hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java]: Made the change.

**Comment by flyrain** [comment]
The failed test should not be related.
```
org.apache.iceberg.flink.TestFlinkUpsert > testPrimaryKeyFieldsAtEndOfTableSchema[catalogName=testhadoop, baseNamespace=default, format=PARQUET, isStreaming=true] FAILED
    java.lang.AssertionError: 
    Expecting:
      [+I[3, bbb, 2022-03-01], +I[1, aaa, 2022-03-01]]
    to contain exactly in any order:
      [+I[2, aaa, 2022-03-01], +I[3, bbb, 2022-03-01]]
    elements not found:
      [+I[2, aaa, 2022-03-01]]
    and elements not expected:
      [+I[1, aaa, 2022-03-01]]
        at org.apache.iceberg.flink.TestHelpers.assertRows(TestHelpers.java:137)
        at org.apache.iceberg.flink.TestFlinkUpsert.testPrimaryKeyFieldsAtEndOfTableSchema(TestFlinkUpsert.java:262)
```

File: hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java [hunk 13]
```diff
@@ -402,9 +411,43 @@ private void setHmsTableParameters(String newMetadataLocation, Table tbl, TableM
...
+  }
+
+  @VisibleForTesting
+  void setSnapshotSummary(Map<String, String> parameters, Snapshot currentSnapshot) {
+    try {
+      String summary = JsonUtil.mapper().writeValueAsString(currentSnapshot.summary());
+      if (summary.length() <= maxHiveTablePropertySize) {
+        parameters.put(TableProperties.CURRENT_SNAPSHOT_SUMMARY, summary);
+      } else {
+        parameters.remove(TableProperties.CURRENT_SNAPSHOT_SUMMARY);
```
szehon-ho, line 441: Optional: we can actually put the parameters.remove() for all the properties above the parameters.put(), here we are doing the remove three times.

szehon-ho (approved):
Looks good to me.  Just left one comment if you have time, otherwise it's fine.  Probably can restart the test too.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Good idea. Made the change.

szehon-ho (approved):

szehon-ho:
Merged, thanks @flyrain for the changes

flyrain:
Thanks @szehon-ho! Thanks all for the review, @singhpk234, @RussellSpitzer, @anuragmantri.

PR #1618, opened 2020-10-15
dynamically load custom FileIO implementation

jackye1995:
- refactor the class loader logic in `LocationProviders` to a util class
- add functionality to dynamically load FileIO in Hadoop and Hive table operations
- introduce a catalog property `io-impl`

PR #4459, opened 2022-04-01
Core: HadoopTables support dynamic FileIO

shidayang:
HadoopTables support dynamic FileIO

### DIFF
--- a/core/src/main/java/org/apache/iceberg/hadoop/HadoopTables.java
+++ b/core/src/main/java/org/apache/iceberg/hadoop/HadoopTables.java
@@ -44,6 +44,7 @@ import org.apache.iceberg.Transactions;
 import org.apache.iceberg.catalog.Catalog;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
+import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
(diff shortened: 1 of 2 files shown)


shidayang:
cc @jackye1995

core/src/main/java/org/apache/iceberg/hadoop/HadoopTables.java
```diff
@@ -68,12 +69,23 @@
 
   private Configuration conf;
 
+  private FileIO fileIO;
+
   public HadoopTables() {
     this(new Configuration());
   }
 
+  public HadoopTables(FileIO fileIO) {
```
openinx, line 78: Could you please explain why do we need the other `FileIO` rather than initializing the `HadoopFileIO` by default ?

shidayang: Thanks for your review, My FileSystem which has some external function such as authentication is not a standard implementation. In other cases, Many Engines has owner FileSystem Manager so they don't need HadoopTables to initialize FileSystem.

openinx: If the `FileSystem` is not a standard HDFS implementation, then I will think the `table` is not a stand hadoop table. So why not just define the customized catalog & table implementation ?

shidayang: In the one hand, The only difference is the FileIO. The constructor of HadoopTableOperations is protected and constructor args has FileIO, So the  design philosophy of HadoopTableOperations is support Dynamic FileIO. 

In other hand, like apache/iceberg#1618

PR #4461, opened 2022-04-01
Aliyun / GCP / DELL : make client initialization thread safe in all FileIOs

singhpk234:
recently a change went in which showed / fixed thread safety concern with S3FileIO. This made me check other FileIO looks like this handling is missing here as well.
- https://github.com/apache/iceberg/pull/4454 

- As per my understanding the FileIO obj when serialized and sent to the executors (via broadcast) , it will be shared by all tasks within the executor, since the client is marked `transient` it will be re-initialized when it reaches executor and now the client initialization can be triggered by any of task threads hence raising the possibility of race condition.

This change attempts to fix all remaining FileIO's as well.

--- 

cc @jfz @rdblue

### DIFF
--- a/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSFileIO.java
+++ b/aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSFileIO.java
@@ -47,7 +47,7 @@ public class OSSFileIO implements FileIO {
 
   private SerializableSupplier<OSS> oss;
   private AliyunProperties aliyunProperties;
-  private transient OSS client;
+  private transient volatile OSS client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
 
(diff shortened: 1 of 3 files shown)


jfz:
Looks good to me.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSFileIO.java
```diff
@@ -89,7 +89,11 @@ public void deleteFile(String path) {
 
   private OSS client() {
     if (client == null) {
-      client = oss.get();
+      synchronized (this) {
+        if (client == null) {
+          client = oss.get();
+        }
+      }
```
rdblue, line 96: In the other PR, we were wondering whether we should just initialize the client eagerly in `initialize` What do you think about that vs. lazy initialization?

singhpk234: As per my understanding the since s3 client is not serializable and initialize being called in only at driver when FileIO is initialized via CatlogUtil, we capture the s3 client against the wrapper `SerializableSupplier` and this will now be part of serialized FileIO obj (which will be used when it's broadcasted). Now in the executor the task threads starts calling the FileIO func directly, without calling intialize etc.

IIUC, it might not be possible to do it in initialize which is called at driver, Am I missing something here ?

rdblue: Ah, you're right. This is to handle serialization. Good catch.

PR #4463, opened 2022-04-01
Docs: update Spark Write doc for partitioned tables

polaris6:
I am using spark sql to write data, the amount of data is very large, so I don't want to sort the data globally, but I didn't find an example of partial sorting in the documentation, I want to add it here, thanks!

Sort by definition in spark sql documentation:
https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sortby.html
image

The test:
The first step is to create a table:
wecom-temp-0f83c792cf74f323d1f2dc9431255605

Then query the data by sort by:
wecom-temp-7a6e180f1068256a1eb80b87156a8bf6

### DIFF
--- a/docs/spark/spark-writes.md
+++ b/docs/spark/spark-writes.md
@@ -313,7 +313,7 @@ distribution & sort order to Spark.
 {{< hint info >}}
 Both global sort (sorting all the data in the write) and local sort (sorting the data within a Spark task) can be used to write against a partitioned table.
 
-In SQL, the [`ORDER BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-orderby.html) will achieve global sorting and [`SORT BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sortby.html) will achieve local sorting.
+In SQL, [`ORDER BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-orderby.html) will achieve global sorting and [`SORT BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sortby.html) will achieve local sorting.
 
 In the Dataframe API, the `orderBy`/`sort` functions will achieve global sorting and `sortWithinPartitions` will achieve local sorting.
 {{< /hint >}}
--- a/docs/spark/spark-writes.md
+++ b/docs/spark/spark-writes.md
@@ -311,7 +311,7 @@ distribution & sort order to Spark.
 {{< /hint >}}
 
 {{< hint info >}}
-Both global sort and local sort (data sorted within each partition) can be used to write against partitioned table.
+Both global sort (sorting all the data in the write) and local sort (sorting the data within a Spark task) can be used to write against partitioned table.
 
 In SQL, the [`ORDER BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-orderby.html) will achieve global sorting and [`SORT BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sortby.html) will achieve local sorting.
 
--- a/docs/spark/spark-writes.md
+++ b/docs/spark/spark-writes.md
@@ -311,7 +311,11 @@ distribution & sort order to Spark.
 {{< /hint >}}
 
 {{< hint info >}}
-Both global sort (`orderBy`/`sort`) and local sort (`sortWithinPartitions`) work for the requirement.
+Both global sort and local sort (data sorted within each partition) can be used to write against partitioned table.
+
+In SQL, the [`ORDER BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-orderby.html) will achieve global sorting and [`SORT BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sortby.html) will achieve local sorting.
+
+In the Dataframe API, the `orderBy`/`sort` functions will achieve global sorting and `sortWithinPartitions` will achieve local sorting.
 {{< /hint >}}
 
 Let's go through writing the data against below sample table:
@@ -326,9 +330,9 @@ USING iceberg
 PARTITIONED BY (days(ts), category)
 ```
 
-To write data to the sample table, your data needs to be sorted by `days(ts), category`.
+#### In Spark SQL
 
-If you're inserting data with SQL statement, you can use `ORDER BY` to achieve it, like below:
+To globally sort data based on `ts` and `category`:
 
 ```sql
 INSERT INTO prod.db.sample
@@ -336,8 +340,39 @@ SELECT id, data, category, ts FROM another_table
 ORDER BY ts, category
 ```
 
-If you're inserting data with DataFrame, you can use either `orderBy`/`sort` to trigger global sort, or `sortWithinPartitions`
-to trigger local sort. Local sort for example:
+To locally sort data based on `ts` and `category`:
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY ts, category
+```
+
+`SORT BY` clauses can also be used with partition transforms. The [date-and-timestamp-functions](https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html#date-and-timestamp-functions) will be used when partition transforms is time related; truncate related functions such as `substr` will be used when partition transform is `truncate[W]`; [define and register UDFs](https://spark.apache.org/docs/latest/sql-ref-functions-udf-scalar.html) can also take effect.
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY day(ts), category
+```
+
+#### In the Dataframe API
+
+To globally sort data based on `ts` and `category`:
+
+```scala
+data.sort("ts", "category")
+    .writeTo("prod.db.sample")
+    .append()
+```
+
+```scala
+data.orderBy("ts", "category")
+    .writeTo("prod.db.sample")
+    .append()
+```
+
+To locally sort data based on `ts` and `category`:
 
 ```scala
 data.sortWithinPartitions("ts", "category")
@@ -345,9 +380,11 @@ data.sortWithinPartitions("ts", "category")
     .append()
 ```
 
-You can simply add the original column to the sort condition for the most partition transformations, except `bucket`.
+#### Bucket Transform
 
-For `bucket` partition transformation, you need to register the Iceberg transform function in Spark to specify it during sort.
+It's easy to add the original column to the sort condition for the most partition transformations, except `bucket`.
+
+For `bucket` partition transformation, it is required to register the Iceberg transform function in Spark to specify it during sort.
 
 Let's go through another sample table having bucket partition:
 
@@ -361,7 +398,7 @@ USING iceberg
 PARTITIONED BY (bucket(16, id))
 ```
 
-You need to register the function to deal with bucket, like below:
+Register the function to deal with bucket, like below:
 
 ```scala
 import org.apache.iceberg.spark.IcebergSpark
@@ -376,9 +413,9 @@ Explicit registration of the function is necessary because Spark doesn't allow I
 which can be used in query.
 {{< /hint >}}
 
-Here we just registered the bucket function as `iceberg_bucket16`, which can be used in sort clause.
+Here the bucket function is registered as `iceberg_bucket16`, which can be used in sort clause.
 
-If you're inserting data with SQL statement, you can use the function like below:
+In Spark SQL, the function can be used as follows:
 
 ```sql
 INSERT INTO prod.db.sample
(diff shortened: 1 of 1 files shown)


polaris6:
cc @samredai @jackye1995

docs/spark/spark-writes.md
```diff
@@ -311,7 +311,9 @@ distribution & sort order to Spark.
 {{< /hint >}}
 
 {{< hint info >}}
-Both global sort (`orderBy`/`sort`) and local sort (`sortWithinPartitions`) work for the requirement.
+To write with SQL, both global sort (`ORDER BY`) and local sort (`SORT BY`) work for the requirement.
```
samredai, line 314: "the requirement" was always a bit confusing to me here. What do you think about this re-wording?

> With SQL, you can use `ORDER BY` to achieve global sorting or `SORT BY` to achieve local sorting.
> 
> With dataframes, you can use `orderBy`/`sort` to achieve global sorting or `sortWithinPartitions` to achieve local sorting.

docs/spark/spark-writes.md
```diff
@@ -328,14 +330,30 @@ PARTITIONED BY (days(ts), category)
...
 
+Local sort like below:
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY ts, category
+```
+
+Of course, you can use `SORT BY` with partition transforms:
```
samredai, line 349:
```suggestion
Of course, you can also use `SORT BY` with partition transforms.
```

docs/spark/spark-writes.md
samredai, line 314: "the requirement" was always a bit confusing to me here. What do you think about my suggestions in this section?

samredai:
Thanks for adding this @polaris6 ! I left a couple of suggestions, let me know what you think 👍

docs/spark/spark-writes.md
polaris6: yes i think so too.  The revised sentence is better.

samredai (approved):
LGTM

polaris6:
@openinx There is another pr here, please help me to review it, thanks!

docs/spark/spark-writes.md
```diff
@@ -311,7 +311,9 @@ distribution & sort order to Spark.
 {{< /hint >}}
 
 {{< hint info >}}
-Both global sort (`orderBy`/`sort`) and local sort (`sortWithinPartitions`) work for the requirement.
+With SQL, you can use `ORDER BY` to achieve global sorting or `SORT BY` to achieve local sorting.
```
openinx, line 314: Could you please add the `ORDER BY` and `SORT BY` links from apache spark pages ?

docs/spark/spark-writes.md
```diff
@@ -328,14 +330,30 @@ PARTITIONED BY (days(ts), category)
...
+SELECT id, data, category, ts FROM another_table
+SORT BY ts, category
+```
+
+Of course, you can also use `SORT BY` with partition transforms.
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY days(ts), category
```
openinx, line 354: Should we register the `days(..)` UDF into spark environment firstly before we call it ?  I will suggest to add the SQL about how to register the `days(..)` UDF.

docs/spark/spark-writes.md
polaris6: Of course, I updated the commit, please help to review it again, thanks!

docs/spark/spark-writes.md
polaris6: I perfected the description, what do you think of this description?

polaris6:
@openinx Hi openinx, excuse me, can you help me to review this pr again? thank you very much!

polaris6:
@rdblue Hi Ryan Blue, excuse me, can you help me to review this pr? thank you very much!

polaris6:
Hello @openinx @rdblue @RussellSpitzer, I’m sorry to bother you, this PR has been around for a long time, could somebody please help me to review and merge it, thank you very much!

RussellSpitzer:
I think there is a misconception here between data layout sort orders and the ordering of the data in a query. The orders documented by iceberg dictate how the data is physically stored in files on the disc and happens based on distribution and ordering requirements in the DSV2 api. The functions mentioned here and their sql only effect Spark on read and cause shuffles to order output for the reader but do not effect the on disc layout.

The file based ordering is to make pruning of files during scan planning more efficient while the ordering during read is to satisfy user query requirements.

RussellSpitzer:
I believe your example may show a bug in the way distribution and ordering is being presented to query engine ... but that is separate from this doc changes noted here

RussellSpitzer:
Ah never mind I think I misred this wrong this is about manually imposing sort orders for writes without using table sort orders. I think for that this is correct

docs/spark/spark-writes.md
```diff
@@ -311,7 +311,9 @@ distribution & sort order to Spark.
 {{< /hint >}}
 
 {{< hint info >}}
-Both global sort (`orderBy`/`sort`) and local sort (`sortWithinPartitions`) work for the requirement.
```
RussellSpitzer, line 314: I think for this we should note above that this is not required for recent spark releases and the information below will come in conflict with a built in table Sort Order or distribution. (it should not override a built in ordering)

docs/spark/spark-writes.md
```diff
@@ -311,7 +311,9 @@ distribution & sort order to Spark.
...
+With SQL, you can use [`ORDER BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-orderby.html) to achieve global sorting or [`SORT BY`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-sortby.html) to achieve local sorting.
```
RussellSpitzer, line 314: Global and Local sorting need to be defined, I don't think we have it in this part of the doc but I believe we could just refer to it in the distribution section?

RussellSpitzer: We are trying to refrain from using pronouns in the documentation

"With SQL, you can use" -> "In SQL the keyword [`ORDER BY`] ..... "

docs/spark/spark-writes.md
```diff
@@ -328,14 +330,32 @@ PARTITIONED BY (days(ts), category)
 
 To write data to the sample table, your data needs to be sorted by `days(ts), category`.
 
-If you're inserting data with SQL statement, you can use `ORDER BY` to achieve it, like below:
+If you're inserting data with SQL statement, you can use either `ORDER BY` to trigger global sort, or `SORT BY` to trigger local sort. Global sort like below:
```
RussellSpitzer, line 333: Since we are changing this line let's drop the pronouns. 

I think the examples may be clearer as well if we just name them explicitly

"To globally sort data based on ts and category
Example
"

docs/spark/spark-writes.md
```diff
@@ -328,14 +330,32 @@ PARTITIONED BY (days(ts), category)
...
 
+Local sort like below:
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY ts, category
+```
+
+Of course, you can also use `SORT BY` with partition transformations.
```
RussellSpitzer, line 349: Pronouns here
"Of Course" I think is a bit colloquial and doesn't add much here

Instead 
"SORT BY clauses can also be used with partition transforms" 

I think this also probably should be accompanied by an example ^ since I don't think a novice reader would understand quite what this means.

docs/spark/spark-writes.md
```diff
@@ -328,14 +330,32 @@ PARTITIONED BY (days(ts), category)
...
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY ts, category
+```
+
+Of course, you can also use `SORT BY` with partition transformations. 
+
+You can use the [date-and-timestamp-functions](https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html#date-and-timestamp-functions) in Spark SQL when your partition transformation is time related, or use truncate related functions such as substr in Spark SQL when your partition transformation is `truncate[W]`, you can also [define and register UDFs](https://spark.apache.org/docs/latest/sql-ref-functions-udf-scalar.html) and invoke them in Spark SQL.
```
RussellSpitzer, line 351: Pronouns again

docs/spark/spark-writes.md
polaris6: Sorry I don't know much about this part yet, I may need to research it first.

RussellSpitzer: That's fine we probably can do a separate doc to actually explain the Distribution/Ordering built in's if we haven't yet

docs/spark/spark-writes.md
```diff
@@ -311,7 +311,11 @@ distribution & sort order to Spark.
 {{< /hint >}}
 
 {{< hint info >}}
-Both global sort (`orderBy`/`sort`) and local sort (`sortWithinPartitions`) work for the requirement.
+Both global sort and local sort (data sorted within each partition) can be used to write against partitioned table.
```
RussellSpitzer, line 314: Do we have global or local defined anywhere? I still think we should have something like
"global sort (sorting all the data in the write) and local sort (sorting the data within a Spark task)"

polaris6: I've simply changed the description, perhaps we can go into more detail in this separate doc:
https://github.com/apache/iceberg/pull/4463#discussion_r856268809

docs/spark/spark-writes.md
polaris6: Yes, that sounds great.

polaris6: Yes, that sounds great.

polaris6:
hi @RussellSpitzer, could you please help me to merge this PR? Thank you very much!
We can add a separate doc in another PR to explain the Distribution/Ordering and global sort/local sort, l am glad to do this if no one else does it, but I need to take the time to research it first.

docs/spark/spark-writes.md
```diff
@@ -326,28 +330,61 @@ USING iceberg
...
-to trigger local sort. Local sort for example:
+To locally sort data based on `ts` and `category`:
+
+```sql
+INSERT INTO prod.db.sample
+SELECT id, data, category, ts FROM another_table
+SORT BY ts, category
+```
+
+`SORT BY` clauses can also be used with partition transforms. The [date-and-timestamp-functions](https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html#date-and-timestamp-functions) will be used when partition transforms is time related; truncate related functions such as `substr` will be used when partition transform is `truncate[W]`; [define and register UDFs](https://spark.apache.org/docs/latest/sql-ref-functions-udf-scalar.html) can also take effect.
```
RussellSpitzer, line 351: This paragraph has a few issues,
`will` -> `should` ? There are multiple "will be used"  when I believe the intent is "can be used" or "should be used"

partition transforms is time related -> are time related // Plural `transforms` needs "are"

related functions such as substr _should_ be used when _the_ partition // "the" before partition and "should" instead of will

The last sentence "define and register udfs can also take effect" doesn't make sense to me in the context of the other sentences. What does this mean in the context of sort by clauses?

Maggie, my resident english expert, also tells me the semi-colons are correct but you could use periods if you liked as well.

docs/spark/spark-writes.md
```diff
@@ -326,28 +330,61 @@ USING iceberg
...
 ```scala
 data.sortWithinPartitions("ts", "category")
     .writeTo("prod.db.sample")
     .append()
 ```
 
-You can simply add the original column to the sort condition for the most partition transformations, except `bucket`.
+#### Bucket Transform
+
+It's easy to add the original column to the sort condition for the most partition transformations, except `bucket`.
```
RussellSpitzer, line 385: This is a personal thing, but I dislike the use of words like "easy" or "simple" in documentation. Computers are hard and I generally want folks to not feel like they don't understand something that is labeled as "easy" and to avoid this I try to never use "easy" or "simple". It also doesn't add much to this documentation.

For this sentence I would probably go with something like
`bucket` transforms require additional steps which are not necessary with other partition transforms.

If we want to keep this sentence then we should correct the grammar,
"It's easy to add the original column to the sort condition for ~the~ most partition transforms, except bucket"

RussellSpitzer (commented):
left more grammar cleanup comments

polaris6:
Hi @RussellSpitzer, thank you very much for your suggestion, it's very helpful! I resubmitted the commit.

docs/spark/spark-writes.md
```diff
@@ -326,28 +330,61 @@ USING iceberg
 PARTITIONED BY (days(ts), category)
 ```
 
-To write data to the sample table, your data needs to be sorted by `days(ts), category`.
+#### In Spark SQL
```
samredai, line 333:
```suggestion
#### Sort Order Using Spark SQL
```

docs/spark/spark-writes.md
```diff
@@ -376,17 +413,17 @@ Explicit registration of the function is necessary because Spark doesn't allow I
 which can be used in query.
 {{< /hint >}}
 
-Here we just registered the bucket function as `iceberg_bucket16`, which can be used in sort clause.
+Here the bucket function is registered as `iceberg_bucket16`, which can be used in sort clause.
```
samredai, line 416:
```suggestion
Here the bucket function is registered as `iceberg_bucket16`, which can be used in a sort clause.
```

polaris6:
Hi @samredai, thank you very much for your suggestion, it's very helpful, I really appreciate it!

PR #4464, opened 2022-04-01
Core: Fix Conflicting transactions are ignored during commit

ajantha-bhat:
Currently when the conflicting transaction is present, it is silently ignored.

scenario:
a. create table
b. create transaction (txn-1) of add column on that table (don't commit the txn)
c. create another txn (txn-2) of add column and commit it.
d. commit the txn-1

txn-1 is expected to fail because it is conflicting schema id. 
Currently the commit of txn-1 is success. But the table is not containing the txn-1 schema changes.
Instead this txn-1 should fail (even after retry) and throw exception.


core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -368,14 +368,7 @@ private void commitSimpleTransaction() {
               2.0 /* exponential */)
           .onlyRetryOn(CommitFailedException.class)
           .run(underlyingOps -> {
-            if (base != underlyingOps.refresh()) {
-              this.base = underlyingOps.current(); // just refreshed
-              this.current = base;
-              for (PendingUpdate update : updates) {
```
ajantha-bhat, line 374: Initially thought of moving the update outside the if check, by direct write test case (`testTransactionRewriteManifestsAppendedDirectly`) will fail in that case.

Hence, falling back to the base metadata when the commit is failed. So, that updates will be added again.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -437,6 +430,25 @@ private void commitSimpleTransaction() {
     }
   }
 
+  private void applyUpdates(TableOperations underlyingOps) {
```
ajantha-bhat, line 433: moved to a separate function, just to fix checkstyle complexity warning.

core/src/test/java/org/apache/iceberg/TestTransaction.java
```diff
@@ -420,6 +421,35 @@ public void testMultipleUpdateTransactionRetryMergeCleanup() {
     Assert.assertFalse("Append manifest should be deleted", new File(appendManifest.path()).exists());
   }
 
+  @Test
+  public void testTransactionRetrySchemaUpdate() {
```
ajantha-bhat, line 425: this test case fails without the fix.

ajantha-bhat:
cc: @RussellSpitzer , @rdblue

core/src/test/java/org/apache/iceberg/TestTransaction.java
```diff
@@ -420,6 +421,35 @@ public void testMultipleUpdateTransactionRetryMergeCleanup() {
...
+    // use only one retry
+    table.updateProperties()
+            .set(TableProperties.COMMIT_NUM_RETRIES, "1")
+            .commit();
```
rdblue, line 429: Please fix indentation.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -437,6 +430,25 @@ private void commitSimpleTransaction() {
...
+      // use refreshed the metadata
+      this.base = underlyingOps.current();
+      this.current = underlyingOps.current();
+      for (PendingUpdate update : updates) {
+        // re-commit each update in the chain to apply it and update current
+        try {
+          update.commit();
+        } catch (CommitFailedException e) {
+          // fallback to old base, so that it refreshes again on retry and apply pending updates.
+          base = oldBase;
```
rdblue, line 445: @ajantha-bhat, if I understand correctly, this is happening:
1. The transaction fails to commit because of a concurrent update
2. The base is updated to the refreshed metadata to re-apply commits, but one commit fails
3. The retry runs again but detects that base is the same as the refreshed metadata, so the updates are not reapplied
4. Because the updates aren't reapplied, `commit` is called with the refreshed base and the last inner commit to succeed

Is that right?

In that case, I think that the right thing to do is to avoid catching the update's exception and throw it instead of retrying. If an update cannot be re-committed we should fail immediately rather than resetting to the original metadata to exhaust retries.

That is what we do in the REST catalog `CatalogHandlers`: https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java#L289-L294

I think that's a more correct fix.

rdblue (changes requested):
I think the test case is good, but I would fix this slightly differently.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
ajantha-bhat: >  The transaction fails to commit because of a concurrent update
The base is updated to the refreshed metadata to re-apply commits, but one commit fails
The retry runs again but detects that base is the same as the refreshed metadata, so the updates are not reapplied
Because the updates aren't reapplied, commit is called with the refreshed base and the last inner commit to succeed

> Is that right?

Yes, exactly that is what happening.

> In that case, I think that the right thing to do is to avoid catching the update's exception and throw it instead of retrying. If an update cannot be re-committed we should fail immediately rather than resetting to the original metadata to exhaust retries.

Thanks for the suggestion. But I am wondering what is the point of retry configuration?
In case of single update, your suggestion holds good. But I think it will impact multiple updates.

Say I have 3 updates to apply, first update is success and the second update is failed because the first update is success. But doing a retry with rebase may help in second commit to get success right?  

There are some operations like concurrent append, which can get success on retry and some operations like concurrent schema update which cannot even success after retry. So, we may need to retry when more than one update is present and it fails right?

WDYT?

ajantha-bhat: My main worry is we are totally ignoring the retry configuration for all the scenario (as it is a common code) if I break out of the retry loop.
So, when will the retry configuration will be used then?
Let me think more on this. I am bit new to this side of the code.

ajantha-bhat: got it. `CommitFailedException` from `TableOperations.commit()` should be retired but not from the `PendingUpdate.commit()` as pending updates cannot successful even after the retry.

rdblue: Yes, exactly. So you can do the same thing as the code I referenced above.

ajantha-bhat:
> I think the test case is good, but I would fix this slightly differently.

@rdblue: Thanks for the review. Fixed as per suggestion.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -404,6 +397,9 @@ private void commitSimpleTransaction() {
           .onFailure((file, exc) -> LOG.warn("Failed to delete uncommitted file: {}", file, exc))
           .run(ops.io()::deleteFile);
 
+      if (e instanceof ValidationFailureException) {
```
rdblue, line 400: Looks like this is here because you want to share the cleanup code, but I think it would be cleaner to move that to a method that gets called from separate `catch` blocks.

ajantha-bhat: yeah. It is cleaner that way. Fixed now.

ajantha-bhat:
@rdblue : Can you please help in merging this? Thanks.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -728,4 +744,20 @@ TableOperations ops() {
   Set<String> deletedFiles() {
     return deletedFiles;
   }
+
+  /**
+   * Exception used to avoid retrying {@link PendingUpdate} when it is failed with {@link CommitFailedException}.
+   */
+  private static class ValidationFailureException extends RuntimeException {
```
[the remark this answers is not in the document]
ajantha-bhat: `ValidationFailureException` in `CatalogHandlers` is more about Illegal arguments. So, it is validation failure exception.

Here it is more of failed to apply pending update. Even though the purpose is to abort, they can have different explanations based on the context.

Hence renamed `ValidationFailureException` to `PendingUpdateFailedException` in BaseTransaction instead of making common code.

ajantha-bhat:
ping @rdblue ..

core/src/main/java/org/apache/iceberg/BaseTransaction.java
rdblue: Yeah, reapplying updates needs to happen after refreshing the underlying metadata on each attempt.

core/src/main/java/org/apache/iceberg/BaseTransaction.java
```diff
@@ -388,22 +381,11 @@ private void commitSimpleTransaction() {
     } catch (CommitStateUnknownException e) {
       throw e;
 
+    } catch (PendingUpdateFailedException e) {
+      cleanUpOnCommitFailure();
+      throw e.wrapped();
```
rdblue, line 386: Looks like other catch blocks have extra whitespace. Can you match the style?

rdblue (approved):

rdblue:
Thanks, @ajantha-bhat! Nice work.

PR #4465, opened 2022-04-01
Add FileIO implementation for Azure Blob Storage

### LINKED ISSUE #4257: Implement FileIO for Azure

sumeetgajjar:
### What changes were proposed in this pull request?

Currently, HadoopFileIO is used to talk to azure blob storage. This PR introduces AzureFileIO which uses Azure native SDK to communicate with Azure blob storage.

### Does this PR introduce _any_ user-facing change?
Yes, users can now configure the catalog `io-impl` property to `org.apache.iceberg.azure.blob.AzureBlobFileIO` to enable this feature.

### How was this patch tested?

- Ran the newly added tests against [Azurite Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio)
- Ran the newly added tests against Azure Blob Storage
- Used newly added AzureFileIO to run queries on iceberg tables stored on azure using Spark

Closes #4257


sumeetgajjar:
cc: @danielcweeks @jackye1995

.baseline/checkstyle/checkstyle.xml
```diff
@@ -122,7 +122,8 @@
                 org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.*,
                 org.apache.spark.sql.functions.*,
                 org.apache.spark.sql.connector.iceberg.write.RowLevelOperation.Command.*,
-                org.junit.Assert.*"/>
+                org.junit.Assert.*,
+                org.assertj.core.api.Assertions.*"/>
```
rdblue, line 126: I don't think we need to change this.

build.gradle
```diff
@@ -606,6 +613,101 @@ project(':iceberg-dell') {
...
 }
 
+project(':iceberg-azure') {
+  apply plugin: "com.bmuschko.docker-remote-api"
+  apply plugin: "com.github.johnrengelman.shadow"
+
+  dependencies {
+    api project(':iceberg-api')
+    implementation project(path: ':iceberg-bundled-guava', configuration: 'shadow')
+    implementation platform('com.azure:azure-sdk-bom')
```
rdblue, line 623: I would rather track dependency versions specifically rather than relying on some external BOM. BOMs are fine for end uses, but libraries should generally not delegate dependency versions to other projects.

build.gradle
```diff
@@ -38,6 +42,9 @@ buildscript {
 
 plugins {
   id 'nebula.dependency-recommender' version '9.0.2'
+  // Since 7.x gradle-docker-plugin is compiled using JDK11, thus using the latest version will fail Java8 builds
+  // https://bmuschko.github.io/gradle-docker-plugin/current/user-guide/#change_log
```
rdblue, line 46: This comment doesn't make it clear why this plugin is used. Can you explain in more detail?

azure/src/main/java/org/apache/iceberg/azure/blob/AzureURI.java
```diff
@@ -0,0 +1,151 @@
...
+  private final String storageAccount;
+  private final String path;
+
+  private AzureURI(String location) {
+    Preconditions.checkNotNull(location, "Location cannot be null.");
+    final URI uri;
+    try {
+      uri = new URI(location);
+    } catch (URISyntaxException e) {
+      throw new ValidationException("Invalid Azure URI: %s.", location);
```
rdblue, line 57: There's no need for end punctuation in logs, and in fact in cases like this it is misleading because `.` could be interpreted as part of the URI. Can you remove end punctuation?

azure/src/main/java/org/apache/iceberg/azure/blob/AzureURI.java
```diff
@@ -0,0 +1,151 @@
...
+  private final String location;
+  private final String container;
+  private final String storageAccount;
+  private final String path;
+
+  private AzureURI(String location) {
+    Preconditions.checkNotNull(location, "Location cannot be null.");
+    final URI uri;
+    try {
+      uri = new URI(location);
```
rdblue, line 55: We generally discourage the use of `URI` because it handles URI encoding in strange ways. I think it is a best practice to ignore the URI and parse manually using `split` and delimiters.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureURI.java
```diff
@@ -0,0 +1,151 @@
...
+    } catch (URISyntaxException e) {
+      throw new ValidationException("Invalid Azure URI: %s.", location);
+    }
+    this.location = location;
+
+    ValidationException.check(
```
rdblue, line 61: `ValidationException` is not a substitute for `IllegalArgumentException`. `ValidationException` indicates that while an argument may be valid, it is inconsistent with other arguments or config.

For example, when creating a partition spec, `"column"` is a valid column reference, but `null` would result in an `IllegalArgumentException`. But, if `"column"` is not a defined name in the schema then a `ValidationException` is thrown because you can't partition by an unknown column.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
```diff
@@ -0,0 +1,86 @@
...
+import com.azure.storage.blob.BlobClientBuilder;
+import com.azure.storage.common.StorageSharedKeyCredential;
+import java.util.Optional;
+import org.apache.iceberg.azure.AuthType;
+import org.apache.iceberg.azure.AzureProperties;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AzureBlobClientFactory {
```
rdblue, line 33: Should this be package-private? I don't see a reason why people would need to use it directly.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(AzureBlobClientFactory.class);
+
+  private AzureBlobClientFactory() {
+  }
+
+  public static BlobClient createBlobClient(AzureURI azureURI, AzureProperties azureProperties) {
+    final String storageAccount = azureURI.storageAccount();
+    final BlobClientBuilder builder = new BlobClientBuilder();
```
rdblue, line 42: In Iceberg, we don't use `final` for local variables. Recent JVM versions (8+) handle this without problems and it is also not very valuable because it isn't actually translated into bytecode.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
```diff
@@ -0,0 +1,86 @@
...
+    if (connectionString.isPresent()) {
+      LOG.debug("Using azure storage connection string to build the blob client for {}", storageAccount);
+      Preconditions.checkArgument(!connectionString.get().isEmpty(), "Connection string cannot be empty.");
+      builder.connectionString(connectionString.get());
+    } else {
+      final String endpoint = storageEndpoint(storageAccount, azureProperties);
+      LOG.debug("Using {} endpoint for {}", endpoint, storageAccount);
+      builder.endpoint(endpoint);
+      final AuthType authType = azureProperties.authType(storageAccount);
+      setAuth(storageAccount, authType, azureProperties, builder);
```
rdblue, line 55: Why is `setAuth` not required when using a connection string?

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputFile.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public static AzureBlobInputFile from(AzureURI azureURI, BlobClient blobClient, AzureProperties azureProperties) {
+    return new AzureBlobInputFile(azureURI, blobClient, azureProperties);
+  }
+
+  @Override
+  public long getLength() {
+    if (length == null) {
+      length = blobClient().getProperties().getBlobSize();
```
rdblue, line 42: What happens if the blob doesn't exist? Could this throw `NotFoundException` to standardize?

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputStream.java
```diff
@@ -0,0 +1,130 @@
...
+    if (!closed) {
+      close(); // releasing resources is more important than printing the warning
+      String trace = TRACE_JOINER.join(Arrays.copyOfRange(createStack, 1, createStack.length));
+      LOG.warn("Unclosed input stream created by:\n\t{}", trace);
+    }
+  }
+
+  private void openStream(long offset) {
+    final BlobInputStreamOptions options = new BlobInputStreamOptions().setRange(new BlobRange(offset))
+        .setBlockSize(azureProperties.readBlockSize(azureURI.storageAccount()));
```
rdblue, line 126: We usually put each chained method on a separate line.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputStream.java
```diff
@@ -0,0 +1,130 @@
...
+      // Already at the specified position.
+      return;
+    }
+
+    if (newPos > pos) {
+      // Seeking forward.
+      final long bytesToSkip = newPos - pos;
+      // BlobInputStream#skip only repositions the internal pointers,
+      // the actual bytes are skipped when BlobInputStream#read is invoked.
+      final long bytesSkipped = stream.skip(bytesToSkip);
```
rdblue, line 77: Does this read through or just change the next request?

build.gradle
```diff
@@ -606,6 +613,101 @@ project(':iceberg-dell') {
...
+  task startContainer(type: DockerStartContainer) {
+    dependsOn createContainer
+    targetContainerId createContainer.getContainerId()
+  }
+
+  task stopContainer(type: DockerStopContainer) {
+    targetContainerId createContainer.getContainerId()
+  }
+
+  task integrationTest(type: Test) {
```
rdblue, line 689: Can you make sure that there is a workflow that runs these integration tests?

azure/src/main/java/org/apache/iceberg/azure/AuthType.java
```diff
@@ -0,0 +1,26 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.azure;
+
+public enum AuthType {
```
rdblue, line 22: Does this need to be public?

azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+
+public class AzureProperties implements Serializable {
+
+  // Start of storage account configuration
+  public static final String STORAGE_CONNECTION_STRING = "azure.storage.%s.connection-string";
```
rdblue, line 32: These seem long. What about `abfs.%s.uri` or `abfs.%s.connection-string` instead? Does it really need to be "connection string"?

rdblue (changes requested):
Thank for working on this, @sumeetgajjar! Overall it is looking good.

.baseline/checkstyle/checkstyle.xml
sumeetgajjar: Hi Ryan - without the above change, the `checkstyleIntegration` task fails with the following error:
```
[ant:checkstyle] [ERROR] /<redacted>/upstream-iceberg/azure/src/integration/java/org/apache/iceberg/azure/blob/TestAzureBlobOutputStream.java:38:46: Using a static member import should be avoided - org.assertj.core.api.Assertions.assertThat. [AvoidStaticImport]

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':iceberg-azure:checkstyleIntegration'.
> Checkstyle rule violations were found. See the report at: file:///<redacted>/upstream-iceberg/azure/build/reports/checkstyle/integration.html
  Checkstyle files with violations: 5
  Checkstyle violations by severity: [error:12]
```

rdblue: Yes, you shouldn't use a static import for those methods.

azure/src/main/java/org/apache/iceberg/azure/AuthType.java
sumeetgajjar: Yes, this needs to be public.
`AuthType` is accessed in `AzureProperties`, `AzureBlobClientFactory`, and `IntegrationTests`.

Making it package-private would leave `AzureProperties` unaffected since both of them reside in `org.apache.iceberg.azure` package.
However, `AzureBlobClientFactory`, and `IntegrationTests` would throw a compilation error since they reside in `org.apache.iceberg.azure.blob` package.

azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java
sumeetgajjar: > These seem long. What about abfs.%s.uri or abfs.%s.connection-string instead?

I can replace `azure.storage` with `abfs` for this and the rest of the configs to make them shorter.

> Does it really need to be "connection string"?

It would be a good idea to keep `connection-string` as is, since it would directly map to the `connection-string` config available on the Azure portal, thereby reducing the chances of confusing this config with a different one.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
sumeetgajjar: Ack

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
sumeetgajjar: The general form of a connection is as follows:
```
DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;
AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;
BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;
QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1;
TableEndpoint=http://127.0.0.1:10002/devstoreaccount1;
```

Thus the connection string contains all the necessary auth-related and storage account endpoints information required for establishing a connection and hence it is not necessary to explicitly set the auth.
https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string#configure-a-connection-string-for-an-azure-storage-account

Note: the above connection string does not leak any actual keys, it is the default connection string used by the Azurite Emulator thus safe to share on public forums.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
sumeetgajjar: Ack, I'll remove it for the local variables.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputFile.java
sumeetgajjar: It throws a `BlobStorageException: Status code 404, (BlobNotFound)` as of now. I can standardize it to throw `NotFoundException`

.baseline/checkstyle/checkstyle.xml
sumeetgajjar: Ack, 
will make the necessary modifications.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputStream.java
sumeetgajjar: It does not read through, it repositions internal pointers to change the next read request

https://github.com/Azure/azure-sdk-for-java/blob/627f10710712d00eea89dfd892587f18da2a456a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java#L419

https://github.com/Azure/azure-sdk-for-java/blob/627f10710712d00eea89dfd892587f18da2a456a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java#L379

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputStream.java
sumeetgajjar: Ack

azure/src/main/java/org/apache/iceberg/azure/blob/AzureURI.java
sumeetgajjar: Ack

build.gradle
sumeetgajjar: `java-ci` workflow already takes care of this.

The `check` task depends on the above `integrationTest` task. The `java-ci` workflow runs the `check` task on each module in Iceberg which would in turn run the integration tests for azure. 

https://github.com/apache/iceberg/blob/7c2ea0128133f630041ecb77de1fae754073e904/.github/workflows/java-ci.yml#L68

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobClientFactory.java
sumeetgajjar: Along with this, I've also made `BaseAzureBlobFile` package-private.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureURI.java
sumeetgajjar: Ack.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureURI.java
sumeetgajjar: Ack.

build.gradle
sumeetgajjar: Ack.

build.gradle
sumeetgajjar: Ack. 
Will switch to direct dependencies.

Note: Currently GCP also relies on bom. We should also make a change in the GCP module to avoid using bom.

https://github.com/apache/iceberg/blob/7c2ea0128133f630041ecb77de1fae754073e904/build.gradle#L375

sumeetgajjar: If required, I can file another PR to eliminate the usage of bom in the GCP module.

sumeetgajjar:
@rdblue thanks for the review and your comments. 
I have addressed all of your requested changes in the latest commit, please review those at your convenience.

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputStream.java
```diff
@@ -0,0 +1,131 @@
...
+    openStream(/* offset = */ 0);
+  }
+
+  @Override
+  public long getPos() {
+    return pos;
+  }
+
+  @Override
+  public void seek(long newPos) {
```
rbalamohan, line 63: Does this take care of reverse seek issues, where connections can be terminated and reopened?

sumeetgajjar: Hi @rbalamohan - can you please elaborate more on the reverse seek issues?

For the reverse seek case, we close the current stream and re-open the stream from the earlier offset.
```
// Seeking backward.
stream.close();
openStream(newPos); // newPos is a position back in the stream. 
```

rbalamohan: Connection close/reopens are expensive in cloud stores. https://issues.apache.org/jira/browse/HADOOP-12444 has more details on backward seek. Good set of tickets went in terms of fixing/improving for backward seek especially for columnar formats like ORC, Parq.

sumeetgajjar: I can follow a similar approach in the current implementation where I can skip openStream in the seek method and lazily open a stream in the subsequent read request.

electrum:
Does this support Data Lake Storage Gen2?

sumeetgajjar:
> Does this support Data Lake Storage Gen2?

Hi @electrum - yes, DataLake Storage Gen2 is supported. In fact, we only support Gen2. 
Gen1 is set to retire on Feb 29, 2024 thus no point in adding support for it: https://docs.microsoft.com/en-us/azure/data-lake-store/data-lake-store-overview

azure/src/main/java/org/apache/iceberg/azure/blob/AzureBlobInputStream.java
sumeetgajjar: Done.

sumeetgajjar:
@rdblue gentle ping to re-review the PR. 
All the suggested changes have been incorporated.

blcksrx:
Any plan to merge this PR?

kiarash-rezahanjani:
Curious about the state of this PR and if there is a plan to merge?

karlschriek:
Doesn't look like there is anything left to do. Will this be merged anytime soon? @rdblue ?

rni-HMC:
Will this be merged anytime soon? I'd love to have support for blob storage instead of only adlfs (as implemented in #8303 )

PR #4467, opened 2022-04-01
Core: Add ResourcePaths helper for REST

rdblue:
This adds a helper class, `ResourcePaths` to construct resource paths in the REST catalog. This cleans up path construction to be based on REST resource and adds support for the `prefix` parameter in the REST spec.

This also adds tests for each resource and validates the multipart namespace (null byte) handling.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
@@ -65,6 +65,7 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
   private RESTClient client = null;
   private String catalogName = null;
   private Map<String, String> properties = null;
+  private ResourcePaths paths = null;
   private Object conf = null;
   private FileIO io = null;
 
@@ -77,6 +78,7 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
     this.client = clientBuilder.apply(props);
     this.catalogName = name;
     this.properties = ImmutableMap.copyOf(props);
+    this.paths = ResourcePaths.forCatalogProperties(props);
     String ioImpl = props.get(CatalogProperties.FILE_IO_IMPL);
     this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), props, conf);
   }
@@ -92,19 +94,17 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
   }
 
   @Override
-  public List<TableIdentifier> listTables(Namespace namespace) {
-    String ns = RESTUtil.urlEncode(namespace);
-    ListTablesResponse response = client
-        .get("v1/namespaces/" + ns + "/tables", ListTablesResponse.class, ErrorHandlers.namespaceErrorHandler());
+  public List<TableIdentifier> listTables(Namespace ns) {
+    ListTablesResponse response = client.get(
+        paths.tables(ns), ListTablesResponse.class, ErrorHandlers.namespaceErrorHandler());
     return response.identifiers();
   }
 
   @Override
   public boolean dropTable(TableIdentifier identifier, boolean purge) {
-    String tablePath = tablePath(identifier);
     // TODO: support purge flagN
     try {
-      client.delete(tablePath, null, ErrorHandlers.tableErrorHandler());
+      client.delete(paths.table(identifier), null, ErrorHandlers.tableErrorHandler());
       return true;
     } catch (NoSuchTableException e) {
       return false;
@@ -117,8 +117,7 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
   }
 
   private LoadTableResponse loadInternal(TableIdentifier identifier) {
-    String tablePath = tablePath(identifier);
-    return client.get(tablePath, LoadTableResponse.class, ErrorHandlers.tableErrorHandler());
+    return client.get(paths.table(identifier), LoadTableResponse.class, ErrorHandlers.tableErrorHandler());
   }
 
   @Override
@@ -127,7 +126,7 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
     Pair<RESTClient, FileIO> clients = tableClients(response.config());
 
     return new BaseTable(
-        new RESTTableOperations(clients.first(), tablePath(identifier), clients.second(), response.tableMetadata()),
+        new RESTTableOperations(clients.first(), paths.table(identifier), clients.second(), response.tableMetadata()),
         fullTableName(identifier));
   }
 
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/apache/iceberg/rest/TestResourcePaths.java
```diff
@@ -0,0 +1,149 @@
...
+    Namespace ns = Namespace.of("n/s");
+    Assert.assertEquals("v1/ws/catalog/namespaces/n%2Fs", withPrefix.namespace(ns));
+    Assert.assertEquals("v1/namespaces/n%2Fs", withoutPrefix.namespace(ns));
+  }
+
+  @Test
+  public void testNamespaceWithMultipartNamespace() {
+    Namespace ns = Namespace.of("n", "s");
+    Assert.assertEquals("v1/ws/catalog/namespaces/n%00s", withPrefix.namespace(ns));
+    Assert.assertEquals("v1/namespaces/n%00s", withoutPrefix.namespace(ns));
```
rdblue, line 63: @kbendick, these tests failed at first and I had to update the null byte joiner. I think the problem was that the URL encoder was running on each namespace part and not on the whole thing, so null bytes were left unencoded.

kbendick: Ah ok good catch. Thank you for fixing it!

core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package org.apache.iceberg.rest;
+
+import java.util.Map;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.relocated.com.google.common.base.Joiner;
+
+public class ResourcePaths {
```
kbendick, line 27: Do we want to make this `V1ResourcePaths` or just deal with the change to `V2` if / when there is a V2?

core/src/test/java/org/apache/iceberg/rest/TestResourcePaths.java
```diff
@@ -0,0 +1,149 @@
...
+public class TestResourcePaths {
+  private final String prefix = "ws/catalog";
+  private final ResourcePaths withPrefix = ResourcePaths.forCatalogProperties(ImmutableMap.of("prefix", prefix));
+  private final ResourcePaths withoutPrefix = ResourcePaths.forCatalogProperties(ImmutableMap.of());
+
+  @Test
+  public void testConfigPath() {
+    // prefix does not affect the config route because config is merged into catalog properties
+    Assert.assertEquals(ResourcePaths.config(), "v1/config");
+  }
```
kbendick, line 37: We might consider allowing prefix for the config route as well, in case somebody wants to use prefix to support several different configs based on the full path (maybe a config returned for a different organization depending on the prefix).

So they could have:
`v1/dbas/config & v1/analysts/config` or maybe `v1/customer1/config & v1/customer2/config` or even `v1/resource_queue1/config & v1/resource_queue2/config` or however else they choose to interpret the use of `prefix`.

core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java
rdblue: We're not sure how we're going to evolve them just yet, so I just hid version entirely in this class. We can extend that way later.

core/src/test/java/org/apache/iceberg/rest/TestResourcePaths.java
rdblue: The reason why we don't support prefix like that right now is that we need an original config route, which may send back a `prefix`. We could add what you're talking about later, but we'd need to create an initial `ResourcePaths` and then a new one with the config that it returns. I think we should stick to the simple version for now and we can extend it later.

kbendick: Ok. That logic is sound to me. If somebody winds up requesting it later on, we can consider supporting it then.

core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java
kbendick: Sounds good to me. When I was working on similar, having `V1` in the class name was rather ugly. So I actually prefer to avoid it if there’s a reasonable argument against it, which you’ve provided. Thanks.

kbendick (approved):
I ran all three of table level, namespace, and config, and they all generated appropriate URLs in some other work I'm doing and they all generated the expected URL.

Here is a sample from listing tables in namespace `default` with prefix `rauls_space`:
```
https://localhost/v1/rauls_space/namespaces/default
```

Thanks @rdblue!

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -139,61 +138,57 @@ public void createNamespace(Namespace namespace, Map<String, String> metadata) {
...
+    client.post(paths.namespaces(), request, CreateNamespaceResponse.class, ErrorHandlers.namespaceErrorHandler());
   }
 
   @Override
   public List<Namespace> listNamespaces(Namespace namespace) throws NoSuchNamespaceException {
     Preconditions.checkArgument(namespace.isEmpty(), "Cannot list namespaces under parent: %s", namespace);
     // String joined = NULL.join(namespace.levels());
     ListNamespacesResponse response = client
-        .get("v1/namespaces", ListNamespacesResponse.class, ErrorHandlers.namespaceErrorHandler());
+        .get(paths.namespaces(), ListNamespacesResponse.class, ErrorHandlers.namespaceErrorHandler());
```
kbendick, line 149: This route is incorrect. List namespaces underneath a parent namespace is `GET /v1/namespaces?parent=accounting`, but this will generate `GET /v1/namespaces/accounting`.

iirc this is because listing the empty namespace is `GET /v1/namspaces`, to make it easier for systems that might _need_ the variable there in the path when it's expected.

kbendick (approved):
Actually, I just noticed that my example is wrong 😅 .

All of the other routes are correct, but list namespaces isn't correct per the contract. I am going to go in and validate all of them as I am working on tests.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: I think this is correct. We need to support passing parent still, but the path should be the namespaces resource, not the one with a specific namespace.

core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java
```diff
@@ -0,0 +1,70 @@
...
+  private static final Joiner SLASH = Joiner.on("/").skipNulls();
+  private static final String PREFIX = "prefix";
```
danielcweeks, line 29: I don't think we should force the use of `v1` as a required prefix.  The full path should be configurable with `v1` being a convention used by default.

danielcweeks: Never mind.  Ryan explianed that the initial prefix can be controlled during the REST client configuration.  LGTM

danielcweeks (approved):

Issue #4435, opened 2022-03-30
[Python] Restructure Literal and Expression


dramaticlly:
for context :https://github.com/apache/iceberg/pull/4262#pullrequestreview-899679092

Current python src structure

```
src
├── iceberg
│ ├── __init__.py
│ ├── expression
│ │ ├── __init__.py
│ │ └── literals.py
│ ├── expressions.py
│ ├── io
│ │ ├── __init__.py
│ │ ├── base.py
│ │ └── pyarrow.py
│ ├── types.py
│ └── utils
│     ├── __init__.py
│     └── bin_packing.py
```

@samredai proposed following new structure 

```
src
├── iceberg
│ ├── __init__.py
│ ├── expressions
│ │ ├── __init__.py
│ │ ├── operations.py (renamed from expressions.py)
│ │ └── literals.py
│ ├── io
│ │ ├── __init__.py
│ │ ├── base.py
│ │ └── pyarrow.py
│ ├── types.py
│ └── utils
│     ├── __init__.py
│     └── bin_packing.py
```

The PR change shall be mechanical refactoring and introduce no logical code change

samredai:
Thanks for opening this @dramaticlly, to add more, I think we should section off the ABC classes to a `base.py` file (doesn't have to be the same PR as the general restructuring). Mainly because the expressions module has a pretty gnarly sequence of class inheritance that's hard to follow when the interfaces are scattered among the concrete classes. This also keeps the files with the actual logic easy to follow since the top of those files are not full of super classes. In turn that also makes it easier to find where pdb breakpoints need to be set while debugging.

Here's what I imagine the file contents would look like:
- **base.py** (ABC classes that are never instantiated directly)
  - Literal
  - Expression
  - ExpressionVisitor
  - BoundExpressionVisitor
  - Evaluator
  - Bound
  - Unbound
  - Term
  - Predicate
- **literals.py**
  - IntegerLiteral
  - BooleanLiteral
  - DecimalLiteral
  - ...
- **operation.py**
  - Operation
  - And
  - Or
  - Not
  - TrueExp
  - FalseExp
- **predicate.py**
  - BoundPredicate
  - UnboundPredicate
- **references.py**
  - Boundreference
  - UnboundReference
- **visitors.py**
  - EvalVisitor
  - BindVisitor
  - ReferenceVisitor
  - ManifestEvalVisitor
  - MetricsEvalVisitor
- **evaluators.py**
  - InclusiveManifestEvaluator
[...]

PR #4468, opened 2022-04-01
Python: reorganize expressions module structure

dramaticlly:
This help with https://github.com/apache/iceberg/issues/4435 to provide better structure under the expressions module.

Rename old expression module to include both literals and operations

This is mechanical refactor without logical change. I also want to extract the base class out of literals but looks like there's another PR #4466  already doing so, so to minimize the conflict I will leave this PR as reorganization only

CC @samredai  @rdblue


python/tests/expressions/test_operations.py
```diff
@@ -17,26 +17,26 @@
 
 import pytest
 
-from iceberg import expressions
+from iceberg.expressions import operations
```
rdblue, line 20: Shouldn't `operations` go into `base.py` and then get exposed in `__init__.py`? I would probably leave the tests unchanged and use `expressions.Operation.*`.

samredai: +1 that operations should go into base.py but would it be helpful for the tests to use the full imports? In other words `from iceberg.expressions import base`. It makes it a bit easier to jump from looking at a test to the relevant source code.

dramaticlly:
moved `operations` into `base.py` and leave the unit test coverage for operations unchanged.

@rdblue  and @samredai  appreciate another look

python/tests/expressions/test_operations.py
dramaticlly: Thank you, updated per your suggestions!

samredai (approved):
LGTM. One small nit though is we should now rename `test_operations.py` to `test_expressions_base.py`. The pattern we've been using so far is to keep the tests directory in sync with the src directory, but `expressions/test_base.py` will run into a conflict with `io/test_base.py` (pytest wants unique filenames). We should probably also rename `io/test_base.py` to `io/test_io_base.py`.

Thanks @dramaticlly!

dramaticlly:
Thank you @samredai  for your quick review, I renamed
- test_expressions_base 
- test_io_base.py

as suggested!

rdblue (approved):

rdblue:
Thanks, @dramaticlly!

PR #4473, opened 2022-04-02
Spark 3.2: Commit consumed offsets to the checkpoint location

SreeramGarlapati:
This is Port of bug fix from our internal fork. Without this - our table maintenance expired the initial snapshot that was checkpoint'ed by the tables - which intern rendered our streaming readers to unrecoverable state.

cc: @rajarshisarkar, @daksha121, @rdblue, @aokolnychyi, @RussellSpitzer

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
@@ -79,6 +79,7 @@ public class SparkMicroBatchStream implements MicroBatchStream {
   private final boolean skipDelete;
   private final boolean skipOverwrite;
   private final Long fromTimestamp;
+  private final InitialOffsetStore initialOffsetStore;
 
   SparkMicroBatchStream(JavaSparkContext sparkContext, Table table, SparkReadConf readConf,
                         Schema expectedSchema, String checkpointLocation) {
@@ -92,7 +93,7 @@ public class SparkMicroBatchStream implements MicroBatchStream {
     this.splitOpenFileCost = readConf.splitOpenFileCost();
     this.fromTimestamp = readConf.streamFromTimestamp();
 
-    InitialOffsetStore initialOffsetStore = new InitialOffsetStore(table, checkpointLocation, fromTimestamp);
+    this.initialOffsetStore = new InitialOffsetStore(table, checkpointLocation, fromTimestamp);
     this.initialOffset = initialOffsetStore.initialOffset();
 
     this.skipDelete = readConf.streamingSkipDeleteSnapshots();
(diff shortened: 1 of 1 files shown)


RussellSpitzer (commented):
there should be a test with this patch which illustrates the issue being fixed if possible

SreeramGarlapati:
@RussellSpitzer - sparkSession is holding on to the state of the stream - internally. 
I am hitting failures when trying to reproduce the problem locally - as it needs multiple sparkSessions in the same unittest. Is there any precedence in the codebase for this pattern that you could kindly point me to pl.

putting the problem being explained above aside - the code itself has coverage in this unittest - `testResumingStreamReadFromCheckpoint` - using which I was able to debug thru & fix the code - where I need to use `File.createOrReplace()` instead of the existing `File.create()`.

cc: @rajarshisarkar

rdblue:
@SreeramGarlapati can you please update the description with a clear statement of the bug and how this fixes it?

rdblue:
I agree with @RussellSpitzer. This should have a test to verify the behavior.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -273,14 +275,13 @@ public StreamingOffset initialOffset() {
...
-      writeOffset(offset, outputFile);
-
+      writeOffset(offset);
       return offset;
     }
 
-    private void writeOffset(StreamingOffset offset, OutputFile file) {
-      try (OutputStream outputStream = file.create()) {
+    private void writeOffset(StreamingOffset offset) {
+      OutputFile file = io.newOutputFile(initialOffsetLocation);
```
singhpk234, line 283: [question] I have a doubt, can multiple streaming writing to same check-point location, result in nondeterministic state.

Let's say initial snapshot id was : snapshot(1)
Table state  (snapshot(1) -> snapshot(2) -> snapshot(3) -> snapshot(4))
Now stream 1 started in cluster 1 and read / commited till snapshot(2), 
Now at the same time stream 2 started in cluster 2, it started from snapshot(2), before it could commit snapshot(3) , stream 1 committed snapshot(3) & snapshot(4) now when stream2 tried to commit snapshot(3) it will be overwriting the state of stream 1 or vice versa, starting state of new stream let's say stream3 in cluster 3 nondeterministic,  as one stream is running ahead of other (stream 1 / stream 2).

Also earlier if two stream would have started in the same time (with no checkpoint file) one would have failed since we did a file.create() now they can co-exist. Your thoughts ?

SreeramGarlapati: @singhpk234 - multiple spark streaming clusters cannot run based on the same checkpoint location.

ajantha-bhat:
Closing this as Spark 3.2 is no longer supported in the latest version.

If issue exist on other spark version, please handle there.

PR #2330, opened 2021-03-13
Spark: Add table location to reserved properties so that the table location shows in Spark SQL Properties.

flyrain:
https://github.com/apache/iceberg/issues/2309
A unit test is added. Besides here is the integration test with Apache Spark 3.0.2:
```
spark-sql> describe table extended local.db.table;
21/03/13 00:22:33 INFO BaseMetastoreCatalog: Table loaded by catalog: local.db.table
21/03/13 00:22:35 INFO CodeGenerator: Code generated in 242.060932 ms
21/03/13 00:22:35 INFO CodeGenerator: Code generated in 11.30419 ms
id	bigint
data	string

# Partitioning
Not partitioned

# Detailed Table Information
Name	local.db.table
Location	/Users/yufei/Downloads/spark-3.0.2-bin-hadoop2.7/warehouse/db/table
Provider	iceberg
Owner	yufei
Table Properties	[current-snapshot-id=none,format=iceberg/parquet]
Time taken: 3.503 seconds, Fetched 12 row(s)
21/03/13 00:22:35 INFO SparkSQLCLIDriver: Time taken: 3.503 seconds, Fetched 12 row(s)
```

PR #4475, opened 2022-04-03
Spark: DESCRIBE TABLE EXTENDED shows identifier fields

liuml07:
Unit tests was updated. Also tested manually in Spark SQL. The first table does not have identifier fields while the second table (same schema) has identifier fields.
Screen Shot 2022-04-02 at 5 19 57 PM

Was following the pattern in #2330 for `location`. Thanks @flyrain 

Not sure is example `[userName,productId]` a good format? The comma is used to join multiple table properties in the shell.


rdblue (approved):

rdblue:
Thanks, @liuml07!

PR #3912, opened 2022-01-17
Hive: Support 'identifier-field-ids' when creating table in hive

humengyu2012:
Supporting 'identifier-field-ids' when creating table in hive.
Although this feature is supported with json, it is too difficult for most people to write a json string in hive sql.
Creating a iceberg table in hive like: 
```sql
CREATE TABLE test_iceberg (
  c1 int,
  c2 int,
  c3 string
) 
PARTITIONED BY (c4 string)
STORED BY 'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler'
TBLPROPERTIES (
  'iceberg.identifier-field-names' = 'c1,c2'
);
```

We could obtain an iceberg table:
```json
{
  "format-version" : 1,
  "table-uuid" : "7b05fa2b-9504-4ea2-8e24-5f760a101f2f",
  "location" : "hdfs://xxx:8020/user/hive/warehouse/test_iceberg",
  "last-updated-ms" : 1642454127758,
  "last-column-id" : 4,
  "schema" : {
    "type" : "struct",
    "schema-id" : 0,
    "identifier-field-ids" : [ 1, 2 ],
    "fields" : [ {
      "id" : 1,
      "name" : "c1",
      "required" : true,
      "type" : "int"
    }, {
      "id" : 2,
      "name" : "c2",
      "required" : true,
      "type" : "int"
    }, {
      "id" : 3,
      "name" : "c3",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 4,
      "name" : "c4",
      "required" : false,
      "type" : "string"
    } ]
  },
  "current-schema-id" : 0,
  "schemas" : [ {
    "type" : "struct",
    "schema-id" : 0,
    "identifier-field-ids" : [ 1, 2 ],
    "fields" : [ {
      "id" : 1,
      "name" : "c1",
      "required" : true,
      "type" : "int"
    }, {
      "id" : 2,
      "name" : "c2",
      "required" : true,
      "type" : "int"
    }, {
      "id" : 3,
      "name" : "c3",
      "required" : false,
      "type" : "string"
    }, {
      "id" : 4,
      "name" : "c4",
      "required" : false,
      "type" : "string"
    } ]
  } ],
  "partition-spec" : [ {
    "name" : "c4",
    "transform" : "identity",
    "source-id" : 4,
    "field-id" : 1000
  } ],
  "default-spec-id" : 0,
  "partition-specs" : [ {
    "spec-id" : 0,
    "fields" : [ {
[...]

PR #4478, opened 2022-04-03
Spark: Support identifier fields when creating table with SQL

liuml07:
Currently Spark SQL does not allow user to specify 'identifier-fields' when creating table. This is a PoC effort.

```sql
CREATE TABLE t1(
        id BIGINT NOT NULL,
        name STRING NOT NULL,
        data STRING)
USING iceberg
TBLPROPERTIES ('identifier-fields'='id,name')
```

After this, we can add support for other engines like Hive easily.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/Schema.java
+++ b/api/src/main/java/org/apache/iceberg/Schema.java
@@ -129,6 +129,22 @@ public class Schema implements Serializable {
     }
   }
 
+  public static Set<Integer> identifierFieldNamesToIds(StructType struct, Set<String> identifierFiledNames) {
+    Map<String, Integer> nameToId = TypeUtil.indexByName(struct);
+    Set<Integer> identifierFieldIds = Sets.newHashSet();
+    // Validate identifier field by name exists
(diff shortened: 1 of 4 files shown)


melin:
Options is recommended. TBLPROPERTIES does not match the use keyword

rdblue:
Table properties aren't a substitute for DDL clauses. I don't think that we should merge something like this. If you want to add key fields to Spark's DDL then I'm all for it. But otherwise I think we should continue to use the `ALTER TABLE` syntax.

liuml07:
Thank you both for the feedback, really appreciate that!

Originally I was following the pattern in #3912. I think again and it makes more sense to add identifier fields to DDL instead of the table properties. Meanwhile, if users need to create table with `TBLPROPERTIES ('identifier-fields'='id')` while updating with  `ALTER TABLE SET|DROP IDENTIFIER FIELDS`, it could be confusing.

rdblue:
I agree, it would be confusing. Thanks for working on this area, @liuml07!

Issue #2789, opened 2021-07-06
Implement rate limiting while reading stream from Iceberg table as Spark3 DSv2 source


SreeramGarlapati:
This is a continuation of work done in PR #2660: reading from iceberg table as an incremental streaming source in spark.

Here's an [idea proposal](https://github.com/apache/iceberg/pull/2660#discussion_r650153016) by @aokolnychyi - to use limit API once we move to 3.1.

[Please refer to the Concepts section to fully understand the issue]

## The Problem
Current implementation of Iceberg DSv2 micro_batch source - hands off all available data on the table to Spark3 DSv2. This could result in highly unpredictable/irregular stream sizes. This issue is an ask to explore - what are the concepts available in Iceberg - to rate limit the stream size.

## Implementation proposal

Flow control can be maintained at `File Level` / `Record Level`.
Iceberg table metadata layer mainly operates at `File Level` - i.e., each table is a list of data (/delete) files. 
While ideal implementation of flow control should be at Record level - keeping in mind the ROI, complexity of implementation & our current requirement - we only need `File Level` flow control.

Implementation idea is that:
1. introduce a spark option to enable rate limiting - `max-bytes-per-micro-batch` - which switches to this implementation
2. change the implementation of `SparkMicroBatchStream` to remember the tables stream position - the `latestReturnedOffset`
3. change the implementation of `latestoffset()` to **NOT** return the iceberg tables' actual `latestOffset` - but, apply flow control and measure what the NEW `latestReturnedOffset` can be.
4. change the OffsetStoreProvider to also add the `latestCommittedOffset` to be able to resume stream.

#### Background on Spark 3 - MicroBatchStream

1. Spark expects micro batch streaming source to implement the interface `MicroBatchStream`
2. Spark driver - as part of `MicroBatchExecution` - instantiates the Iceberg's `MicroBatchStream` implementation & invokes the **methods** in this `MicroBatchStream` class **in the below order**:
          - `latestOffset()` - asks Iceberg streaming source to return what is the `latestOffset` availabe on the stream 
          - `initialOffset()` - asks iceberg streaming source to return what is the first ever `initialOffset` when this stream started
          - `planInputPartitions(start, end)` - spark runtime picks a `start` offset and `end` offset based on the stream position it is at - & asks iceberg streaming source to return a dataStructure `InputPartition[]` - which will later be distributed across spark executors
[...]

PR #2660, opened 2021-06-02
Spark3 structured streaming micro_batch read support

SreeramGarlapati:
This work is an extension of the idea in issue https://github.com/apache/iceberg/issues/179 & the Spark2 work done in PR #2272 - only that - this is for Spark3.

**Implementation Notes:**
* If an Iceberg table has all Snapshots of type `Append` - `spark3.readStream...` on that Iceberg table will return all files in the First Snapshot - followed by all the data files added in the subsequent Snapshots.
* If the reader encounters any Snapshot of type `NotEqual` to `Append` - the reader fails.
* For checkpointing the Offset - this implementation plugs into default Spark3 offset checkpointing. Reader only maintains the InitialOffset; implementation & reasoning inspired from [KafkaStreamMicroBatchStream](https://github.com/apache/spark/blob/master/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaMicroBatchStream.scala).

**What's NEXT!?**
* Change the [UnitTests](https://github.com/apache/iceberg/pull/2660#discussion_r655790388) - which relies on Table APIs to simulate generation of Delete/Overwrite Snapshots to SQL Statements.
* Handling snapshots of type: `DELETES`, `REPLACE` & `OVERWRITES`.
* Refactoring `SparkBatchScan` class
   1. Separate out `Batch` & `Scan` into different implementations
   2. Draw cleaner line b/w `SparkBatchScan` & `MicroBatchStreaming` classes
* plug into spark - [stats estimation](https://github.com/apache/iceberg/pull/2660#discussion_r650084415) for `MicroBatchStreaming`
* Implement [full scan mode](https://github.com/apache/iceberg/pull/2660/files#r653979695) on first snapshot
   -  [this commit has the implementation](https://github.com/apache/iceberg/pull/2660/commits/353f2741ef6ce80ec22fcbf1d8e53c780b6edc4a) & [keep this input by @rdblue in mind](https://github.com/apache/iceberg/pull/2660#discussion_r653962262)
* Documentation update
* Rate limiting
   -  there is [an idea proposal by](https://github.com/apache/iceberg/pull/2660#discussion_r650153016) @aokolnychyi - to use limit API when we move to 3.1. 
* Accepting a different FiloIO implementation for checkpoints 
* Columnar reads are not enabled. Something for future.

cc: @aokolnychyi & @RussellSpitzer & @holdenk @rdblue @rdsr @jackye1995  #2611 

[...]

PR #4479, opened 2022-04-03
Spark 3.3: support rate limit in Spark Streaming

singhpk234:
### About the change

Supports rate limiting in SS via Admission Control API. Rate limiting is a required to have an uniform micro-batch sizes, At present spark Admission Control API exposes rate limiting on 2 things (File Read / Records Read). 

Now we can have offset that read snapshot's partially (based on number of Files / Records limits specified) as well as now a Microbatch can span across mutiple snapshots where each snapshot could be partially read. Since we took into consideration already in past that a microbatch can read from [startFileIndex, endFileIndex) no changes to microbatch interface was made.

Please ref below for more about Admission Control APIs / Read Limit : 
- https://github.com/apache/spark/pull/27380
- https://issues.apache.org/jira/browse/SPARK-30669

--- 
### Concepts used in this PR
(copied from : https://github.com/apache/iceberg/pull/2660)

#### Background on Spark 3 - MicroBatchStream

Spark expects micro batch streaming source to implement the interface MicroBatchStream

Spark driver - as part of MicroBatchExecution - instantiates the Iceberg's MicroBatchStream implementation & invokes the methods in this MicroBatchStream class in the below order:
    
 - latestOffset(Offset startOffset, ReadLimit readLimit) - asks Iceberg streaming source to return what is the latestOffset  availabe on the stream, given that previous offset was startOffset.
 - initialOffset() - asks iceberg streaming source to return what is the first ever initialOffset when this stream started
 - planInputPartitions(start, end) - spark runtime picks a start offset and end offset based on the stream position it is at - & asks iceberg streaming source to return a dataStructure InputPartition[] - which will later be distributed across spark executors
 - createReaderFactory() - iceberg micro batch streaming source should implement this method - to educate spark executor     - as to - how to read a given InputPartition
- there are other standard methods on the MicroBatchStream interface - which will be invoked by spark - like deserializeOffset, commit, stop - which are self explanatory.

#### Iceberg's Spark stream offset - StreamingOffset

Spark expects the streaming source - to implement Offset class. Offset - is a logical representation of position of the stream 

- at which spark is reading.
Iceberg's implementation (which already existed before this PR) is: StreamingOffset.

It stores 3 properties about the Iceberg table position:
   - SnapshotId: the Iceberg table Snapshot Id
   - position: file position in a given iceberg snapshot
   - ScanAllFiles: whether to stream
        All files in that Snapshot - which includes - (a) files added in older Snapshots + (b) net NEW files added in the current   

Snapshot
        or to stream - only the files that are newly added in the current Snapshot.

---
Testing :
Added UT's to assert expected number of microbatches are made.

Closes : https://github.com/apache/iceberg/issues/2789


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -251,6 +273,91 @@ private static StreamingOffset determineStartingOffset(Table table, Long fromTim
...
+          curFilesAdded += 1;
+          curRecordCount += genericDataFile.recordCount();
+        }
+
+        ++curPos;
+      }
+      // if the currentSnapShot was also the mostRecentSnapshot then break
+      if (curSnapshot.snapshotId() ==  table.currentSnapshot().snapshotId()) {
+        break;
+      }
```
singhpk234, line 330: Thinking of achieving the same via manifest index like we did it here : 

https://github.com/apache/iceberg/blob/56d8f07ce6cf7ac99f439848ebf7c8b86f5046df/core/src/main/java/org/apache/iceberg/MicroBatches.java#L116-L121

singhpk234:
cc @SreeramGarlapati, you might be interested in this change. Happy to get your thoughts on the same. This  change seems to be mostly aligned with the issue you filed a while back 

- https://github.com/apache/iceberg/issues/2789

singhpk234:
once read limit is respected via this PR:

will add support of `Trigger.AvailableNow()` introduced in spark 3.3 : https://issues.apache.org/jira/browse/SPARK-36533

singhpk234:
cc @rdblue

cccs-jc:
This is really cool. Currently, without being able limit the input rows, it's not possible to be down for too long..

.palantir/revapi.yml
```diff
@@ -261,6 +261,12 @@ acceptedBreaks:
         \ T) throws java.io.IOException, com.fasterxml.jackson.core.JacksonException\
         \ @ org.apache.iceberg.rest.RESTSerializers.UpdateRequirementDeserializer"
       justification: "False positive - JacksonException is a subclass of IOException"
+    - code: "java.method.numberOfParametersChanged"
```
jackye1995, line 264: can we add a method with the new parameters instead of changing existing parameters?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -205,12 +205,16 @@ private List<FileScanTask> planFiles(StreamingOffset startOffset, StreamingOffse
         continue;
       }
 
+      Snapshot currentSnapshot = table.snapshot(currentOffset.snapshotId());
```
jackye1995, line 208: can we avoid changes to multiple Spark versions at this moment? I guess making a new method instead of changing the method parameters would help that

cccs-eric:
@singhpk234 This would be a very useful feature for us, thanks for the contribution!  Do you have any plan on finishing the work and addressing @jackye1995 's comments?

singhpk234:
> Do you have any plan on finishing the work and addressing @jackye1995 's comments?

@cccs-eric, definitely will address @jackye1995 's feedback by this week for sure.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
singhpk234: sure restricting to one spark version at the moment.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkReadOptions.java
```diff
@@ -77,4 +77,9 @@ private SparkReadOptions() {}
 
   // Timestamp in milliseconds; start a stream from the snapshot that occurs after this timestamp
   public static final String STREAM_FROM_TIMESTAMP = "stream-from-timestamp";
+
+  // maximum file per micro_batch
+  public static final String MAX_FILES_PER_MICRO_BATCH = "max-files-per-micro-batch";
```
jackye1995, line 82: given this is a streaming feature, can we use `streaming-` prefix, and name it something like `streaming-max-files-per-batch`?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -200,25 +210,43 @@ private List<FileScanTask> planFiles(StreamingOffset startOffset, StreamingOffse
 
     StreamingOffset currentOffset = null;
 
+    // [(startOffset : startFileIndex), (endOffset : endFileIndex) )
```
jackye1995, line 213: the last ) should be ]?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -282,6 +310,128 @@ private static StreamingOffset determineStartingOffset(Table table, Long fromTim
...
+    if (startOffset.equals(StreamingOffset.START_OFFSET)) {
+      startingOffset = determineStartingOffset(table, fromTimestamp);
+    }
+
+    Snapshot curSnapshot = table.snapshot(startingOffset.snapshotId());
+    int startPosOfSnapOffset = (int) startingOffset.position();
+
+    boolean scanAllFiles = startingOffset.shouldScanAllFiles();
+
+    boolean isOk = true;
```
jackye1995, line 343: nit: can we be more concrete about this naming?

core/src/main/java/org/apache/iceberg/MicroBatchesUtil.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.util.Pair;
+
+public class MicroBatchesUtil {
```
jackye1995, line 30: can we avoid creating another util class just for this? We can have the util methods in `MicroBatches` directly.

singhpk234:
> @singhpk234 it's looking good. Just a few more small changes are suggested. Do you think you'll have time to address them next week. Cheers

Definitely @cccs-jc ! Will address these feedback by end of this week as well as raise this against 3.3. Apologies have been keeping busy due to some internal deliverables

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
singhpk234: added `)` in place of `]` to show the range endOffset is not inclusive.

core/src/main/java/org/apache/iceberg/MicroBatches.java
```diff
@@ -36,6 +37,96 @@
 public class MicroBatches {
   private MicroBatches() {}
 
+  public static List<Pair<ManifestFile, Integer>> skippedManifestIndexesFromSnapshot(
+      FileIO io, Snapshot snapshot, long startFileIndex, boolean scanAllFiles) {
+    //  Preconditions.checkArgument(startFileIndex >= 0, "startFileIndex is unexpectedly smaller
```
RussellSpitzer, line 42: Leftover code?

core/src/main/java/org/apache/iceberg/MicroBatches.java
```diff
@@ -36,6 +37,96 @@
...
+    for (ManifestFile manifest : manifestFiles) {
+      manifestIndexes.add(Pair.of(manifest, currentFileIndex));
+      currentFileIndex += manifest.addedFilesCount() + manifest.existingFilesCount();
+    }
+
+    return manifestIndexes;
+  }
+
+  /**
+   * Method to skip the manifest file in which the index is smaller than startFileIndex. For
```
RussellSpitzer, line 101: manifest file's whose index is smaller than startFileIndex

?

core/src/main/java/org/apache/iceberg/MicroBatches.java
```diff
@@ -36,6 +37,96 @@
...
+   * example, if the index list is : (m1, 0), (m2, 3), (m3, 5), and startFileIndex is 4, then the
+   * returned manifest index list is: (m2, 3), (m3, 5).
+   *
+   * @param indexedManifests List of input manifests.
+   * @param startFileIndex Index used to skip the processed manifests.
```
RussellSpitzer, line 106: Skip all manifests with an index less than or equal to this value?

core/src/main/java/org/apache/iceberg/MicroBatches.java
```diff
@@ -36,6 +37,96 @@
...
+    return manifestGroup.planFiles();
+  }
+
+  /**
+   * Method to index the data files for each manifest. For example, if manifest m1 has 3 data files,
+   * manifest m2 has 2 data files, manifest m3 has 1 data file, then the index will be (m1, 0), (m2,
+   * 3), (m3, 5).
+   *
+   * @param manifestFiles List of input manifests used to index.
+   * @return a list of manifest index with key as manifest file, value as file counts.
```
RussellSpitzer, line 85: returns a list pairing each manifest with the index number of the first data file entry in that manifest.

core/src/main/java/org/apache/iceberg/MicroBatches.java
```diff
@@ -188,25 +231,23 @@ private static List<Pair<ManifestFile, Integer>> skipManifests(
      *
      * @param indexedManifests A list of indexed manifests to generate MicroBatch
      * @param startFileIndex A startFileIndex used to skip processed files.
+     * @param endFileIndex An endFileIndex used to find files to include, it not inclusive.
```
RussellSpitzer, line 234: , it not inclusive => exclusive

jackye1995 (approved):
Looks good to me! Waiting for any other comment from @RussellSpitzer

jackye1995:
@RussellSpitzer do you have any further comment? If not I think this is good to go

cccs-jc:
@RussellSpitzer good to go?

jackye1995:
Looks like this has been inactive and I think it's ready to go. 

I know we just added Spark 3.4, but given the fact that this PR has been hanging for a long time, I think it's probably better to get this in first. I will go ahead to merge it. @singhpk234 or @cccs-jc could you also add this patch to 3.4 afterwards?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -283,6 +302,128 @@ private static StreamingOffset determineStartingOffset(Table table, Long fromTim
...
+                    table.specs(),
+                    caseSensitive,
+                    curSnapshot,
+                    indexedManifests.get(idx).first(),
+                    scanAllFiles);
+            CloseableIterator<FileScanTask> taskIter = taskIterable.iterator()) {
+          while (taskIter.hasNext()) {
+            FileScanTask task = taskIter.next();
+            if (curPos >= startPosOfSnapOffset) {
+              // TODO : use readLimit provided in function param, the readLimits are derived from
```
namrathamyske, line 362: @singhpk234 any reason we don't use this from function parameter and marked this as TODO?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/source/TestStructuredStreamingRead3.java
```diff
@@ -138,6 +143,57 @@ public void testReadStreamOnIcebergTableWithMultipleSnapshots() throws Exception
...
+  }
+
+  @Test
+  public void testReadStreamOnIcebergTableWithMultipleSnapshots_WithNumberOfRows_1()
+      throws Exception {
+    appendDataAsMultipleSnapshots(TEST_DATA_MULTIPLE_SNAPSHOTS);
+
+    // only 1 micro-batch will be formed and we will read data partially
+    Assert.assertEquals(
+        1,
```
RahmanQureshi, line 175: @singhpk234 👋 Can you help me understand why the expected result here 1 and not 7? For example, the test "WithNumberOfRows_4" expects all of the test data to have been processed in 2 micro batches

RahmanQureshi: Ohh, I see! The next file has 2 records, and the rate limiting isn't sophisticated enough to break a file into multiple microbatches (which makes sense).

But I am curious why the first list of records gets broken into two files deterministically. The first file has simple-record-1, and the second file has simple-record-2 and simple-record-3.

PR #4480, opened 2022-04-03
Python: Refactor to use common decimal and datetime util

rdblue:
This fixes a couple of minor comments from #4416 and refactors to use common decimal and datetime util functions.

### DIFF
--- a/python/src/iceberg/utils/decimal.py
+++ b/python/src/iceberg/utils/decimal.py
@@ -48,7 +48,7 @@ def unscaled_to_decimal(unscaled: int, scale: int) -> Decimal:
     return Decimal((sign, digits, -scale))
 
 
-def min_size(value: Union[int, Decimal]) -> int:
+def bytes_required(value: Union[int, Decimal]) -> int:
     """Returns the minimum number of bytes needed to serialize a decimal or unscaled value
 
     Args:
@@ -74,4 +74,4 @@ def decimal_to_bytes(value: Decimal) -> bytes:
         bytes: the unscaled value of the Decimal as bytes
     """
     unscaled_value = decimal_to_unscaled(value)
-    return unscaled_value.to_bytes(min_size(unscaled_value), byteorder="big", signed=True)
+    return unscaled_value.to_bytes(bytes_required(unscaled_value), byteorder="big", signed=True)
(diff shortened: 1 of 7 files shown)


jun-he:
SGTM, thanks @rdblue

python/src/iceberg/utils/decimal.py
```diff
@@ -0,0 +1,70 @@
...
+    if isinstance(value, int):
+        return (value.bit_length() + 7) // 8
+    elif isinstance(value, Decimal):
+        return (decimal_to_unscaled(value).bit_length() + 7) // 8
+
+    raise ValueError(f"Unsupported value: {value}")
+
+
+def decimal_to_bytes(value: Decimal) -> bytes:
+    unscaled_value = decimal_to_unscaled(value)
```
samredai, line 69: Suggesting a docstring here:

    """Returns a byte representation of a decimal

    Args:
        value (Decimal): a decimal value
    Returns:
        bytes: A byte representation of the decimal unscaled
    """

samredai (approved):
One suggestion but otherwise LGTM

python/src/iceberg/utils/decimal.py
rdblue: Thanks! I updated this.

python/src/iceberg/utils/decimal.py
```diff
@@ -0,0 +1,77 @@
...
+        scale (int): A scale to set for the returned Decimal instance
+
+    Returns:
+        Decimal: A scaled Decimal instance
+    """
+    sign, digits, _ = Decimal(unscaled).as_tuple()
+    return Decimal((sign, digits, -scale))
+
+
+def min_size(value: Union[int, Decimal]) -> int:
```
danielcweeks, line 51: Should we call this `min_bytes`?

rdblue: To me, that implies that this returns some min as bytes, rather than a size. What about `required_size`?

rdblue: Updated to `bytes_required`.

rdblue:
Thanks for reviewing, @jun-he, @samredai, and @danielcweeks!

PR #4489, opened 2022-04-04
Dell: Add metrics tracking for EcsFileIO

rajarshisarkar:
This PR adds metrics tracking for `EcsFileIO`. This would help reporting IO metrics via the Hadoop `FileSystem.Statistics`.

---
cc: @danielcweeks @rdblue @kbendick

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -612,12 +612,6 @@ project(':iceberg-dell') {
     testImplementation "javax.xml.bind:jaxb-api"
     testImplementation "javax.activation:activation"
     testImplementation "org.glassfish.jaxb:jaxb-runtime"
-    testImplementation("org.apache.hadoop:hadoop-common") {
-      exclude group: 'org.apache.avro', module: 'avro'
-      exclude group: 'org.slf4j', module: 'slf4j-log4j12'
-      exclude group: 'javax.servlet', module: 'servlet-api'
-      exclude group: 'com.google.code.gson', module: 'gson'
-    }
   }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -612,6 +612,12 @@ project(':iceberg-dell') {
     testImplementation "javax.xml.bind:jaxb-api"
     testImplementation "javax.activation:activation"
     testImplementation "org.glassfish.jaxb:jaxb-runtime"
+    testImplementation("org.apache.hadoop:hadoop-common") {
+      exclude group: 'org.apache.avro', module: 'avro'
+      exclude group: 'org.slf4j', module: 'slf4j-log4j12'
+      exclude group: 'javax.servlet', module: 'servlet-api'
+      exclude group: 'com.google.code.gson', module: 'gson'
+    }
   }
 }
 
--- a/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
+++ b/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
@@ -90,7 +90,8 @@ public class EcsFileIO implements FileIO {
       this.metrics = ctor.newInstance("ecs");
 
       metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
+    } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
+      this.metrics = MetricsContext.nullMetrics();
       LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
     }
   }
--- a/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
+++ b/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
@@ -91,7 +91,6 @@ public class EcsFileIO implements FileIO {
 
       metrics.initialize(properties);
     } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
-      this.metrics = MetricsContext.nullMetrics();
       LOG.warn("Unable to load metrics class: '{}', falling back to null metrics", DEFAULT_METRICS_IMPL, e);
     }
   }
--- a/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
+++ b/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
@@ -22,13 +22,17 @@ package org.apache.iceberg.dell.ecs;
 import com.emc.object.s3.S3Client;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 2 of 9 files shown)


dell/src/main/java/org/apache/iceberg/dell/ecs/EcsOutputFile.java
```diff
@@ -25,18 +25,24 @@
...
 
 class EcsOutputFile extends BaseEcsFile implements OutputFile {
   public static EcsOutputFile fromLocation(String location, S3Client client) {
-    return new EcsOutputFile(client, new EcsURI(location), new DellProperties());
+    return new EcsOutputFile(client, new EcsURI(location), new DellProperties(), MetricsContext.nullMetrics());
   }
 
   public static EcsOutputFile fromLocation(String location, S3Client client, DellProperties dellProperties) {
-    return new EcsOutputFile(client, new EcsURI(location), dellProperties);
+    return new EcsOutputFile(client, new EcsURI(location), dellProperties, MetricsContext.nullMetrics());
```
rdblue, line 36: Should we deprecate this factory method?

build.gradle
```diff
@@ -612,6 +612,12 @@ project(':iceberg-dell') {
     testImplementation "javax.xml.bind:jaxb-api"
     testImplementation "javax.activation:activation"
     testImplementation "org.glassfish.jaxb:jaxb-runtime"
+    testImplementation("org.apache.hadoop:hadoop-common") {
+      exclude group: 'org.apache.avro', module: 'avro'
+      exclude group: 'org.slf4j', module: 'slf4j-log4j12'
+      exclude group: 'javax.servlet', module: 'servlet-api'
+      exclude group: 'com.google.code.gson', module: 'gson'
+    }
```
rdblue, line 620: Is this needed? The tests all use null metrics, so it seems like we don't need this for testing.

rajarshisarkar: `TestEcsCatalog` seems to [initialise a FileIO instance](https://github.com/apache/iceberg/blob/master/dell/src/test/java/org/apache/iceberg/dell/ecs/TestEcsCatalog.java#L62) thereby causing the following exception without the change:

```
org/apache/hadoop/fs/FileSystem
java.lang.NoClassDefFoundError: org/apache/hadoop/fs/FileSystem
	at org.apache.iceberg.hadoop.HadoopMetricsContext.initialize(HadoopMetricsContext.java:50)
	at org.apache.iceberg.dell.ecs.EcsFileIO.initialize(EcsFileIO.java:92)
	at org.apache.iceberg.dell.ecs.EcsCatalog.initializeFileIO(EcsCatalog.java:131)
	at org.apache.iceberg.dell.ecs.EcsCatalog.initialize(EcsCatalog.java:108)
	at org.apache.iceberg.dell.ecs.TestEcsCatalog.before(TestEcsCatalog.java:62)
```

This was because `NoClassDefFoundError` was not handled in the catch block due to which metrics was not falling back to null metrics in case the exception happens while doing `metrics.initialize(properties)`. Reference:

https://github.com/apache/iceberg/blob/dfdeb149843ce54e4f7fdd7cc5508477bb6baff6/dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java#L86-L95

I think the fallback should be handled in other FileIO's as well. Do let me know your thoughts on this.

dell/src/main/java/org/apache/iceberg/dell/ecs/EcsOutputFile.java
rajarshisarkar: Yes, that should be in-line with the other FileIO's. I have deprecated both the public static methods.

dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
```diff
@@ -74,6 +82,18 @@ public void initialize(Map<String, String> properties) {
...
+
+    // Report Hadoop metrics if Hadoop is available
+    try {
+      DynConstructors.Ctor<MetricsContext> ctor =
+          DynConstructors.builder(MetricsContext.class).hiddenImpl(DEFAULT_METRICS_IMPL, String.class).buildChecked();
+      this.metrics = ctor.newInstance("ecs");
+
+      metrics.initialize(properties);
+    } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
+      this.metrics = MetricsContext.nullMetrics();
```
rdblue, line 94: This is what metrics is already initialized to.

dell/src/main/java/org/apache/iceberg/dell/ecs/EcsOutputFile.java
```diff
@@ -25,18 +25,33 @@
...
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.io.PositionOutputStream;
+import org.apache.iceberg.metrics.MetricsContext;
 
 class EcsOutputFile extends BaseEcsFile implements OutputFile {
+
+  /**
+   * @deprecated moving to package-private in 0.15.0
+   */
+  @Deprecated
```
rdblue, line 35: Actually, the class is already package-private, so we can just remove it. Sorry I didn't notice that before.

rajarshisarkar: I see, I have reverted the changes.

openinx (approved):
Looks good to me. Will merge this if CI says good.

dell/src/main/java/org/apache/iceberg/dell/ecs/EcsFileIO.java
rajarshisarkar: Actually that is being overridden in the current logic. I have added some in-line comments below to highlight the issue:

```
private MetricsContext metrics = MetricsContext.nullMetrics(); // metrics initialised as nullMetrics at class level
...
...
  @Override
...
...
    try {
      this.metrics = ctor.newInstance("ecs"); // metrics re-initialised as new HadoopMetricsContext("ecs")

      metrics.initialize(properties); // throws exception in case of classpath issues 
    } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
      // We swallow the exception and allow new HadoopMetricsContext() to flow down to EcsInputFile/EcsOutputFile
    }
  }
```

We can handle the fallback in the catch block (do suggest if there is some better way). Without that the CI seems to fail for commit: [c7e355f](https://github.com/apache/iceberg/pull/4489/commits/c7e355fc0820c5630218ac0e5ae1f125852e210a)

The CI passes is we handle the nullMetrics fallback in the catch block for commit: [2ad9cda](https://github.com/apache/iceberg/pull/4489/commits/2ad9cdae82ed3714ee58f96abfc95daf1b880a58)

openinx:
Merged ! Thanks all for reviewing, and thanks @rajarshisarkar for the contribution !

PR #4491, opened 2022-04-04
Nessie: Extract Catalog client code to NessieClient for Trino Consumption

nastra:
This adds a new `NessieIcebergClient` class that encapsulates some of
the features that have been used across the `NessieCatalog`.
It also slightly improves the error messages to include the reference or
the table name when something fails.

Note that the `testFailure()` method was updated, because we're now
always using the same `UpdateableReference` between `NessieCatalog` and
`NessieTableOperations`, so that it's higly unlikely that those two get
out-of-sync.

### DIFF
--- a/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
+++ b/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
@@ -23,7 +23,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 import org.apache.hadoop.conf.Configurable;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.iceberg.BaseMetastoreCatalog;
@@ -34,37 +33,19 @@ import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.SupportsNamespaces;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.common.DynMethods;
-import org.apache.iceberg.exceptions.AlreadyExistsException;
-import org.apache.iceberg.exceptions.CommitFailedException;
-import org.apache.iceberg.exceptions.CommitStateUnknownException;
 import org.apache.iceberg.exceptions.NamespaceNotEmptyException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
-import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.hadoop.HadoopFileIO;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
-import org.apache.iceberg.util.Tasks;
 import org.projectnessie.client.NessieClientBuilder;
 import org.projectnessie.client.NessieConfigConstants;
-import org.projectnessie.client.api.CommitMultipleOperationsBuilder;
 import org.projectnessie.client.api.NessieApiV1;
 import org.projectnessie.client.http.HttpClientBuilder;
-import org.projectnessie.client.http.HttpClientException;
-import org.projectnessie.error.BaseNessieClientServerException;
-import org.projectnessie.error.NessieConflictException;
-import org.projectnessie.error.NessieNamespaceAlreadyExistsException;
-import org.projectnessie.error.NessieNamespaceNotEmptyException;
-import org.projectnessie.error.NessieNamespaceNotFoundException;
-import org.projectnessie.error.NessieNotFoundException;
-import org.projectnessie.error.NessieReferenceNotFoundException;
-import org.projectnessie.model.Branch;
 import org.projectnessie.model.ContentKey;
-import org.projectnessie.model.GetNamespacesResponse;
-import org.projectnessie.model.IcebergTable;
-import org.projectnessie.model.Operation;
 import org.projectnessie.model.TableReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -127,7 +108,7 @@ public class NessieCatalog extends BaseMetastoreCatalog implements AutoCloseable
     NessieApiV1 api = createNessieClientBuilder(options.get(NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL))
         .fromConfig(x -> options.get(removePrefix.apply(x)))
         .build(NessieApiV1.class);
-    this.client = new NessieIcebergClient(api, requestedRef, null);
+    this.client = new NessieIcebergClient(api, requestedRef, null, catalogOptions);
   }
 
   private static NessieClientBuilder<?> createNessieClientBuilder(String customBuilder) {
@@ -159,13 +140,9 @@ public class NessieCatalog extends BaseMetastoreCatalog implements AutoCloseable
     TableReference tr = TableReference.parse(tableIdentifier.name());
     Preconditions.checkArgument(!tr.hasTimestamp(), "Invalid table name: # is only allowed for hashes (reference by " +
         "timestamp is not supported)");
-    NessieIcebergClient newClient = client;
-    if (tr.getReference() != null) {
-      newClient = new NessieIcebergClient(client.getApi(), tr.getReference(), tr.getHash());
-    }
     return new NessieTableOperations(
         ContentKey.of(org.projectnessie.model.Namespace.of(tableIdentifier.namespace().levels()), tr.getName()),
-        newClient,
+        client.withReference(tr.getReference(), tr.getHash()),
         fileIO,
         catalogOptions);
   }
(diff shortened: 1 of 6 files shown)


nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -235,20 +228,20 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
-    reference.checkMutable();
+    client.getRef().checkMutable();
 
     TableIdentifier to = NessieUtil.removeCatalogName(toOriginal, name());
 
-    IcebergTable existingFromTable = table(from);
+    IcebergTable existingFromTable = client.table(from);
     if (existingFromTable == null) {
-      throw new NoSuchTableException("table %s doesn't exists", from.name());
+      throw new NoSuchTableException("table '%s' doesn't exists", from.name());
```
rdblue, line 237: As long as we're touching these, could we use the more standard exception messages? This is usually `Table does not exist: %s`. That's nice because you don't need quotes to see when it is empty.

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -290,32 +285,34 @@ public void renameTable(TableIdentifier from, TableIdentifier toOriginal) {
   @Override
   public void createNamespace(Namespace namespace, Map<String, String> metadata) {
     try {
-      reference.checkMutable();
-      api.createNamespace()
-          .reference(reference.getReference())
+      client.getRef().checkMutable();
```
rdblue, line 288: Shouldn't the client handle this?

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -290,32 +285,34 @@ public void renameTable(TableIdentifier from, TableIdentifier toOriginal) {
...
+      client.getApi().createNamespace()
```
rdblue, line 289: It seems to me that the client should expose `createNamespace()` that does the check above and exposes this API? It seems weird that this is basically asking the client for its internal state and modifying that state (violates "tell don't ask").

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -290,32 +285,34 @@ public void renameTable(TableIdentifier from, TableIdentifier toOriginal) {
...
           .namespace(org.projectnessie.model.Namespace.of(namespace.levels()))
           .create();
-      refresh();
+      client.refresh();
     } catch (NessieNamespaceAlreadyExistsException e) {
       throw new AlreadyExistsException(e, "Namespace '%s' already exists.", namespace);
     } catch (NessieNotFoundException e) {
-      throw new RuntimeException(String.format("Cannot create Namespace '%s': ref is no longer valid.", namespace), e);
+      throw new RuntimeException(String.format("Cannot create Namespace '%s': " +
+          "ref '%s' is no longer valid.", namespace, client.getRef().getName()), e);
```
rdblue, line 298: Can this be moved into the Iceberg client?

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
nastra: sure, will update those accordingly

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
nastra: I agree. my initial approach was to just move a few things over to the `NessieIcebergClient`, but now it makes perfect sense to have most of the code in the client itself, which is then just used by the catalog

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
nastra: yep, done

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -133,7 +105,10 @@ public void initialize(String inputName, Map<String, String> options) {
       throw new IllegalStateException("Parameter 'warehouse' not set, Nessie can't store data.");
     }
     final String requestedRef = options.get(removePrefix.apply(NessieConfigConstants.CONF_NESSIE_REF));
-    this.reference = loadReference(requestedRef, null);
+    NessieApiV1 api = createNessieClientBuilder(options.get(NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL))
```
snazy, line 108: Maybe move this (lines 107..111) into a static builder-ish method in NessieIcebergClient?

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -165,14 +140,13 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
     TableReference tr = TableReference.parse(tableIdentifier.name());
     Preconditions.checkArgument(!tr.hasTimestamp(), "Invalid table name: # is only allowed for hashes (reference by " +
         "timestamp is not supported)");
-    UpdateableReference newReference = this.reference;
+    NessieIcebergClient newClient = client;
     if (tr.getReference() != null) {
-      newReference = loadReference(tr.getReference(), tr.getHash());
+      newClient = new NessieIcebergClient(client.getApi(), tr.getReference(), tr.getHash(), catalogOptions);
```
snazy, line 145: Looks like this can become a `forTableReference(tr)` (that also accepts `null`) - then it seems you can get rid of the `catalogOptions` field in this clas.

I.e. the below would become:
```java
    return new NessieTableOperations(
        ContentKey.of(org.projectnessie.model.Namespace.of(tableIdentifier.namespace().levels()), tr.getName()),
        client.forTableReference(tr),,
        fileIO,
        catalogOptions);
```

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
nastra: I have another cleanup PR in the pipeline and will adjust this part of the code accordingly there

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
nastra: I have another cleanup PR in the pipeline and will adjust this part of the code accordingly there

snazy (approved):

nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
```diff
@@ -0,0 +1,334 @@
...
+          "ref '%s' no longer exists.", from.name(), to.name(), getRef().getName()), e);
+    } catch (BaseNessieClientServerException e) {
+      throw new CommitFailedException(e, "Cannot rename table '%s' to '%s': " +
+          "the current reference is not up to date.", from.name(), to.name());
+    } catch (HttpClientException ex) {
+      // Intentionally catch all nessie-client-exceptions here and not just the "timeout" variant
+      // to catch all kinds of network errors (e.g. connection reset). Network code implementation
+      // details and all kinds of network devices can induce unexpected behavior. So better be
+      // safe than sorry.
+      throw new CommitStateUnknownException(ex);
```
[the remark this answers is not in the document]
nastra: we can probably change this in a follow-up. This is just code that was moved from the `NessieCatalog` to this place

nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
```diff
@@ -0,0 +1,334 @@
...
+import org.projectnessie.model.EntriesResponse;
+import org.projectnessie.model.GetNamespacesResponse;
+import org.projectnessie.model.IcebergTable;
+import org.projectnessie.model.Operation;
+import org.projectnessie.model.Reference;
+import org.projectnessie.model.Tag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NessieIcebergClient implements AutoCloseable {
```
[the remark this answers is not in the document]
nastra: as discussed on Slack, the idea was initially to clean things up in the NessieCatalog. However, I realized that we can re-use this entire code in Trino (where we're currently integrating Nessie as well). That's the main reason for making this class public

nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
```diff
@@ -0,0 +1,334 @@
...
+          ref = Branch.of(ref.getName(), hash);
+        } else {
+          ref = Tag.of(ref.getName(), hash);
+        }
+      }
+      return new UpdateableReference(ref, hash != null);
+    } catch (NessieNotFoundException ex) {
+      if (requestedRef != null) {
+        throw new IllegalArgumentException(String.format("Nessie ref '%s' does not exist. This ref must exist " +
+            "before creating a NessieCatalog.", requestedRef), ex);
```
RussellSpitzer, line 110: Message here may be confusing since this is no longer part of a catalog class, although it only would be called by a catalog class ....

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -323,22 +321,35 @@ public void testExistingTableUpdate() {
...
-    Assertions.assertThatThrownBy(() -> icebergTable.updateSchema().addColumn("data", Types.LongType.get()).commit())
-        .isInstanceOf(CommitFailedException.class)
-        .hasMessage(
-            "Cannot commit: Reference hash is out of date. Update the reference iceberg-table-test and try again");
+    Assertions.assertThat(api.getCommitLog().refName(BRANCH).get().getLogEntries())
+        .extracting(e -> e.getCommitMeta().getHash())
+        .hasSize(2)
+        .containsExactly(commit.getHash(), branch.getHash());
+
+    // previously this would fail with "Cannot commit: Reference hash is out of date. Update the reference ..."
```
RussellSpitzer, line 343: May want to remove this prior to commit

nastra: done

nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
nastra: updated

RussellSpitzer (approved):
Looks good to me.

RussellSpitzer:
@nastra Merged! Thanks for the PR and I hope to see that Nessie Trino Support soon!

nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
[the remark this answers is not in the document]
nastra: changed in https://github.com/apache/iceberg/pull/4592

PR #4492, opened 2022-04-04
Core: Add stage-create flag to REST create route

rdblue:
This fixes an outstanding item in `RESTCatalog`, which was to determine whether to create a separate route for create table transactions. This reuses the existing create table route with a flag to stage the create to begin the transaction. This includes both spec and code updates.


danielcweeks (approved):
LGTM

rdblue:
Thanks for reviewing, @danielcweeks! Since there was already a lot of time to discuss [alternatives on the original thread](https://github.com/apache/iceberg/pull/4348#pullrequestreview-912096963), I think we have consensus and I'll merge this.

PR #4495, opened 2022-04-04
Core - REST Catalog call /v1/config route

kbendick:
Adds the initial call to `/v1/config` and merges the server-side configuration with the client configuration.

Also adds in an object mapper specific to the RESTCatalog, so that the object mapper in `JsonUtils` doesn't need to be modified.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
+++ b/core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
+
+import java.util.Map;
+import java.util.function.Function;
+import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/**
+ * Takes in the full configuration for the {@link RESTCatalog}, which should already have
+ * called the server's initial configuration route.
+ * Using the merged configuration, an instance of {@link RESTClient} is obtained that can be used with the
+ * RESTCatalog.
+ */
+public class HTTPClientFactory implements Function<Map<String, String>, RESTClient> {
+
+  @Override
+  public RESTClient apply(Map<String, String> properties) {
+    Preconditions.checkArgument(properties != null, "Invalid configuration: null");
+    Preconditions.checkArgument(properties.containsKey(CatalogProperties.URI), "REST Catalog server URI is required");
+
+    String baseURI = properties.get(CatalogProperties.URI).trim();
+
+    HTTPClient.Builder builder = HTTPClient.builder()
+        .uri(baseURI);
+
+    // Only apply bearer auth token if one is provided.
+    String token = properties.get(RESTCatalogProperties.AUTH_TOKEN);
+    if (token != null && !token.trim().isEmpty()) {
+      builder.withBearerAuth(token.trim());
+    }
+
+    return builder.build();
+  }
+}
--- a/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
@@ -19,6 +19,8 @@
 
 package org.apache.iceberg.rest;
 
+import java.io.Closeable;
+import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -57,10 +59,14 @@ import org.apache.iceberg.rest.responses.GetNamespaceResponse;
 import org.apache.iceberg.rest.responses.ListNamespacesResponse;
 import org.apache.iceberg.rest.responses.ListTablesResponse;
 import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.apache.iceberg.rest.responses.RESTCatalogConfigResponse;
 import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
 import org.apache.iceberg.util.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Configuration> {
+public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Configuration>, Closeable {
+  private static final Logger LOG = LoggerFactory.getLogger(RESTCatalog.class);
   private final Function<Map<String, String>, RESTClient> clientBuilder;
   private RESTClient client = null;
   private String catalogName = null;
@@ -69,18 +75,28 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
   private Object conf = null;
   private FileIO io = null;
 
+  public RESTCatalog() {
+    this(new HTTPClientFactory());
+  }
+
   RESTCatalog(Function<Map<String, String>, RESTClient> clientBuilder) {
     this.clientBuilder = clientBuilder;
   }
 
   @Override
   public void initialize(String name, Map<String, String> props) {
-    this.client = clientBuilder.apply(props);
+    RESTCatalogConfigResponse config = fetchConfig(props);
+    Map<String, String> mergedProps = config.merge(props);
+    this.client = clientBuilder.apply(mergedProps);
     this.catalogName = name;
-    this.properties = ImmutableMap.copyOf(props);
-    this.paths = ResourcePaths.forCatalogProperties(props);
-    String ioImpl = props.get(CatalogProperties.FILE_IO_IMPL);
-    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), props, conf);
+    this.properties = mergedProps;
+    this.paths = ResourcePaths.forCatalogProperties(properties);
+    String ioImpl = properties.get(CatalogProperties.FILE_IO_IMPL);
+    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), properties, conf);
+  }
+
+  public Map<String, String> properties() {
+    return properties;
   }
 
   @Override
@@ -197,6 +213,13 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
   @Override
   public TableBuilder buildTable(TableIdentifier identifier, Schema schema) {
     return new Builder(identifier, schema);
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (client != null) {
+      client.close();
+    }
   }
 
   private class Builder implements TableBuilder {
@@ -404,5 +427,24 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
     RESTClient tableClient = clientBuilder.apply(fullConf);
 
     return Pair.of(tableClient, tableIO);
+  }
+
+  private RESTCatalogConfigResponse fetchConfig(Map<String, String> props) {
+    // Create a client for one time use, as we will reconfigure the client using the merged server and application
+    // defined configuration.
+    RESTClient singleUseClient = clientBuilder.apply(props);
+
+    try {
+      RESTCatalogConfigResponse configResponse = singleUseClient
+          .get(ResourcePaths.config(), RESTCatalogConfigResponse.class, ErrorHandlers.defaultErrorHandler());
+      configResponse.validate();
+      return configResponse;
+    } finally {
+      try {
+        singleUseClient.close();
+      } catch (IOException e) {
+        LOG.error("Failed to close HTTP client used for getting catalog configuration. Possible resource leak.", e);
+      }
+    }
   }
 }
--- a/core/src/main/java/org/apache/iceberg/rest/RESTCatalogProperties.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTCatalogProperties.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
+
+/**
+ * Catalog properties which are specific to the RESTCatalog, which can be used in conjunction with
+ * {@link org.apache.iceberg.CatalogProperties}
+ */
+public class RESTCatalogProperties {
+
+  private RESTCatalogProperties() {
+  }
+
+  /**
+   * A Bearer authorization token which will be used to authenticate requests with the server.
+   */
+  public static final String AUTH_TOKEN = "token";
+}
--- a/core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class RESTObjectMapper {
+  private static final JsonFactory FACTORY = new JsonFactory();
+  private static final ObjectMapper MAPPER = new ObjectMapper(FACTORY);
+  private static volatile boolean isInitialized = false;
+
+  private RESTObjectMapper() {
+  }
+
+  static ObjectMapper mapper() {
+    if (!isInitialized) {
+      synchronized (RESTObjectMapper.class) {
+        if (!isInitialized) {
+          MAPPER.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+          MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+          RESTSerializers.registerAll(MAPPER);
+          isInitialized = true;
+        }
+      }
+    }
+
+    return MAPPER;
+  }
+}
--- a/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java
+++ b/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java
@@ -19,7 +19,6 @@
 
 package org.apache.iceberg.rest;
 
-import java.io.Closeable;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
@@ -47,6 +46,7 @@ import org.apache.iceberg.rest.requests.CreateTableRequest;
 import org.apache.iceberg.rest.requests.UpdateNamespacePropertiesRequest;
 import org.apache.iceberg.rest.requests.UpdateTableRequest;
 import org.apache.iceberg.rest.responses.ErrorResponse;
+import org.apache.iceberg.rest.responses.RESTCatalogConfigResponse;
 import org.apache.iceberg.util.Pair;
 
 /**
@@ -154,8 +154,7 @@ public class RESTCatalogAdapter implements RESTClient {
                                                   Object body, Class<T> responseType) {
     switch (route) {
       case CONFIG:
-        // TODO: use the correct response object
-        return castResponse(responseType, ImmutableMap.of());
+        return castResponse(responseType, RESTCatalogConfigResponse.builder().build());
 
       case LIST_NAMESPACES:
         if (asNamespaceCatalog != null) {
@@ -281,9 +280,9 @@ public class RESTCatalogAdapter implements RESTClient {
 
   @Override
   public void close() throws IOException {
-    if (catalog instanceof Closeable) {
-      ((Closeable) catalog).close();
-    }
+    // The calling test is responsible for closing the underlying catalog backing this REST catalog
+    // so that the underlying backend catalog is not closed and reopened during the REST catalog's
+    // initialize method when fetching the server configuration.
   }
 
   private static class BadResponseType extends RuntimeException {
--- a/core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
+++ b/core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
@@ -20,15 +20,23 @@
 package org.apache.iceberg.rest;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.UUID;
+import java.util.function.Consumer;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.iceberg.AssertHelpers;
 import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.catalog.CatalogTests;
 import org.apache.iceberg.jdbc.JdbcCatalog;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.rest.responses.ErrorResponse;
+import org.apache.iceberg.rest.responses.RESTCatalogConfigResponse;
+import org.junit.Assert;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
 public class TestRESTCatalog extends CatalogTests<RESTCatalog> {
@@ -36,13 +44,14 @@ public class TestRESTCatalog extends CatalogTests<RESTCatalog> {
   public Path temp;
 
   private RESTCatalog restCatalog;
+  private JdbcCatalog backendCatalog;
 
   @BeforeEach
   public void createCatalog() {
     File warehouse = temp.toFile();
     Configuration conf = new Configuration();
 
-    JdbcCatalog backendCatalog = new JdbcCatalog();
+    this.backendCatalog = new JdbcCatalog();
     backendCatalog.setConf(conf);
     Map<String, String> backendCatalogProperties = ImmutableMap.of(
         CatalogProperties.WAREHOUSE_LOCATION, warehouse.getAbsolutePath(),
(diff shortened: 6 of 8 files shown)


core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java
```diff
@@ -0,0 +1,48 @@
...
+      synchronized (RESTObjectMapper.class) {
+        if (!isInitialized) {
+          MAPPER.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+          RESTSerializers.registerAll(MAPPER);
+          isInitialized = true;
+        }
+      }
+    }
+
+    return MAPPER;
```
kbendick, line 46: We can move this to another PR if we'd like.

We need an object mapper that is configured with the necessary visibility modifiers.

I made it lazy like this so that people who don't need this don't wind up loading two object mappers, but that could be overkill.

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ * RESTCatalog.
+ */
+public class HTTPClientFactory implements RESTClientFactory {
+
+  @Override
+  public RESTClient apply(Map<String, String> properties) {
+    Preconditions.checkArgument(properties != null, "Invalid initial configuration: null");
+    Preconditions.checkArgument(properties.containsKey(CatalogProperties.URI), "REST Catalog server URI is required");
+    Preconditions.checkArgument(properties.containsKey(RESTCatalogProperties.AUTH_TOKEN),
+        "Invalid http client configuration. Missing authentication token");
```
rdblue, line 39: Is this required? I think we should use it if it's there, but we don't need to force other people into using an auth token.

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ * Takes in the full configuration for the {@link RESTCatalog}, which should already have
+ * called the server's initial configuration route.
+ * Using the merged configuration, an instance of {@link RESTClient} is obtained that can be used with the
+ * RESTCatalog.
+ */
+public class HTTPClientFactory implements RESTClientFactory {
+
+  @Override
+  public RESTClient apply(Map<String, String> properties) {
+    Preconditions.checkArgument(properties != null, "Invalid initial configuration: null");
```
rdblue, line 36: Remove "initial"? I don't think that makes sense if this is called with a merged configuration after fetching server-side config.

core/src/main/java/org/apache/iceberg/rest/RESTClientFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
+
+import java.util.Map;
+import java.util.function.Function;
+
+public interface RESTClientFactory extends Function<Map<String, String>, RESTClient> {
```
rdblue, line 25: What is the purpose of this interface? I don't really see how this would be used.

core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.apache.iceberg.rest;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class RESTObjectMapper {
```
rdblue, line 27: This leaks Jackson, so it should not be public. Can you make the class and the `mapper` method package-private?

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    Preconditions.checkArgument(properties.containsKey(CatalogProperties.URI), "REST Catalog server URI is required");
+    Preconditions.checkArgument(properties.containsKey(RESTCatalogProperties.AUTH_TOKEN),
+        "Invalid http client configuration. Missing authentication token");
+
+    String baseURI = properties.get(CatalogProperties.URI);
+
+    return HTTPClient.builder()
+        .withBearerAuth(properties.get(RESTCatalogProperties.AUTH_TOKEN))
+        .uri(baseURI)
+        .mapper(RESTObjectMapper.mapper())
```
rdblue, line 46: Can you remove `mapper` from the client builder API? I don't think it is something that we need to expose and it leaks Jackson in the API.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,66 @@
...
+    Preconditions.checkArgument(initialProps.containsKey(RESTCatalogProperties.AUTH_TOKEN),
+        "Invalid http client configuration. Missing authentication token");
+
+    String baseURI = initialProps.get(CatalogProperties.URI).trim();
+
+    HTTPClient singleUseClient = HTTPClient.builder()
+        .uri(baseURI)
+        .mapper(RESTObjectMapper.mapper())
+        .withBearerAuth(initialProps.get(RESTCatalogProperties.AUTH_TOKEN).trim())
+        .build();
```
rdblue, line 124: Why does this duplicate the logic from `HTTPClientFactory`? The intent of having the client builder was to be able to use it in both places.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,66 @@
...
+    initializeInternal(name, mergedProps);
+  }
+
+  /**
+   * Performs the normal initialization steps after calling the config route and merging the initial
+   * configuration with any configuration provided by the server.
+   * <p>
+   * See {@link RESTCatalogConfigResponse} for details on merging the initial client and server configuration.
+   *
+   * @param name  Name of this RESTCatalog
```
rdblue, line 98: I think that arg names should match the variable names. Could you make the case match?

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,66 @@
...
+   * @param props Properties to use, merged with any server-provided configuration.
+   */
+  @VisibleForTesting
+  void initializeInternal(String name, Map<String, String> props) {
```
rdblue, line 102: I think we can avoid adding both of these `@VisibleForTesting` methods. To test, I think you can either update `RESTCatalogAdaptor` to accept a config map to return, or just pass in a different `RESTClient` when creating a test catalog. It could be something roughly like this:

```java
  @Test
  public void testConfigRoute() {
    RESTClient testClient = new RESTClient() {
      @Override
      public void head(String path, Consumer<ErrorResponse> errorHandler) {
        throw new UnsupportedOperationException("Should not be called for testConfigRoute");
      }

      @Override
      public <T extends RESTResponse> T delete(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
        throw new UnsupportedOperationException("Should not be called for testConfigRoute");
      }

      @Override
      public <T extends RESTResponse> T get(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
      }

      @Override
      public <T extends RESTResponse> T post(String path, RESTRequest body, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
        throw new UnsupportedOperationException("Should not be called for testConfigRoute");
      }

      @Override
      public void close() {
      }
    };

    RESTCatalog restCatalog = new RESTCatalog((config) -> testClient);

    restCatalog.setConf(new Configuration());
    restCatalog.initialize("prod", ImmutableMap.of());
    
    Assert.assertNotNull("Should have config key foo", restCatalog.properties().get("foo"));
  }
```

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
kbendick: I was wondering the same thing. I had originally had it as optional.

I'll change it so that it is only applied if the user provides it.

core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java
kbendick: Yes I can.

core/src/main/java/org/apache/iceberg/rest/RESTClientFactory.java
kbendick: Originally, this had more recognizable benefit. I decided to keep it in so that the hierarchy of `REST*` being the interface and then `HTTP*` being the implementing class, like `RESTClient` <-> `HTTPClient`, would be kept.

However, since this is a plain functional interface, I agree it doesn't provide much benefit. Will remove.

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
kbendick: Agreed. Originally this handled the merging, but since the client builder / factory is used in other places, I removed that.

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
kbendick: Sure. Now that we have one, it's not really needed.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Maybe this comment is dated, but if you're referring to the parameter names in the javadoc comment, they do match.

If you're referring to making the parameter named `properties`, I think that will cause a shadowed variable error in error prone since this isn't a constructor. Let me double check.

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
kbendick: Updated to use the auth token only if present.

core/src/main/java/org/apache/iceberg/rest/HTTPClientFactory.java
kbendick: Removed.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Yeah. I noticed that after posting it. Originally what I was working with for the `HTTPClientFactory` had some stuff that made it not possible to share the code. But that's been fixed so things have been deduplicated!

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
rdblue: Oh, looks like my bad. Sorry about that, not sure why I didn't see `name`.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: So this did still require 2 `VisibleForTesting` functions, but one of them is just an accessor for `properties()`. I was able to remove the annotation for the function for merging in server config, as well as the usage of mockito.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Ah no worries. It happens.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -72,4 +78,62 @@ protected boolean supportsNamespaceProperties() {
...
+  public void testInitializeWithBadArguments() {
+    AssertHelpers.assertThrows("Configuration passed to initialize cannot be null",
+        IllegalArgumentException.class,
+        "Invalid configuration: null",
+        () -> catalog().initialize("prod", null));
+
+    AssertHelpers.assertThrows("Configuration passed to initialize must have uri",
+        IllegalArgumentException.class,
+        "REST Catalog server URI is required",
+        () -> catalog().initialize("prod", ImmutableMap.of()));
```
kbendick, line 137: Now that we've refactored to not have such a large surface area in `VisibleForTesting`, I'm going to move these tests over to a test class for `TestHTTPClientFactory`.

They're failing now because I updated `RESTCatalog`'s fetch server configuration function to use the passed in adaptor (via `clientBuilder`).

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -55,7 +61,7 @@ public void createCatalog() {
 
     this.restCatalog = new RESTCatalog((config) -> adaptor);
     restCatalog.setConf(conf);
-    restCatalog.initialize("prod", ImmutableMap.of());
+    restCatalog.initializeInternal("prod", ImmutableMap.of());
```
rdblue, line 64: Why does this call `initializeInternal` still? I think that normal initialization should be used for testing.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -72,4 +78,63 @@ protected boolean supportsNamespaceProperties() {
...
+      }
+    };
+
+    RESTCatalog restCat = new RESTCatalog((config) -> testClient);
+    Map<String, String> initialConfig = ImmutableMap.of("uri", "http://localhost:8080");
+
+    restCat.setConf(new Configuration());
+    restCat.initialize("prod", initialConfig);
+
+    Assert.assertNotNull("Should have sever-provided config keys", restCat.properties().get("cache-enabled"));
```
rdblue, line 124: Typo: `sever` -> `server`.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -72,4 +78,63 @@ protected boolean supportsNamespaceProperties() {
...
+                                               Consumer<ErrorResponse> errorHandler) {
+        throw new UnsupportedOperationException("Should not be called for testConfigRoute");
+      }
+
+      @Override
+      public <T extends RESTResponse> T get(String path, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+        return (T) RESTCatalogConfigResponse
+            .builder()
+            .withDefaults(ImmutableMap.of())
+            .withOverrides(ImmutableMap.of("cache-enabled", "false"))
```
rdblue, line 103: Right now, this test uses disjoint property sets, but I'd like to see it validate the override behavior. I realize that's tested in the config response, but at least a basic check would be good here.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,42 @@
...
-    String ioImpl = props.get(CatalogProperties.FILE_IO_IMPL);
-    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), props, conf);
+    this.catalogName = name;
+    this.paths = ResourcePaths.forCatalogProperties(properties);
+    String ioImpl = properties.get(CatalogProperties.FILE_IO_IMPL);
+    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), properties, conf);
+  }
+
+  @VisibleForTesting
+  Map<String, String> properties() {
```
rdblue, line 112: This could just be public, too. I think it's reasonable for the catalog to expose its config properties.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,42 @@
...
     this.client = clientBuilder.apply(props);
-    this.catalogName = name;
     this.properties = ImmutableMap.copyOf(props);
-    this.paths = ResourcePaths.forCatalogProperties(props);
-    String ioImpl = props.get(CatalogProperties.FILE_IO_IMPL);
-    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), props, conf);
+    this.catalogName = name;
+    this.paths = ResourcePaths.forCatalogProperties(properties);
+    String ioImpl = properties.get(CatalogProperties.FILE_IO_IMPL);
+    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), properties, conf);
```
rdblue, line 108: It looks like these lines don't need to change at all. The current set of changes is to reorder some of them (move catalogName assignment) and rename `props` to `properties`. I think we can avoid making at least some of these changes. But I would merge the methods together first, since that will affect this.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -405,4 +435,23 @@ private String fullTableName(TableIdentifier ident) {
...
+    try {
+      RESTCatalogConfigResponse configResponse = singleUseClient
+          .get(ResourcePaths.config(), RESTCatalogConfigResponse.class, ErrorHandlers.defaultErrorHandler());
+      configResponse.validate();
+      return configResponse.merge(props);
+    } finally {
+      try {
+        singleUseClient.close();
+      } catch (IOException e) {
+        LOG.error("Failed to close http client used for getting catalog configuration. Possible resource leak.", e);
```
rdblue, line 453: `http` should be all caps.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -405,4 +435,23 @@ private String fullTableName(TableIdentifier ident) {
 
     return Pair.of(tableClient, tableIO);
   }
+
+  private Map<String, String> fetchAndMergeWithServerConf(Map<String, String> props) {
```
rdblue, line 439: I think this would be cleaner if it just fetched the config and returned the response. That way it does one only thing. The only call that would be moved outside is the `merge` call, which probably shouldn't be in the `try` block anyway.

Then the caller would look like this:

```java
    ConfigResponse config = fetchConfig(properties);
    Map<String, String> props = config.merge(properties);
```

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: Actually these didn't need to be moved. I will consider it if need be.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: Ok. I wasn't sure how much testing to add here because there are tests over there, but I agree it would be a good idea to have a bit more extensive testing here.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Ok. Given that the properties can change (mostly from the config call), I wasn't sure if we wanted to expose it. But that seems fair.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,42 @@
...
+   * Performs the normal initialization steps after calling the config route and merging the initial
+   * configuration with any configuration provided by the server.
+   * <p>
+   * See {@link RESTCatalogConfigResponse} for details on merging the initial client and server configuration.
+   *
+   * @param name  Name of this RESTCatalog
+   * @param props Properties to use, merged with any server-provided configuration.
+   */
+  @VisibleForTesting
+  void initializeInternal(String name, Map<String, String> props) {
```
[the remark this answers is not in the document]
kbendick: Yeah this was removed.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Yeah I had already moved it to the `try` block, but I will move the merging out so that this function has more of a single purpose.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -42,20 +50,29 @@ public void createCatalog() {
...
-        JdbcCatalog.PROPERTY_PREFIX + "password", "password");
-    backendCatalog.initialize("backend", backendCatalogProperties);
+    this.restCatalog = new RESTCatalog((config) -> {
+      JdbcCatalog backendCatalog = new JdbcCatalog();
+      backendCatalog.setConf(conf);
+      Map<String, String> backendCatalogProperties = ImmutableMap.of(
+          CatalogProperties.WAREHOUSE_LOCATION, warehouse.getAbsolutePath(),
+          CatalogProperties.URI, "jdbc:sqlite:file::memory:?ic" + UUID.randomUUID().toString().replace("-", ""),
+          JdbcCatalog.PROPERTY_PREFIX + "username", "user",
+          JdbcCatalog.PROPERTY_PREFIX + "password", "password");
```
kbendick, line 60: The `backendCatalog` creation needed to be moved into the constructor so that the single use connection that is opened and then closed to fetch the server config doesn't permanently close the underlying JDBC catalog connection.

This way, the lifecycle of the connection to the JDBC catalog mirrors that of the HTTP client.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Updated.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Updated.

core/src/main/java/org/apache/iceberg/rest/RESTCatalogProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class RESTCatalogProperties {
+
+  private RESTCatalogProperties() {
+  }
+
+  /**
+   * A Bearer authorization token which will be used to authenticate requests with the server.
+   */
+  public static final String AUTH_TOKEN = "token";
```
kbendick, line 34: Do we want this to start with `rest.`, similar to the way that `jdbc` catalog prefaces its configs?

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: No longer calls `initializeInternal`.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Updated so that these don't change.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -89,6 +90,11 @@ public CommitFailedException wrapped() {
     }
   }
 
+  public static RESTCatalogConfigResponse fetchServerConfig() {
```
rdblue, line 93: What is the purpose of updating `CatalogHandlers`? Since this doesn't actually do translation to the `Catalog` API, it doesn't seem like something we should provide a helper for. Plus, this doesn't give the caller any opportunity to customize the config.

core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java
```diff
@@ -154,8 +154,7 @@ public <T extends RESTResponse> T handleRequest(Route route, Map<String, String>
                                                   Object body, Class<T> responseType) {
     switch (route) {
       case CONFIG:
-        // TODO: use the correct response object
-        return castResponse(responseType, ImmutableMap.of());
+        return castResponse(responseType, CatalogHandlers.fetchServerConfig());
```
rdblue, line 157: I commented above that it probably isn't very helpful to include `fetchServerConfig()` in `CatalogHandlers`. I would recommend creating the response directly here. You could also include a default config map for testing purposes if you wanted, although I think you're testing it different way so you may not need it.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
rdblue: I'm a bit concerned about this because it creates an entirely new JDBC catalog. It works for tests now, but if we were to send slightly different config back when loading a table, the catalog would create a new connection and wipe the DB.

Could this be solved instead by skipping `catalog.close()` in the `RESTCatalogAdaptor.close` method? We could simply close the `backendCatalog` in an `@AfterEach` method.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -72,4 +90,72 @@ protected boolean supportsNamespaceProperties() {
   protected boolean supportsServerSideRetry() {
     return true;
   }
+
+  /* RESTCatalog specific tests */
+
+  @Test
+  public void testConfigRoute() throws IOException {
```
rdblue, line 97: These tests look good to me.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: Fixed.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: Added more testing for defaults and overrides being applied. Between that and the tests actually in the `RESTCatalogConfigResponse`, things should be adequately tested.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: Yeah that would be fine. I had it that way at one point in time (but never pushed it), so let me get it back that way again.

core/src/main/java/org/apache/iceberg/rest/RESTCatalogProperties.java
kbendick: Yeah I don't think so either. Should be easy to add if need be.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
kbendick: The tests fail if there's not something in place of this call, since we call `initialize` and not the older `initializeInternal`.

So something needs to be returned here.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,12 +75,31 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
...
+  /**
+   * Performs the normal initialization steps after calling the config route and merging the initial
+   * configuration with any configuration provided by the server.
+   * <p>
+   * See {@link RESTCatalogConfigResponse} for details on merging the initial client and server configuration.
+   *
+   * @param name  Name of this RESTCatalog
+   * @param props Properties to use, merged with any server-provided configuration.
+   */
+  private void initializePostConfig(String name, Map<String, String> props) {
```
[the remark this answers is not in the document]
kbendick: My main reasoning for making it a separate function was so that the non-merged properties didn't get reused accidentally. I can merge it into one function if we prefer.

core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java
kbendick: Sure I can inline it here. That makes sense.

And since the server configuration being applied depends on what configuration is passed in, I chose to forgo it for now since it's tested elsewhere.

core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java
```diff
@@ -281,9 +281,9 @@ public void head(String path, Consumer<ErrorResponse> errorHandler) {
 
   @Override
   public void close() throws IOException {
-    if (catalog instanceof Closeable) {
-      ((Closeable) catalog).close();
-    }
+    // The calling test is responsible for closing the underlying catalog backing this REST catalog
+    // so that the underlying backend catalog is not closed and reopened during the REST catalog's
+    // initialize method when fetching the server configuration.
```
rdblue, line 286: Looks good.

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -55,7 +64,19 @@ public void createCatalog() {
...
+  }
+
+  @AfterEach
+  public void closeCatalog() throws IOException {
+    if (restCatalog != null) {
+      restCatalog.close();
+    }
+
+    if (backendCatalog != null) {
+      this.backendCatalog.close();
```
rdblue, line 78: Nit: unnecessary `this.` prefix. We only require it when assigning to an instance field.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -69,18 +75,28 @@ public class RESTCatalog implements Catalog, SupportsNamespaces, Configurable<Co
...
-    this.client = clientBuilder.apply(props);
+    RESTCatalogConfigResponse config = fetchConfig(props);
+    Map<String, String> mergedProps = config.merge(props);
+    this.client = clientBuilder.apply(mergedProps);
     this.catalogName = name;
-    this.properties = ImmutableMap.copyOf(props);
-    this.paths = ResourcePaths.forCatalogProperties(props);
-    String ioImpl = props.get(CatalogProperties.FILE_IO_IMPL);
-    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), props, conf);
+    this.properties = ImmutableMap.copyOf(mergedProps);
```
rdblue, line 92: Nit: `RESTCatalogConfigResponse.merge` returns an `ImmutableMap` already.

rdblue (approved):
+1 once checkstyle is fixed. Here's the error:

```
[ERROR] /home/runner/work/iceberg/iceberg/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java:22:8: Unused import - java.io.Closeable. [UnusedImports]
```

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
kbendick: Good catch.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
kbendick: Removed.

kbendick:
> +1 once checkstyle is fixed. Here's the error:
> 
> ```
> [ERROR] /home/runner/work/iceberg/iceberg/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java:22:8: Unused import - java.io.Closeable. [UnusedImports]
> ```

This has been removed. Once CI completes this should be good to go.

rdblue:
Thanks, @kbendick!

PR #4499, opened 2022-04-05
Core - Fix JsonUtils collection or null functions when an explicit null value is used

kbendick:
When testing the `ErrorResponse`, it came up that `JsonUtils#getStringListOrNull` fails when the node in question (element by name) is explicitly null.

Fixes the issue for both `getXCollectionOrNull` functions in `JsonUtils`.


rdblue:
Good catch, @kbendick! Thanks for fixing this.

PR #4500, opened 2022-04-05
AWS, GCP, Aliyun: Handle null metrics fallback for MetricsContext

rajarshisarkar:
Recently encountered the following exception in https://github.com/apache/iceberg/pull/4489#discussion_r842375076:
```
org/apache/hadoop/fs/FileSystem
java.lang.NoClassDefFoundError: org/apache/hadoop/fs/FileSystem
	at org.apache.iceberg.hadoop.HadoopMetricsContext.initialize(HadoopMetricsContext.java:50)
	at org.apache.iceberg.dell.ecs.EcsFileIO.initialize(EcsFileIO.java:92)
	at org.apache.iceberg.dell.ecs.EcsCatalog.initializeFileIO(EcsCatalog.java:131)
	at org.apache.iceberg.dell.ecs.EcsCatalog.initialize(EcsCatalog.java:108)
	at org.apache.iceberg.dell.ecs.TestEcsCatalog.before(TestEcsCatalog.java:62)
```
	
This was because `NoClassDefFoundError` was not handled in the catch block due to which metrics was not falling back to null metrics in case the exception happens while doing `metrics.initialize(properties)`. 

Reference:

https://github.com/apache/iceberg/blob/7c2ea0128133f630041ecb77de1fae754073e904/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java#L274-L283

This PR handles the null metrics fallback for MetricsContext for the other FileIO's.

---
cc: @rdblue @danielcweeks @kbendick


aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -278,7 +278,8 @@ public void initialize(Map<String, String> properties) {
       this.metrics = ctor.newInstance("s3");
 
       metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
+    } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
+      this.metrics = MetricsContext.nullMetrics();
```
singhpk234, line 282: [question] I believe we have set metrics to nullMetrics already here : 
https://github.com/apache/iceberg/blob/7c2ea0128133f630041ecb77de1fae754073e904/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java#L76

I belive then this would not be required here ? Should we do the same for all remaining filesIO's as well just in case someone tries to access the FileIO obj , without initialization (not sure how practical this case is though), Your thoughts

rajarshisarkar: I don't see any harm in initialising at a class level (might be helpful for some use cases, maybe a test?). I can be convinced otherwise though.

The scenario here is - during runtime metrics is overridden at `this.metrics = ctor.newInstance("scheme");` and then the exception happens in `metrics.initialize(properties);` leading to no fallback.

aliyun/src/main/java/org/apache/iceberg/aliyun/oss/OSSFileIO.java
```diff
@@ -111,7 +111,8 @@ public void initialize(Map<String, String> properties) {
       this.metrics = ctor.newInstance("oss");
 
       metrics.initialize(properties);
-    } catch (NoSuchMethodException | ClassCastException e) {
+    } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
+      this.metrics = MetricsContext.nullMetrics();
```
rdblue, line 115: Isn't this already set? I don't think we need this.

rajarshisarkar: Actually that is being overridden in the current logic. I have added some in-line comments below to highlight the issue:

```
private MetricsContext metrics = MetricsContext.nullMetrics(); // metrics initialised as nullMetrics at class level
...
...
  @Override
...
...
    try {
      this.metrics = ctor.newInstance("ecs"); // metrics re-initialised as new HadoopMetricsContext("ecs")

      metrics.initialize(properties); // throws exception in case of classpath issues 
    } catch (NoClassDefFoundError | NoSuchMethodException | ClassCastException e) {
      // We swallow the exception and allow new HadoopMetricsContext() to flow down to EcsInputFile/EcsOutputFile
    }
  }
```

We can handle the fallback in the catch block (do suggest if there is some better way). Without that the CI seems to fail for commit: [c7e355f](https://github.com/apache/iceberg/pull/4489/commits/c7e355fc0820c5630218ac0e5ae1f125852e210a)

The CI passes if we handle the nullMetrics fallback in the catch block for commit: [2ad9cda](https://github.com/apache/iceberg/pull/4489/commits/2ad9cdae82ed3714ee58f96abfc95daf1b880a58)

rajarshisarkar: Do suggest if there is some better way to handle the fallback. I can implement the same for all the FileIO's in this PR.

rdblue: What about waiting until after `initialize` succeeds to set `this.metrics`? That seems better to me.

rajarshisarkar: If I understand this correctly:

```
...
try {
....
this.metrics = ctor.newInstance("scheme");
} catch() {
...
}
```

rdblue: No, you need to initialize a `MetricsContext` and then set `this.metrics` to the _initialized_ context.

If it were me, I would use add a local variable. That can keep track of the context and then you can update the instance field afterward:

```java
MetricsContext context = ctor.newInstance("scheme");
context.initialize(properties);
this.metrics = context;
```

The local variable helps you avoid modifying `metrics` until after `initialize` is called, but you still use the initialized metrics context instead of a completely new and uninitialized one.

rajarshisarkar: Ah, thanks for the detailed response. Yes, using the local variable makes perfect sense!

rajarshisarkar:
I have made changes to `OSSFileIO`, `S3FileIO`, `EcsFileIO`, and `GCSFileIO`. PR is ready for next round of review.

rdblue (approved):

Issue #4346, opened 2022-03-16
Make DeleteOrphanFiles in Spark reliable


aokolnychyi:
There have been multiple attempts to make our `DeleteOrphanFiles` action more reliable. One such [discussion](https://the-asf.slack.com/archives/CF01LKV9S/p1601663482098100) happened more than a year ago. However, we never reached consensus.

I will try to summarize my current thoughts but I encourage everyone to comment as well.

### Location Generation

There are three location types in Iceberg.

#### Table location

Table locations are either provided by the user or defaulted in `TableOperations`. When defaulting, we currently manipulate raw strings via methods such as `String$format`. That means there is no normalization/validation for root table locations.

#### Metadata

Classes that extend `BaseMetastoreTableOperations` use `metadataFileLocation` to generate a new location for all types of metadata files. Under the hood, it simply uses `String$format` and has no location normalization.

```
private String metadataFileLocation(TableMetadata metadata, String filename) {
  String metadataLocation = metadata.properties().get(TableProperties.WRITE_METADATA_LOCATION);

  if (metadataLocation != null) {
    return String.format("%s/%s", metadataLocation, filename);
  } else {
    return String.format("%s/%s/%s", metadata.location(), METADATA_FOLDER_NAME, filename);
  }
}
```

In `HadoopTableOperations`, we rely on `Path` instead of `String$format` as we have access to Hadoop classes. 

```
private Path metadataPath(String filename) {
  return new Path(metadataRoot(), filename);
}

private Path metadataRoot() {
  return new Path(location, "metadata");
}
```

That means some normalization is happening for metadata file locations generated in `HadoopTableOperations`.

#### Data

Data file locations depend on `LocationProvider` returned by `TableOperations`. While users can inject a custom location provider, Iceberg has two built-in implementations:

- DefaultLocationProvider
- ObjectStoreLocationProvider

Both built-in implementations use `String$format` and have no normalization/validation.

### Problem

Right now, `DeleteOrphanFiles` uses Hadoop `FileSystem` to list all actual files in a given location and compares them to the locations stored in the metadata. As discussed above, Iceberg does not do any normalization for locations persisted in the metadata. That means locations retuned during listing may have cosmetic differences compared to locations stored in the metadata, even though both can point to the same files. As a consequence,  `DeleteOrphanFiles` can corrupt a table.

### Proposed Approach

[...]

PR #4503, opened 2022-04-05
Spark 3.2: Add `compareWithFileList` param to DeleteOrphanFiles action

bijanhoule:
This allows the user to specify an existing table of actual files for the `DeleteOrphanFiles` action, skipping the directory listing.

This can be used in cases where a cloud storage provider-specific implementation might be more efficient than a recursive FileSystem listing (e.g. leveraging [S3 Inventory list](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory.html) for S3 to create a temp table).

### DIFF
--- a/api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
+++ b/api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
@@ -81,6 +81,19 @@ public interface DeleteOrphanFiles extends Action<DeleteOrphanFiles, DeleteOrpha
   DeleteOrphanFiles executeDeleteWith(ExecutorService executorService);
 
   /**
+   * Passes a table which contains the list of actual files in the table. This skips the directory listing - any
+   * files in the actualFilesTable provided which are not found in table metadata will be deleted. Not compatible
+   * with `location` or `older_than` arguments - this assumes that the provided table of actual files has been
+   * filtered down to the table’s location and only includes files older than a reasonable retention interval.
+   *
+   * @param tableName the table containing the actual files dataset.  Should have a single `file_path` string column
+   * @return this for method chaining
+   */
+  default DeleteOrphanFiles actualFilesTable(String tableName) {
+    throw new UnsupportedOperationException(this.getClass().getName() + " does not implement actualFilesTable");
+  }
+
+  /**
    * The action result that contains a summary of the execution.
    */
   interface Result {
--- a/api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
+++ b/api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
@@ -81,19 +81,6 @@ public interface DeleteOrphanFiles extends Action<DeleteOrphanFiles, DeleteOrpha
   DeleteOrphanFiles executeDeleteWith(ExecutorService executorService);
 
   /**
-   * Passes a table which contains the list of actual files in the table. This skips the directory listing - any
-   * files in the actualFilesTable provided which are not found in table metadata will be deleted. Not compatible
-   * with `location` or `older_than` arguments - this assumes that the provided table of actual files has been
-   * filtered down to the table’s location and only includes files older than a reasonable retention interval.
-   *
-   * @param tableName the table containing the actual files dataset.  Should have a single `file_path` string column
-   * @return this for method chaining
-   */
-  default DeleteOrphanFiles actualFilesTable(String tableName) {
-    throw new UnsupportedOperationException(this.getClass().getName() + " does not implement actualFilesTable");
-  }
-
-  /**
    * The action result that contains a summary of the execution.
    */
   interface Result {
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
@@ -53,6 +53,8 @@ import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.expressions.UserDefinedFunction;
 import org.apache.spark.sql.functions;
 import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
 import org.apache.spark.util.SerializableConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -69,7 +71,14 @@ import static org.apache.iceberg.TableProperties.GC_ENABLED_DEFAULT;
  * removes unreachable files that are older than 3 days using {@link Table#io()}. The behavior can be modified
  * by passing a custom location to {@link #location} and a custom timestamp to {@link #olderThan(long)}.
  * For example, someone might point this action to the data folder to clean up only orphan data files.
- * In addition, there is a way to configure an alternative delete method via {@link #deleteWith(Consumer)}.
+ * <p>
+ * Configure an alternative delete method using {@link #deleteWith(Consumer)}.
+ * <p>
+ * For full control of the set of files being evaluated, use the {@link #actualFilesTable(String)} argument.  This
+ * skips the directory listing - any files in the actualFilesTable provided which are not found in table metadata will
+ * be deleted.
+ * Not compatible with `location` or `older_than` arguments - this assumes that the provided table of actual files
+ * has been filtered down to the table’s location and only includes files older than a reasonable retention interval.
  * <p>
  * <em>Note:</em> It is dangerous to call this action with a short retention interval as it might corrupt
  * the state of the table if another operation is writing at the same time.
@@ -99,6 +108,7 @@ public class BaseDeleteOrphanFilesSparkAction
 
   private String location = null;
   private long olderThanTimestamp = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(3);
+  private Dataset<Row> providedActualFilesDF;
   private Consumer<String> deleteFunc = defaultDelete;
   private ExecutorService deleteExecutorService = null;
 
@@ -145,6 +155,27 @@ public class BaseDeleteOrphanFilesSparkAction
   }
 
   @Override
+  public BaseDeleteOrphanFilesSparkAction actualFilesTable(String tableName) {
+    ValidationException.check(
+        spark().catalog().tableExists(tableName),
+        "actualFilesTable `" + tableName + "` does not exist");
+
+    try {
+      StructType schema = spark().table(tableName).schema();
+      StructField filePathField = schema.apply("file_path");
+      ValidationException.check(
+          filePathField.dataType() == DataTypes.StringType,
+          "Invalid schema for actual files table - 'file_path' column is not a string type");
+    } catch (IllegalArgumentException e) {
+      throw new ValidationException(
+          "actualFilesTable `" + tableName + "` is missing required `file_path` column");
+    }
+
+    this.providedActualFilesDF = spark().table(tableName);
+    return this;
+  }
+
+  @Override
   public DeleteOrphanFiles.Result execute() {
     JobGroupInfo info = newJobGroupInfo("DELETE-ORPHAN-FILES", jobDesc());
     return withJobGroupInfo(info, this::doExecute);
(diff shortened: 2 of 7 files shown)


api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
```diff
@@ -80,6 +80,19 @@
    */
   DeleteOrphanFiles executeDeleteWith(ExecutorService executorService);
 
+  /**
+   * Passes a table which contains the list of actual files in the table. This skips the directory listing - any
+   * files in the actualFilesTable provided which are not found in table metadata will be deleted. Not compatible
+   * with `location` or `older_than` arguments - this assumes that the provided table of actual files has been
```
rdblue, line 86: What if the table had a timestamp? For example, if the table has a `modified_time` then we could apply the `older_than` filter fairly easily rather than making the user handle it.

api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
```diff
@@ -80,6 +80,19 @@
...
+   * filtered down to the table’s location and only includes files older than a reasonable retention interval.
+   *
+   * @param tableName the table containing the actual files dataset.  Should have a single `file_path` string column
+   * @return this for method chaining
+   */
+  default DeleteOrphanFiles actualFilesTable(String tableName) {
```
rdblue, line 92: I think it is a good idea to be able to use a temporary table for the stored procedure, but for the action API I would expect to be able to pass in a `Dataset` of some kind. `Dataset<String>` would work, and we could also expose a couple of Java classes to make `Dataset<FileLocationAndModifiedTime>` and `Dataset<FileLocation>` work.

I think that we can also make what's happening here a bit more clear in the API by naming the method something like `compareToFileList(Dataset<String>)`:

```java
Dataset<String> files = ...
SparkActions.get()
    .deleteOrphanFiles(catalog.loadTable(identifier))
    .compareToFileList(files)
    .execute();
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RemoveOrphanFilesProcedure.java
```diff
@@ -49,7 +49,8 @@
       ProcedureParameter.optional("older_than", DataTypes.TimestampType),
       ProcedureParameter.optional("location", DataTypes.StringType),
       ProcedureParameter.optional("dry_run", DataTypes.BooleanType),
-      ProcedureParameter.optional("max_concurrent_deletes", DataTypes.IntegerType)
+      ProcedureParameter.optional("max_concurrent_deletes", DataTypes.IntegerType),
+      ProcedureParameter.optional("actual_files_table", DataTypes.StringType)
```
rdblue, line 53: What about renaming this to `file_list_view` or `listed_files_view`? I think that we want some identifier that matches the name of the action method; that's why I'm thinking `file_list` or `listed_files` that are similar to `compareToFileList`. But we also want to make it clear that this is a table or temporary view. I think it's more likely that the DF is registered as a temp view rather than stored in a table, so I'd probably use the suffix `_view`.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -150,10 +150,20 @@ public void testDryRun() throws IOException, InterruptedException {
     Assert.assertEquals("Action should find 1 file", invalidFiles, result2.orphanFileLocations());
     Assert.assertTrue("Invalid file should be present", fs.exists(new Path(invalidFiles.get(0))));
 
+    String actualFilesTableName = "actualFilesTable";
+    spark.createDataset(allFiles, Encoders.STRING()).toDF("file_path").createOrReplaceTempView(actualFilesTableName);
```
rdblue, line 154: I think that this test case should run _after_ the actual orphan file delete. Otherwise, listing produces the same result and we're not actually verifying that we get a different result from the new option.

rdblue:
Thanks, @bijanhoule! This looks close to being ready, but I think we should make a few minor modifications to make the actions API a bit more intuitive. I would expect that to accept a `Dataset` rather than requiring temp table registration.

aokolnychyi:
I'd be curious to take a look as well tomorrow.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -150,10 +150,20 @@ public void testDryRun() throws IOException, InterruptedException {
...
     DeleteOrphanFiles.Result result3 = actions.deleteOrphanFiles(table)
+        .deleteWith(s -> { })
+        .actualFilesTable(actualFilesTableName)
+        .execute();
+
+    Assert.assertEquals("Action should find 1 file", invalidFiles, result3.orphanFileLocations());
+    Assert.assertTrue("Invalid file should be present", fs.exists(new Path(invalidFiles.get(0))));
+
+    DeleteOrphanFiles.Result result4 = actions.deleteOrphanFiles(table)
```
kbendick, line 163: If we add support for things like `older_than`, we should probably make a separate test case for the actual files table option so we can exercise various scenarios focused on that one area.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RemoveOrphanFilesProcedure.java
```diff
@@ -80,16 +81,23 @@ public StructType outputType() {
...
     boolean dryRun = args.isNullAt(3) ? false : args.getBoolean(3);
     Integer maxConcurrentDeletes = args.isNullAt(4) ? null : args.getInt(4);
+    String tableWithActualFilePaths = args.isNullAt(5) ? null : args.getString(5);
 
     Preconditions.checkArgument(maxConcurrentDeletes == null || maxConcurrentDeletes > 0,
             "max_concurrent_deletes should have value > 0,  value: " + maxConcurrentDeletes);
 
+    Preconditions.checkArgument(
+            tableWithActualFilePaths == null || (location == null && olderThanMillis == null),
+            "actual_files_table cannot be used with `location` or `older_than`"
```
kbendick, line 98: Nit: I think this is over-indented. Code continued on the next line should be indented 4 spaces over from its starting point.

That said, it does line up with the argument on the line above it.

api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
aokolnychyi: Adding `modified_time` sounds like a good idea as ignoring `older_than` is dangerous. Files just written but not yet committed could be cleaned up, corrupting the table.

api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
aokolnychyi: I guess the problem is that the Action API is engine agnostic so we cannot refer to `Dataset`. We considered offering engine-specific actions at some point but that idea was discarded.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -163,7 +194,7 @@ private String jobDesc() {
     Dataset<Row> validContentFileDF = buildValidContentFileDF(table);
     Dataset<Row> validMetadataFileDF = buildValidMetadataFileDF(table);
     Dataset<Row> validFileDF = validContentFileDF.union(validMetadataFileDF);
-    Dataset<Row> actualFileDF = buildActualFileDF();
+    Dataset<Row> actualFileDF = this.providedActualFilesDF == null ? buildActualFileDF() : providedActualFilesDF;
```
aokolnychyi, line 197: nit: redundant `this`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RemoveOrphanFilesProcedure.java
aokolnychyi: +1, I think it uses 8 instead of 4. The line above is a mistake too.

api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
aokolnychyi: I would even consider making it required. How many people will filter the locations manually before calling the action?

api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
rdblue: I think we should move this into the Spark-specific API then.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
rdblue: Agreed. We usually want to add separate test cases anyway.

bijanhoule:
Thanks for the feedback all!  I'll take another pass and adjust a few things:
- move to the spark-specific api and use a dataset input rather than temp table
- add support for `modified_time` and enforce `older_than`.  Originally the thought was to keep this out for simplicity, but I agree it's probably safer to just make it a requirement.
- separate out test cases and add coverage for the above

api/src/main/java/org/apache/iceberg/actions/DeleteOrphanFiles.java
aokolnychyi: @rdblue, you mean exposing this method in `BaseDeleteOrphanFilesSparkAction` similarly to `expire` in `BaseExpireSnapshotsSparkAction` or introduce a Spark-specific interface?

It is a bummer as this functionality that can be useful for all engines.

rdblue: We may be able to do this later, but I think the most natural way to do this for the Actions API is to use a Dataset. We can take another look at this later, too. Maybe we should parameterize this method so that you can specify what collection you want. Flink could use DataStream and Spark could use Dataset.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -144,6 +154,37 @@ public BaseDeleteOrphanFilesSparkAction deleteWith(Consumer<String> newDeleteFun
...
+
+    this.compareToFileList = files;
+    return this;
+  }
+
+  private Dataset<Row> filteredCompareToFileList() {
+    Dataset<Row> files = compareToFileList;
+    if (location != null) {
+      files = files.filter(files.col(FILE_PATH).startsWith(location));
+    }
```
bijanhoule, line 182: @aokolnychyi -- re: https://github.com/apache/iceberg/pull/4503#discussion_r846544708 does this `startsWith` location filtering seem reasonable to you for a short-term solution?  I know there's some discussion about making some larger changes [here](https://github.com/apache/iceberg/issues/4346).

bijanhoule:
FYI, updated this PR based on feedback above:
- action: rename `actualFilesTable` -> `compareWithFileList`
- procedure: `rename actual_files_table` -> `file_list_view`
- interface change:  remove from `DeleteOrphanFiles`, add to `BaseDeleteOrphanFilesSparkAction`
- apply location and last_modified filtering to compareWithFileList
- refactor test cases

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRemoveOrphanFilesAction.java
```diff
@@ -703,4 +705,120 @@ public void testGarbageCollectionDisabled() {
...
+    List<FilePathLastModifiedRecord> invalidFiles = Lists.newArrayList(allFiles);
+    invalidFiles.removeAll(validFiles);
+    List<String> invalidFilePaths =
+        invalidFiles.stream()
+            .map(FilePathLastModifiedRecord::getFilePath)
+            .collect(Collectors.toList());
+    Assert.assertEquals("Should be 1 invalid file", 1, invalidFiles.size());
+
+    // sleep for 1 second to ensure files will be old enough
+    Thread.sleep(1000);
```
rdblue, line 758: This should use `waitUntilAfter` rather than sleeping.

rdblue (approved):

rdblue:
This looks great. Thanks for adding this, @bijanhoule!

PR #4505, opened 2022-04-06
Python: Add Namespace class

samredai:
This adds the `Namespace` class in `catalog/base.py`.

```py
from iceberg.catalog.base import Namespace

namespace = Namespace("foo", "bar", "baz")
print(namespace)  # 'foo.bar.baz'
len(namespace) # 3
hash(namespace)  # 1445470134
namespace.level(2)  # 'baz'
namespace == namespace  # True
namespace.empty()  # False
```

### DIFF
--- a/python/src/iceberg/catalog/base.py
+++ b/python/src/iceberg/catalog/base.py
@@ -1,0 +1,88 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from typing import Tuple
+
+import mmh3  # type: ignore
+
+
+class Namespace:
+    """A namespace in a Catalog
+
+    Example:
+        >>> from iceberg.catalog.base import Namespace
+        >>> namespace = Namespace("foo", "bar", "baz")
+        >>> print(namespace)
+        foo.bar.baz
+    """
+
+    def __init__(self, *levels: str):
+        self._levels = levels
+        for level in levels:
+            if not level:
+                raise ValueError("Cannot create a namespace with a None level")
+            if b"\x00" in bytes(level.encode("utf-8")):
+                raise ValueError(f"Cannot create a namespace with the null-byte character: {level}")
+
+    def __str__(self) -> str:
+        """Each level in the namespace is combined into a dot-separatd string"""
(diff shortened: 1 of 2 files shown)


samredai:
cc: @dramaticlly @rdblue @jun-he @CircArgs

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,88 @@
...
+
+import mmh3  # type: ignore
+
+
+class Namespace:
+    """A namespace in a Catalog
+
+    Example:
+        >>> from iceberg.catalog.base import Namespace
+        >>> namespace = Namespace("foo", "bar", "baz")
```
dramaticlly, line 28: just curious about, for namespace, are we anticipating something like `iceberg.datdabase.table` and potentially the 4th level with metadata tables such as `partitions` or `files`? or is it being used different?

Trying to see if we shall build some assumptions into length of the namespace, too long or too short wont make any sence?

python/tests/catalog/test_catalog_base.py
```diff
@@ -0,0 +1,131 @@
...
+    namespace = base.Namespace("foo", "bar", "baz")
+
+    assert namespace.levels == ("foo", "bar", "baz")
+    assert namespace.empty() is False
+    assert namespace.level(0) == "foo"
+    assert namespace.level(1) == "bar"
+    assert namespace.level(2) == "baz"
+    assert hash(namespace) == 1445470134
+    assert len(namespace) == 3
+    assert str(namespace) == "foo.bar.baz"
```
dramaticlly, line 33: nit: shall we have one for `assert repr(namespace)` as well? 

I do see there's `test_namespace_repr ` below but we also had`test_namespace_hash` and we tested hash here :)

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,88 @@
...
+        >>> print(namespace)
+        foo.bar.baz
+    """
+
+    def __init__(self, *levels: str):
+        self._levels = levels
+        for level in levels:
```
dramaticlly, line 35: nit: shall we switch to have validation run before the assignment ?

dramaticlly (commented):
Thank you Sam!

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,88 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from typing import Tuple
+
+import mmh3  # type: ignore
+
+
+class Namespace:
```
rdblue, line 23: I'm wondering if we want this in Python. If I had it to do over, I would not put this class in the Java implementation and would use a list of strings for `TableIdentifier` and `Namespace` instead. The problem is that this doesn't do anything useful, but users have to create instances and import it.

I suggest we use a tuple or list of strings instead.

samredai: That makes total sense to me! I was on the fence as well but I figured it was an easy enough PR to open. But yeah I agree that where this is used should just take an `Iterable[str]` and we could always have a `validate_namespace` and `validate_table_identifier` utility function.

python/src/iceberg/catalog/base.py
samredai: This is a good question. My opinion is that it's ok to nest the source code files as much as logically makes sense and makes navigating the code base easier. We always have an option of using __all__ to bubble items up to the sub-module level which I think we should.

PR #4507, opened 2022-04-06
Core: Skipping manifest clean-up for all Error or Exception.

southernriver:
In our production environment, we found operation committed success but Spark Job throw `OutOfMemoryError` exception,  some data files were deleted, we could not read current-snapshot  due to FileNotFoundException.
The manifest clean-up may throw Exception or Error, and we should also catch RuntimeError like `OutOfMemoryError`, otherwise,  it would lead to abort and delete datafile.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/SnapshotProducer.java
+++ b/core/src/main/java/org/apache/iceberg/SnapshotProducer.java
@@ -356,8 +356,8 @@ abstract class SnapshotProducer<ThisT> implements SnapshotUpdate<ThisT> {
       if (event != null) {
         Listeners.notifyAll(event);
       }
-    } catch (RuntimeException e) {
-      LOG.warn("Failed to notify listeners", e);
+    } catch (Throwable t) {
+      LOG.warn("Failed to notify listeners", t);
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/SnapshotProducer.java
+++ b/core/src/main/java/org/apache/iceberg/SnapshotProducer.java
@@ -343,8 +343,8 @@ abstract class SnapshotProducer<ThisT> implements SnapshotUpdate<ThisT> {
         LOG.warn("Failed to load committed snapshot, skipping manifest clean-up");
       }
 
-    } catch (RuntimeException e) {
-      LOG.warn("Failed to load committed table metadata, skipping manifest clean-up", e);
+    } catch (Throwable t) {
+      LOG.warn("Failed to load committed table metadata, skipping manifest clean-up", t);
     }
 
     notifyListeners();


southernriver:
@rdblue  @openinx  Could you please have a look at this, it's a serious bug.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -343,8 +343,8 @@ public void commit() {
         LOG.warn("Failed to load committed snapshot, skipping manifest clean-up");
       }
 
-    } catch (RuntimeException e) {
-      LOG.warn("Failed to load committed table metadata, skipping manifest clean-up", e);
+    } catch (Throwable t) {
```
kbendick, line 346: My understanding has always been that we should only catch non-fatal exceptions.

Is there anyway we can slim down the scope / amount of code that this larger `try-catch` block covers?

kbendick (commented):
Thanks for reporting this finding @southernriver!

I have a few initial questions. Hoping that you can help clarify my understanding of some things. Mainly, I'm hoping to get a better understanding of the situation you had.

Can you provide a more detailed description of the state of your table after the Spark job OOM'd? For example, do you have a stack trace, or possibly more details about what files were missing when you attempted to read from the table again?

Also, is there a way we can test this?

It would help me to be sure that we're fixing the issue in the right place and that we won't have a regression if we could come up with a unit test that simulates the OOM somehow - something that fails before applying this patch, that would be really helpful.

southernriver:
@kbendick  Thank you. Manifest clean-up operation could throw `OutOfMemoryError` , For Spark-Side,  Job  callback to  SparkWrite.abort:

```
  private void abort(WriterCommitMessage[] messages) {
    Map<String, String> props = table.properties();
    Tasks.foreach(files(messages))
        .retry(PropertyUtil.propertyAsInt(props, COMMIT_NUM_RETRIES, COMMIT_NUM_RETRIES_DEFAULT))
        .exponentialBackoff(
            PropertyUtil.propertyAsInt(props, COMMIT_MIN_RETRY_WAIT_MS, COMMIT_MIN_RETRY_WAIT_MS_DEFAULT),
            PropertyUtil.propertyAsInt(props, COMMIT_MAX_RETRY_WAIT_MS, COMMIT_MAX_RETRY_WAIT_MS_DEFAULT),
            PropertyUtil.propertyAsInt(props, COMMIT_TOTAL_RETRY_TIME_MS, COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT),
            2.0 /* exponential */)
        .throwFailureWhenFinished()
        .run(file -> {
          table.io().deleteFile(file.path().toString());
        });
  }

```
It would delete datafile. And then if we read from current snapshot,  `java.io.FileNotFoundException` would be thrown:
image

it's easy to reproduce this problem, you can  just throw `OutOfMemoryError` in try block.
`      throw  new OutOfMemoryError("oom");
`


BTW, I think clean-up is not necessary after committed success, and we can catch all exception or error.

rdblue:
I'm not sure how the changes in this PR would help, since it just makes a warning more broad. The catch clauses that are updated here don't actually change what happens other than to log a warning in unrelated cases.

I'm also skeptical that this is the right thing to do. The chance of an OOM causing a commit to appear like it did not succeed when it actually did is incredibly small... and besides, cleanup is run when a `RuntimeException` is caught. That means that an Error wouldn't actually cause the commit path to remove live data files.

@southernriver, It sounds like the problem that you actually hit was with Spark's error handling and cleanup instead? Maybe we should change the Spark writer instead to avoid triggering its abort path in certain cases. That's a very different change.

southernriver:
> I'm not sure how the changes in this PR would help, since it just makes a warning more broad. The catch clauses that are updated here don't actually change what happens other than to log a warning in unrelated cases.
> 
> I'm also skeptical that this is the right thing to do. The chance of an OOM causing a commit to appear like it did not succeed when it actually did is incredibly small... and besides, cleanup is run when a `RuntimeException` is caught. That means that an Error wouldn't actually cause the commit path to remove live data files.
> 
> @southernriver, It sounds like the problem that you actually hit was with Spark's error handling and cleanup instead? Maybe we should change the Spark writer instead to avoid triggering its abort path in certain cases. That's a very different change.

@rdblue Thanks for your comment.  Iceberg itself implements the SparkWrite#abort method and could delete datafile with uncatched `java.lang.Error` or `java.lang.Exception`.  ` RuntimeException` can not cover`OutOfMemoryError` , so I did catch all `java.lang.Throwable` to avoid clean-up live data files.

PR #4512, opened 2022-04-06
Core: Optimise BaseFileScanTask memory usage

rbalamohan:
BaseFileScanTask has a reference to "ResidualEvaluator" and is used only in "residual()" function. However, there can be corner cases where upstream has added large number of literals to residual evaluator (e.g in one of the cases Hive added 79000 LongLiterals in residual evaluator). When these tasks are serialized, they cause memory pressure. Intent is to avoid this, by computing the expression right upfront, and not retaining "ResidualEvalutor" in BaseFileScanTask.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
@@ -34,7 +34,7 @@ public class BaseFileScanTask implements FileScanTask {
   private final DeleteFile[] deletes;
   private final String schemaString;
   private final String specString;
-  private final ResidualEvaluator residuals;
+  private final Expression expr;
 
   private transient PartitionSpec spec = null;
 
@@ -44,7 +44,7 @@ public class BaseFileScanTask implements FileScanTask {
     this.deletes = deletes != null ? deletes : new DeleteFile[0];
     this.schemaString = schemaString;
     this.specString = specString;
-    this.residuals = residuals;
+    this.expr = (residuals == null) ? null : residuals.residualFor(file.partition());
   }
 
   @Override
@@ -77,7 +77,7 @@ public class BaseFileScanTask implements FileScanTask {
 
   @Override
   public Expression residual() {
-    return residuals.residualFor(file.partition());
+    return expr;
   }
 
   @Override


pvary:
@rbalamohan:
Do I understand correctly that the maximum size of `expr` is the same as the `residuals`, and it could be smaller if the partition values would prevent specific parts of the expression to change the results of the expression? Like if the `residuals` would say `date<2022 Jan 15` and the file contains `year(date)=2021`?

Is there any edge case when the `expr` is bigger than the `residuals`?

What is the CPU cost of calculating the `expr`? @rdblue: Are there any places in the Iceberg code where we create the filter but do not call the `residuals()` method?

rbalamohan:
Residuals are evaluated earlier to create the "expression" upfront. Computed expression (string) is much smaller than residual memory as it need not retain the literals. 

In the case I mentioned earlier, it retained 79000 LongLiterals which caused memory spikes as this was serialized in every split in Hive. With this change, expression is computed upfront and residuals need not be carried forward and thus reduces memory pressure. In the case you mentioned, expression gets evaluated to "true/false" upfront and thus reduces memory pressure.

pvary:
Thanks @rbalamohan, this is very promising!

@rdblue: Any concerns?

pvary (approved):

pvary:
I will push this change on Monday if nobody raises concerns till then.

rdblue:
I don't think that this is the right approach.

In most cases, the residual evaluator is small because the initial expression is small. Plus, the residual evaluator is reused for all of the tasks, so there is only one in memory. Eagerly evaluating the residual could create a huge number of expressions in memory, one for each partition at a minimum and probably on per file split.

I also don't think this would help very much because if the large expression with thousands of literal values is not eliminated by the residual evaluator, it would be retained in the resulting expressions.

Overall, I think that the right balance is to keep the residual evaluator here. There will be some cases where that is expensive, but much better in the average case. The cases where it is expensive are cases where the large expression is already in memory anyway.

pvary:
Thanks @rdblue for the review!
Will have to find another solution then. We need to serialize the tasks and send them to the executors. When we have this big residual, then we have to serialize it for every task which is a serious unnecessary overhead, especially when the resulting expression is small.

@szlta suggested that - in case the residuals are the same for every task in a table scan - we could serialize the residuals independently and send them by a different / table scan level method, and before executing the task we can restore the residuals.

pvary:
@rdblue: Definitely! Let us check the possibilities and come back with a suggestion.

rbalamohan:
@rdblue , @pvary : We may have to look at other possibilities. One potential issue in adding it to table property is that, we may not know the query pattern of end users. So it would be difficult to specify it in table property.

rdblue:
So is the idea to ignore residuals and to run the original filter on every split? That seems reasonable to me.

Issue #4513, opened 2022-04-06
row delta delete files were dropped in sequential commits after table format updated to v2


vanliu-tx:
I have one v1 table with some data files, after updating it to v2 and added some row delta deletes, it works fine.

But when I delete some data files to expire data, the row delta delete-files were dropped somehow. This doesn't happen in the first commit, it happens in the second commit as the first commit wrote wrong min data sequence number in manifest.

vanliu-tx:
fixed

Issue #33, opened 2018-12-07
Add Avro support to Pig reader


rdblue:
The initial implementation for pig only supports parquet. We need an avro read path.

This issue was migrated from https://github.com/Netflix/iceberg/issues/51

Fokko:
I can pick this up

PR #4514, opened 2022-04-06
Core: row delta delete files were dropped in sequential commits af…

vanliu-tx:
…ter table format updated to v2 #33

Closing #4513

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
+++ b/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
@@ -690,7 +690,7 @@ abstract class MergingSnapshotProducer<ThisT> extends SnapshotProducer<ThisT> {
         base.schema(), current != null ? current.dataManifests() : null);
     long minDataSequenceNumber = filtered.stream()
         .map(ManifestFile::minSequenceNumber)
-        .filter(seq -> seq > 0) // filter out unassigned sequence numbers in rewritten manifests
+        .filter(seq -> seq >= 0) // filter out unassigned sequence numbers in rewritten manifests
         .reduce(base.lastSequenceNumber(), Math::min);
     deleteFilterManager.dropDeleteFilesOlderThan(minDataSequenceNumber);
     List<ManifestFile> filteredDeletes = deleteFilterManager.filterManifests(
--- a/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
+++ b/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
@@ -690,7 +690,7 @@ abstract class MergingSnapshotProducer<ThisT> extends SnapshotProducer<ThisT> {
         base.schema(), current != null ? current.dataManifests() : null);
     long minDataSequenceNumber = filtered.stream()
         .map(ManifestFile::minSequenceNumber)
-        .filter(seq -> seq != ManifestWriter.UNASSIGNED_SEQ) // filter out unassigned sequence numbers in rewritten manifests
+        .filter(seq -> seq != ManifestWriter.UNASSIGNED_SEQ) // filter out unassigned in rewritten manifests
         .reduce(base.lastSequenceNumber(), Math::min);
     deleteFilterManager.dropDeleteFilesOlderThan(minDataSequenceNumber);
     List<ManifestFile> filteredDeletes = deleteFilterManager.filterManifests(
--- a/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
+++ b/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
@@ -690,7 +690,7 @@ abstract class MergingSnapshotProducer<ThisT> extends SnapshotProducer<ThisT> {
         base.schema(), current != null ? current.dataManifests() : null);
     long minDataSequenceNumber = filtered.stream()
         .map(ManifestFile::minSequenceNumber)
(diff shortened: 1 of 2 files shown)


vanliu-tx:
> Thank you for this contribution @vanliu-tx! This is a great find.
> 
> The test is quite extensive. I left a few nits, but mostly I'm wondering if there's any kind of simpler way to test for this, like checking the sequence number of planned files after applying a delete.
> 
> It might not be something that can be made much simpler though without refactoring the code more.
> 
> cc @rdblue @RussellSpitzer @szehon-ho @jackye1995 you'll likely be interested in this.

added some verification for manifest sequence number and min sequence number.

vanliu-tx:
@rdblue @RussellSpitzer @aokolnychyi  could you help to merge this bug fix?

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
```diff
@@ -690,7 +690,7 @@ public List<ManifestFile> apply(TableMetadata base) {
         base.schema(), current != null ? current.dataManifests() : null);
     long minDataSequenceNumber = filtered.stream()
         .map(ManifestFile::minSequenceNumber)
-        .filter(seq -> seq > 0) // filter out unassigned sequence numbers in rewritten manifests
+        .filter(seq -> seq >= 0) // filter out unassigned sequence numbers in rewritten manifests
```
rdblue, line 693: @vanliu-tx, I don't see how this would matter. Sequence number 0 is reserved for v1 files, so when you convert to v2 all of the v1 files have sequence number 0. And in v1, there are no delete files. So reporting a min sequence number of 1 and removing all delete files older than 1 should have no change.

core/src/test/java/org/apache/iceberg/TestV1ToV2RowDeltaDelete.java
```diff
@@ -0,0 +1,240 @@
...
+    Assert.assertEquals("Should have one data manifest file",
+        1, dataManifests.size());
+    Assert.assertEquals("Should have one delete manifest file",
+        1, deleteManifests.size());
+    Assert.assertEquals(dataManifest, dataManifests.get(0)); // data manifest not changed
+    ManifestFile deleteManifest = deleteManifests.get(0);
+    verifyManifestSequenceNumber(deleteManifest, 1, 1);
+    List<FileScanTask> tasks = Lists.newArrayList(table.newScan().planFiles().iterator());
+    Assert.assertEquals("Should have three task", 3, tasks.size());
+    FileScanTask task = tasks.get(0);
```
rdblue, line 98: This should not assume that the task for FILE_A is produced first. Instead, this should find the task by path.

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
rdblue: The check in `ManifestFilterManager` that uses this value is: `(isDelete && entry.sequenceNumber() > 0 && entry.sequenceNumber() < minSequenceNumber)`

We know that `entry.sequenceNumber()` is non-zero. It shouldn't matter whether `minDataSequenceNumber` is 0 or 1.

vanliu-tx: Assuming the following steps
1. create v1 table and append some data files.
2. upgrade to v2 and commit.
3. append some row-delta delete and commit.
4. remove one data file and commit.
5. remove another data file and commit.

```
    long minDataSequenceNumber = filtered.stream()
        .map(ManifestFile::minSequenceNumber)
        .filter(seq -> seq > 0) // filter out unassigned sequence numbers in rewritten manifests
        .reduce(base.lastSequenceNumber(), Math::min);
```
Because all v1 data files has sequence number 0, the minDataSequenceNumber will use base.lastSequenceNumber() as it's value. In step 4 this value is 1, deleteFilterManager.dropDeleteFilesOlderThan(minDataSequenceNumber) will not drop delete-files in step 3. But in step 5, this value turns to 2, which will drop delete-files in step 3.

core/src/test/java/org/apache/iceberg/TestV1ToV2RowDeltaDelete.java
vanliu-tx: will fix this.

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
rdblue: Okay, I understand. The `base.lastSequenceNumber()` is what causes this. Thanks!

rdblue, line 693: Is this `filter` call even needed? If we need to keep results with sequence number 0, then it seems like we should just remove it. Everything should match that filter.

vanliu-tx: some manifest file's minSequenceNumber is -1, if we remove this line, the minDataSequenceNumber will be -1 in some cases, and fail the validation in

vanliu-tx: maybe we can change this line to

rdblue: That sounds good to me.

rdblue (approved):

rdblue:
Thanks for fixing this, @vanliu-tx!

Issue #4292, opened 2022-03-08
PartitionsTable doesn't take old specs into consideration


szlta:
Currently PartitionsTable will be constructed with a schema based on the latest table spec only:
https://github.com/apache/iceberg/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Ficeberg%2FPartitionsTable.java#L48
Especially with V2, this can result in funny partition table prints when we drop a partition field and create a different one.
E.g:
Table schema: a int, b timestamp
spec0: year(b) -> partitions table schema: b_year int
spec1: day(b) -> partitions table schema: b_day date

Any rows inserted in spec0 times will store partition values that is epoch years, like 52 for 2022.
Then after spec change the value 52 would be considered an epoch day instead, and will yield to "1970-02-22"

I believe the solution would be to use what DataFilesTable has already:
Partitioning.partitionType(table()); 
see: https://github.com/apache/iceberg/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Ficeberg%2FDataFilesTable.java#L55

cc: @RussellSpitzer @aokolnychyi

aokolnychyi:
I agree with your analysis, @szlta. I am surprised `PartitionsTable` does not use `Partitioning$partitionType`.

I'd do the following in separate PRs:
- Rename conflicting columns in old specs in `Partitioning$partitionType` as [discussed](https://github.com/apache/iceberg/pull/3411#discussion_r821845093).
- Use `Partitioning$partitionType` in all metadata tables.
- Consider adding `partition_spec_id` to metadata tables like `data_files` and `partitions`.
- Consider reusing equivalent columns with column IDs during spec evolutions.

szehon-ho:
Third point was already done in #3015 for *files_table, though I think not yet Partitions table

aokolnychyi:
@szlta, would you be interested to work on the steps above? If you don't have time, I may ask someone on our team to do this.

szlta:
Hi @aokolnychyi sorry for the longer pause here, I was on paternity leave ;)
Before I left, I was looking into how proposal 2 in https://github.com/apache/iceberg/pull/3411#discussion_r823692312 could be implemented in Iceberg. It seemed to be quite tricky, so first I'd jump onto the smaller fixes of the above proposed list.
I'd be happy to take this up and work on a PR that should fix these two points:

- Use Partitioning$partitionType in all metadata tables.
[...]

PR #3015, opened 2021-08-24
Add data_file.spec_id to metadata tables

szehon-ho:
* Partition Spec Id was added to DataFile in https://github.com/apache/iceberg/pull/1317 as a derived field from the parent manifest file
* This will be useful to surface to the various metadata tables that have data_file field to users
* Some internal examples of use: #1421 (distributed spark planning), #2608 (rewrite manifests)
* From offline discussion with @RussellSpitzer , who pointed to #1421 , from which the code logic was mostly extracted

PR #4516, opened 2022-04-06
Adding specId for partitions metadata table

szlta:
As per #4292, PartitionsTable should call Partitioning.partitionType() to gather the partition struct in its schema. Also the result should show specId too, as the partition key might not be considered a unique key on its own without specId in V2 format. (E.g a partition column removed then re-added along with another column..)

This is just the initial change so that I can see what tests would need fixing and to check with community if the new schema is okay like this:

1. spec_id
2. partition
3. record_count
4. file_count

I think in other meta tables the spec_id precedes the partition key in the output, so I prepended the existing schema here with it. I'm not sure if this counts as a backward-incompatible change though. What do you think @aokolnychyi ?

### DIFF
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -45,9 +45,10 @@ public class PartitionsTable extends BaseMetadataTable {
     super(ops, table, name);
 
     this.schema = new Schema(
-        Types.NestedField.required(1, "partition", table.spec().partitionType()),
-        Types.NestedField.required(2, "record_count", Types.LongType.get()),
-        Types.NestedField.required(3, "file_count", Types.IntegerType.get())
+        Types.NestedField.required(1, "spec_id", Types.IntegerType.get()),
+        Types.NestedField.required(2, "partition", Partitioning.partitionType(table)),
+        Types.NestedField.required(3, "record_count", Types.LongType.get()),
+        Types.NestedField.required(4, "file_count", Types.IntegerType.get())
     );
   }
 
@@ -89,7 +90,7 @@ public class PartitionsTable extends BaseMetadataTable {
   }
 
   private static StaticDataTask.Row convertPartition(Partition partition) {
-    return StaticDataTask.Row.of(partition.key, partition.recordCount, partition.fileCount);
+    return StaticDataTask.Row.of(partition.specId, partition.key, partition.recordCount, partition.fileCount);
   }
 
   private static Iterable<Partition> partitions(StaticTableScan scan) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -164,17 +165,20 @@ Partition get(StructLike key) {
   }
 
   static class Partition {
+    private int specId;
     private final StructLike key;
     private long recordCount;
     private int fileCount;
 
     Partition(StructLike key) {
+      this.specId = 0;
```
szehon-ho, line 174: I'm wondering one thing , didn't test it myself.  

If we have two partitions with different spec id but different key, it will show up with a random one.  Could this case happen (ie, adding and removing a field in partition spec should give a new spec id but can still produce the same partition value?)?  If so, we may need to add spec-id to key?

szehon-ho (commented):
Had one doubt on this change, looks good otherwise.

Yea when adding the spec_id to files table (#3015) , @rdblue had mentioned its ok to insert into the first field , as nobody should be depending on the field order, so I am guessing its ok in this table?

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -45,9 +45,10 @@
     super(ops, table, name);
 
     this.schema = new Schema(
-        Types.NestedField.required(1, "partition", table.spec().partitionType()),
-        Types.NestedField.required(2, "record_count", Types.LongType.get()),
-        Types.NestedField.required(3, "file_count", Types.IntegerType.get())
+        Types.NestedField.required(1, "spec_id", Types.IntegerType.get()),
```
rdblue, line 48: It is fine to add a new field, but you should not change the IDs of the other fields.

szlta: Okay, in this case spec_id will be appended to the schema then.

core/src/main/java/org/apache/iceberg/PartitionsTable.java
szehon-ho: I see, yea you are right, seems Iceberg re-uses the spec id if we revert the spec to a former one.

szehon-ho (approved):
It looks ok to me.

szehon-ho:
Thanks @szlta for change and @rdblue for taking a look, let's work on the subsequent issue.

szlta:
Thanks a lot for the review @szehon-ho and @rdblue

szehon-ho:
I tried to test these cases, while this fixes the PartitionsTable schema, some of the partition entries may have problems (as current spec is still used in a few other places).   I made #4560 pr as a follow up (@szlta sorry I was not sure if you were working on going to work on this or something else).

PR #4517, opened 2022-04-06
Spark 3.2: Use SnapshotSummary to deduce latestOffset in Spark SS

singhpk234:
- At present we iterate the iterator of addedFiles to find the no of added files in the snapshot, this info is already present in snapshot summary. We should use this rather than iterating the addedFiles iterator to determine files added.

Came across this while working on 
- https://github.com/apache/iceberg/pull/4479

--- 

cc @rdblue, @RussellSpitzer, @jackye1995, @kbendick

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
@@ -51,6 +51,7 @@ import org.apache.iceberg.spark.SparkReadConf;
 import org.apache.iceberg.spark.SparkReadOptions;
 import org.apache.iceberg.spark.source.SparkScan.ReadTask;
 import org.apache.iceberg.spark.source.SparkScan.ReaderFactory;
+import org.apache.iceberg.util.PropertyUtil;
 import org.apache.iceberg.util.SnapshotUtil;
 import org.apache.iceberg.util.TableScanUtil;
 import org.apache.iceberg.util.Tasks;
@@ -112,10 +113,12 @@ public class SparkMicroBatchStream implements MicroBatchStream {
     }
 
     Snapshot latestSnapshot = table.currentSnapshot();
-    return new StreamingOffset(latestSnapshot.snapshotId(),
-        Long.parseLong(latestSnapshot.summary().getOrDefault(SnapshotSummary.ADDED_FILES_PROP,
-            String.valueOf(Iterables.size(latestSnapshot.addedFiles())))),
-        false);
+    long addedFilesCount = PropertyUtil.propertyAsLong(latestSnapshot.summary(), SnapshotSummary.ADDED_FILES_PROP, -1);
+    // If snapshotSummary doesn't have SnapshotSummary.ADDED_FILES_PROP, iterate through addedFiles iterator to find
+    // addedFilesCount.
+    addedFilesCount = addedFilesCount == -1 ? Iterables.size(latestSnapshot.addedFiles()) : addedFilesCount;
+
+    return new StreamingOffset(latestSnapshot.snapshotId(), addedFilesCount, false);
   }
 
   @Override
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
@@ -37,6 +37,7 @@ import org.apache.iceberg.Schema;
 import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.SerializableTable;
 import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotSummary;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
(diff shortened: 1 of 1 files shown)


SreeramGarlapati (changes requested):
@singhpk234 - we **explicitly opted out of** reading the `SnapshotSummary` - based on discussion with @aokolnychyi, @RussellSpitzer & @rdblue. 

rationale: `SnapshotSummary` is a free form dictionary & `ADDED_FILES_PROP` as a key in this dictionary - **is NOT** added to [Iceberg Spec](https://iceberg.apache.org/spec/#snapshots) & is not NOT populated by all engines yet.

Pl. read thru this comment thread for full context.
https://github.com/apache/iceberg/pull/2660#discussion_r650154461

singhpk234:
Thanks @SreeramGarlapati !!! for sharing the https://github.com/apache/iceberg/pull/2660#discussion_r650154461 thread, it really helps. Glad to know it was already considered while the implementation. 

It looks like we were on the fence to establish the reliability of this metric and hence decided to re-visit it in future, refering this comment https://github.com/apache/iceberg/pull/2660#discussion_r650322788

 > SnapshotSummary is a free form dictionary & ADDED_FILES_PROP as a key in this dictionary - is NOT added to [Iceberg Spec](https://iceberg.apache.org/spec/#snapshots)

Agree with you on this.

> is not NOT populated by all engines yet

As per my understanding engines don't own the responsibility of calculating / updating the stats, IIUC it's done entirely by core library. The flow for ex : [commit() [from engine]](https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java#L251) -> [apply() in SnapshotProducer (core)](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L296) -> [Summary calculation (core)](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L202-L210).
Am I missing something here ?

rdblue:
@singhpk234, this is going to get the size of `latestSnapshot.addedFiles()` either way because it needs to calculate that value to pass it into the property check.

@SreeramGarlapati makes a good point, but if the added files property is present, then it seems to me like we could use it and fall back to reading the snapshot.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -111,7 +112,10 @@ public Offset latestOffset() {
     }
 
     Snapshot latestSnapshot = table.currentSnapshot();
-    return new StreamingOffset(latestSnapshot.snapshotId(), Iterables.size(latestSnapshot.addedFiles()), false);
+    return new StreamingOffset(latestSnapshot.snapshotId(),
+        Long.parseLong(latestSnapshot.summary().getOrDefault(SnapshotSummary.ADDED_FILES_PROP,
```
jackye1995, line 116: can use `PropertyUtil.propertyAsLong`. Also this is too long to squeeze in a single line call. Can we have 1 line getting the added files, another line to call `return new StreamingOffset(latestSnapshot.snapshotId(), addedFilesCount, false);`

singhpk234: I see, I didn't use `PropertyUtil.propertyAsLong` as to provide the default, I will have to iterate the addedFiles iterator to compute it's size (implying materilizing this), presently this util doesn't supports sending it a supplier so that it can be computed only when key is not present. what do you recommend here ?

singhpk234: I also see the similar problem will be there in my current approach in the way hashMap is implemented it. 
```
    default V getOrDefault(Object key, V defaultValue) {
        V v;
        return (((v = get(key)) != null) || containsKey(key))
            ? v
            : defaultValue;
    }
```

Let me try to fix it

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java
```diff
@@ -111,7 +113,12 @@ public Offset latestOffset() {
     }
 
     Snapshot latestSnapshot = table.currentSnapshot();
-    return new StreamingOffset(latestSnapshot.snapshotId(), Iterables.size(latestSnapshot.addedFiles()), false);
+    long addedFilesCount = PropertyUtil.propertyAsLong(latestSnapshot.summary(), SnapshotSummary.ADDED_FILES_PROP, -1);
+    // If snapshotSummary doesn't have SnapshotSummary.ADDED_FILES_PROP, iterate through addedFiles iterator to find
+    // addedFilesCount.
+    addedFilesCount = addedFilesCount == -1 ? Iterables.size(latestSnapshot.addedFiles()) : addedFilesCount;
```
jackye1995, line 119: ah I see, good point we cannot just put `Iterables.size(latestSnapshot.addedFiles())` as the input to the last method otherwise it will be evaluated.

jackye1995:
@SreeramGarlapati @rdblue any additional comments? If not I think it is good to be merged.

jackye1995:
I think this is good to go, and there is no further comment for quite a while now. I will merge it, thanks for the reviews! @rdblue @SreeramGarlapati

PR #4518, opened 2022-04-06
core: Provide mechanism to cache manifest file content

### LINKED ISSUE #4508: Provide mechanism to cache manifest file content across several planFiles call
Iceberg planFiles API has benefit of filtering out unnecessary files given a set of filter predicates. During query planning, a query engine might need to call planFiles against the same table multiple times but with slightly different filter predicates for each scan. This lead to repeated manifest files reading over the same table. It is often expensive, especially if it involve remote reads through network. Certain stream operation such as backward seek and small reads characteristics by Avro library can have high latency depending on the target file system / cloud connector.

Having ability to cache manifest content in memory for several planFiles calls can improve performance. One idea is to add method in TableScanContext.java to supply a caffeine cache, say `Cache<String, byte[]>`, and pass it further to AvroIO.java. In turn, AvroIO can consult with this cache and use the cached manifest file content if it is available.

rizaon:
This is a draft PR for ManifestCache implementation.

The `ManifestCache` interface closely follow `com.github.benmanes.caffeine.cache.Cache` interface, with addition of specifying `maxContentLength()`. If stream length is longer than `maxContentLength()`, AvroIO will skip caching it to avoid memory pressure. An example of implementation, `CaffeineManifestCache`, can be seen in TestAvroFileSplit.java.

I will tidy this up and add documentations as we go. Looking forward for any feedback.

Closes #4508

### DIFF
--- a/core/src/main/java/org/apache/iceberg/CatalogProperties.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogProperties.java
@@ -19,6 +19,7 @@
 package org.apache.iceberg;
 
 import java.util.concurrent.TimeUnit;
+import org.apache.iceberg.io.ContentCache;
 
 public class CatalogProperties {
 
@@ -56,6 +57,64 @@ public class CatalogProperties {
 
   public static final long CACHE_EXPIRATION_INTERVAL_MS_DEFAULT = TimeUnit.SECONDS.toMillis(30);
   public static final long CACHE_EXPIRATION_INTERVAL_MS_OFF = -1;
+
+  /**
+   * Controls whether to use {@link ContentCache} during manifest file reads or not.
+   *
+   * <p>This value will be ignored and the file cache will be disabled if any of the following is
+   * true:
+   *
+   * <ul>
+   *   <li>{@link #IO_CACHE_EXPIRATION_INTERVAL_MS} is set to negative number.
+   *   <li>{@link #IO_CACHE_MAX_TOTAL_BYTES} is set to non-positive value.
+   *   <li>{@link #IO_CACHE_MAX_CONTENT_LENGTH} is set to non-positive value.
+   * </ul>
+   */
+  public static final String IO_CACHE_ENABLED = "io.cache-enabled";
+
+  public static final boolean IO_CACHE_ENABLED_DEFAULT = false;
+
+  /**
+   * Controls the duration for which entries in the {@link ContentCache} are cached.
+   *
+   * <p>Behavior of specific values of cache.expiration-interval-ms:
+   *
+   * <ul>
+   *   <li>Negative Values - Caching disabled
+   *   <li>Zero - Cache entries expires only if it gets evicted due to memory pressure from {@link
+   *       #IO_CACHE_MAX_TOTAL_BYTES} setting.
+   *   <li>Positive Values - Cache entries expire if not accessed via the cache after this many
+   *       milliseconds
+   * </ul>
+   */
+  public static final String IO_CACHE_EXPIRATION_INTERVAL_MS = "io.cache.expiration-interval-ms";
+
+  public static final long IO_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT = TimeUnit.SECONDS.toMillis(60);
+
+  /**
+   * Controls the maximum total amount of bytes to cache in {@link ContentCache}.
+   *
+   * <p>Behavior of specific values of cache.content.max-total-bytes:
+   *
+   * <ul>
+   *   <li>Non-Positive Values - Caching disabled
+   *   <li>Positive Values - Maximum total amount of bytes allowed in cache before cache start doing
+   *       eviction
+   * </ul>
+   */
+  public static final String IO_CACHE_MAX_TOTAL_BYTES = "io.cache.max-total-bytes";
+
+  public static final long IO_CACHE_MAX_TOTAL_BYTES_DEFAULT = 100 * 1024 * 1024;
+
+  /**
+   * Controls the maximum length of file to be considered for caching.
+   *
+   * <p>An {@link org.apache.iceberg.io.InputFile} will not be cached if the length is longer than
+   * this limit.
+   */
+  public static final String IO_CACHE_MAX_CONTENT_LENGTH = "io.cache.max-content-length";
+
+  public static final long IO_CACHE_MAX_CONTENT_LENGTH_DEFAULT = 8 * 1024 * 1024;
 
   public static final String URI = "uri";
   public static final String CLIENT_POOL_SIZE = "clients";
--- a/core/src/main/java/org/apache/iceberg/CatalogProperties.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogProperties.java
@@ -65,14 +65,14 @@ public class CatalogProperties {
    * true:
    *
    * <ul>
-   *   <li>{@link #IO_CACHE_EXPIRATION_INTERVAL_MS} is set to negative number.
-   *   <li>{@link #IO_CACHE_MAX_TOTAL_BYTES} is set to non-positive value.
-   *   <li>{@link #IO_CACHE_MAX_CONTENT_LENGTH} is set to non-positive value.
+   *   <li>{@link #IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS} is set to negative number.
+   *   <li>{@link #IO_MANIFEST_CACHE_MAX_TOTAL_BYTES} is set to non-positive value.
+   *   <li>{@link #IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH} is set to non-positive value.
    * </ul>
    */
-  public static final String IO_CACHE_ENABLED = "io.cache-enabled";
+  public static final String IO_MANIFEST_CACHE_ENABLED = "io.manifest.cache-enabled";
 
-  public static final boolean IO_CACHE_ENABLED_DEFAULT = false;
+  public static final boolean IO_MANIFEST_CACHE_ENABLED_DEFAULT = false;
 
   /**
    * Controls the duration for which entries in the {@link ContentCache} are cached.
@@ -82,14 +82,16 @@ public class CatalogProperties {
    * <ul>
    *   <li>Negative Values - Caching disabled
    *   <li>Zero - Cache entries expires only if it gets evicted due to memory pressure from {@link
-   *       #IO_CACHE_MAX_TOTAL_BYTES} setting.
+   *       #IO_MANIFEST_CACHE_MAX_TOTAL_BYTES} setting.
    *   <li>Positive Values - Cache entries expire if not accessed via the cache after this many
    *       milliseconds
    * </ul>
    */
-  public static final String IO_CACHE_EXPIRATION_INTERVAL_MS = "io.cache.expiration-interval-ms";
+  public static final String IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS =
+      "io.manifest.cache.expiration-interval-ms";
 
-  public static final long IO_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT = TimeUnit.SECONDS.toMillis(60);
+  public static final long IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT =
+      TimeUnit.SECONDS.toMillis(60);
 
   /**
    * Controls the maximum total amount of bytes to cache in {@link ContentCache}.
@@ -102,9 +104,10 @@ public class CatalogProperties {
    *       eviction
    * </ul>
    */
-  public static final String IO_CACHE_MAX_TOTAL_BYTES = "io.cache.max-total-bytes";
+  public static final String IO_MANIFEST_CACHE_MAX_TOTAL_BYTES =
+      "io.manifest.cache.max-total-bytes";
 
-  public static final long IO_CACHE_MAX_TOTAL_BYTES_DEFAULT = 100 * 1024 * 1024;
+  public static final long IO_MANIFEST_CACHE_MAX_TOTAL_BYTES_DEFAULT = 100 * 1024 * 1024;
 
   /**
    * Controls the maximum length of file to be considered for caching.
@@ -112,9 +115,10 @@ public class CatalogProperties {
    * <p>An {@link org.apache.iceberg.io.InputFile} will not be cached if the length is longer than
    * this limit.
    */
-  public static final String IO_CACHE_MAX_CONTENT_LENGTH = "io.cache.max-content-length";
+  public static final String IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH =
+      "io.manifest.cache.max-content-length";
 
-  public static final long IO_CACHE_MAX_CONTENT_LENGTH_DEFAULT = 8 * 1024 * 1024;
+  public static final long IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH_DEFAULT = 8 * 1024 * 1024;
 
   public static final String URI = "uri";
   public static final String CLIENT_POOL_SIZE = "clients";
--- a/core/src/main/java/org/apache/iceberg/CatalogProperties.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogProperties.java
@@ -19,7 +19,6 @@
 package org.apache.iceberg;
 
 import java.util.concurrent.TimeUnit;
-import org.apache.iceberg.io.ContentCache;
 
 public class CatalogProperties {
 
@@ -59,15 +58,14 @@ public class CatalogProperties {
   public static final long CACHE_EXPIRATION_INTERVAL_MS_OFF = -1;
 
   /**
-   * Controls whether to use {@link ContentCache} during manifest file reads or not.
+   * Controls whether to use caching during manifest reads or not.
    *
-   * <p>This value will be ignored and the file cache will be disabled if any of the following is
-   * true:
+   * <p>Enabling manifest file caching require the following configuration constraints to be true:
    *
    * <ul>
-   *   <li>{@link #IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS} is set to negative number.
-   *   <li>{@link #IO_MANIFEST_CACHE_MAX_TOTAL_BYTES} is set to non-positive value.
-   *   <li>{@link #IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH} is set to non-positive value.
+   *   <li>{@link #IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS} must be a non-negative value.
+   *   <li>{@link #IO_MANIFEST_CACHE_MAX_TOTAL_BYTES} must be a positive value.
+   *   <li>{@link #IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH} must be a positive value.
    * </ul>
    */
   public static final String IO_MANIFEST_CACHE_ENABLED = "io.manifest.cache-enabled";
@@ -75,12 +73,11 @@ public class CatalogProperties {
   public static final boolean IO_MANIFEST_CACHE_ENABLED_DEFAULT = false;
 
   /**
-   * Controls the duration for which entries in the {@link ContentCache} are cached.
+   * Controls the maximum duration for which an entry stays in the manifest cache.
    *
-   * <p>Behavior of specific values of cache.expiration-interval-ms:
+   * <p>Must be a non-negative value. Following are specific behaviors of this config:
    *
    * <ul>
-   *   <li>Negative Values - Caching disabled
    *   <li>Zero - Cache entries expires only if it gets evicted due to memory pressure from {@link
    *       #IO_MANIFEST_CACHE_MAX_TOTAL_BYTES} setting.
    *   <li>Positive Values - Cache entries expire if not accessed via the cache after this many
@@ -94,15 +91,9 @@ public class CatalogProperties {
       TimeUnit.SECONDS.toMillis(60);
 
   /**
-   * Controls the maximum total amount of bytes to cache in {@link ContentCache}.
+   * Controls the maximum total amount of bytes to cache in manifest cache.
    *
-   * <p>Behavior of specific values of cache.content.max-total-bytes:
-   *
-   * <ul>
-   *   <li>Non-Positive Values - Caching disabled
-   *   <li>Positive Values - Maximum total amount of bytes allowed in cache before cache start doing
-   *       eviction
-   * </ul>
+   * <p>Must be a positive value.
    */
   public static final String IO_MANIFEST_CACHE_MAX_TOTAL_BYTES =
       "io.manifest.cache.max-total-bytes";
@@ -113,7 +104,7 @@ public class CatalogProperties {
    * Controls the maximum length of file to be considered for caching.
    *
    * <p>An {@link org.apache.iceberg.io.InputFile} will not be cached if the length is longer than
-   * this limit.
+   * this limit. Must be a positive value.
    */
   public static final String IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH =
       "io.manifest.cache.max-content-length";
--- a/core/src/main/java/org/apache/iceberg/ManifestFiles.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFiles.java
@@ -31,9 +31,11 @@ import org.apache.iceberg.io.ContentCache;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
+import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.util.PropertyUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,20 +53,28 @@ public class ManifestFiles {
               ManifestFile.PARTITION_SUMMARY_TYPE,
               GenericPartitionFieldSummary.class.getName()));
 
-  private static final Cache<FileIO, ContentCache> CONTENT_CACHES =
-      Caffeine.newBuilder().weakKeys().maximumSize(maxFileIO()).recordStats().build();
+  @VisibleForTesting
+  static Cache<FileIO, ContentCache> newManifestCache() {
+    return Caffeine.newBuilder()
+        .weakKeys()
+        .softValues()
+        .maximumSize(maxFileIO())
+        .removalListener(
+            (io, contentCache, cause) ->
+                LOG.debug("Evicted {} from FileIO-level cache ({})", io, cause))
+        .recordStats()
+        .build();
+  }
 
-  /**
-   * Get or create a manifest file cache for a given FileIO.
-   *
-   * <p>Returned cache object will be kept in memory until {@link #dropCache(FileIO)} is called.
-   * Return null if the given FileIO properties does not allow caching.
-   *
-   * @param fileIO A FileIO.
-   * @return A manifest file cache or null if the given FileIO properties does not allow caching.
-   */
-  public static ContentCache getOrCreateCache(FileIO fileIO) {
-    return CONTENT_CACHES.get(fileIO, io -> ContentCache.createCache(io.properties()));
+  private static final Cache<FileIO, ContentCache> CONTENT_CACHES = newManifestCache();
+
+  @VisibleForTesting
+  static ContentCache contentCache(FileIO io) {
+    return CONTENT_CACHES.get(
+        io,
+        fileIO ->
+            new ContentCache(
+                cacheDurationMs(fileIO), cacheTotalBytes(fileIO), cacheMaxContentLength(fileIO)));
   }
 
   /** Drop manifest file cache object for a FileIO if exists. */
@@ -331,21 +341,21 @@ public class ManifestFiles {
   }
 
   private static InputFile newInputFile(FileIO io, String path, long length) {
-    try {
-      ContentCache cache = getOrCreateCache(io);
-      if (cache != null) {
-        LOG.debug("FileIO-level cache stats: {}", CONTENT_CACHES.stats());
-        return cache.tryCache(io, path, length);
-      } else {
-        return io.newInputFile(path, length);
-      }
-    } catch (UnsupportedOperationException e) {
-      return io.newInputFile(path, length);
+    boolean enabled = cachingEnabled(io);
+
+    if (enabled) {
+      ContentCache cache = contentCache(io);
+      Preconditions.checkNotNull(cache);
+      LOG.debug("FileIO-level cache stats: {}", CONTENT_CACHES.stats());
+      return cache.tryCache(io, path, length);
     }
+
+    // caching is not enable for this io or caught RuntimeException.
+    return io.newInputFile(path, length);
   }
 
   private static int maxFileIO() {
-    String value = System.getProperty(SystemProperties.IO_CACHE_MAX_FILEIO);
+    String value = System.getProperty(SystemProperties.IO_MANIFEST_CACHE_MAX_FILEIO);
     if (value != null) {
       try {
         return Integer.parseUnsignedInt(value);
@@ -353,6 +363,34 @@ public class ManifestFiles {
         // will return the default
       }
     }
-    return SystemProperties.IO_CACHE_MAX_FILEIO_DEFAULT;
+    return SystemProperties.IO_MANIFEST_CACHE_MAX_FILEIO_DEFAULT;
+  }
+
+  public static boolean cachingEnabled(FileIO io) {
+    return PropertyUtil.propertyAsBoolean(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_ENABLED,
+        CatalogProperties.IO_MANIFEST_CACHE_ENABLED_DEFAULT);
+  }
+
+  public static long cacheDurationMs(FileIO io) {
+    return PropertyUtil.propertyAsLong(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS,
+        CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT);
+  }
+
+  public static long cacheTotalBytes(FileIO io) {
+    return PropertyUtil.propertyAsLong(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES,
+        CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES_DEFAULT);
+  }
+
+  public static long cacheMaxContentLength(FileIO io) {
+    return PropertyUtil.propertyAsLong(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH,
+        CatalogProperties.IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH_DEFAULT);
   }
 }
--- a/core/src/main/java/org/apache/iceberg/ManifestFiles.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFiles.java
@@ -341,11 +341,20 @@ public class ManifestFiles {
   }
 
   private static InputFile newInputFile(FileIO io, String path, long length) {
-    boolean enabled = cachingEnabled(io);
+    boolean enabled = false;
+
+    try {
+      enabled = cachingEnabled(io);
+    } catch (UnsupportedOperationException e) {
+      // There is an issue reading io.properties(). Disable caching.
+      enabled = false;
+    }
 
     if (enabled) {
       ContentCache cache = contentCache(io);
-      Preconditions.checkNotNull(cache);
+      Preconditions.checkNotNull(
+          cache,
+          "ContentCache creation failed. Check that all manifest caching configurations has valid value.");
       LOG.debug("FileIO-level cache stats: {}", CONTENT_CACHES.stats());
       return cache.tryCache(io, path, length);
     }
--- a/core/src/main/java/org/apache/iceberg/ManifestFiles.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFiles.java
@@ -18,6 +18,8 @@
  */
 package org.apache.iceberg;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import java.io.IOException;
 import java.util.Map;
 import org.apache.iceberg.ManifestReader.FileType;
@@ -25,15 +27,20 @@ import org.apache.iceberg.avro.AvroEncoderUtil;
 import org.apache.iceberg.avro.AvroSchemaUtil;
 import org.apache.iceberg.exceptions.RuntimeIOException;
 import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.io.ContentCache;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ManifestFiles {
   private ManifestFiles() {}
+
+  private static final Logger LOG = LoggerFactory.getLogger(ManifestFiles.class);
 
   private static final org.apache.avro.Schema MANIFEST_AVRO_SCHEMA =
       AvroSchemaUtil.convert(
@@ -43,6 +50,28 @@ public class ManifestFiles {
               GenericManifestFile.class.getName(),
               ManifestFile.PARTITION_SUMMARY_TYPE,
               GenericPartitionFieldSummary.class.getName()));
+
+  private static final Cache<FileIO, ContentCache> CONTENT_CACHES =
+      Caffeine.newBuilder().weakKeys().maximumSize(maxFileIO()).recordStats().build();
+
+  /**
+   * Get or create a manifest file cache for a given FileIO.
+   *
+   * <p>Returned cache object will be kept in memory until {@link #dropCache(FileIO)} is called.
+   * Return null if the given FileIO properties does not allow caching.
+   *
+   * @param fileIO A FileIO.
+   * @return A manifest file cache or null if the given FileIO properties does not allow caching.
+   */
+  public static ContentCache getOrCreateCache(FileIO fileIO) {
+    return CONTENT_CACHES.get(fileIO, io -> ContentCache.createCache(io.properties()));
+  }
+
+  /** Drop manifest file cache object for a FileIO if exists. */
+  public static synchronized void dropCache(FileIO fileIO) {
+    CONTENT_CACHES.invalidate(fileIO);
+    CONTENT_CACHES.cleanUp();
+  }
 
   /**
    * Returns a {@link CloseableIterable} of file paths in the {@link ManifestFile}.
@@ -86,7 +115,7 @@ public class ManifestFiles {
         manifest.content() == ManifestContent.DATA,
         "Cannot read a delete manifest with a ManifestReader: %s",
         manifest);
-    InputFile file = io.newInputFile(manifest.path(), manifest.length());
+    InputFile file = newInputFile(io, manifest.path(), manifest.length());
     InheritableMetadata inheritableMetadata = InheritableMetadataFactory.fromManifest(manifest);
     return new ManifestReader<>(file, specsById, inheritableMetadata, FileType.DATA_FILES);
   }
@@ -140,7 +169,7 @@ public class ManifestFiles {
         manifest.content() == ManifestContent.DELETES,
         "Cannot read a data manifest with a DeleteManifestReader: %s",
         manifest);
-    InputFile file = io.newInputFile(manifest.path(), manifest.length());
+    InputFile file = newInputFile(io, manifest.path(), manifest.length());
     InheritableMetadata inheritableMetadata = InheritableMetadataFactory.fromManifest(manifest);
     return new ManifestReader<>(file, specsById, inheritableMetadata, FileType.DELETE_FILES);
   }
@@ -299,5 +328,31 @@ public class ManifestFiles {
     }
 
     return writer.toManifestFile();
+  }
+
+  private static InputFile newInputFile(FileIO io, String path, long length) {
+    try {
+      ContentCache cache = getOrCreateCache(io);
+      if (cache != null) {
+        LOG.debug("FileIO-level cache stats: {}", CONTENT_CACHES.stats());
+        return cache.tryCache(io, path, length);
+      } else {
+        return io.newInputFile(path, length);
+      }
+    } catch (UnsupportedOperationException e) {
+      return io.newInputFile(path, length);
+    }
+  }
+
+  private static int maxFileIO() {
+    String value = System.getProperty(SystemProperties.IO_CACHE_MAX_FILEIO);
+    if (value != null) {
+      try {
+        return Integer.parseUnsignedInt(value);
+      } catch (NumberFormatException e) {
+        // will return the default
+      }
+    }
+    return SystemProperties.IO_CACHE_MAX_FILEIO_DEFAULT;
   }
 }
--- a/core/src/main/java/org/apache/iceberg/ManifestFiles.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFiles.java
@@ -366,28 +366,28 @@ public class ManifestFiles {
     return SystemProperties.IO_MANIFEST_CACHE_MAX_FILEIO_DEFAULT;
   }
 
-  public static boolean cachingEnabled(FileIO io) {
+  static boolean cachingEnabled(FileIO io) {
     return PropertyUtil.propertyAsBoolean(
         io.properties(),
         CatalogProperties.IO_MANIFEST_CACHE_ENABLED,
         CatalogProperties.IO_MANIFEST_CACHE_ENABLED_DEFAULT);
   }
 
-  public static long cacheDurationMs(FileIO io) {
+  static long cacheDurationMs(FileIO io) {
     return PropertyUtil.propertyAsLong(
         io.properties(),
         CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS,
         CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT);
   }
 
-  public static long cacheTotalBytes(FileIO io) {
+  static long cacheTotalBytes(FileIO io) {
     return PropertyUtil.propertyAsLong(
         io.properties(),
         CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES,
         CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES_DEFAULT);
   }
 
-  public static long cacheMaxContentLength(FileIO io) {
+  static long cacheMaxContentLength(FileIO io) {
     return PropertyUtil.propertyAsLong(
         io.properties(),
         CatalogProperties.IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH,
--- a/core/src/main/java/org/apache/iceberg/SystemProperties.java
+++ b/core/src/main/java/org/apache/iceberg/SystemProperties.java
@@ -33,6 +33,14 @@ public class SystemProperties {
   /** Whether to use the shared worker pool when planning table scans. */
   public static final String SCAN_THREAD_POOL_ENABLED = "iceberg.scan.plan-in-worker-pool";
 
+  /**
+   * Maximum number of distinct {@link org.apache.iceberg.io.FileIO} that is allowed to have
+   * associated {@link org.apache.iceberg.io.ContentCache} in memory at a time.
+   */
+  public static final String IO_CACHE_MAX_FILEIO = "iceberg.io.cache.max-fileio";
+
+  public static final int IO_CACHE_MAX_FILEIO_DEFAULT = 8;
+
   static boolean getBoolean(String systemProperty, boolean defaultValue) {
     String value = System.getProperty(systemProperty);
     if (value != null) {
--- a/core/src/main/java/org/apache/iceberg/SystemProperties.java
+++ b/core/src/main/java/org/apache/iceberg/SystemProperties.java
@@ -37,9 +37,9 @@ public class SystemProperties {
    * Maximum number of distinct {@link org.apache.iceberg.io.FileIO} that is allowed to have
    * associated {@link org.apache.iceberg.io.ContentCache} in memory at a time.
    */
-  public static final String IO_CACHE_MAX_FILEIO = "iceberg.io.cache.max-fileio";
+  public static final String IO_MANIFEST_CACHE_MAX_FILEIO = "iceberg.io.manifest.cache.fileio-max";
 
-  public static final int IO_CACHE_MAX_FILEIO_DEFAULT = 8;
+  public static final int IO_MANIFEST_CACHE_MAX_FILEIO_DEFAULT = 8;
 
   static boolean getBoolean(String systemProperty, boolean defaultValue) {
     String value = System.getProperty(systemProperty);
--- a/core/src/main/java/org/apache/iceberg/io/ContentCache.java
+++ b/core/src/main/java/org/apache/iceberg/io/ContentCache.java
@@ -32,6 +32,7 @@ import java.util.function.Function;
 import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.util.PropertyUtil;
@@ -64,11 +65,28 @@ public class ContentCache {
             CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH,
             CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH_DEFAULT);
 
-    if (enabled && durationMs >= 0 && totalBytes > 0 && contentLength > 0) {
-      return new ContentCache(durationMs, totalBytes, contentLength);
+    if (!enabled) {
+      LOG.debug("No ContentCache created. {} is false.", CatalogProperties.IO_CACHE_ENABLED);
+    } else if (durationMs < 0) {
+      LOG.debug(
+          "No ContentCache created. {} ({}) is less than 0.",
+          CatalogProperties.IO_CACHE_EXPIRATION_INTERVAL_MS,
+          durationMs);
+    } else if (totalBytes <= 0) {
+      LOG.debug(
+          "No ContentCache created. {} ({}) is equal or less than 0.",
+          CatalogProperties.IO_CACHE_MAX_TOTAL_BYTES,
+          totalBytes);
+    } else if (contentLength <= 0) {
+      LOG.debug(
+          "No ContentCache created. {} ({}) is equal or less than 0.",
+          CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH,
+          contentLength);
     } else {
-      return null;
+      return new ContentCache(durationMs, totalBytes, contentLength);
     }
+
+    return null;
   }
 
   private final long expireAfterAccessMs;
@@ -145,20 +163,12 @@ public class ContentCache {
 
   @Override
   public String toString() {
-    return getClass().getSimpleName()
-        + '{'
-        + "expireAfterAccessMs="
-        + expireAfterAccessMs
-        + ", "
-        + "maxContentLength="
-        + maxContentLength
-        + ", "
-        + "maxTotalBytes="
-        + maxTotalBytes
-        + ", "
-        + "cacheStats="
-        + cache.stats()
-        + '}';
+    return MoreObjects.toStringHelper(this)
+        .add("expireAfterAccessMs", expireAfterAccessMs)
+        .add("maxContentLength", maxContentLength)
+        .add("maxTotalBytes", maxTotalBytes)
+        .add("cacheStats", cache.stats())
+        .toString();
   }
 
   private static class CacheEntry {
--- a/core/src/main/java/org/apache/iceberg/io/ContentCache.java
+++ b/core/src/main/java/org/apache/iceberg/io/ContentCache.java
@@ -24,77 +24,55 @@ import com.github.benmanes.caffeine.cache.Weigher;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.List;
-import java.util.Map;
 import java.util.function.Function;
-import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.exceptions.NotFoundException;
-import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
-import org.apache.iceberg.util.PropertyUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Class that provides file-content caching during reading.
+ *
+ * <p>The file-content caching is initiated by calling {@link ContentCache#tryCache(FileIO, String,
+ * long)}. Given a FileIO, a file location string, and file length that is within allowed limit,
+ * ContentCache will return a {@link CachingInputFile} that is backed by the cache. Calling {@link
+ * CachingInputFile#newStream()} will return a {@link ByteBufferInputStream} backed by list of
+ * {@link ByteBuffer} from the cache if such file-content exist in the cache. If the file-content
+ * does not exist in the cache yet, a regular InputFile will be instantiated, read-ahead, and loaded
+ * into the cache before returning ByteBufferInputStream. The regular InputFile is also used as a
+ * fallback if cache loading fail.
+ */
 public class ContentCache {
   private static final Logger LOG = LoggerFactory.getLogger(ContentCache.class);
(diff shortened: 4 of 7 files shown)


rizaon:
[485e2ab](https://github.com/apache/iceberg/pull/4518/commits/485e2abf4bb3c67f7ab0db2e5cf2e78397bf25db) Switch ManifestCache class to use Guava Cache instead. This is easier to do in iceberg-api since we can simply include the Guava cache classes into the shadow jar of iceberg-bundled-guava.

rdblue:
Switching to use more Guava classes is probably not a good idea.

@rizaon, do you have use cases where this has helped? If so, what were they? This adds quite a bit of complexity and memory overhead that we purposely avoided up to now so I want to make sure it is worth the change.

Have you considered adding a caching FileIO instance? That could be used for more use cases than just manifest files and job planning. For example, a FileIO read-through cache could cache delete files that might be reused for tasks. This could be configured by file path, making it easy to determine what to cache. Plus, we could use more options than just in-memory, like a distributed cache or local disk.

rizaon:
Hi @rdblue, thank you for your feedback.

We found a slow query compilation issue against the Iceberg table in our recent Apache Impala build. Impala uses Iceberg's HiveCatalog and HadoopFileIO instance with an S3A input stream to access data from S3. We did a full 10 TB TPC-DS benchmark and found that query compilation can go for several seconds, while it used to be less than a second with native hive tables. This slowness in single query compilation is due to the requirement to call planFiles several times, even for scan nodes targetting the same table. We also see several socket read operations that spend hundreds of milliseconds during planFiles, presumably due to S3A HTTP HEAD request overhead and backward seek overhead (issue #4508). This is especially hurt fast-running queries.

We tried this caching solution and it help speed up Impala query compilation almost 5x faster compared to without on Iceberg tables. Our original solution, however, is to put a Caffeine cache as a singleton in AvroIO.java. I thought it is better to supply the cache from outside.

I have not considered the solution of adding a caching FileIO instance. I'm pretty new to the Iceberg codebase but interested to follow up on that if it can yield a better integration. Will it require a new class of Catalog/Table as well, or can we improve on the existing HiveCatalog & HadoopFileIO?

rizaon:
A relevant Impala's JIRA is here:
https://issues.apache.org/jira/browse/IMPALA-11171

rdblue:
@rizaon, caching in the FileIO layer would be much more general. You could do things like detect that the file size is less than some threshold and cache it in memory, or detect file names under `metadata` and cache those. I highly recommend looking into the FileIO API!

rizaon:
[c032b7a](https://github.com/apache/iceberg/pull/4518/commits/c032b7ab2f2db941fe5433c13a38dfcb8bf538ef) implement caching as a new FileIO class, CachingHadoopFileIO. A new Tables class, CachingHadoopTables, is also added to assist with testing.

We tried to avoid `lazyStats()` call as much as possible by checking the cache first before checking for stream length. This comes with a risk of keeping stale data in memory when the actual file is already deleted, as shown in `CachingHiveTableTest.testMissingMetadataWontCauseHang()`. This is probably fine, given the immutable nature of iceberg metadata. Otherwise, application should invalidate cache entry first to force re-read / `lazyStats()` call.

core/src/main/java/org/apache/iceberg/hadoop/HadoopInputFile.java
```diff
@@ -241,4 +272,118 @@ public void setNativeCryptoParameters(NativeFileCryptoParameters nativeCryptoPar
   public String toString() {
     return path.toString();
   }
+
+  protected void setContentCache(CachingHadoopFileIO.ContentCache cache) {
+    this.contentCache = cache;
```
rizaon, line 277: We can probably move the cache existence check here and additionally set `length` if appropriate.

core/src/main/java/org/apache/iceberg/hadoop/CachingHadoopFileIO.java
```diff
@@ -0,0 +1,202 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Variant of {@link HadoopFileIO} that adds metadata content caching features.
+ * <p>
+ * This FileIO intend to speedup scan planning by caching most of the table metadata contents in memory and reduce
+ * remote reads.
+ */
+public class CachingHadoopFileIO extends HadoopFileIO {
```
rdblue, line 39: Why is this specific to Hadoop? Why not cache with other IO?

core/src/main/java/org/apache/iceberg/hadoop/CachingHadoopTables.java
```diff
@@ -0,0 +1,77 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.hadoop;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.iceberg.io.FileIO;
+
+public class CachingHadoopTables extends HadoopTables {
```
rdblue, line 25: It's a red flag that `HadoopTables` is used. What does caching in `FileIO` have to do with the method used to load tables ? I think  those are orthogonal concerns.

core/src/main/java/org/apache/iceberg/hadoop/ConfigProperties.java
```diff
@@ -26,4 +26,8 @@ private ConfigProperties() {
 
   public static final String ENGINE_HIVE_ENABLED = "iceberg.engine.hive.enabled";
   public static final String KEEP_HIVE_STATS = "iceberg.hive.keep.stats";
+  public static final String CACHE_MAX_CONTENT_LENGTH = "iceberg.cache.max-content-length";
```
rdblue, line 29: This should not be configured through Hadoop. Instead, it should be configured with catalog properties.

core/src/main/java/org/apache/iceberg/hadoop/CachingHadoopFileIO.java
rizaon: Apologies, you're right. This should be generic. I fix this by making it wrap another FileIO in [8cf5196](https://github.com/apache/iceberg/pull/4518/commits/8cf5196b80bbd0bbd41db23dd072df2533c4be71).

core/src/main/java/org/apache/iceberg/hadoop/CachingHadoopTables.java
rizaon: This is now dropped. We are now testing directly against `CachingFileIO`.

rizaon: This is now dropped. We are now testing directly against `CachingFileIO`.

core/src/main/java/org/apache/iceberg/hadoop/ConfigProperties.java
rizaon: Done

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -53,6 +56,57 @@ private CatalogProperties() {
...
   public static final long CACHE_EXPIRATION_INTERVAL_MS_OFF = -1;
 
+  /**
+   * When using {@link CachingFileIO} as io-impl, define the internal FileIO that is
+   * wrapped by CachingFileIO.
+   * <p>
+   * Default to {@link HadoopFileIO} if not defined.
+   */
+  public static final String CONTENT_CACHE_WRAPPED_FILE_IO_IMPL = "cache.content.wrapped-io-impl";
+  public static final String CONTENT_CACHE_WRAPPED_FILE_IO_IMPL_DEFAULT = HadoopFileIO.class.getName();
```
rdblue, line 66: I don't think that configuration needs to be this complicated. Instead of replacing the `FileIO` implementation and adding a second dynamic load, I would prefer using `cache-enabled=true` and handling this where we instantiate `FileIO`.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -53,6 +56,57 @@ private CatalogProperties() {
...
+   * Behavior of specific values of cache.expiration-interval-ms:
+   * <ul>
+   *   <li> Negative Values - Caching disabled</li>
+   *   <li> Zero - Cache entries expires only if it gets evicted due to memory pressure from
+   *        {@link #CONTENT_CACHE_MAX_TOTAL_BYTES} setting.
+   *   </li>
+   *   <li> Positive Values - Cache entries expire if not accessed via the cache after this many milliseconds</li>
+   * </ul>
+   */
+  public static final String CONTENT_CACHE_EXPIRATION_INTERVAL_MS = "cache.content.expiration-interval-ms";
```
rdblue, line 80: `content` seems unnecessary. What else would be cached?

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rizaon: `cache-enabled` is already taken by `CachingCatalog`

https://github.com/apache/iceberg/blob/ed54952b51a6cdc7d2350b8223c276421845a8e2/core/src/main/java/org/apache/iceberg/CatalogProperties.java#L39

Is it OK to enable both `CachingCatalog` and `CachingFileIO` by sharing this property key? Or should I make new key like `file-cache-enabled=true`?

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rizaon: I see catalog properties already have key `CACHE_ENABLED` and `CACHE_EXPIRATION_INTERVAL_MS` tied to `CachingCatalog` (which has `Cache<TableIdentifier, Table>` inside). Thus, I prefix this with `CONTENT_` to signify that this is different kind of caching. It is caching the file content (bytes) and not file handle or other java objects.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rdblue: I think this would be an `io.cache-enabled` property.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+
+/**
+ * Implementation of FileIO that adds metadata content caching features.
+ * <p>
+ * This FileIO intend to speedup scan planning by caching most of the table metadata contents in memory and reduce
+ * remote reads.
+ */
+public class CachingFileIO implements FileIO, HadoopConfigurable {
+  private static final Logger LOG = LoggerFactory.getLogger(CachingFileIO.class);
+  private static ContentCache sharedCache;
```
rdblue, line 55: Is there a way to pass in the cache rather than making it static?

@jackye1995, do you have any ideas here?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+        CatalogProperties.CONTENT_CACHE_MAX_CONTENT_LENGTH_DEFAULT);
+
+    if (durationMs >= 0 && totalBytes > 0) {
+      return new ContentCache(durationMs, totalBytes, contentLength);
+    } else {
+      return null;
+    }
+  }
+
+  public static synchronized ContentCache getSharedCache(Map<String, String> properties) {
```
rdblue, line 75: Iceberg avoids using `get` in method names. In most cases it is unnecessary and not adding value. In other cases it can be replaced with a more specific verb. `getOrCreate` seems to work here, if we end up keeping this method.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+   */
+  public CachingFileIO() {
+  }
+
+  public CachingFileIO(Configuration hadoopConf) {
+    this(new SerializableConfiguration(hadoopConf)::get);
+  }
+
+  public CachingFileIO(SerializableSupplier<Configuration> hadoopConf) {
+    this.hadoopConf = hadoopConf;
```
rdblue, line 100: Why does this reference a Hadoop Configuration? It seems like this doesn't need to rely on Hadoop.

rdblue: Nevermind, I see this hasn't been refactored to pass in the inner `FileIO` yet.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+      } catch (IOException ex) {
+        throw new RuntimeIOException(ex);
+      }
+    }
+
+    private SeekableInputStream cachedStream() throws IOException {
+      try {
+        byte[] buf = contentCache.cache().get(location(), k -> getFileContent());
+        LOG.debug("Cache stats: {}", contentCache.cache().stats());
+        return new SeekableCachedInputStream(new SeekableByteArrayInputStream(buf));
```
rdblue, line 323: I think this should use a generic in-memory input stream, rather than one that is specific to caching.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+        throw ex.getCause();
+      }
+    }
+
+    private byte[] getFileContent() {
+      try {
+        int totalByteRead = 0;
+        boolean eos = false;
+        int fileLength = (int) getLength();
+        byte[] buf = new byte[fileLength];
```
rdblue, line 299: I would prefer not to use a single byte[] allocation. If `getLength` is incorrect (which can evidently be the case for some encrypted streams) then we need to be able to handle that case. Plus, this could easily result in humongous allocations in the JVM that are hard to garbage collect. It's better to allocate ByteBuffer instances as needed and use a stream like Parquet's `ByteBufferInputStream` to stitch them together.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+    public long getLength() {
+      byte[] buf = contentCache.cache.getIfPresent(location());
+      return (buf != null) ? buf.length : wrappedInputFile.getLength();
+    }
+
+    @Override
+    public SeekableInputStream newStream() {
+      try {
+        if (getLengthChecked() <= contentCache.maxContentLength()) {
+          return cachedStream();
```
rdblue, line 264: Looks like the intent is to implement a read-through cache. Can you clarify that in the Javadoc for this `FileIO`?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    private byte[] getFileContent() {
+      try {
+        int totalByteRead = 0;
+        boolean eos = false;
+        int fileLength = (int) getLength();
+        byte[] buf = new byte[fileLength];
+        SeekableInputStream stream = wrappedInputFile.newStream();
+        while (totalByteRead < fileLength && !eos) {
```
rdblue, line 301: Can this use `ByteStreams.readFully` instead of implementing its own `readFully` method?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rdblue: Yeah, thinking about this more, I think it should allocate small buffers. In a loop, allocate a buffer that is 4 or 8 MB, then call `ByteStreams.readFully` to fill the buffer, and then wrap it in a `ByteBuffer`. Then use `ByteBufferInputStream` and the associated classes from Parquet (probably just copy those into `util` along with tests since we will use them in other places).

You may need a cache entry to have a fast size lookup for the caffeine `Cache`.

rizaon: Are you referring to this class?
https://github.com/apache/parquet-mr/blob/master/parquet-common/src/main/java/org/apache/parquet/bytes/ByteBufferInputStream.java

I look around, but can't seem to find class that can adapt that `ByteBufferInputStream` to a `SeekableInputStream` needed for `newStream()` later. Can you confirm that I should copy that class (possibly with `SingleBufferInputStream` and `MultiBufferInputStream` as well), and implement `seek()` on it so it can be adapted to `SeekableInputStream`?

rdblue:
@rizaon, there are a couple of PRs that should help you with this. #4608 adds `IOUtil.readFully` and `IOUtil.readRemaining` that you can use to fill buffers from an input stream. I've also opened #4623 to add the `ByteBufferInputStream` classes I was talking about. Those should allow you to easily split the data across several small allocations and wrap them in an `InputStream` implementation.

rizaon:
Got it, thank you. Will rebase and update this PR once they are merged.

rdblue:
@rizaon, the other PRs are in so you should be able to take advantage of `IOUtil.readFully` and the `ByteBufferInputStream` implementations. I hope that helps!

rdblue:
@findepi, @alexjo2144, FYI. This could be an option for Trino performance.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rizaon: I renamed these with `IO_CACHE_` and `io.cache.` prefix to match with `io.cache-enabled`.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: This remain static in 47b8008. Please let me know if there is a better way to access the cache from Catalog object or outside.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: This is now removed.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Refactored to use `ByteBufferInputStream`.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Refactored to use `IOUtil.readRemaining`. However, it still relies on `getLength` being accurate. Please let me know if it should keep looping until EOF met.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Refactored to use `IOUtil.readRemaining`

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,363 @@
...
+          // read the stream in 8MB chunk
+          int bytesToRead = Math.min(BUFFER_CHUNK_SIZE, totalBytesToRead);
+          byte[] buf = new byte[bytesToRead];
+          int bytesRead = IOUtil.readRemaining(stream, buf, 0, bytesToRead);
+          totalBytesToRead -= bytesRead;
+
+          if (bytesRead < bytesToRead) {
+            // We are reading less than we should, possibly hitting EOF.
+            // Truncate buf and break from the loop.
+            buf = Arrays.copyOf(buf, bytesRead);
```
rdblue, line 335: Rather than copying, can't this just create the buffer with a smaller limit?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,363 @@
...
+    return fileContentCache;
+  }
+
+  public long cacheEstimatedSize() {
+    return fileContentCache == null ? 0 : fileContentCache.cache.estimatedSize();
+  }
+
+  public static class CacheEntry {
+    private final int length;
+    private final List<ByteBuffer> buffer;
```
rdblue, line 189: I think this should be `buffers` since it is a list.

Also, if this class is private, then it's okay to access fields directly. But if it is public then everything should use getter methods. I don't see a reason for the class to be public, right?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Do you mean to use second variant of `ByteBuffer.wrap`? Like `ByteBuffer.wrap(buf, 0, bytesRead)`?
I suppose we can, but that does not seem to reduce the backing array size?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Yes, `CacheEntry` can be private.

I just wish that `ContentCache` itself is accessible to the outside world so that query engine can force cache invalidation by themself if necessary (ie., upon [REFRESH](https://impala.apache.org/docs/build3x/html/topics/impala_refresh.html) query).

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: I'm keeping the copying method in [1b98dc2](https://github.com/apache/iceberg/pull/4518/commits/1b98dc218a0e7b2ab66dead1aa2a40b623aaba3e), but reduce the chunk size from 8MB to 4MB.

The intention is that, if we indeed unexpectedly hit EOF early, the worst case is we are doing slightly under 4MB copy rather than slightly under 8MB copy. But if `getLength` is accurate, we should not reach this copying code.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Done

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -272,6 +273,13 @@ public static FileIO loadFileIO(
     configureHadoopConf(fileIO, hadoopConf);
 
     fileIO.initialize(properties);
+
+    // Check if we should wrap this with CachingFileIO
```
rdblue, line 277: Minor: personal pronouns like "I" and "we" usually make comments longer and harder to read. More direct statements like `Check whether to wrap with CachingFileIO` are preferred.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+    if (wrappedIO != null) {
+      wrappedIO.close();
+      wrappedIO = null;
+    }
+  }
+
+  @Override
+  public InputFile newInputFile(String path) {
+    InputFile inFile = wrappedIO.newInputFile(path);
+    if (fileContentCache != null && path.contains("/metadata/")) {
```
rdblue, line 154: Should this be a config setting since there is a TODO item about it? A regex could be really useful.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+
+  @Override
+  public void deleteFile(String path) {
+    wrappedIO.deleteFile(path);
+    if (fileContentCache != null) {
+      fileContentCache.cache.invalidate(path);
+    }
+  }
+
+  public ContentCache getCache() {
```
rdblue, line 178: Style: `get` in a method name should be omitted.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+    return fileContentCache;
+  }
+
+  public long cacheEstimatedSize() {
```
rdblue, line 182: This should be `estimatedCacheSize`, right? This isn't caching the size.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+  @Override
+  public OutputFile newOutputFile(String path) {
+    return wrappedIO.newOutputFile(path);
+  }
+
+  @Override
+  public void deleteFile(String path) {
+    wrappedIO.deleteFile(path);
+    if (fileContentCache != null) {
+      fileContentCache.cache.invalidate(path);
```
rdblue, line 174: If this is going to reach into `fileContentCache` to call methods on `cache` directly, why have `fileContentCache`? Just to check whether it is null? If it is a useful class, I would expect this to call its API instead of modifying its internal state.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+      return wrappedInputFile.location();
+    }
+
+    @Override
+    public boolean exists() {
+      CacheEntry buf = contentCache.cache.getIfPresent(location());
+      return buf != null || wrappedInputFile.exists();
+    }
+
+    public long getLengthChecked() throws IOException {
```
rdblue, line 301: The only place this is used catches `IOException` to re-wrap it in `RuntimeIOException`. How about just calling `getLength()` instead of having this method?

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+      try {
+        return getLength();
+      } catch (RuntimeIOException ex) {
+        throw ex.getCause();
+      }
+    }
+
+    private CacheEntry getFileContent() {
```
rdblue, line 309: Style: unnecessary `get` in method name.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+      try {
+        int fileLength = (int) getLength();
```
rdblue, line 311: I don't think this should cast to `int`. The length should be tracked as a `long`. I know that we expect `maxContentLength` to be much shorter, but even that is tracked as a `long`. I think this should just use a `long` instead of risking truncation.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+        int totalBytesToRead = fileLength;
+        SeekableInputStream stream = wrappedInputFile.newStream();
+        List<ByteBuffer> buffers = Lists.newArrayList();
+
+        while (totalBytesToRead > 0) {
```
rdblue, line 316: I think this check is good. You could keep reading buffers until one is under-sized, but then you have to worry about weird cases like when the content fits exactly in some number of buffers. This should be fine.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
```diff
@@ -0,0 +1,355 @@
...
+          int bytesToRead = Math.min(BUFFER_CHUNK_SIZE, totalBytesToRead);
+          byte[] buf = new byte[bytesToRead];
+          int bytesRead = IOUtil.readRemaining(stream, buf, 0, bytesToRead);
+          totalBytesToRead -= bytesRead;
+
+          if (bytesRead < bytesToRead) {
+            // We are reading less than we should, possibly hitting EOF.
+            // Truncate buffer and break from the loop. In worst case, this will do byte copying of slightly under 4MB.
+            byte[] smallerBuf = new byte[bytesRead];
+            System.arraycopy(buf, 0, smallerBuf, 0, bytesRead);
```
rdblue, line 327: Looks like this wasn't updated yet -- I'll come back this weekend after you've had a chance to update.

danielcweeks:
@rizaon and @rdblue I made a comment in one of the replies above but wanted to ask about potentially bad behaviors around reading metadata tables.  Since in those cases the data files are really the metadata files, the caching would trigger.  I feel like this could contribute to some interesting problems with high churn in those cases both in the driver and potentially the worker side.

It might be better just to wrap on access where we read metadata files in `MetadataFiles.java`.  That would also avoid guessing at the pathing for metadata files.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Hi @danielcweeks , thank you for your feedback!

I can't seem to find an easy way to pass catalog properties (where we define caching related knobs) to ManifestFiles.java. How about we let the CachingFileIO wrapping just like now, but create new method, say `newMetadataInputFile()`, that specially used by `ManifestReader.read()`?

Maybe modification like this in ManifestFiles.java:85
```
InputFile file = (io instanceof CachingFileIO)
    ? ((CachingFileIO) io).newMetadataInputFile(manifest.path())
    : io.newInputFile(manifest.path());
```

For the `newInputFile()` implementation by CachingFileIO, we can just pass-through it to the inner FileIO. Thus, it ensure that caching only applies to metadata files.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
steveloughran: something bigger than an int long should automatically be excluded from caching, presumably. though if someone did want to declare a cache of > 2GB then it may be needed

core/src/main/java/org/apache/iceberg/CatalogUtil.java
rizaon: Done

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Done

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Done

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Added more API method to avoid accessing internal `cache`.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: After looking around, I ended up go the regex route.

There seems to be many call sites where Iceberg is reading metadata files, not just in ManifestFiles.java. Among them are:
```
$ git grep -e "newInputFile.*manifest" -e "newInputFile.*meta"
core/src/main/java/org/apache/iceberg/AllManifestsTable.java:              .withInputFile(io.newInputFile(snap.manifestListLocation()))
core/src/main/java/org/apache/iceberg/AllManifestsTable.java:              io.newInputFile(tableOps().current().metadataFileLocation()),
core/src/main/java/org/apache/iceberg/AllManifestsTable.java:          .read(io.newInputFile(manifestListTask.file().path().toString()))
core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java:    InputFile toCopy = ops.io().newInputFile(manifest.path());
core/src/main/java/org/apache/iceberg/BaseSnapshot.java:      this.allManifests = ManifestLists.read(io.newInputFile(manifestListLocation));
core/src/main/java/org/apache/iceberg/FastAppend.java:    InputFile toCopy = ops.io().newInputFile(manifest.path());
core/src/main/java/org/apache/iceberg/HistoryTable.java:        ops.io().newInputFile(ops.current().metadataFileLocation()),
core/src/main/java/org/apache/iceberg/ManifestFiles.java:    InputFile file = io.newInputFile(manifest.path());
core/src/main/java/org/apache/iceberg/ManifestFiles.java:    InputFile file = io.newInputFile(manifest.path());
core/src/main/java/org/apache/iceberg/ManifestsTable.java:        ops.io().newInputFile(location != null ? location : ops.current().metadataFileLocation()),
core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java:    InputFile toCopy = ops.io().newInputFile(manifest.path());
core/src/main/java/org/apache/iceberg/PartitionsTable.java:          io().newInputFile(ops.current().metadataFileLocation()),
core/src/main/java/org/apache/iceberg/PartitionsTable.java:          io().newInputFile(ops.current().metadataFileLocation()),
core/src/main/java/org/apache/iceberg/RemoveSnapshots.java:      return Avro.read(ops.io().newInputFile(snapshot.manifestListLocation()))
core/src/main/java/org/apache/iceberg/SnapshotsTable.java:        ops.io().newInputFile(ops.current().metadataFileLocation()),
hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java:    InputFile metadataFile = fileIO.newInputFile(metadataFileLocation);
```

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Removed `getLengthChecked()`. However, without checking the cause, `newStream()` will throw `RuntimeIOException` instead of `NotFoundException`.

Updated assertion at `CachingHiveTableTest.testMissingMetadataWontCauseHang` accordingly.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Renamed to `cacheEntry()`.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Changed to `long`.

core/src/main/java/org/apache/iceberg/io/CachingFileIO.java
rizaon: Updated by wrapping it with smaller ByteBuffer limit instead of truncating the array.

melin:
CachedFileIO caches the data accessed for the first time. For the mpp database, it affects the corresponding time of the first access. Can snapshot changes be notified to CachedFileIO? Clean cached data, metadata can be cached for a long time. 
All metadata is cached in dremio and updated regularly. Snapshot changes, update time is not timely, avoid reading metadata and cause query delay

steveloughran:
> it affects the corresponding time of the first access. 

if you are talking about any accessed timestamps in filesystems, know the cloud stores don't have them. and s3 objects creation times are when a multipart upload was initiated, not completed. not relevant for smaller files, but for many mb/gb ones, it is

rizaon:
Hi @danielcweeks , given the immutability of snapshot and metadata, does the "potentially bad behaviors around reading metadata tables" still a concern? Immutability should make consistency easy to manage (can just evict cache entry when space is needed without having to worry about invalidating on update, etc).

About "high churn" in terms of high eviction rates, I'd like to understand what scenario can lead to that issue. Within a single table snapshot, what is the largest amount of manifest files and their individual size have been observed before?

rizaon:
One way to avoid churn / pressure in cache is to detect what is the reason of most recent cache eviction. Say, if CachingFileIO tend to evict most entries in short time due to total memory size constraint, CachingFileIO should consider just bypassing cache for some period or until some entries evicted due to expiration rather than size limit.

We can do that with caffeine cache by implementing RemovalListener
https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/2.3.1/com/github/benmanes/caffeine/cache/RemovalListener.html

Another way is to impose high watermark limit on cache total weight. Thus, bypassing the cache once this high watermark limit exceeded.

danielcweeks:
Hey @rizaon, some recent changes actually exposed FileIO properties, which I believe would simplify how we can cache manifest files.

Where we [read manifest files](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/ManifestFiles.java#L85), we now have the file legnth and properties from the FileIO.  It seems like we could just add a simple cache to the ManifestFiles and allow caching configuration via `FileIO::properties()`.

I feel like that would really simplify this and avoid depending on things like regex+pathing to isolate which files to cache.

rizaon:
Hi @danielcweeks , thank you for the update. I will check PR #5207 and study how we can utilize it.

So with that PR, am I correct that we should implement the caching in ManifestFiles.java instead of creating a new CachingFileIO.java as this PR did?

danielcweeks:
@rizaon because we want to isolate the caching to metadata files only, I think MetadataFiles.java would be a good place to start.  We might want some utilities so if we find we want to cache FileIO elsewhere, we can reuse the same approach, but this seems like a full FileIO implementation isn't necessary (and introduces other problems).

rizaon:
@danielcweeks [4d4c4fa](https://github.com/apache/iceberg/commit/4d4c4fac0ab7b5c99aeaa3c6d3fb03d9dffa27c9) implement the manifest file cache in ManifestFiles.java. Since ManifestFiles is stateless and we might deal with multiple FileIO, the cache is implemented as `Cache<FileIO, ContentCache>` (nested cache). It expire a FileIO's ContentCache after 1 hour of inactivity or `dropCache(FileIO)` is called.

rizaon:
337bfe8 is a rebase + `spotlessApply` of 4d4c4fa

rizaon:
Hello @danielcweeks, can I get a new feedback from my latest rebase please?

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -300,4 +328,14 @@ private static ManifestFile copyManifestInternal(
 
     return writer.toManifestFile();
   }
+
+  private static InputFile newInputFile(FileIO io, String path, long length) {
+    InputFile inputFile = io.newInputFile(path, length);
```
pvary, line 333: I have seen cases where calling `newInputFile` was the bottleneck, since the HadoopFileIo created a new `FileSystem` object every time. Why did we decide to key the cache with the `inputFile` instead of the `path`?

rizaon: `inputFile` is not used as cache key, but rather as a fallback within `CachingInputFile` class (field `wrappedInputFile `) if `path` does not exist in `ContentCache` yet.

I understand that `HadoopInputFile.fromLocation()` will create a new `FileSystem` upon initialization. Maybe I can tweak the code to lazily instantiate `wrappedInputFile` by passing `io`, `path`, and `length` instead as constructor parameters of `CachingInputFile`?

rizaon: [d86461e](https://github.com/apache/iceberg/pull/4518/commits/d86461ece6643f86a051b4e38fa00d76bcec31d1) address this improvement.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,286 @@
...
+        if (getLength() <= contentCache.maxContentLength()) {
+          return cachedStream();
+        }
+
+        // fallback to non-caching input stream.
+        return wrappedInputFile().newStream();
+      } catch (FileNotFoundException e) {
+        throw new NotFoundException(e, "Failed to open input stream for file: %s", location);
+      } catch (IOException e) {
+        throw new RuntimeIOException(e, "Failed to open input stream for file: %s", location);
```
steveloughran, line 226: could you add e.toString to the error ext, as there's always the risk that something (hive...) will not print the innermost exception after wrapping things a few more times.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    private CacheEntry cacheEntry() {
+      try {
+        long fileLength = getLength();
+        long totalBytesToRead = fileLength;
+        SeekableInputStream stream = wrappedInputFile().newStream();
+        List<ByteBuffer> buffers = Lists.newArrayList();
+
+        while (totalBytesToRead > 0) {
```
steveloughran, line 248: might be good to log at debug this/its duration

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,286 @@
...
+          // read the stream in 4MB chunk
+          int bytesToRead = (int) Math.min(BUFFER_CHUNK_SIZE, totalBytesToRead);
+          byte[] buf = new byte[bytesToRead];
+          int bytesRead = IOUtil.readRemaining(stream, buf, 0, bytesToRead);
+          totalBytesToRead -= bytesRead;
+
+          if (bytesRead < bytesToRead) {
+            // Read less than it should be, possibly hitting EOF.
```
steveloughran, line 256: if less came back than expected, that's possibly a sign of a serious problem (file overwrite?).

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,286 @@
...
+            // Set smaller ByteBuffer limit and break out of the loop.
+            buffers.add(ByteBuffer.wrap(buf, 0, bytesRead));
+            break;
+          } else {
+            buffers.add(ByteBuffer.wrap(buf));
+          }
+        }
+
+        stream.close();
```
steveloughran, line 265: must be in a finally clause or try-with-resources, otherwise s3a will keep the http connection around.

steveloughran (commented):
reviewed the IO section...some opportunities to harden it there

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -43,6 +47,30 @@ private ManifestFiles() {}
               GenericManifestFile.class.getName(),
               ManifestFile.PARTITION_SUMMARY_TYPE,
               GenericPartitionFieldSummary.class.getName()));
+  private static final int CONTENT_CACHES_EXPIRATION_INTERVAL_SECOND = 60 * 60; // 1 hour
```
danielcweeks, line 50: It seems like if the content cache lives for only 60s and this cache expires after an hour, we're holding on to a lot of FileIO handles for much longer than necessary.

Is there a reason why this need to be so long?  Also it doesn't look like we have an upper bound on this cache, which could be problematic due to how much manifest IO activity can happen during any planning phase.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Done in [f76fa5d](https://github.com/apache/iceberg/pull/4518/commits/f76fa5d7f8257c63fd90df32fd841ac989cb26af).

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Done in [f76fa5d](https://github.com/apache/iceberg/pull/4518/commits/f76fa5d7f8257c63fd90df32fd841ac989cb26af).

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: [f76fa5d](https://github.com/apache/iceberg/pull/4518/commits/f76fa5d7f8257c63fd90df32fd841ac989cb26af) change this by abandoning content caching if bytesRead is less than bytesToRead.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: [f76fa5d](https://github.com/apache/iceberg/pull/4518/commits/f76fa5d7f8257c63fd90df32fd841ac989cb26af) change this by using try-with-resource.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
rizaon: [f76fa5d](https://github.com/apache/iceberg/pull/4518/commits/f76fa5d7f8257c63fd90df32fd841ac989cb26af) change `CONTENT_CACHE` to evict by size and reference.

Since content caching is per catalog/FileIO, different FileIO might have different caching configuration, including expiration time. This expiration time can be longer than 60s. Thus, using time-based eviction for  `CONTENT_CACHE` might conflict with with per-FileIO cache configuration. I don't think we can use maximum weight as well for `CONTENT_CACHE` since according to the [documentation](https://github.com/ben-manes/caffeine/wiki/Eviction) entry's weight is only calculated on entry creation and update time, and static thereafter. All `CONTENT_CACHE` entries then  will have 0 weight initially and potentially stay that way.

[f76fa5d](https://github.com/apache/iceberg/pull/4518/commits/f76fa5d7f8257c63fd90df32fd841ac989cb26af) add default upper bound to 8 FileIO to cache (no specific reason about why it should be 8, I'm open to change it). Additionally, it also use `weakKeys()` to allow garbage collection of an entry if there are no other strong references to the keys.

steveloughran (commented):
changes all LGTM; thanks for doing them

rizaon:
Core test failed at `HadoopFileIOTest > testHadoopFileIOKryoSerialization()`. Should I do rebase to pick up the fix from #5437 ?

rizaon:
[bb737ab](https://github.com/apache/iceberg/pull/4518/commits/bb737ab85f7f1585d0058d09572e9b965da3bb53) is a rebase + squash of all review so far.

[15d67b3](https://github.com/apache/iceberg/pull/4518/commits/15d67b3c41bada725c566ac3fa14aa71127c17a0) is a fix for kyro serialization failure in `testHadoopFileIOKryoSerialization()`.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,300 @@
...
+    long contentLength =
+        PropertyUtil.propertyAsLong(
+            properties,
+            CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH,
+            CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH_DEFAULT);
+
+    if (enabled && durationMs >= 0 && totalBytes > 0 && contentLength > 0) {
+      return new ContentCache(durationMs, totalBytes, contentLength);
+    } else {
+      return null;
```
pvary, line 70: Would it worth to add a DEBUG log here why the result is `null`?

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,300 @@
...
+    cache.cleanUp();
+  }
+
+  public long estimatedCacheSize() {
+    return cache.estimatedSize();
+  }
+
+  @Override
+  public String toString() {
+    return getClass().getSimpleName()
```
pvary, line 148: AFAIK  we tend to use `MoreObjects.toStringHelper(..)` for generating `toString` methods

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,300 @@
...
+      this.io = io;
+      this.location = location;
+      this.length = length;
+    }
+
+    private InputFile wrappedInputFile() {
+      if (fallbackInputFile == null) {
+        fallbackInputFile = io.newInputFile(location, length);
+      }
+      return fallbackInputFile;
```
pvary, line 192: How long do we keep this reference for the `InputFile`. Does this have any reference for important resources?

rizaon: This `InputFile` reference is kept as long as the enclosing `CachingInputFile`. The `CachingInputFile` itself is not cached. Only the byte buffers of the input stream are cached on the first call to `CachingInputFile#cachedStream()`

pvary: Sounds reasonable to me

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Fixed in 99057de

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Fixed in 99057de

pvary (approved):

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+          } else {
+            buffers.add(ByteBuffer.wrap(buf));
+          }
+        }
+
+        CacheEntry newEntry = new CacheEntry(fileLength - totalBytesToRead, buffers);
+        LOG.debug("cacheEntry took {} ms for {}", (System.currentTimeMillis() - start), location);
+        return newEntry;
+      } catch (IOException ex) {
+        throw new RuntimeIOException(ex);
```
danielcweeks, line 292: (Deprecation) switch to `UncheckedIOException`

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+                    + " bytes, but only "
+                    + (fileLength - totalBytesToRead)
+                    + " bytes "
+                    + "read.");
+          } else {
+            buffers.add(ByteBuffer.wrap(buf));
+          }
+        }
+
+        CacheEntry newEntry = new CacheEntry(fileLength - totalBytesToRead, buffers);
```
danielcweeks, line 288: `totalBytesToRead` should always be zero so this should just be `fileLength`

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+          return cachedStream();
+        }
+
+        // fallback to non-caching input stream.
+        return wrappedInputFile().newStream();
+      } catch (FileNotFoundException e) {
+        throw new NotFoundException(
+            e, "Failed to open input stream for file %s: %s", location, e.toString());
+      } catch (IOException e) {
+        throw new RuntimeIOException(
```
danielcweeks, line 243: deprecation: use UncheckedIOException

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+      return location;
+    }
+
+    @Override
+    public boolean exists() {
+      CacheEntry buf = contentCache.getIfPresent(location);
+      return buf != null || wrappedInputFile().exists();
+    }
+
+    private CacheEntry cacheEntry() {
```
danielcweeks, line 259: Something to think about (possibly create an issue), but this isn't really read-through, but rather read-ahead since we're reading the whole file before allowing the reader to process the manifest file.  We could improve this by making it a read-through buffer.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+import org.apache.iceberg.exceptions.NotFoundException;
+import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.util.PropertyUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ContentCache {
```
danielcweeks, line 42: ContentCache is rather generic and I feel like we should have some javadoc explaining what this cache is and how it is used.

danielcweeks (changes requested):
@rizaon just added some feedback.  Take a look and let me know if something doesn't make sense.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+    this.maxContentLength = maxContentLength;
+
+    Caffeine<Object, Object> builder = Caffeine.newBuilder();
+    if (expireAfterAccessMs > 0) {
+      builder = builder.expireAfterAccess(Duration.ofMillis(expireAfterAccessMs));
+    }
+
+    this.cache =
+        builder
+            .maximumWeight(maxTotalBytes)
```
danielcweeks, line 109: should we include weak keys/values here as well?

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -57,6 +58,64 @@ private CatalogProperties() {}
...
+   * <p>This value will be ignored and the file cache will be disabled if any of the following is
+   * true:
+   *
+   * <ul>
+   *   <li>{@link #IO_CACHE_EXPIRATION_INTERVAL_MS} is set to negative number.
+   *   <li>{@link #IO_CACHE_MAX_TOTAL_BYTES} is set to non-positive value.
+   *   <li>{@link #IO_CACHE_MAX_CONTENT_LENGTH} is set to non-positive value.
+   * </ul>
+   */
+  public static final String IO_CACHE_ENABLED = "io.cache-enabled";
```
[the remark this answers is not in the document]
rizaon: Done.

core/src/main/java/org/apache/iceberg/SystemProperties.java
```diff
@@ -33,6 +33,14 @@ private SystemProperties() {}
   /** Whether to use the shared worker pool when planning table scans. */
   public static final String SCAN_THREAD_POOL_ENABLED = "iceberg.scan.plan-in-worker-pool";
 
+  /**
+   * Maximum number of distinct {@link org.apache.iceberg.io.FileIO} that is allowed to have
+   * associated {@link org.apache.iceberg.io.ContentCache} in memory at a time.
+   */
+  public static final String IO_CACHE_MAX_FILEIO = "iceberg.io.cache.max-fileio";
```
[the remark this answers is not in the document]
rizaon: Done

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+        PropertyUtil.propertyAsLong(
+            properties,
+            CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH,
+            CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH_DEFAULT);
+
+    if (!enabled) {
+      LOG.debug("No ContentCache created. {} is false.", CatalogProperties.IO_CACHE_ENABLED);
+    } else if (durationMs < 0) {
+      LOG.debug(
+          "No ContentCache created. {} ({}) is less than 0.",
```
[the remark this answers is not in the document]
rizaon: Done.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -300,4 +329,30 @@ private static ManifestFile copyManifestInternal(
 
     return writer.toManifestFile();
   }
+
+  private static InputFile newInputFile(FileIO io, String path, long length) {
+    try {
+      ContentCache cache = getOrCreateCache(io);
```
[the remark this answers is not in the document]
rizaon: Done.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -44,6 +51,28 @@ private ManifestFiles() {}
...
+  /**
+   * Get or create a manifest file cache for a given FileIO.
+   *
+   * <p>Returned cache object will be kept in memory until {@link #dropCache(FileIO)} is called.
+   * Return null if the given FileIO properties does not allow caching.
+   *
+   * @param fileIO A FileIO.
+   * @return A manifest file cache or null if the given FileIO properties does not allow caching.
+   */
+  public static ContentCache getOrCreateCache(FileIO fileIO) {
```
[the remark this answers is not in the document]
rizaon: This is still useful for testing, to gain access to values of `CONTENT_CACHES`. I made this to package private and rename it to `contentCache(FileIO io)`.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
```diff
@@ -0,0 +1,310 @@
...
+            properties,
+            CatalogProperties.IO_CACHE_MAX_TOTAL_BYTES,
+            CatalogProperties.IO_CACHE_MAX_TOTAL_BYTES_DEFAULT);
+    long contentLength =
+        PropertyUtil.propertyAsLong(
+            properties,
+            CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH,
+            CatalogProperties.IO_CACHE_MAX_CONTENT_LENGTH_DEFAULT);
+
+    if (!enabled) {
```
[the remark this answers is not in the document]
rizaon: Done.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Added small documentation in `ContentCache`, `ContentCache.tryCache()`, and `CachingInputFile`.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -44,6 +51,28 @@ private ManifestFiles() {}
               ManifestFile.PARTITION_SUMMARY_TYPE,
               GenericPartitionFieldSummary.class.getName()));
 
+  private static final Cache<FileIO, ContentCache> CONTENT_CACHES =
+      Caffeine.newBuilder().weakKeys().maximumSize(maxFileIO()).recordStats().build();
```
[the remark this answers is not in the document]
rizaon: I think `softValues()` is more appropriate.

Neither `ContentCache` nor `CacheEntry` held for a long time outside its respective Caffeine cache. So using `weakValues()` will always reap them in every gc even when free memory is plenty.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Added `softValues()` to this builder.

The key still use strong reference to allow key matching by String `equals()` rather than object match (`==`).

core/src/main/java/org/apache/iceberg/ManifestFiles.java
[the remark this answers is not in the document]
rizaon: The [documentation](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/2.2.0/com/github/benmanes/caffeine/cache/Caffeine.html) says

> Entries with reclaimed keys or values may be removed from the cache on each cache modification, on occasional cache accesses, or on calls to [Cache.cleanUp()](https://www.javadoc.io/static/com.github.ben-manes.caffeine/caffeine/2.2.0/com/github/benmanes/caffeine/cache/Cache.html#cleanUp--);

Added `testWeakFileIOReferenceCleanUp()` to verify that using `weakKeys()` can still evicts cache entries.

core/src/main/java/org/apache/iceberg/io/ContentCache.java
rizaon: Filed #5814 as a follow up.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -57,6 +58,68 @@ private CatalogProperties() {}
...
+
+  public static final boolean IO_MANIFEST_CACHE_ENABLED_DEFAULT = false;
+
+  /**
+   * Controls the duration for which entries in the {@link ContentCache} are cached.
+   *
+   * <p>Behavior of specific values of cache.expiration-interval-ms:
+   *
+   * <ul>
+   *   <li>Negative Values - Caching disabled
```
danielcweeks, line 83: nit:  I don't think this is true after the changes around enabling/disabling caches.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
```diff
@@ -57,6 +58,68 @@ private CatalogProperties() {}
...
+  public static final long IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT =
+      TimeUnit.SECONDS.toMillis(60);
+
+  /**
+   * Controls the maximum total amount of bytes to cache in {@link ContentCache}.
+   *
+   * <p>Behavior of specific values of cache.content.max-total-bytes:
+   *
+   * <ul>
+   *   <li>Non-Positive Values - Caching disabled
```
danielcweeks, line 102: This also may be inconsistent with the behavior

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -300,4 +339,58 @@ private static ManifestFile copyManifestInternal(
...
+      try {
+        return Integer.parseUnsignedInt(value);
+      } catch (NumberFormatException e) {
+        // will return the default
+      }
+    }
+    return SystemProperties.IO_MANIFEST_CACHE_MAX_FILEIO_DEFAULT;
+  }
+
+  public static boolean cachingEnabled(FileIO io) {
```
danielcweeks, line 369: This can be package level protection.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -300,4 +339,58 @@ private static ManifestFile copyManifestInternal(
...
+    return PropertyUtil.propertyAsBoolean(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_ENABLED,
+        CatalogProperties.IO_MANIFEST_CACHE_ENABLED_DEFAULT);
+  }
+
+  public static long cacheDurationMs(FileIO io) {
```
danielcweeks, line 376: This can be package level protection.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -300,4 +339,58 @@ private static ManifestFile copyManifestInternal(
...
+    return PropertyUtil.propertyAsLong(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS,
+        CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT);
+  }
+
+  public static long cacheTotalBytes(FileIO io) {
```
danielcweeks, line 383: This can be package level protection.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
```diff
@@ -300,4 +339,58 @@ private static ManifestFile copyManifestInternal(
...
+    return PropertyUtil.propertyAsLong(
+        io.properties(),
+        CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES,
+        CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES_DEFAULT);
+  }
+
+  public static long cacheMaxContentLength(FileIO io) {
```
danielcweeks, line 390: This can be package level protection.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rizaon: Clarified the documentation.

core/src/main/java/org/apache/iceberg/CatalogProperties.java
rizaon: Clarified the documentation.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
rizaon: Done.

core/src/main/java/org/apache/iceberg/ManifestFiles.java
rizaon: Done.

danielcweeks:
Hey @rizaon I think this is getting real close now (thanks for sticking with us on this).  Now that we've gotten this far, I see there's a little more we can do to consolidate the logic in ContentCache.  For example, if you pull the property configuration into ContentCache constructor, we can remove a lot from ManifestFiles.  For example:

```java
 public ContentCache(FileIO fileio) {
    this.fileio = fileio;

    long expireAfterAccessMs = PropertyUtil.propertyAsLong(
        fileio.properties(),
        CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS,
        CatalogProperties.IO_MANIFEST_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT);

    long maxTotalBytes = PropertyUtil.propertyAsLong(
        fileio.properties(),
        CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES,
        CatalogProperties.IO_MANIFEST_CACHE_MAX_TOTAL_BYTES_DEFAULT);

    long maxContentLength = PropertyUtil.propertyAsLong(
        fileio.properties(),
        CatalogProperties.IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH,
        CatalogProperties.IO_MANIFEST_CACHE_MAX_CONTENT_LENGTH_DEFAULT);

    ValidationException.check(expireAfterAccessMs >= 0, "expireAfterAccessMs is less than 0");
    ValidationException.check(maxTotalBytes > 0, "maxTotalBytes is equal or less than 0");
    ValidationException.check(maxContentLength > 0, "maxContentLength is equal or less than 0");
    this.expireAfterAccessMs = expireAfterAccessMs;
    this.maxTotalBytes = maxTotalBytes;
    this.maxContentLength = maxContentLength;
```

You can also pull `CONTENT_CACHES`, `contentCache(FileIO io)`, and `dropCache(FileIO fileIO)` into `ContentCache` as well.

I played around with moving more of the logic into `ContentCache` and was able to move everything (minus the `newInputFile`) logic into ContentCache, which I feel is a cleaner separation of concerns.

rizaon:
Hi @danielcweeks that looks like a nice cleanup. But I have 2 concerns:

1. By pushing the config verification down to `ContentCache` constructor, that will make `ContentCache` less generic and more specific towards manifest caching use. Is that OK?
2. If we keep `fileio` as a class member of `ContentCache`, that becomes a strong reference to `fileio`. Thus, the `weakKeys()` in `CONTENT_CACHES` is not useful anymore because there is always a strong reference to that `FileIO` key living in `ContentCache` object (the value)?

danielcweeks:
> 



> Hi @danielcweeks that looks like a nice cleanup. But I have 2 concerns:
> 
> 1. By pushing the config verification down to `ContentCache` constructor, that will make `ContentCache` less generic and more specific towards manifest caching use. Is that OK?
> 2. If we keep `fileio` as a class member of `ContentCache`, that becomes a strong reference to `fileio`. Thus, the `weakKeys()` in `CONTENT_CACHES` is not useful anymore because there is always a strong reference to that `FileIO` key living in `ContentCache` object (the value)?

Good points.  Let me take one more pass (I was just hoping to consolidate more of the logic in ContentCache, but you're probably right that it makes more sense to keep the manifest settings with ManifestFiles).

rizaon:
In the last github check runs, there was a checkstye issue and exception handling when given `FileIO` does not implement `properties()` method.

I have rebased this PR and add [ba7329d](https://github.com/apache/iceberg/pull/4518/commits/ba7329daffe383ab24b1b778079d31e91944dd5d) to fix this issue.

danielcweeks (approved):

danielcweeks:
@rizaon Thanks for your contribution here and keeping this updated!  I'm excited to see how this works out.

rizaon:
@danielcweeks thank you for accepting this PR!

PR #2926, opened 2021-08-03
Core: Predicate pushdown for files metadata table

szehon-ho:
* This re-uses the logic of https://github.com/apache/iceberg/pull/2358 (partitions metadata table)

PR #4520, opened 2022-04-07
Core: Files table predicate filtering wrong for tables with evolved partition specs

szehon-ho:
The change https://github.com/apache/iceberg/pull/2926 introduced partition predicate push down for the "files" metadata table.  However, it is incorrect in the case that the partition spec evolves and data written in both specs.  The projectionFilter is constructed by projecting the expression on the current table's partition spec, and evaluating the values against the manifest file.

Say for example, we have a table that had a former partition spec that is "data", and a new partition spec that is "id", with manifests written with both specs.

A query like "select * from my_table.files where files.partition.id=1" projects successfully against the current partition spec "id" to have a ManifestEvaluator looking for value 1.  It looks for this against all manifest-files, even those written with the old partition-spec "data".  This erroneously filters those old manifests where data is 1, and correctly filters new manifests where id is 1, giving wrong results.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
@@ -19,17 +19,19 @@
 
 package org.apache.iceberg;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import java.util.Map;
 import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.ManifestEvaluator;
-import org.apache.iceberg.expressions.Projections;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.transforms.Transforms;
 import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types.StructType;
 
@@ -92,10 +94,26 @@ abstract class BaseFilesTable extends BaseMetadataTable {
     protected CloseableIterable<FileScanTask> planFiles(TableOperations ops, Snapshot snapshot, Expression rowFilter,
                                                         boolean ignoreResiduals, boolean caseSensitive,
                                                         boolean colStats) {
-      CloseableIterable<ManifestFile> filtered =
-          CloseableIterable.filter(
-              CloseableIterable.withNoopClose(manifests()),
-              m -> filter(m, rowFilter, caseSensitive));
+      Map<Integer, PartitionSpec> specsById = table().specs();
+
+      LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
+        PartitionSpec spec = specsById.get(specId);
+        PartitionSpec transformedSpec = transformSpec(fileSchema, spec, PARTITION_FIELD_PREFIX);
+        return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
+      });
+
+      CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(
+          manifests(),
+          manifest -> {
+            PartitionSpec spec = specsById.get(manifest.partitionSpecId());
+
+            if (spec.fields().stream().anyMatch(f -> f.transform().equals(Transforms.alwaysNull()))) {
+              // In V1, removing Partition Fields meant replacing them with Void transformations
+              // We cannot use these for filtering as it looks for manifest files with 'null' partition value.
+              return true;
+            }
+            return evalCache.get(manifest.partitionSpecId()).eval(manifest);
+          });
 
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
@@ -115,21 +133,6 @@ abstract class BaseFilesTable extends BaseMetadataTable {
      * Returns an iterable of manifest files to explore for this Files metadata table scan
      */
     protected abstract CloseableIterable<ManifestFile> manifests();
-
-
-    private boolean filter(ManifestFile manifestFile, Expression rowFilter, boolean caseSensitive) {
-      if (!table().spec().equals(table().specs().get(manifestFile.partitionSpecId()))) {
-        return true;
-      }
-
-      // use an inclusive projection to remove the partition name prefix and filter out any non-partition expressions
-      PartitionSpec spec = transformSpec(fileSchema, table().spec(), PARTITION_FIELD_PREFIX);
-      Expression partitionFilter = Projections.inclusive(spec, caseSensitive).project(rowFilter);
-      ManifestEvaluator manifestEval = ManifestEvaluator.forPartitionFilter(
-          partitionFilter, spec, caseSensitive);
-
-      return manifestEval.eval(manifestFile);
-    }
   }
 
   static class ManifestReadTask extends BaseFileScanTask implements DataTask {
--- a/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
@@ -92,7 +92,10 @@ abstract class BaseFilesTable extends BaseMetadataTable {
     protected CloseableIterable<FileScanTask> planFiles(TableOperations ops, Snapshot snapshot, Expression rowFilter,
                                                         boolean ignoreResiduals, boolean caseSensitive,
                                                         boolean colStats) {
-      CloseableIterable<ManifestFile> filtered = filterManifests(manifests(), rowFilter, caseSensitive);
+      CloseableIterable<ManifestFile> filtered =
+          CloseableIterable.filter(
+              CloseableIterable.withNoopClose(manifests()),
+              m -> filter(m, rowFilter, caseSensitive));
 
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
@@ -113,17 +116,17 @@ abstract class BaseFilesTable extends BaseMetadataTable {
      */
     protected abstract CloseableIterable<ManifestFile> manifests();
 
-    private CloseableIterable<ManifestFile> filterManifests(CloseableIterable<ManifestFile> manifests,
-                                                            Expression rowFilter,
-                                                            boolean caseSensitive) {
+
+    private boolean filter(ManifestFile manifestFile, Expression rowFilter, boolean caseSensitive) {
+      PartitionSpec originalSpec = table().specs().get(manifestFile.partitionSpecId());
+
       // use an inclusive projection to remove the partition name prefix and filter out any non-partition expressions
-      PartitionSpec spec = transformSpec(fileSchema, table().spec(), PARTITION_FIELD_PREFIX);
+      PartitionSpec spec = transformSpec(fileSchema, originalSpec, PARTITION_FIELD_PREFIX);
       Expression partitionFilter = Projections.inclusive(spec, caseSensitive).project(rowFilter);
-
       ManifestEvaluator manifestEval = ManifestEvaluator.forPartitionFilter(
-          partitionFilter, table().spec(), caseSensitive);
+          partitionFilter, originalSpec, caseSensitive);
 
-      return CloseableIterable.filter(manifests, manifestEval::eval);
+      return manifestEval.eval(manifestFile);
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
@@ -118,13 +118,15 @@ abstract class BaseFilesTable extends BaseMetadataTable {
 
 
     private boolean filter(ManifestFile manifestFile, Expression rowFilter, boolean caseSensitive) {
-      PartitionSpec originalSpec = table().specs().get(manifestFile.partitionSpecId());
+      if (!table().spec().equals(table().specs().get(manifestFile.partitionSpecId()))) {
+        return true;
+      }
 
       // use an inclusive projection to remove the partition name prefix and filter out any non-partition expressions
-      PartitionSpec spec = transformSpec(fileSchema, originalSpec, PARTITION_FIELD_PREFIX);
+      PartitionSpec spec = transformSpec(fileSchema, table().spec(), PARTITION_FIELD_PREFIX);
       Expression partitionFilter = Projections.inclusive(spec, caseSensitive).project(rowFilter);
       ManifestEvaluator manifestEval = ManifestEvaluator.forPartitionFilter(
-          partitionFilter, originalSpec, caseSensitive);
+          partitionFilter, spec, caseSensitive);
 
       return manifestEval.eval(manifestFile);
     }
--- a/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
@@ -30,6 +30,7 @@ import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types.StructType;
@@ -101,7 +102,7 @@ abstract class BaseFilesTable extends BaseMetadataTable {
         return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
       });
 
-      CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(manifests(),
+      CloseableIterable<ManifestFile> filteredManifests = CloseableIterable.filter(manifests(),
           manifest -> evalCache.get(manifest.partitionSpecId()).eval(manifest));
 
       String schemaString = SchemaParser.toJson(schema());
@@ -113,9 +114,8 @@ abstract class BaseFilesTable extends BaseMetadataTable {
       // This data task needs to use the table schema, which may not include a partition schema to avoid having an
       // empty struct in the schema for unpartitioned tables. Some engines, like Spark, can't handle empty structs in
       // all cases.
-      return CloseableIterable.transform(filtered, manifest ->
-          new ManifestReadTask(ops.io(), ops.current().specsById(),
-              manifest, schema(), schemaString, specString, residuals));
+      return CloseableIterable.transform(filteredManifests, manifest ->
+          new ManifestReadTask(table(), manifest, schema(), schemaString, specString, residuals));
     }
 
     /**
@@ -130,11 +130,11 @@ abstract class BaseFilesTable extends BaseMetadataTable {
     private final ManifestFile manifest;
     private final Schema schema;
 
-    ManifestReadTask(FileIO io, Map<Integer, PartitionSpec> specsById, ManifestFile manifest,
+    ManifestReadTask(Table table, ManifestFile manifest,
                      Schema schema, String schemaString, String specString, ResidualEvaluator residuals) {
       super(DataFiles.fromManifest(manifest), null, schemaString, specString, residuals);
-      this.io = io;
-      this.specsById = specsById;
+      this.io = table.io();
+      this.specsById = Maps.newHashMap(table.specs());
       this.manifest = manifest;
       this.schema = schema;
     }
--- a/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
@@ -31,7 +31,6 @@ import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
-import org.apache.iceberg.transforms.Transforms;
 import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types.StructType;
 
@@ -102,18 +101,8 @@ abstract class BaseFilesTable extends BaseMetadataTable {
         return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
       });
 
-      CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(
-          manifests(),
-          manifest -> {
-            PartitionSpec spec = specsById.get(manifest.partitionSpecId());
-
-            if (spec.fields().stream().anyMatch(f -> f.transform().equals(Transforms.alwaysNull()))) {
-              // In V1, removing Partition Fields meant replacing them with Void transformations
-              // We cannot use these for filtering as it looks for manifest files with 'null' partition value.
-              return true;
-            }
-            return evalCache.get(manifest.partitionSpecId()).eval(manifest);
-          });
+      CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(manifests(),
+          manifest -> evalCache.get(manifest.partitionSpecId()).eval(manifest));
 
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
--- a/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFilesTable.java
@@ -98,7 +98,7 @@ abstract class BaseFilesTable extends BaseMetadataTable {
 
       LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
         PartitionSpec spec = specsById.get(specId);
-        PartitionSpec transformedSpec = transformSpec(fileSchema, spec, PARTITION_FIELD_PREFIX);
+        PartitionSpec transformedSpec = transformSpec(fileSchema, spec);
         return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
       });
 
--- a/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
@@ -35,7 +35,6 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
  * deserialization.
  */
 abstract class BaseMetadataTable implements Table, HasTableOperations, Serializable {
-  protected static final String PARTITION_FIELD_PREFIX = "partition.";
   private final PartitionSpec spec = PartitionSpec.unpartitioned();
   private final SortOrder sortOrder = SortOrder.unsorted();
   private final TableOperations ops;
@@ -52,18 +51,17 @@ abstract class BaseMetadataTable implements Table, HasTableOperations, Serializa
    * This method transforms the table's partition spec to a spec that is used to rewrite the user-provided filter
    * expression against the given metadata table.
    * <p>
-   * The resulting partition spec maps $partitionPrefix.X fields to partition X using an identity partition transform.
+   * The resulting partition spec maps partition.X fields to partition X using an identity partition transform.
    * When this spec is used to project an expression for the given metadata table, the projection will remove
-   * predicates for non-partition fields (not in the spec) and will remove the "$partitionPrefix." prefix from fields.
+   * predicates for non-partition fields (not in the spec) and will remove the "partition." prefix from fields.
    *
    * @param metadataTableSchema schema of the metadata table
    * @param spec spec on which the metadata table schema is based
-   * @param partitionPrefix prefix to remove from each field in the partition spec
    * @return a spec used to rewrite the metadata table filters to partition filters using an inclusive projection
    */
-  static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec, String partitionPrefix) {
+  static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec) {
     PartitionSpec.Builder identitySpecBuilder = PartitionSpec.builderFor(metadataTableSchema).checkConflicts(false);
-    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), partitionPrefix + pf.name(), "identity"));
+    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), pf.name(), "identity"));
     return identitySpecBuilder.build();
   }
 
--- a/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
@@ -63,7 +63,7 @@ abstract class BaseMetadataTable implements Table, HasTableOperations, Serializa
    */
   static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec, String partitionPrefix) {
     PartitionSpec.Builder identitySpecBuilder = PartitionSpec.builderFor(metadataTableSchema).checkConflicts(false);
-    spec.fields().forEach(pf -> identitySpecBuilder.identity(partitionPrefix + pf.name(), pf.name()));
+    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), partitionPrefix + pf.name(), "identity"));
     return identitySpecBuilder.build();
   }
 
--- a/core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
+++ b/core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
@@ -64,7 +64,9 @@ public class TestMetadataTableFilters extends TableTestBase {
   @Override
   public void setupTable() throws Exception {
     super.setupTable();
-    table.updateProperties().set(TableProperties.MANIFEST_MERGE_ENABLED, "false").commit();
+    table.updateProperties()
+        .set(TableProperties.MANIFEST_MERGE_ENABLED, "false")
+        .commit();
     table.newFastAppend()
         .appendFile(FILE_A)
         .commit();
@@ -258,12 +260,14 @@ public class TestMetadataTableFilters extends TableTestBase {
   public void testPartitionSpecEvolutionRemovalV1() {
     Assume.assumeTrue(formatVersion == 1);
 
-    // Change spec and add two data and delete files each
-    table.updateSpec().removeField(Expressions.bucket("data", 16))
-        .addField("id").commit();
+    // Change spec and add two data files
+    table.updateSpec()
+        .removeField(Expressions.bucket("data", 16))
+        .addField("id")
+        .commit();
     PartitionSpec newSpec = table.spec();
 
-    // Add two data files and two delete files with new spec
+    // Add two data files with new spec
     PartitionKey data10Key = new PartitionKey(newSpec, table.schema());
     data10Key.set(1, 10);
     DataFile data10 = DataFiles.builder(newSpec)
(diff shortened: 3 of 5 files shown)


core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,74 @@ public void testPlanTasks() {
     validateCombinedScanTasks(tasks, 0);
   }
 
+  @Test
+  public void testPartitionSpecEvolution() {
+    Assume.assumeTrue(formatVersion == 2);
```
szehon-ho, line 259: It is pretty tricky to have the same code work in v1 and v2 to change spec and write new files, so choosing to do this in v2 (should be orthogonal to the problem).

szehon-ho:
Thanks to @aokolnychyi for finding it in https://github.com/apache/iceberg/pull/4382#issuecomment-1087866951 and before.

FYI @RussellSpitzer @rdblue seems its bad issue for tables with changed partitions specs, as files table may return less results with a filter.  It's present in 0.12 and 0.13 release, and I think a good candidate for 0.13.2 release.  Sorry I did not check this case initially.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -113,17 +116,19 @@ public TableScan appendsAfter(long fromSnapshotId) {
      */
     protected abstract CloseableIterable<ManifestFile> manifests();
 
-    private CloseableIterable<ManifestFile> filterManifests(CloseableIterable<ManifestFile> manifests,
-                                                            Expression rowFilter,
-                                                            boolean caseSensitive) {
+
+    private boolean filter(ManifestFile manifestFile, Expression rowFilter, boolean caseSensitive) {
```
aokolnychyi, line 120: Instead of giving up if the manifest spec ID does not match the current table spec, I think we should use the manifest spec to project the row filter and push down whatever we can. Remember that the partition type will be a union of all partition columns ever used in the table.

Another problem with the current implementation is that we call `transformSpec` and create a manifest evaluator for every manifest file. That's expensive. We usually use `LoadingCache` to reuse manifest evaluators.

Also, why not use `ManifestEvalutor.forRowFilter` instead of doing the projection ourselves?

Will something like this work if we modify `planFiles`?

```
Map<Integer, PartitionSpec> specsById = ops.current().specsById();

LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
  PartitionSpec spec = specsById.get(specId);
  PartitionSpec transformedSpec = transformSpec(fileSchema, spec, PARTITION_FIELD_PREFIX);
  return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
});

CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(
    manifests(),
    manifest -> evalCache.get(manifest.partitionSpecId()).eval(manifest));
```

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
aokolnychyi: I'd probably prefer not to reuse `populateTableNewSpec` and duplicate some parts of the code but cover v1 and v2 tables to be safe. Using extra helper methods in tests reduces code duplication but makes it hard to review such tests as I had to jump to those methods to understand the current table state.

I'd also add tests that refer to the old partition spec. The `partition` column is a union of all partition types. We should push down conditions on `data_bucket` to manifests that belong to the old spec.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Hi @aokolnychyi yea ideally we would try harder to filter even on older specs, I was just looking to fix the problem earlier.

So the problem I run into here, is if a field is renamed (ie, TestMetadataTablesWithPartitionEvolution.testFilesMetadataTable).  In this case, transformSpec() method tries to bind the expression with the old spec on the fileSchema.  But it can't find the old field anymore.  As you know, the fileSchema is derived from Partitioning.partitionType, which has union, but probably not the old names.  

Stack trace:
```
Cannot find source column: partition.category_bucket_8
java.lang.IllegalArgumentException: Cannot find source column: partition.category_bucket_8
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
	at org.apache.iceberg.PartitionSpec$Builder.findSourceColumn(PartitionSpec.java:397)
	at org.apache.iceberg.PartitionSpec$Builder.identity(PartitionSpec.java:402)
	at org.apache.iceberg.BaseMetadataTable.lambda$transformSpec$0(BaseMetadataTable.java:66)
	at org.apache.iceberg.relocated.com.google.common.collect.ImmutableList.forEach(ImmutableList.java:422)
	at org.apache.iceberg.BaseMetadataTable.transformSpec(BaseMetadataTable.java:66)
	at org.apache.iceberg.BaseFilesTable$BaseFilesTableScan.lambda$planFiles$0(BaseFilesTable.java:103)
	at com.github.benmanes.caffeine.cache.LocalLoadingCache.lambda$newMappingFunction$2(LocalLoadingCache.java:141)
	at com.github.benmanes.caffeine.cache.UnboundedLocalCache.lambda$computeIfAbsent$2(UnboundedLocalCache.java:238)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
	at com.github.benmanes.caffeine.cache.UnboundedLocalCache.computeIfAbsent(UnboundedLocalCache.java:234)
	at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:108)
	at com.github.benmanes.caffeine.cache.LocalLoadingCache.get(LocalLoadingCache.java:54)
	at org.apache.iceberg.BaseFilesTable$BaseFilesTableScan.lambda$planFiles$1(BaseFilesTable.java:109)
	at org.apache.iceberg.io.CloseableIterable$3.shouldKeep(CloseableIterable.java:82)
	at org.apache.iceberg.io.FilterIterator.advance(FilterIterator.java:67)
	at org.apache.iceberg.io.FilterIterator.hasNext(FilterIterator.java:50)
	at org.apache.iceberg.io.CloseableIterable$4$1.hasNext(CloseableIterable.java:108)
	at org.apache.iceberg.relocated.com.google.common.collect.Iterators.addAll(Iterators.java:366)
	at org.apache.iceberg.relocated.com.google.common.collect.Lists.newArrayList(Lists.java:147)
	at org.apache.iceberg.relocated.com.google.common.collect.Lists.newArrayList(Lists.java:133)
	at org.apache.iceberg.spark.source.SparkBatchQueryScan.files(SparkBatchQueryScan.java:115)
	at org.apache.iceberg.spark.source.SparkBatchQueryScan.tasks(SparkBatchQueryScan.java:128)
	at org.apache.iceberg.spark.source.SparkScan.toBatch(SparkScan.java:108)
	at org.apache.spark.sql.execution.datasources.v2.BatchScanExec.batch$lzycompute(BatchScanExec.scala:40)
	at org.apache.spark.sql.execution.datasources.v2.BatchScanExec.batch(BatchScanExec.scala:40)
	at org.apache.spark.sql.execution.datasources.v2.BatchScanExec.partitions$lzycompute(BatchScanExec.scala:52)
	at org.apache.spark.sql.execution.datasources.v2.BatchScanExec.partitions(BatchScanExec.scala:52)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceV2ScanExecBase.supportsColumnar(DataSourceV2ScanExecBase.scala:93)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceV2ScanExecBase.supportsColumnar$(DataSourceV2ScanExecBase.scala:92)
	at org.apache.spark.sql.execution.datasources.v2.BatchScanExec.supportsColumnar(BatchScanExec.scala:35)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceV2Strategy.apply(DataSourceV2Strategy.scala:123)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$1(QueryPlanner.scala:63)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:491)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.plan(QueryPlanner.scala:93)
	at org.apache.spark.sql.execution.SparkStrategies.plan(SparkStrategies.scala:68)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$3(QueryPlanner.scala:78)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:196)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:194)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:199)
	at scala.collection.TraversableOnce.foldLeft$(TraversableOnce.scala:192)
	at scala.collection.AbstractIterator.foldLeft(Iterator.scala:1431)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$2(QueryPlanner.scala:75)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.plan(QueryPlanner.scala:93)
	at org.apache.spark.sql.execution.SparkStrategies.plan(SparkStrategies.scala:68)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$3(QueryPlanner.scala:78)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:196)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:194)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:199)
	at scala.collection.TraversableOnce.foldLeft$(TraversableOnce.scala:192)
	at scala.collection.AbstractIterator.foldLeft(Iterator.scala:1431)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$2(QueryPlanner.scala:75)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.plan(QueryPlanner.scala:93)
	at org.apache.spark.sql.execution.SparkStrategies.plan(SparkStrategies.scala:68)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$3(QueryPlanner.scala:78)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:196)
	at scala.collection.TraversableOnce$folder$1.apply(TraversableOnce.scala:194)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:199)
	at scala.collection.TraversableOnce.foldLeft$(TraversableOnce.scala:192)
	at scala.collection.AbstractIterator.foldLeft(Iterator.scala:1431)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.$anonfun$plan$2(QueryPlanner.scala:75)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at org.apache.spark.sql.catalyst.planning.QueryPlanner.plan(QueryPlanner.scala:93)
	at org.apache.spark.sql.execution.SparkStrategies.plan(SparkStrategies.scala:68)
	at org.apache.spark.sql.execution.QueryExecution$.createSparkPlan(QueryExecution.scala:468)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$sparkPlan$2(QueryExecution.scala:157)
	at org.apache.spark.sql.catalyst.QueryPlanningTracker.measurePhase(QueryPlanningTracker.scala:111)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$executePhase$1(QueryExecution.scala:196)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
	at org.apache.spark.sql.execution.QueryExecution.executePhase(QueryExecution.scala:196)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$sparkPlan$1(QueryExecution.scala:157)
	at org.apache.spark.sql.execution.QueryExecution.withCteMap(QueryExecution.scala:73)
	at org.apache.spark.sql.execution.QueryExecution.sparkPlan$lzycompute(QueryExecution.scala:150)
	at org.apache.spark.sql.execution.QueryExecution.sparkPlan(QueryExecution.scala:150)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$executedPlan$2(QueryExecution.scala:170)
	at org.apache.spark.sql.catalyst.QueryPlanningTracker.measurePhase(QueryPlanningTracker.scala:111)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$executePhase$1(QueryExecution.scala:196)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
	at org.apache.spark.sql.execution.QueryExecution.executePhase(QueryExecution.scala:196)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$executedPlan$1(QueryExecution.scala:170)
	at org.apache.spark.sql.execution.QueryExecution.withCteMap(QueryExecution.scala:73)
	at org.apache.spark.sql.execution.QueryExecution.executedPlan$lzycompute(QueryExecution.scala:163)
	at org.apache.spark.sql.execution.QueryExecution.executedPlan(QueryExecution.scala:163)
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$5(SQLExecution.scala:101)
	at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:163)
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$1(SQLExecution.scala:90)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
	at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:64)
	at org.apache.spark.sql.Dataset.withAction(Dataset.scala:3704)
	at org.apache.spark.sql.Dataset.collectAsList(Dataset.scala:2982)
	at org.apache.iceberg.spark.source.TestMetadataTablesWithPartitionEvolution.assertPartitions(TestMetadataTablesWithPartitionEvolution.java:328)
	at org.apache.iceberg.spark.source.TestMetadataTablesWithPartitionEvolution.testFilesMetadataTable(TestMetadataTablesWithPartitionEvolution.java:185)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:110)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:38)
	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at com.sun.proxy.$Proxy2.processTestClass(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$2.run(TestWorker.java:176)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:129)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:100)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:60)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:133)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:71)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
```

aokolnychyi: It looks like a bug in `transformSpec` that assumes old partition fields are never renamed. If we switch to using column IDs, will it help?

```
static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec) {
  spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), pf.name(), "identity"));
  return identitySpecBuilder.build();
}
```

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -92,7 +94,26 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
+        PartitionSpec transformedSpec = transformSpec(fileSchema, spec, PARTITION_FIELD_PREFIX);
+        return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
+      });
+
+      CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(
+          manifests(),
+          manifest -> {
+            PartitionSpec spec = specsById.get(manifest.partitionSpecId());
+
+            if (spec.fields().stream().anyMatch(f -> f.transform().equals(Transforms.alwaysNull()))) {
```
szehon-ho, line 110: This case is a bit ugly.  For V1 tables in the case of removing partition fields, it replaces the transform with a "void" transform.  What happens is that when we use this spec to evaluate a manifest file, it filters them all out as it searches for null value for this partition key, which seems incorrect.

Example, if we remove a field "data_bucket" from partition spec, then this code would check whether manifest files have null "data_bucket" values.

So hence skipping this check for now.

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -63,7 +63,7 @@ protected BaseMetadataTable(TableOperations ops, Table table, String name) {
    */
   static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec, String partitionPrefix) {
     PartitionSpec.Builder identitySpecBuilder = PartitionSpec.builderFor(metadataTableSchema).checkConflicts(false);
-    spec.fields().forEach(pf -> identitySpecBuilder.identity(partitionPrefix + pf.name(), pf.name()));
+    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), partitionPrefix + pf.name(), "identity"));
```
aokolnychyi, line 66: Shouldn't this be just `pf.name()`?

szehon-ho: Whoops yes, you are right.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Thought: haven't looked but maybe we could potentially push this logic down to ManifestEvaluator, return "ROWS_MIGHT_MATCH" match for any partition comparison involving  a partition field that has become VoidTransformation (removed in V1).  (I am thinking this case of querying removed partition field should only affect metadata tables primarily).

aokolnychyi: Well, it looked wrong initially to filter out such manifests but I am no longer sure.
If we have a filter `partition.data_bucket = 1` and `data_bucket` is null for evolved tables, maybe it is correct to skip such manifests as technically the value for that partition value is null?

aokolnychyi: Suppose I partitioned the table by `bucket(8, id)` and then switched to partitioning by `id` and have two files like this.

```
partition(data_bucket) = {1, null} - file written with bucket(8, id)
partition(id) = {null, 100} - file written with identity(id)
```

If I query `partition.data_bucket = 1`, it should be safe to skip the second file and manifest?

szehon-ho: Yea you may be right.  The difference is we are not skipping the manifest for V2 tables but are for V1, though maybe in the end they get filtered out anyway, I can add an end to end test for this.

aokolnychyi: An end to end test would be great. I think it will produce the same output for V1 and V2 tables.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -92,7 +93,16 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
-      CloseableIterable<ManifestFile> filtered = filterManifests(manifests(), rowFilter, caseSensitive);
+      Map<Integer, PartitionSpec> specsById = table().specs();
+
+      LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
+        PartitionSpec spec = specsById.get(specId);
+        PartitionSpec transformedSpec = transformSpec(fileSchema, spec);
+        return ManifestEvaluator.forRowFilter(rowFilter, transformedSpec, caseSensitive);
+      });
+
+      CloseableIterable<ManifestFile> filtered = CloseableIterable.filter(manifests(),
```
aokolnychyi, line 104: nit: shall we call it `filteredManifests`?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -74,14 +72,29 @@
                 "default-namespace", "default"
             ),
             ORC,
-            formatVersion()
+            1
+        },
+        { "testhive", SparkCatalog.class.getName(),
```
szehon-ho, line 77: Just adding two more runs , should not take too long, and I think will make this test much more deterministic (right now randomly picks a version)

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -92,7 +93,16 @@ public TableScan appendsAfter(long fromSnapshotId) {
     protected CloseableIterable<FileScanTask> planFiles(TableOperations ops, Snapshot snapshot, Expression rowFilter,
                                                         boolean ignoreResiduals, boolean caseSensitive,
                                                         boolean colStats) {
-      CloseableIterable<ManifestFile> filtered = filterManifests(manifests(), rowFilter, caseSensitive);
+      Map<Integer, PartitionSpec> specsById = table().specs();
```
aokolnychyi, line 96: In a few other places, we copy specs into a new map via `Maps.newHashMap` to make sure Java collections are used. This is done to avoid any potential Kryo issues as `ScanTasks` are serializable. Maybe, we should do the same here.

At the end of the method, we call `ops.io(), ops.current().specsById()` to construct a task. In theory, we could pass `Table` object to `ManifestReadTask` and then access `FileIO` via `table.io()` and copy specs into a Java map in the constructor. Up to you. At least, no need to call `ops.current().specsById()` as we already have specs.

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -63,6 +64,7 @@ public TestMetadataTableFilters(MetadataTableType type, int formatVersion) {
   @Override
   public void setupTable() throws Exception {
     super.setupTable();
+    table.updateProperties().set(TableProperties.MANIFEST_MERGE_ENABLED, "false").commit();
```
aokolnychyi, line 67: nit: can we split this into multiple lines like the invocation below?

```
table.updateProperties()
    .set(..)
    .commit();
```

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
...
+        .withPath("/path/to/data-11-deletes.parquet")
+        .withFileSizeInBytes(10)
+        .withPartition(data11Key)
+        .withRecordCount(1)
+        .build();
+
+    table.newFastAppend().appendFile(data10).commit();
+    table.newFastAppend().appendFile(data11).commit();
+
+    if (formatVersion == 2) {
```
aokolnychyi, line 302: nit: do we need this branch? looks like the test is for v1.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Removed block and added test to show it doesnt matter

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
     validateCombinedScanTasks(tasks, 0);
   }
 
+  @Test
+  public void testPartitionSpecEvolutionRemovalV1() {
+    Assume.assumeTrue(formatVersion == 1);
+
+    // Change spec and add two data and delete files each
+    table.updateSpec().removeField(Expressions.bucket("data", 16))
```
aokolnychyi, line 262: nit: same here, I think it would be more readable to have each call on a separate line.

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -52,18 +51,17 @@ protected BaseMetadataTable(TableOperations ops, Table table, String name) {
...
    * @param metadataTableSchema schema of the metadata table
    * @param spec spec on which the metadata table schema is based
-   * @param partitionPrefix prefix to remove from each field in the partition spec
    * @return a spec used to rewrite the metadata table filters to partition filters using an inclusive projection
    */
-  static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec, String partitionPrefix) {
+  static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec) {
     PartitionSpec.Builder identitySpecBuilder = PartitionSpec.builderFor(metadataTableSchema).checkConflicts(false);
-    spec.fields().forEach(pf -> identitySpecBuilder.identity(partitionPrefix + pf.name(), pf.name()));
+    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), pf.name(), "identity"));
```
szehon-ho, line 64: This uses the field id to do the transform from metadata table schema to table schema to catch partition field rename cases (old way was to manual remove 'partition.' from the field name).

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
...
+    PartitionKey data11Key = new PartitionKey(newSpec, table.schema());
+    data10Key.set(1, 11);
+    DataFile data11 = DataFiles.builder(newSpec)
+        .withPath("/path/to/data-11.parquet")
+        .withRecordCount(10)
+        .withFileSizeInBytes(10)
+        .withPartition(data11Key)
+        .build();
+
+    DeleteFile delete10 = FileMetadata.deleteFileBuilder(newSpec)
```
aokolnychyi, line 284: I don't think you need delete files for v1 tests.

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
...
+  public void testPartitionSpecEvolutionRemovalV1() {
+    Assume.assumeTrue(formatVersion == 1);
+
+    // Change spec and add two data and delete files each
+    table.updateSpec().removeField(Expressions.bucket("data", 16))
+        .addField("id").commit();
+    PartitionSpec newSpec = table.spec();
+
+    // Add two data files and two delete files with new spec
+    PartitionKey data10Key = new PartitionKey(newSpec, table.schema());
```
aokolnychyi, line 267: An alternative way to define partitioning is to use `withPartitionPath` and specify a string.
I don't have a preference, just FYI.

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
     validateCombinedScanTasks(tasks, 0);
   }
 
+  @Test
+  public void testPartitionSpecEvolutionRemovalV1() {
```
aokolnychyi, line 258: nit: some comments in this test refer to delete files which cannot be part of v1 tables

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
...
+    // 1 original data/delete files written by old spec (V1 filters out new specs which don't have this value)
+    Assert.assertEquals(expectedScanTaskCount(1), Iterables.size(tasks));
+  }
+
+  @Test
+  public void testPartitionSpecEvolutionRemovalV2() {
+    Assume.assumeTrue(formatVersion == 2);
+
+    // Change spec and add two data and delete files each
+    table.updateSpec().removeField(Expressions.bucket("data", 16))
```
aokolnychyi, line 340: nit: same as above

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
```diff
@@ -252,6 +254,310 @@ public void testPlanTasks() {
...
+        Expressions.greaterThan("record_count", 0));
+    scan = metadataTable.newScan().filter(filter);
+    tasks = scan.planFiles();
+
+    // 1 original data/delete files written by old spec, plus both of new data file/delete file written by new spec
+    Assert.assertEquals(expectedScanTaskCount(3), Iterables.size(tasks));
+  }
+
+  @Test
+  public void testPartitionSpecEvolutionAdditiveV1() {
```
aokolnychyi, line 410: some of the above comments to earlier tests apply to this one too

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -92,17 +105,11 @@
                 "cache-enabled", "false" // Spark will delete tables using v1, leaving the cache out of sync
             ),
             AVRO,
-            formatVersion()
+            1
```
aokolnychyi, line 108: Shall we test v2 for Avro as well?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -189,6 +196,109 @@ public void testFilesMetadataTable() throws ParseException {
...
+
+    table.updateSpec()
+        .addField("category")
+        .commit();
+    sql("REFRESH TABLE %s", tableName);
+    sql("INSERT INTO TABLE %s VALUES (1, 'c1', 'd1')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'c2', 'd2')", tableName);
+
+    Dataset<Row> result = loadMetadataTable(ALL_ENTRIES);
+    result.collectAsList().forEach(f -> System.out.println(f));
```
aokolnychyi, line 240: Do we need this?

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
szehon-ho: Yea I couldn't get it to work due to the void transform being there, didn't try that hard though.

core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java
szehon-ho: Good catch

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
szehon-ho: Yea sorry left over debug statement

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
szehon-ho: Yea was trying to not increase the test time (seems it was original concern), but added it.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
szehon-ho: Hm, not sure I completely see, this specsById is just used to lookup a cache right, before the ManifestReadTask is returned?  

So I didn't change this line but changed the ManifestReadTask constructor to take table and there do Maps.newHashMap, let me know if thats what you wanted

szehon-ho:
@rdblue would you also want to take a look at this too if you have some time (as its a bit tricky code)?  Thanks

szehon-ho:
Adding an observation, the case where wrong results are filtered out is limited to V2 tables, where partition fields are removed (ie, in the initial description).

In tests for any partition spec modification in V1 (in V1, fields are never removed), or V2 tables where partition fields are modified or added but not removed, we get a NPE in the ManifestEvaluator if we try to query files written by different partition spec than current one.

Still worth to fix, just making a note for the impact.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
aokolnychyi: This looks good to me.

aokolnychyi (approved):
LGTM

aokolnychyi:
I merged this one to unblock one local change I have. @RussellSpitzer @rdblue, if you could still review, it would be great.

szehon-ho:
Thanks all, ill also look at merging this to the 0.13.x branch for the 0.13.2 release (as the metadata tables hierarchy is different in master).  @rdblue Let me know if this is not the right branch.

PR #4522, opened 2022-04-07
Core: Rewrite v2 table files.

openinx:


chenjunjiedada:
@openinx For the spark side, we could set the binpack option `delete-file-threshold` to 1 to rewrite the deletes. Is this for merge delete and data file but not for binpacking?

rdblue:
@openinx, can you add a bit more description about what this is doing? What does this cover and how are rewrites performed?

ajantha-bhat:
a. Can we separate Flink and spark changes ?
b. Spark already supports merging the v2 delete files to data files in `rewriteDataFiles` action. If we are supporting compacting the v2 delete files to a big delete file in this PR, I think it can be a separate action.

core/src/main/java/org/apache/iceberg/actions/BaseRewriteDataFilesAction.java
```diff
@@ -270,43 +275,69 @@ private Map<StructLikeWrapper, Collection<FileScanTask>> groupTasksByPartition(
...
-        .validateFromSnapshot(startingSnapshotId)
-        .rewriteFiles(Sets.newHashSet(deletedDataFiles), Sets.newHashSet(addedDataFiles));
-    commit(rewriteFiles);
+  private boolean doPartitionNeedRewrite(Collection<FileScanTask> partitionTasks) {
+    int files = 0;
+    for (FileScanTask scanTask : partitionTasks) {
+      files += 1; // One for data file.
+      files += scanTask.deletes().size();
+    }
+    return files > 1;
```
wuwenchi, line 317: If I only want to split a large file, is there a problem here?

wuwenchi, line 317: If I only want to split a large file, is there a problem here?

PR #4525, opened 2022-04-08
API: Add default value core api and schema serialization

rzhang10:
This is the first PR to implement the spec pr #4301 

It augments the `NestedField` API with default values and adds logic to `SchemaParser` for serializing default values with schema, and it also implements the logic to deserialize it again to the in-memory representation in `NestedField`.

The default value is represented in memory as vanilla java `Object` and is validated/parsed/converted from the user-facing JSON representation via utilities in `DefaultValueParser`. The JSON representation is also what gets serialized to disk along with the schema, the representation aligns with Appendix C in spec #4301.

This PR will be the basis of the later PRs to support default value r/w semantics in each data type X compute engine.

Please take a look @rdblue , thank you!

### DIFF
--- a/api/src/main/java/org/apache/iceberg/expressions/Literal.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Literal.java
@@ -71,6 +71,18 @@ public interface Literal<T> extends Serializable {
     return new Literals.DecimalLiteral(value);
   }
 
+  static Literal<Integer> ofDateLiteral(int value) {
+    return new Literals.DateLiteral(value);
+  }
+
+  static Literal<Long> ofTimeLiteral(long value) {
+    return new Literals.TimeLiteral(value);
+  }
+
+  static Literal<Long> ofTimestampLiteral(long value) {
+    return new Literals.TimestampLiteral(value);
+  }
+
   /**
    * Returns the value wrapped by this literal.
    */
--- a/api/src/main/java/org/apache/iceberg/expressions/Literals.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Literals.java
@@ -149,7 +149,6 @@ class Literals {
     public int hashCode() {
       return Objects.hashCode(value);
     }
-
   }
 
   private abstract static class ComparableLiteral<C extends Comparable<C>> extends BaseLiteral<C> {
@@ -399,6 +398,8 @@ class Literals {
     public <T> Literal<T> to(Type type) {
       if (type.typeId() == Type.TypeID.DATE) {
         return (Literal<T>) this;
+      } else if (type.typeId() == Type.TypeID.STRING) {
+        return (Literal<T>) new StringLiteral(LocalDate.ofEpochDay(value()).format(DateTimeFormatter.ISO_LOCAL_DATE));
       }
       return null;
     }
@@ -419,6 +420,9 @@ class Literals {
     public <T> Literal<T> to(Type type) {
       if (type.typeId() == Type.TypeID.TIME) {
         return (Literal<T>) this;
+      } else if (type.typeId() == Type.TypeID.STRING) {
+        return (Literal<T>) new StringLiteral(LocalTime.ofNanoOfDay(value() * 1000)
+            .format(DateTimeFormatter.ISO_LOCAL_TIME));
       }
       return null;
     }
@@ -443,6 +447,10 @@ class Literals {
         case DATE:
           return (Literal<T>) new DateLiteral((int) ChronoUnit.DAYS.between(
               EPOCH_DAY, EPOCH.plus(value(), ChronoUnit.MICROS).toLocalDate()));
+        case STRING:
+          // Always return the literal without timezone.
+          return (Literal<T>) new StringLiteral(LocalDateTime.ofEpochSecond(value() / 1000000,
+              (int) (value() % 1000000) * 1000, ZoneOffset.UTC).format(DateTimeFormatter.ISO_LOCAL_DATE_TIME));
         default:
       }
       return null;
--- a/api/src/main/java/org/apache/iceberg/types/Types.java
+++ b/api/src/main/java/org/apache/iceberg/types/Types.java
@@ -19,7 +19,6 @@
 
 package org.apache.iceberg.types;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.List;
@@ -28,14 +27,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
-import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
-import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.types.Type.NestedType;
 import org.apache.iceberg.types.Type.PrimitiveType;
 
@@ -428,7 +423,7 @@ public class Types {
     }
 
     public static NestedField optional(int id, String name, Type type, String doc,
-                                       JsonNode initialDefault, JsonNode writeDefault) {
+                                       Object initialDefault, Object writeDefault) {
       return new NestedField(true, id, name, type, doc, initialDefault, writeDefault);
     }
 
@@ -441,7 +436,7 @@ public class Types {
     }
 
     public static NestedField required(int id, String name, Type type, String doc,
-                                       JsonNode initialDefault, JsonNode writeDefault) {
+                                       Object initialDefault, Object writeDefault) {
       return new NestedField(false, id, name, type, doc, initialDefault, writeDefault);
     }
 
@@ -454,7 +449,7 @@ public class Types {
     }
 
     public static NestedField of(int id, boolean isOptional, String name, Type type, String doc,
-                                 JsonNode initialDefault, JsonNode writeDefault) {
+                                 Object initialDefault, Object writeDefault) {
       return new NestedField(isOptional, id, name, type, doc, initialDefault, writeDefault);
     }
 
@@ -463,12 +458,12 @@ public class Types {
     private final String name;
     private final Type type;
     private final String doc;
-    private final JsonNode initialDefault;
-    private final JsonNode writeDefault;
+    private final Object initialDefault;
+    private final Object writeDefault;
 
 
     private NestedField(boolean isOptional, int id, String name, Type type, String doc,
-                        JsonNode initialDefault, JsonNode writeDefault) {
+        Object initialDefault, Object writeDefault) {
       Preconditions.checkNotNull(name, "Name cannot be null");
       Preconditions.checkNotNull(type, "Type cannot be null");
       this.isOptional = isOptional;
@@ -476,8 +471,8 @@ public class Types {
       this.name = name;
       this.type = type;
       this.doc = doc;
-      this.initialDefault = validateDefault(name, type, initialDefault);
-      this.writeDefault = validateDefault(name, type, writeDefault);
+      this.initialDefault = initialDefault;
+      this.writeDefault = writeDefault;
     }
 
     public boolean isOptional() {
@@ -502,9 +497,8 @@ public class Types {
       return new NestedField(false, id, name, type, doc, initialDefault, writeDefault);
     }
 
-    public NestedField updateWriteDefault(JsonNode newWriteDefault) {
-      JsonNode validatedWriteDefault = validateDefault(name, type, newWriteDefault);
-      return new NestedField(isOptional, id, name, type, doc, initialDefault, validatedWriteDefault);
+    public NestedField updateWriteDefault(Object newWriteDefault) {
+      return new NestedField(isOptional, id, name, type, doc, initialDefault, newWriteDefault);
     }
 
     public int fieldId() {
@@ -523,11 +517,11 @@ public class Types {
       return doc;
     }
 
-    public JsonNode initialDefaultValue() {
+    public Object initialDefaultValue() {
       return initialDefault;
     }
 
-    public JsonNode writeDefaultValue() {
+    public Object writeDefaultValue() {
       return writeDefault;
     }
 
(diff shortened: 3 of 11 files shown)


build.gradle
```diff
@@ -193,6 +193,7 @@ project(':iceberg-bundled-guava') {
 project(':iceberg-api') {
   dependencies {
     implementation project(path: ':iceberg-bundled-guava', configuration: 'shadow')
+    implementation "com.fasterxml.jackson.core:jackson-databind"
```
rdblue, line 196: Why is an additional Jackson library needed? I think we should be able to add default value serialization just like the other parsers.

core/src/main/java/org/apache/iceberg/SchemaParser.java
```diff
@@ -92,6 +94,14 @@ private static void toJson(Types.StructType struct, Integer schemaId, Set<Intege
       if (field.doc() != null) {
         generator.writeStringField(DOC, field.doc());
       }
+      if (field.initialDefaultValue() != null) {
+        generator.writeFieldName(INITIAL_DEFAULT);
+        generator.writeRawValue(field.initialDefaultValue().toString());
+      }
+      if (field.writeDefaultValue() != null) {
+        generator.writeFieldName(WRITE_DEFAULT);
+        generator.writeRawValue(field.writeDefaultValue().toString());
```
rdblue, line 103: Instead of using `toString` that produces JSON, I think that this should expose a `ValueParser` that converts values from the internal Iceberg representation into JSON values.

api/src/test/java/org/apache/iceberg/types/TestTypeDefaultValuesValidation.java
```diff
@@ -0,0 +1,89 @@
...
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][] {
+        {Types.BooleanType.get(), parseJsonStringToJsonNode("true")},
+        {Types.IntegerType.get(), parseJsonStringToJsonNode("1")},
+        {Types.LongType.get(), parseJsonStringToJsonNode("9999999")},
+        {Types.FloatType.get(), parseJsonStringToJsonNode("1.23")},
+        {Types.DoubleType.get(), parseJsonStringToJsonNode("123.456")},
+        {Types.DateType.get(), parseJsonStringToJsonNode("3650")},
+        {Types.TimeType.get(), parseJsonStringToJsonNode("36000000000")},
+        {Types.TimestampType.withoutZone(), parseJsonStringToJsonNode("1649374911000000")},
```
rdblue, line 56: Looking at this again, I think that date/time fields should be converted to (strict) ISO-8601 representations. That way they are readable in JSON.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -524,6 +563,91 @@ public boolean equals(Object o) {
...
+
+    private static JsonNode validateDefault(String name, Type type, JsonNode defaultValue) {
+      if (defaultValue != null && !isValidDefault(type, defaultValue)) {
+        throw new ValidationException("Invalid default value for field %s: %s not a %s", name, defaultValue, type);
+      }
+      return defaultValue;
+    }
+
+    @SuppressWarnings("checkstyle:CyclomaticComplexity")
+    private static boolean isValidDefault(Type type, JsonNode defaultValue) {
```
rdblue, line 575: Iceberg uses visitors to separate the logic for schema traversal from the logic for individual decisions. Can you please use a visitor?

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -492,6 +523,14 @@ public String doc() {
       return doc;
     }
 
+    public JsonNode initialDefaultValue() {
```
rdblue, line 526: This API should not use Jackson classes. Instead, it should use the standard Iceberg in-memory representation:
* boolean - Java primitive boolean
* int - Java primitive int
* long - Java primitive long
* float - Java primitive float
* double - Java primitive double
* date - Java primitive int
* time - Java primitive long
* timestamp/timestamptz - Java primitive long
* string - Java CharSequence
* uuid - Java UUID
* binary - Java ByteBuffer
* fixed(L) - Java ByteBuffer
* decimal(P, S) - Java BigDecimal
* list - Java Collection
* struct - StructLike
* map - Java Map

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -524,6 +563,91 @@ public boolean equals(Object o) {
     public int hashCode() {
       return Objects.hash(NestedField.class, id, isOptional, name, type);
     }
+
+    private static JsonNode validateDefault(String name, Type type, JsonNode defaultValue) {
```
rdblue, line 567: I think this logic should be moved to a `ValueParser` with `toJson` and `fromJson` methods, like the others. We don't want to leak any Jackson classes in the API.

rzhang10:
@rdblue 

I see your major concern in the reviews are regarding using Jackson in Iceberg-api module.

Let me put my thoughts here in one place to answer all the questions above, the core problem is whether we should represent the default value in memory in `JsonNode` or Java primitives.

The reason I chose Jackson is that:
- Jackson has more robust json validation utils, which come in handy here.
- Jackson `JsonNoe` is also `java.io.Serializable`.
- Jackson is also used in the Iceberg-core module as a dependency. I believe whenever iceberg is used as a library, iceberg-api and iceberg-core will both be dependent upon. So I wouldn't believe introducing Jackson in the API module will (additionally) pollute downstream dependencies.

If we take the java primitive approach, I think my current code could be adapted this way:

Maintain the current json validation logic, then add a `ValueParser` with `toJson` and `fromJson` to convert between json and their respective java primitive representations. The structure of the current code will pretty much stay the same.

Let me know your thoughts and I'm open to further discussions too.

api/src/main/java/org/apache/iceberg/types/Types.java
rzhang10: I feel using a `SchemaVisitor` here would be an overkill, since this logic is only validating a `NestField`, it's technically only part of the schema, and even if it's a complex type, we only recursively validate downward, but not upward. Also I feel readability-wise, this recursive implementation is clean enough? it clearly shows the logic is to validate each type of default value case-by-case.

To me, I feel `SchemaVisitor` is mostly used to do holistic schema manipulations/transformations.

rdblue:
@rzhang10, it's okay to add the library to the API module if that's what is needed, although I'd like to see if we actually need it there or if we can convert values to/from JSON in core.

For the in-memory representation, we do need to use the common representation and not JsonNode. We also cannot leak Jackson classes in the API module.

api/src/main/java/org/apache/iceberg/types/Types.java
rzhang10: > * list - Java Collection

I think using Java `List` is more correct, as List cares about the order.



> * struct - StructLike

I see this comment:

```
/**
 * Interface for accessing data by position in a schema.
 * <p>
 * This interface supports accessing data in top-level fields, not in nested fields.
 */
```
in that interface, also I see the API exposes is getting fields by position, which doesn't suit our casing of accessing fields by field id.

I think using a native Java map would be better.

api/src/main/java/org/apache/iceberg/expressions/Literal.java
```diff
@@ -71,6 +71,18 @@ static Literal<BigDecimal> of(BigDecimal value) {
     return new Literals.DecimalLiteral(value);
   }
 
+  static Literal<Integer> ofDateLiteral(int value) {
```
rdblue, line 74: I don't think you should need any new literal constructors. You can always convert using `to` and the type.

api/src/main/java/org/apache/iceberg/expressions/Literals.java
```diff
@@ -419,6 +420,9 @@ static class TimeLiteral extends ComparableLiteral<Long> {
     public <T> Literal<T> to(Type type) {
       if (type.typeId() == Type.TypeID.TIME) {
         return (Literal<T>) this;
+      } else if (type.typeId() == Type.TypeID.STRING) {
+        return (Literal<T>) new StringLiteral(LocalTime.ofNanoOfDay(value() * 1000)
+            .format(DateTimeFormatter.ISO_LOCAL_TIME));
```
rdblue, line 425: I don't think this conversion should be done using literals. Instead, use the methods in `DateTimeUtil` and add any new methods there. This shouldn't require changes to the public API in expressions.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -415,43 +415,64 @@ public int hashCode() {
...
     private final String name;
     private final Type type;
     private final String doc;
+    private final Object initialDefault;
+    private final Object writeDefault;
+
 
-    private NestedField(boolean isOptional, int id, String name, Type type, String doc) {
+    private NestedField(boolean isOptional, int id, String name, Type type, String doc,
+        Object initialDefault, Object writeDefault) {
```
rdblue, line 466: Style: In Iceberg always start all argument lines at the same point, either all indented after the method declaration (aligned with `boolean isOptional, ...`) or all starting at the same continuation indent level. This mixes the two.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -415,43 +415,64 @@ public int hashCode() {
...
     }
 
     private final boolean isOptional;
     private final int id;
     private final String name;
     private final Type type;
     private final String doc;
+    private final Object initialDefault;
+    private final Object writeDefault;
+
```
rdblue, line 463: Style: avoid adding unnecessary newlines. One newline to separate blocks is sufficient.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -473,7 +494,11 @@ public NestedField asRequired() {
       if (!isOptional) {
         return this;
       }
-      return new NestedField(false, id, name, type, doc);
+      return new NestedField(false, id, name, type, doc, initialDefault, writeDefault);
+    }
+
+    public NestedField updateWriteDefault(Object newWriteDefault) {
```
rdblue, line 500: I don't think there are other methods like this. Is it worth adding a new pattern for just this? Why not add `updateDoc`? I think I'd rather copy where needed.

rdblue, line 500: There's precedent for returning a slightly modified version of the field with the `asRequired` and `asOptional` methods, but this method name doesn't make it clear to me that the original field isn't being modified. I think `withWriteDefault` is slightly better, although not great.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -492,6 +517,14 @@ public String doc() {
       return doc;
     }
 
+    public Object initialDefaultValue() {
```
rdblue, line 520: Do we need `Value` here? It seems redundant. Maybe just `initialDefault` and `writeDefault`.

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
```diff
@@ -0,0 +1,253 @@
...
+import org.apache.iceberg.exceptions.ValidationException;
+import org.apache.iceberg.expressions.Literal;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.relocated.com.google.common.io.BaseEncoding;
+import org.apache.iceberg.types.Type;
+import org.apache.iceberg.types.Types;
+
+public class DefaultValueParser {
```
rdblue, line 48: I think this class could be in a separate PR to start with. The representation and tests are fairly well-defined. Can you separate it out?

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  public static Object parseDefaultFromJson(Type type, JsonNode jsonNode) {
+    if (jsonNode == null) {
+      return null;
+    }
+
+    switch (type.typeId()) {
+      case BOOLEAN:
+        return jsonNode.booleanValue();
```
rdblue, line 73: I think this should use the same pattern and checks that are used in other JSON parsers. Those use `JsonUtil` and validate the type. For example:

```java
  public static int getInt(String property, JsonNode node) {
    Preconditions.checkArgument(node.has(property), "Cannot parse missing int %s", property);
    JsonNode pNode = node.get(property);
    Preconditions.checkArgument(pNode != null && !pNode.isNull() && pNode.isNumber(),
        "Cannot parse %s to an integer value: %s", property, pNode);
    return pNode.asInt();
  }
```

That checks that the node actually is an integer. We want to be similarly strict here.

core/src/main/java/org/apache/iceberg/util/JsonUtil.java
```diff
@@ -115,6 +115,17 @@ public static String getStringOrNull(String property, JsonNode node) {
     return pNode.asText();
   }
 
+  public static JsonNode getChildNodeOrNull(String property, JsonNode node) {
+    if (!node.has(property)) {
+      return null;
+    }
+    JsonNode pNode = node.get(property);
```
rdblue, line 122: Style: control flow blocks should be separated from the next statement by an empty newline.

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  private static final JsonFactory FACTORY = new JsonFactory();
+  private static final ObjectMapper MAPPER;
+
+  static {
+    MAPPER = new ObjectMapper(FACTORY);
+    SimpleModule customModule = new SimpleModule();
+    customModule.addSerializer(ByteBuffer.class, new HexStringCustomByteBufferSerializer());
+    MAPPER.registerModule(customModule);
```
rdblue, line 59: The majority of the Iceberg parsers don't use `ObjectMapper`. Why is it needed here?

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  public static JsonNode validateDefault(String name, Type type, JsonNode defaultValue) {
+    if (defaultValue != null && !isValidDefault(type, defaultValue)) {
+      throw new ValidationException("Invalid default value for field %s: %s not a %s", name, defaultValue, type);
+    }
+    return defaultValue;
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
```
rdblue, line 147: We generally don't suppress this warning unless we are trying to avoid major code changes. Since this is all new, it is best to follow the recommendation.

core/src/test/java/org/apache/iceberg/TestDefaultValuesValidationAndParsing.java
```diff
@@ -0,0 +1,99 @@
...
+         stringToJsonNode("{\"1\": 1, \"2\": \"bar\"}")}
+    });
+  }
+
+  private static JsonNode stringToJsonNode(String json) {
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      return mapper.readTree(json);
+    } catch (JsonProcessingException e) {
+      System.out.println("Failed to parse: " + json + "; reason: " + e.getMessage());
```
rdblue, line 80: Please remove print statements.

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
rzhang10: There's another dedicated method which does validation work `isValidDefault` which does this similar check already, and it should always be called before calling the current method `parseDefaultFromJson` which does the actual parsing.

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
rzhang10: The reason I put them together is that, the default value serialization and deserialization will depends on the APIs to access the default value stored inside the schema. So in order for the `jsonRoundTrip` test to work, it must depends on the methods introduced in the API..

Do you think we can first create a standalone PR with only API changes (I think that PR should be very minimum and fast for approval and merge), then I can rebase the changes and submit subsequent PRs? (The spark reader implementation PRs will also surely depends on the new API methods).

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
rzhang10: I use this mainly to help serialize the java in-mem `fixed` and `binary` type defaults which are represented by java `ByteBuffer` to our custom json representation that looks like `0x111f`, you can see the customized serialized logic in `HexStringCustomByteBufferSerializer`.

api/src/main/java/org/apache/iceberg/expressions/Literals.java
rzhang10: Yeah since previously I remember you pointed me to the `Literals` code so I thought I might be able to utilize the existing literals conversion logic there. 
My changes basically add the conversion from the 3 time/date-related type to string literal. Since I noticed the existing string literal code has the conversion to those date/time types, so I thought it might be a plus to add my current changes that can make the conversions more symmetric and enable people to do round-trip conversions?

WDYT?

api/src/main/java/org/apache/iceberg/expressions/Literal.java
rzhang10: See my below comment.

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
rzhang10: The reason I'm asking is even if I split all my later PRs, if the API changes are not checked in, those PRs won't compile themselves, or else I have to include the same API code changes in all the split PRs respectively.

rzhang10: Anyways, I have created a minimum API-change-only PR: https://github.com/apache/iceberg/pull/4732 , which took into account all your above style comments. I think this PR should be quite straightforward to be merged. And I plan to submit all subsequent split PRs by rebasing this after this has been merged.

api/src/main/java/org/apache/iceberg/expressions/Literals.java
rdblue: I don't think this conversion should be done using literals. You can see how we're doing it here, but there is no need to change the `Literals` API

api/src/main/java/org/apache/iceberg/types/Types.java
rdblue: @rzhang10, please don't resolve threads that are not updated yet.

core/src/main/java/org/apache/iceberg/DefaultValueParser.java
rdblue: Let's match the approach that the other parsers take so we have consistency in the code.

PR #4529, opened 2022-04-08
Add a new TableScan#appendInRange method that accepts nullable fromSn…

stevenzwu:
…apshotId as Long type (instead of current primitive long).

There are two reasons to support null fromSnapshotId
- Support inclusive streaming starting strategy. We can simply pass in the parentId of the starting snapshot, which can be null
- Support incremental streaming scan for a empty table, where the fromSnapshotId could be null

This change is binary compatible. Old interfaces are marked as deprecated.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/TableScan.java
+++ b/api/src/main/java/org/apache/iceberg/TableScan.java
@@ -147,8 +147,25 @@ public interface TableScan {
    * @param toSnapshotId read append data up to this snapshot id
    * @return a table scan which can read append data from {@code fromSnapshotId}
    * exclusive and up to {@code toSnapshotId} inclusive
+   *
+   * @deprecated Please use the appendsBetween with non-primitive fromSnapshotId of Long type
    */
-  TableScan appendsBetween(long fromSnapshotId, long toSnapshotId);
+  @Deprecated
+  default TableScan appendsBetween(long fromSnapshotId, long toSnapshotId) {
+    return appendsBetween(Long.valueOf(fromSnapshotId), toSnapshotId);
+  }
+
+  /**
+   * Create a new {@link TableScan} to read appended data from {@code fromSnapshotId} exclusive to {@code toSnapshotId}
+   * inclusive.
+   *
+   * @param fromSnapshotId the last snapshot id read by the user, exclusive. If it is null,
+   *                       all the ancestor snapshots of the toSnapshotId should be discovered
+   * @param toSnapshotId read append data up to this snapshot id
+   * @return a table scan which can read append data from {@code fromSnapshotId}
+   * exclusive and up to {@code toSnapshotId} inclusive
+   */
+  TableScan appendsBetween(Long fromSnapshotId, long toSnapshotId);
 
   /**
    * Create a new {@link TableScan} to read appended data from {@code fromSnapshotId} exclusive to the current snapshot
--- a/api/src/main/java/org/apache/iceberg/TableScan.java
+++ b/api/src/main/java/org/apache/iceberg/TableScan.java
@@ -152,7 +152,7 @@ public interface TableScan {
    */
   @Deprecated
   default TableScan appendsBetween(long fromSnapshotId, long toSnapshotId) {
-    return appendsBetween(Long.valueOf(fromSnapshotId), toSnapshotId);
+    return appendsInRange(Long.valueOf(fromSnapshotId), toSnapshotId);
   }
 
   /**
@@ -165,7 +165,7 @@ public interface TableScan {
    * @return a table scan which can read append data from {@code fromSnapshotId}
    * exclusive and up to {@code toSnapshotId} inclusive
    */
-  TableScan appendsBetween(Long fromSnapshotId, long toSnapshotId);
+  TableScan appendsInRange(Long fromSnapshotId, long toSnapshotId);
 
   /**
    * Create a new {@link TableScan} to read appended data from {@code fromSnapshotId} exclusive to the current snapshot
--- a/api/src/main/java/org/apache/iceberg/events/IncrementalTableScanEvent.java
+++ b/api/src/main/java/org/apache/iceberg/events/IncrementalTableScanEvent.java
@@ -24,19 +24,16 @@ import org.apache.iceberg.expressions.Expression;
 
 /**
  * Event sent to listeners when an incremental table scan is planned.
- *
- * @deprecated by {@link IncrementalTableScanEvent}
  */
-@Deprecated
-public final class IncrementalScanEvent {
+public final class IncrementalTableScanEvent {
   private final String tableName;
-  private final long fromSnapshotId;
+  private final Long fromSnapshotId;
   private final long toSnapshotId;
   private final Expression filter;
   private final Schema projection;
 
-  public IncrementalScanEvent(String tableName, long fromSnapshotId, long toSnapshotId, Expression filter,
-                              Schema projection) {
+  public IncrementalTableScanEvent(String tableName, Long fromSnapshotId, long toSnapshotId, Expression filter,
+                                   Schema projection) {
     this.tableName = tableName;
     this.fromSnapshotId = fromSnapshotId;
     this.toSnapshotId = toSnapshotId;
@@ -48,7 +45,7 @@ public final class IncrementalScanEvent {
     return tableName;
   }
 
-  public long fromSnapshotId() {
+  public Long fromSnapshotId() {
     return fromSnapshotId;
   }
 
--- a/core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
+++ b/core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
@@ -40,7 +40,7 @@ abstract class BaseAllMetadataTableScan extends BaseMetadataTableScan {
   protected abstract String tableType();
 
   @Override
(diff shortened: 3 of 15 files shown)


api/src/main/java/org/apache/iceberg/events/IncrementalTableScanEvent.java
```diff
@@ -0,0 +1,63 @@
...
+    this.toSnapshotId = toSnapshotId;
+    this.filter = filter;
+    this.projection = projection;
+  }
+
+  public String tableName() {
+    return tableName;
+  }
+
+  public Long fromSnapshotId() {
```
stevenzwu, line 48: This class is cloned from `IncrementalScanEvent` because we can't change the return type from `Long` to `long` without breaking binary compatibility.

stevenzwu:
We can't really overload the `appendBetween` with long and Long. Otherwise, we are going to see compiling error (Ambiguous) like this.

```
> Task :iceberg-flink:iceberg-flink-1.14:compileJava
        scan = scan.appendsBetween(context.startSnapshotId(), context.endSnapshotId());
                   ^
  both method appendsBetween(long,long) in org.apache.iceberg.TableScan and method appendsBetween(java.lang.Long,long) in org.apache.iceberg.TableScan match
```

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -147,8 +147,25 @@ default TableScan select(String... columns) {
...
+   * inclusive.
+   *
+   * @param fromSnapshotId the last snapshot id read by the user, exclusive. If it is null,
+   *                       all the ancestor snapshots of the toSnapshotId should be discovered
+   * @param toSnapshotId read append data up to this snapshot id
+   * @return a table scan which can read append data from {@code fromSnapshotId}
+   * exclusive and up to {@code toSnapshotId} inclusive
    */
-  TableScan appendsBetween(long fromSnapshotId, long toSnapshotId);
+  TableScan appendsInRange(Long fromSnapshotId, long toSnapshotId);
```
stevenzwu, line 168: We can't use the same method name. otherwise, we ran into the following resolution error for method overloading.

```
/home/runner/work/iceberg/iceberg/spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/source/Reader.java:172: error: reference to appendsBetween is ambiguous

        scan = scan.appendsBetween(startSnapshotId, endSnapshotId);
                   ^
  both method appendsBetween(long,long) in org.apache.iceberg.TableScan and method appendsBetween(java.lang.Long,long) in org.apache.iceberg.TableScan match
```

stevenzwu: Another naming option is `appendsBetweenScan`

core/src/test/java/org/apache/iceberg/TestIncrementalDataTableScan.java
```diff
@@ -60,7 +60,7 @@ public void testInvalidScans() {
     AssertHelpers.assertThrows(
         "from and to snapshots cannot be the same, since from snapshot is exclusive and not part of the scan",
         IllegalArgumentException.class, "from and to snapshot ids cannot be the same",
-        () -> appendsBetweenScan(1, 1));
+        () -> appendsBetweenScan(1L, 1L));
```
stevenzwu, line 63: need to change the `fromSnapshotId` from `1` to `1L` because the type is `Long` now. there is no auto type conversion from `int` to `Long`.

Since I am changing the constants for `fromSnapshotId`, I also changed the `toSnapshotId` as well to be consistent.

stevenzwu:
@rdblue @openinx @yittg @kbendick please take a look when you get a chance

yittg:
thanks @stevenzwu for providing this change.
Personally, i'm not quite sure should we use a such big change to achieve this target.
I wonder whether it is a better option to use a special invalid snapshot id value to do the same thing, such as `-1`. Checking this special value can be equivalent with null checking.

Or please correct me if it has other effects. Also looking forward others' opinions.

stevenzwu:
@yittg this proposal is after the discussion with Ryan. The original plan was to provide an overloaded `appendsBetween` with `long` and `Long`. But during implementation, I found that overload caused compiling error as I mentioned in one of the comments. So I have to rename the new method to `appendsInRange`, which is not as good a name as `appendsBetween`. This PR is actually relatively small change even though it touched many files.

Regarding special value, I am not sure if there is a special invalid snapshot id that Iceberg defines in the table format spec. The only special value is `null`. E.g., the earliest snapshot has `parent-snapshot-id` as null.

yittg:
@stevenzwu Thanks for your explanation and great work.

core/src/test/java/org/apache/iceberg/TestIncrementalDataTableScan.java
```diff
@@ -303,10 +306,8 @@ private List<String> appendsAfterScan(long fromSnapshotId) {
...
   }
 
-  private List<String> appendsBetweenScan(long fromSnapshotId, long toSnapshotId) {
-    Snapshot s1 = table.snapshot(fromSnapshotId);
-    Snapshot s2 = table.snapshot(toSnapshotId);
-    TableScan appendsBetween = table.newScan().appendsBetween(s1.snapshotId(), s2.snapshotId());
+  private List<String> appendsBetweenScan(Long fromSnapshotId, long toSnapshotId) {
+    TableScan appendsBetween = table.newScan().appendsInRange(fromSnapshotId, toSnapshotId);
     return filesToScan(appendsBetween);
   }
```
yittg, line 312: How about adding another method, and keeping the original?
```suggestion

  private List<String> appendsBeforeScan(long toSnapshotId) {
    TableScan appendsBetween = table.newScan().appendsInRange(null, toSnapshotId);
    return filesToScan(appendsBetween);
  }

  private List<String> appendsBetweenScan(long fromSnapshotId, long toSnapshotId) {
    TableScan appendsBetween = table.newScan().appendsInRange(fromSnapshotId, toSnapshotId);
    return filesToScan(appendsBetween);
  }
```

yittg (approved):
LGTM

core/src/test/java/org/apache/iceberg/TestIncrementalDataTableScan.java
stevenzwu: This change seems unnecessary. let me try to revert it and minimize the change in this test class

yittg: Are there still cases where startSnapshotId is `null` after reverting?

stevenzwu: Not in the existing unit test of `TestIncrementalDataTableScan`. But this reminds me that maybe we should add unit test with null startSnapshotId for the `appendsInRange`. But we do not need to change this method. if there is a need, we can create a new method as you suggested.

yittg: it's better now from my side. thanks

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -147,8 +147,25 @@ default TableScan select(String... columns) {
...
+  default TableScan appendsBetween(long fromSnapshotId, long toSnapshotId) {
+    return appendsInRange(Long.valueOf(fromSnapshotId), toSnapshotId);
+  }
+
+  /**
+   * Create a new {@link TableScan} to read appended data from {@code fromSnapshotId} exclusive to {@code toSnapshotId}
+   * inclusive.
+   *
+   * @param fromSnapshotId the last snapshot id read by the user, exclusive. If it is null,
+   *                       all the ancestor snapshots of the toSnapshotId should be discovered
```
chenjunjiedada, line 163: do you mean all ancestor snapshots should be included? right?

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
```diff
@@ -40,7 +40,7 @@ abstract class BaseAllMetadataTableScan extends BaseMetadataTableScan {
   protected abstract String tableType();
 
   @Override
-  public TableScan appendsBetween(long fromSnapshotId, long toSnapshotId) {
+  public TableScan appendsInRange(Long fromSnapshotId, long toSnapshotId) {
```
chenjunjiedada, line 43: In the future, we want to support reading overwrite snapshots so maybe we could change to `dataBetween(startSnaphotId, endSnasphotId)`.

stevenzwu: if the community agrees with PR #4580 , we will abort this PR/approach. In PR #4580 , we can add more flexible `IncrementalScan` options (not just append) to support use cases like CDC read.

api/src/main/java/org/apache/iceberg/TableScan.java
stevenzwu: correct. let's say the snapshot history is Sn -> Sn+1 -> Sn+2 -> Sn+3. `appendsBetween(null, Sn+3) should discover files appended by all 4 snapshots.

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
chenjunjiedada: Cool, let me take a look.

stevenzwu:
close this one in favor of PR #4580

PR #4530, opened 2022-04-10
Spark [3.2] : Improve #4446 stats estimation for spark scan with delete files

felixYyu:
Improve #4446 with delete files.  cc @singhpk234 @rdblue

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
@@ -24,13 +24,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
-import org.apache.iceberg.CombinedScanTask;
-import org.apache.iceberg.FileScanTask;
-import org.apache.iceberg.Schema;
-import org.apache.iceberg.SchemaParser;
-import org.apache.iceberg.Snapshot;
-import org.apache.iceberg.SnapshotSummary;
-import org.apache.iceberg.Table;
+
+import org.apache.iceberg.*;
 import org.apache.iceberg.expressions.Expression;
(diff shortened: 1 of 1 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -148,8 +149,9 @@ protected Statistics estimateStatistics(Snapshot snapshot) {
 
     for (CombinedScanTask task : tasks()) {
       for (FileScanTask file : task.files()) {
-        // TODO: if possible, take deletes also into consideration.
-        double fractionOfFileScanned = ((double) file.length()) / file.file().fileSizeInBytes();
+        // take deletes also into consideration.
+        long deletesSizeInBytes = file.deletes().stream().mapToLong(DeleteFile::fileSizeInBytes).sum();
+        double fractionOfFileScanned = ((double) file.length()) / (file.file().fileSizeInBytes() + deletesSizeInBytes);
```
singhpk234, line 154: Thanks @felixYyu for your contribution.

As per my understanding a single delete file can be associated with multiple data files, (also position delete doesn't store complete deleted row info also (ref [this](https://iceberg.apache.org/spec/#position-delete-files)) there is also case of equality deletes (ref : [this](https://iceberg.apache.org/spec/#equality-delete-files)) where we just store the matching pred, hence it will not be so straight forward to estimate size / row count.

Hence row count post applying delete files will be hard to estimate without actually applying them, hence I left it as a todo.

From the looks of it, this approach looks incorrect to me (may be I am wrong here) will leave it to the better judgement of more experienced folks of iceberg.

rdblue:
I agree with @singhpk234. Because delete files can affect more than one file and not all deletes in a file are for a given data file, this is not a good approach for correctly estimating the final row count. In addition, file size is used to scale the row count within a single file. Delete files could have a completely different compression ratio so this is going to be very unreliable.

PR #4534, opened 2022-04-11
Use memory-backed streams in tests

findepi:
Sometimes an actual file is needed, but sometimes it's not. In the case
where it isn't, we can use an in-memory `InputFile`, `OutputFile`
implementation.

Extracted from https://github.com/apache/iceberg/pull/4537

### DIFF
--- a/core/src/test/java/org/apache/iceberg/TestManifestWriterVersions.java
+++ b/core/src/test/java/org/apache/iceberg/TestManifestWriterVersions.java
@@ -24,6 +24,7 @@ import java.util.List;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileAppender;
 import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.io.InMemoryOutputFile;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
@@ -235,7 +236,7 @@ public class TestManifestWriterVersions {
   }
 
   private InputFile writeManifestList(ManifestFile manifest, int formatVersion) throws IOException {
-    OutputFile manifestList = Files.localOutput(temp.newFile());
+    OutputFile manifestList = new InMemoryOutputFile();
     try (FileAppender<ManifestFile> writer = ManifestLists.write(
         formatVersion, manifestList, SNAPSHOT_ID, SNAPSHOT_ID - 1, formatVersion > 1 ? SEQUENCE_NUMBER : 0)) {
       writer.add(manifest);
--- a/core/src/test/java/org/apache/iceberg/avro/TestGenericAvro.java
+++ b/core/src/test/java/org/apache/iceberg/avro/TestGenericAvro.java
@@ -19,25 +19,22 @@
 
 package org.apache.iceberg.avro;
 
-import java.io.File;
 import java.io.IOException;
 import java.util.List;
 import org.apache.avro.generic.GenericData.Record;
-import org.apache.iceberg.Files;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.io.FileAppender;
+import org.apache.iceberg.io.InMemoryOutputFile;
+import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
-import org.junit.Assert;
 
 public class TestGenericAvro extends AvroDataTest {
   @Override
   protected void writeAndValidate(Schema schema) throws IOException {
     List<Record> expected = RandomAvroData.generate(schema, 100, 0L);
 
-    File testFile = temp.newFile();
-    Assert.assertTrue("Delete should succeed", testFile.delete());
-
-    try (FileAppender<Record> writer = Avro.write(Files.localOutput(testFile))
+    OutputFile outputFile = new InMemoryOutputFile();
+    try (FileAppender<Record> writer = Avro.write(outputFile)
         .schema(schema)
         .named("test")
         .build()) {
@@ -47,7 +44,7 @@ public class TestGenericAvro extends AvroDataTest {
     }
 
     List<Record> rows;
-    try (AvroIterable<Record> reader = Avro.read(Files.localInput(testFile))
+    try (AvroIterable<Record> reader = Avro.read(outputFile.toInputFile())
         .project(schema)
         .build()) {
       rows = Lists.newArrayList(reader);
--- a/core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
+++ b/core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.io;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.UUID;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+public class InMemoryInputFile implements InputFile {
+
+  private final String location;
+  private final byte[] contents;
+
+  public InMemoryInputFile(byte[] contents) {
+    this("memory:" + UUID.randomUUID(), contents);
+  }
+
+  public InMemoryInputFile(String location, byte[] contents) {
+    Preconditions.checkNotNull(location, "location is null");
+    Preconditions.checkNotNull(contents, "contents is null");
+    this.location = location;
+    this.contents = contents.clone();
+  }
+
+  @Override
+  public long getLength() {
+    return contents.length;
+  }
+
+  @Override
+  public SeekableInputStream newStream() {
+    return new InMemorySeekableInputStream(contents);
+  }
+
+  @Override
+  public String location() {
+    return location;
+  }
+
+  @Override
+  public boolean exists() {
+    return true;
+  }
+
+  private static class InMemorySeekableInputStream extends SeekableInputStream {
+
+    private final int length;
+    private final ByteArrayInputStream delegate;
+
+    InMemorySeekableInputStream(byte[] contents) {
+      this.length = contents.length;
+      this.delegate = new ByteArrayInputStream(contents);
+    }
+
+    @Override
+    public long getPos() throws IOException {
+      return length - delegate.available();
+    }
+
+    @Override
+    public void seek(long newPos) throws IOException {
+      delegate.reset(); // resets to a marked position
+      Preconditions.checkState(delegate.skip(newPos) == newPos,
+          "Invalid position %s within stream of length %s", newPos, length);
+    }
+
+    @Override
+    public int read() {
+      return delegate.read();
+    }
+
+    @Override
+    public int read(byte[] b) throws IOException {
+      return delegate.read(b);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) {
+      return delegate.read(b, off, len);
+    }
+
+    @Override
+    public long skip(long n) {
+      return delegate.skip(n);
+    }
+
+    @Override
+    public int available() {
+      return delegate.available();
+    }
+
+    @Override
+    public boolean markSupported() {
+      return false;
+    }
+
+    @Override
+    public void mark(int readAheadLimit) {
+      // The delegate's mark is used to implement seek
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void reset() {
+      delegate.reset();
+    }
+
+    @Override
+    public void close() throws IOException {
+      delegate.close();
+    }
+  }
+}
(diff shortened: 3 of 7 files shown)


core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+package org.apache.iceberg.io;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.UUID;
+
+import static java.util.Objects.requireNonNull;
+import static org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkState;
```
rdblue, line 27: We typically don't use static method imports. I think they are mostly disallowed by the Iceberg style enforcement.

findepi: I noticed the `Test*` classes are exempt from the checks. I will fix this one nonetheless.

findepi:
AC

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -0,0 +1,130 @@
...
+    public long getPos() throws IOException {
+      return length - delegate.available();
+    }
+
+    @Override
+    public void seek(long newPos) throws IOException {
+      delegate.reset();
+      Preconditions.checkState(delegate.skip(newPos) == newPos,
+          "Invalid position %s within stream of length %s", newPos, length);
+    }
```
kbendick, line 82: Nit: We should consider adding an explicit check that the new position is non-negative.

The `checkState` call seems like it will cover that as `skip` shouldn't return a negative number, but it might be better for tests to add an explicit check. WDYT?

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Objects;
+import java.util.UUID;
+
+public class InMemoryOutputFile implements OutputFile {
+
+  private final String location;
+
+  private boolean exists;
```
kbendick, line 31: Do we explicitly set `exists` to `false` anywhere or are we relying on the default value being false?

I'd prefer we explicitly set the value false somewhere for readability purposes.

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  public InMemoryOutputFile(String location) {
+    this.location = Objects.requireNonNull(location, "location is null");
+  }
+
+  @Override
+  public PositionOutputStream create() {
+    if (exists) {
+      throw new RuntimeException("Already exists");
```
kbendick, line 45: Can we throw `AlreadyExistsException` to match the current `LocalOutputFile#create()` method?

https://github.com/apache/iceberg/blob/e45c19e1d02a130d33ac886f676dfb141708090d/api/src/main/java/org/apache/iceberg/Files.java#L57-L59

kbendick (commented):
Thanks @findepi! I know this is something that has been considered before, so it's good to see it being added.

Hopefully this will make tests that much faster. Left some style notes and a few comments.

core/src/test/java/org/apache/iceberg/TestManifestWriterVersions.java
```diff
@@ -235,7 +236,7 @@ void checkRewrittenManifest(ManifestFile manifest, long expectedSequenceNumber,
   }
 
   private InputFile writeManifestList(ManifestFile manifest, int formatVersion) throws IOException {
-    OutputFile manifestList = Files.localOutput(temp.newFile());
+    InMemoryOutputFile manifestList = new InMemoryOutputFile();
```
RussellSpitzer, line 239: Could just still be OutputFile right?

core/src/test/java/org/apache/iceberg/avro/TestGenericAvro.java
```diff
@@ -19,25 +19,21 @@
...
 public class TestGenericAvro extends AvroDataTest {
   @Override
   protected void writeAndValidate(Schema schema) throws IOException {
     List<Record> expected = RandomAvroData.generate(schema, 100, 0L);
 
-    File testFile = temp.newFile();
-    Assert.assertTrue("Delete should succeed", testFile.delete());
-
-    try (FileAppender<Record> writer = Avro.write(Files.localOutput(testFile))
+    InMemoryOutputFile outputFile = new InMemoryOutputFile();
```
RussellSpitzer, line 35: Could just be output File

RussellSpitzer (approved):
This makes sense to me, although i was wondering when I saw the title whether you were going to implement a full FileIO since we could be using that now. I think this is fine for these tests though

kbendick (approved):
Outside of the out of date comment related to `mark` and the question around throwing for `markSupported`, this looks good for the current testing purposes.

My other comments are more informational or for clarification.

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Override
+    public boolean markSupported() {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void mark(int readAheadLimit) {
+      // We use mark to implement seek
```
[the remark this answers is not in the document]
findepi: in seek(), the `delegate.reset()` resets to a marked position, will add a comment

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public InputFile toInputFile() {
+    return new InMemoryInputFile(location(), getContents());
+  }
+
+  public byte[] getContents() {
+    return contents.toByteArray();
+  }
```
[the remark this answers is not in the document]
findepi: changed to `toByteArray`, will this be OK?

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Override
+  public String location() {
+    return location;
+  }
+
+  @Override
+  public boolean exists() {
+    return true;
+  }
```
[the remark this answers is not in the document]
findepi: currently `InMemoryInputFIle` can only represent a file that exists.
in the future, if tests need that, we can evolve the `InMemoryInputFIle` interface to allow representing a file that doesn't exist

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
findepi: I think that it's just tests, so exception message doesn't matter that much, and adding another check doesn't make the existing check unnecessary. i would leave this as is, if you don't mind

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
findepi: Added an explicit initializer.

Note that in Trino we have a hard rule that a field must not have an explicit initializer that is same as the default value for given type, so i am afraid I will need to be reminded about this from time to time. Please bear with me

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
findepi: Thanks, indeed this is what interface requires me to do

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
```diff
@@ -0,0 +1,108 @@
...
+package org.apache.iceberg.io;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Objects;
+import java.util.UUID;
+
+public class InMemoryOutputFile implements OutputFile {
+
+  private final String location;
```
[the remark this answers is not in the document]
findepi: > I believe this is to match the behavior of `file.toString()`, which might be relied on elsewhere.

toString representation mustn't be relied on, unless its contract is defined by an interface (it isn't here)

location-returning toString wouldn't be useful, as this is just a UUID.

core/src/test/java/org/apache/iceberg/TestManifestWriterVersions.java
findepi: reverted

findepi: yes, fixed (not reverted)

core/src/test/java/org/apache/iceberg/avro/TestGenericAvro.java
findepi: fixed

findepi:
> This makes sense to me, although i was wondering when I saw the title whether you were going to implement a full FileIO since we could be using that now. I think this is fine for these tests though

Not until i need that :)

findepi:
AC

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -0,0 +1,131 @@
...
+public class InMemoryInputFile implements InputFile {
+
+  private final String location;
+  private final byte[] contents;
+
+  public InMemoryInputFile(byte[] contents) {
+    this("memory:" + UUID.randomUUID(), contents);
+  }
+
+  public InMemoryInputFile(String location, byte[] contents) {
```
kbendick, line 36: As a side note: Typically we use `ByteBuffer`s instead of raw `byte[]`, so that we can get views over slices etc for free and help avoid additional allocations in many situations.

ByteBuffers also have the advantage of presenting views over type-specific byte arrays (for `char`, `int`, `float`, etc). And these views are indexed in terms of the type-specific size of the values, which is easier to reason about. The type-specific views also offer direct access to the backing byte buffer as well as bulk get and put methods for transferring large contiguous sequences of values.

This is test code, but if we were to ever move this to be code used in the project, we'd almost certainly want to use `ByteBuffer` instead of the raw `byte[]` throughout.

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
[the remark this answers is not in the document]
kbendick: That makes sense to me. We don't need it yet by any means then.

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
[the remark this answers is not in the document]
kbendick: Oh yeah I see that. I was thinking `mark` this function, and not the `mark` that's underlying inside the delegate ByteArrayInputStream.

Maybe updating this comment to be `The delegate's mark is used to implement seek` would be more clear? Not required though. You could also just get rid of the comment though. And the statement isn't technically incorrect anyway.

Thanks for the clarification 👍

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -0,0 +1,131 @@
...
+    Preconditions.checkNotNull(location, "location is null");
+    Preconditions.checkNotNull(contents, "contents is null");
+    this.location = location;
+    this.contents = contents.clone();
+  }
```
kbendick, line 41: As a side note: Typically we use `ByteBuffer`s instead of raw `byte[]`, so that we can get views over slices etc for free and help avoid additional allocations in many situations.

ByteBuffers also have the advantage of presenting views over type-specific byte arrays (for `char`, `int`, `float`, etc). And these views are indexed in terms of the type-specific size of the values, which is easier to reason about. The type-specific views also offer direct access to the backing byte buffer as well as bulk get and put methods for transferring large contiguous sequences of values.

This is test code, but if we were to ever move this to be code used in the project, we'd almost certainly want to use `ByteBuffer` instead of the raw `byte[]` throughout where possible.

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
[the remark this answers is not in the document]
kbendick: That's fair. I'm pretty sure we removed any and all reliance on inappropriate usages of `toString` quite some time ago. Since it's not part of the contract, it's just added overhead then. 👍

core/src/test/java/org/apache/iceberg/io/InMemoryOutputFile.java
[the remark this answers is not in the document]
kbendick: Yeah that's great. Thank you!

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
findepi: For internal representation `byte[]` is preferred, since the i can `ByteArray{Output,Input}Stream}`.
For constructor -- whatever it's easier to provide by the caller. I think we eventually will accept both.

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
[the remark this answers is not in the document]
findepi: changed to `// The delegate's mark is used to implement seek`, thanks for the suggestion!

findepi:
AC

All comments applied or answered to.
Do you want me to mark them as resolved?

@RussellSpitzer @kbendick do you mind to take another look and let me know what else I could change here?

findepi:
@RussellSpitzer @rdblue can you please give me more advice here?

RussellSpitzer (approved):

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
kbendick: Yeah I'm good with that. If this were ported outside of test, I just wanted to make sure you were aware that request might come up (though it might not depending on the situation) 🙂

kbendick (approved):
Reconfirming my +1. This looks good to me. Thanks @findepi!

RussellSpitzer:
Thanks @findepi for the contribution and @kbendick and @rdblue for review

findepi:
thank you for the merge!

PR #4535, opened 2022-04-11
Core: do not allow setting fields nested in optional struct as identi…

zhongyujiang:
…fier fields and do not allow deleting columns with nested identifier fields

This PR does 2 things:
1.  Disable setting required fields nested in optional struct as identifier fields. Currently we only check if the identifier fields is required to prevent nullable identifier fields, but a required field nested in an optional struct is also nullable;
2. Disable deletion of struct type columns with nested identifier fields.

@jackye1995 Could you help review this? Thanks!

### DIFF
--- a/core/src/main/java/org/apache/iceberg/SchemaUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/SchemaUpdate.java
@@ -439,12 +439,21 @@ class SchemaUpdate implements UpdateSchema {
                                      Multimap<Integer, Move> moves,
                                      Set<String> identifierFieldNames) {
     // validate existing identifier fields are not deleted
+    Map<Integer, Integer> idToParent = TypeUtil.indexParents(schema.asStruct());
+
     for (String name : identifierFieldNames) {
       Types.NestedField field = schema.findField(name);
       if (field != null) {
         Preconditions.checkArgument(!deletes.contains(field.fieldId()),
             "Cannot delete identifier field %s. To force deletion, " +
                 "also call setIdentifierFields to update identifier fields.", field);
+        Integer parentId = idToParent.get(field.fieldId());
+        while (parentId != null) {
+          Preconditions.checkArgument(!deletes.contains(parentId),
+              "Cannot delete field %s as it will delete nested identifier field %s",
+              schema.findField(parentId), field);
+          parentId = idToParent.get(parentId);
+        }
       }
     }
 
@@ -462,7 +471,6 @@ class SchemaUpdate implements UpdateSchema {
       freshIdentifierFieldIds.add(nameToId.get(name));
     }
 
-    Map<Integer, Integer> idToParent = TypeUtil.indexParents(schema.asStruct());
     Map<Integer, Types.NestedField> idToField = TypeUtil.indexById(struct);
     freshIdentifierFieldIds.forEach(id -> Schema.validateIdentifierField(id, idToField, idToParent));
 
--- a/core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
+++ b/core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
@@ -1261,17 +1261,24 @@ public class TestSchemaUpdate {
   @Test
   public void testAddNestedIdentifierFieldColumns() {
     Schema newSchema = new SchemaUpdate(SCHEMA, SCHEMA_LAST_COLUMN_ID)
-        .setIdentifierFields("preferences.feature1")
+        .allowIncompatibleChanges()
+        .addRequiredColumn("required_struct", Types.StructType.of(
+            Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 2, "field", Types.StringType.get())
+        ))
+        .apply();
+
+    newSchema = new SchemaUpdate(newSchema, SCHEMA_LAST_COLUMN_ID + 2)
+        .setIdentifierFields("required_struct.field")
         .apply();
 
     Assert.assertEquals("set existing nested field as identifier should succeed",
-        Sets.newHashSet(newSchema.findField("preferences.feature1").fieldId()),
+        Sets.newHashSet(newSchema.findField("required_struct.field").fieldId()),
         newSchema.identifierFieldIds());
 
     newSchema = new SchemaUpdate(SCHEMA, SCHEMA_LAST_COLUMN_ID)
         .allowIncompatibleChanges()
         .addRequiredColumn("new", Types.StructType.of(
-            Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 1, "field", Types.StringType.get())
+            Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 2, "field", Types.StringType.get())
         ))
         .setIdentifierFields("new.field")
         .apply();
@@ -1283,8 +1290,8 @@ public class TestSchemaUpdate {
     newSchema = new SchemaUpdate(SCHEMA, SCHEMA_LAST_COLUMN_ID)
         .allowIncompatibleChanges()
         .addRequiredColumn("new", Types.StructType.of(
-            Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 1, "field", Types.StructType.of(
-                Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 2, "nested", Types.StringType.get())))))
+            Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 2, "field", Types.StructType.of(
+                Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 3, "nested", Types.StringType.get())))))
         .setIdentifierFields("new.field.nested")
         .apply();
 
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -438,13 +438,18 @@ private static Schema applyChanges(Schema schema, List<Integer> deletes,
...
-        Preconditions.checkArgument(!deletes.contains(field.fieldId()),
-            "Cannot delete identifier field %s. To force deletion, " +
-                "also call setIdentifierFields to update identifier fields.", field);
+        Integer idNotToDelete = field.fieldId();
+        while (idNotToDelete != null) {
+          Preconditions.checkArgument(!deletes.contains(idNotToDelete),
+              "Cannot delete identifier field %s. To force deletion, " +
+                  "also call setIdentifierFields to update identifier fields.", field);
+          idNotToDelete = idToParent.get(idNotToDelete);
+        }
```
zhongyujiang, line 452: Now deleting struct type columns with nested identifier fields will also fail bacause this check, just the failure message can be a little confusing.
https://github.com/apache/iceberg/blob/910f2712c7b731181c4a72fb6b8c66addeb76dd9/core/src/main/java/org/apache/iceberg/SchemaUpdate.java#L456-L463

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -438,13 +438,18 @@ private static Schema applyChanges(Schema schema, List<Integer> deletes,
...
-    // validate existing identifier fields are not deleted
+    Map<Integer, Integer> idToParent = TypeUtil.indexParents(schema.asStruct());
+
     for (String name : identifierFieldNames) {
       Types.NestedField field = schema.findField(name);
       if (field != null) {
-        Preconditions.checkArgument(!deletes.contains(field.fieldId()),
-            "Cannot delete identifier field %s. To force deletion, " +
-                "also call setIdentifierFields to update identifier fields.", field);
+        Integer idNotToDelete = field.fieldId();
```
szehon-ho, line 446: Can we keep the original Precondition check as is, and then have code after to check parents.  

The message can have more information in the new case, and the code is clearer
```
 Preconditions.checkArgument(!deletes.contains(field.fieldId()),
            "Cannot delete identifier field %s. To force deletion, " +
                "also call setIdentifierFields to update identifier fields.", field);
Integer parentId = fieldId;
while (parentId != null) {
   Preconditions.checkArgument(... "Cannot delete field %s as it will delete nested identifier field %s",...)
   parentId = idToParent.get(parentId);
}
```

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
```diff
@@ -1261,14 +1261,6 @@ public void testAddNewIdentifierFieldColumns() {
   @Test
   public void testAddNestedIdentifierFieldColumns() {
     Schema newSchema = new SchemaUpdate(SCHEMA, SCHEMA_LAST_COLUMN_ID)
-        .setIdentifierFields("preferences.feature1")
-        .apply();
-
-    Assert.assertEquals("set existing nested field as identifier should succeed",
```
szehon-ho, line 1267: I dont think we should remove this test?

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
```diff
@@ -1476,6 +1480,20 @@ public void testDeleteIdentifierFieldColumnsFails() {
         "Cannot delete identifier field 1: id: required int. To force deletion, " +
             "also call setIdentifierFields to update identifier fields.",
         () -> new SchemaUpdate(schemaWithIdentifierFields, SCHEMA_LAST_COLUMN_ID).deleteColumn("id").apply());
+
+    Schema newSchema = new SchemaUpdate(SCHEMA, SCHEMA_LAST_COLUMN_ID)
```
szehon-ho, line 1484: Can we split this out into new test method, to make it easier?

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
```diff
@@ -1438,12 +1435,19 @@ public void testSetIdentifierFieldsFails() {
...
 
-    AssertHelpers.assertThrows("add a nested field in struct of a map should fail",
+    AssertHelpers.assertThrows("add a nested field in struct of a list should fail",
         IllegalArgumentException.class,
         "must not be nested in " + newSchema.findField("new.fields"),
         () -> new SchemaUpdate(newSchema, lastColId)
             .setIdentifierFields("new.fields.element.nested")
             .apply());
+
+    AssertHelpers.assertThrows("add a nested field in an optional struct which should fail",
```
szehon-ho, line 1445: We don't need "which"

szehon-ho (changes requested):
I think it makes sense, some style comments.

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
zhongyujiang: Will update.

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
zhongyujiang: This will fail now because `preferences ` is optional, I'll add test for a required struct field.

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
```diff
@@ -1478,6 +1497,23 @@ public void testDeleteIdentifierFieldColumnsFails() {
...
+  public void testDeleteContainingNestedIdentifierFieldColumnsFails() {
+    Schema newSchema = new SchemaUpdate(SCHEMA, SCHEMA_LAST_COLUMN_ID)
+        .allowIncompatibleChanges()
+        .addRequiredColumn("out", Types.StructType.of(
+            Types.NestedField.required(SCHEMA_LAST_COLUMN_ID + 2, "nested", Types.StringType.get())
+        ))
+        .setIdentifierFields("out.nested")
+        .apply();
+
+    AssertHelpers.assertThrows("delete an identifier column without setting identifier fields should fail",
```
szehon-ho, line 1510: nit: the message can be clearer like 'delete a struct with a nested identifier column without setting identifier fields should fail'?

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
```diff
@@ -1388,34 +1395,39 @@ public void testSetIdentifierFieldsFails() {
             .setIdentifierFields("locations.key.zip")
             .apply());
 
-    AssertHelpers.assertThrows("add a nested field in list should fail",
```
szehon-ho, line 1391: How about this test, it should not need to be removed right?  (if i understand, the first check is still about list).  

Seems the new test "required_list.element.x" is about required list, this can be kept and test about optional list.

szehon-ho (commented):
Thanks for the changes, just two more very small comments, should be good to go after.

core/src/test/java/org/apache/iceberg/TestSchemaUpdate.java
zhongyujiang: Yes it still fails, only now the reason it fails is because the hidden middle  field `element` is optional:
> "Cannot add field x as an identifier field: must not be nested in an optional field 14: element: optional struct<15: x: required long, 16: y: required long>"
to contain:

This test is against "list" type, so I added the "required_list.element.x" test to make sure the Precondtion will fail for the "list" type reason, to replace it.

szehon-ho: Ah i see,  it checks from bottom upward , and hits the "element" optional check first.  What do you think, we still keep the test as "add a nested field in optional list should fail", and change the assert message?  This test  may still have value as there's not another that test this case, unless I am missing seeing it.

zhongyujiang: I think we can keep this test, and I think changing the validation logic might be a better option than changing the assert message. 
I feel we shouldn't expose the `element` or  `key-value` field in Preconditions error message because these fields are invisible to the users and may confuse them. I think a validation from top to bottom can give a more correct error message. 
```
Deque<Integer> deque = Lists.newLinkedList();
while (parentId != null) {
  deque.push(parentId);
  parentId = idToParent.get(parentId);
}

while (!deque.isEmpty()) {
  Types.NestedField parent = idToField.get(deque.pop());
  Preconditions.checkArgument(parent.type().isStructType(),
      "Cannot add field %s as an identifier field: must not be nested in %s", field.name(), parent);
  Preconditions.checkArgument(parent.isRequired(),
      "Cannot add field %s as an identifier field: must not be nested in an optional field %s",
      field.name(), parent);
}
```
WDYT?

szehon-ho: Yea, that sounds good to me

szehon-ho: Maybe can add a small code comment, like 'Exploring from root for better error message for list and map types'?

zhongyujiang: Will add.

zhongyujiang: Done.

szehon-ho (approved):

szehon-ho:
Merged, thanks @zhongyujiang

PR #4536, opened 2022-04-11
Docs: add Cloduera native docs section

marton-bod:


marton-bod:
PR for the iceberg-docs: https://github.com/apache/iceberg-docs/pull/72

docs/cloudera/_index.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "Cloudera"
+bookIconImage: ../img/cloudera-logo.png
+bookFlatSection: true
+weight: 410
```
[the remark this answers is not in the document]
marton-bod: Sure, will change it. Maybe worth adding a comment in a readme so that others will know about this requirement too? (unless this becomes obsolete with the new docs page design anyways)

docs/cloudera/_index.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "Cloudera"
+bookIconImage: ../img/cloudera-logo.png
+bookFlatSection: true
+weight: 415
+bookExternalUrlNewWindow: https://docs.cloudera.com/cdp/latest/cdp-iceberg/topics/iceberg-in-cdp.html
```
rdblue, line 6: Do we want to link to vendor docs rather than to the open source docs for Impala and Hive?

This seems like it doesn't fit with the current set of links, which are links to projects rather than to companies. Maybe we could use a "vendor" page to link instead?

pvary: We are currently working on updating the docs:
- One part of it is updating the OS docs - which is a bigger task. We plan to keep most of the info here, as our strategy is OS first, as always. Our doc team is currently working on putting this together
- Also we would like to add a single link in the group of the vendor specific docs, which would point to the CDPD. Similarly as we (Iceberg community) did in case of other vendors:
  - [ Dremio](https://docs.dremio.com/data-formats/apache-iceberg/)
  - [ Amazon Athena](https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg.html)
  - [ Amazon EMR](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-iceberg-create-cluster.html)

What do you think @rdblue?

rdblue: I think this is slightly different. Dremio, EMR, and Athena are all compute engines, but Cloudera isn't. It would make more sense to me to link to Impala and Hive docs than to link directly to Cloudera's platform (good to hear those are coming). That's why I would suggest having a Vendor section if we want to add Cloudera.

marton-bod: Thanks Ryan! I think having a Vendor section would make sense. Is there any way I can help with that?

samredai: @marton-bod Here's the worksheet for the proposal from the community sync this week for the new docs site flow. You can find the diagram in there so feel free to add some comments/suggestions on adding a vendor section and where it would be fit into the logical flow of the site. 👍

marton-bod: Thanks @samredai for the proposal, it looks awesome! 
One idea I had is to have the top-level entrypoint called "Integrations" instead of "Compute frameworks", and within "Integrations" we could have too general subsections: 
1) Compute engine integrations - Spark, Flink, Trino, Presto, Athena, etc.
2) Vendor platform integrations - Cloudera, etc.

What do you think?

samredai: I see the distinction and I considered calling that section "Integrations" initially. As I think about the way someone would navigate here though, I don't think they would go to "Integrations" and then decide if they're looking for a computer engine or a vendor platform, that feels like an extra unnecessary step and not really intuitive.

I would lean more towards having a top level "Vendors" section that takes you to a single page with a collection of external links (maybe hyperlinked cards with logos).

marton-bod: Yes, I think that makes sense. Having a Vendor top-level item is likely more intuitive for users

marton-bod: How do you think should we move forward with this PR given the new layout? Should we merge this and then it will get reorganised under the new Vendor section from here?

samredai: I'm making some style changes to the landing-page them such as [this](https://github.com/apache/iceberg-docs/pull/73) PR, and I'll follow-up with a PR that includes a simple left navbar. Then we can add non-versioned pages incrementally there and a vendor page would actually be a good first candidate since such a page doesn't exist yet.

pvary: @samredai: How could we help to move forward with the Vendor page? I would be happy to help any possible way.
Tried to ask write permission for the layout spreadsheet to add the page to the list, but no answer yet.

Thanks,
Peter

pvary: @samredai: Ryan merged the Cloudera icon - could we move forward with the Vendor page as well? How could I help?

samredai: @pvary sorry I was combing through my notifications and realized I missed this one. We should be good to add the vendors page now that the new theme is in. We just need to add a `vendors.md` file to the landing-page site in the [common](https://github.com/apache/iceberg-docs/tree/main/landing-page/content/common) directory.

I'm thinking something simple like the graphql ["Services"](https://graphql.org/code/#services) section. where it's a hyper-linked title (with alternative links, i.e. GitHub) on the left and a short description on the right.

pvary: Thanks @samredai.
I am on a longer PTO, but maybe @szlta, or @lcspinter will be able to pick this up next week

PR #4538, opened 2022-04-11
Core: Add Serializer for TableMetadata to REST Catalogs Serializers

kbendick:
There's presently no serializer for `TableMetadata` in the set of serializers registered for the REST catalogs object mapper.

Given that we already have a `toJson` method that handles this without the need for any FileIO objects etc, this can be added in now.

Work on a deserializer that the REST catalog can use is on going.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/RESTSerializers.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTSerializers.java
@@ -33,6 +33,8 @@ import org.apache.iceberg.PartitionSpecParser;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.SortOrderParser;
+import org.apache.iceberg.TableMetadata;
+import org.apache.iceberg.TableMetadataParser;
 import org.apache.iceberg.UnboundPartitionSpec;
 import org.apache.iceberg.UnboundSortOrder;
 import org.apache.iceberg.catalog.Namespace;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/rest/RESTSerializers.java
```diff
@@ -61,10 +63,19 @@ public static void registerAll(ObjectMapper mapper) {
...
+        .addDeserializer(UnboundSortOrder.class, new UnboundSortOrderDeserializer())
+        .addSerializer(TableMetadata.class, new TableMetadataSerializer());
     mapper.registerModule(module);
   }
 
+  public static class TableMetadataSerializer extends JsonSerializer<TableMetadata> {
+    @Override
+    public void serialize(TableMetadata metadata, JsonGenerator gen, SerializerProvider serializers)
+        throws IOException {
+      gen.writeRawValue(TableMetadataParser.toJson(metadata));
```
rdblue, line 75: Can't this pass the generator into `TableMetadataParser` like the others do?

kbendick: Sure. That would be fine.

kbendick: It _does_ require making a new signature `public`, but given that it matches the rest I think that's fine.

kbendick:
This has been superseded by https://github.com/apache/iceberg/pull/4641

Issue #3941, opened 2022-01-21
[Feature Request] Support for change data capture


NOD507:
A change data feed option like the one on delta lake is very usefull for incremental syncs between tables or systems.

CDC is one feature thats not on open source delta lake or iceberg. I dont think theres a format that supports CDC on spark open source at this moment.

Hidden partitioning is currently a plus for iceberg vs delta lake, and CDC could be another one.

https://docs.databricks.com/delta/delta-change-data-feed.html

flyrain:
I'm working on a Iceberg CDC design doc. It will be like the CDF(change data feed) in the proprietary Delta. Will keep you posted. It'd be helpful if you can share your CDC use case.

SreeramGarlapati:
@flyrain - we [started this design doc, which scopes the problem down to the tables that are written with MoR](https://docs.google.com/document/d/1A2oXgzMOhldSdpxhVnuRNYKozAaTAn3zWfE1LIeFLVs/edit?usp=sharing), a while back - after a community sync up. pl. see if you wanna adapt it or even change the same doc.

flyrain:
Thanks for sharing, @SreeramGarlapati! That is a good start and I'm glad you guys have started working on it. Can we have a discuss when my design doc is ready?

mohitgargk:
+1

@flyrain We are looking into two possible approaches here.
1.  Materialised where, like DeltaLake Change Data Feed that is generated in the write path.
2. Change Data Feed is evaluated at the read path by consumers.

At this point of time, we are grokking and spiking on approach #2. Would love to know your early thoughts on this.

flyrain:
Hi @mohitgargk, we had multiple discussions internally and are working on a doc. Will share it later. Generally speaking, CDC on MOR is in a better position, we should target it first. CDC on COW is almost impossible without additional metadata column or extra logging.

mohitgargk:
@flyrain Agree on that. With CoW, without any row-identifier, it is not possible to establish link with the new state and old state of the records without ambiguities.

For the CDF record structure, we grokked a bit on the design and came up with [this](https://docs.google.com/document/d/14ZRQc1anvXN52HGoNpSbKy7Ua6_7IIkr6mlXe_gHrJA/edit?usp=sharing). 

Above proposal is based on our use case where we are interested in change of a certain column but also want the values of other columns. If the whole state is part of CDF record, the consumer can avoid query/lookup/join with the table to get rest of the state. 

[...]

PR #3287, opened 2021-10-13
Support pos-delete in vectorized read.

flyrain:
### Why do we need this PR?
As #3141 mentioned, vectorized reader does NOT support row-level delete currently. This PR adds the support for pos-delete. This implementation supports <del>both Spark 2 and</del> Spark 3.2.

### Tests added
I parameterized TestSparkReaderDeletes, so that all tests under class TestSparkReaderDeletes now support both vectorized read and non-vectorized read. The ones with eq-delete still fall back to non-vectorized read at this moment.

### Future work
I will add the equality-delete support in a follow-up PR.

### Benchmark
https://github.com/apache/iceberg/pull/3287#issuecomment-960433304

cc @rdblue, @aokolnychyi, @RussellSpitzer, @karuppayya, @szehon-ho @jackye1995

PR #4539, opened 2022-04-11
Change Data Capture(CDC)[Draft]

flyrain:
The draft PR for change data capture. It largely aligns with the MVP we discussed in the [design doc](https://docs.google.com/document/d/1bN6rdLNcYOHnT3xVBfB33BoiPO06aKBo56SZmuU9pnY/edit?usp=sharing) and mail list. 
1. To emit delete and insert CDC records only
2. Create a Spark action for CDC generation
3. Leverage the `_deleted` metadata column for both pos deletes and eq deletes. Both deleted(pos and eq) rows are in the same format.

It is a draft PR though. So there are limitations
1. For row-level deletes, it only support parquet vectorized read at this moment. 
2. Multiple optimization can be done, for example, meta column `_delete` pushdown.
3. Need to expand interface to support query by timestamp instead of snapshot ids.
5. Test cases need to be added.

Happy to take feedbacks and file the formal PRs.
cc @aokolnychyi @RussellSpitzer @szehon-ho @jackye1995 @kbendick @karuppayya @chenjunjiedada

### DIFF
--- a/api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
+++ b/api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
@@ -74,4 +74,11 @@ public interface ActionsProvider {
   default DeleteReachableFiles deleteReachableFiles(String metadataLocation) {
     throw new UnsupportedOperationException(this.getClass().getName() + " does not implement deleteReachableFiles");
   }
+
+  /**
+   * Instantiates an action to generate CDC records.
+   */
+  default Cdc generateCdcRecords(Table table) {
+    throw new UnsupportedOperationException(this.getClass().getName() + " does not implement generateCdcRecords");
+  }
 }
--- a/api/src/main/java/org/apache/iceberg/actions/Cdc.java
+++ b/api/src/main/java/org/apache/iceberg/actions/Cdc.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.actions;
+
+public interface Cdc extends Action<Cdc, Cdc.Result> {
+  /**
+   * Emit changed data set by a snapshot id.
+   *
+   * @param snapshotId id of the snapshot to generate changed data
+   * @return this for method chaining
+   */
+  Cdc useSnapshot(long snapshotId);
+
+  /**
+   * Emit changed data set by a range of snapshots
+   *
+   * @param fromSnapshotId id of the first snapshot
+   * @param toSnapshotId id of the last snapshot
+   * @return this for method chaining
+   */
+  Cdc between(long fromSnapshotId, long toSnapshotId);
+
+  /**
+   * The action result that contains a dataset of changed rows.
+   */
+  interface Result {
+    /**
+     * Returns CDC records.
+     */
+    Object cdcRecords();
+  }
+}
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
@@ -104,6 +104,10 @@ public class VectorHolder {
     return new ConstantVectorHolder(numRows, constantValue);
   }
 
+  public static <T> VectorHolder deleteMetaColumnHolder(int numRows) {
+    return new DeletedVectorHolder(numRows);
+  }
+
   public static VectorHolder dummyHolder(int numRows) {
     return new ConstantVectorHolder(numRows);
   }
@@ -143,6 +147,19 @@ public class VectorHolder {
   public static class PositionVectorHolder extends VectorHolder {
     public PositionVectorHolder(FieldVector vector, Type type, NullabilityHolder nulls) {
       super(vector, type, nulls);
+    }
+  }
+
+  public static class DeletedVectorHolder extends VectorHolder {
+    private final int numRows;
+
+    public DeletedVectorHolder(int numRows) {
+      this.numRows = numRows;
+    }
+
+    @Override
+    public int numValues() {
+      return this.numRows;
     }
   }
 
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
@@ -517,5 +517,32 @@ public class VectorizedArrowReader implements VectorizedReader<VectorHolder> {
     }
   }
 
+  /**
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns a dummy
+   * VectorHolder which indicates whether the row is deleted.
+   */
+  public static class DeletedVectorReader<T> extends VectorizedArrowReader {
+    public DeletedVectorReader() {
+    }
+
+    @Override
+    public VectorHolder read(VectorHolder reuse, int numValsToRead) {
+      return VectorHolder.deleteMetaColumnHolder(numValsToRead);
+    }
+
+    @Override
+    public void setRowGroupInfo(PageReadStore source, Map<ColumnPath, ColumnChunkMetaData> metadata, long rowPosition) {
+    }
+
+    @Override
+    public String toString() {
+      return "DeletedVectorReader";
+    }
+
+    @Override
+    public void setBatchSize(int batchSize) {
+    }
+  }
+
 }
 
--- a/core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
@@ -47,6 +47,10 @@ public class BaseFileScanTask implements FileScanTask {
     this.residuals = residuals;
   }
 
+  public BaseFileScanTask cloneWithoutDeletes() {
+    return new BaseFileScanTask(file, null, schemaString, specString, residuals);
+  }
+
   @Override
   public DataFile file() {
     return file;
--- a/core/src/main/java/org/apache/iceberg/ManifestGroup.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestGroup.java
@@ -41,7 +41,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.ParallelIterable;
 
-class ManifestGroup {
+public class ManifestGroup {
   private static final Types.StructType EMPTY_STRUCT = Types.StructType.of();
 
   private final FileIO io;
@@ -55,6 +55,8 @@ class ManifestGroup {
   private Expression partitionFilter;
   private boolean ignoreDeleted;
   private boolean ignoreExisting;
+  private boolean ignoreAdded;
+  private boolean onlyWithDeletes;
   private boolean ignoreResiduals;
   private List<String> columns;
   private boolean caseSensitive;
@@ -66,7 +68,7 @@ class ManifestGroup {
         Iterables.filter(manifests, manifest -> manifest.content() == ManifestContent.DELETES));
   }
 
-  ManifestGroup(FileIO io, Iterable<ManifestFile> dataManifests, Iterable<ManifestFile> deleteManifests) {
+  public ManifestGroup(FileIO io, Iterable<ManifestFile> dataManifests, Iterable<ManifestFile> deleteManifests) {
     this.io = io;
     this.dataManifests = Sets.newHashSet(dataManifests);
     this.deleteIndexBuilder = DeleteFileIndex.builderFor(io, deleteManifests);
@@ -75,6 +77,8 @@ class ManifestGroup {
     this.partitionFilter = Expressions.alwaysTrue();
     this.ignoreDeleted = false;
     this.ignoreExisting = false;
+    this.ignoreAdded = false;
+    this.onlyWithDeletes = false;
     this.ignoreResiduals = false;
     this.columns = ManifestReader.ALL_COLUMNS;
     this.caseSensitive = true;
@@ -82,13 +86,13 @@ class ManifestGroup {
     this.manifestEntryPredicate = e -> true;
   }
 
-  ManifestGroup specsById(Map<Integer, PartitionSpec> newSpecsById) {
+  public ManifestGroup specsById(Map<Integer, PartitionSpec> newSpecsById) {
     this.specsById = newSpecsById;
     deleteIndexBuilder.specsById(newSpecsById);
     return this;
   }
 
-  ManifestGroup filterData(Expression newDataFilter) {
+  public ManifestGroup filterData(Expression newDataFilter) {
     this.dataFilter = Expressions.and(dataFilter, newDataFilter);
     deleteIndexBuilder.filterData(newDataFilter);
     return this;
@@ -122,6 +126,16 @@ class ManifestGroup {
 
   ManifestGroup ignoreExisting() {
     this.ignoreExisting = true;
+    return this;
+  }
+
+  public ManifestGroup ignoreAdded() {
+    this.ignoreAdded = true;
+    return this;
+  }
+
+  public ManifestGroup onlyWithDeletes() {
+    this.onlyWithDeletes = true;
     return this;
   }
 
@@ -180,7 +194,7 @@ class ManifestGroup {
         return CloseableIterable.transform(entries, e -> new BaseFileScanTask(
             e.file().copy(), deleteFiles.forEntry(e), schemaString, specString, residuals));
       }
-    });
+    }, deleteFiles);
 
     if (executorService != null) {
       return new ParallelIterable<>(tasks, executorService);
@@ -198,11 +212,13 @@ class ManifestGroup {
    * @return a CloseableIterable of manifest entries.
    */
   public CloseableIterable<ManifestEntry<DataFile>> entries() {
-    return CloseableIterable.concat(entries((manifest, entries) -> entries));
+    return CloseableIterable.concat(entries((manifest, entries) -> entries, null));
   }
 
+  @SuppressWarnings({"unchecked", "checkstyle:CyclomaticComplexity"})
   private <T> Iterable<CloseableIterable<T>> entries(
-      BiFunction<ManifestFile, CloseableIterable<ManifestEntry<DataFile>>, CloseableIterable<T>> entryFn) {
+      BiFunction<ManifestFile, CloseableIterable<ManifestEntry<DataFile>>, CloseableIterable<T>> entryFn,
+      DeleteFileIndex deleteFiles) {
     LoadingCache<Integer, ManifestEvaluator> evalCache = specsById == null ?
         null : Caffeine.newBuilder().build(specId -> {
           PartitionSpec spec = specsById.get(specId);
@@ -237,6 +253,12 @@ class ManifestGroup {
           manifest -> manifest.hasAddedFiles() || manifest.hasDeletedFiles());
     }
 
+    if (ignoreAdded) {
+      // only scan manifests that have entries other than added
+      matchingManifests = Iterables.filter(matchingManifests,
+          manifest -> manifest.hasExistingFiles() || manifest.hasDeletedFiles());
+    }
+
     matchingManifests = Iterables.filter(matchingManifests, manifestPredicate::test);
 
     return Iterables.transform(
@@ -258,6 +280,16 @@ class ManifestGroup {
                 entry -> entry.status() != ManifestEntry.Status.EXISTING);
           }
 
+          if (ignoreAdded) {
+            entries = CloseableIterable.filter(entries,
+                entry -> entry.status() != ManifestEntry.Status.ADDED);
+          }
+
+          if (onlyWithDeletes) {
+            entries = CloseableIterable.filter(entries,
+                entry -> hasDeletes(entry, deleteFiles));
+          }
+
           if (evaluator != null) {
             entries = CloseableIterable.filter(entries,
                 entry -> evaluator.eval((GenericDataFile) entry.file()));
@@ -266,5 +298,10 @@ class ManifestGroup {
           entries = CloseableIterable.filter(entries, manifestEntryPredicate);
           return entryFn.apply(manifest, entries);
         });
+  }
+
+  private boolean hasDeletes(ManifestEntry<DataFile> entry, DeleteFileIndex deleteFiles) {
+    DeleteFile[] deleteFileArray = deleteFiles.forEntry(entry);
+    return deleteFileArray.length > 0;
   }
 }
--- a/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
+++ b/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
@@ -290,8 +290,6 @@ public abstract class DeleteFilter<T> {
       requiredIds.addAll(eqDelete.equalityFieldIds());
     }
 
-    requiredIds.add(MetadataColumns.IS_DELETED.fieldId());
-
     Set<Integer> missingIds = Sets.newLinkedHashSet(
         Sets.difference(requiredIds, TypeUtil.getProjectedIds(requestedSchema)));
 
--- a/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCdcWithMerge.java
+++ b/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCdcWithMerge.java
@@ -1,0 +1,160 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.extensions;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.TableProperties;
+import org.apache.iceberg.actions.ActionsProvider;
+import org.apache.iceberg.actions.Cdc;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.spark.Spark3Util;
+import org.apache.iceberg.spark.SparkCatalog;
+import org.apache.iceberg.spark.actions.SparkActions;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException;
+import org.apache.spark.sql.catalyst.parser.ParseException;
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runners.Parameterized;
+
+import static org.apache.iceberg.TableProperties.WRITE_DISTRIBUTION_MODE_NONE;
+import static org.apache.iceberg.spark.actions.BaseCdcSparkAction.RECORD_TYPE;
+
+public class TestCdcWithMerge extends SparkRowLevelOperationsTestBase {
+  public TestCdcWithMerge(String catalogName, String implementation, Map<String, String> config, String fileFormat,
+                          boolean vectorized, String distributionMode) {
+    super(catalogName, implementation, config, fileFormat, vectorized, distributionMode);
+  }
+
+  /**
+   * Overwrite the parameters in the parent class. We need different combinations.
+   */
+  @Parameterized.Parameters(name = "catalogName = {0}, implementation = {1}, config = {2}," +
+      " format = {3}, vectorized = {4}, distributionMode = {5}")
+  public static Object[][] parameters() {
+    return new Object[][] {
+        { "testhive", SparkCatalog.class.getName(),
+            ImmutableMap.of(
+                "type", "hive",
+                "default-namespace", "default"
+            ),
+            "parquet",
+            true,
+            WRITE_DISTRIBUTION_MODE_NONE
+        }
+    };
+  }
+
+  @Override
+  protected Map<String, String> extraTableProperties() {
+    return ImmutableMap.of(TableProperties.MERGE_MODE, "copy-on-write");
+  }
+
+  @BeforeClass
+  public static void setupSparkConf() {
+    spark.conf().set("spark.sql.shuffle.partitions", "4");
+  }
+
+  @After
+  public void removeTables() {
+    sql("DROP TABLE IF EXISTS %s", tableName);
+    sql("DROP TABLE IF EXISTS source");
+  }
+
+  private ActionsProvider actions() {
+    return SparkActions.get();
+  }
+
+  @Test
+  public void testMergeWithOnlyUpdateClause() throws ParseException, NoSuchTableException {
+    createAndInitTable("id INT, dep STRING",
+        "{ \"id\": 1, \"dep\": \"emp-id-one\" }\n" +
+            "{ \"id\": 6, \"dep\": \"emp-id-six\" }");
+
+    createOrReplaceView("source", "id INT, dep STRING",
+        "{ \"id\": 2, \"dep\": \"emp-id-2\" }\n" +
+            "{ \"id\": 1, \"dep\": \"emp-id-1\" }\n" +
+            "{ \"id\": 6, \"dep\": \"emp-id-6\" }");
+
+    sql("MERGE INTO %s AS t USING source AS s " +
+        "ON t.id == s.id " +
+        "WHEN MATCHED AND t.id = 1 THEN " +
+        "  UPDATE SET *", tableName);
+
+    Table testTable = Spark3Util.loadIcebergTable(spark, tableName);
+    Cdc.Result result = actions().generateCdcRecords(testTable).execute();
+    Dataset<Row> resultDF = (Dataset<Row>) result.cdcRecords();
+
+    // verify results
+    List<Object[]> actualRecords = rowsToJava(resultDF.sort(RECORD_TYPE, "id").collectAsList());
+    long snapshotId = testTable.currentSnapshot().snapshotId();
+    long snapshotTs = testTable.currentSnapshot().timestampMillis();
+    ImmutableList<Object[]> expectedRows = ImmutableList.of(
+        row(1, "emp-id-one", "D", snapshotId, snapshotTs, 0),
+        row(6, "emp-id-six", "D", snapshotId, snapshotTs, 0),
+        row(1, "emp-id-1", "I", snapshotId, snapshotTs, 0),
+        row(6, "emp-id-six", "I", snapshotId, snapshotTs, 0)
+    );
+    assertEquals("Should have expected rows", expectedRows, actualRecords);
+  }
+
+  @Test
+  public void testMergeWithPositionDelete() throws ParseException, NoSuchTableException {
+    createAndInitTable("id INT, dep STRING",
+        "{ \"id\": 1, \"dep\": \"emp-id-one\" }\n" +
+            "{ \"id\": 6, \"dep\": \"emp-id-six\" }");
+
+    withV2MergeOnRead();
+
+    createOrReplaceView("source", "id INT, dep STRING",
+        "{ \"id\": 2, \"dep\": \"emp-id-2\" }\n" +
+            "{ \"id\": 1, \"dep\": \"emp-id-1\" }\n" +
+            "{ \"id\": 6, \"dep\": \"emp-id-6\" }");
+
+    sql("MERGE INTO %s AS t USING source AS s " +
+        "ON t.id == s.id " +
+        "WHEN MATCHED AND t.id = 1 THEN " +
+        "  UPDATE SET *", tableName);
+
+    Table testTable = Spark3Util.loadIcebergTable(spark, tableName);
+    Cdc.Result result = actions().generateCdcRecords(testTable).execute();
+    Dataset<Row> resultDF = (Dataset<Row>) result.cdcRecords();
+
+    // verify results
+    List<Object[]> actualRecords = rowsToJava(resultDF.sort(RECORD_TYPE, "id").collectAsList());
+    long snapshotId = testTable.currentSnapshot().snapshotId();
+    long snapshotTs = testTable.currentSnapshot().timestampMillis();
+    ImmutableList<Object[]> expectedRows = ImmutableList.of(
+        row(1, "emp-id-one", "D", snapshotId, snapshotTs, 0),
+        row(1, "emp-id-1", "I", snapshotId, snapshotTs, 0)
+    );
+    assertEquals("Should have expected rows", expectedRows, actualRecords);
+  }
+
+  private void withV2MergeOnRead() {
+    sql("ALTER TABLE %s SET TBLPROPERTIES('%s' '%s')", tableName, TableProperties.FORMAT_VERSION, "2");
+    sql("ALTER TABLE %s SET TBLPROPERTIES('%s' '%s')", tableName, TableProperties.MERGE_MODE, "merge-on-read");
+  }
+}
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
@@ -1,0 +1,262 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.actions;
+
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.apache.iceberg.BaseFileScanTask;
+import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DataOperations;
+import org.apache.iceberg.FileContent;
+import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.HasTableOperations;
+import org.apache.iceberg.ManifestFile;
+import org.apache.iceberg.ManifestGroup;
+import org.apache.iceberg.MetadataColumns;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.actions.Cdc;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.spark.FileScanTaskSetManager;
+import org.apache.iceberg.spark.SparkReadOptions;
+import org.apache.iceberg.util.SnapshotUtil;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.functions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.spark.sql.functions.lit;
+
+public class BaseCdcSparkAction extends BaseSparkAction<Cdc, Cdc.Result> implements Cdc {
+  private static final Logger LOG = LoggerFactory.getLogger(BaseCdcSparkAction.class);
+  public static final String RECORD_TYPE = "_record_type";
+  public static final String COMMIT_SNAPSHOT_ID = "_commit_snapshot_id";
+  public static final String COMMIT_TIMESTAMP = "_commit_timestamp";
+  public static final String COMMIT_ORDER = "_commit_order";
+  private final List<Long> snapshotIds = Lists.newLinkedList();
+  private final Table table;
+  private final List<Dataset<Row>> dfs = Lists.newLinkedList();
+  private boolean ignoreRowsDeletedWithinSnapshot = true;
+
+  private Expression filter = Expressions.alwaysTrue();
+
+  protected BaseCdcSparkAction(SparkSession spark, Table table) {
+    super(spark);
+    this.table = table;
+  }
+
+  @Override
+  public Result execute() {
+    // use the current snapshot by default
+    if (snapshotIds.isEmpty()) {
+      if (table.currentSnapshot() != null) {
+        snapshotIds.add(table.currentSnapshot().snapshotId());
+      }
+    }
+
+    for(int i = 0; i < snapshotIds.size(); i++) {
+      generateCdcRecordsPerSnapshot(snapshotIds.get(i), i);
+    }
+
+    Dataset<Row> outputDf = null;
+    for (Dataset<Row> df : dfs) {
+      if (outputDf == null) {
+        outputDf = df;
+      } else {
+        outputDf = outputDf.unionByName(df, true);
+      }
+    }
+    return new BaseCdcSparkActionResult(outputDf);
+  }
+
+  private void generateCdcRecordsPerSnapshot(long snapshotId, int commitOrder) {
+    Snapshot snapshot = table.snapshot(snapshotId);
+    if (snapshot.operation().equals(DataOperations.REPLACE)) {
+      return;
+    }
+
+    // new data file as the insert
+    Dataset<Row> df = readAppendDataFiles(snapshotId, commitOrder);
+    if (df != null) {
+      dfs.add(df);
+    }
+
+    // metadata deleted data files
+    Dataset<Row> deletedDf = readMetadataDeletedFiles(snapshotId, commitOrder);
+    if (deletedDf != null) {
+      dfs.add(deletedDf);
+    }
+
+    // pos and eq deletes
+    Dataset<Row> rowLevelDeleteDf = readRowLevelDeletes(snapshotId, commitOrder);
+    if (rowLevelDeleteDf != null) {
+      dfs.add(rowLevelDeleteDf);
+    }
+  }
+
+  private Dataset<Row> readAppendDataFiles(long snapshotId, int commitOrder) {
+    List<FileScanTask> appendedFileTasks = planAppendedFiles(snapshotId);
+    if (appendedFileTasks.size() <= 0) {
+      return null;
+    }
+
+    String groupID = UUID.randomUUID().toString();
+    FileScanTaskSetManager manager = FileScanTaskSetManager.get();
+    manager.stageTasks(table, groupID, appendedFileTasks);
+    Dataset<Row> scanDF = spark().read().format("iceberg")
+        .option(SparkReadOptions.FILE_SCAN_TASK_SET_ID, groupID)
+        .load(table.name());
+
+    return withCdcColumns(scanDF, snapshotId, "I", commitOrder);
+  }
+
+  private List<FileScanTask> planAppendedFiles(long snapshotId) {
+    CloseableIterable<FileScanTask> fileScanTasks = table.newScan()
+        .useSnapshot(snapshotId)
+        .filter(filter)
+        .ignoreResiduals()
+        .planFiles();
+
+    Set<CharSequence> dataFiles = Sets.newHashSet();
+    for (DataFile dataFile : table.snapshot(snapshotId).addedFiles()) {
+      dataFiles.add(dataFile.path());
+    }
+
+    List<FileScanTask> appendedFiles = Lists.newLinkedList();
+    if (dataFiles.size() <= 0) {
+      return appendedFiles;
+    }
+
+    fileScanTasks.forEach(fileScanTask -> {
+      if (fileScanTask.file().content().equals(FileContent.DATA) && dataFiles.contains(fileScanTask.file().path())) {
+        FileScanTask newFileScanTask = fileScanTask;
+        if (!ignoreRowsDeletedWithinSnapshot) {
+          // remove delete files so that no delete will apply to the data file
+          Preconditions.checkArgument(fileScanTask instanceof BaseFileScanTask,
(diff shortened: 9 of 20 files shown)


api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
```diff
@@ -74,4 +74,11 @@ default ExpireSnapshots expireSnapshots(Table table) {
   default DeleteReachableFiles deleteReachableFiles(String metadataLocation) {
     throw new UnsupportedOperationException(this.getClass().getName() + " does not implement deleteReachableFiles");
   }
+
+  /**
+   * Instantiates an action to generate CDC records.
+   */
+  default Cdc generateCdcRecords(Table table) {
```
kbendick, line 81: Nit: Consider using `CDC` as all caps instead, like it is in the javadoc comments. For me, it looks a lot cleaner.

flyrain: I'm also open to other names, e.g., `ChangeDataSet`, `ChangeDataCapture`

kbendick: In Flink it's referred to as `Changelog`.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+      snapshotIds.add(snapshotId);
+    }
+    return this;
+  }
+
+  @Override
+  public Cdc between(long fromSnapshotId, long toSnapshotId) {
+    Preconditions.checkArgument(table.snapshot(fromSnapshotId) != null,
+        "The fromSnapshotId(%s) is invalid", fromSnapshotId);
+    Preconditions.checkArgument(table.snapshot(toSnapshotId) != null,
```
rajarshisarkar, line 250: We should also check if `fromSnapshotId` is an ancestor of `toSnapshotId`:

```
Preconditions.checkArgument(SnapshotUtil.isAncestorOf(table, toSnapshotId, fromSnapshotId),
        "from snapshot %s is not an ancestor of to snapshot %s", fromSnapshotId, toSnapshotId);
```

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
```diff
@@ -517,5 +517,32 @@ public void setBatchSize(int batchSize) {
     }
   }
 
+  /**
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns a dummy
+   * VectorHolder which indicates whether the row is deleted.
+   */
+  public static class DeletedVectorReader<T> extends VectorizedArrowReader {
```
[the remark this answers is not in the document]
flyrain: We don't need it. Will remove it.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+    // use the current snapshot by default
+    if (snapshotIds.isEmpty()) {
+      if (table.currentSnapshot() != null) {
+        snapshotIds.add(table.currentSnapshot().snapshotId());
+      }
+    }
+
+    for (int i = 0; i < snapshotIds.size(); i++) {
+      generateCdcRecordsPerSnapshot(snapshotIds.get(i), i);
+    }
```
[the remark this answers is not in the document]
singhpk234: nvm df's would be computed lazily, this wouldn't be required then.

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
kbendick: Yeah I agree that using a full word would be better.

In comments and even method names it's fine in my opinion but as the main class name it probably would be best to use the full name.

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
```diff
@@ -47,6 +47,10 @@ public BaseFileScanTask(DataFile file, DeleteFile[] deletes, String schemaString
     this.residuals = residuals;
   }
 
+  public BaseFileScanTask cloneWithoutDeletes() {
+    return new BaseFileScanTask(file, null, schemaString, specString, residuals);
```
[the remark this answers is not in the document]
flyrain: Nice suggestion. I don't like to pass `null` either.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+        .withColumn(COMMIT_ORDER, lit(commitOrder));
+  }
+
+  @Override
+  public Cdc useSnapshot(long snapshotId) {
+    if (table.snapshot(snapshotId) != null) {
+      snapshotIds.add(snapshotId);
+    }
+    return this;
+  }
```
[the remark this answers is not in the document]
flyrain: Yup, will clear the list before add anything.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+    return ImmutableList.copyOf(manifestGroup.planFiles());
+  }
+
+  private Dataset<Row> readRowLevelDeletes(long snapshotId, int commitOrder) {
+    List<FileScanTask> fileScanTasks = planRowLevelDeleteFiles(snapshotId);
+    if (fileScanTasks.size() <= 0) {
+      return null;
+    }
+
+    String groupID = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
flyrain: I assume this is for easier debugging. We probably don't do that since this uuid is used totally internally, it won't be in any log and UI.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+  public Result execute() {
+    // use the current snapshot by default
+    if (snapshotIds.isEmpty()) {
+      if (table.currentSnapshot() != null) {
+        snapshotIds.add(table.currentSnapshot().snapshotId());
+      }
+    }
+
+    for (int i = 0; i < snapshotIds.size(); i++) {
+      generateCdcRecordsPerSnapshot(snapshotIds.get(i), i);
```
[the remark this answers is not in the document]
flyrain: Yep, it is a part of the design. I simplified it in the draft. Will add it later. To be clear, v1 doesn't have seq number, we will give a zero-based order.

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
stevenzwu: +1 for `Changelog`. Here, it means `generateChangelog`

api/src/main/java/org/apache/iceberg/actions/Cdc.java
```diff
@@ -0,0 +1,49 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.actions;
+
+public interface Cdc extends Action<Cdc, Cdc.Result> {
```
stevenzwu, line 22: Is this an action? Action typically modify the table like rewrite, expire snapshots.

flyrain: It is. Action usually modifies table, but is not necessarily limited by that. This is the first PR, we will also explore a way to use scan for CDC.

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
```diff
@@ -146,4 +150,17 @@ public PositionVectorHolder(FieldVector vector, Type type, NullabilityHolder nul
     }
   }
 
+  public static class DeletedVectorHolder extends VectorHolder {
```
stevenzwu, line 153: nit: a little weird to me to have an extended class nested inside

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -41,7 +41,7 @@
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.ParallelIterable;
 
-class ManifestGroup {
+public class ManifestGroup {
```
stevenzwu, line 44: If we use scan API, we may not need to expose ManifestGroup as public

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
flyrain: I guess it's fine, since we've got nested classes `PositionVectorHolder` and `ConstantVectorHolder` extend `VectorHolder`.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
flyrain: It'd be awesome. We have to design the scan API nicely though.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+    if (deletedDf != null) {
+      dfs.add(deletedDf);
+    }
+
+    // pos and eq deletes
+    Dataset<Row> rowLevelDeleteDf = readRowLevelDeletes(snapshotId, commitOrder);
+    if (rowLevelDeleteDf != null) {
+      dfs.add(rowLevelDeleteDf);
+    }
+  }
```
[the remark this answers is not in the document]
flyrain: I can definitely make the change. Although, I don't think it makes any difference. Basically, these cdc records(e.g. `1, aa, D`, `1, bb, I` ) **within a snapshot** doesn't have any order. Please check the discussion in the [design doc](https://docs.google.com/document/d/1bN6rdLNcYOHnT3xVBfB33BoiPO06aKBo56SZmuU9pnY/edit#heading=h.czd6l3c9xj87). We should treat them same while ingesting them into downstream. Here, we cannot delete the row (`1, aa`) by just checking its id.

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
stevenzwu: agree. it is existing pattern and we don't need to change. regardless, I just thought it a little weird

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -73,6 +74,16 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
...
   }
 
+  public static <T> CloseableIterable<T> marker(CloseableIterable<T> rows, Function<T, Long> rowToPosition,
+                                                PositionDeleteIndex deleteSet, Consumer<T> markDeleted) {
+    if (deleteSet.isEmpty()) {
+      return rows;
+    }
+
+    PositionSetDeleteMarker<T> deleteMarker = new PositionSetDeleteMarker<>(rowToPosition, deleteSet, markDeleted);
+    return deleteMarker.filter(rows);
```
stevenzwu, line 84: since `PositionSetDeleteMarker` always return true, we are not actually doing filter. it seems that we are mainly leverage the filter to traverse the iterable and get the side effect of calling `markDeleted` with matched rows. The semantics is a little odd to me. Maybe we don't need to introduce `PositionSetDeleteMarker` filter?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -94,6 +96,22 @@ protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSc
...
+  }
+
+  private int getIndexOfIsDeletedMetadataColumn() {
+    List<Types.NestedField> icebergFields = requiredSchema.asStruct().fields();
+    for (int i = 0; i < icebergFields.size(); i++) {
+      if (icebergFields.get(i).fieldId() == MetadataColumns.IS_DELETED.fieldId()) {
+        return i;
+      }
+    }
+    return -1;
```
stevenzwu, line 110: nit: Iceberg coding style has a line break after a control block

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -185,14 +203,35 @@ private CloseableIterable<T> applyEqDeletes(CloseableIterable<T> records) {
...
+        if (!remainingRows.test(item)) {
+          // if the row is deleted, set the metadata column to true
+          markRowDeleted(item);
+        }
+        return true;  // keep the row even if it is deleted by equality deletions
+      }
+    };
+  }
+
+  protected void markRowDeleted(T item) {
```
stevenzwu, line 233: should this just be an abstract method and force impl classes to implement this method?

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
stevenzwu: Maybe just use this method from `CloseableIterable` to traverse? it returns the same object after applying the `Consumer` function.

```
  static <I, O> CloseableIterable<O> transform(CloseableIterable<I> iterable, Function<I, O> transform) {
```

api/src/main/java/org/apache/iceberg/actions/Cdc.java
stevenzwu: All the existing actions in Iceberg seem to modify the table. Since this is a public API, I like to double check. Can we keep it in SparkActions only for now if we are in experimental phase?

Conceptually, this is a scan/read (not a maintenance action).

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkActionResult.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class BaseCdcSparkActionResult implements Cdc.Result {
+  private final Dataset<Row> cdcRecords;
+
+  public BaseCdcSparkActionResult(Dataset<Row> cdcRecords) {
+    this.cdcRecords = cdcRecords;
+  }
+
+  @Override
+  public Object cdcRecords() {
```
stevenzwu, line 34: maybe make `Cdc.Result` a generic class to get type info?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+
+  private Dataset<Row> withCdcColumns(Dataset<Row> df, long snapshotId, String cdcType, int commitOrder) {
+    return df.withColumn(RECORD_TYPE, lit(cdcType))
+        .withColumn(COMMIT_SNAPSHOT_ID, lit(snapshotId))
+        .withColumn(COMMIT_TIMESTAMP, lit(table.snapshot(snapshotId).timestampMillis()))
+        .withColumn(COMMIT_ORDER, lit(commitOrder));
+  }
+
+  @Override
+  public Cdc useSnapshot(long snapshotId) {
```
stevenzwu, line 239: `useSnapshot` is typically meant for time travel. I feel `ofSnapshot` might be a more accurate name. generate changelog dataset of a specific snapshot

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+  public Cdc between(long fromSnapshotId, long toSnapshotId) {
+    Preconditions.checkArgument(table.snapshot(fromSnapshotId) != null,
+        "The fromSnapshotId(%s) is invalid", fromSnapshotId);
+    Preconditions.checkArgument(table.snapshot(toSnapshotId) != null,
+        "The toSnapshotId(%s) is invalid", toSnapshotId);
+    Preconditions.checkArgument(SnapshotUtil.isAncestorOf(table, toSnapshotId, fromSnapshotId),
+        "The fromSnapshot(%s) is not an ancestor of the toSnapshot(%s)", fromSnapshotId, toSnapshotId);
+
+    snapshotIds.clear();
+    // include the fromSnapshotId
```
stevenzwu, line 257: exclusive behavior for `fromSnapshotId` might be easier to work with. Assuming last enumerated position (snapshotId) needs to be saved somewhere. Next run would scan the range of `(fromSnapshotId, toSnapshotId]`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+  protected BaseCdcSparkAction(SparkSession spark, Table table) {
+    super(spark);
+    this.table = table;
+  }
+
+  @Override
+  public Result execute() {
+    // use the current snapshot by default
+    if (snapshotIds.isEmpty()) {
+      if (table.currentSnapshot() != null) {
```
stevenzwu, line 79: I am not sure if we want to default to current snapshot. For incremental/CDC read, there is a need to bookkeeping/checkpointing the last position so that the next scan/run would know where to  resume from. Silently pick a location would violate the bookkeeping requirement.

Simply returning empty result / no action seems reasonable to me.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+      if (outputDf == null) {
+        outputDf = df;
+      } else {
+        outputDf = outputDf.unionByName(df, true);
+      }
+    }
+    return new BaseCdcSparkActionResult(outputDf);
+  }
+
+  private void generateCdcRecordsPerSnapshot(long snapshotId, int commitOrder) {
```
stevenzwu, line 99: nit: maybe return the Datasets from this method (instead of updating a class variable). then outside can do the union. E.g. use the reduce method from Java stream API.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+      if (table.currentSnapshot() != null) {
+        snapshotIds.add(table.currentSnapshot().snapshotId());
+      }
+    }
+
+    for (int i = 0; i < snapshotIds.size(); i++) {
+      generateCdcRecordsPerSnapshot(snapshotIds.get(i), i);
+    }
+
+    Dataset<Row> outputDf = null;
```
stevenzwu, line 88: can we use `unionAll`?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+  public Result execute() {
+    // use the current snapshot by default
+    if (snapshotIds.isEmpty()) {
+      if (table.currentSnapshot() != null) {
+        snapshotIds.add(table.currentSnapshot().snapshotId());
+      }
+    }
+
+    for (int i = 0; i < snapshotIds.size(); i++) {
+      generateCdcRecordsPerSnapshot(snapshotIds.get(i), i);
```
stevenzwu, line 85: We pass the index of the snapshot list as `commitOrder`. What is the intended usage of `commitOrder`?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.functions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.spark.sql.functions.lit;
+
+public class BaseCdcSparkAction extends BaseSparkAction<Cdc, Cdc.Result> implements Cdc {
```
stevenzwu, line 57: A high level question. since this algorithm doesn't guarantee ordering within a snapshot (for a good reason), does that mean we can only have parallel executors if it only scans one snapshot? If this scans multiple snapshots, parallel executors could mess up the ordering.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
chenjunjiedada: I think it traverses the rows and adds `is_deleted` value to the row.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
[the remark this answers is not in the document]
stevenzwu: I agree with @hameizi on read/emit eq deletes before the append, as eq deletes are applied to the previous snapshots. We should apply/emit eq deletes before the appended rows in the current snapshot.

I also left a comment regarding the pos deletes in issue #3941 . I am not sure we need to emit pos deletes, as they are applied as delete filter for the inserted rows in the current snapshot already. It is like they are squashed away already.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,262 @@
...
+
+    return withCdcColumns(scanDF, snapshotId, "D", commitOrder);
+  }
+
+  private List<FileScanTask> planMetadataDeletedFiles(long snapshotId) {
+    Snapshot snapshot = table.snapshot(snapshotId);
+    ManifestGroup manifestGroup = new ManifestGroup(table.io(), snapshot.dataManifests(), snapshot.deleteManifests())
+        .filterData(filter)
+        .ignoreAdded()
+        .specsById(((HasTableOperations) table).operations().current().specsById());
```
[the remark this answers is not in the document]
flyrain: `caseSensitive` not necessary here.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+    FileScanTaskSetManager manager = FileScanTaskSetManager.get();
+    manager.stageTasks(table, groupID, appendedFileTasks);
+    Dataset<Row> scanDF = spark().read().format("iceberg")
+        .option(SparkReadOptions.FILE_SCAN_TASK_SET_ID, groupID)
+        .load(table.name());
+
+    return withCdcColumns(scanDF, snapshotId, "I", commitOrder);
+  }
+
+  private List<FileScanTask> planAppendedFiles(long snapshotId) {
```
stevenzwu, line 140: This method is a little weird to me. 

- discover all files reachable by the `snapshotId` via time travel. 
- found all files added in this snapshot. 
- filter the files discovered during the first step with the appended data files discovered in second step

We can use `TableScan#appendsBetween(parentSnapshotId, snapshotId)#planFiles` to get the `FileScanTask` collection. Then we can apply the `ignoreRowsDeletedWithinSnapshot` transformation if needed.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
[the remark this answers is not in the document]
flyrain: Yes, we can plan them in parallel if the planning perf is an issue. It'd be especially usefully in case of a big table and/or a big number of snapshots.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+        appendedFiles.add(newFileScanTask);
+      }
+    });
+
+    return appendedFiles;
+  }
+
+  private Dataset<Row> readMetadataDeletedFiles(long snapshotId, int commitOrder) {
+    List<FileScanTask> appendedFileTasks = planMetadataDeletedFiles(snapshotId);
```
stevenzwu, line 174: appendedFileTasks -> metadataDeletedFileTasks

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
```diff
@@ -0,0 +1,267 @@
...
+        .filterData(filter)
+        .onlyWithDeletes()
+        .specsById(((HasTableOperations) table).operations().current().specsById());
+
+    return ImmutableList.copyOf(manifestGroup.planFiles());
+  }
+
+  private Dataset<Row> withCdcColumns(Dataset<Row> df, long snapshotId, String cdcType, int commitOrder) {
+    return df.withColumn(RECORD_TYPE, lit(cdcType))
+        .withColumn(COMMIT_SNAPSHOT_ID, lit(snapshotId))
```
stevenzwu, line 233: what are the intended usage of those 3 commit related metadata columns?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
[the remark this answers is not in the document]
flyrain: I'm OK with that, will make the change.

For pos deletes, there are two scenarios:
1. It deletes rows from the previous snapshots. We need to emit them as other deleted rows.
2. It deletes rows within the same snapshots. We can squash it by default. I use a flag `ignoreRowsDeletedWithinSnapshot` to control that in case people don't want to squash it.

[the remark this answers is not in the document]
stevenzwu: My understanding is that pos deletes are only applicable to files appended in the same snapshot and eq deletes are only applied to the previous snapshots.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -125,6 +129,16 @@ ManifestGroup ignoreExisting() {
     return this;
   }
 
+  public ManifestGroup ignoreAdded() {
+    this.ignoreAdded = true;
+    return this;
+  }
+
+  public ManifestGroup onlyWithDeletes() {
```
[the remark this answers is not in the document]
flyrain: No. `onlyWithDeletes` means we only get the data files with row-level deletes. Maybe I should rename it to `onlyWithRowLevelDeletes`.  `ignoreExisting` and `ignoreAdded` are used to filter the type of data files.

api/src/main/java/org/apache/iceberg/actions/Cdc.java
flyrain: Agree. This is the start point. As I said in another comment, we need a well designed scan interface first.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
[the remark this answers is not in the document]
stevenzwu: yeah, `onlyWithRowLevelDeletes` would make it very clear

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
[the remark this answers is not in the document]
flyrain: Spark does, and it doesn't use eq delete at all. In case of merge, delete, update, Spark uses pos deletes to remove rows from previous snapshots.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCdcWithMerge.java
```diff
@@ -0,0 +1,169 @@
...
+    sql("DROP TABLE IF EXISTS %s", tableName);
+    sql("DROP TABLE IF EXISTS source");
+  }
+
+  private ActionsProvider actions() {
+    return SparkActions.get();
+  }
+
+  @Test
+  public void testMergeWithOnlyUpdateClause() throws ParseException, NoSuchTableException {
```
flyrain, line 100: @stevenzwu, Spark use pos deletes in the merge command.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkFilesScan.java
```diff
@@ -50,6 +52,16 @@ class SparkFilesScan extends SparkScan {
...
   }
 
+  SparkFilesScan(SparkSession spark, Table table, SparkReadConf readConf, Schema expectedSchema,
+                 List<Expression> filterExpressions) {
+    super(spark, table, readConf, expectedSchema == null ? table.schema() : expectedSchema, ImmutableList.of());
+
+    this.taskSetID = readConf.fileScanTaskSetId();
+    this.splitSize = readConf.splitSize();
+    this.splitLookback = readConf.splitLookback();
+    this.splitOpenFileCost = readConf.splitOpenFileCost();
```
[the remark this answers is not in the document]
flyrain: I can remove the first constructor since it is not used.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: Probably, it needs changes for all subclass of DeleteFilter. This is PR is already big. Maybe we should split it. For example, we can have a separated PR to read deleted rows in case of row-level change. cc @chenjunjiedada

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: Make sense. One concern is that it cannot get the first snapshot of a table. Two solutions:
1. Add a new interface like `toSnapshot()`, which compute all snapshots to the `toSnapshot`. 
2. Add a overwrite method `between(long fromSnapshotId, long toSnapshotId, boolean includeFromSnapshotId)`
What do you think?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: Agree. It is for easier test at the beginning, let me make the change.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: We expected the schema of these datasets are the same. There are no difference between `unionAll` and `unionByName`.  BTW, `unionAll` only union one dataset, not a list of datasets,

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: We got the metadata column `commit order` to indicate the order of cdc records from multiple snapshots. The order won't be messed up.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: see https://github.com/apache/iceberg/pull/4539#discussion_r855673237

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
aokolnychyi: I am OK with `generateChangelog`.

api/src/main/java/org/apache/iceberg/actions/Cdc.java
aokolnychyi: I wouldn't mind having an action like this. We have `RemoveOrphanFiles` that does not modify the table state, for instance. I would match the naming style we have, though. I think it can be `GenerateChangelog` as all other actions start with a verb.

api/src/main/java/org/apache/iceberg/actions/Cdc.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Emit changed data set by a snapshot id.
+   *
+   * @param snapshotId id of the snapshot to generate changed data
+   * @return this for method chaining
+   */
+  Cdc ofSnapshot(long snapshotId);
```
aokolnychyi, line 29: If we decide to include a verb in the action name, I'd also consider renaming these methods like this:

```
// changes for a specific snapshot
GenerateChangelog forSnapshot(long snapshotId);

// changes starting from a particular snapshot (exclusive)
GenerateChangelog afterSnapshot(long fromSnapshotId);

// changes from start snapshot (exclusive) to end snapshot (inclusive)
GenerateChangelog betweenSnapshots(long fromSnapshotId, long toSnapshotId);
```

api/src/main/java/org/apache/iceberg/actions/Cdc.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Emit changed data set by the current snapshot.
+   *
+   * @return this for method chaining
+   */
+  Cdc ofCurrentSnapshot();
```
aokolnychyi, line 36: In which case will this method be useful? I guess consumers will probably always want to consume from a particular point in time so I am not sure CDC records for a current snapshot would be very useful.

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
stevenzwu: or `generateChangeSet` for batch execution. if it is a long-running streaming execution, `changelog` would be more accurate as it implies a stream.

api/src/main/java/org/apache/iceberg/actions/Cdc.java
stevenzwu: +1. for incremental read, we need to bookkeep the position. implicit current snapshot would make the bookkeeping impossible and hence is probably not a valid use case

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
```diff
@@ -517,5 +517,32 @@ public void setBatchSize(int batchSize) {
     }
   }
 
+  /**
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns a dummy
+   * VectorHolder which indicates whether the row is deleted.
+   */
+  public static class DeletedVectorReader extends VectorizedArrowReader {
```
chenjunjiedada, line 524: A basic question, do we need to enable vectorization to read CDC? What about the non-vectorized reader?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
chenjunjiedada: The previous PR which read the deleted rows has such implementation, +1 to use another PR.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
stevenzwu: does that mean we can only have parallel executors if it only scans one snapshot? If this scans multiple snapshots, parallel executors could mess up the ordering.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
stevenzwu: what do you mean "get the first snapshot of a table"?

> Add a new interface like toSnapshot(), which compute all snapshots to the toSnapshot.

This is covered if `fromSnapshotId` is not set, which means null is used. Then `IncrementalScan` would process all snapshots up to the `toSnapshotId`. As if we trace the ancestor chain of `toSnapshotId`, we will eventually trace to `null` snapshotId and stop. 

>Add a overwrite method between(long fromSnapshotId, long toSnapshotId, boolean includeFromSnapshotId)

This API is not not necessary. we can just set the `fromSnapshotId` to the parent snapshot id if we want the inclusive behavior.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCdcWithMerge.java
stevenzwu: Are you say8ing Spark merge into writes pos deletes? if so, it doesn't answer my question why do we need to emit pos deletes, as pos deletes are already applied as delete filters when emit insert records for the same snapshot.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,10 +266,14 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
 
     // if there are fewer deletes than a reasonable number to keep in memory, use a set
     if (posDeletes.stream().mapToLong(DeleteFile::recordCount).sum() < setFilterThreshold) {
-      return Deletes.filter(records, this::pos, Deletes.toPositionIndex(filePath, deletes));
+      return hasMetadataColumnIsDeleted ?
+          Deletes.marker(records, this::pos, Deletes.toPositionIndex(filePath, deletes), this::markRowDeleted) :
```
chenjunjiedada, line 270: I feel we should abstract these later. 
```
if (needWrapDeleteColumn) {
    Deletes.map(records, this::pos, Deletes.toPositionoIndex(filePath, deletes), this::markRowDeleted);
} else {
    Deletes.filter(..)
}
```

What about adding a class DeleteStream that supports `filter` and `map`.

chenjunjiedada: The `DeleteStream` contains the filter logics in this `DeleteFilter` also contains transforming logics.

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
```diff
@@ -70,8 +70,7 @@ public class TestSparkParquetReadMetadataColumns {
   private static final Schema PROJECTION_SCHEMA = new Schema(
       required(100, "id", Types.LongType.get()),
       required(101, "data", Types.StringType.get()),
-      MetadataColumns.ROW_POSITION,
-      MetadataColumns.IS_DELETED
```
chenjunjiedada, line 74: Why we change these unit tests?

api/src/main/java/org/apache/iceberg/actions/Cdc.java
stevenzwu: `RemoveOrphanFiles` is also a maintenance action on the table. Agree that technically it doesn't modify the table state. But it is a garbage collection action for the tables. Those orphaned files may be part of the table before but got unreferenced due to compaction or retention purge, or those orphaned files were intended to be added to the table but got aborted in the middle. To me, it is a modification of the broader env/infra of the table.

aokolnychyi: The primary goal of adding the `Action` API was to provide scalable and ready-to-use recipes for common scenarios. I am not sure we should strictly restrict the usage to maintenance.

That being said, I'd be happy to discuss alternative ways to expose this.

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
aokolnychyi: I liked `generateChangelog` but if it can confuse people, `generateChangeSet` sounds good too.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: `null` can solve the issue. However, I was trying to avoid it since it doesn't have a clear meaning to users. 
1. Users don't know they can input a null.
2. Users don't know what will happen when they input a null.

So users have to rely on the doc to figure that out. I'd suggest to provide the new interface `beforeSnapshot()` or `toSnapshot()`. It did the same thing under the hood, but more meaningful to users. Another benefit is that it'd be logically complete since @aokolnychyi also proposed the method `afterSnapshot()` in here, https://github.com/apache/iceberg/pull/4539#discussion_r855718325.

stevenzwu: yeah. @rdblue also made a similar suggestion: https://github.com/apache/iceberg/pull/4580#discussion_r856412067. We will go with `fromSnapshotInclusive` and `fromSnapshotExclusive` in the new increment scan interface

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
flyrain: Combining feedbacks, changed it to `GetChangeSet`. The name `GenerateChangeSet` is good, but it is way too long. Think about the class name `BaseGenerateChangeSetSparkActionResult`. I admit the verb `get` is plain comparing to `generate`. But I think it is fine, a plain name is suitable for a tool.

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
flyrain: Vectorized read is significant faster than non-vectorized read, check benchmark in #3287. We should have it. Non-vectorized read is handled by the change I made in class `RowDataReader`, `DeleteFilter` and `Deletes`.

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
flyrain: Some tests(e.g., `testReadRowNumbersWithDelete`) failed due to it uses the `PROJECTION_SCHEMA`, which contains IS_DELETED. So the row count was wrong since it still assume only non-deleted rows are there.  We can either remove the column `IS_DELETED` so that deleted rows won't be in the results. Or we can change the test case to filter out the deleted rows in the results. I choose the former so that it can fix multiple failure at once.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkActionResult.java
flyrain: Good idea, will make the change.

api/src/main/java/org/apache/iceberg/actions/GetChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+  GetChangeSet betweenSnapshots(long fromSnapshotId, long toSnapshotId);
+
+  /**
+   * The action result that contains a dataset of changed rows.
+   */
+  interface Result {
+    /**
+     * Returns the change set.
+     */
+    Object changeSet();
```
aokolnychyi, line 62: I think we may want to parameterize this and use `Dataset<Row>` in Spark instead of plain `Object`.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
aokolnychyi: I am not sure about extending `TableScan` (can be convinced) but we can definitely create a utility class in core that will internally use `ManifestGroup`. I don' think we should expose this class. Maybe, it is safer to start with a utility and then see if it is something that can be part of `TableScan`.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+
+public class BaseGetChangeSetSparkAction extends BaseSparkAction<GetChangeSet, GetChangeSet.Result>
+    implements GetChangeSet {
+  private static final Logger LOG = LoggerFactory.getLogger(BaseGetChangeSetSparkAction.class);
+  public static final String RECORD_TYPE = "_record_type";
+  public static final String COMMIT_SNAPSHOT_ID = "_commit_snapshot_id";
+  public static final String COMMIT_TIMESTAMP = "_commit_timestamp";
+  public static final String COMMIT_ORDER = "_commit_order";
+
+  private final List<Long> snapshotIds = Lists.newLinkedList();
```
aokolnychyi, line 65: While using a single snapshot ID list is convenient, the downside is that we can't validate whether there are conflicting calls to `betweenSnapshots`, `forSnapshot`, etc. In `SparkScanBuilder` and `BaseTableScan`, we took a different approach.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+  private final Table table;
+  private final List<Dataset<Row>> dfs = Lists.newLinkedList();
```
aokolnychyi, line 67: Does this have to be an instance var? I guess a local var would be more appropriate.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+
+  private Dataset<Row> readAppendDataFiles(long snapshotId, int commitOrder) {
+    List<FileScanTask> fileScanTasks = planAppendedFiles(snapshotId);
+    if (fileScanTasks.isEmpty()) {
+      return null;
+    }
+
+    String groupID = UUID.randomUUID().toString();
+    FileScanTaskSetManager manager = FileScanTaskSetManager.get();
+    manager.stageTasks(table, groupID, fileScanTasks);
```
aokolnychyi, line 127: We stage tasks but we never invalidate them. This will cause a memory leak. It will be tricky to invalidate them as we don't know whether it is already safe to do so. We can instruct the manager to remove tasks after the first access or make the action result closable. Let's think what will be best.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+    Dataset<Row> scanDF = spark().read().format("iceberg")
+        .option(SparkReadOptions.FILE_SCAN_TASK_SET_ID, groupID)
+        .load(table.name());
+
+    return withCdcColumns(scanDF, snapshotId, "I", commitOrder);
+  }
+
+  private List<FileScanTask> planAppendedFiles(long snapshotId) {
+    CloseableIterable<FileScanTask> fileScanTasks = table.newScan()
```
aokolnychyi, line 136: This is a full table scan. If you use a utility with `ManifestGroup`, you would be able to get a much better performance.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+
+    List<FileScanTask> appendedFiles = Lists.newLinkedList();
+    if (dataFiles.isEmpty()) {
+      return appendedFiles;
+    }
+
+    fileScanTasks.forEach(fileScanTask -> {
+      if (fileScanTask.file().content().equals(FileContent.DATA) && dataFiles.contains(fileScanTask.file().path())) {
+        FileScanTask newFileScanTask = fileScanTask;
+        if (!ignoreRowsDeletedWithinSnapshot) {
```
aokolnychyi, line 155: I am not sure this is correct. Records added and removed in the same snapshot must have a lesser commit order compared to all other records added in that snapshot. I'd not add this functionality to start with.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+    Dataset<Row> scanDF = spark().read().format("iceberg")
+        .option(SparkReadOptions.FILE_SCAN_TASK_SET_ID, groupID)
+        .load(table.name());
+
+    return withCdcColumns(scanDF, snapshotId, "D", commitOrder);
+  }
+
+  private List<FileScanTask> planMetadataDeletedFiles(long snapshotId) {
+    Snapshot snapshot = table.snapshot(snapshotId);
+    ManifestGroup manifestGroup = new ManifestGroup(table.io(), snapshot.dataManifests(), snapshot.deleteManifests())
```
aokolnychyi, line 186: Do we have to access all data manifests? Aren't we interested only in data manifests added in this snapshot? Do we also need delete manifests for this?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
```diff
@@ -0,0 +1,282 @@
...
+    Snapshot snapshot = table.snapshot(snapshotId);
+    List<ManifestFile> manifestFiles = snapshot.deleteManifests().stream()
+        .filter(manifestFile -> manifestFile.snapshotId().equals(snapshotId))
+        .collect(Collectors.toList());
+
+    // todo create a partition filter to filter out unrelated partitions
+    ManifestGroup manifestGroup = new ManifestGroup(table.io(), snapshot.dataManifests(), manifestFiles)
+        .filterData(filter)
+        .onlyWithRowLevelDeletes()
+        .specsById(((HasTableOperations) table).operations().current().specsById());
```
aokolnychyi, line 221: nit: why not use `table.specs()`?

aokolnychyi (commented):
I did a review of the action. I think it shows the algorithm we discussed in the issue can be implemented fairly easy. I'd say supporting `is_deleted` is actually more challenging.

@flyrain, what about creating a separate PR for `is_deleted`? I think these two are independent. We can approach vectorized and non-vectorized readers separately.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: We can definitely refactor here. I've got some ideas in my mind. Trying to understand your point here. Do you mean a class `DeleteStream` combine both functionality from `PositionSetDeleteFilter` and `PositionSetDeleteMarker`? I will test these ideas anyway. Thanks for the suggestion.

api/src/main/java/org/apache/iceberg/actions/ActionsProvider.java
```diff
@@ -74,4 +74,11 @@ default ExpireSnapshots expireSnapshots(Table table) {
   default DeleteReachableFiles deleteReachableFiles(String metadataLocation) {
     throw new UnsupportedOperationException(this.getClass().getName() + " does not implement deleteReachableFiles");
   }
+
+  /**
+   * Instantiates an action to generate a change data set.
+   */
+  default GetChangeSet getChangeSet(Table table) {
```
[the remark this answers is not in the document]
stevenzwu: Anton made a good argument here. `generate` is not much longer

core/src/main/java/org/apache/iceberg/ManifestGroup.java
stevenzwu: @aokolnychyi we are not talking about extending `TableScan`. We are talking about new scan interfaces to support incremental scan. Please see PR 4580 : https://github.com/apache/iceberg/pull/4580#discussion_r856499496

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
stevenzwu: yeah, I made a similar comment here: https://github.com/apache/iceberg/pull/4539#discussion_r855600180

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
flyrain: We should consider them incompatible, right? For example, if you do 
```
getChangeSet(table).betweenSnapshots(s1, s2).forCurrentSnapshot(s3).execute()
```
It should only honor the later one, which is `forCurrentSnapshot(s3)`. If you flip the order like this 
```
getChangeSet(table).forCurrentSnapshot(s3).betweenSnapshots(s1, s2).execute()
```
It should only pick up snapshots between s1 and s3. 
The current logic support this logic already. Basically, it cleans up the lists every time before adding any snapshot id into it.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
flyrain: Sure, will make the change. Double checked, they point to the same object.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
chenjunjiedada: Yes, like the Java steam API.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: No, they won't. The order will be kept by the metadata column `_commit_order` for multiple snapshots in terms of parallel executors. Each snapshot will get a unique `_commit_order` for its change set no matter how parallel the data is read.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseCdcSparkAction.java
flyrain: RECORD_TYPE indicates the change types.(I, D, -U, +U)
COMMIT_SNAPSHOT_ID indicates which snapshot makes the change record.
COMMIT_TIMESTAMP indicates when the change happend.
COMMIT_ORDER indicates change order in case of multiple snapshots.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCdcWithMerge.java
flyrain: As I mentioned here, https://github.com/apache/iceberg/pull/4539#discussion_r855609262, we only apply the pos deletes for data files generated in the same snapshot. We still emit pos deletes when they delete rows from previous snapshots.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseGetChangeSetSparkAction.java
flyrain: Yes, order is tricky when we output both deletes and inserts. I will remove the option to do that. It always applies the deletes within the same snapshot.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -185,14 +204,35 @@ private CloseableIterable<T> applyEqDeletes(CloseableIterable<T> records) {
         .reduce(Predicate::and)
         .orElse(t -> true);
 
-    Filter<T> remainingRowsFilter = new Filter<T>() {
+    Filter<T> remainingRowsFilter = hasMetadataColumnIsDeleted ? getMarker(remainingRows) : getFilter(remainingRows);
```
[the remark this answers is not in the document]
flyrain: Yeah, I think the delete logic is cleaner in that way. We are more or less in that direction.

api/src/main/java/org/apache/iceberg/actions/GetChangeSet.java
flyrain: Made it return a generic type.

flyrain:
There will be no update on this PR. Please check the latest progress here, https://github.com/apache/iceberg/projects/26.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Emit changed data set for the current snapshot.
+   *
+   * @return this for method chaining
+   */
+  GenerateChangeSet forCurrentSnapshot();
+
+  /**
+   * Emit changed data from a particular snapshot(exclusive).
```
RussellSpitzer, line 39: I think "from" here is ambiguous, probably "since" is a bit closer?

PR #4540, opened 2022-04-11
Spark: `DESC TABLE EXTENDED` shows table format-version

liuml07:
Currently Spark SQL does not show table `format-version` in `SHOW TBLPROPERTIES`, `DESCRIBE TABLE` or `SHOW CREATE TABLE`. This could be based on the fact that `format-version` is a reserved table property and it's not persisted as usual table properties.

Still users may find it useful to inspect the table `format-version` via Spark SQL. One way is to show it via `DESC TABLE EXTENDED`. Towards that, this pull requests has 3 small commits:
1. Add a one liner test which explains the intention and fails
2. Show `format-version` for command `DESC TABLE EXTENDED` by adding it to Spark table properties
3. (cosmetic) Replace some literals with constants defined in `TableProperties.java`

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
@@ -72,14 +72,17 @@ import org.apache.spark.sql.util.CaseInsensitiveStringMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.iceberg.TableProperties.CURRENT_SNAPSHOT_ID;
+import static org.apache.iceberg.TableProperties.FORMAT_VERSION;
+
 public class SparkTable implements org.apache.spark.sql.connector.catalog.Table,
     SupportsRead, SupportsWrite, SupportsDelete, SupportsRowLevelOperations, SupportsMetadataColumns {
 
   private static final Logger LOG = LoggerFactory.getLogger(SparkTable.class);
 
   private static final Set<String> RESERVED_PROPERTIES =
-      ImmutableSet.of("provider", "format", "current-snapshot-id", "location", "sort-order", "identifier-fields",
-          "format-version");
+      ImmutableSet.of("provider", "format", CURRENT_SNAPSHOT_ID, "location", FORMAT_VERSION, "sort-order",
+          "identifier-fields");
   private static final Set<TableCapability> CAPABILITIES = ImmutableSet.of(
       TableCapability.BATCH_READ,
       TableCapability.BATCH_WRITE,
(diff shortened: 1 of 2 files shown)


liuml07:
Screen Shot 2022-04-11 at 4 44 08 PM

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
```diff
@@ -490,6 +490,11 @@ public void testSparkTableReservedProperties() throws Exception {
     Assert.assertEquals("Unexpected format", "iceberg/parquet", table.properties().get("format"));
     Assert.assertNotEquals("No current-snapshot-id found", "none", table.properties().get("current-snapshot-id"));
     Assert.assertTrue("Location isn't correct", table.properties().get("location").endsWith(destTableName));
+
+    Assert.assertNotNull("Missing format-version", table.properties().get("format-version"));
```
liuml07, line 494: Not asserting the value being 1 in case we move the default table format-version to 2 (this test may simply break then)

kbendick: I think it's fine to assert the value is 1. That's the current expected behavior (format-version is 1 for tables created by Spark).

If / when we update the default to 2, we'll update the test.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
```diff
@@ -157,8 +162,10 @@ public Map<String, String> properties() {
     propsBuilder.put("provider", "iceberg");
     String currentSnapshotId = icebergTable.currentSnapshot() != null ?
         String.valueOf(icebergTable.currentSnapshot().snapshotId()) : "none";
-    propsBuilder.put("current-snapshot-id", currentSnapshotId);
+    propsBuilder.put(CURRENT_SNAPSHOT_ID, currentSnapshotId);
     propsBuilder.put("location", icebergTable.location());
+    int formatVersion = ((HasTableOperations) icebergTable).operations().current().formatVersion();
```
rdblue, line 167: This is used for other implementations as well that may not implement `HasTableOperations`. Instead, this should check whether `icebergTable` is an instance of `BaseTable` and add the property only if it is. `BaseTable` exposes the table operations.

liuml07: Kyle was suggesting `BaseTable` in Slack as well, but I was thinking `HasTableOperations` is an interface that always exposes the table operation. My hunch was wrong. Glad to know & updated the code. Thank you.

rdblue (approved):

PR #4544, opened 2022-04-12
[REST] Update Snapshot ref in Open-API spec

samredai:
This is a small fix to the OpenAPI spec. It looks like summary is actually a top level requirement for a snapshot object and the property "operation" is mislabeled "summary".

### DIFF
--- a/open-api/rest-catalog-open-api.yaml
+++ b/open-api/rest-catalog-open-api.yaml
@@ -1015,6 +1015,7 @@ components:
         - snapshot-id
         - timestamp-ms
         - manifest-list
+        - summary
       properties:
         snapshot-id:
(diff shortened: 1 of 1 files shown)


open-api/rest-catalog-open-api.yaml
```diff
@@ -1027,10 +1028,8 @@ components:
           type: integer
         summary:
           type: object
-          required:
-            - summary
```
rdblue, line 1031: The operation is actually required, so this should just change to `operation` instead.

samredai: Added operation as required 👍

rdblue (approved):

PR #3677, opened 2021-12-06
Python: Adding TableMetadata object from dict, bytestream, and InputFile implementation

samredai:
**UPDATE:** This has been updated to remove the wrapper mechanism and instead assigns attributes to the class explicitly. The s3 io has also been abstracted to a generic file io in the `from_file()` method which is dependent on PR #3691.

I've also included the popular [attrs](https://github.com/python-attrs/attrs) library to allow for a cleaner class definition and also to freeze `TableMetadata` instances. There's agreement that table metadata instances should be frozen but I anticipate some discussions around how to actually perform table metadata updates so I think it's best to tackle that in follow-up PR's/proposals. 

---

This is the table metadata portion of #3227. This isn't complete but wanted it to be visible to get feedback on the direction. The equivalent logic in the legacy implementation is stored in [table_metadata.py](https://github.com/apache/iceberg/blob/master/python_legacy/iceberg/core/table_metadata.py) and [table_metadata_parser.py](https://github.com/apache/iceberg/blob/master/python_legacy/iceberg/core/table_metadata_parser.py).

The idea here is that, instead of including very explicit parsing and validation logic for table metadata files, we can rely on the standard library in conjunction with [jsonschema](https://json-schema.org/) tooling to accomplish both. The `TABLE_METADATA_V2_SCHEMA` jsonschema definition found in metadata.py is an example of how this can be done (still needs to be tuned to the spec exactly). The `TableMetadata` class itself naively parses a given json object and includes a `validate()` method that validates against the defined jsonschema. (`validate()` simply calls the `validate_v1()` or `validate_v2()` static method.) Table metadata values can then be retrieved using simple dot notation and can be updated as well.
```py
table_metadata = TableMetadata.from_s3("s3://foo/bar/baz.metadata.json", version=2)
print(table_metadata.properties.read_split_target_size) # 134217728
table_metadata.properties.read_split_target_size = 268435456
print(table_metadata.properties.read_split_target_size) # 268435456
```

Once tweaked, the jsonschema definition should prove re-usable since jsonschema parsers exist in almost every language. It may also be valuable to include a blessed jsonschema definition in the Iceberg docs.

A proposal for editing table metadata is to use a collection of functions that update value(s), validate that the schema is still valid, and return the updated `TableMetadata` instance, such as:
```py
from copy import deepcopy
import time

def rollback(table_metadata: TableMetadata, snapshot_id: int):
  new_table_metadata = deepcopy(table_metadata)
  now_millis = int(time.time() * 1000)
[...]

PR #4545, opened 2022-04-12
[Python] Add a symlink to the OpenAPI rest catalog spec and package it with the python library

samredai:
This adds a symlink for the REST catalog OpenAPI spec contained in `open-api/rest-catalog-open-api.yaml` and a utility function to load it. The function uses `@lru_cache` to cache the file after it's loaded once.

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -45,6 +45,8 @@ python_requires = >=3.7
 install_requires =
     mmh3
     singledispatch
+    pyyaml
+include_package_data = true
 [options.extras_require]
 arrow =
     pyarrow
@@ -53,3 +55,6 @@ dev=
     pytest
 [options.packages.find]
 where = src
+[options.package_data]
+""=
+    ../src/iceberg/assets/rest-catalog-open-api.yaml
--- a/python/src/iceberg/utils/openapi.py
+++ b/python/src/iceberg/utils/openapi.py
@@ -1,0 +1,15 @@
+import os
+from functools import lru_cache
+
+import yaml
+
+from iceberg import assets
+
+
+@lru_cache(maxsize=1)
+def read_yaml_file(path: str):
+    return yaml.safe_load(open(path))
+
+
+def load_openapi_spec():
+    return read_yaml_file(os.path.join(list(assets.__path__)[0], "rest-catalog-open-api.yaml"))
--- a/python/src/iceberg/utils/openapi.py
+++ b/python/src/iceberg/utils/openapi.py
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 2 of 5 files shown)


python/setup.cfg
```diff
@@ -45,6 +45,8 @@ python_requires = >=3.7
 install_requires =
     mmh3
     singledispatch
+    pyyaml
```
kbendick, line 48: Does this make pyyaml a dependency of the library itself? If so, is there anyway to make it just a “test” dependency or something or will we need it at runtime?

samredai: We'll need it at runtime. Specifically in the `TableMetadata` class which hasn't been merged in yet but the idea is that reading in a table metadata file will be validated against the open api spec.

kbendick: Are we gonna do that for every file on every read? Or just as a test?

samredai: It's a great question and I think we can decide on that later. The TableMetadata PR is still a WIP but that class includes this [validate](https://github.com/apache/iceberg/pull/3677/files#diff-0a2dcd19c1079e50703f2010f74a47309468503f3ddd753db3aca9951364810fR236) method. So we may have certain paths where we don't validate (loading a metadata file from a path we got from a catalog) and other paths where we do validate (like before writing+committing a new metadata file), just as an example.

Fokko (approved):

python/setup.cfg
Fokko: I think it would be good to pin it on a version. Currently it will just install the latest greatest, but there might be breaking changes (happens all the time in Python land :)
```suggestion
    pyyaml==6.0
```

python/src/iceberg/utils/openapi.py
```diff
@@ -0,0 +1,32 @@
...
+
+import os
+from functools import lru_cache
+
+import yaml
+
+from iceberg import assets
+
+
+@lru_cache(maxsize=1)
```
Fokko, line 26: Once https://github.com/apache/iceberg/pull/4784 has been merged, we can simplify this:
```suggestion
@cache
```
https://docs.python.org/3/library/functools.html#functools.cache

python/src/iceberg/utils/openapi.py
```diff
@@ -0,0 +1,32 @@
...
+def read_yaml_file(path: str):
+    return yaml.safe_load(open(path))
```
Fokko, line 28: Nit: If something happens, we'll clean up the connection:
```python
with open(path) as f:
    return yaml.safe_load(f
```

python/src/iceberg/utils/openapi.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+from functools import lru_cache
+
+import yaml
+
+from iceberg import assets
+
+
+@lru_cache(maxsize=1)
+def read_yaml_file(path: str):
```
Fokko, line 27:
```suggestion
def read_yaml_file(path: str) -> Dict[str, Any]:
```

python/src/iceberg/utils/openapi.py
```diff
@@ -0,0 +1,32 @@
...
+    return yaml.safe_load(open(path))
+
+
+def load_openapi_spec():
```
Fokko, line 31:
```suggestion
def load_openapi_spec() -> Dict[str, Any]:
```

samredai:
Closed this since there's still some discussions around how we want to add schema validation and that can be postponed without blocking anything.

Issue #3014, opened 2021-08-23
PartitionSpec isUnpartitioned returns true for tables which previously had Partitions but no longer do


RussellSpitzer:
Because SpecV1 creates "void" transforms when removing partition transforms, a v1 table whose partitions are removed will always return an unpartitioned spec.

https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L97-L99

This method should always return true if all of the transforms are void as well.

xinbinhuang:
@RussellSpitzer can you assign this to me, and timebox it till Thursday?  Wanna take a look at iceberg code, but don't want to block it if I get occupied by other things

RussellSpitzer:
Done! Let me know if anything is u nclear

RussellSpitzer:
@xinbinhuang How are things going?

xinbinhuang:
Hi @RussellSpitzer  , I had some time to look into the code last week but haven't gotten the time to draft up a PR yet. Planing to do it in the next two days. But I'm aware it has passed quite a few days from last Thursday, so if it'll be too late to get into the release cycle or it blocks other works. Please go ahead to fix it :)

Thanks!
Bin

PR #3059, opened 2021-09-01
Returns isUnpartitioned=true for VoidTransform on all fields

xinbinhuang:
closes: #3014

PR #4546, opened 2022-04-12
Hive: Expose default partition spec and sort order in HMS.

flyrain:
cc @aokolnychyi @RussellSpitzer @szehon-ho @karuppayya @anuragmantri @kbendick @singhpk234

### DIFF
--- a/core/src/main/java/org/apache/iceberg/PartitionSpecParser.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionSpecParser.java
@@ -39,61 +39,48 @@ public class PartitionSpecParser {
   private static final String SPEC_ID = "spec-id";
   private static final String FIELDS = "fields";
   private static final String SOURCE_ID = "source-id";
-  private static final String SOURCE_NAME = "source-name";
   private static final String FIELD_ID = "field-id";
   private static final String TRANSFORM = "transform";
   private static final String NAME = "name";
 
-  public static String toJsonWithSourceName(PartitionSpec spec) {
-    return toJson(spec.toUnbound(), spec.schema(), false);
+  public static void toJson(PartitionSpec spec, JsonGenerator generator) throws IOException {
+    toJson(spec.toUnbound(), generator);
   }
 
   public static String toJson(PartitionSpec spec) {
     return toJson(spec, false);
   }
 
+  public static String toJson(PartitionSpec spec, boolean pretty) {
+    return toJson(spec.toUnbound(), pretty);
+  }
+
+  public static void toJson(UnboundPartitionSpec spec, JsonGenerator generator) throws IOException {
+    generator.writeStartObject();
+    generator.writeNumberField(SPEC_ID, spec.specId());
+    generator.writeFieldName(FIELDS);
+    toJsonFields(spec, generator);
+    generator.writeEndObject();
+  }
+
   public static String toJson(UnboundPartitionSpec spec) {
     return toJson(spec, false);
   }
 
-  public static String toJson(PartitionSpec spec, boolean pretty) {
-    return toJson(spec.toUnbound(), null, pretty);
-  }
-
   public static String toJson(UnboundPartitionSpec spec, boolean pretty) {
-    return toJson(spec, null, pretty);
-  }
-
-  private static String toJson(UnboundPartitionSpec spec, Schema schema, boolean pretty) {
     try {
       StringWriter writer = new StringWriter();
       JsonGenerator generator = JsonUtil.factory().createGenerator(writer);
       if (pretty) {
         generator.useDefaultPrettyPrinter();
       }
-      toJson(spec, generator, schema);
+      toJson(spec, generator);
       generator.flush();
       return writer.toString();
 
     } catch (IOException e) {
       throw new RuntimeIOException(e);
     }
-  }
-
-  public static void toJson(PartitionSpec spec, JsonGenerator generator) throws IOException {
-    toJson(spec.toUnbound(), generator);
-  }
-
-  public static void toJson(UnboundPartitionSpec spec, JsonGenerator generator) throws IOException {
-    toJson(spec, generator, null);
-  }
-
-  private static void toJson(UnboundPartitionSpec spec, JsonGenerator generator, Schema schema) throws IOException {
-    generator.writeStartObject();
-    generator.writeNumberField(SPEC_ID, spec.specId());
-    generator.writeFieldName(FIELDS);
-    toJsonFields(spec, generator, schema);
-    generator.writeEndObject();
   }
 
   public static PartitionSpec fromJson(Schema schema, JsonNode json) {
@@ -125,22 +112,16 @@ public class PartitionSpecParser {
   }
 
   static void toJsonFields(PartitionSpec spec, JsonGenerator generator) throws IOException {
-    toJsonFields(spec.toUnbound(), generator, null);
+    toJsonFields(spec.toUnbound(), generator);
   }
 
-  static void toJsonFields(UnboundPartitionSpec spec, JsonGenerator generator, Schema schema) throws IOException {
+  static void toJsonFields(UnboundPartitionSpec spec, JsonGenerator generator) throws IOException {
     generator.writeStartArray();
     for (UnboundPartitionSpec.UnboundPartitionField field : spec.fields()) {
       generator.writeStartObject();
       generator.writeStringField(NAME, field.name());
       generator.writeStringField(TRANSFORM, field.transformAsString());
       generator.writeNumberField(SOURCE_ID, field.sourceId());
-      if (schema != null) {
-        Types.NestedField nestedField = schema.findField(field.sourceId());
-        if (nestedField != null) {
-          generator.writeStringField(SOURCE_NAME, nestedField.name());
-        }
-      }
       generator.writeNumberField(FIELD_ID, field.partitionId());
       generator.writeEndObject();
     }
--- a/core/src/main/java/org/apache/iceberg/PartitionSpecParser.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionSpecParser.java
@@ -39,48 +39,61 @@ public class PartitionSpecParser {
   private static final String SPEC_ID = "spec-id";
   private static final String FIELDS = "fields";
   private static final String SOURCE_ID = "source-id";
+  private static final String SOURCE_NAME = "source-name";
   private static final String FIELD_ID = "field-id";
   private static final String TRANSFORM = "transform";
   private static final String NAME = "name";
 
-  public static void toJson(PartitionSpec spec, JsonGenerator generator) throws IOException {
-    toJson(spec.toUnbound(), generator);
+  public static String toJsonWithSourceName(PartitionSpec spec) {
+    return toJson(spec.toUnbound(), spec.schema(), false);
   }
 
   public static String toJson(PartitionSpec spec) {
     return toJson(spec, false);
   }
 
-  public static String toJson(PartitionSpec spec, boolean pretty) {
-    return toJson(spec.toUnbound(), pretty);
-  }
-
-  public static void toJson(UnboundPartitionSpec spec, JsonGenerator generator) throws IOException {
-    generator.writeStartObject();
-    generator.writeNumberField(SPEC_ID, spec.specId());
-    generator.writeFieldName(FIELDS);
-    toJsonFields(spec, generator);
-    generator.writeEndObject();
-  }
-
   public static String toJson(UnboundPartitionSpec spec) {
     return toJson(spec, false);
   }
 
+  public static String toJson(PartitionSpec spec, boolean pretty) {
+    return toJson(spec.toUnbound(), null, pretty);
+  }
+
   public static String toJson(UnboundPartitionSpec spec, boolean pretty) {
+    return toJson(spec, null, pretty);
+  }
+
+  private static String toJson(UnboundPartitionSpec spec, Schema schema, boolean pretty) {
     try {
       StringWriter writer = new StringWriter();
       JsonGenerator generator = JsonUtil.factory().createGenerator(writer);
       if (pretty) {
         generator.useDefaultPrettyPrinter();
       }
-      toJson(spec, generator);
+      toJson(spec, generator, schema);
       generator.flush();
       return writer.toString();
 
     } catch (IOException e) {
       throw new RuntimeIOException(e);
     }
+  }
+
+  public static void toJson(PartitionSpec spec, JsonGenerator generator) throws IOException {
+    toJson(spec.toUnbound(), generator);
+  }
+
+  public static void toJson(UnboundPartitionSpec spec, JsonGenerator generator) throws IOException {
+    toJson(spec, generator, null);
+  }
+
+  private static void toJson(UnboundPartitionSpec spec, JsonGenerator generator, Schema schema) throws IOException {
+    generator.writeStartObject();
+    generator.writeNumberField(SPEC_ID, spec.specId());
+    generator.writeFieldName(FIELDS);
+    toJsonFields(spec, generator, schema);
+    generator.writeEndObject();
   }
 
   public static PartitionSpec fromJson(Schema schema, JsonNode json) {
@@ -112,16 +125,22 @@ public class PartitionSpecParser {
   }
 
   static void toJsonFields(PartitionSpec spec, JsonGenerator generator) throws IOException {
-    toJsonFields(spec.toUnbound(), generator);
+    toJsonFields(spec.toUnbound(), generator, null);
   }
 
-  static void toJsonFields(UnboundPartitionSpec spec, JsonGenerator generator) throws IOException {
+  static void toJsonFields(UnboundPartitionSpec spec, JsonGenerator generator, Schema schema) throws IOException {
     generator.writeStartArray();
     for (UnboundPartitionSpec.UnboundPartitionField field : spec.fields()) {
       generator.writeStartObject();
       generator.writeStringField(NAME, field.name());
       generator.writeStringField(TRANSFORM, field.transformAsString());
       generator.writeNumberField(SOURCE_ID, field.sourceId());
+      if (schema != null) {
+        Types.NestedField nestedField = schema.findField(field.sourceId());
+        if (nestedField != null) {
+          generator.writeStringField(SOURCE_NAME, nestedField.name());
+        }
+      }
       generator.writeNumberField(FIELD_ID, field.partitionId());
       generator.writeEndObject();
     }
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -71,12 +71,12 @@ public class TableProperties {
   public static final String CURRENT_SCHEMA = "current-schema";
 
   /**
-   * Reserved table property for the JSON representation of current partition spec.
+   * Reserved table property for the JSON representation of current(default) partition spec.
    */
   public static final String DEFAULT_PARTITION_SPEC = "default-partition-spec";
 
   /**
-   * Reserved table property for the JSON representation of current sort order.
+   * Reserved table property for the JSON representation of current(default) sort order.
    */
   public static final String DEFAULT_SORT_ORDER = "default-sort-order";
 
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -66,6 +66,11 @@ public class TableProperties {
   public static final String CURRENT_SNAPSHOT_TIMESTAMP = "current-snapshot-timestamp-ms";
 
   /**
+   * Reserved table property for the JSON representation of current schema.
+   */
+  public static final String CURRENT_SCHEMA = "current-schema";
+
+  /**
    * Reserved table property for the JSON representation of current partition spec.
    */
   public static final String DEFAULT_PARTITION_SPEC = "default-partition-spec";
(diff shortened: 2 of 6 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -433,6 +437,20 @@ void setSnapshotSummary(Map<String, String> parameters, Snapshot currentSnapshot
     }
   }
 
+  private void setPartitionSpec(TableMetadata metadata, Map<String, String> parameters) {
+    parameters.remove(TableProperties.DEFAULT_PARTITION_SPEC);
+    if (metadata.spec() != null && metadata.spec().isPartitioned()) {
+      parameters.put(TableProperties.DEFAULT_PARTITION_SPEC, PartitionSpecParser.toJson(metadata.spec()));
```
flyrain, line 443: `PartitionSpecParser.toJson()` doesn't convert a source column id to its name. For example, here is an example. 
```
{"spec-id":0,"fields":[{"name":"data_bucket","transform":"bucket[16]","source-id":2,"field-id":1000}]}
```
It only shows the source id(2),  the column name is `data` as the following code shows. It'd be more user-friendly if we show the column name.
```
    Schema schema = new Schema(
        required(1, "id", Types.IntegerType.get(), "unique ID"),
        required(2, "data", Types.StringType.get())
    );
    PartitionSpec spec = PartitionSpec.builderFor(schema)
        .bucket("data", 16)
        .build();
```

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -559,4 +556,32 @@ public void testSetSnapshotSummary() throws Exception {
...
+        .bucket("data", 16)
+        .build();
+    TableIdentifier tableIdent = TableIdentifier.of(DB_NAME, "tbl");
+
+    try {
+      catalog.buildTable(tableIdent, schema)
+          .withPartitionSpec(spec)
+          .create();
+
+      Assert.assertEquals(PartitionSpecParser.toJson(spec),
```
singhpk234, line 576: [minor] how about asserting the case where we don't set default partition spec un-partitioned spec

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -433,6 +437,20 @@ void setSnapshotSummary(Map<String, String> parameters, Snapshot currentSnapshot
     }
   }
 
+  private void setPartitionSpec(TableMetadata metadata, Map<String, String> parameters) {
+    parameters.remove(TableProperties.DEFAULT_PARTITION_SPEC);
+    if (metadata.spec() != null && metadata.spec().isPartitioned()) {
```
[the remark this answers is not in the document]
flyrain: Thanks @singhpk234 for pointing out. #3059 is trying to fix #3014, and it is almost ready to merge. It should be fine in that case.

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -75,6 +75,20 @@ private TableProperties() {
    */
   public static final String CURRENT_SNAPSHOT_TIMESTAMP = "current-snapshot-timestamp-ms";
 
+  /**
+   * Reserved table property for default partition spec.
+   * <p>
+   * This reserved property is used to store the default partition spec.
+   */
+  public static final String DEFAULT_PARTITION_SPEC = "default-partition-spec";
```
kbendick, line 83: It might be helpful to clarify what is meant by `default` for both of these.

Each javadoc states that it’s for the “default” spec / sort order, but I’m still not entirely sure what that means.

Is it the current sort order / partition spec that would be used if the user doesn’t override it for an individual query ?

kbendick: Maybe something like “JSON representation of the table’s current configured partition spec, which will be used if not overridden for individual writes”. Kind of wordy but something along those lines would be helpful for me if quickly looking through the JavaDocs etc. Will leave that decision to you though.

flyrain: Yes, it is for the current partition spec and sort order. Make sense to me. Will make the change.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: Would we need to store schema on HMS side to make sense of the ids?  Otherwise today we are depending on the naming convention.

szehon-ho (commented):
I agree , once we change the word  default to current, it looks ok to me.

One concern I was thinking in general even from #4456 , whether all this JSON parsing increase the commit times which can be sensitive for Streaming applications (I guess having that flag to control partition length is good in $4456 as user can set it to 0 if so).  At some point should measure the commit time perf for large tables and maybe disable have flag to disable it together if its a concern, as right now there's no way to opt-out.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Schema has been stored in HMS, https://github.com/apache/iceberg/blob/116026b40a43e477f68a34f8eead831ef139bc60/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L247, but schema id is missing when it converts Iceberg schema to Hive table schema. It may make sense to replace the source-id with source-name.

flyrain:
> One concern I was thinking in general even from #4456 , whether all this JSON parsing increase the commit times which can be sensitive for Streaming applications (I guess having that flag to control partition length is good in #4456 as user can set it to 0 if so, but it doesn't help these ones). At some point should measure the commit time perf for large tables. Or just have a follow up pr for a flag to disable all these stats persistence altogether if its a concern, as right now there's no way to opt-out, wdyt?

Schema, snapshot summary, partition spec are more likely impact the commit perf. We may think of a general way to control that.  I'd suggest to handle that in another PR.

szehon-ho:
> Out of all table properties synced to HMS, schema, snapshot summary, partition spec will more likely impact the commit perf. We may think of a general way to control that. I'd suggest to handle that in another PR.

Yea I was mentioning, a follow-up PR is fine, or if someone has any other idea about this

core/src/main/java/org/apache/iceberg/PartitionSpecParser.java
```diff
@@ -83,6 +80,22 @@ public static String toJson(UnboundPartitionSpec spec, boolean pretty) {
...
+  }
+
+  private static void toJson(UnboundPartitionSpec spec, JsonGenerator generator, Schema schema) throws IOException {
+    generator.writeStartObject();
+    generator.writeNumberField(SPEC_ID, spec.specId());
+    generator.writeFieldName(FIELDS);
+    toJsonFields(spec, generator, schema);
+    generator.writeEndObject();
+  }
+
```
flyrain, line 98: Adjusted the order of methods, so that it looks better. From top to bottom, the methods `toJson` have 1 parameter, 2 parameters and 3 parameters respectively. Each layer depends on the under layer.
```
toJson(PartitionSpec spec)
toJson(UnboundPartitionSpec spec)

toJson(PartitionSpec spec, boolean pretty)
toJson(UnboundPartitionSpec spec, boolean pretty)

toJson(UnboundPartitionSpec spec, Schema schema, boolean pretty)
```

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Added the source name in the json string.

szehon-ho: (Chat offline) 

Imo, I think instead of changing it, we should just store schema object then for users interested in separate field, its more accurate then Hive schema and a bit less hacky than changing the partition-spec format to add it.

flyrain: Added a new property `current-schema` to avoid adding column name into partition spec and sort order. It duplicates the schema with Hive format, but we think it's worth to do that. The Hive format loses important column info(id).

core/src/main/java/org/apache/iceberg/TableProperties.java
szehon-ho: I think you reverted too many?  Should be 'current'.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -433,6 +439,43 @@ void setSnapshotSummary(Map<String, String> parameters, Snapshot currentSnapshot
     }
   }
 
+  private void setSchema(TableMetadata metadata, Map<String, String> parameters) {
+    parameters.remove(TableProperties.CURRENT_SCHEMA);
+    if (metadata.schema() != null) {
```
szehon-ho, line 444: Lets' have common method and re-use to set various fields like
```
private void setField(TableMetadata metadata, Map<String, String> parameters, String key, String value) {
   parameters.remove(key);
   if (value.length <= maxHiveTablepropertySize) {
      parameters.put(key, value);
   } else {
      LOG.warn("Not exposing {} in HMS since it exceeds {} characters", maxHiveTablePropertySize);
   }
}
```

szehon-ho (commented):
Looks better , thanks @flyrain , few comments on the code side.

As discussed offline and as I understand, these changes are for users of HiveCatalog who are interested in getting a better understanding of Iceberg metadata directly from HIve API.  This could be done by RestCatalog functionalities once that has more adoptions.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: Yea also i think Hive columns may not have exact Iceberg schema in some cases.  For instance, the type conversions are not exact 1-to-1 mapping.

core/src/main/java/org/apache/iceberg/TableProperties.java
flyrain: I was trying to be consistent with the name in metadata.json, which is `default-partition-spec`. The same for sort order.  I changed the comments though.

szehon-ho: OK I see, yea I always find it confusing. 

In the comment, maybe we can add that they are equivalent, otherwise the comment is even more confusing:
```
Reserved table property for the JSON representation of current (default) schema.
```

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Made the change.

core/src/main/java/org/apache/iceberg/TableProperties.java
flyrain: It is confusing. I like the `current` more. I keep the original name just for consistency.

szehon-ho (approved):
Optional: Maybe we can add a test, about schema being too big to serialize, like snapshotSummary test

szehon-ho:
FYI @singhpk234 if you want to take another look ?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -433,6 +439,38 @@ void setSnapshotSummary(Map<String, String> parameters, Snapshot currentSnapshot
...
+      setField(parameters, TableProperties.DEFAULT_SORT_ORDER, sortOrder);
+    }
+  }
+
+  private void setField(Map<String, String> parameters, String key, String value) {
+    if (value.length() <= maxHiveTablePropertySize) {
+      parameters.put(key, value);
+    } else {
+      LOG.warn("Not exposing {} in HMS since it exceeds {} characters", key, maxHiveTablePropertySize);
+    }
```
singhpk234, line 471: [nit] should we make this : 

https://github.com/apache/iceberg/blob/449a7438b83c475d4bf15e41a5c7aa0021ca0b0c/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L425-L430

also use this helper

singhpk234 (approved):
Looks good to me as well :) !!!

Thanks @flyrain

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Was trying to do that, but the warn message needs snapshot id here. But it requires changes for method `setFiled()` like the below, and changes for all other callers.  It's like removing one duplication, but adding a few complication. I'd suggest to keep it as is.
```
setField(Map<String, String> parameters, String key, String value, String warnMessage) 
```

flyrain:
> Optional: Maybe we can add a test, about schema being too big to serialize, like snapshotSummary test

Made the change

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -433,6 +439,38 @@ void setSnapshotSummary(Map<String, String> parameters, Snapshot currentSnapshot
...
+
+  private void setSortOrder(TableMetadata metadata, Map<String, String> parameters) {
+    parameters.remove(TableProperties.DEFAULT_SORT_ORDER);
+    if (metadata.sortOrder() != null && metadata.sortOrder().isSorted()) {
+      String sortOrder = SortOrderParser.toJson(metadata.sortOrder());
+      setField(parameters, TableProperties.DEFAULT_SORT_ORDER, sortOrder);
+    }
+  }
+
+  private void setField(Map<String, String> parameters, String key, String value) {
```
szehon-ho, line 466: One performance suggestion, if the user sets to 0 (disable this feature), we can skip the serialization for performance.  

Maybe , easiest, we can we add some boolean function like exposeViaTableProperties() that checks if value is 0, and use it in all the methods?

```
    if (exposeInHmsProperties() && metadata.sortOrder() != null && metadata.sortOrder().isSorted()) {
```

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -52,8 +52,11 @@
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
```
szehon-ho, line 52: Two other suggestion for this class: can we add in comment of "HIVE_TABLE_PROPERTY_MAX_SIZE" , one more sentence to let user know how to turn off feature?

```
// set to 0 to not expose Iceberg metadata in HMS properties
```
And also, a precondition in HiveTableOperations constructor to check if  value is non-negative.

szehon-ho (commented):
Thanks @flyrain i had a few more comments , can you see if they make sense.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Setting it to 0 can control some of them(snapshot summary, schema, partition spec, and sort order), but not all of them. I'd suggest to create another PR to make sure all are taken care.

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -562,6 +562,60 @@ public void testSetSnapshotSummary() throws Exception {
...
+          hmsTableParameters().get(TableProperties.CURRENT_SCHEMA));
+
+      // add many new fields to make the schema json string exceed the limit
+      UpdateSchema updateSchema = table.updateSchema();
+      for (int i = 0; i < 600; i++) {
+        updateSchema.addColumn("new_col_" + i, Types.StringType.get());
+      }
+      updateSchema.commit();
+
+      Assert.assertTrue(SchemaParser.toJson(table.schema()).length() > 32672);
```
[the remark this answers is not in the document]
flyrain: It is not re-calculated here. Am I missing something?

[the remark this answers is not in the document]
szehon-ho: Line 597 and line 607 both call: SchemaParser.toJson(table.schema()), isn't it?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: Make sense, but could we at least take care of the ones in this PR?  ( schema, partition spec, sort order)?  We can have a follow up for the other ones not touched by this PR.

Just didn't want to leave it in a state where we are wasting CPU cycle (JSON serialization) needlessly if the user turns off this feature.  As this is done in the critical commit block, unlike the original serialization which happens before.  The other HMS table properties to me are also less CPU intensive as they are just getting a field.

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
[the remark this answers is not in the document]
flyrain: They are different, we updated schema in the middle.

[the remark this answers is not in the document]
szehon-ho: Yea, sorry you are right!

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Make sense, made the change.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
flyrain: Added the comment. Negative is fine, right?

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -557,11 +563,93 @@ public void testSetSnapshotSummary() throws Exception {
...
     // the limit has been updated to 4000 instead of the default value(32672)
     Assert.assertTrue(summarySize > 4000 && summarySize < 32672);
-    parameters.remove(TableProperties.CURRENT_SNAPSHOT_SUMMARY);
+    parameters.remove(CURRENT_SNAPSHOT_SUMMARY);
     spyOps.setSnapshotSummary(parameters, snapshot);
     Assert.assertEquals("The snapshot summary must not be in parameters due to the size limit", 0, parameters.size());
   }
 
+  @Test
+  public void testNotExposeTableProperties() {
```
flyrain, line 572: A new test added for method `exposeInHmsProperties()`

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCatalog.java
```diff
@@ -557,11 +563,93 @@ public void testSetSnapshotSummary() throws Exception {
...
+    Configuration conf = new Configuration();
+    conf.set("iceberg.hive.table-property-max-size", "0");
+    HiveTableOperations spyOps = spy(new HiveTableOperations(conf, null, null, catalog.name(), DB_NAME, "tbl"));
```
szehon-ho, line 575: Nit: Unnecessary spy

szehon-ho (approved):
Thanks @flyrain looks good to me now, will merge tomorrow unless others have some review comment.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
szehon-ho: I think probably better to disallow negative as it makes little sense?  But to me its ok either way.

szehon-ho:
Merged, thanks @flyrain for contribution and @singhpk234 for additional review

flyrain:
Thanks @szehon-ho for the detailed review and merge! Thanks @kbendick and @singhpk234 for the review.

pvary:
Maybe a little late in the game (I was on a longer PTO), and only partially related - since this is a Hive 4 feature -, but we did some work to display the partitioning information in the `DESCRIBE FORMATTED` command:
See: https://issues.apache.org/jira/browse/HIVE-25326

```
# Partition Transform Information	 	 
# col_name            	transform_type      	 
b                   	IDENTITY            	 
c                   	IDENTITY  
```

flyrain:
Thanks @pvary. We want to support the use case that doesn't need storage access. Does [HIVE-25326](https://issues.apache.org/jira/browse/HIVE-25326) need to access the metadata.json file to show partition spec?

pvary:
@flyrain: Yes, it uses the [HiveIcebergStorageHandler.getPartitionTransformSpec](https://github.com/apache/hive/blob/dbdcf00dd6334acaded4369fc0c1ccbdd142255e/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java#L330) which in turn uses [IcebergTableUtil.getTable](https://github.com/apache/hive/blob/dbdcf00dd6334acaded4369fc0c1ccbdd142255e/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java#L54) to load the table and read the spec from the table snapshot.

When we were working on creating Hive tables, I have started working on the Hive table creation we used the `spec` and the `schema` table property fields, but then we decided on removing them and keep the metadata json as a single source of truth. See: [HiveIcebergMetaHook. PROPERTIES_TO_REMOVE](https://github.com/apache/iceberg/blob/master/mr/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L62). I hope this does not cause issues with the new code when the table is created.

flyrain:
Got it. It should not. This PR is using a different name than `InputFormatConfig.PARTITION_SPEC(iceberg.mr.table.partition.spec)`. Thanks for the reminder.

PR #4551, opened 2022-04-13
Flink - Remove support for Flink 1.12

kbendick:
Remove support for Flink 1.12, which has been deprecated.

Flink 1.12 has been deprecated and is no longer actively maintained upstream, as the Flink community retains support for 2 major versions at a time.


versions.props
```diff
@@ -1,7 +1,7 @@
 org.slf4j:* = 1.7.25
 org.apache.avro:avro = 1.10.1
 org.apache.calcite:* = 1.10.0
-org.apache.flink:* = 1.12.5
+org.apache.flink:* = 1.14.3
```
kbendick, line 4: I don't think this is being used at all, but I updated it to the latest released Flink version.

stevenzwu (approved):

versions.props
stevenzwu: yeah. it does seem that we can remove this property

hililiwei (approved):

versions.props
kbendick: There are a few that I think can be removed (eg the Spark dependencies).

One that I’m less sure about is Calcite

I can remove the Flink property in this PR or attempt to clean up several in a secondary PR (or both).

openinx (approved):
Looks good to me overall ! Thanks @kbendick for the contribution, and thanks @stevenzwu & @hililiwei for the reviewing !

PR #4554, opened 2022-04-13
Core - Rename REST Catalog server configuration class to ConfigResponse

kbendick:
In the REST Catalog's `initialize` lifecycle, the server sends back some configuration overrides & fallback (default) values, which are merged with whatever the application provided.

This renames the class from `RESTCatalogConfigResponse` to `ConfigResponse`, as the class already is in the package `org.apache.iceberg.rest.responses` anyway.1

Also updates the spec and fixes a slightly incorrect example `ConfigResponse` object.

cc @rdblue

### DIFF
--- a/open-api/rest-catalog-open-api.yaml
+++ b/open-api/rest-catalog-open-api.yaml
@@ -98,15 +98,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CatalogConfiguration'
+                $ref: '#/components/schemas/CatalogConfig'
               example: {
-                "data": {
-                  "overrides": {
-                    "warehouse": "s3://bucket/warehouse/"
-                  },
-                  "defaults": {
-                    "clients": "4"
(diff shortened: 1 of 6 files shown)


open-api/rest-catalog-open-api.yaml
```diff
@@ -98,15 +98,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CatalogConfiguration'
+                $ref: '#/components/schemas/CatalogConfig'
               example: {
-                "data": {
```
kbendick, line 103: This top-level `data` object field is no longer used in the responses, but wasn't updated for this one example.

kbendick: Screenshot of the new example in the swagger UI.

image

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #4558, opened 2022-04-14
Core: Fix `Preconditions.checkArgument` error message format

liuml07:
I learnt from a PR review that `Precondition.checkArgument` only support `%s` in the error message template. See [Guava doc](https://github.com/google/guava/wiki/PreconditionsExplained). I see there are a few places in our code that uses `%d`. This simple PR is to replace `%d` with `%s` so that error message will be formatted exactly as intended.

The patch was generated using following command:
```bash
# one line
find . -name "*.java" | xargs sed -i '' '/Preconditions/ s/%d/%s/g'
# two lines
find . -name "*.java" | xargs sed -i '' '/Preconditions.*,$/{n;s/%d/%s/g;}'
```

I "reviewed" the patch and it seems good. I do not see other format pattern like `%f` or `%t` in existing code. One `checkState` is also fixed.

Let me know if this auto-generated code makes sense.


szehon-ho (approved):
Yea I think I gave the earlier review comment :), no idea it was so widespread, thanks for looking.

Though its disruptive, It seems worth to make this fix, I gave a try and the message also is wrong if not %s:

```Preconditions.checkArgument(false, "test %d %d", 1, 2);``` => ```test %d %d [1, 2]```

RussellSpitzer (approved):

RussellSpitzer:
Definitely worth doing
Thanks @liuml07 ! I always forget this

Thanks @szehon-ho for reviewing as well

szehon-ho:
Hi guys, I filed #4618 to make checkstyle rule for this , if someone wants to pick it up.

PR #4560, opened 2022-04-14
Core: Fix Partitions table for evolved partition specs

szehon-ho:
https://github.com/apache/iceberg/pull/4516 fixes the schema of the partitions table in the case of changed partition specs to use Partitioning.partitionType (a union of all previous partition specs), but the data is still wrong in some cases:

* The PartitionsMap is constructed with the table's current spec, which is used to generate hashcode of partition values.  This leads to colliding of some values which get lost. 
  * Ref:  https://github.com/apache/iceberg/blob/apache-iceberg-0.13.1/core/src/main/java/org/apache/iceberg/PartitionsTable.java#L99.  
  * Example:  p1=foo, p2=bar collides with p1=foo if PartitionMap is instantiated with current spec of {p1}, and other instance is lost.
* The partition values are just listed in order without any transformation, meaning they may in the wrong field in the unified schema.
  * Ref: https://github.com/apache/iceberg/blob/a78aa2dbdb98634f26066c457cc1aef93166be9f/core/src/main/java/org/apache/iceberg/PartitionsTable.java#L93.  
  * Example, spec evolution of p1,p2 => p2,p1 leads to wrong ordering of earlier partitions.

This fixes the problem by :
1.  Instantiating the PartitionMap by Partitioning.partitionType, so all types are used in the hashcode generation.
2. Transforming the PartitionDatas to fit the final schema Partioning.partitionType()

### DIFF
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -134,20 +134,20 @@ public class PartitionsTable extends BaseMetadataTable {
   }
 
   /**
-   * Convert a partition data written by an old spec, to table's normalized partition form, which is a common partition
+   * Convert a partition data written by an old spec, to table's normalized partition type, which is a common partition
    * type for all specs of the table.
    * @param originalPartition un-normalized partition data
-   * @param normalizedPartitionSchema table's normalized partition form {@link Partitioning#partitionType(Table)}
+   * @param normalizedPartitionType table's normalized partition type {@link Partitioning#partitionType(Table)}
    * @param originalPartitionFieldPositions an array of positional indexes of the spec's partition fields indexed by
-   *                                       position in the normalized partition type
+   *                                       position in the normalized partition schema
    * @return the normalized partition data
    */
   private static PartitionData normalizePartition(PartitionData originalPartition,
-                                                  Types.StructType normalizedPartitionSchema,
+                                                  Types.StructType normalizedPartitionType,
                                                   Integer[] originalPartitionFieldPositions) {
-    PartitionData normalizedPartition = new PartitionData(normalizedPartitionSchema);
+    PartitionData normalizedPartition = new PartitionData(normalizedPartitionType);
 
-    IntStream.range(0, normalizedPartitionSchema.fields().size()).forEach(normalizedPartitionFieldIndex -> {
+    IntStream.range(0, normalizedPartitionType.fields().size()).forEach(normalizedPartitionFieldIndex -> {
       Integer originalPartitionPosition = originalPartitionFieldPositions[normalizedPartitionFieldIndex];
       if (originalPartitionPosition != null) {
         Object originalPartitionValue = originalPartition.get(originalPartitionPosition);
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -127,11 +127,10 @@ public class PartitionsTable extends BaseMetadataTable {
       originalFieldIdsToPosition.put(originalField.fieldId(), originalPartitionIndex);
       originalPartitionIndex++;
     }
-    Integer[] result = normalizedPartitionType.fields().stream().map(f -> {
-      Integer originalFieldPosition = originalFieldIdsToPosition.get(f.fieldId());
-      return (originalFieldPosition == null) ? null : originalFieldPosition;
-    }).toArray(Integer[]::new);
-    return result;
+
+    return normalizedPartitionType.fields().stream()
+        .map(f -> originalFieldIdsToPosition.get(f.fieldId()))
+        .toArray(Integer[]::new);
   }
 
   /**
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -95,11 +95,11 @@ public class PartitionsTable extends BaseMetadataTable {
 
   private static Iterable<Partition> partitions(Table table, StaticTableScan scan) {
     CloseableIterable<FileScanTask> tasks = planFiles(scan);
-
-    PartitionMap partitions = new PartitionMap(Partitioning.partitionType(table));
+    Types.StructType partitionType = Partitioning.partitionType(table);
+    PartitionMap partitions = new PartitionMap(partitionType);
     for (FileScanTask task : tasks) {
       PartitionData original = (PartitionData) task.file().partition();
-      PartitionData normalized = normalizePartition(original, Partitioning.partitionType(table));
+      PartitionData normalized = normalizePartition(original, partitionType);
       partitions.get(normalized).update(task.file());
     }
     return partitions.all();
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -24,7 +24,6 @@ import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.expressions.Projections;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
-import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.StructLikeWrapper;
@@ -106,9 +105,7 @@ public class PartitionsTable extends BaseMetadataTable {
     return partitions.all();
   }
 
-  private static PartitionData normalizePartition(PartitionData partition, Types.StructType newSchema) {
-    Preconditions.checkArgument(partition.getPartitionType().fields().size() == partition.size(),
-        "Partition values must match size");
+  private static PartitionData normalizePartition(PartitionData partition, Types.StructType normalizedPartitionSchema) {
     Map<Integer, Object> fieldIdToValues = Maps.newHashMap();
     int originalPartitionIndex = 0;
     for (Types.NestedField f : partition.getPartitionType().fields()) {
@@ -116,14 +113,14 @@ public class PartitionsTable extends BaseMetadataTable {
       originalPartitionIndex++;
     }
 
-    PartitionData result = new PartitionData(newSchema);
+    PartitionData normalizedPartition = new PartitionData(normalizedPartitionSchema);
 
-    int finalPartitionIndex = 0;
-    for (Types.NestedField f : newSchema.fields()) {
-      result.set(finalPartitionIndex, fieldIdToValues.get(f.fieldId()));
-      finalPartitionIndex++;
+    int normalizedPartitionIndex = 0;
+    for (Types.NestedField f : normalizedPartitionSchema.fields()) {
+      normalizedPartition.set(normalizedPartitionIndex, fieldIdToValues.get(f.fieldId()));
+      normalizedPartitionIndex++;
     }
-    return result;
+    return normalizedPartition;
   }
 
   @VisibleForTesting
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -22,12 +22,14 @@ package org.apache.iceberg;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import java.util.Map;
+import java.util.stream.IntStream;
 import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.expressions.Projections;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.types.Types;
+import org.apache.iceberg.util.Pair;
 
 /**
  * A {@link Table} implementation that exposes a table's partitions as rows.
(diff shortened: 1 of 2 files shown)


szehon-ho:
FYI @rdblue @aokolnychyi @RussellSpitzer @szlta if can you help review, thanks

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +94,38 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
+    Preconditions.checkArgument(partition.getPartitionType().fields().size() == partition.size(),
+        "Partition values must match size");
+    Map<Integer, Object> fieldIdToValues = Maps.newHashMap();
+    int originalPartitionIndex = 0;
+    for (Types.NestedField f : partition.getPartitionType().fields()) {
+      fieldIdToValues.put(f.fieldId(), partition.get(originalPartitionIndex));
+      originalPartitionIndex++;
+    }
+
+    PartitionData result = new PartitionData(newSchema);
```
RussellSpitzer, line 119: result -> normalizedPartition?

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +94,38 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
     for (FileScanTask task : tasks) {
-      partitions.get(task.file().partition()).update(task.file());
+      PartitionData original = (PartitionData) task.file().partition();
+      PartitionData normalized = normalizePartition(original, Partitioning.partitionType(table));
+      partitions.get(normalized).update(task.file());
     }
     return partitions.all();
   }
 
+  private static PartitionData normalizePartition(PartitionData partition, Types.StructType newSchema) {
```
RussellSpitzer, line 109: newSchema -> normalizedPartitionSchema?

Maybe that's too long?

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +94,38 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
+    Map<Integer, Object> fieldIdToValues = Maps.newHashMap();
+    int originalPartitionIndex = 0;
+    for (Types.NestedField f : partition.getPartitionType().fields()) {
+      fieldIdToValues.put(f.fieldId(), partition.get(originalPartitionIndex));
+      originalPartitionIndex++;
+    }
+
+    PartitionData result = new PartitionData(newSchema);
+
+    int finalPartitionIndex = 0;
```
RussellSpitzer, line 121: normalizedPartitionIndex (or normalizedIndex)

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -261,6 +262,88 @@ public void testEntriesMetadataTable() throws ParseException {
     }
   }
 
+  @Test
+  public void testPartitionMetadataTable() throws ParseException {
```
RussellSpitzer, line 266: As one more thing to test, can you check reordering partition transforms? 
Going from data, category -> category,data

I think you have this covered but I want to make sure we have a test in there since I think this is a pretty common usecase

RussellSpitzer (commented):
One suggested additional test, and I'm not quite sure about the Precondition check but other than that looks good

core/src/main/java/org/apache/iceberg/PartitionsTable.java
szehon-ho: Good point, renamed relevant fields in this method.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
szehon-ho: Added the test.  Unfortunately it's a bit weird until https://github.com/apache/iceberg/issues/4292. completely fixed (I believe @szlta is taking a look at the remaining point there which would fix this issue).

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +93,36 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
-  private static Iterable<Partition> partitions(StaticTableScan scan) {
+  private static Iterable<Partition> partitions(Table table, StaticTableScan scan) {
     CloseableIterable<FileScanTask> tasks = planFiles(scan);
 
-    PartitionMap partitions = new PartitionMap(scan.table().spec().partitionType());
+    PartitionMap partitions = new PartitionMap(Partitioning.partitionType(table));
     for (FileScanTask task : tasks) {
-      partitions.get(task.file().partition()).update(task.file());
+      PartitionData original = (PartitionData) task.file().partition();
+      PartitionData normalized = normalizePartition(original, Partitioning.partitionType(table));
```
szlta, line 102: Should we reuse the result of the Partitioning.partitionType(table) invocation at line 99 rather than calling it in every iteration?

szehon-ho: Yea good point

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +93,36 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
-      partitions.get(task.file().partition()).update(task.file());
+      PartitionData original = (PartitionData) task.file().partition();
+      PartitionData normalized = normalizePartition(original, partitionType);
+      partitions.get(normalized).update(task.file());
     }
     return partitions.all();
   }
 
+  private static PartitionData normalizePartition(PartitionData partition, Types.StructType normalizedPartitionSchema) {
+    Map<Integer, Object> fieldIdToValues = Maps.newHashMap();
```
RussellSpitzer, line 109: One thing we may want to consider here is caching these normalization by spec rather than recomputing the mapping for every partition value. Thinking about this code being run on millions of files

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -261,6 +262,118 @@ public void testEntriesMetadataTable() throws ParseException {
...
+        .commit();
+    sql("REFRESH TABLE %s", tableName);
+
+    sql("INSERT INTO TABLE %s VALUES (1, 'c1', 'd1')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'c2', 'd2')", tableName);
+
+    // Re-added partition fields currently not re-associated: https://github.com/apache/iceberg/issues/4292
+    // In V1, dropped partition fields show separately when field is re-added
+    // In V2, re-added field currently conflicts with its deleted form
+    if (formatVersion == 1) {
```
RussellSpitzer, line 360: Putting this in a new test with the "Assume(formatVersion ==1)" would be a bit cleaner.

RussellSpitzer (approved):
Two suggestions but I think this is good as is, 

1. Spitting up the test so that we have individual tests for the different alterations
2. Caching the field mapping by partition spec

szehon-ho:
@RussellSpitzer added additional cache of field positional mapping per spec-id as suggested , it's changed a bit, let me know if you can take another look

core/src/main/java/org/apache/iceberg/PartitionsTable.java
szehon-ho: Discussed offline, added cache of positional mappings to the final partition type, for each spec-id

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +96,77 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
 
-  private static Iterable<Partition> partitions(StaticTableScan scan) {
+  private static Iterable<Partition> partitions(Table table, StaticTableScan scan) {
     CloseableIterable<FileScanTask> tasks = planFiles(scan);
+    Types.StructType normalizedPartitionType = Partitioning.partitionType(table);
+    PartitionMap partitions = new PartitionMap();
+
+    LoadingCache<Integer, Integer[]> originalPartitionFieldPositionsBySpec = Caffeine.newBuilder().build(specId ->
+        originalPositions(table, specId, normalizedPartitionType));
+    LoadingCache<Pair<PartitionData, Integer>, PartitionData> normalizedPartitions = Caffeine.newBuilder().build(
```
RussellSpitzer, line 106: This I don't think is super important to cache since the getters and setters are pretty fast and you have the integer mapping already. So I probably would drop this.

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +96,77 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
     return StaticDataTask.Row.of(partition.key, partition.recordCount, partition.fileCount, partition.specId);
   }
 
-  private static Iterable<Partition> partitions(StaticTableScan scan) {
+  private static Iterable<Partition> partitions(Table table, StaticTableScan scan) {
     CloseableIterable<FileScanTask> tasks = planFiles(scan);
+    Types.StructType normalizedPartitionType = Partitioning.partitionType(table);
+    PartitionMap partitions = new PartitionMap();
+
+    LoadingCache<Integer, Integer[]> originalPartitionFieldPositionsBySpec = Caffeine.newBuilder().build(specId ->
```
RussellSpitzer, line 104: I think the number of partition specs should be very small so I would probably just use a 
Maps.newHashMap()

Then .computeIfAbsent

Not that I think there is somethign wrong with Caffine, just seems a bit heavy weight to me in this use case

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +96,77 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
+      PartitionData normalized = normalizedPartitions.get(Pair.of(original, specId));
+      partitions.get(normalized).update(task.file());
     }
     return partitions.all();
   }
 
+  /**
+   * Returns an array of original partition field positions, indexed by normalized partition field positions
+   */
+  private static Integer[] originalPositions(Table table,
```
RussellSpitzer, line 121: I think if we make this non-static we can just rely on the table coming from this class

RussellSpitzer: I see this is all called from a static context so maybe we can't change that ...

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +96,77 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
+    LoadingCache<Integer, Integer[]> originalPartitionFieldPositionsBySpec = Caffeine.newBuilder().build(specId ->
+        originalPositions(table, specId, normalizedPartitionType));
+    LoadingCache<Pair<PartitionData, Integer>, PartitionData> normalizedPartitions = Caffeine.newBuilder().build(
+        pData -> normalizePartition(pData, normalizedPartitionType, originalPartitionFieldPositionsBySpec));
 
-    PartitionMap partitions = new PartitionMap(scan.table().spec().partitionType());
     for (FileScanTask task : tasks) {
-      partitions.get(task.file().partition()).update(task.file());
+      PartitionData original = (PartitionData) task.file().partition();
+      int specId = task.spec().specId();
```
RussellSpitzer, line 111: I was thinking this would look something more like
```
positionMapping = originalPositions.computeIfAbsent(specId, specId -> originalPositions())
make normalizedPartitionData
for (originalIndex = 0; originalIndex < original.length; originalIndex ++) {
   normalizedPartitionData.put(positionMapping[originalIndex], original.get(originalIndex))
}
```

This could be a separate function like we have now but I don't know if it's that bad if we just inline it here

szehon-ho: Done, yea it's a bit easier to read in separate function so I kept it outside.

core/src/main/java/org/apache/iceberg/PartitionsTable.java
szehon-ho: It was saving the construction of the new partitionData object, at the cost of memory. but yea it might not be worth it, dropped it.

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -93,16 +93,71 @@ private static StaticDataTask.Row convertPartition(Partition partition) {
...
+   * @param originalPartitionFieldPositions an array of positional indexes of the spec's partition fields indexed by
+   *                                       position in the normalized partition schema
+   * @return the normalized partition data
+   */
+  private static PartitionData normalizePartition(PartitionData originalPartition,
+                                                  Types.StructType normalizedPartitionType,
+                                                  Integer[] originalPartitionFieldPositions) {
+    PartitionData normalizedPartition = new PartitionData(normalizedPartitionType);
+
+    IntStream.range(0, normalizedPartitionType.fields().size()).forEach(normalizedPartitionFieldIndex -> {
```
RussellSpitzer, line 150: I would just use a normal for loop here, the Intstream I feel like is more verbose that is useful. I wish there was a zipWithIndex or Enumerate but java is still not ready for concise versions of those :(

core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -140,20 +195,13 @@ private class PartitionsScan extends StaticTableScan {
...
-      this.reused = StructLikeWrapper.forType(type);
-    }
+    private final Map<StructLike, Partition> partitions = Maps.newHashMap();
 
     Partition get(StructLike key) {
-      Partition partition = partitions.get(reused.set(key));
+      Partition partition = partitions.get(key);
       if (partition == null) {
         partition = new Partition(key);
-        partitions.put(StructLikeWrapper.forType(type).set(key), partition);
```
RussellSpitzer, line 156: I'm a bit confused by this change, I believe the issue here is that StructType does not have a well defined hashFunction (since implementations can do whatever they like) which is why we use the Wrapper to make sure we have a valid hash.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -212,7 +214,6 @@ public void testFilesMetadataTableFilter() throws ParseException {
     sql("CREATE TABLE %s (id bigint NOT NULL, category string, data string) USING iceberg " +
         "TBLPROPERTIES ('commit.manifest-merge.enabled' 'false')", tableName);
     initTable();
-
```
RussellSpitzer, line 215: Unnecessary white space change here

RussellSpitzer: I think if we want to make this change we need to do it in all the tests and currently this is missing format changes for testFilesMetadataTable and testWithUnknownTransfer. Probably fine to just keep it as is and match in the new tests

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
```diff
@@ -378,10 +378,177 @@ public void testEntriesMetadataTable() throws ParseException {
...
+    sql("INSERT INTO TABLE %s VALUES (1, 'c1', 'd1')", tableName);
+    sql("INSERT INTO TABLE %s VALUES (2, 'c2', 'd2')", tableName);
+
+    // verify the metadata tables while the current spec is still unpartitioned
+    Dataset<Row> df = loadMetadataTable(PARTITIONS);
+    Assert.assertTrue("Partition must be skipped", df.schema().getFieldIndex("partition").isEmpty());
+
+    Table table = validationCatalog.loadTable(tableIdent);
+
+    table.updateSpec()
```
RussellSpitzer, line 395: I think you could just do this in SparkSQL if you like and skip the refresh, but this is fine too.

szehon-ho: Will just keep it for now then, matches existing tests

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestMetadataTablesWithPartitionEvolution.java
szehon-ho: Removed whitespace

core/src/main/java/org/apache/iceberg/PartitionsTable.java
szehon-ho: Changed to map of PartitionData (I feel, it should have been that way in the beginning)

RussellSpitzer (approved):
LGTM! Thanks for baring with my many comments. Feel free to merge whenever you feel ready.

szehon-ho:
No problem, thanks for all the suggestions !

PR #3614, opened 2021-11-27
Flink: Fix ALTER for tables with primary key fields

Initial-neko:
fix #3521 

it makes flink table changes its properties via not comparing PK's name because its random name
using new schema api because old schema method is Deprecated

PR #4561, opened 2022-04-14
Flink: support modify table properties with primary key.

wuwenchi:
Because we have iceberg table with primary key, and we want to alter table properties.
This PR supports `ALTER TABLE ... SET ...` and `ALTER TABLE ... RESET ...` to modify a table properties which has primary key.   #4562

### DIFF
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
@@ -414,8 +414,8 @@ public class FlinkCatalog extends AbstractCatalog {
     // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
     // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
     if (!equalsPrimary &&
-            ts1.getTableColumns().equals(ts2.getTableColumns()) &&
-            ts1.getWatermarkSpecs().equals(ts2.getWatermarkSpecs())) {
+          ts1.getTableColumns().equals(ts2.getTableColumns()) &&
+          ts1.getWatermarkSpecs().equals(ts2.getWatermarkSpecs())) {
       throw new UnsupportedOperationException("Altering schema is not supported yet.");
     }
 
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
@@ -404,18 +404,18 @@ public class FlinkCatalog extends AbstractCatalog {
     TableSchema ts1 = ct1.getSchema();
     TableSchema ts2 = ct2.getSchema();
     boolean equalsPrimary = false;
+
     if (ts1.getPrimaryKey().isPresent() && ts2.getPrimaryKey().isPresent()) {
-      equalsPrimary = ts1.getPrimaryKey().get().getType().equals(ts2.getPrimaryKey().get().getType()) &&
-              ts1.getPrimaryKey().get().getColumns().equals(ts2.getPrimaryKey().get().getColumns());
+      equalsPrimary =
+          Objects.equals(ts1.getPrimaryKey().get().getType(), ts2.getPrimaryKey().get().getType()) &&
+          Objects.equals(ts1.getPrimaryKey().get().getColumns(), ts2.getPrimaryKey().get().getColumns());
     } else if (!ts1.getPrimaryKey().isPresent() && !ts2.getPrimaryKey().isPresent()) {
       equalsPrimary = true;
     }
 
-    // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
-    // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
-    if (!equalsPrimary &&
-          ts1.getTableColumns().equals(ts2.getTableColumns()) &&
-          ts1.getWatermarkSpecs().equals(ts2.getWatermarkSpecs())) {
+    if (!(Objects.equals(ts1.getTableColumns(), ts2.getTableColumns()) &&
+          Objects.equals(ts1.getWatermarkSpecs(), ts2.getWatermarkSpecs()) &&
+          equalsPrimary)) {
       throw new UnsupportedOperationException("Altering schema is not supported yet.");
     }
 
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
@@ -405,17 +405,17 @@ public class FlinkCatalog extends AbstractCatalog {
     TableSchema ts2 = ct2.getSchema();
     boolean equalsPrimary = false;
     if (ts1.getPrimaryKey().isPresent() && ts2.getPrimaryKey().isPresent()) {
-      equalsPrimary = ts1.getPrimaryKey().get().getType().equals(ts2.getPrimaryKey().get().getType())
-              && ts1.getPrimaryKey().get().getColumns().equals(ts2.getPrimaryKey().get().getColumns());
+      equalsPrimary = ts1.getPrimaryKey().get().getType().equals(ts2.getPrimaryKey().get().getType()) &&
+              ts1.getPrimaryKey().get().getColumns().equals(ts2.getPrimaryKey().get().getColumns());
     } else if (!ts1.getPrimaryKey().isPresent() && !ts2.getPrimaryKey().isPresent()) {
       equalsPrimary = true;
     }
 
     // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
     // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
-    if (!equalsPrimary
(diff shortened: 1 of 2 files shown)


flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -400,6 +400,30 @@ void createIcebergTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
...
+              ts1.getPrimaryKey().get().getColumns().equals(ts2.getPrimaryKey().get().getColumns());
+    } else if (!ts1.getPrimaryKey().isPresent() && !ts2.getPrimaryKey().isPresent()) {
+      equalsPrimary = true;
+    }
+
+    // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
+    // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
+    if (!equalsPrimary &&
+            ts1.getTableColumns().equals(ts2.getTableColumns()) &&
+            ts1.getWatermarkSpecs().equals(ts2.getWatermarkSpecs())) {
```
rdblue, line 418: Please fix indentation. This should be 2 indents, not 4.

rdblue:
@wuwenchi, can you please credit the author of the unit test?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: @rdblue Thanks for your review, and I will modify it with 4

wuwenchi:
> @wuwenchi, can you please credit the author of the unit test?

@openinx Could you please review this PR?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
rdblue: @wuwenchi, could you please fix the other cases where lines are over-indented?

rdblue:
@wuwenchi, please credit the original author of the test before we can merge this.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -400,6 +400,30 @@ void createIcebergTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
...
+  private static void validateTableSchemaAndPartition(CatalogTable ct1, CatalogTable ct2) {
+    TableSchema ts1 = ct1.getSchema();
+    TableSchema ts2 = ct2.getSchema();
+    boolean equalsPrimary = false;
+    if (ts1.getPrimaryKey().isPresent() && ts2.getPrimaryKey().isPresent()) {
+      equalsPrimary = ts1.getPrimaryKey().get().getType().equals(ts2.getPrimaryKey().get().getType()) &&
+              ts1.getPrimaryKey().get().getColumns().equals(ts2.getPrimaryKey().get().getColumns());
+    } else if (!ts1.getPrimaryKey().isPresent() && !ts2.getPrimaryKey().isPresent()) {
+      equalsPrimary = true;
+    }
```
yittg, line 412: how about wrapping a function `primaryKeyEqualsIgnoreName`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -400,6 +400,30 @@ void createIcebergTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
...
+
+    // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
+    // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
+    if (!equalsPrimary &&
+          ts1.getTableColumns().equals(ts2.getTableColumns()) &&
+          ts1.getWatermarkSpecs().equals(ts2.getWatermarkSpecs())) {
+      throw new UnsupportedOperationException("Altering schema is not supported yet.");
```
yittg, line 419: it should be `if (!(equalsPrimary && columns && watermarks)) {}`?

yittg: And would you mind validating it by `Objects.equals` as flink does originally?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
openinx: Why not just use the one line sentence to check the equality ? 

```
boolean equalsPrimary = Objects.equals(ts1.getPrimaryKey(), ts2.getPrimaryKey());
```

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -400,6 +400,30 @@ void createIcebergTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
...
+    boolean equalsPrimary = false;
+    if (ts1.getPrimaryKey().isPresent() && ts2.getPrimaryKey().isPresent()) {
+      equalsPrimary = ts1.getPrimaryKey().get().getType().equals(ts2.getPrimaryKey().get().getType()) &&
+              ts1.getPrimaryKey().get().getColumns().equals(ts2.getPrimaryKey().get().getColumns());
+    } else if (!ts1.getPrimaryKey().isPresent() && !ts2.getPrimaryKey().isPresent()) {
+      equalsPrimary = true;
+    }
+
+    // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
+    // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
```
openinx, line 415: We've added the duplicated comments in two places, right ?

openinx: Let's just remove the duplicate comments for line447~448.

openinx:
> @wuwenchi, please credit the original author of the test before we can merge this.

@rdblue , this unit test was added by @wuwenchi for addressing the table properties changes for *primary key* tables,  in theory the  original author of it is @wuwenchi .  So I think we don't need to credit others, right ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: @openinx   
Primarykey is a `UniqueConstraint` , so if use `boolean equalsPrimary = Objects.equals(ts1.getPrimaryKey(), ts2.getPrimaryKey());`, it will call `super.equals(o)`. Then
in `AbstractConstraint` 's equals function, it will compare `name` properity which generated by `TableSchema`'s `primaryKey`:
```
public Builder primaryKey(String... columns) {
            return primaryKey(UUID.randomUUID().toString(), columns);
        }
```
So, we should not compare the name attribute, and I only compare the type and columns from primarykeys.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: @yittg yes, I will fix it.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
openinx: Okay,  that sounds reasonable. 

Then I think we can use `Objects.equals` to check the equality for both `type` and `columns` fields.

rdblue:
@openinx, the test looks identical to the [test](https://github.com/apache/iceberg/pull/3614/files#diff-2c8ac6c96a4608e31ba079ccb7cd04dcd50cd9dc57c1c81eeeb5bc75c151c275R322) in #3614 from a different author.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
yittg: @wuwenchi  Would you mind adding some test cases for validating the expected exception?

wuwenchi:
@rdblue   Sorry, I didn't notice this pr #3614  @Initial-neko before...

I want to add iceberg support for watermark and computed columns, and save these informations to table properties.
I found this problem when I alter a table with a primary key.

 In fact, I write the test according to original test (
https://github.com/apache/iceberg/blob/master/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkCatalogTable.java#L296-L318 ) which written by @openinx , and I only add `primary key` in DDL sql.

openinx (approved):
Looks good to me overall !  Thanks @wuwenchi for the contribution !

openinx:
I think it's reasonable to add the new unit test based on the original unit tests. So I'm okay to skip the credit for the old PR.  Thanks all for the reviewing !

rdblue:
@wuwenchi, thanks for clarifying how the test was written! Since this was mostly a copy of an existing test then I can see why it looks so similar to the one in the other PR.

PR #4566, opened 2022-04-14
Core: Refactor metadata tables and scans

aokolnychyi:
This PR refactors metadata tables and scans. The project has evolved since the initial implementation was done and I think that metadata tables and scans can be cleaned up a little bit.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
@@ -66,7 +66,7 @@ public class AllDataFilesTable extends BaseFilesTable {
 
     @Override
     protected CloseableIterable<ManifestFile> manifests() {
-      return allManifests(Snapshot::dataManifests);
+      return reachableManifests(Snapshot::dataManifests);
     }
   }
 }
--- a/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
@@ -19,12 +19,7 @@
 
 package org.apache.iceberg;
 
-import java.io.IOException;
-import org.apache.iceberg.exceptions.RuntimeIOException;
 import org.apache.iceberg.io.CloseableIterable;
-import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
-import org.apache.iceberg.relocated.com.google.common.collect.Sets;
-import org.apache.iceberg.util.ParallelIterable;
 
 /**
  * A {@link Table} implementation that exposes a table's valid data files as rows.
@@ -53,47 +48,25 @@ public class AllDataFilesTable extends BaseFilesTable {
     return MetadataTableType.ALL_DATA_FILES;
   }
 
-  public static class AllDataFilesTableScan extends BaseFilesTableScan {
+  public static class AllDataFilesTableScan extends BaseAllFilesTableScan {
 
-    AllDataFilesTableScan(TableOperations ops, Table table, Schema fileSchema) {
-      super(ops, table, fileSchema, MetadataTableType.ALL_DATA_FILES);
+    AllDataFilesTableScan(TableOperations ops, Table table, Schema schema) {
+      super(ops, table, schema, MetadataTableType.ALL_DATA_FILES);
     }
 
-    private AllDataFilesTableScan(TableOperations ops, Table table, Schema schema, Schema fileSchema,
+    private AllDataFilesTableScan(TableOperations ops, Table table, Schema schema,
                                   TableScanContext context) {
-      super(ops, table, schema, fileSchema, context, MetadataTableType.ALL_DATA_FILES);
+      super(ops, table, schema, MetadataTableType.ALL_DATA_FILES, context);
     }
 
     @Override
     protected TableScan newRefinedScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
-      return new AllDataFilesTableScan(ops, table, schema, fileSchema(), context);
-    }
-
-    @Override
-    public TableScan useSnapshot(long scanSnapshotId) {
-      throw new UnsupportedOperationException("Cannot select snapshot: all_data_files is for all snapshots");
-    }
-
-    @Override
-    public TableScan asOfTime(long timestampMillis) {
-      throw new UnsupportedOperationException("Cannot select snapshot: all_data_files is for all snapshots");
-    }
-
-    @Override
-    public CloseableIterable<FileScanTask> planFiles() {
-      return super.planFilesAllSnapshots();
+      return new AllDataFilesTableScan(ops, table, schema, context);
     }
 
     @Override
     protected CloseableIterable<ManifestFile> manifests() {
-      try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
-          Iterables.transform(table().snapshots(),
-              snapshot -> (Iterable<ManifestFile>) () -> snapshot.dataManifests().iterator()),
-          context().planExecutor())) {
-        return CloseableIterable.withNoopClose(Sets.newHashSet(iterable));
-      } catch (IOException e) {
-        throw new RuntimeIOException(e, "Failed to close parallel iterable");
-      }
+      return allManifests(Snapshot::dataManifests);
     }
   }
 }
--- a/core/src/main/java/org/apache/iceberg/AllEntriesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllEntriesTable.java
@@ -84,7 +84,7 @@ public class AllEntriesTable extends BaseMetadataTable {
 
     @Override
     protected CloseableIterable<FileScanTask> doPlanFiles() {
-      CloseableIterable<ManifestFile> manifests = allManifests(Snapshot::allManifests);
+      CloseableIterable<ManifestFile> manifests = reachableManifests(Snapshot::allManifests);
 
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
--- a/core/src/main/java/org/apache/iceberg/AllEntriesTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllEntriesTable.java
@@ -19,19 +19,14 @@
 
 package org.apache.iceberg;
 
-import java.io.IOException;
-import java.util.List;
-import java.util.concurrent.ExecutorService;
-import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.ManifestEntriesTable.ManifestReadTask;
 import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.io.CloseableIterable;
-import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types.StructType;
-import org.apache.iceberg.util.ParallelIterable;
 
 /**
  * A {@link Table} implementation that exposes a table's manifest entries as rows, for both delete and data files.
@@ -74,11 +69,11 @@ public class AllEntriesTable extends BaseMetadataTable {
   private static class Scan extends BaseAllMetadataTableScan {
 
     Scan(TableOperations ops, Table table, Schema schema) {
-      super(ops, table, schema);
+      super(ops, table, schema, MetadataTableType.ALL_ENTRIES);
     }
 
     private Scan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
-      super(ops, table, schema, context);
+      super(ops, table, schema, MetadataTableType.ALL_ENTRIES, context);
     }
 
     @Override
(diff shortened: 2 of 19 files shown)


core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -53,47 +48,25 @@ MetadataTableType metadataTableType() {
     return MetadataTableType.ALL_DATA_FILES;
   }
 
-  public static class AllDataFilesTableScan extends BaseFilesTableScan {
+  public static class AllDataFilesTableScan extends BaseAllFilesTableScan {
```
aokolnychyi, line 51: I added `BaseAllFilesTableScan` for all aggregated file tables as I think it is slightly cleaner. That way, we can properly override the planning method in the parent class and consume in all children.

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -53,47 +48,25 @@ MetadataTableType metadataTableType() {
...
 
-    AllDataFilesTableScan(TableOperations ops, Table table, Schema fileSchema) {
-      super(ops, table, fileSchema, MetadataTableType.ALL_DATA_FILES);
+    AllDataFilesTableScan(TableOperations ops, Table table, Schema schema) {
```
aokolnychyi, line 53: I took a closer look and it seems it is not really `fileSchema`. It is more the metadata table schema before column pruning. That's the purpose of that variable in `BaseTableScan`.

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -53,47 +48,25 @@ MetadataTableType metadataTableType() {
...
+      super(ops, table, schema, MetadataTableType.ALL_DATA_FILES, context);
     }
 
     @Override
-    protected TableScan newRefinedScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
-      return new AllDataFilesTableScan(ops, table, schema, fileSchema(), context);
-    }
-
-    @Override
-    public TableScan useSnapshot(long scanSnapshotId) {
```
aokolnychyi, line 73: These got moved to the parent as none of the aggregated metadata tables support selecting a snapshot.

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -53,47 +48,25 @@ MetadataTableType metadataTableType() {
...
-    @Override
-    public CloseableIterable<FileScanTask> planFiles() {
-      return super.planFilesAllSnapshots();
+    protected TableScan newRefinedScan(Schema schema, TableScanContext context) {
+      return new AllDataFilesTableScan(ops(), table(), schema, context);
     }
 
     @Override
     protected CloseableIterable<ManifestFile> manifests() {
-      try (CloseableIterable<ManifestFile> iterable = new ParallelIterable<>(
```
aokolnychyi, line 89: This logic was is in a few places. I moved it to the parent class and share in all places.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -85,58 +85,41 @@ MetadataTableType metadataTableType() {
...
-        boolean ignoreResiduals, boolean caseSensitive, boolean colStats) {
+    protected CloseableIterable<FileScanTask> doPlanFiles() {
+      FileIO io = table().io();
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
       Map<Integer, PartitionSpec> specs = Maps.newHashMap(table().specs());
+      Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : filter();
+      ResidualEvaluator residuals = ResidualEvaluator.unpartitioned(filter);
 
-      return CloseableIterable.withNoopClose(Iterables.transform(ops.current().snapshots(), snap -> {
```
aokolnychyi, line 125: I checked and `table()` returns the main table in all cases so `table().snapshots()` is equivalent to accessing table operations but shorter and uses a higher level API.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -85,58 +85,41 @@ MetadataTableType metadataTableType() {
...
-    @Override
-    protected CloseableIterable<FileScanTask> planFiles(
-        TableOperations ops, Snapshot snapshot, Expression rowFilter,
-        boolean ignoreResiduals, boolean caseSensitive, boolean colStats) {
+    protected CloseableIterable<FileScanTask> doPlanFiles() {
+      FileIO io = table().io();
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
       Map<Integer, PartitionSpec> specs = Maps.newHashMap(table().specs());
+      Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : filter();
```
aokolnychyi, line 107: Moved it from the loop. Don't think there is a reason to create these instances for each snapshot.

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
```diff
@@ -19,40 +19,56 @@
...
+    super(ops, table, schema, tableType, context);
   }
 
-  /**
-   * Type of scan being performed, such as {@link MetadataTableType#ALL_DATA_FILES} when scanning
-   * a table's {@link org.apache.iceberg.AllDataFilesTable}.
-   * <p>
-   * Used for logging and error messages.
-   */
-  protected abstract String tableType();
```
aokolnychyi, line 40: Moved it to the base metadata table scan as all metadata tables will need this.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -56,72 +56,73 @@ public Schema schema() {
     }
   }
 
-  abstract static class BaseFilesTableScan extends BaseMetadataTableScan {
+  private static CloseableIterable<FileScanTask> planFiles(Table table, CloseableIterable<ManifestFile> manifests,
```
aokolnychyi, line 59: Moved the common planning logic to a static method to share in both scans.

aokolnychyi: I also removed the comment about column pruning as I think it is no longer accurate. We do support column pruning for metadata tables.

core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java
```diff
@@ -19,39 +19,50 @@
...
+   * a table's {@link org.apache.iceberg.AllDataFilesTable}.
+   * <p>
+   * Used for logging and error messages.
+   */
+  protected MetadataTableType tableType() {
+    return tableType;
+  }
+
+  @Override
+  public TableScan appendsBetween(long fromSnapshotId, long toSnapshotId) {
```
aokolnychyi, line 50: None of the metadata tables support incremental scans so we can share this logic.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -83,22 +87,23 @@ protected Collection<String> selectedColumns() {
...
   }
 
   @SuppressWarnings("checkstyle:HiddenField")
-  protected abstract TableScan newRefinedScan(
-      TableOperations ops, Table table, Schema schema, TableScanContext context);
+  protected abstract TableScan newRefinedScan(TableOperations ops, Table table, Schema schema,
+                                              TableScanContext context);
 
-  @SuppressWarnings("checkstyle:HiddenField")
-  protected abstract CloseableIterable<FileScanTask> planFiles(
```
aokolnychyi, line 99: I am not sure the signature of this method makes sense after we introduced the context object. This method passes some arguments but not all so the implementations are not entirely pure. Also, the passed arguments are useless for some metadata tables. That's why I see little value in having so many arguments. Just makes calling this method harder.

aokolnychyi:
cc @kbendick @szehon-ho @karuppayya @RussellSpitzer @flyrain @rdblue

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
```diff
@@ -53,47 +48,25 @@ MetadataTableType metadataTableType() {
...
-  public static class AllDataFilesTableScan extends BaseFilesTableScan {
+  public static class AllDataFilesTableScan extends BaseAllFilesTableScan {
 
-    AllDataFilesTableScan(TableOperations ops, Table table, Schema fileSchema) {
-      super(ops, table, fileSchema, MetadataTableType.ALL_DATA_FILES);
+    AllDataFilesTableScan(TableOperations ops, Table table, Schema schema) {
+      super(ops, table, schema, MetadataTableType.ALL_DATA_FILES);
     }
 
-    private AllDataFilesTableScan(TableOperations ops, Table table, Schema schema, Schema fileSchema,
```
aokolnychyi, line 62: I no longer pass both `schema` and `fileSchema` as those are same and can be accessed by `schema()` from the common parent class.

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
```diff
@@ -19,40 +19,56 @@
...
   @Override
   public CloseableIterable<FileScanTask> planFiles() {
-    return super.planFilesAllSnapshots();
+    LOG.info("Scanning metadata table {} with filter {}.", table(), filter());
+    Listeners.notifyAll(new ScanEvent(table().name(), 0L, filter(), schema()));
+
+    return doPlanFiles();
+  }
+
+  protected CloseableIterable<ManifestFile> allManifests(Function<Snapshot, Iterable<ManifestFile>> toManifests) {
```
szehon-ho, line 64: Name is a bit confusing as 'allManifests' is already used to denote data + delete manifests.  How about 'allSnapshotManifests' or 'allSnapshots' , or something to denote the fact that is explores all snapshots.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
```diff
@@ -56,72 +56,73 @@ public Schema schema() {
...
+                                                           Schema tableSchema, Schema projectedSchema,
+                                                           TableScanContext context) {
+    Expression rowFilter = context.rowFilter();
+    boolean caseSensitive = context.caseSensitive();
+    boolean ignoreResiduals = context.ignoreResiduals();
+
+    LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
```
szehon-ho, line 66: Not related to this change, but my next pr plans to use this logic as well to fix filtering for PartitionsTable.  Maybe at that time will move the evalCache part of this further up to BaseMetadataTables.

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
aokolnychyi: Agreed. Will change.

core/src/main/java/org/apache/iceberg/AllEntriesTable.java
```diff
@@ -88,34 +83,16 @@ protected TableScan newRefinedScan(TableOperations ops, Table table, Schema sche
...
-    @Override
-    protected CloseableIterable<FileScanTask> planFiles(
-        TableOperations ops, Snapshot snapshot, Expression rowFilter,
-        boolean ignoreResiduals, boolean caseSensitive, boolean colStats) {
-      CloseableIterable<ManifestFile> manifests = allManifestFiles(
-          ops.current().snapshots(), context().planExecutor());
       String schemaString = SchemaParser.toJson(schema());
       String specString = PartitionSpecParser.toJson(PartitionSpec.unpartitioned());
-      Expression filter = ignoreResiduals ? Expressions.alwaysTrue() : rowFilter;
+      Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : filter();
```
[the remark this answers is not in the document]
aokolnychyi: I think it may be something related to having `ignoreResiduals()` that acts like a setter. I'll have to check the existing implementation. I used what was already available.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
aokolnychyi: Do you plan to reuse as-is? If so, maybe, I can move it around in this PR.

szehon-ho: Yea no problem, I can look at it later, it's part of the code but not the whole thing as PartitionsTable creates another kind of FileScanTask.

core/src/main/java/org/apache/iceberg/AllEntriesTable.java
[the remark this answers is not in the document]
szehon-ho: Yea I see, never mind if its too big a change.

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
rdblue: I think it was called `fileSchema` because it was the projection of the `DataFile` struct.

core/src/main/java/org/apache/iceberg/AllEntriesTable.java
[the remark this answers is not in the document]
aokolnychyi: Yeah, it is because of a conflicting setter method.

```
@Override
public TableScan ignoreResiduals() {
  return newRefinedScan(ops, table, schema, context.ignoreResiduals(true));
}
```

core/src/main/java/org/apache/iceberg/AllDataFilesTable.java
aokolnychyi: Okay, that's why. I think it still makes sense to rename as I got rid of a separate `fileSchema` var.

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
```diff
@@ -19,40 +19,56 @@
...
   @Override
   public CloseableIterable<FileScanTask> planFiles() {
-    return super.planFilesAllSnapshots();
+    LOG.info("Scanning metadata table {} with filter {}.", table(), filter());
+    Listeners.notifyAll(new ScanEvent(table().name(), 0L, filter(), schema()));
+
+    return doPlanFiles();
+  }
+
+  protected CloseableIterable<ManifestFile> reachableManifests(Function<Snapshot, Iterable<ManifestFile>> toManifests) {
```
aokolnychyi, line 64: @szehon-ho, I went with `reachableManifests` as `allSnapshotsManifests` was too long to stay on one line. I think we use the term `reachableXXX` in a few places that scan through all snapshots.

core/src/main/java/org/apache/iceberg/BaseFilesTable.java
aokolnychyi: Let's take a look at it in the PR that will use it then.

core/src/main/java/org/apache/iceberg/BaseAllMetadataTableScan.java
szehon-ho: That's a good name.

szehon-ho (approved):

core/src/main/java/org/apache/iceberg/DataTableScan.java
```diff
@@ -78,23 +76,23 @@ protected TableScan newRefinedScan(TableOperations ops, Table table, Schema sche
...
+
+    ManifestGroup manifestGroup = new ManifestGroup(table().io(), snapshot.dataManifests(), snapshot.deleteManifests())
+        .caseSensitive(isCaseSensitive())
+        .select(colStats() ? SCAN_WITH_STATS_COLUMNS : SCAN_COLUMNS)
+        .filterData(filter())
+        .specsById(table().specs())
         .ignoreDeleted();
 
-    if (ignoreResiduals) {
+    if (shouldIgnoreResiduals()) {
```
flyrain, line 89: Nit: may rename `shouldIgnoreResiduals()` to `ignoreResiduals()` to be consistent.

aokolnychyi: I cann't because of a conflicting setter method discussed [here](https://github.com/apache/iceberg/pull/4566#discussion_r855626731).

flyrain (approved):
LGTM. Thanks for the PR, @aokolnychyi. I like the part that removes all parameters from `planFiles(...)`.

aokolnychyi:
Thanks for reviewing, @szehon-ho @flyrain @rdblue!

PR #3411, opened 2021-10-28
Fixes read metadata failed after dropped partition for V1 format

ConeyLiu:
This patch fixes two problems:
1. For V1 tables, we use a `VoidTransform` to replace the removed partition field. While the result type of `VoidTransform` same as the type of the field. For example, the result type of `Bucket(2, string_type_field)` is int, while is string for `VoidTransform(string_type_field)`. So we should use the original partition field type(int instead of string) to build the common partitioning type.
2. We should use the matched `PartitionSpec` to convert the `PartitionFiedlSummary` to human string, not the current table spec.

Closes #3374

PR #4569, opened 2022-04-15
[0.13] Core: Fix filter pushdown for metadata tables with evolved specs (#4520)

szehon-ho:
Backport of #4520 to 0.13.x branch, which returns sometimes wrong results for Files metadata table queries with filters, for V2 tables with specific partition spec evolutions involving dropping fields, and IndexOutOfBoundException in other cases where partition spec has changed.

Adapted to old code in the branch.

This works, except some test case fails in querying dropped partition fields in V1 tables because of related but different issue #3411 .  So one option is to backport that one first for the test, though fix is orthogonal.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
@@ -35,7 +35,6 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
  * deserialization.
  */
 abstract class BaseMetadataTable implements Table, HasTableOperations, Serializable {
-  protected static final String PARTITION_FIELD_PREFIX = "partition.";
   private final PartitionSpec spec = PartitionSpec.unpartitioned();
   private final SortOrder sortOrder = SortOrder.unsorted();
   private final TableOperations ops;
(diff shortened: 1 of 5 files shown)


szehon-ho:
@kbendick @rdblue can this backport go in for 0.13.2?

kbendick:
Yes, at first glance I do think they both should be backported.

The release manager is @nastra fyi, but I personally think if people are encountering this and it’s an obvious bug, we backport it.

I’d personally also like to see #3411 backported, as I’ve recently encountered issues on partition summaries using the void transform (which could be unrelated to both as I need to take a closer look): https://github.com/apache/iceberg/issues/4689

But if it’s ready I don’t see why not. Any ideas on a possible timeline for both? I know that’s highly dependent on things you can’t control, but we’ve admittedly delayed 0.13.2 for quite some time.

TLDR - On first pass I would think to include both. But I do have some concerns around holding up the release too much.

I’ll spend time on a proper pass over this tomorrow :)

szehon-ho:
Hi @kbendick yea I requested the backport for #3411 already and  @ConeyLiu did this in : https://github.com/apache/iceberg/pull/4572. 

For this one, I think @rdblue was ok and marked the original issue #4520 for 0.13.2 backport by the tag, so I did the backport by convenience

kbendick (approved):
Looking at the combination of #4520, the other PRs that were backported, and this, this seems good.

Given it's been some time and other things have been added, I'd suggest rebasing. I'd also consider porting the tests for the other Spark versions to this PR, just so we can ensure that they work. But barring that (or any disagreement with that), I'm +1 on this.

Thank you @szehon-ho!

szehon-ho:
Thanks @kbendick for the review, rebased.  Re: Spark versions, the main fix is in core so I think the Spark side is agnostic (I added the Spark 3.2 test mainly as an end-to-end).  

Anyway it's the same fix as on the master branch to #4520, and I can look at adding some tests for other Spark versions over there.

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -52,18 +51,17 @@ protected BaseMetadataTable(TableOperations ops, Table table, String name) {
...
    * @param metadataTableSchema schema of the metadata table
    * @param spec spec on which the metadata table schema is based
-   * @param partitionPrefix prefix to remove from each field in the partition spec
    * @return a spec used to rewrite the metadata table filters to partition filters using an inclusive projection
    */
-  static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec, String partitionPrefix) {
+  static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec) {
     PartitionSpec.Builder identitySpecBuilder = PartitionSpec.builderFor(metadataTableSchema).checkConflicts(false);
-    spec.fields().forEach(pf -> identitySpecBuilder.identity(partitionPrefix + pf.name(), pf.name()));
+    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), pf.name(), "identity"));
```
rdblue, line 64: This looks suspicious to me because it doesn't pass in the field ID as the partition field ID. Shouldn't it pass in both IDs? Otherwise the partition ID will be assigned sequentially, which probably doesn't work.

rdblue: Actually, I thought about this more last night and I think it's fine. This spec is used to produce an unbound predicate. So the only part that matters is that the correct name is passed in.

rdblue (approved):

PR #4572, opened 2022-04-16
[0.13] Core: Fixes read metadata failed after dropped partition transform for V1 format

ConeyLiu:
This backs port #3411 to 0.13.x branch.

This patch fixes two problems:
1. For V1 tables, we use a `VoidTransform` to replace the removed partition field. While the result type of `VoidTransform` same as the type of the field. For example, the result type of `Bucket(2, string_type_field)` is int, while is string for `VoidTransform(string_type_field)`. So we should use the original partition field type(int instead of string) to build the common partitioning type.
2. We should use the matched `PartitionSpec` to convert the `PartitionFiedlSummary` to human string, not the current table spec.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/AllManifestsTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllManifestsTable.java
@@ -21,6 +21,7 @@ package org.apache.iceberg;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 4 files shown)


ConeyLiu:
cc @szehon-ho @rdblue

szehon-ho (approved):
Thanks a lot @ConeyLiu

ConeyLiu:
Thanks @szehon-ho.

PR #4575, opened 2022-04-17
Core: Add context headers to RESTCatalog

rdblue:
This refactors the `RESTCatalog` so that it can be used in engines like Trino, where a catalog is shared between sessions and context is passed to each catalog operation.

`SessionCatalog` is a new interface that is like `Catalog`, but has a `SessionContext` object in each call. This session context can be used to carry identity and credentials, which are passed through to the catalog service. `SessionCatalog` does not extend `Catalog` to avoid bugs where context is accidentally not passed.

`RESTSessionCatalog` is an `SessionCatalog` implementation that comes from `RESTCatalog`, and `RESTCatalog` is now a wrapper around `RESTSessionCatalog`.

In addition, this PR also adds headers to the HTTP methods in `RESTClient`.


rdblue:
An alternative to this that we should also consider is adding `SessionContext` to catalog calls. That would add an interface like this:

```java
public interface SessionCatalog extends Catalog {
  final class SessionContext {
    private final String sessionId;
    private final String identity;
    private final String credential;
    private final Map<String, String> sessionConfig;

    public SessionContext(String sessionId, String identity, String credential, Map<String, String> sessionConfig) {
      this.sessionId = sessionId;
      this.identity = identity;
      this.credential = credential;
      this.sessionConfig = sessionConfig;
    }

    public String sessionId() {
      return sessionId;
    }

    public String identity() {
      return identity;
    }

    public String credential() {
      return credential;
    }

    public Map<String, String> sessionConfig() {
      return sessionConfig;
    }
  }

  void initialize(String name, Map<String, String> properties);

  String name();
  
  Catalog withContext(SessionContext context);

  List<TableIdentifier> listTables(SessionContext context, Namespace namespace);

  TableBuilder buildTable(SessionContext context, TableIdentifier ident, Schema schema);

  Table registerTable(TableIdentifier ident, String metadataFileLocation);

  default boolean tableExists(SessionContext context, TableIdentifier ident) {
    try {
      loadTable(context, ident);
      return true;
    } catch (NoSuchTableException e) {
      return false;
    }
  }

  Table loadTable(SessionContext context, TableIdentifier ident);

  boolean dropTable(SessionContext context, TableIdentifier ident);

  boolean purgeTable(SessionContext context, TableIdentifier ident);

  void renameTable(SessionContext context, TableIdentifier from, TableIdentifier to);

  void invalidateTable(SessionContext context, TableIdentifier ident);
}
```

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,479 @@
...
+import org.apache.iceberg.rest.responses.GetNamespaceResponse;
+import org.apache.iceberg.rest.responses.ListNamespacesResponse;
+import org.apache.iceberg.rest.responses.ListTablesResponse;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
+import org.apache.iceberg.util.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RESTSessionCatalog extends BaseSessionCatalog implements Configurable<Configuration>, Closeable {
```
rdblue, line 69: This class is the original `RESTCatalog`. The logic is unchanged other than to pass headers into the `RESTClient`.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -26,425 +26,217 @@
...
+  private final RESTSessionCatalog sessionCatalog;
+  private final SessionCatalog.SessionContext context;
+  private final Catalog delegate;
+  private final SupportsNamespaces nsDelegate;
+  private String name;
+  private Map<String, String> properties;
 
   public RESTCatalog() {
-    this(new HTTPClientFactory());
+    this(SessionCatalog.SessionContext.createEmpty(), new HTTPClientFactory());
```
rdblue, line 53: This class was moved to be `RESTSessionCatalog`. The implementation here is completely new. The purpose of this class is to wrap a `RESTSessionCatalog` and always pass a global session context.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,479 @@
...
+    String ioImpl = properties.get(CatalogProperties.FILE_IO_IMPL);
+    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), properties, conf);
+  }
+
+  Map<String, String> headers(SessionContext context) {
+    return headers.get(context.sessionId(), id -> {
+      ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+      builder.putAll(baseHeaders);
+
+      if (context.identity() != null) {
```
jackye1995, line 108: I notice here that although we use session Id as the key for the headers map, what we are setting are just headers for identity and credentials. Basically the identity and credentials are the determining factor of the cache content (`Catalog` in the `BaseSessionCatalog`, and header map here). This is very similar to https://github.com/prestodb/presto/blob/master/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergResourceFactory.java#L88.

That means if those information do not change, the same cached content could be shared across sessions. For example, here if 2 sessions have the same identity and credential, they actually have the same headers. So does it make more sense to use `content.identity() + content.credential()` as the cache key? This could greatly improve the cache hit rate, instead of creating something new for each session. This is especially helpful if the session is short-lived, but the caller credential is shared across sessions.

rdblue: Yes, I agree that this is probably going to change. I'm not sure if `identity` and `credential` are the right things to track here. And in Trino we would probably want to set session ID to one or both of them to have sessions that are not simply single queries. I'm definitely open to suggestions!

My main concern right now was to get people looking at the new catalog interface and make sure we have consensus that it's a good idea. What metadata we track in a session is important, but secondary if we don't end up going this direction.

api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+   * Return the properties for this catalog.
+   *
+   * @return this catalog's config properties
+   */
+  Map<String, String> properties();
+
+  /**
+   * Return all the identifiers under this namespace.
```
RussellSpitzer, line 124: nit: under a given namespace

Since this object doesn't have a specific namespace

api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
```diff
@@ -0,0 +1,323 @@
...
+ * A Catalog API for table and namespace operations that includes session context.
+ */
+public interface SessionCatalog {
+  /**
+   * Context for a session.
+   */
+  final class SessionContext {
+    private final String sessionId;
+    private final String identity;
+    private final String credential;
```
danielcweeks, line 44: We might want to consider making credentials a map (this is pretty common and also how Trino represents credentials).

rdblue: This is really helpful. I'm not at all sure how we want to track credentials and identity, so more ideas here are great.

jackye1995: The credential is also a part of identity, for reference: https://github.com/trinodb/trino/blob/master/core/trino-spi/src/main/java/io/trino/spi/security/Identity.java

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
jackye1995: I see. If we are focusing on the consensus part, I think it is pretty aligned with the [TrinoCatalog](https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/TrinoCatalog.java) interface, there is a converter needed for a Trino [ConnectorSession](https://github.com/trinodb/trino/blob/master/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSession.java) but that is overall straightforward.

The reason I bring up this identity and credential part here is that maybe Iceberg's `SessionContext` can be an interface but not an implementation, and we have `SessionCatalog<T extends SessionContext>`, and `SessionContext` has one important method `cacheKey()` that can be used to define if 2 sessions should use the same underlying catalog, and the cache key can on the extreme end be the session ID to have 1 different catalog for each session, or be a composite of identity and credential values.

In Trino, then we can have `IcebergConnectorSession implements SessionContext` and `TrinoCatalog implements SessionCatalog<IcebergConnectorSession>` as a way to simplify the `TrinoCatalog` class.

core/src/main/java/org/apache/iceberg/catalog/BaseSessionCatalog.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
+
+public abstract class BaseSessionCatalog implements SessionCatalog {
+  private final Cache<String, Catalog> catalogs = Caffeine.newBuilder()
+      .expireAfterAccess(10, TimeUnit.MINUTES)
+      .removalListener((id, catalog, cause) -> {
+        if (catalog instanceof Closeable) {
+          try {
+            ((Closeable) catalog).close();
```
RussellSpitzer, line 44: Does this mean any session is closed after 10 minutes of idle time? or that the session may not be closed but the Catalog will need to be reinitialized again?

rdblue: This closes the catalog, but we don't control anything about the session. We're just caching a catalog instance to use with a session. If we have to create a new one because it timed out, it should be initialized with the session details so nothing would change in the catalog's behavior.

core/src/main/java/org/apache/iceberg/catalog/BaseSessionCatalog.java
```diff
@@ -0,0 +1,170 @@
...
+      .build();
+
+  private String name = null;
+  private Map<String, String> properties = null;
+
+  @Override
+  public void initialize(String catalogName, Map<String, String> props) {
+    this.name = catalogName;
+    this.properties = props;
+  }
```
kbendick, line 59: Should we be inserting the catalog into the `catalogs` cache? I don't see where that is happening.

core/src/main/java/org/apache/iceberg/catalog/BaseSessionCatalog.java
kbendick: Can we add a debug logging statement here? This seems like something that might wind up being hard to debug.

core/src/main/java/org/apache/iceberg/catalog/BaseSessionCatalog.java
rdblue: The cache is creating the catalog if it doesn't exist when `get` is called.

core/src/main/java/org/apache/iceberg/catalog/BaseSessionCatalog.java
rdblue: I've moved this interface and base class into its own PR, #4773.

I also removed this `close` call because it doesn't need to close. The `AsCatalog` adapter doesn't have any resources of its own. It's just a light wrapper to use a session context with the `SessionCatalog`.

rdblue:
I've separated this into #4773 for the session catalog interface and #4772 for adding headers to the `HTTPClient`. I'll follow up to #4773 with an implementation of the new interface for `RESTCatalog`.

rdblue:
This has been fully replaced now that #4830 is posted. I'm going to close it.

Issue #4576, opened 2022-04-18
Read metadata table failed due to illegal character


ConeyLiu:
Reproduce code:
```scala
case class Data(id: Long)
case class Item(id: Long, data: Data)

spark.sql("create table local.db.test_table(id long, data struct<id: long>) using iceberg partitioned by (identity(id))").show()
spark.range(10).map(i => Item(i, Data(i))).writeTo("local.db.test_table").append()
spark.sql("alter table local.db.test_table add partition field identity(data.id)").show()
spark.range(10).map(i => Item(i, Data(i))).writeTo("local.db.test_table").append()
spark.sql("select * from local.db.test_table.files").show()
```

Exception stack:
```
Caused by: org.apache.iceberg.shaded.org.apache.avro.SchemaParseException: Illegal character in: data.id_r1001
  at org.apache.iceberg.shaded.org.apache.avro.Schema.validateName(Schema.java:1561)
  at org.apache.iceberg.shaded.org.apache.avro.Schema.access$400(Schema.java:87)
  at org.apache.iceberg.shaded.org.apache.avro.Schema$Field.<init>(Schema.java:541)
  at org.apache.iceberg.shaded.org.apache.avro.Schema$Field.<init>(Schema.java:580)
  at org.apache.iceberg.avro.BuildAvroProjection.record(BuildAvroProjection.java:111)
  at org.apache.iceberg.avro.BuildAvroProjection.record(BuildAvroProjection.java:43)
  at org.apache.iceberg.avro.AvroCustomOrderSchemaVisitor.visit(AvroCustomOrderSchemaVisitor.java:51)
  at org.apache.iceberg.avro.AvroCustomOrderSchemaVisitor$VisitFuture.get(AvroCustomOrderSchemaVisitor.java:109)
  at org.apache.iceberg.avro.BuildAvroProjection.field(BuildAvroProjection.java:140)
  at org.apache.iceberg.avro.BuildAvroProjection.field(BuildAvroProjection.java:43)
  at org.apache.iceberg.avro.AvroCustomOrderSchemaVisitor$VisitFieldFuture.get(AvroCustomOrderSchemaVisitor.java:124)
  at org.apache.iceberg.relocated.com.google.common.collect.Iterators$6.transform(Iterators.java:826)
  at org.apache.iceberg.relocated.com.google.common.collect.TransformedIterator.next(TransformedIterator.java:52)
[...]

PR #4577, opened 2022-04-18
Fixes read metadata table failed due to illegal character

ConeyLiu:
As #4576 described, we got a `SchemaParseException ` when reading the metadata table. The root cause is that Avro does not support a name containing dot(.).  For example, `Identity(data.id)`. In this case, the partition field named with `data.id` is illegal for Avro, so we need to rewrite it to meet the Avro field name requirements.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
+++ b/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
@@ -107,7 +107,10 @@ class BuildAvroProjection extends AvroCustomOrderSchemaVisitor<Schema, Schema.Fi
             field.name());
         // Create a field that will be defaulted to null. We assign a unique suffix to the field
         // to make sure that even if records in the file have the field it is not projected.
-        String fieldName = AvroSchemaUtil.makeCompatibleName(field.name());
+        String origFieldName = field.name();
+        boolean isValidFieldName = AvroSchemaUtil.validAvroName(origFieldName);
+        String fieldName =
+            isValidFieldName ? origFieldName : AvroSchemaUtil.sanitize(origFieldName);
         Schema.Field newField =
             new Schema.Field(
                 fieldName + "_r" + field.fieldId(),
@@ -115,7 +118,9 @@ class BuildAvroProjection extends AvroCustomOrderSchemaVisitor<Schema, Schema.Fi
                 null,
                 JsonProperties.NULL_VALUE);
         newField.addProp(AvroSchemaUtil.FIELD_ID_PROP, field.fieldId());
-        newField.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, field.name());
+        if (!isValidFieldName) {
+          newField.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, origFieldName);
+        }
         updatedFields.add(newField);
         hasChange = true;
       }
--- a/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
+++ b/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
@@ -107,13 +107,20 @@ class BuildAvroProjection extends AvroCustomOrderSchemaVisitor<Schema, Schema.Fi
             field.name());
         // Create a field that will be defaulted to null. We assign a unique suffix to the field
         // to make sure that even if records in the file have the field it is not projected.
+        String origFieldName = field.name();
+        boolean isValidFieldName = AvroSchemaUtil.validAvroName(origFieldName);
+        String fieldName =
+            isValidFieldName ? origFieldName : AvroSchemaUtil.sanitize(origFieldName);
         Schema.Field newField =
             new Schema.Field(
-                field.name() + "_r" + field.fieldId(),
+                fieldName + "_r" + field.fieldId(),
                 AvroSchemaUtil.toOption(AvroSchemaUtil.convert(field.type())),
                 null,
                 JsonProperties.NULL_VALUE);
         newField.addProp(AvroSchemaUtil.FIELD_ID_PROP, field.fieldId());
+        if (!isValidFieldName) {
+          newField.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, origFieldName);
+        }
         updatedFields.add(newField);
         hasChange = true;
       }
--- a/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
+++ b/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
@@ -107,20 +107,12 @@ class BuildAvroProjection extends AvroCustomOrderSchemaVisitor<Schema, Schema.Fi
             field.name());
         // Create a field that will be defaulted to null. We assign a unique suffix to the field
         // to make sure that even if records in the file have the field it is not projected.
-        String origFieldName = field.name();
-        boolean isValidFieldName = AvroSchemaUtil.validAvroName(origFieldName);
-        String fieldName =
-            isValidFieldName ? origFieldName : AvroSchemaUtil.sanitize(origFieldName);
-        Schema.Field newField =
-            new Schema.Field(
-                fieldName + "_r" + field.fieldId(),
-                AvroSchemaUtil.toOption(AvroSchemaUtil.convert(field.type())),
-                null,
-                JsonProperties.NULL_VALUE);
+        String fieldName = AvroSchemaUtil.makeCompatibleName(field.name());
+        Schema.Field newField = new Schema.Field(
+            fieldName + "_r" + field.fieldId(),
+            AvroSchemaUtil.toOption(AvroSchemaUtil.convert(field.type())), null, JsonProperties.NULL_VALUE);
         newField.addProp(AvroSchemaUtil.FIELD_ID_PROP, field.fieldId());
-        if (!isValidFieldName) {
-          newField.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, origFieldName);
-        }
+        newField.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, field.name());
         updatedFields.add(newField);
         hasChange = true;
       }
--- a/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
+++ b/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
@@ -95,7 +95,8 @@ class BuildAvroProjection extends AvroCustomOrderSchemaVisitor<Schema, Schema.Fi
         hasChange = true;
       }
 
-      Schema.Field avroField = updateMap.get(AvroSchemaUtil.makeCompatibleName(field.name()));
+      String fieldName = AvroSchemaUtil.makeCompatibleName(field.name());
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -106,10 +106,16 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
             "Missing required field: %s", field.name());
         // Create a field that will be defaulted to null. We assign a unique suffix to the field
         // to make sure that even if records in the file have the field it is not projected.
+        String origFieldName = field.name();
+        boolean isValidFieldName = AvroSchemaUtil.validAvroName(origFieldName);
+        String fieldName =  isValidFieldName ? origFieldName : AvroSchemaUtil.sanitize(origFieldName);
```
chenjunjiedada, line 111: extra space after '='

chenjunjiedada: Also, I think isValidFieldName can be inline.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -106,10 +106,16 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
             "Missing required field: %s", field.name());
         // Create a field that will be defaulted to null. We assign a unique suffix to the field
         // to make sure that even if records in the file have the field it is not projected.
+        String origFieldName = field.name();
```
ConeyLiu, line 109: This problem only happens on metadata table, because we use a common partition schema for metadata table. For example:
```scala
// create table
spark.sql("create table local.db.test_table(id long, data struct<id: long>) using iceberg partitioned by (identity(id))").show()
// manifest 0
spark.range(10).map(i => Item(i, Data(i))).writeTo("local.db.test_table").append()
// update table to add a partiition field on the nested field
spark.sql("alter table local.db.test_table add partition field identity(data.id)").show()
// manifest 1
spark.range(10).map(i => Item(i, Data(i))).writeTo("local.db.test_table").append()
spark.sql("select * from local.db.test_table.files").show()
```
The common partition schema is `(Identity(id), Identity(data.id))` when we used the common schema projecting on the manifest 0, we could not find the field(data.id) from the Avro schema which read from the manifest file. Then we need to create a new Avro field from the missing one which is why we reached here.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
ConeyLiu: Updated.

ConeyLiu:
Thanks, @chenjunjiedada for the review. And cc @rdblue @szehon-ho @aokolnychyi @RussellSpitzer, could you help to review this? Thanks in advance.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -106,10 +106,16 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
...
+        boolean isValidFieldName = AvroSchemaUtil.validAvroName(origFieldName);
+        String fieldName = isValidFieldName ? origFieldName : AvroSchemaUtil.sanitize(origFieldName);
```
RussellSpitzer, line 111: Is there an issue with us just always calling sanitize?

chenjunjiedada: Aha, +1 to just use `AvroSchemaUtil.sanitize(origFieldName)`.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -106,10 +106,16 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
...
         Schema.Field newField = new Schema.Field(
-            field.name() + "_r" + field.fieldId(),
+            fieldName + "_r" + field.fieldId(),
             AvroSchemaUtil.toOption(AvroSchemaUtil.convert(field.type())), null, JsonProperties.NULL_VALUE);
         newField.addProp(AvroSchemaUtil.FIELD_ID_PROP, field.fieldId());
+        if (!isValidFieldName) {
```
ConeyLiu, line 116: @RussellSpitzer @chenjunjiedada We need to determine whether the original field name prop needs to be set. This same as `BuildAvroProjection`.

chenjunjiedada: We set according to comparing target `fieldName` and `originFiledName`, right? Also, can we always set the property to `originFieldName`?

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
ConeyLiu: Updated to use `AvroSchemaUtil.makeCompatibleName`.

ConeyLiu: Updated to use `AvroSchemaUtil.makeCompatibleName`.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -107,13 +107,15 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
...
+        String fieldName = AvroSchemaUtil.makeCompatibleName(field.name());
         Schema.Field newField =
             new Schema.Field(
-                field.name() + "_r" + field.fieldId(),
+                fieldName + "_r" + field.fieldId(),
                 AvroSchemaUtil.toOption(AvroSchemaUtil.convert(field.type())),
                 null,
                 JsonProperties.NULL_VALUE);
         newField.addProp(AvroSchemaUtil.FIELD_ID_PROP, field.fieldId());
+        newField.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, field.name());
```
szehon-ho, line 118: Nit: Can we use fieldName here?  (Actually prefer to rename to 'oldFieldName' if you don't mind.

Also , do you know the significance of this property?  Looks like in the other case its only set if the field name has been changed, should we do that?

core/src/test/java/org/apache/iceberg/TestMetadataTableScans.java
```diff
@@ -978,6 +1091,32 @@ private Set<String> expectedManifestListPaths(Iterable<Snapshot> snapshots, Long
         .collect(Collectors.toSet());
   }
 
+  private void createTableAndAddPartitionOnNestedFiled() throws IOException {
```
szehon-ho, line 1094: Great to add all these tests.  I think it may fit better in "TestMetadataTablesWithPartitionEvolution" because if i understand its a partition evolution case?  But maybe its too much work to move the new tests there.

szehon-ho (commented):
I think it looks good, just had a question on the field property and whether we always need to set it.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
ConeyLiu: > Nit: Can we use fieldName here? (Actually prefer to rename to 'oldFieldName' if you don't mind.

This aligns the code in https://github.com/apache/iceberg/blob/dbb8a404f6632a55acb36e949f0e7b84b643cede/core/src/main/java/org/apache/iceberg/avro/TypeToSchema.java#L110

> Also , do you know the significance of this property? 

Actually, I don't know the true meaning. It seems the property is set, but could not find any usage. Maybe for debugging or similar.

> Looks like in the other case its only set if the field name has been changed, should we do that?

I do this in previously commit: https://github.com/apache/iceberg/pull/4577/commits/d54148421d8740f6e792d1567e5c7fe610c12a48. And changed by some comments and advice.

core/src/test/java/org/apache/iceberg/TestMetadataTableScans.java
ConeyLiu: OK, let me look it.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -107,13 +107,15 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
             field.name());
         // Create a field that will be defaulted to null. We assign a unique suffix to the field
         // to make sure that even if records in the file have the field it is not projected.
+        String fieldName = AvroSchemaUtil.makeCompatibleName(field.name());
```
nastra, line 110: this is already being done in L98, can we maybe do `String avroFieldName = AvroSchemaUtil.makeCompatibleName(field.name());` in L97 and use `avroFieldName` in L113?

core/src/test/java/org/apache/iceberg/TestMetadataTableScans.java
nastra: `TestMetadataTablesWithPartitionEvolution` seems to be a Spark test. I would actually prefer to have the tests in `iceberg-core` (maybe in a subclass of `TestReadProjection`?)

nastra (commented):
thanks for fixing this, left a few minor comments

core/src/test/java/org/apache/iceberg/TestMetadataTableScans.java
szehon-ho: Ah didnt realize, yea its fine then for me where it is.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
ConeyLiu: Updated

core/src/test/java/org/apache/iceberg/TestMetadataTableScans.java
ConeyLiu: Split into a dedicated class.

core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
ConeyLiu: Changed to set the property `AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP` only when the field name is sanitized.

szehon-ho (approved):
Thanks @ConeyLiu  , code changes look good to me.  Can merge once other comments resolved

core/src/test/java/org/apache/iceberg/TestMetadataTableScansWithPartitionEvolution.java
```diff
@@ -0,0 +1,171 @@
...
+        .commit();
+  }
+
+  @Test
+  public void testManifestsTableWithAddPartitionOnNestedField() throws IOException {
+    Table manifestsTable = new ManifestsTable(table.ops(), table);
+    TableScan scan = manifestsTable.newScan();
+
+    try (CloseableIterable<FileScanTask> tasks = scan.planFiles()) {
+      Assert.assertEquals("Should have one tasks", 1, Iterables.size(tasks));
```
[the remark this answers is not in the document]
ConeyLiu: Updated to use assertions.

ConeyLiu:
Thanks @szehon-ho @nastra @chenjunjiedada @RussellSpitzer for your time.

PR #4578, opened 2022-04-18
Core: Update Remove Snapshots procedure for branching and tagging

amogh-jahagirdar:
This PR contains the updated expire snapshots procedure for branching and tagging. Implements the algorithm outlined here https://github.com/apache/iceberg/blob/master/format/spec.md#snapshot-retention-policy.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
+++ b/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
@@ -21,10 +21,12 @@ package org.apache.iceberg;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 import org.apache.iceberg.avro.Avro;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.NotFoundException;
@@ -33,7 +35,9 @@ import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.relocated.com.google.common.util.concurrent.MoreExecutors;
 import org.apache.iceberg.util.DateTimeUtil;
@@ -54,6 +58,8 @@ import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS;
 import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT;
 import static org.apache.iceberg.TableProperties.GC_ENABLED;
 import static org.apache.iceberg.TableProperties.GC_ENABLED_DEFAULT;
+import static org.apache.iceberg.TableProperties.MAX_REF_AGE_MS;
+import static org.apache.iceberg.TableProperties.MAX_REF_AGE_MS_DEFAULT;
 import static org.apache.iceberg.TableProperties.MAX_SNAPSHOT_AGE_MS;
 import static org.apache.iceberg.TableProperties.MAX_SNAPSHOT_AGE_MS_DEFAULT;
 import static org.apache.iceberg.TableProperties.MIN_SNAPSHOTS_TO_KEEP;
@@ -75,10 +81,12 @@ class RemoveSnapshots implements ExpireSnapshots {
 
   private final TableOperations ops;
   private final Set<Long> idsToRemove = Sets.newHashSet();
+  private final long now;
+  private final long defaultMaxRefAgeMs;
   private boolean cleanExpiredFiles = true;
   private TableMetadata base;
-  private long expireOlderThan;
-  private int minNumSnapshots;
+  private long defaultExpireOlderThan;
+  private int defaultMinNumSnapshots;
   private Consumer<String> deleteFunc = defaultDelete;
   private ExecutorService deleteExecutorService = DEFAULT_DELETE_EXECUTOR_SERVICE;
   private ExecutorService planExecutorService = ThreadPools.getWorkerPool();
@@ -86,21 +94,27 @@ class RemoveSnapshots implements ExpireSnapshots {
   RemoveSnapshots(TableOperations ops) {
     this.ops = ops;
     this.base = ops.current();
+    ValidationException.check(
+        PropertyUtil.propertyAsBoolean(base.properties(), GC_ENABLED, GC_ENABLED_DEFAULT),
+        "Cannot expire snapshots: GC is disabled (deleting files may corrupt other tables)");
 
-    long maxSnapshotAgeMs = PropertyUtil.propertyAsLong(
+    long defaultMaxSnapshotAgeMs = PropertyUtil.propertyAsLong(
         base.properties(),
         MAX_SNAPSHOT_AGE_MS,
         MAX_SNAPSHOT_AGE_MS_DEFAULT);
-    this.expireOlderThan = System.currentTimeMillis() - maxSnapshotAgeMs;
 
-    this.minNumSnapshots = PropertyUtil.propertyAsInt(
+    this.now = System.currentTimeMillis();
+    this.defaultExpireOlderThan = now - defaultMaxSnapshotAgeMs;
+    this.defaultMinNumSnapshots = PropertyUtil.propertyAsInt(
         base.properties(),
         MIN_SNAPSHOTS_TO_KEEP,
         MIN_SNAPSHOTS_TO_KEEP_DEFAULT);
 
-    ValidationException.check(
-        PropertyUtil.propertyAsBoolean(base.properties(), GC_ENABLED, GC_ENABLED_DEFAULT),
-        "Cannot expire snapshots: GC is disabled (deleting files may corrupt other tables)");
+    this.defaultMaxRefAgeMs = PropertyUtil.propertyAsLong(
+        base.properties(),
+        MAX_REF_AGE_MS,
+        MAX_REF_AGE_MS_DEFAULT
+    );
   }
 
   @Override
@@ -120,7 +134,7 @@ class RemoveSnapshots implements ExpireSnapshots {
   public ExpireSnapshots expireOlderThan(long timestampMillis) {
     LOG.info("Expiring snapshots older than: {} ({})",
         DateTimeUtil.formatTimestampMillis(timestampMillis), timestampMillis);
-    this.expireOlderThan = timestampMillis;
+    this.defaultExpireOlderThan = timestampMillis;
     return this;
   }
 
@@ -128,7 +142,7 @@ class RemoveSnapshots implements ExpireSnapshots {
   public ExpireSnapshots retainLast(int numSnapshots) {
     Preconditions.checkArgument(1 <= numSnapshots,
             "Number of snapshots to retain must be at least 1, cannot be: %s", numSnapshots);
-    this.minNumSnapshots = numSnapshots;
+    this.defaultMinNumSnapshots = numSnapshots;
     return this;
   }
 
@@ -161,21 +175,148 @@ class RemoveSnapshots implements ExpireSnapshots {
 
   private TableMetadata internalApply() {
     this.base = ops.refresh();
-
-    Set<Long> idsToRetain = Sets.newHashSet();
-    List<Long> ancestorIds = SnapshotUtil.ancestorIds(base.currentSnapshot(), base::snapshot);
-    if (minNumSnapshots >= ancestorIds.size()) {
-      idsToRetain.addAll(ancestorIds);
-    } else {
-      idsToRetain.addAll(ancestorIds.subList(0, minNumSnapshots));
+    if (base.snapshots().isEmpty()) {
+      return base;
     }
 
-    TableMetadata updateMeta = base.removeSnapshotsIf(snapshot ->
-        idsToRemove.contains(snapshot.snapshotId()) ||
-        (snapshot.timestampMillis() < expireOlderThan && !idsToRetain.contains(snapshot.snapshotId())));
-    List<Snapshot> updateSnapshots = updateMeta.snapshots();
-    List<Snapshot> baseSnapshots = base.snapshots();
-    return updateSnapshots.size() != baseSnapshots.size() ? updateMeta : base;
+    Set<Long> idsToRetain = Sets.newHashSet();
+
+    // Compute the snapshots for each reference
+    Map<String, Set<Long>> refSnapshots = computeRefSnapshots(base.refs());
+
+    // Identify unreferenced snapshots which should be retained
+    Set<Long> unreferencedSnapshotsToRetain = computeUnreferencedSnapshotsToRetain(refSnapshots);
+    idsToRetain.addAll(unreferencedSnapshotsToRetain);
+
+    // Identify refs that should be removed
+    Map<String, SnapshotRef> retainedRefs  = computeRetainedRefs(base.refs());
+    Map<Long, List<String>> retainedIdToRefs = Maps.newHashMap();
+    for (Map.Entry<String, SnapshotRef> retainedRefEntry : retainedRefs.entrySet()) {
+      long snapshotId = retainedRefEntry.getValue().snapshotId();
+      retainedIdToRefs.putIfAbsent(snapshotId, Lists.newArrayList());
+      retainedIdToRefs.get(snapshotId).add(retainedRefEntry.getKey());
+      idsToRetain.add(snapshotId);
+    }
+
+    for (long idToRemove : idsToRemove) {
+      List<String> refsForId = retainedIdToRefs.get(idToRemove);
+      Preconditions.checkArgument(refsForId == null,
+          "Cannot expire %s. Still referenced by refs: %s", idToRemove, refsForId);
+    }
+
+    Set<Long> branchSnapshotsToRetain = computeAllBranchSnapshotsToRetain(retainedRefs, refSnapshots);
+    idsToRetain.addAll(branchSnapshotsToRetain);
+    TableMetadata.Builder updatedMetaBuilder = TableMetadata.buildFrom(base);
+
+    base.snapshots().stream()
+        .map(Snapshot::snapshotId)
+        .filter(snapshot -> !idsToRetain.contains(snapshot))
+        .forEach(idsToRemove::add);
+    updatedMetaBuilder.removeSnapshots(idsToRemove);
+
+    base.refs().keySet().stream()
+        .filter(ref -> !retainedRefs.containsKey(ref))
+        .forEach(updatedMetaBuilder::removeRef);
+
+    return updatedMetaBuilder.build();
+  }
+
+  /**
+   * Helper to compute the mapping of a ref to its snapshots. If it's a branch, the snapshots is an ordered set
+   * of all the snapshots on the branch. If it's a tag, the snapshot is a set of the single snapshot the tag refers to
+   */
+  private Map<String, Set<Long>> computeRefSnapshots(Map<String, SnapshotRef> refs) {
+    Map<String, Set<Long>> refSnapshots = Maps.newHashMap();
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      String name = refEntry.getKey();
+      SnapshotRef ref = refEntry.getValue();
+      if (ref.isBranch()) {
+        Set<Long> branchAncestors = Sets.newLinkedHashSet();
+        Iterable<Snapshot> snapshots = SnapshotUtil.ancestorsOf(ref.snapshotId(), base::snapshot);
+        snapshots.forEach(snapshot -> branchAncestors.add(snapshot.snapshotId()));
+        refSnapshots.put(name, branchAncestors);
+      } else {
+        refSnapshots.put(name, ImmutableSet.of(ref.snapshotId()));
+      }
+    }
+
+    return refSnapshots;
+  }
+
+  private Map<String, SnapshotRef> computeRetainedRefs(Map<String, SnapshotRef> refs) {
+    Map<String, SnapshotRef> retainedRefs = Maps.newHashMap();
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      String name = refEntry.getKey();
+      SnapshotRef ref = refEntry.getValue();
+      if (name.equals(SnapshotRef.MAIN_BRANCH)) {
+        retainedRefs.put(name, ref);
+        continue;
+      }
+
+      Snapshot snapshot = base.snapshot(ref.snapshotId());
+      long maxRefAgeMs = ref.maxRefAgeMs() != null ? ref.maxRefAgeMs() : defaultMaxRefAgeMs;
+      if (snapshot != null) {
+        long refAgeMs = now - snapshot.timestampMillis();
+        if (refAgeMs <= maxRefAgeMs) {
+          retainedRefs.put(name, ref);
+        }
+      } else {
+        LOG.warn("Removing invalid ref {}: snapshot {} does not exist", name, ref.snapshotId());
+      }
+    }
+
+    return retainedRefs;
+  }
+
+  private Set<Long> computeUnreferencedSnapshotsToRetain(Map<String, Set<Long>> refSnapshots) {
+    Set<Long> referencedSnapshots = Sets.newHashSet();
+    refSnapshots.values().forEach(referencedSnapshots::addAll);
+
+    return base.snapshots().stream()
+        .filter(snapshot -> !referencedSnapshots.contains(snapshot.snapshotId()))
+        .filter(snapshot -> snapshot.timestampMillis() >= defaultExpireOlderThan)
+        .map(Snapshot::snapshotId)
+        .collect(Collectors.toSet());
+  }
+
+  private Set<Long> computeAllBranchSnapshotsToRetain(
+      Map<String, SnapshotRef> refs,
+      Map<String, Set<Long>> refSnapshots) {
+
+    Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      final String name = refEntry.getKey();
+      final SnapshotRef ref = refEntry.getValue();
+
+      if (ref.isBranch()) {
+        long expireSnapshotsOlderThan = ref.maxSnapshotAgeMs() != null ? now - ref.maxSnapshotAgeMs() :
+            defaultExpireOlderThan;
+        int minSnapshotsToKeep = ref.minSnapshotsToKeep() != null ? ref.minSnapshotsToKeep() :
+            defaultMinNumSnapshots;
+        Set<Long> branchAncestors = refSnapshots.get(name);
+        branchSnapshotsToRetain.addAll(
+            computeBranchSnapshotsToRetain(branchAncestors, expireSnapshotsOlderThan, minSnapshotsToKeep));
+      }
+    }
+
+    return branchSnapshotsToRetain;
+  }
+
+  private Set<Long> computeBranchSnapshotsToRetain(
+      Set<Long> branchSnapshots,
+      long expireSnapshotsOlderThan,
+      int minSnapshotsToKeep) {
+    Set<Long> idsToRetain = Sets.newHashSet();
+    for (long snapshot : branchSnapshots) {
+      Snapshot ancestor = base.snapshot(snapshot);
+      if (idsToRetain.size() < minSnapshotsToKeep || ancestor.timestampMillis() >= expireSnapshotsOlderThan) {
+        idsToRetain.add(ancestor.snapshotId());
+      } else {
+        return idsToRetain;
+      }
+    }
+
+    return idsToRetain;
   }
 
   @Override
@@ -190,6 +331,10 @@ class RemoveSnapshots implements ExpireSnapshots {
         .onlyRetryOn(CommitFailedException.class)
         .run(item -> {
           TableMetadata updated = internalApply();
+          if (cleanExpiredFiles && updated.refs().size() > 1) {
+            throw new UnsupportedOperationException("Cannot incrementally clean files for tables with more than 1 ref");
+          }
+
           ops.commit(base, updated);
         });
     LOG.info("Committed snapshot changes");
--- a/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
+++ b/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
@@ -20,13 +20,13 @@
 package org.apache.iceberg;
 
 import java.io.IOException;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 import org.apache.iceberg.avro.Avro;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.NotFoundException;
@@ -35,7 +35,6 @@ import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
-import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
@@ -180,16 +179,8 @@ class RemoveSnapshots implements ExpireSnapshots {
     }
 
     Set<Long> idsToRetain = Sets.newHashSet();
-
-    // Compute the snapshots for each reference
-    Map<String, Set<Long>> refSnapshots = computeRefSnapshots(base.refs());
-
-    // Identify unreferenced snapshots which should be retained
-    Set<Long> unreferencedSnapshotsToRetain = computeUnreferencedSnapshotsToRetain(refSnapshots);
-    idsToRetain.addAll(unreferencedSnapshotsToRetain);
-
     // Identify refs that should be removed
-    Map<String, SnapshotRef> retainedRefs  = computeRetainedRefs(base.refs());
+    Map<String, SnapshotRef> retainedRefs = computeRetainedRefs(base.refs());
     Map<Long, List<String>> retainedIdToRefs = Maps.newHashMap();
     for (Map.Entry<String, SnapshotRef> retainedRefEntry : retainedRefs.entrySet()) {
       long snapshotId = retainedRefEntry.getValue().snapshotId();
@@ -204,9 +195,14 @@ class RemoveSnapshots implements ExpireSnapshots {
           "Cannot expire %s. Still referenced by refs: %s", idToRemove, refsForId);
     }
 
-    Set<Long> branchSnapshotsToRetain = computeAllBranchSnapshotsToRetain(retainedRefs, refSnapshots);
-    idsToRetain.addAll(branchSnapshotsToRetain);
+    idsToRetain.addAll(computeAllBranchSnapshotsToRetain(retainedRefs.values()));
+    idsToRetain.addAll(unreferencedSnapshotsToRetain(retainedRefs.values()));
+
     TableMetadata.Builder updatedMetaBuilder = TableMetadata.buildFrom(base);
+
+    base.refs().keySet().stream()
+        .filter(ref -> !retainedRefs.containsKey(ref))
+        .forEach(updatedMetaBuilder::removeRef);
 
     base.snapshots().stream()
         .map(Snapshot::snapshotId)
@@ -214,33 +210,7 @@ class RemoveSnapshots implements ExpireSnapshots {
         .forEach(idsToRemove::add);
     updatedMetaBuilder.removeSnapshots(idsToRemove);
 
-    base.refs().keySet().stream()
-        .filter(ref -> !retainedRefs.containsKey(ref))
-        .forEach(updatedMetaBuilder::removeRef);
-
     return updatedMetaBuilder.build();
-  }
-
-  /**
-   * Helper to compute the mapping of a ref to its snapshots. If it's a branch, the snapshots is an ordered set
-   * of all the snapshots on the branch. If it's a tag, the snapshot is a set of the single snapshot the tag refers to
-   */
-  private Map<String, Set<Long>> computeRefSnapshots(Map<String, SnapshotRef> refs) {
-    Map<String, Set<Long>> refSnapshots = Maps.newHashMap();
-    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
-      String name = refEntry.getKey();
-      SnapshotRef ref = refEntry.getValue();
-      if (ref.isBranch()) {
-        Set<Long> branchAncestors = Sets.newLinkedHashSet();
-        Iterable<Snapshot> snapshots = SnapshotUtil.ancestorsOf(ref.snapshotId(), base::snapshot);
-        snapshots.forEach(snapshot -> branchAncestors.add(snapshot.snapshotId()));
-        refSnapshots.put(name, branchAncestors);
-      } else {
-        refSnapshots.put(name, ImmutableSet.of(ref.snapshotId()));
-      }
-    }
-
-    return refSnapshots;
   }
 
   private Map<String, SnapshotRef> computeRetainedRefs(Map<String, SnapshotRef> refs) {
@@ -268,34 +238,16 @@ class RemoveSnapshots implements ExpireSnapshots {
     return retainedRefs;
   }
 
-  private Set<Long> computeUnreferencedSnapshotsToRetain(Map<String, Set<Long>> refSnapshots) {
-    Set<Long> referencedSnapshots = Sets.newHashSet();
-    refSnapshots.values().forEach(referencedSnapshots::addAll);
-
-    return base.snapshots().stream()
-        .filter(snapshot -> !referencedSnapshots.contains(snapshot.snapshotId()))
-        .filter(snapshot -> snapshot.timestampMillis() >= defaultExpireOlderThan)
-        .map(Snapshot::snapshotId)
-        .collect(Collectors.toSet());
-  }
-
-  private Set<Long> computeAllBranchSnapshotsToRetain(
-      Map<String, SnapshotRef> refs,
-      Map<String, Set<Long>> refSnapshots) {
-
+  private Set<Long> computeAllBranchSnapshotsToRetain(Collection<SnapshotRef> refs) {
     Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
-    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
-      final String name = refEntry.getKey();
-      final SnapshotRef ref = refEntry.getValue();
-
+    for (SnapshotRef ref : refs) {
       if (ref.isBranch()) {
         long expireSnapshotsOlderThan = ref.maxSnapshotAgeMs() != null ? now - ref.maxSnapshotAgeMs() :
             defaultExpireOlderThan;
         int minSnapshotsToKeep = ref.minSnapshotsToKeep() != null ? ref.minSnapshotsToKeep() :
             defaultMinNumSnapshots;
-        Set<Long> branchAncestors = refSnapshots.get(name);
         branchSnapshotsToRetain.addAll(
-            computeBranchSnapshotsToRetain(branchAncestors, expireSnapshotsOlderThan, minSnapshotsToKeep));
+            computeBranchSnapshotsToRetain(ref.snapshotId(), expireSnapshotsOlderThan, minSnapshotsToKeep));
       }
     }
 
@@ -303,12 +255,11 @@ class RemoveSnapshots implements ExpireSnapshots {
   }
 
   private Set<Long> computeBranchSnapshotsToRetain(
-      Set<Long> branchSnapshots,
+      long snapshot,
       long expireSnapshotsOlderThan,
       int minSnapshotsToKeep) {
     Set<Long> idsToRetain = Sets.newHashSet();
-    for (long snapshot : branchSnapshots) {
-      Snapshot ancestor = base.snapshot(snapshot);
+    for (Snapshot ancestor : SnapshotUtil.ancestorsOf(snapshot, base::snapshot)) {
       if (idsToRetain.size() < minSnapshotsToKeep || ancestor.timestampMillis() >= expireSnapshotsOlderThan) {
         idsToRetain.add(ancestor.snapshotId());
       } else {
@@ -317,6 +268,25 @@ class RemoveSnapshots implements ExpireSnapshots {
     }
 
     return idsToRetain;
+  }
+
+  private Set<Long> unreferencedSnapshotsToRetain(Collection<SnapshotRef> refs) {
+    Set<Long> referencedSnapshots = Sets.newHashSet();
+    for (SnapshotRef ref : refs) {
+      for (Snapshot snapshot : SnapshotUtil.ancestorsOf(ref.snapshotId(), base::snapshot)) {
+        referencedSnapshots.add(snapshot.snapshotId());
+      }
+    }
+
+    Set<Long> snapshotsToRetain = Sets.newHashSet();
+    for (Snapshot snapshot : base.snapshots()) {
+      if (!referencedSnapshots.contains(snapshot.snapshotId()) && // unreferenced
+          snapshot.timestampMillis() >= defaultExpireOlderThan) { // not old enough to expire
+        snapshotsToRetain.add(snapshot.snapshotId());
+      }
+    }
+
+    return snapshotsToRetain;
   }
 
   @Override
--- a/core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java
+++ b/core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java
@@ -1204,4 +1204,159 @@ public class TestRemoveSnapshots extends TableTestBase {
             .build(),
         deletedFiles);
   }
+
+  @Test
+  public void testRefExpiration() {
+    table.newAppend()
+        .appendFile(FILE_A)
+        .commit();
+    long now = System.currentTimeMillis();
+    long tagARefAgeMs = 100;
+    table.manageSnapshots()
+        .createTag("TagA", table.currentSnapshot().snapshotId())
+        .setMaxRefAgeMs("TagA", tagARefAgeMs)
+        .commit();
+    long expirationTime = now + tagARefAgeMs;
+    table.newAppend()
+        .appendFile(FILE_B)
+        .commit();
+    table.manageSnapshots()
+        .createBranch("BranchB", table.currentSnapshot().snapshotId())
+        .commit();
+
+    waitUntilAfter(expirationTime);
+
+    table.expireSnapshots().cleanExpiredFiles(false).commit();
+
+    Assert.assertNull(table.ops().current().ref("TagA"));
+    Assert.assertNotNull(table.ops().current().ref("BranchB"));
+    Assert.assertNotNull(table.ops().current().ref(SnapshotRef.MAIN_BRANCH));
+  }
+
+  @Test
+  public void testMultipleRefsAndCleanExpiredFilesFails() {
+    table.newAppend()
+        .appendFile(FILE_A)
+        .commit();
+    table.manageSnapshots()
+        .createTag("TagA", table.currentSnapshot().snapshotId())
+        .commit();
+
+    AssertHelpers.assertThrows(
+        "Should fail removing snapshots and files when there is more than 1 ref",
+        UnsupportedOperationException.class,
+        "Cannot incrementally clean files for tables with more than 1 ref",
+        () -> table.expireSnapshots().cleanExpiredFiles(true).commit());
+  }
+
+  @Test
+  public void testFailRemovingSnapshotWhenStillReferenced() {
+    table.newAppend()
+        .appendFile(FILE_A)
+        .commit();
+    table.newAppend()
+        .appendFile(FILE_B)
+        .stageOnly()
+        .commit();
+    table.manageSnapshots().createTag("stagedB", 2).commit();
+
+    AssertHelpers.assertThrows(
+        "Should fail removing snapshot when it is still referenced",
+          IllegalArgumentException.class,
+        "Cannot expire 2. Still referenced by refs: [stagedB]",
+        () -> table.expireSnapshots().expireSnapshotId(2).commit());
+  }
+
+  @Test
+  public void testRetainStagedSnapshotsWithinExpirationAge() {
+    table.newAppend()
+        .appendFile(FILE_A)
+        .commit();
+    long expireTimestampSnapshotA = waitUntilAfter(table.currentSnapshot().timestampMillis());
+    table.newAppend()
+        .appendFile(FILE_B)
+        .stageOnly()
+        .commit();
+
+    table.newAppend()
+        .appendFile(FILE_C)
+        .commit();
+
+    table.expireSnapshots().expireOlderThan(expireTimestampSnapshotA).commit();
+
+    Assert.assertEquals(2, table.ops().current().snapshots().size());
+  }
+
+  // ToDo: Add tests which commit to branches once committing snapshots to a branch is supported
+
+  @Test
+  public void testMinSnapshotsToKeepMultipleBranches() {
+    table.newAppend().appendFile(FILE_A).commit();
+    table.newAppend().appendFile(FILE_B).commit();
+    table.newAppend().appendFile(FILE_C).stageOnly().commit();
+    int branchSnapshot = 3;
+    String branchName = "branchC";
+    long maxSnapshotAgeMs = 1;
+    long expirationTime = System.currentTimeMillis() + maxSnapshotAgeMs;
+
+    // Retain 1 snapshot on main
+    table.manageSnapshots()
+        .setMinSnapshotsToKeep(SnapshotRef.MAIN_BRANCH, 1)
+        .setMaxSnapshotAgeMs(SnapshotRef.MAIN_BRANCH, 1)
+        .commit();
+
+    // Retain all snapshots on branchC
+    table.manageSnapshots()
+        .createBranch(branchName, branchSnapshot)
+        .setMinSnapshotsToKeep(branchName, 3)
+        .setMaxSnapshotAgeMs(branchName, 1)
+        .commit();
+
+    waitUntilAfter(expirationTime);
+    table.expireSnapshots().cleanExpiredFiles(false).commit();
+
+    Assert.assertEquals(3, table.ops().current().snapshots().size());
+
+    table.manageSnapshots().setMinSnapshotsToKeep(branchName, 1).commit();
+    table.expireSnapshots().cleanExpiredFiles(false).commit();
+
+    Assert.assertEquals(2, table.ops().current().snapshots().size());
+    Assert.assertNull(table.ops().current().snapshot(1));
+  }
+
+  @Test
+  public void testMaxSnapshotAgeMultipleBranches() {
+    table.newAppend().appendFile(FILE_A).commit();
+    long ageFirstSnapshot = 10;
+    long expirationTime = System.currentTimeMillis() + ageFirstSnapshot;
+    table.manageSnapshots()
+        .setMaxSnapshotAgeMs(SnapshotRef.MAIN_BRANCH, ageFirstSnapshot)
+        .setMinSnapshotsToKeep(SnapshotRef.MAIN_BRANCH, 1)
+        .commit();
+    String branchName = "branchC";
+
+    waitUntilAfter(expirationTime);
+
+    table.newAppend().appendFile(FILE_B).commit();
+    table.newAppend().appendFile(FILE_C).stageOnly().commit();
+
+    table.manageSnapshots()
+        .createBranch(branchName, 3)
+        .setMinSnapshotsToKeep(branchName, 1)
+        .setMaxSnapshotAgeMs(branchName, Long.MAX_VALUE)
+        .commit();
+
+    table.expireSnapshots().cleanExpiredFiles(false).commit();
+
+    Assert.assertEquals(3, table.ops().current().snapshots().size());
+
+    table.manageSnapshots()
+        .setMaxSnapshotAgeMs(branchName, ageFirstSnapshot)
+        .commit();
+
+    table.expireSnapshots().cleanExpiredFiles(false).commit();
+
+    Assert.assertEquals(2, table.ops().current().snapshots().size());
+    Assert.assertNull(table.ops().current().snapshot(1));
+  }
 }
--- a/core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java
+++ b/core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java
@@ -1206,30 +1206,66 @@ public class TestRemoveSnapshots extends TableTestBase {
   }
 
   @Test
-  public void testRefExpiration() {
+  public void testTagExpiration() {
     table.newAppend()
         .appendFile(FILE_A)
         .commit();
+
     long now = System.currentTimeMillis();
-    long tagARefAgeMs = 100;
+    long maxAgeMs = 100;
+    long expirationTime = now + maxAgeMs;
+
     table.manageSnapshots()
-        .createTag("TagA", table.currentSnapshot().snapshotId())
-        .setMaxRefAgeMs("TagA", tagARefAgeMs)
+        .createTag("tag", table.currentSnapshot().snapshotId())
+        .setMaxRefAgeMs("tag", maxAgeMs)
         .commit();
-    long expirationTime = now + tagARefAgeMs;
+
     table.newAppend()
         .appendFile(FILE_B)
         .commit();
+
     table.manageSnapshots()
-        .createBranch("BranchB", table.currentSnapshot().snapshotId())
+        .createBranch("branch", table.currentSnapshot().snapshotId())
         .commit();
 
     waitUntilAfter(expirationTime);
 
     table.expireSnapshots().cleanExpiredFiles(false).commit();
 
-    Assert.assertNull(table.ops().current().ref("TagA"));
-    Assert.assertNotNull(table.ops().current().ref("BranchB"));
+    Assert.assertNull(table.ops().current().ref("tag"));
+    Assert.assertNotNull(table.ops().current().ref("branch"));
+    Assert.assertNotNull(table.ops().current().ref(SnapshotRef.MAIN_BRANCH));
+  }
+
+  @Test
+  public void testBranchExpiration() {
+    table.newAppend()
+        .appendFile(FILE_A)
+        .commit();
+
+    long now = System.currentTimeMillis();
+    long maxAgeMs = 100;
+    long expirationTime = now + maxAgeMs;
+
+    table.manageSnapshots()
+        .createBranch("branch", table.currentSnapshot().snapshotId())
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -75,32 +77,47 @@ public void accept(String file) {
...
+
+    if (mainBranch != null) {
+      maxSnapshotAgeMs = mainBranch.maxSnapshotAgeMs();
+    }
+    if (maxSnapshotAgeMs == null) {
+      maxSnapshotAgeMs = PropertyUtil.propertyAsLong(
+              base.properties(),
+              MAX_SNAPSHOT_AGE_MS,
+              MAX_SNAPSHOT_AGE_MS_DEFAULT);
+    }
```
amogh-jahagirdar, line 107: This preserves the behavior that if there is no max snapshot age defined on the main branch, the procedure will use table level property, and if that's still not defined, the procedure will use the default property. This was simply done to accommodate whatever existing tests we had.

 However, I'm not sure if this is the behavior we actually want. If there's null set on the main branch for maxSnapshotAgeMs, or for minNUmSnapshotsToKeep, it would make more sense for the procedure to treat that as nothing on main should be expired. Then at this point, the table properties would not matter for the expiration logic. Everything would have to be done via branch retention properties.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
...
+
+    for (Map.Entry<String, SnapshotRef> branchEntry : branches) {
+      String name = branchEntry.getKey();
+      SnapshotRef branch = branchEntry.getValue();
+      Long expireSnapshotsOlderThan = null;
+      Integer minSnapshotsToKeep = null;
+      if (name.equals(SnapshotRef.MAIN_BRANCH)) {
+        expireSnapshotsOlderThan = this.expireOlderThan;
+        minSnapshotsToKeep = this.minNumSnapshots;
+      } else if (branch.maxSnapshotAgeMs() != null) {
```
amogh-jahagirdar, line 215: Probably for better readability I should break this up into smaller functions

1.) Identifying refs to remove
2.) Identifying branch snapshots to keep
3.) The builder updates

amogh-jahagirdar:
Still missing unit tests on this but wanted to get feedback on the code while I add the tests.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
...
+  }
+
+  private Set<Long> computeBranchSnapshotsToRetain(SnapshotRef branch, Long expireSnapshotsOlderThan,
+      Integer minSnapshotsToKeep) {
+
+    Snapshot snapshot = base.snapshot(branch.snapshotId());
+    List<Long> ancestorIds = SnapshotUtil.ancestorIds(snapshot, base::snapshot);
+    Set<Long> idsToRetain = Sets.newHashSet();
+    // Everything on this branch should be retained
+    if (expireSnapshotsOlderThan == null || minSnapshotsToKeep == null || minSnapshotsToKeep >= ancestorIds.size()) {
```
rajarshisarkar, line 247: I have a doubt,

The condition would evaluate to `true` if we have `expireSnapshotsOlderThan != null` and `minSnapshotsToKeep >= ancestorIds.size()`

In that case, can some of the `ancestorIds` have `ancestor.timestampMillis() < expireOlderThan`? If yes, should we retain them?

amogh-jahagirdar: > The condition would evaluate to true if we have expireSnapshotsOlderThan != null and minSnapshotsToKeep >= ancestorIds.size()

That is correct! 

> In that case, can some of the ancestorIds have ancestor.timestampMillis() < expireOlderThan? If yes, should we retain them?

Yes we would be retaining them even if some of the ancestors are older than the threshold. I believe this should be the desired behavior because a user explicitly set this number of minimum number of snapshots to retain we would respect that (since retention is inclusive of what the user set).

rajarshisarkar: I see, makes sense to me.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -75,32 +77,47 @@ public void accept(String file) {
...
         PropertyUtil.propertyAsBoolean(base.properties(), GC_ENABLED, GC_ENABLED_DEFAULT),
         "Cannot expire snapshots: GC is disabled (deleting files may corrupt other tables)");
+
+    SnapshotRef mainBranch = base.ref(SnapshotRef.MAIN_BRANCH);
+    Long maxSnapshotAgeMs = null;
+
+    if (mainBranch != null) {
+      maxSnapshotAgeMs = mainBranch.maxSnapshotAgeMs();
+    }
+    if (maxSnapshotAgeMs == null) {
```
rdblue, line 102: Don't forget style guidelines.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
...
+    // Identify refs that should be removed
+    Set<String> refsToRemove = Sets.newHashSet();
+    for (String name : base.refs().keySet()) {
+      SnapshotRef ref = base.ref(name);
+      if (!name.equals(SnapshotRef.MAIN_BRANCH) && ref.maxRefAgeMs() != null) {
+        Snapshot snapshot = base.snapshot(ref.snapshotId());
+        if (removalTimeReference - snapshot.timestampMillis() >= ref.maxRefAgeMs()) {
+          refsToRemove.add(name);
+        }
+
```
rdblue, line 192: Nit: Useless newline.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
     this.base = ops.refresh();
 
     Set<Long> idsToRetain = Sets.newHashSet();
-    List<Long> ancestorIds = SnapshotUtil.ancestorIds(base.currentSnapshot(), base::snapshot);
-    if (minNumSnapshots >= ancestorIds.size()) {
+    // Identify refs that should be removed
+    Set<String> refsToRemove = Sets.newHashSet();
+    for (String name : base.refs().keySet()) {
+      SnapshotRef ref = base.ref(name);
+      if (!name.equals(SnapshotRef.MAIN_BRANCH) && ref.maxRefAgeMs() != null) {
```
rdblue, line 187: I don't think this is quite right. The max ref age table setting should be applied if the ref itself has no maximum age. Instead, this just skips the ref.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -75,32 +77,47 @@ public void accept(String file) {
...
+    if (mainBranch != null) {
+      maxSnapshotAgeMs = mainBranch.maxSnapshotAgeMs();
+    }
+    if (maxSnapshotAgeMs == null) {
+      maxSnapshotAgeMs = PropertyUtil.propertyAsLong(
+              base.properties(),
+              MAX_SNAPSHOT_AGE_MS,
+              MAX_SNAPSHOT_AGE_MS_DEFAULT);
+    }
+    this.removalTimeReference = System.currentTimeMillis();
```
rdblue, line 108: How about `now`?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
     this.base = ops.refresh();
 
     Set<Long> idsToRetain = Sets.newHashSet();
-    List<Long> ancestorIds = SnapshotUtil.ancestorIds(base.currentSnapshot(), base::snapshot);
-    if (minNumSnapshots >= ancestorIds.size()) {
+    // Identify refs that should be removed
+    Set<String> refsToRemove = Sets.newHashSet();
```
rdblue, line 184: What about `expiredRefs`?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
...
+      SnapshotRef ref = base.ref(name);
+      if (!name.equals(SnapshotRef.MAIN_BRANCH) && ref.maxRefAgeMs() != null) {
+        Snapshot snapshot = base.snapshot(ref.snapshotId());
+        if (removalTimeReference - snapshot.timestampMillis() >= ref.maxRefAgeMs()) {
+          refsToRemove.add(name);
+        }
+
+      }
+
+      if (!refsToRemove.contains(name) && !idsToRemove.contains(ref.snapshotId())) {
```
rdblue, line 195: Should `idsToRemove` be allowed to contain snapshots from refs? I think you should be able to remove a snapshot that is a ref and have that drop the ref. I think that should be an error when you expire the snapshot by ID.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +180,86 @@ private TableMetadata internalApply() {
...
+        idsToRetain.add(ref.snapshotId());
+      }
+
+    }
+    List<Map.Entry<String, SnapshotRef>> branches = base
```
rdblue, line 200: Don't forget style.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: You are correct, I missed in the spec, there's a table level property for max-ref-age-ms that we want to default to. The same applies for the other retention properties. This behavior makes more sense. I will update to fix this.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: If I understand what you are saying we are distinguishing the 2 cases, 

1.) the metadata operation to **remove** a snapshot should also drop any refs associated with it. 

2.) **Expiration** of a snapshot which has references should result in failure

This makes sense to me because removing implies a more "blunt" kind of operation where we actually intend on deleting that information. Expiration does not really have that implication, that way unknowingly running an unsafe expiration is prevented.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated, added missing newline after if block

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -112,6 +137,11 @@ public ExpireSnapshots cleanExpiredFiles(boolean clean) {
   @Override
   public ExpireSnapshots expireSnapshotId(long expireSnapshotId) {
     LOG.info("Expiring snapshot with id: {}", expireSnapshotId);
+    List<Map.Entry<String, SnapshotRef>> refsForSnapshot = base.refs().entrySet().stream()
+        .filter(refEntry -> refEntry.getValue().snapshotId() == expireSnapshotId)
+        .collect(Collectors.toList());
+    Preconditions.checkArgument(refsForSnapshot.isEmpty(), "Cannot expire %d. There are %d references to it",
+        expireSnapshotId, refsForSnapshot.size());
```
amogh-jahagirdar, line 144: Wrong place to put this check, we should do it in the internalApply after computing which refs will be retained. then we can check if the snapshots of the refs to be retained contains any of the individual snapshots in idsToRemove.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+
+  private Set<Long> computeAllBranchSnapshotsToRetain(List<Map.Entry<String, SnapshotRef>> branches) {
+    Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
+    for (Map.Entry<String, SnapshotRef> branchEntry : branches) {
+      String name = branchEntry.getKey();
+      SnapshotRef branch = branchEntry.getValue();
+      Long expireSnapshotsOlderThan;
+      Integer minSnapshotsToKeep;
+      if (name.equals(SnapshotRef.MAIN_BRANCH) || branch.maxSnapshotAgeMs() == null) {
+        expireSnapshotsOlderThan = this.expireOlderThan;
```
amogh-jahagirdar, line 231: This isn't right, this.expireOlder than can be set to the main branch's retention, and we don't want to use that in case the branch snapshot age does not exist; we should use the table level property here as well. Will fix this.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -195,6 +299,18 @@ public void commit() {
...
+      TableMetadata updated = ops.refresh();
+      if (updated.refs() != null) {
+        List<SnapshotRef> branches = updated.refs()
+            .values().stream().filter(SnapshotRef::isBranch)
+            .collect(Collectors.toList());
+
+        if (branches.size() > 1) {
+          throw new UnsupportedOperationException(
+              "Deleting expired files when there is more than 1 branch is currently not supported");
+        }
```
kbendick, line 311: Should this throw after doing all of the work (e.g. the commit etc)?

If I understand correctly, throwing here will cause the actual remove snapshots query / action to "fail" from the POV of the end user.

If that's the case, i think we should log an error and then proceed given that we've already committed the snapshot changes, or check this earlier on and then throw if need be.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -75,32 +79,53 @@ public void accept(String file) {
...
     ValidationException.check(
         PropertyUtil.propertyAsBoolean(base.properties(), GC_ENABLED, GC_ENABLED_DEFAULT),
         "Cannot expire snapshots: GC is disabled (deleting files may corrupt other tables)");
+
+    SnapshotRef mainBranch = base.ref(SnapshotRef.MAIN_BRANCH);
+    Long maxSnapshotAgeMs = null;
+
+    if (mainBranch != null) {
+      maxSnapshotAgeMs = mainBranch.maxSnapshotAgeMs();
+      minNumSnapshots = mainBranch.minSnapshotsToKeep();
```
rdblue, line 104: Why are these set from the main branch instead of table properties?

I thought that the table properties were used for any branch that doesn't have its own retention settings. Main shouldn't be treated any differently.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+          computeBranchSnapshotsToRetain(branch, expireSnapshotsOlderThan, minSnapshotsToKeep));
+    }
+    return branchSnapshotsToRetain;
+  }
+
+  private Set<String> computeExpiredRefs(Map<String, SnapshotRef> refs) {
+    Set<String> expiredRefs = Sets.newHashSet();
+    for (String name : refs.keySet()) {
+      SnapshotRef ref = refs.get(name);
+      Long maxRefAgeMs = ref.maxRefAgeMs() != null ? ref.maxRefAgeMs() : this.defaultMaxRefAgeMs;
```
rdblue, line 252: Style: only use `this.` when setting an instance field. It is used to show that the object state is being modified, but we don't use it when accessing that state.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+      if (!name.equals(SnapshotRef.MAIN_BRANCH)) {
```
rdblue, line 253: I generally prefer a flatter structure by moving these checks to the start:

```java
  for (String name : refs.keySet()) {
    if (SnapshotRef.MAIN_BRANCH.equals(name)) {
      continue;
    }

    ...
  }
```

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+        Snapshot snapshot = base.snapshot(ref.snapshotId());
+        if (now - snapshot.timestampMillis() >= maxRefAgeMs) {
```
rdblue, line 255: I think this should also check whether `snapshot` is null to avoid NPE when reading bad table data.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+      branchSnapshotsToRetain.addAll(
+          computeBranchSnapshotsToRetain(branch, expireSnapshotsOlderThan, minSnapshotsToKeep));
+    }
+    return branchSnapshotsToRetain;
+  }
+
+  private Set<String> computeExpiredRefs(Map<String, SnapshotRef> refs) {
+    Set<String> expiredRefs = Sets.newHashSet();
+    for (String name : refs.keySet()) {
+      SnapshotRef ref = refs.get(name);
```
rdblue, line 251: Why iterate over `keySet` and then immediately call `refs.get` rather than iterating over `entrySet`?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
rdblue, line 252: Is this really a `Long`? I think `defaultMaxRefAgeMs` is always defined, so this should be a `long`. That way you can avoid a NPE when comparing this to `now - snapshot.timestampMillis()`, which will cast it to a `long` anyway.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+      Long maxRefAgeMs = ref.maxRefAgeMs() != null ? ref.maxRefAgeMs() : this.defaultMaxRefAgeMs;
+      if (!name.equals(SnapshotRef.MAIN_BRANCH)) {
+        Snapshot snapshot = base.snapshot(ref.snapshotId());
+        if (now - snapshot.timestampMillis() >= maxRefAgeMs) {
+          expiredRefs.add(name);
+        }
+      }
+    }
+    return expiredRefs;
```
rdblue, line 260: Style: don't forget space between control flow blocks and the following statements.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
     this.base = ops.refresh();
 
     Set<Long> idsToRetain = Sets.newHashSet();
-    List<Long> ancestorIds = SnapshotUtil.ancestorIds(base.currentSnapshot(), base::snapshot);
-    if (minNumSnapshots >= ancestorIds.size()) {
+    // Identify refs that should be removed
+    Set<String> expiredRefs = computeExpiredRefs(base.refs());
+    Set<Long> idsOfRetainedRefs = base.refs().entrySet().stream()
```
rdblue, line 193: Rather than using `xOfY` you can usually rephrase as `yX`: `retainedRefIds` for example.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+        .filter(ref -> !expiredRefs.contains(ref.getKey()))
+        .map(refEntry -> refEntry.getValue().snapshotId())
+        .collect(Collectors.toSet());
+
+    for (Long idToRemove : idsToRemove) {
```
rdblue, line 198: Try to use primitives where possible. Here, you know that each ref had a non-null snapshot ID, so you know all of these are going to be non-null.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+    TableMetadata.Builder updatedMetaBuilder = TableMetadata.buildFrom(base);
+    idsToRemove.addAll(base.snapshots().stream().map(Snapshot::snapshotId)
+        .filter(snapshot -> !idsToRetain.contains(snapshot)).collect(Collectors.toSet()));
+    updatedMetaBuilder.removeSnapshots(idsToRemove);
+    expiredRefs.forEach(updatedMetaBuilder::removeRef);
+    TableMetadata updatedMetadata = updatedMetaBuilder.build();
+    List<Snapshot> updateSnapshots = updatedMetadata.snapshots();
+    List<Snapshot> baseSnapshots = base.snapshots();
+    boolean updated = updateSnapshots.size() != baseSnapshots.size() || expiredRefs.size() > 0;
+    return updated ? updatedMetadata : base;
```
rdblue, line 220: The builder should return the original metadata if there are no updates, so you don't need to duplicate the logic here.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+        .filter(refEntry -> refEntry.getValue().isBranch())
+        .collect(Collectors.toList());
+
+    Set<Long> branchSnapshotsToRetain = computeAllBranchSnapshotsToRetain(branches);
+    idsToRetain.addAll(branchSnapshotsToRetain);
+    TableMetadata.Builder updatedMetaBuilder = TableMetadata.buildFrom(base);
+    idsToRemove.addAll(base.snapshots().stream().map(Snapshot::snapshotId)
+        .filter(snapshot -> !idsToRetain.contains(snapshot)).collect(Collectors.toSet()));
+    updatedMetaBuilder.removeSnapshots(idsToRemove);
+    expiredRefs.forEach(updatedMetaBuilder::removeRef);
```
rdblue, line 215: You should remove the refs before removing the snapshots, just in case we add a consistency check later.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+    TableMetadata updatedMetadata = updatedMetaBuilder.build();
+    List<Snapshot> updateSnapshots = updatedMetadata.snapshots();
+    List<Snapshot> baseSnapshots = base.snapshots();
+    boolean updated = updateSnapshots.size() != baseSnapshots.size() || expiredRefs.size() > 0;
+    return updated ? updatedMetadata : base;
+  }
+
+  private Set<Long> computeAllBranchSnapshotsToRetain(List<Map.Entry<String, SnapshotRef>> branches) {
```
rdblue, line 223: It's a bit strange to pass `List<Map.Entry<String, SnapshotRef>>` here, but `Map<String, SnapshotRef>` to other methods. I'd probably just move branch filtering into this, since you're already looping over the map entries.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+    Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
+    for (Map.Entry<String, SnapshotRef> branchEntry : branches) {
+      String name = branchEntry.getKey();
+      SnapshotRef branch = branchEntry.getValue();
+      Long expireSnapshotsOlderThan;
```
rdblue, line 228: I think this should use the same logic that you have for `maxRefAgeMs`:

```
long maxSnapshotAgeMs = branch.maxSnapshotAgeMs() != null ? branch.maxSnapshotAgeMs() : defaultMaxSnapshotAgeMs;
long minSnapshotsToKeep = branch.minSnapshotsToKeep() != null ? branch.minSnapshotsToKeep() : defaultMinSnapshotsToKeep;
```

And those defaults should be from table settings. I don't think we treat `main` differently other than ensuring that we don't age it off.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
rdblue: Actually, I was wrong. `defaultMaxRefAgeMs` can be `null` if the table doesn't have a max age defined. In that case, we shouldn't age off any refs. The max age is infinite.

So the problem is actually that the instance field is a `long` but it could actually be `null` so this _should_ be `Long`. And the code below should handle the case where max ref age is null and keep the ref.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+        if (now - snapshot.timestampMillis() >= maxRefAgeMs) {
+          expiredRefs.add(name);
+        }
+      }
+    }
+    return expiredRefs;
+  }
+
+  private Set<Long> computeBranchSnapshotsToRetain(SnapshotRef branch, Long expireSnapshotsOlderThan,
+      Integer minSnapshotsToKeep) {
```
rdblue, line 264: I think the timestamp and snapshots to keep values can be primitives. There's probably also no reason to pass the branch in, right? We only need the ref snapshot ID.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+    Snapshot snapshot = base.snapshot(branch.snapshotId());
+    List<Long> ancestorIds = SnapshotUtil.ancestorIds(snapshot, base::snapshot);
+    Set<Long> idsToRetain = Sets.newHashSet();
+    // Everything on this branch should be retained
+    if (minSnapshotsToKeep >= ancestorIds.size()) {
       idsToRetain.addAll(ancestorIds);
     } else {
-      idsToRetain.addAll(ancestorIds.subList(0, minNumSnapshots));
+      idsToRetain.addAll(ancestorIds.subList(0, minSnapshotsToKeep));
+      for (Long ancestorId : ancestorIds) {
```
rdblue, line 273: This is also guaranteed to be non-null.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
-    TableMetadata updateMeta = base.removeSnapshotsIf(snapshot ->
-        idsToRemove.contains(snapshot.snapshotId()) ||
-        (snapshot.timestampMillis() < expireOlderThan && !idsToRetain.contains(snapshot.snapshotId())));
-    List<Snapshot> updateSnapshots = updateMeta.snapshots();
-    List<Snapshot> baseSnapshots = base.snapshots();
-    return updateSnapshots.size() != baseSnapshots.size() ? updateMeta : base;
+    return idsToRetain;
   }
 
+
```
rdblue, line 284: Style: unnecessary newline.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -163,21 +188,100 @@ private TableMetadata internalApply() {
...
+        }
+      }
+    }
+    return expiredRefs;
+  }
+
+  private Set<Long> computeBranchSnapshotsToRetain(SnapshotRef branch, Long expireSnapshotsOlderThan,
+      Integer minSnapshotsToKeep) {
+    Snapshot snapshot = base.snapshot(branch.snapshotId());
+    List<Long> ancestorIds = SnapshotUtil.ancestorIds(snapshot, base::snapshot);
```
rdblue, line 266: You might consider doing this a bit differently. I think it would be more direct to use the Iterable:

```java
  Set<Long> idsToRetain = Sets.newHashSet();
  for (Snapshot snap : SnapshotUtil.ancestorsOf(snapshotId, base::snapshot)) {
    if (idsToRetain.size() < minSnapshotsToKeep || snap.timestampMillis() >= expireOlderThan) {
      idsToRetain.add(snap.snapshotId());
    } else {
      break;
    }
  }

  return idsToRetain;
```

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -195,6 +299,18 @@ public void commit() {
...
     if (cleanExpiredFiles) {
+      TableMetadata updated = ops.refresh();
+      if (updated.refs() != null) {
+        List<SnapshotRef> branches = updated.refs()
+            .values().stream().filter(SnapshotRef::isBranch)
+            .collect(Collectors.toList());
+
+        if (branches.size() > 1) {
+          throw new UnsupportedOperationException(
+              "Deleting expired files when there is more than 1 branch is currently not supported");
```
rdblue, line 310: I don't think that this is supported when there are tags either. For example, I could have snapshots A, B, C, D and need to expire A and B. If A is tagged, then removing B might drop files that are referenced by A. Incremental expiration only works if there is only one branch, `main`.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -195,6 +299,18 @@ public void commit() {
     LOG.info("Committed snapshot changes");
 
     if (cleanExpiredFiles) {
+      TableMetadata updated = ops.refresh();
+      if (updated.refs() != null) {
```
rdblue, line 303: Why would `refs` be null?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
rdblue: The problem is that we have several cases that can't be reliably cleaned up without using the reference set. For example:

```
  A - B - C
  ` - D
```

If A is aged off, there is no way to know that B, C, and D share all of the files from A. So expiring B could delete files that are referenced by D but not C.

To take care of cases like this, we need to do reference set comparison. But rather than implementing that all at once here, we decided to fail snapshot expiration in the short term.

However, one thing does need to be fixed: this check should happen _before_ committing the changes. If we detect that we can't reliably clean up the table, we shouldn't drop the snapshots anyway. So this check should come first in the commit. (@amogh-jahagirdar)

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Yeah I had a comment for fixing this below, this current logic for the max snapshot age, and min snapshots to keep is over-engineered. Here we can set the default retention properties from the table level. Then when we go to compute which branch snapshots to retain we would only fall back to these if the branch properties aren't set. 

When determining which retention property for min snapshots/max snapshot age to use we don't need to distinguish between main and other branches. In either case, we would fall back to using the table property if it's missing for the branch. As you mentioned the only differentiation with main is that the ref shouldn't be aged out.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Correct, I commented this earlier, something I need to fix: https://github.com/apache/iceberg/pull/4578/files#r862970455

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: > defaultMaxRefAgeMs can be null if the table doesn't have a max age defined. In that case, we shouldn't age off any refs. The max age is infinite.

I thought this originally as well, but the table configuration spec https://github.com/apache/iceberg/blob/6f4ba696f8819a580aad3032b86838826c194c6a/docs/tables/configuration.md implies that the default for history.expire.max-ref-age-ms would be Long.MAX_VALUE to indicate that the max ref age is "infinite".

> So the problem is actually that the instance field is a long but it could actually be null so this should be Long. And the code below should handle the case where max ref age is null and keep the ref.

Given the first part, I don't think it can be null (we would default to Long.MAX_VALUE). The comparison we are making in the code to determine if a ref should be expired: `now - snapshot.timestampMillis() >= maxRefAgeMs`, if maxRefAgeMs is Long.MAX_VALUE that effectively means the ref won't be expired. Or is there some overflow case I am missing (I don't think it should be possible) and we want to have null be the default for no expiration.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: @rdblue Thanks for the explanation on why the reachability analysis needs to be done and why we can't delete files if there are multiple branches.

> However, one thing does need to be fixed: this check should happen before committing the changes. If we detect that we can't reliably clean up the table, we shouldn't drop the snapshots anyway. So this check should come first in the commit. 

I was thinking about this before, I'm not sure if the operation to remove snapshots should be coupled with the ability to clean up the files. We know reliably which snapshots which should be removed based on the updated algorithm so it seems a bit unnecessarily constraining to fail the entire operation in case we can't do the deletion of files.

Currently, if I'm not mistaken, deletion of the files themselves is already best effort and can fail, so it's already not a guarantee that we can reliably clean up the table. So following this, I think it makes sense to allow the removal of snapshot metadata and then prevent the deletion of files in the case of multiple branches. Right now we're preventing the operation by throwing an exception which as @kbendick pointed out, to a user would seem like the operation failed but it didn't from a snapshot removal perspective. Another way to handle this is to ignore the cleanFiles flag in case there are multiple branches and warn log that nothing is being deleted because of the current constraint (instead of throwing an exception). @kbendic @rdblue Let me know your thoughts!

amogh-jahagirdar: I re-read and thought about @rdblue example. The problem is I was assuming there would be another branch and so I was thinking we would always be able to tell when we can't delete the files just based off the number of refs but that we could remove the snapshots themselves.

My assumption there isn't valid; the 2nd branch's ref could be aged off independently and there would then be an isolated snapshot lineage. Then later on when the procedure is run  we have 

`
   <main> B - C
   D
`

There's one ref (let's say the B-C lineage is main). When we go to expire there still could be files that B and C does not refer to but D does. There is no longer a 2nd branch, so the removal of files would go through and delete files which D has. 

So then yes, we must prevent the actual snapshot removal operation from going through in case of multiple refs.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Good point, updated!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: No particular reason, using the entrySet is cleaner so updated to do that!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Ah yeah good point, this check is redundant. removed

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Yes, I somehow missed cleaning this up. Updated to just loop over the map entries directly and if it's a branch do the desired computation

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Yes at this point we only really care for the branch for the snapshot Id it refers to, so it's cleaner just to pass in the snapshot. Updated to use primitives (they can safely be primitivs)

amogh-jahagirdar:
Updated, still need to add some tests but @rdblue @kbendick let me know what you think!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Definitley, the iterable approach is much cleaner. Updated

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: It cannot be, we're always maintaining the guarantee that  main will exist so there should be refs. Removed this check

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +173,94 @@ public List<Snapshot> apply() {
 
   private TableMetadata internalApply() {
     this.base = ops.refresh();
+    if (cleanExpiredFiles && base.refs().size() > 1) {
+      throw new UnsupportedOperationException(
+          "Cannot perform snapshot expiration and have clean files set when there is more than 1 ref");
+    }
```
amogh-jahagirdar, line 179: My doubt here is that say we have the following commit graph

> A - B - C 

> ` - D

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +173,101 @@ public List<Snapshot> apply() {
...
+  private boolean multipleLineages(TableMetadata metadata) {
+    List<Snapshot> allSnapshots = metadata.snapshots();
+    Set<Snapshot> mainSnapshots = Sets.newHashSet(SnapshotUtil.ancestorsOf(
+        metadata.ref(SnapshotRef.MAIN_BRANCH).snapshotId(), metadata::snapshot));
+    return allSnapshots.stream()
+        .filter(snapshot -> !mainSnapshots.contains(snapshot))
+        .filter(snapshot -> snapshot.parentId() != null)
+        .filter(snapshot -> metadata.snapshot(snapshot.parentId()) == null ||
+            !mainSnapshots.contains(metadata.snapshot(snapshot.parentId())))
+        .findAny().isPresent();
```
amogh-jahagirdar, line 213: This should filter out any snapshots which are outside the main lineage, which are not staged commits off of main, which have parent snapshots where the parent snapshot is either outside of main OR the parent snapshot is non-existent in the metadata since it was removed. The existence of such a snapshot which meets this criteria means the procedure with cleanExpired cannot be performed.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -190,6 +282,10 @@ public void commit() {
         .onlyRetryOn(CommitFailedException.class)
         .run(item -> {
           TableMetadata updated = internalApply();
+          if (cleanExpiredFiles && (updated.refs().size() > 1 || multipleLineages(updated))) {
```
amogh-jahagirdar, line 285: The check is done after the computation of snapshots to remove but before the commit. I think this has to be done, since we need the updated commit graph after what the expiration would look like if we were to commit it to determine if there will be any detached lineages. If the check is done before the computation the procedure would miss some cases.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar:
```
metadata.snapshot(snapshot.parentId()) == null ||
!mainSnapshots.contains(metadata.snapshot(snapshot.parentId())
```

is a redundant check, since mainSnapshots will never contain anything null !mainSnapshots.contains(metadata.snapshot(snapshot.parentId()) should be sufficient

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +173,101 @@ public List<Snapshot> apply() {
...
+        .map(ref -> base.ref(ref).snapshotId())
+        .collect(Collectors.toSet());
+
+    for (long idToRemove : idsToRemove) {
+      Preconditions.checkArgument(!retainedRefIds.contains(idToRemove),
+          "Cannot expire %s. There is a reference to it", idToRemove);
+    }
+
+    idsToRetain.addAll(retainedRefIds);
+    Set<Long> branchSnapshotsToRetain = computeAllBranchSnapshotsToRetain(base.refs());
```
amogh-jahagirdar, line 193: I don't think using base.refs() is right, we should be using the retained refs. If the refs are aged off, the branch snapshots themselves should not be retained. The inverse is possible by design: the snapshots can be aged off but the references can have their own lifecycle

amogh-jahagirdar: Updated

core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java
```diff
@@ -1204,4 +1204,49 @@ public void testExpireWithDeleteFiles() {
             .build(),
         deletedFiles);
   }
+
```
amogh-jahagirdar, line 1207: Need to add more tests around min snapshots to keep and expire older than for different branches.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
     TableMetadata updated = internalApply();
     List<Snapshot> removed = Lists.newArrayList(base.snapshots());
     removed.removeAll(updated.snapshots());
-
```
rdblue, line 158: Nit: unnecessary whitespace change.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -75,32 +80,40 @@ public void accept(String file) {
...
     long maxSnapshotAgeMs = PropertyUtil.propertyAsLong(
         base.properties(),
         MAX_SNAPSHOT_AGE_MS,
         MAX_SNAPSHOT_AGE_MS_DEFAULT);
-    this.expireOlderThan = System.currentTimeMillis() - maxSnapshotAgeMs;
 
-    this.minNumSnapshots = PropertyUtil.propertyAsInt(
+    this.now = System.currentTimeMillis();
+    this.expireOlderThan = now - maxSnapshotAgeMs;
+    this.defaultMinNumSnapshots = PropertyUtil.propertyAsInt(
```
rdblue, line 107: If you're renaming this, why not also rename `maxSnapshotAgeMs` to `defaultMaxSnapshotAgeMs`?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
...
+
+  private Map<String, SnapshotRef> computeRetainedRefs(Map<String, SnapshotRef> refs) {
+    Map<String, SnapshotRef> retainedRefs = Maps.newHashMap();
+    retainedRefs.put(SnapshotRef.MAIN_BRANCH, refs.get(SnapshotRef.MAIN_BRANCH));
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      String name = refEntry.getKey();
+      SnapshotRef ref = refEntry.getValue();
+      long maxRefAgeMs = ref.maxRefAgeMs() != null ? ref.maxRefAgeMs() : defaultMaxRefAgeMs;
+      Snapshot snapshot = base.snapshot(ref.snapshotId());
+      if (snapshot != null && now - snapshot.timestampMillis() <= maxRefAgeMs) {
```
rdblue, line 238: Nit: this relies on precedence. I prefer to use parens to make precedence clear: `(now - snapshot.timestampMillis()) <= maxRefAgeMs`

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
...
-      idsToRetain.addAll(ancestorIds.subList(0, minNumSnapshots));
+    // Identify refs that should be removed
+    Map<String, SnapshotRef> retainedRefs = computeRetainedRefs(base.refs());
+    Set<Long> retainedRefIds = retainedRefs.entrySet().stream()
+        .map(refEntry -> refEntry.getValue().snapshotId())
+        .collect(Collectors.toSet());
+
+    for (long idToRemove : idsToRemove) {
+      Preconditions.checkArgument(!retainedRefIds.contains(idToRemove),
+          "Cannot expire %s. There is a reference to it", idToRemove);
```
rdblue, line 189: How about `"Cannot expire snapshot %s: referenced by %s %s", idToRemove, ref.isTag() ? "tag" : "branch", refName`?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
...
+
+        branchSnapshotsToRetain.addAll(
+            computeBranchSnapshotsToRetain(ref.snapshotId(), expireSnapshotsOlderThan, minSnapshotsToKeep));
+      }
+    }
+
+    return branchSnapshotsToRetain;
+  }
+
+  private Map<String, SnapshotRef> computeRetainedRefs(Map<String, SnapshotRef> refs) {
```
rdblue, line 230: Nit: this comes between `computeAllBranchSnapshotsToRetain` and its helper method, `computeBranchSnapshotsToRetain`. It's also called first. I'd move it above `computeAllBranchSnapshotsToRetain`.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
...
+
+  private Set<Long> computeAllBranchSnapshotsToRetain(Map<String, SnapshotRef> refs) {
+    Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      SnapshotRef ref = refEntry.getValue();
+      if (ref.isBranch()) {
+        long expireSnapshotsOlderThan;
+        int minSnapshotsToKeep;
+        if (ref.maxSnapshotAgeMs() == null) {
+          expireSnapshotsOlderThan = expireOlderThan;
```
rdblue, line 211: `defaultExpireOlderThan` would help make this more clear.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
...
+    return updatedMetaBuilder.build();
+  }
+
+  private Set<Long> computeAllBranchSnapshotsToRetain(Map<String, SnapshotRef> refs) {
+    Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      SnapshotRef ref = refEntry.getValue();
+      if (ref.isBranch()) {
+        long expireSnapshotsOlderThan;
+        int minSnapshotsToKeep;
```
rdblue, line 209: For these, I'd probably use the ternary operator like you do elsewhere:

```java
long expireOlderThan = ref.maxSnapshotAgeMs() != null ? now - ref.maxSnapshotAgeMs() : defaultExpireOlderThan;
```

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -155,27 +168,93 @@ public List<Snapshot> apply() {
...
+      if (snapshot != null && now - snapshot.timestampMillis() <= maxRefAgeMs) {
+        retainedRefs.put(name, ref);
+      }
+    }
+
+    return retainedRefs;
+  }
+
+  private Set<Long> computeBranchSnapshotsToRetain(long snapshot, long expireSnapshotsOlderThan,
+      int minSnapshotsToKeep) {
```
rdblue, line 247: Style: the starting indentation of arguments should be aligned. It could be at the continuing indent (for all args) or after the method name (for all args).

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -190,6 +269,11 @@ public void commit() {
         .onlyRetryOn(CommitFailedException.class)
         .run(item -> {
           TableMetadata updated = internalApply();
+          if (cleanExpiredFiles && updated.refs().size() > 1) {
+            throw new UnsupportedOperationException(
+                "Cannot remove snapshots and clean files when there is more than 1 ref or multiple lineages");
```
rdblue, line 274: This needs to mention that it is incremental. `"Cannot incrementally clean files for tables with more than 1 ref"`

rdblue (changes requested):
@amogh-jahagirdar, this is really close! I had a few comments but I think overall it is correct. Let me know when the tests are ready to review!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: There can be multiple references for the given id; so to get this information the grouping from retained snapshot id to the list of refs has to be computed. I do think the extra computation is worth it, it is odd to to throw an exception "There is a reference to it" without mentioning what the references are. Will update. I think we can log the list of refs for the given id directly, I think in practice there would not be so many refs for a given snapshot id that it would pollute the logs to log it.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated to use continuing indent for all the args

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Good point, I've updated this and the defaultExpireOlderThan

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated the message to reflect that this is a limitation for specifically incremental file cleanup

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Updated

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Good point, cleaned that up

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: I've updated to maintain the refs associated with a given snapshot ID so that the exception message could be clearer. @rdblue let me know what you think!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Reverted this

amogh-jahagirdar:
Thanks @rdblue I added some more tests and also added some more logic for getting the refs associated with a snapshot, that way if the procedure is called to expire a snapshot which is still referenced, the error will highlight what are the references still pointing to it.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
 
   private TableMetadata internalApply() {
     this.base = ops.refresh();
+    if (base.refs().isEmpty()) {
```
rdblue, line 177: What happens when a table without refs is read? Do we create a `main` ref using the `current-snaphot-id`? We need to make sure that this doesn't prevent older metadata from getting cleaned up.

amogh-jahagirdar: Ah yes, good catch, this should actually be base.snapshots().isEmpty() to early return in case there is no snapshots to expire. Currently, we do not create a main branch when we read the metadata. Only when there are snapshots produced do we actually create the main ref at that moment. 

To prevent the situation where a user uses the new procedure on existing metadata where no new snapshots have been produced using the newer library (and thus no main branch has been created), I think just using the snapshot list size as a source of truth should suffice for any early return.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
...
+    idsToRemove.addAll(base.snapshots().stream().map(Snapshot::snapshotId)
+        .filter(snapshot -> !idsToRetain.contains(snapshot)).collect(Collectors.toSet()));
+    base.refs().keySet().stream().filter(ref -> !retainedRefs.containsKey(ref)).forEach(updatedMetaBuilder::removeRef);
+    updatedMetaBuilder.removeSnapshots(idsToRemove);
+    return updatedMetaBuilder.build();
+  }
+
+  private Map<String, SnapshotRef> computeRetainedRefs(Map<String, SnapshotRef> refs) {
+    Map<String, SnapshotRef> retainedRefs = Maps.newHashMap();
+    retainedRefs.put(SnapshotRef.MAIN_BRANCH, refs.get(SnapshotRef.MAIN_BRANCH));
```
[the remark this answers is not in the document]
rdblue: Actually, I think it would be easier to move this inside the loop:

```java
    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
      String name = refEntry.getKey();
      SnapshotRef ref = refEntry.getValue();
      if (SnapshotRef.MAIN_BRANCH.equals(name)) {
        retainedRefs.put(name, ref);
        continue;
      }
      ...
    }
```

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
...
+          "Cannot expire %s. Still referenced by refs: %s", idToRemove, refsForId);
     }
 
-    TableMetadata updateMeta = base.removeSnapshotsIf(snapshot ->
-        idsToRemove.contains(snapshot.snapshotId()) ||
-        (snapshot.timestampMillis() < expireOlderThan && !idsToRetain.contains(snapshot.snapshotId())));
-    List<Snapshot> updateSnapshots = updateMeta.snapshots();
-    List<Snapshot> baseSnapshots = base.snapshots();
-    return updateSnapshots.size() != baseSnapshots.size() ? updateMeta : base;
+    idsToRetain.addAll(retainedIdToRefs.keySet());
```
rdblue, line 197: Why not add the snapshot ID in the loop above when you're constructing `retainedIdToRefs`?

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
...
+      }
+    }
+
+    return retainedRefs;
+  }
+
+  private Set<Long> computeAllBranchSnapshotsToRetain(Map<String, SnapshotRef> refs) {
+    Set<Long> branchSnapshotsToRetain = Sets.newHashSet();
+    for (Map.Entry<String, SnapshotRef> refEntry : refs.entrySet()) {
+      SnapshotRef ref = refEntry.getValue();
```
rdblue, line 232: You could just iterate over `refs.values()`.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
...
-        (snapshot.timestampMillis() < expireOlderThan && !idsToRetain.contains(snapshot.snapshotId())));
-    List<Snapshot> updateSnapshots = updateMeta.snapshots();
-    List<Snapshot> baseSnapshots = base.snapshots();
-    return updateSnapshots.size() != baseSnapshots.size() ? updateMeta : base;
+    idsToRetain.addAll(retainedIdToRefs.keySet());
+    Set<Long> branchSnapshotsToRetain = computeAllBranchSnapshotsToRetain(retainedRefs);
+    idsToRetain.addAll(branchSnapshotsToRetain);
+    TableMetadata.Builder updatedMetaBuilder = TableMetadata.buildFrom(base);
+    idsToRemove.addAll(base.snapshots().stream().map(Snapshot::snapshotId)
+        .filter(snapshot -> !idsToRetain.contains(snapshot)).collect(Collectors.toSet()));
```
rdblue, line 202: This statement is hard to read. When using streams, put each stream operation on its own line. It's also hard to read since this is using `addAll` after `collect`. Instead, you can directly add to `idsToRemove`:

```java
  base.snapshots().stream()
      .map(...)
      .filter(...)
      .forEach(idsToRemove::add)
```

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
...
+    base.refs().keySet().stream().filter(ref -> !retainedRefs.containsKey(ref)).forEach(updatedMetaBuilder::removeRef);
```
rdblue, line 203: This should also use a separate line for each stream operation.

rdblue (changes requested):
@amogh-jahagirdar, this is really close. Mostly just style at this point, but I also think the handling of the main ref is a blocker.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Good point, we should definitely just be adding there to avoid having to iterate through all the retained snapshots again!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Ah yes my bad, I missed formatting this and the stream is much cleaner if we do the foreach(add) approach.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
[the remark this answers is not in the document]
amogh-jahagirdar: The issue of the case where there is no main branch yet has come up in a few cases, I'm thinking it makes more sense that when parsing the metadata in case there is no main branch, we populate it with the current snapshot id. That way much of the API/core classes can operate under the assumption that main exists. Thoughts @rdblue?

I'll raise another PR for this, and we can discuss there if that makes sense (if it doesn't we can close it).

But regardless of that PR,  the check in the loop does look cleaner and tackles the case where it doesn't exist so I'll just do that.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Yeah I think we can actually just let this method take in the list of refs directly, and then just iterate over that.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +174,89 @@ public List<Snapshot> apply() {
...
+    updatedMetaBuilder.removeSnapshots(idsToRemove);
```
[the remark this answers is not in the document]
amogh-jahagirdar: Good point, the code becomes more readable that way since a reader doesn't have to do much scanning to see how idsToRemove is used. Updated.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
amogh-jahagirdar: Ah yes fixed!

amogh-jahagirdar:
@rdblue I agree, the handling of the main ref is certainly a blocker, I've updated the logic so that it handles the case where main does not exist. That being said, I am about to publish a PR for updating the parsing of metadata so that main will always be set. In case it's not found, we set it to the current snapshot ID. I think this is useful so that other areas in the code can reasonably operate with this assumption.

amogh-jahagirdar:
Looks like TestBaseIncrementalAppendScan.testMultipleRootSnapshots is failing, which looks to be recently added as part of the Incremental scanning work. Looking into it

amogh-jahagirdar:
Ok, I think I see what's going on. The failing incremental scan test can't find a staged snapshot because the updated expiration logic actually will remove it.

Expire snapshots previously would retain staged snapshots so long as it's not outside of the "expire older than" age.  However, in this updated logic, we're not considering that for *unreferenced* staged snapshots. Age retention is only applied to snapshots which are part of branches. I think we definitely want to preserve the existing behavior to avoid deleting snapshots that people may expect should be retained, so I will add logic for that.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +175,120 @@ public List<Snapshot> apply() {
...
+
+    return branchSnapshotsToRetain;
+  }
+
+  private Set<Long> computeDanglingSnapshotsToRetain(Collection<SnapshotRef> refs) {
+    Set<Long> snapshotsPartOfBranches = Sets.newHashSet();
+    for (SnapshotRef ref : refs) {
+      if (ref.isBranch()) {
+        Iterable<Snapshot> branchSnapshots = SnapshotUtil.ancestorsOf(ref.snapshotId(), base::snapshot);
+        branchSnapshots.forEach(snapshot -> snapshotsPartOfBranches.add(snapshot.snapshotId()));
```
amogh-jahagirdar, line 267: We can be more efficient (time wise, at the tradeoff of memory) by maintaining the mapping between a branch and all the ancestors for the branch so that it can be reused across both the dangling snapshots and the branch snapshots to retain computation. Otherwise we have to get the ancestors multiple times.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +175,138 @@ public List<Snapshot> apply() {
...
+        }
+      } else {
+        LOG.warn("Removing invalid ref {}: snapshot {} does not exist", name, ref.snapshotId());
+      }
+    }
+
+    return retainedRefs;
+  }
+
+  private Set<Long> computeDanglingSnapshotsToRetain(Map<SnapshotRef, Set<Long>> branchSnapshots) {
```
amogh-jahagirdar, line 263: @rdblue Let me know what you think of this logic to handle the case of unreferenced staged snapshots which should still be retained based on the default expiration age. I reorganized the code to minimize the computation of branch snapshots. I also should add a test for this case.

rdblue:
@amogh-jahagirdar, I agree that we shouldn't remove unreferenced snapshots until they are older than the default expiration timestamp. Good catch!

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +176,143 @@ public List<Snapshot> apply() {
...
-    if (minNumSnapshots >= ancestorIds.size()) {
-      idsToRetain.addAll(ancestorIds);
-    } else {
-      idsToRetain.addAll(ancestorIds.subList(0, minNumSnapshots));
+
+    // Compute the snapshots for each reference
+    Map<SnapshotRef, Set<Long>> refSnapshots = computeRefSnapshots(base.refs().values());
+
+    // Identify unreferenced snapshots which should be retained
+    Set<Long> unreferencedSnapshotsToRetain = computeUnreferencedSnapshotsToRetain(refSnapshots);
```
rdblue, line 189: I think this should be passed a single set of retained snapshots and look for any others, rather than using a complex map. Just do this analysis last, after you know all the other snapshots that are retained for other reasons.

amogh-jahagirdar: Yeah I'm still thinking if there's a way to simplify this. 

I think we'll need the structure for maintaining what are the referenced snapshots ("referenced" meaning for a branch all it's ancestors, and for a tag, the snapshot it refers to). I don't think we can construct the snapshot retention set and then after that look through all the table's snapshots , and correctly derive which snapshots are the unreferenced snapshots which should still be retained based on expiration age.

The reason is some of the table's snapshots may no longer need to be retained due to branch retention policy but if we go through those snapshots and they're still within defaultExpirationAge we'd retain them which I don't think we want since the branch has it's own lifecycle.  I think if a snapshot was part of a branch which expired due to it's own retention policy then we don't want to retain it on the basis of the defaultExpireOlderThan.   

**In other words, in this approach I don't think we can differentiate between a snapshot which should actually be removed based on branch retention or if it's a snapshot which was never part of any reference.**

I think we need to be able to identify what are the snapshots which are not part of any branch lineage or are tagged and for that we need a set of all the snapshots which encapsulate all the current branch lineages. We do the grouping so that it can be reused when determining which branch snapshots to keep (so that the ancestors don't have to be iterated through again). Let me know what you think @rdblue  !

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -161,21 +176,143 @@ public List<Snapshot> apply() {
...
+        .forEach(updatedMetaBuilder::removeRef);
+
+    return updatedMetaBuilder.build();
+  }
+
+  /**
+   * Helper to compute the mapping of a ref to its snapshots. If it's a branch, the snapshots is an ordered set
+   * of all the snapshots on the branch. If it's a tag, the snapshot is a set of the single snapshot the tag refers to
+   */
+  private Map<SnapshotRef, Set<Long>> computeRefSnapshots(Collection<SnapshotRef> refs) {
```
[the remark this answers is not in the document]
amogh-jahagirdar: Yeah it did feel weird to use the ref as a key, I did mostly to avoid having to pass in the entire ref map in other helper functions and do lookups in those functions, but that's not something to optimize for. Updated to using the name instead.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
rdblue: I agree with you on the behavior, but I think there's a simpler way to implement it. I went ahead and added it in https://github.com/amogh-jahagirdar/iceberg/pull/1, and fixed some tests. Can you review and merge that PR so we can get this in?

amogh-jahagirdar: Merged! I think the new approach simplifies the code. what I was originally going for was not having to traverse the branch ancestors multiple times, which is why the map was retained but that may be overkill. Thanks for adding the tests!

rdblue:
Thanks, @amogh-jahagirdar!

namrathamyske:
@amogh-jahagirdar @rdblue With this PR, the breaking change is to set cleanExpiredFiles to false if the table has more than one ref. Is this addressed anywhere?

amogh-jahagirdar:
@namrathamyske 

> With this PR, the breaking change is to set cleanExpiredFiles to false if the table has more than one ref.

No, the operation will fail if cleanExpiredFiles is set and there are multiple retained refs. It shouldn't break the existing linear history case and should be backwards compatible but please report in case you ran into any issues! 

See this [comment](https://github.com/apache/iceberg/pull/4578#discussion_r863337777) . To do a correct removal of files with multiple references, a reachability analysis needs to be done which requires quite a bit more implementation.

core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
```diff
@@ -190,6 +301,10 @@ public void commit() {
         .onlyRetryOn(CommitFailedException.class)
         .run(item -> {
           TableMetadata updated = internalApply();
+          if (cleanExpiredFiles && updated.refs().size() > 1) {
+            throw new UnsupportedOperationException("Cannot incrementally clean files for tables with more than 1 ref");
```
zinking, line 305: @amogh-jahagirdar  I have a simple use case blocked by this line. 

I am tagging one of the snapshots on the mainline as *never expire*, but as soon as I tag it, the normal expire procedure (expireOlderThan) stopped working, as there are two refs now.

is the constraint correct here ? because in my scenario, I am merely indicating SKIP this snapshot, with other terms stay the same.

zinking: I think, if there's only one branch main, and tags are all on main, it is safe to carry on. 
is that correct ?

PR #3095, opened 2021-09-10
Flink: flink read iceberg upsert data use streaming mode

hameizi:
Now, flink read iceberg use streaming mode ignore 'overwrite' snapshot, so user can't read the delete data real-time.
This PR first emit the eqdelete as delete rowdata(-U), and then emit the rowdata(+I) that join adddata and posdelete by function applyPosdelete. And then flink can keep one primary table from read iceberg primary table.

PR #4580, opened 2022-04-18
API: Introduce a new IncrementalAppendScan interface

stevenzwu:
During review of Flink FLIP-27 source PR #4329 , we agreed that the streaming start strategy should be inclusive. Hence we would need the `TableScan#appendsBetween` to support nullable `fromSnapshotId`. Right now, `fromSnapshotId` is a primitive type of `long`.
https://github.com/apache/iceberg/pull/4329#discussion_r834877847

Initially, @rdblue and I were thinking just overload `appendsBetween` with a `Long fromSnapshotId`. That would cause compiling error due to ambiguity of type resolution. As we want to maintain binary backward compatibility, I tried to add a new method name `appendsInRange` in PR #4529 , which is not as an intuitive name as `appendsBetween`.

This is a different direction with PR #4529 that is suggested by @rdblue. Instead of modifying `TableScan` interface, maybe we can consider introduce a new `IncrementalTableScan` interface and `Table#newIncrementalScan` method.

- We can decouple regular `TableScan` and `IncrementalTableScan`. This also help avoids the need of `UnsupportedOperationException` for some `TableScan` methods
- We may have more complex incremental scan in the future like the CDC data that can be added to the new `IncrementalTableScan` API

To avoid code duplication, a new super interface of `Scan<T extends Scan>` was extracted as the parent of the current `TableScan` and the new `IncrementalTableScan`. I ran `japi-compliance-checker` too check binary compatibility after the interface refactoring. Will attach the result

### DIFF
--- a/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
+++ b/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
@@ -26,9 +26,9 @@ package org.apache.iceberg;
 public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
 
   /**
-   * Optional. if not set, null value will be used for the start snapshot id.
-   * That would include the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)},
-   * as its parent snapshot id is null which matches the null start snapshot id
+   * Optional. if from snapshot id (inclusive or exclusive) is not provided,
+   * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
+   * will be included as the from snapshot.
    *
    * @param fromSnapshotId the start snapshot id inclusive
    * @return an incremental table scan from {@code fromSnapshotId} inclusive
@@ -36,9 +36,9 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
   IncrementalAppendScan fromSnapshotInclusive(long fromSnapshotId);
 
   /**
-   * Optional. if not set, null value will be used for the start snapshot id.
-   * That would include the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)},
-   * as its parent snapshot id is null which matches the null start snapshot id
+   * Optional. if from snapshot id (inclusive or exclusive) is not provided,
+   * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
+   * will be included as the from snapshot.
    *
    * @param fromSnapshotId the start snapshot id (exclusive)
    * @return an incremental table scan from {@code fromSnapshotId} exclusive
@@ -46,7 +46,7 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
   IncrementalAppendScan fromSnapshotExclusive(long fromSnapshotId);
 
   /**
-   * Optional. if not set, current table snapshot id is used as the end snapshot id
+   * Required
    *
    * @param toSnapshotId the end snapshot id (inclusive)
    * @return an incremental table scan up to {@code toSnapshotId} inclusive
--- a/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
+++ b/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
@@ -29,11 +29,10 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
    * Refine the incremental scan with the start snapshot inclusive.
    * <p>
    * If the start snapshot (inclusive or exclusive) is not provided,
-   * the oldest snapshot will be used as the start snapshot.
-   * Oldest snapshot is a snapshot whose parent-snapshot-id is null.
-   * If there are multiple snapshots has parent-snapshot-id as null
-   * (e.g. due to disjointed lineage history), snapshot timestamp will
-   * be used as a tiebreaker.
+   * a snapshot with no parent is the starting snapshot.
+   * If there are multiple snapshots with no parent
+   * (e.g. due to disjointed lineage history),
+   * snapshot timestamp is used as a tiebreaker.
    *
    * @param fromSnapshotId the start snapshot id inclusive
    * @return an incremental table scan from {@code fromSnapshotId} inclusive
@@ -46,11 +45,10 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
    * Refine the incremental scan with the start snapshot exclusive.
    * <p>
    * If the start snapshot (inclusive or exclusive) is not provided,
-   * the oldest snapshot will be used as the start snapshot.
-   * Oldest snapshot is a snapshot whose parent-snapshot-id is null.
-   * If there are multiple snapshots has parent-snapshot-id as null
-   * (e.g. due to disjointed lineage history), snapshot timestamp will
-   * be used as a tiebreaker.
+   * a snapshot with no parent is the starting snapshot.
+   * If there are multiple snapshots with no parent
+   * (e.g. due to disjointed lineage history),
+   * snapshot timestamp is used as a tiebreaker.
    *
    * @param fromSnapshotId the start snapshot id (exclusive)
    * @return an incremental table scan from {@code fromSnapshotId} exclusive
--- a/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
+++ b/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
@@ -21,19 +21,29 @@
 package org.apache.iceberg;
 
 /**
- * API for configuring an incremental table scan
+ * API for configuring an incremental table scan for appends only snapshots
  */
-public interface IncrementalScan extends Scan<IncrementalScan> {
+public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
 
   /**
    * Optional. if not set, null value will be used for the start snapshot id.
-   * That would include the oldest ancestor of the {@link IncrementalScan#toSnapshotId(long)},
+   * That would include the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)},
+   * as its parent snapshot id is null which matches the null start snapshot id
+   *
+   * @param fromSnapshotId the start snapshot id inclusive
+   * @return an incremental table scan from {@code fromSnapshotId} inclusive
+   */
+  IncrementalAppendScan fromSnapshotInclusive(long fromSnapshotId);
+
+  /**
+   * Optional. if not set, null value will be used for the start snapshot id.
+   * That would include the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)},
    * as its parent snapshot id is null which matches the null start snapshot id
    *
    * @param fromSnapshotId the start snapshot id (exclusive)
    * @return an incremental table scan from {@code fromSnapshotId} exclusive
    */
-  IncrementalScan fromSnapshotId(long fromSnapshotId);
+  IncrementalAppendScan fromSnapshotExclusive(long fromSnapshotId);
 
   /**
    * Optional. if not set, current table snapshot id is used as the end snapshot id
@@ -41,17 +51,5 @@ public interface IncrementalScan extends Scan<IncrementalScan> {
    * @param toSnapshotId the end snapshot id (inclusive)
    * @return an incremental table scan up to {@code toSnapshotId} inclusive
    */
-  IncrementalScan toSnapshotId(long toSnapshotId);
-
-  /**
-   * Only interested in snapshots with append operation
-   */
-  IncrementalScan appendsOnly();
-
-  /**
-   * Ignore snapshots with overwrite operation.
-   *
-   * Default behavior for incremental scan fails if there are overwrite operations in the incremental snapshot range
-   */
-  IncrementalScan ignoreOverwrites();
+  IncrementalAppendScan toSnapshot(long toSnapshotId);
 }
--- a/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
+++ b/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
@@ -26,9 +26,11 @@ package org.apache.iceberg;
 public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
 
   /**
-   * Optional. if from snapshot id (inclusive or exclusive) is not provided,
+   * Refine the incremental scan with the start snapshot inclusive.
+   * <p>
+   * If the start snapshot (inclusive or exclusive) is not provided,
    * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
-   * will be included as the from snapshot.
+   * will be included as the start snapshot.
    *
    * @param fromSnapshotId the start snapshot id inclusive
    * @return an incremental table scan from {@code fromSnapshotId} inclusive
@@ -36,9 +38,11 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
   IncrementalAppendScan fromSnapshotInclusive(long fromSnapshotId);
 
   /**
-   * Optional. if from snapshot id (inclusive or exclusive) is not provided,
+   * Refine the incremental scan with the start snapshot exclusive.
+   * <p>
+   * If the start snapshot (inclusive or exclusive) is not provided,
    * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
-   * will be included as the from snapshot.
+   * will be included as the start snapshot.
    *
    * @param fromSnapshotId the start snapshot id (exclusive)
    * @return an incremental table scan from {@code fromSnapshotId} exclusive
@@ -46,7 +50,9 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
   IncrementalAppendScan fromSnapshotExclusive(long fromSnapshotId);
 
   /**
-   * Required
+   * Refine the incremental scan with the end snapshot inclusive.
+   * <p>
+   * If the end snapshot is not provided, the current table snapshot will be used.
    *
    * @param toSnapshotId the end snapshot id (inclusive)
    * @return an incremental table scan up to {@code toSnapshotId} inclusive
--- a/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
+++ b/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
@@ -29,11 +29,16 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
    * Refine the incremental scan with the start snapshot inclusive.
    * <p>
    * If the start snapshot (inclusive or exclusive) is not provided,
-   * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
-   * will be included as the start snapshot.
+   * the oldest snapshot will be used as the start snapshot.
+   * Oldest snapshot is a snapshot whose parent-snapshot-id is null.
+   * If there are multiple snapshots has parent-snapshot-id as null
+   * (e.g. due to disjointed lineage history), snapshot timestamp will
+   * be used as a tiebreaker.
    *
    * @param fromSnapshotId the start snapshot id inclusive
    * @return an incremental table scan from {@code fromSnapshotId} inclusive
+   * @throws IllegalArgumentException if the start snapshot is not an ancestor
+   * of the end snapshot
    */
   IncrementalAppendScan fromSnapshotInclusive(long fromSnapshotId);
 
@@ -41,11 +46,16 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
    * Refine the incremental scan with the start snapshot exclusive.
    * <p>
    * If the start snapshot (inclusive or exclusive) is not provided,
-   * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
-   * will be included as the start snapshot.
+   * the oldest snapshot will be used as the start snapshot.
+   * Oldest snapshot is a snapshot whose parent-snapshot-id is null.
+   * If there are multiple snapshots has parent-snapshot-id as null
+   * (e.g. due to disjointed lineage history), snapshot timestamp will
+   * be used as a tiebreaker.
    *
    * @param fromSnapshotId the start snapshot id (exclusive)
    * @return an incremental table scan from {@code fromSnapshotId} exclusive
+   * @throws IllegalArgumentException if the start snapshot is not an ancestor
+   * of the end snapshot
    */
   IncrementalAppendScan fromSnapshotExclusive(long fromSnapshotId);
 
--- a/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
+++ b/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
@@ -28,11 +28,8 @@ public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
   /**
    * Refine the incremental scan with the start snapshot inclusive.
    * <p>
-   * If the start snapshot (inclusive or exclusive) is not provided,
-   * a snapshot with no parent is the starting snapshot.
-   * If there are multiple snapshots with no parent
-   * (e.g. due to disjointed lineage history),
-   * snapshot timestamp is used as a tiebreaker.
+   * If neither {@link #fromSnapshotInclusive(long)} or {@link #fromSnapshotExclusive(long)} is provided,
+   * start snapshot inclusive is defaulted to the oldest ancestor of the end snapshot.
    *
    * @param fromSnapshotId the start snapshot id inclusive
    * @return an incremental table scan from {@code fromSnapshotId} inclusive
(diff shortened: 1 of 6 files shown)


stevenzwu:
Here is the japi-compliance-checker report for iceberg-api jar before and after this change

[api-compatibility-report.pdf](https://github.com/apache/iceberg/files/8506868/api-compatibility-report.pdf)

stevenzwu:
@rdblue @openinx @yittg @aokolnychyi @flyrain @flashJd can you please take a look at this approach and see if it makes sense?

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -107,24 +68,6 @@ default TableScan select(String... columns) {
     return select(Lists.newArrayList(columns));
```
stevenzwu, line 68: couldn't move this select method to the `Scan` interface, as API compatibility check tool shows that the return type of `TableScan` is changed to `Scan` probably due to varargs.

stevenzwu:
cc @hameizi the author of PR #3095

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
```diff
@@ -0,0 +1,57 @@
...
+   * Only interested in snapshots with append operation
+   */
+  IncrementalTableScan appendsOnly();
+
+  /**
+   * Ignore snapshots with overwrite operation.
+   *
+   * Default behavior for incremental scan fails if there are overwrite operations in the incremental snapshot range
+   */
+  IncrementalTableScan ignoreOverwrites();
```
stevenzwu, line 56: For V2 tables, incremental scan is potentially interested in append, overwrite, delete operations. append is always included. I guess we need to discuss what is the natural API to control overwrite and delete.

this API of `ignoreOverwrites` might be too restrictive. then we also might need to expose an `includeOverwrites`. I am wondering we should api like
```
include(DataOperations... operations) // default is append only. for CDC read, overwrite and delete can be added
fail(DataOperations... operations) // default is fail nothing
```

stevenzwu: or maybe we should just need the `appendsOnly` API above. otherwise, overwrite and delete snapshots are also included automatically for V2 tables.

hameizi: > or maybe we should just need the `appendsOnly` API above. otherwise, overwrite and delete snapshots are also included automatically for V2 tables.

 I agree. Because i think for V2 table include overwrite and delete snapshots should be a default behavior.

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -0,0 +1,84 @@
...
+  T select(Collection<String> columns);
+
+  /**
+   * Create a new {@link TableScan} from the results of this filtered by the {@link Expression}.
+   *
+   * @param expr a filter expression
+   * @return a new scan based on this with results filtered by the expression
+   */
+  T filter(Expression expr);
+}
```
hameizi, line 84: I think planFiles() is necessary. Just for TableScan and IncrementalTableScan there should be different logic.

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * @param schema a projection schema
+   * @return a new scan based on this with the given projection
+   */
+  T project(Schema schema);
+
+  /**
+   * Create a new {@link TableScan} from this that, if data columns where selected
+   * via {@link #select(java.util.Collection)}, controls whether the match to the schema will be done
+   * with case sensitivity.
```
yittg, line 52: how about documenting the default value?

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Create a new {@link TableScan} from this that will read the given data columns. This produces
+   * an expected schema that includes all fields that are either selected or used by this scan's
+   * filter expression.
+   *
+   * @param columns column names from the table's schema
+   * @return a new scan based on this with the given projection columns
+   */
+  T select(Collection<String> columns);
```
yittg, line 75: Is Collection enough? The order of columns maybe matter?

api/src/main/java/org/apache/iceberg/Scan.java
stevenzwu: agree. I forgot to move a few other methods to the super interface

api/src/main/java/org/apache/iceberg/Scan.java
stevenzwu: This is moved from TableScan. for compatibility reason, we can't change the type/signature.

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Collection;
+import java.util.concurrent.ExecutorService;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.io.CloseableIterable;
+
+/**
+ * Scan objects are immutable and can be shared between threads. Refinement methods, like
+ * {@link #select(Collection)} and {@link #filter(Expression)}, create new TableScan instances.
+ */
+interface Scan<T extends Scan> {
```
openinx, line 32: Maybe it's better to define this `Scan` as `interface Scan<T>`, and add a javadoc saying `<T> the concrete Java class of a Scan instance.`. ( I commented this because it seems to be wired to introduce a IDEA warning saying `Raw use of parameterized class 'Scan' `)

There is a good example here:
https://github.com/apache/iceberg/blob/90225d6c9413016d611e2ce5eff37db1bc1b4fc5/api/src/main/java/org/apache/iceberg/ContentFile.java#L29

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Optional. if not set, null value will be used for the start snapshot id.
+   * That would include the oldest ancestor of the {@link IncrementalTableScan#toSnapshotId(long)},
+   * as its parent snapshot id is null which matches the null start snapshot id
+   *
+   * @param fromSnapshotId the start snapshot id (exclusive)
+   * @return an incremental table scan from {@code fromSnapshotId} exclusive
+   */
+  IncrementalTableScan fromSnapshotId(long fromSnapshotId);
```
openinx, line 36: Will it be more generic to add a `inclusive` flag in this `fromSnapshotId` method ?  (So that we can meet the requirement for both including & excluding fromSnapshotId incremental scan )

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
```diff
@@ -0,0 +1,57 @@
...
+ * under the License.
+ */
+
+
+package org.apache.iceberg;
+
+/**
+ * API for configuring an incremental table scan
+ */
+public interface IncrementalTableScan extends Scan<IncrementalTableScan> {
```
openinx, line 26: Nit: How about just naming it `IncrementalScan` ? In my thought, the `IncrementalTableScan` must be created from the `Table` API, so in theory we can remove the redundent keyword `Table` in the interface..

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -53,6 +53,17 @@ default String name() {
...
   TableScan newScan();
 
+  /**
+   * Create a new {@link IncrementalTableScan scan} for this table.
+   * <p>
+   * Once a scan is created, it can be refined to project columns and filter data.
+   *
+   * @return an incremental scan for this table
+   */
+  default IncrementalTableScan newIncrementalScan() {
```
openinx, line 63: Changing the `IncrementalTableScan` to `IncrementalScan` is matching the `newIncrementalScan` method name ?

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -148,7 +89,9 @@ default TableScan select(String... columns) {
    * @return a table scan which can read append data from {@code fromSnapshotId}
    * exclusive and up to {@code toSnapshotId} inclusive
    */
-  TableScan appendsBetween(long fromSnapshotId, long toSnapshotId);
+  default TableScan appendsBetween(long fromSnapshotId, long toSnapshotId) {
+    throw new UnsupportedOperationException("Incremental scan is not supported");
```
openinx, line 93: We will need to keep this implementation for at least one minor release ?

openinx: According to the apache project compatibility rules..

api/src/main/java/org/apache/iceberg/Scan.java
yittg: Then how about `Scan<T extends Scan<T>>`?

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
hameizi: agree. Base on this, is there should add method like useSnapshot to process just one snapshot. I think it's useful for read iceberg table in streaming mode.

stevenzwu: actually, the `fromSnapshotId` is exclusive behavior (from, to] for the incremental scan, as the `toSnapshotId will become the `fromSnapshotId` in the next scan.

For the inclusive behavior, we were mainly talking about the starting strategy. E.g., if we said a specific start snapshot id, we want to include the files in this snapshot (if append). To support that, we need the incremental scan to support nullable `fromSnapshotId`, as we will just pass in the parent snapshot id (which can be null). That wasn't possible with the `TableScan#appendsBetween(long fromSnapshotId, long toSnapshotId)`.

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
stevenzwu: makes sense

api/src/main/java/org/apache/iceberg/TableScan.java
stevenzwu: this change is not strictly required for this change. I am ok to revert it. I added it here as I was thinking about removing the duplicate code of `UnsupportedOperationException` in many `TableScan` impl classes.

We can mark those two `appends` methods as deprecated once the new IncrementalScan impls are ready. Yes, we can follow the compatibility rules.

api/src/main/java/org/apache/iceberg/Scan.java
stevenzwu: Will try @yittg 's suggestion. I was copying the builder inheritance pattern. It wasn't done fully correctly.

api/src/main/java/org/apache/iceberg/Scan.java
stevenzwu: it was moved from TableScan as it is. I can add the default value to the Javadoc.

flyrain:
Thanks @stevenzwu for the PR. I’m OK with the change, but I doubt if CDC can use the the interface IncrementalTableScan. Basically CDC requires much finer control of planning, check my CDC PR(https://github.com/apache/iceberg/pull/4539) for more details. We can keep evolving on the interface IncrementalTableScan to make it suitable for cdc in the future. It is hard to connect them at this moment.

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
flyrain: For the logical completeness of filtering snapshot type, should we have the following 3 methods?
1. ignoreAppend
2. ignoreOverwrite
3. ignoreDelete

And the `appendsOnly` is basically a combination of `ignoreOverwrite` and `ignoreDelete`

stevenzwu:
@flyrain This is just a starting point. I am sure the current `IncrementalScan` interface is NOT good for the CDC read today, which needs more complex planning control. That was also part of the motivation when Ryan suggested. We can enhance the `IncrementalScan` interface in the future.

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
stevenzwu: agree the 3 ignore methods are complete and flexible? wondering if we need the combinations or we only need two modes (appendsOnly and append + overwrite + delete). E.g., does `append + overwrite` or `overwrite + delete` ever make sense?

Do we need fail APIs (like `failOverwrite`)? My current take is no. Let's just skip the snapshots (not interested).

api/src/main/java/org/apache/iceberg/Scan.java
stevenzwu: Updated the code according to Yi's suggestion. japi-compliance-checker is still good.

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
flyrain: Yes, it is flexible. `append + overwrite` makes sense for user who only want to get inserted rows with some additional filtering. `overwrite + delete` makes sense for getting only deleted rows.
I'm not aware of a use case with `failOverwrite`. We may skip it now.

chenjunjiedada: How about dataBetween(Long startSnapshotId, long endSnapshot, List<RowKind> rowKinds)?

api/src/main/java/org/apache/iceberg/IncrementalScan.java
```diff
@@ -0,0 +1,57 @@
...
+   * Only interested in snapshots with append operation
+   */
+  IncrementalScan appendsOnly();
+
+  /**
+   * Ignore snapshots with overwrite operation.
+   *
+   * Default behavior for incremental scan fails if there are overwrite operations in the incremental snapshot range
+   */
+  IncrementalScan ignoreOverwrites();
```
chenjunjiedada, line 56: Maybe we could add another option `IncrementalScan rowKinds(RowKind... rowKinds)` to support CDC case.

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
stevenzwu: @chenjunjiedada `RowKind` is a Flink API. 

Based on the conversation, it seems most people prefer the fluent style API for the scan builder, like `fromSnapshotId(long fromSnapshotId)`

api/src/main/java/org/apache/iceberg/IncrementalScan.java
stevenzwu: `RowKind` is a Flink API. we can't use it here.

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
chenjunjiedada: I meant we could borrow the `RowKind` definition of the produced data. Like what @flyrain mentioned before. We could use `+I` to target the `append` + `overwrite` with some filter, use `-D, -U` to target `delete` and some data in `overwrite`.

api/src/main/java/org/apache/iceberg/IncrementalScan.java
chenjunjiedada: Can we borrow the definition of data from Flink side? The incremental scan actual target that four kinds of data, right?

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -148,7 +89,9 @@ default TableScan select(String... columns) {
    * @return a table scan which can read append data from {@code fromSnapshotId}
    * exclusive and up to {@code toSnapshotId} inclusive
    */
-  TableScan appendsBetween(long fromSnapshotId, long toSnapshotId);
```
rdblue, line 151: I think we should probably deprecate this because we want people to move to incremental. (Eventually)

api/src/main/java/org/apache/iceberg/IncrementalScan.java
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @param toSnapshotId the end snapshot id (inclusive)
+   * @return an incremental table scan up to {@code toSnapshotId} inclusive
+   */
+  IncrementalScan toSnapshotId(long toSnapshotId);
+
+  /**
+   * Only interested in snapshots with append operation
+   */
+  IncrementalScan appendsOnly();
```
rdblue, line 49: As we talked about, I think it makes sense to remove these two methods since the default for scanning appends is to ignore deletes and overwrites and to read only append snaoshots.

api/src/main/java/org/apache/iceberg/IncrementalScan.java
```diff
@@ -0,0 +1,57 @@
...
+ * under the License.
+ */
+
+
+package org.apache.iceberg;
+
+/**
+ * API for configuring an incremental table scan
+ */
+public interface IncrementalScan extends Scan<IncrementalScan> {
```
rdblue, line 26: `IncrementalAppendScan`?

api/src/main/java/org/apache/iceberg/IncrementalTableScan.java
rdblue: If we want to add exclusive behavior, then we should add alternative methods like `afterSnapshotId`. To me, `fromSnapshotId` should be inclusive of the snapshot that is identified. We can come up with better names for these, like `fromSnapshotInclusive` and `fromSnapshotExclusive` if you like those better.

api/src/main/java/org/apache/iceberg/TableScan.java
stevenzwu: yes, once new IncrementAppendScan is implemented, we can mark these two appends methods as deprecated

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -53,6 +53,17 @@ default String name() {
...
   TableScan newScan();
 
+  /**
+   * Create a new {@link IncrementalAppendScan scan} for this table.
+   * <p>
+   * Once a scan is created, it can be refined to project columns and filter data.
+   *
+   * @return an incremental scan for appends only snapshots
+   */
+  default IncrementalAppendScan newIncrementalAppendScan() {
```
stevenzwu, line 63: @flyrain @aokolnychyi @openinx @chenjunjiedada @yittg @hameizi After discussing with @rdblue, we think it is probably cleaner to have separate newScan method for appends only and CDC read. 

In the future, we can add `Table#newIncrementalChangelogScan` and `IncrementalChangelogScan` interface

api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
```diff
@@ -0,0 +1,55 @@
...
+   * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
+   * will be included as the from snapshot.
+   *
+   * @param fromSnapshotId the start snapshot id (exclusive)
+   * @return an incremental table scan from {@code fromSnapshotId} exclusive
+   */
+  IncrementalAppendScan fromSnapshotExclusive(long fromSnapshotId);
+
+  /**
+   * Required
```
stevenzwu, line 49: @rdblue I marked the `toSnapshot` as required. basically, we will not support the use case of old `TableScan#appendsAfter`. The reason is that incremental scan typically need to track the last end position of the incremental scan. Then the last position will be used as the from position in the next (exclusive). Hence, I think the `TableScan#appendsAfter` is probably not a valid use case.

api/src/main/java/org/apache/iceberg/Scan.java
stevenzwu: done

api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
```diff
@@ -0,0 +1,55 @@
...
+ * under the License.
+ */
+
+
+package org.apache.iceberg;
+
+/**
+ * API for configuring an incremental table scan for appends only snapshots
+ */
+public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
```
Reo-LEI, line 26: I think this can be abstracted into `IncrementalScan`, because not only append-only, changelog incremental scan also needs to specify `from` and `to` snapshots. We can define the `IncrementalScan` interface, and then return different concrete implementations in `Table` through different methods, for example:

````
public interface IncrementalScan extends Scan<IncrementalScan> {...}

abstract class BaseIncrementalScan implements IncrementalScan {...}
public class AppendOnlyIncrementalScan extends BaseIncrementalScan {...}
public class ChangelogIncrementalScan extends BaseIncrementalScan {...}

public interface Table {
   ...
  IncrementalScan newAppendIncrementalScan();
  IncrementalScan newChangelogIncrementalScan();
  ...
}
````

Comment (stevenzwu) [reply api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java:26]: @Reo-LEI conceptually I agree with you. @rdblue prefer to do the refactoring when we come to the changelog incremental scan. 

The reason we didn't add `BaseIncrementalScan` is that right now there is no difference between `BaseIncrementalScan` and `IncrementalAppendScan`. In the future, we can extract `BaseIncrementalScan` out or we can have `IncrementalChangelogScan` extending `IncrementalAppendScan`. Personally, I also prefer the `BaseIncrementalScan`.

File: api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java (hunk 16 above)
Comment (rdblue) [inline api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java:49]: I think this needs a short Javadoc description.

File: api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java [hunk 18]
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Refine the incremental scan with the start snapshot inclusive.
+   * <p>
+   * If the start snapshot (inclusive or exclusive) is not provided,
+   * the oldest ancestor of the {@link IncrementalAppendScan#toSnapshot(long)}
+   * will be included as the start snapshot.
```
rdblue, line 33: @stevenzwu, this isn't true. If the starting snapshot is not set, then it defaults to `null`, which will scan from the start of table history or fail if table history has expired.

stevenzwu: @rdblue yes, it defaults to `null`. The described behavior is based on `IncrementalDataTableScan` using `SnapshotUtil.ancestorsBetween`, which scan from the oldest ancestor of the `toSnapshotId`. I think the current behavior makes sense. Start of table history may not be an ancestor of the `toSnapshotId`, right?

stevenzwu: Maybe the description can be simplified as `if ..., this scans all the ancestors of the {@link IncrementalAppendScan#toSnapshot(long)}`

rdblue: I don't think the proposed behavior makes sense for this interface. Otherwise, there is no way to incrementally scan from the start of the table. If I want to start from the beginning of history, I need to specify starting snapshot `null`. But there's no way to do that without leaving out the "from" snapshot. If that's how to configure scanning from the start of history, then this can't scan from the oldest known snapshot by default.

stevenzwu: if the oldest table snapshot is not an ancestor of the `toSnapshot`, what does it mean? I thought incremental scan is only meaningful along the linear ancestor line.

E.g., we have two disjointed lineages. 
S1 -> S2     ---------->     S5
               S3 -> S4 -> S6 -> S7

If the `toSnapshotId` is set to `S7` and `fromSnapshotId` is not set, I thought we want to scan [S3, S4, S6, S7]. Is that correct?

rdblue: The problem is when the history has expired, not when there is no ancestor relationship. When the starting point is not an ancestor, that's a different problem that results in an exception.

stevenzwu: Just to make sure I understand you correctly. if `fromSnapshotId` is not set and defaults to null, we want start from the snapshot with the oldest timestamp using this `Table` API: `Iterable<Snapshot> snapshots()`. if the oldest snapshot is not an ancestor of the current table snapshot, we throw an exception.

I assume we don't want to use this `Table` API to find out the oldest snapshot (by timestamp)
```
List<HistoryEntry> history()
```

rdblue: If no `fromSnapshot` method is called, the incremental read should start from the beginning of table history, the `null` snapshot. So we need to find a snapshot with `parent-snapshot-id=null`.

stevenzwu: if the table history has disjointed lineage lines, then we can have multiple snapshots with `parent-snapshot-id=null`. I guess then we can use timestamp to break the tie.

stevenzwu: @rdblue I updated the Javadoc based on the discussion here.

stevenzwu: @rdblue can you take another look and see if the comments are addressed adequately.

api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * API for configuring an incremental table scan for appends only snapshots
+ */
+public interface IncrementalAppendScan extends Scan<IncrementalAppendScan> {
+
+  /**
+   * Refine the incremental scan with the start snapshot inclusive.
+   * <p>
+   * If the start snapshot (inclusive or exclusive) is not provided,
+   * the oldest snapshot will be used as the start snapshot.
```
rdblue, line 32: I don't think this statement is clear enough. The table's first snapshot is used. You could argue that's the "oldest" but I think it is better to be clear that you're asking to process snapshots back to the start of the table. Clarifying on the next line helps, but it exposes an internal detail about how we track history: a snapshot with no parent is the starting snapshot.

stevenzwu: Can you see if the latest change is more clear?

rdblue (approved):

PR #4584, opened 2022-04-19
System Printing: Removing System.out and System.err

felixYyu:
in checkstyle.xml, `Logging with System.out and System.err are not allowed because it has no metadata and can't be configured at runtime. Please use an SLF4J logger instead`. so use SLF4J logger instead System.out and System.err. 

- Current strategy:  Removing System.out and System.err

cc: @RussellSpitzer @rdblue

### DIFF
--- a/api/src/test/java/org/apache/iceberg/types/TestReadabilityChecks.java
+++ b/api/src/test/java/org/apache/iceberg/types/TestReadabilityChecks.java
@@ -24,14 +24,11 @@ import org.apache.iceberg.Schema;
 import org.apache.iceberg.types.Type.PrimitiveType;
 import org.junit.Assert;
 import org.junit.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import static org.apache.iceberg.types.Types.NestedField.optional;
 import static org.apache.iceberg.types.Types.NestedField.required;
 
 public class TestReadabilityChecks {
-  private static final Logger LOG = LoggerFactory.getLogger(TestReadabilityChecks.class);
   private static final Type.PrimitiveType[] PRIMITIVES = new Type.PrimitiveType[] {
       Types.BooleanType.get(),
       Types.IntegerType.get(),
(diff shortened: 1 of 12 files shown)


RussellSpitzer (changes requested):
I think it basically all of these examples the output should just be removed. The test output isn't used in any way and is basically useless since no one checks integration test logs and if it was wrong we would be outputting those values immediately afterwards in the assert.

api/src/test/java/org/apache/iceberg/types/TestReadabilityChecks.java
```diff
@@ -362,7 +365,7 @@ public void testStructWriteReordering() {
     List<String> errors = CheckCompatibility.writeCompatibilityErrors(read, write);
     Assert.assertEquals("Should produce 1 error message", 1, errors.size());
 
-    System.err.println(errors);
```
rdblue, line 365: Printing in tests should be removed rather than logged.

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestParquetAvroReader.java
```diff
@@ -116,8 +120,8 @@ public void testStructSchema() throws IOException {
         long end = System.currentTimeMillis();
         long duration = end - start;
 
-        System.err.println("XOR val: " + val);
-        System.err.println(String.format("Reassembled %d records in %d ms", count, duration));
+        LOG.error("XOR val: {}", val);
+        LOG.error("Reassembled {} records in {} ms", count, duration);
```
rdblue, line 124: This is ignored, so there is no need to update this. It is for sanity checking the implementation.

api/src/test/java/org/apache/iceberg/types/TestReadabilityChecks.java
felixYyu: removed

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestParquetAvroReader.java
felixYyu: Although this is ignored, but I also think this should be removed.

felixYyu:
cc @RussellSpitzer can you take a look again, thanks.

RussellSpitzer (approved):

Issue #4582, opened 2022-04-18
Remove extraneous trailing slash in table location


puchengy:
Creating an Iceberg table from spark-3.2 with iceberg runtime 0.13.1  with location `s3://bucket/db/tbl/` will cause underlying files be like `s3://bucket/db/tbl//data/xxx`. This leads to other system (e.g. Trino) file not found exception as those system will remove the duplicated slash in the string (`s3://bucket/db/tbl/data/xxx`).

I think the right behavior should be letting iceberg remove extraneous trailing slash in table location. i.e. when setting the table location as `s3://bucket/db/tbl/`. the trailing `/` should be removed.

nicor88:
@rdblue When running on Iceberg `0.14.0` the issue seems fixed for the data folder, but somehow if I have a trailing slash in my location, the trailing slash is kept for the metadata folder.

PR #4585, opened 2022-04-19
Core : Strip trailing slash from tableLocation in LocationProvider

singhpk234:
Closes https://github.com/apache/iceberg/issues/4582

This change attempts to fix :
1. In `locationsFor` we don't sanitize the location (i.e strip the trailing slash), this attempts to do this stripping
2. The warehouse location for hiveCatalog is also missing this handling rest all catalogs such as Hadoop / Glue have these handling 
3. Had to refactor a bit, slash stripping logic was duplicated in all catalogs. 
   Note : nessie Catalog also  doesn't have this handling, will do the same in next iteration of this pr if experts from nessie   suggests this is required (appologies I am new to nessie)
4. Fixes a minor typo.

cc @szehon-ho , @RussellSpitzer , @jackye1995 , @flyrain

### DIFF
--- a/core/src/main/java/org/apache/iceberg/CatalogUtil.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogUtil.java
@@ -341,15 +341,4 @@ public class CatalogUtil {
 
     setConf.invoke(conf);
   }
-
-  public static String cleanWarehousePath(String path) {
-    Preconditions.checkArgument(path != null && path.length() > 0,
-        "Cannot initialize Catalog because warehousePath must not be null or empty");
-    int len = path.length();
-    if (path.charAt(len - 1) == '/') {
-      return path.substring(0, len - 1);
-    } else {
-      return path;
-    }
-  }
 }
--- a/core/src/main/java/org/apache/iceberg/CatalogUtil.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogUtil.java
@@ -341,4 +341,15 @@ public class CatalogUtil {
 
     setConf.invoke(conf);
   }
+
+  public static String cleanWarehousePath(String path) {
+    Preconditions.checkArgument(path != null && path.length() > 0,
+        "Cannot initialize Catalog because warehousePath must not be null or empty");
+    int len = path.length();
+    if (path.charAt(len - 1) == '/') {
+      return path.substring(0, len - 1);
+    } else {
+      return path;
+    }
+  }
 }
--- a/core/src/main/java/org/apache/iceberg/LocationProviders.java
+++ b/core/src/main/java/org/apache/iceberg/LocationProviders.java
@@ -35,6 +35,7 @@ public class LocationProviders {
   }
 
   public static LocationProvider locationsFor(String location, Map<String, String> properties) {
+    String sanitizedLocation = stripTrailingSlash(location);
     if (properties.containsKey(TableProperties.WRITE_LOCATION_PROVIDER_IMPL)) {
       String impl = properties.get(TableProperties.WRITE_LOCATION_PROVIDER_IMPL);
       DynConstructors.Ctor<LocationProvider> ctor;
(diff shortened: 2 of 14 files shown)


core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -341,4 +341,15 @@ public static void configureHadoopConf(Object maybeConfigurable, Object conf) {
 
     setConf.invoke(conf);
   }
+
+  public static String cleanWarehousePath(String path) {
```
rdblue, line 345: `LocationProviders` already exposes `stripTrailingSlash`, which removes multiple trailing slashes. Could you use that instead?

core/src/main/java/org/apache/iceberg/LocationProviders.java
```diff
@@ -51,7 +52,7 @@ public static LocationProvider locationsFor(String location, Map<String, String>
             impl, LocationProvider.class), e);
       }
       try {
-        return ctor.newInstance(location, properties);
+        return ctor.newInstance(sanitizedLocation, properties);
```
rdblue, line 55: Nit: I generally prefer to rename the method argument so that there are fewer changes.

rdblue:
Thanks, @singhpk234! Looks mostly good, but I think we should use just one `stripTrailingSlash` method.

core/src/main/java/org/apache/iceberg/hadoop/HadoopCatalog.java
```diff
@@ -98,11 +98,8 @@ public HadoopCatalog() {
 
   @Override
   public void initialize(String name, Map<String, String> properties) {
-    String inputWarehouseLocation = properties.get(CatalogProperties.WAREHOUSE_LOCATION);
-    Preconditions.checkArgument(inputWarehouseLocation != null && !inputWarehouseLocation.equals(""),
-        "Cannot instantiate hadoop catalog. No location provided for warehouse (Set warehouse config)");
     this.catalogName = name;
-    this.warehouseLocation = inputWarehouseLocation.replaceAll("/*$", "");
+    this.warehouseLocation = CatalogUtil.cleanWarehousePath(properties.get(CatalogProperties.WAREHOUSE_LOCATION));
```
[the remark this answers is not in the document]
singhpk234: sure, done :) !!

core/src/main/java/org/apache/iceberg/CatalogUtil.java
singhpk234: sure done :) !!

core/src/main/java/org/apache/iceberg/hadoop/HadoopCatalog.java
```diff
@@ -98,11 +98,8 @@ public HadoopCatalog() {
 
   @Override
   public void initialize(String name, Map<String, String> properties) {
-    String inputWarehouseLocation = properties.get(CatalogProperties.WAREHOUSE_LOCATION);
-    Preconditions.checkArgument(inputWarehouseLocation != null && !inputWarehouseLocation.equals(""),
-        "Cannot instantiate hadoop catalog. No location provided for warehouse (Set warehouse config)");
```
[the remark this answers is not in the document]
singhpk234: These assertions were being done now as part of `cleanWarehousePath` function.

Have placed them back

core/src/main/java/org/apache/iceberg/LocationProviders.java
```diff
@@ -160,7 +161,7 @@ private static String pathContext(String tableLocation) {
     }
   }
 
-  private static String stripTrailingSlash(String path) {
+  public static String stripTrailingSlash(String path) {
```
rdblue, line 164: Is there a better place for this rather than making it public from `LocationProviders`? We could move it to `CatalogUtil` if you think that's the right place. Or maybe another util class?

singhpk234: +1, was thinking, how about if we create a new util class let's say `LocationUtils` under iceberg.utils package, more such helper functions if required could be placed here.

rdblue (approved):

rdblue:
Thanks, @singhpk234!

PR #4586, opened 2022-04-19
Spark: Add Spark procedures for snapshot branching and tagging

rajarshisarkar:
This PR adds the Spark procedures for snapshot branching and tagging.

I have added `CreateTagProcedure` for the initial round of review. I will add the other procedures if this looks good.

Reference: [Snapshot Management Spark Procedures](https://iceberg.apache.org/docs/latest/spark-procedures/#snapshot-management)

---

cc: @rdblue @jackye1995 @amogh-jahagirdar

### DIFF
--- a/api/src/main/java/org/apache/iceberg/SnapshotRefType.java
+++ b/api/src/main/java/org/apache/iceberg/SnapshotRefType.java
@@ -19,7 +19,7 @@
 
 package org.apache.iceberg;
 
-enum SnapshotRefType {
+public enum SnapshotRefType {
     BRANCH,
     TAG
 }
--- a/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCreateTagProcedure.java
+++ b/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestCreateTagProcedure.java
@@ -1,0 +1,185 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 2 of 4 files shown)


amogh-jahagirdar:
For operations like creating and dropping branches and tags I was thinking it may make more sense to have a DDL syntax instead of a call procedure (which I think we can leverage for other operations):

For example:

```
CREATE TAG [IF NOT EXISTS] AT <SnapshotId> RETAIN _numHours_ [SECONDS | MINUTES | HOURS | DAYS | etc]
```

rajarshisarkar:
Thanks, for the inputs @amogh-jahagirdar! 

I am carrying forward the call procedure approach from the [existing snapshot management APIs](https://iceberg.apache.org/docs/latest/spark-procedures/#snapshot-management): `rollback_to_snapshot`, `rollback_to_timestamp`, `set_current_snapshot`, `cherrypick_snapshot`. However, the SQL approach looks really interesting to me.

Here are my thoughts:
(1) Introduce call procedures for all the branching & tagging APIs as effectively they are snapshot management APIs and we already have [some existing procedural calls for the same](https://iceberg.apache.org/docs/latest/spark-procedures/#snapshot-management). The procedural calls could be API specific or generic enough to cater different use cases via named/positional arguments.
(2) Introduce new SQL Syntax specific to branching and tagging - make changes in `IcebergSparkSessionExtensions` (and also in parser, analyzer, planner to understand the new syntax). The Syntax could be something similar to what Amogh suggested.

I don't have a strong opinion on (1) or (2) - would like to just use Spark to create/modify/remove a branch/tag. I would also like to list/view details of the available branches/tags using Spark.

@rdblue @jackye1995 Please let me know your thoughts.

jackye1995:
+1 for what Amogh suggested, I would vote for directly going with Spark SQL extension if possible. Just from user standpoint, SQL is considered an easier user experience compared to procedure calls, although I understand technically there is not much difference for us developers. I would also hope the current procedures such as `rewrite_data_files`, `migrate` and `expire_snapshots` to map to `OPTIMIZE`, `IMPORT` and `VACUUM` SQL extensions in the future.

@amogh-jahagirdar I remember you have a doc about the possible syntaxes, could you share with the audience here? Once the syntax is determined, the implementation is pretty straightforward.

amogh-jahagirdar:
Sure! I will update here with a doc on the DDL syntaxes I had in mind for branching/tagging.

rajarshisarkar:
> Just from user standpoint, SQL is considered an easier user experience compared to procedure calls

+1 for this aspect.

rajarshisarkar:
@amogh-jahagirdar Can you please share the doc once you get time. That would help us start the discussion on Syntax.

nastra:
+1 for having actual SQL to perform these actions

amogh-jahagirdar:
Published last week on the dev-list but forgot to link the doc here: https://docs.google.com/document/d/1tbATFPrKF3vNlzkgZQdaW8CAJmbjvryfrlg6C2Ci_aA/edit#

PR #4590, opened 2022-04-19
Python: Add PartitionField

dramaticlly:
This is 1st step of reintroduce https://github.com/apache/iceberg/issues/3228

adds `PartitionField` in `iceberg/partitioning.py`

```
from iceberg.partitioning import PartitionField
partition_field = PartitionField(3, 1, None, "id")
partition_field
PartitionField(field_id=1, name=id, transform=None, source_id=3)
```

Purposefully keep the change list small to focus only on PartitionField and once checked in we can look into PartitionSpec next.

CC @samredai @rdblue @jun-he


python/src/iceberg/partitioning.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __eq__(self, other):
+        return self.field_id == other.field_id
+
+    def __str__(self):
+        return f"{self.field_id}: {self.name}: {self.transform}: ({self.source_id})"
+
+    def __repr__(self):
+        return (
+            f"PartitionField(field_id={self.field_id}, name={self.name}, transform={self._transform}, source_id={self.source_id})"
```
samredai, line 56: nit: to be consistent with the other variables in the repr, `self._transform` should be `self.transform`

python/src/iceberg/partitioning.py
```diff
@@ -0,0 +1,57 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from iceberg.transforms import Transform
+
+
+class PartitionField:
```
samredai, line 20: @rdblue do we need to define a hash method for this class? Maybe something like:
```py
def __hash__(self):
    return hash((self.source_id, self.field_id, self.name, self.transform))
```

samredai (changes requested):
Overall this looks good, I suggested a few changes. Also I'm wondering if this should go into the table sub-module, I was thinking `table/partition.py`.

python/src/iceberg/partitioning.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    PartitionField is a single element with name and unique id,
+    It represents how one partition value is derived from the source column via transformation
+    """
```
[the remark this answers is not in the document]
dramaticlly: added `Attributes` in docstring as suggested!

python/src/iceberg/partitioning.py
dramaticlly: Looks like it's mainly used in list or object array from what I can tell, but please let me know if I should add it!

szehon-ho: Not sure the policy for adding hash across python classes, but maybe it will be useful as PartitionField is part of Iceberg public api ,(example if a user tries to make dicts keyed by partitionSpec?)    Java version has it, for comparison.

dramaticlly:
@samredai added hash and moved `partitioning` to the `table` module, if you could take another look :)

python/src/iceberg/table/partitioning.py
```diff
@@ -0,0 +1,66 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def transform(self) -> Transform:
+        return self._transform
+
+    def __eq__(self, other):
+        return self.field_id == other.field_id
```
rdblue, line 55: I don't think this is correct. It should be source_id, field_id, transform, and name.

python/src/iceberg/table/partitioning.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __str__(self):
+        return f"{self.field_id}: {self.name}: {self.transform}: ({self.source_id})"
```
rdblue, line 58: I think this should match the JVM one, which doesn't have `: ` after transform.

python/src/iceberg/table/partitioning.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __repr__(self):
+        return (
+            f"PartitionField(field_id={self.field_id}, name={self.name}, transform={self.transform}, source_id={self.source_id})"
```
rdblue, line 62: I think you need to use `repr(self.transform)` to build the repr.

python/src/iceberg/table/partitioning.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+    def __hash__(self):
+        return hash((self.source_id, self.field_id, self.name, self.transform))
```
rdblue, line 66: These aren't going to be used in a dict, so you shouldn't need to override `__hash__`.

python/src/iceberg/table/partitioning.py
dramaticlly: Thank you, updated the equality as suggested

python/src/iceberg/table/partitioning.py
dramaticlly: removed `: ` after transform to match Java implementation

python/src/iceberg/table/partitioning.py
dramaticlly: updated to use `repr` for transform

python/src/iceberg/table/partitioning.py
dramaticlly: removed `__hash__` override

dramaticlly:
@rdblue  and @samredai updated to address review feedback, appreciate another look :)

rdblue:
Looks great. Thanks, @dramaticlly!

PR #4592, opened 2022-04-20
Nessie: Replace CommitStateUnknownException with CommitFailedException

nastra:


nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
```diff
@@ -274,12 +273,9 @@ public void renameTable(TableIdentifier from, TableIdentifier to) {
       // to catch all kinds of network errors (e.g. connection reset). Network code implementation
       // details and all kinds of network devices can induce unexpected behavior. So better be
       // safe than sorry.
-      throw new CommitStateUnknownException(ex);
+      throw new CommitFailedException(ex, "Cannot rename table '%s' to '%s': " +
```
RussellSpitzer, line 276: I'm not sure this is right, I think this was originally added because some of the exceptions here do happen when the commit succeeded on the server but it failed to propagate the ack back to the client. At least that's what I was told when this got added way back when. If things have changed let me know.

nastra: seems you have more historical background than I do and in fact you are totally right, because that's exactly the case that's happening here. Closing the PR then

PR #4593, opened 2022-04-20
Nessie: Use memoized Supplier to avoid unnecessary API calls

nastra:
Note that this also reverts the test changes that were done in https://github.com/apache/iceberg/commit/9618147b6de8f8627052a205b86e45263394b0c2#diff-a92127e5d74e2be0c013d0eddf4c9d2af29a71f05e21d7289808271d4c97db78L326-R348

### DIFF
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
@@ -38,6 +38,7 @@ import org.apache.iceberg.TableMetadataParser;
 import org.apache.iceberg.TableOperations;
 import org.apache.iceberg.avro.AvroSchemaUtil;
 import org.apache.iceberg.catalog.TableIdentifier;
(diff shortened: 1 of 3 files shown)


nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -321,31 +322,22 @@ public void testExistingTableUpdate() {
   }
 
   @Test
-  public void testCommitsOutsideOfCatalogApi() throws NessieNotFoundException, NessieConflictException {
```
RussellSpitzer, line 324: Is this set of changes required? I think we just changed this in the last PR

nastra: yes those are required as the "reference loading" behavior is how it was previously and therefore that test that I rewrote is now valid again. This was unfortunately an oversight on my side as we should have been using a memoized supplier from the very beginning

RussellSpitzer (approved):

PR #4594, opened 2022-04-20
Bump Nessie to 0.28.0, improve commit conflict detection

snazy:
Enhances Iceberg commit conflict detection by maintaining the commit-id from which a
table metadata has been loaded, to use it as the "expected hash" in a Nessie commit.

Makes Nessie specific properties available in `TableMetadata` properties:
* `nessie.content.id` - the Nessie `Content.getId()`
* `nessie.commit.id` - the commit ID used to retrieve the table metadata
* `nessie.reference.name` - the reference name from which the table metadata has been loaded

Also fixes an issue via `org.apache.iceberg.nessie.NessieTableOperations#loadTableMetadata`
that caused too many "previous files", because the `TableMetadata.buildFrom()` assumed that
it's only called for ongoing modifications.


core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -819,7 +827,7 @@ private Builder(TableMetadata base) {
       this.startingChangeCount = changes.size();
 
       this.snapshotLog = Lists.newArrayList(base.snapshotLog);
-      this.previousFileLocation = base.metadataFileLocation;
+      this.previousFileLocation = noPreviousFile ? null : base.metadataFileLocation;
```
nastra, line 830: a little bit annoying that this needs to be set, so I wonder whether it would be better to have a builder method that allows overriding this

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -156,14 +189,15 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       Branch branch = client.getApi().commitMultipleOperations()
           .operation(Operation.Put.of(key, newTable, table))
           .commitMeta(NessieUtil.catalogOptions(builder, catalogOptions).build())
-          .branch(client.getRef().getAsBranch())
+          .branch(expectedHead)
```
nastra, line 192: I believe you can now remove the ignore on https://github.com/apache/iceberg/blob/master/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieCatalog.java#L156-L161 as the test would pass now

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -173,7 +207,7 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       throw new CommitStateUnknownException(ex);
     } catch (NessieNotFoundException ex) {
       throw new RuntimeException(
-          String.format("Cannot commit: Reference '%s' no longer exists", client.getRef().getName()), ex);
+          String.format("Cannot commit: Reference '%s' no longer exists", updateableReference.getName()), ex);
```
nastra, line 210: nit: I would probably revert all of those changes as I don't think they are necessary

snazy: The result of `client.getRef()` _could_ change in the meantime.

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -103,13 +123,14 @@ protected void doRefresh() {
           "is no longer valid.", client.getRef().getName()), e);
     }
     String metadataLocation = null;
+    Reference reference = client.getRef().getReference();
```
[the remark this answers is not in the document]
snazy: The result of `client.getRef()` _could_ change in the meantime.

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -75,21 +87,27 @@ protected String tableName() {
...
+      deserialized = TableMetadataParser.read(io(), metadataLocation);
+    }
+    Map<String, String> newProperties = Maps.newHashMap(deserialized.properties());
+    newProperties.put(NESSIE_COMMIT_ID_PROPERTY, reference.getHash());
+    TableMetadata.Builder builder = TableMetadata.buildFrom(deserialized, true)
         .setCurrentSchema(table.getSchemaId())
         .setDefaultSortOrder(table.getSortOrderId())
         .setDefaultPartitionSpec(table.getSpecId())
-        .withMetadataLocation(metadataLocation);
+        .withMetadataLocation(metadataLocation)
```
kbendick, line 104: I think you don't need to pass `true` in here. If `base` doesn't have that field set, it will just be null. Otherwise it will get overridden after this builder is called with `.build()`.

core/src/main/java/org/apache/iceberg/TableMetadata.java
kbendick: I need to override that too for something I'm working on. But I have just used the builder's `withMetadataLocation`. In my case, the value of `base.metadataFIleLocation` is already null so just calling the function works fine.

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
snazy: The problem here is that the `buildFrom`/`Builder.<init>()` sets the final field `previousFileLocation`, which is then used in `Builder.build()` to add a `MetadataLogEntry`. `Builder.previousFileLocation` and `Builder.metadataLocation` are different.

core/src/main/java/org/apache/iceberg/TableMetadataParser.java
```diff
@@ -155,7 +155,7 @@ public static String toJson(TableMetadata metadata) {
     }
   }
 
-  private static void toJson(TableMetadata metadata, JsonGenerator generator) throws IOException {
```
RussellSpitzer, line 158: I don't think we want to change accessibility on these methods, why is the custom JsonGenerator needed in the Nessie code?

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -75,21 +87,27 @@ protected String tableName() {
...
-    super.refreshFromMetadataLocation(newLocation, shouldRetry, numRetries, this::loadTableMetadata);
-  }
-
-  private TableMetadata loadTableMetadata(String metadataLocation) {
+  private TableMetadata loadTableMetadata(String metadataLocation, Reference reference) {
     // Update the TableMetadata with the Content of NessieTableState.
-    TableMetadata.Builder builder = TableMetadata.buildFrom(TableMetadataParser.read(io(), metadataLocation))
+    TableMetadata deserialized;
+    if (table.getMetadata() != null) {
+      deserialized = TableMetadataParser.fromJson(io(), metadataLocation, table.getMetadata().getMetadata());
```
RussellSpitzer, line 94: So we have this path in to request metadata out of a Nessie specific object which stores the metadata as a Json String that needs to be parsed?

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -75,21 +87,27 @@ protected String tableName() {
...
+    } else {
+      deserialized = TableMetadataParser.read(io(), metadataLocation);
```
RussellSpitzer, line 96: Otherwise we just read from the location on disk?

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -755,6 +755,10 @@ public static Builder buildFrom(TableMetadata base) {
     return new Builder(base);
   }
 
+  public static Builder buildFrom(TableMetadata base, boolean noPreviousFile) {
```
RussellSpitzer, line 758: A new public api here needs a java doc, I'm not sure I follow what noPreviousFile is expected to do here

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -139,15 +170,27 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       }
       Snapshot snapshot = metadata.currentSnapshot();
       long snapshotId = snapshot != null ? snapshot.snapshotId() : -1L;
+
+      JsonNode newMetadata;
+      try {
+        TokenBuffer tokenBuffer = new TokenBuffer(JsonUtil.factory().getCodec(), false);
+        TableMetadataParser.toJson(metadata, tokenBuffer);
```
RussellSpitzer, line 177: So we could avoid this usage if we use the "toJson(metadata) method and then pass that into the token buffer. I know it's a bit more wasteful but we are only doing this once per operation and it let's us keep the toJson(metadata, generator) method package private. Again I may be being paranoid here but ideally we don't open up any more methods than we absolutely have to.

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
snazy: (Not yet, but in the future, yes.)

It's already a JSON node tree (as a generic `JsonNode`) - so the text->nodes parsing already happened.
As I didn't want to serialize the `JsonNode` to a `String` to get a `TableMetadata`, I've updated the `TableMetadataParser`.

core/src/main/java/org/apache/iceberg/TableMetadata.java
snazy: That's a "smart thing" in the builder. If it's present, then it add it as a history item. That behavior is okay, when `buildFrom()` is only used when it is used within a "modify `TableMetadata` operation". But the Nessie catalog only uses it to modify the `TableMetadata` as it is - so it's not a change, which requires a new history-item.
Without the `noPreviousFile==true`, a couple of catalog tests fail, because those assert on the number of history items.

RussellSpitzer: @snazy I've been thinking on this and I'm wondering why can't implement a builder like  `.withPreviousFileLocation` to set that to previousFileLocation to null

snazy: WFM, updated the PR

RussellSpitzer (approved):

PR #2905, opened 2021-08-01
Flink: refactor DataIterator to use composition (instead of inheritance)

stevenzwu:
With this composition mode, `DataIterator` deals with `CombinedScanTask` and `IteratorReader` deals with individual `FileScanTask`. `DataIterator` use composition to reference `IteratorReader`

Here are the motivations behind this change
1) address @openinx 's [review comment](https://github.com/apache/iceberg/pull/2105#discussion_r568303231)
2) make it easier to extend. E.g., internally we would like to add another converter from RowData to some other type Foo. We can't define some `FooDataIterator extends RowDataIterator` as the generic type is already fixed to `RowData`. With this new composition model, we can define a `FooIteratorReader implements IteratorReader` and `FooIteratorReader` can use `RowDataIteratorReader` by composition.

PR #4595, opened 2022-04-20
[Spark][MR]: Reuse InputFilesDecryptor

zhongyujiang:
Reuse InputFilesDecryptor in Spark BaseDataReader and MR IcebergInputFormat to reduce code duplication.

### DIFF
--- a/mr/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/mr/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -57,8 +57,7 @@ import org.apache.iceberg.data.InternalRecordWrapper;
 import org.apache.iceberg.data.avro.DataReader;
 import org.apache.iceberg.data.orc.GenericOrcReader;
 import org.apache.iceberg.data.parquet.GenericParquetReaders;
-import org.apache.iceberg.encryption.EncryptedFiles;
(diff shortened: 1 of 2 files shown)


flyrain:
LGTM. A nice refactor, which follows the same approach in #2905. Do we have any test from Spark/Hive side? Also cc @stevenzwu since he is the author of #2905.

ggershinsky:
Nice optimization, LGTM

zhongyujiang:
Thanks for your review and sorry for my late reply.
> Do we have any test from Spark/Hive side?

This just did some small code optimizations so I didn't add the test. What kind of tests do you think should be added? @flyrain

flyrain:
> This just did some small code optimizations so I didn't add the test. What kind of tests do you think should be added? @flyrain 

It's OK to not add tests. I'm just curious that which test covers the changed code.

zhongyujiang:
> I'm just curious that which test covers the changed code.

I think we don't have relevant tests cause encryption hasn't been implemented yet, only PlaintextEncryptionManager at the moment.

flyrain:
> I think we don't have relevant tests cause encryption hasn't been implemented yet, only PlaintextEncryptionManager at the moment.

It is a bit scary, but probably is fine since we've got https://github.com/apache/iceberg/pull/2905 in.

flyrain (approved):

PR #4599, opened 2022-04-20
Orc : Fix inner struct field as partition (#4604)

tprelle:
Fix : https://github.com/apache/iceberg/issues/4604
When you you add a inner struct field as partition field without any transformation on an Iceberg orc table, you can not select anymore only this field because the as is a partition, iceberg do not read it from the file as it's inside a struct OrcValueReaders it's not able to reconstruct a constant struct.

It's working when you select also another field from the struct because as OrcValueReaders is able to read from the file using the another field reader

So instead of trying to add it as a constant, read directly from the files
the value of a hidden partition based on inner struct value even if you only read this column

### DIFF
--- a/api/src/main/java/org/apache/iceberg/types/TypeUtil.java
+++ b/api/src/main/java/org/apache/iceberg/types/TypeUtil.java
@@ -38,6 +38,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 
 public class TypeUtil {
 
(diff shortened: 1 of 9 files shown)


rdblue:
@shardulm94 can you take a look at this?

kbendick (commented):
Hi @tprelle! Thanks for this patch.

Normally we only change one version per project at a time and then backport. Becase of changes in `api`, that's not always possible but it does make reviewing easier. Would that be possible?

api/src/main/java/org/apache/iceberg/types/TypeUtil.java
```diff
@@ -130,15 +131,45 @@ private static Set<Integer> getIdsInternal(Type type, boolean includeStructIds)
...
+      struct.fields().forEach(f -> innerFilter(filteredIds, f));
+    }
+    projectedIds.removeAll(filteredIds);
+  }
+
   public static Types.StructType selectNot(Types.StructType struct, Set<Integer> fieldIds) {
+    return selectNot(struct, fieldIds, false);
+  }
+
+  public static Types.StructType selectNot(Types.StructType struct, Set<Integer> fieldIds, boolean filteredStruct) {
```
kbendick, line 160: Nit: This boolean name is a bit strange to me. Is there perhaps a more descriptive name for what this corresponds to? Maybe `doesHaveInnerStruct` or `doesHaveInnerProjectedStruct`?

I was going to ask you to inline the field name with the raw boolean values in the tests, like `true /* filteredStruct /*`, but that still left me with some confusion.

tprelle: Hi @kbendick,
Thanks for the review.
I had trouble to name this variable. It's may more doesNeedToFilteredInnerStruct. 
I add this variable because on some cases we need to filtered inner struct (ORC) in some cases we do not need.

Is doesNeedToFilteredInnerStruct is better ?

tprelle:
> Hi @tprelle! Thanks for this patch.
> 
> Normally we only change one version per project at a time and then backport. Becase of changes in `api`, that's not always possible but it does make reviewing easier. Would that be possible?

So @kbendick, as i keep the old signature methode on the api it's possible.
To check : On the first PR i fix the major version of flink and spark. Then i will publish a backport PR for others versions ?

api/src/main/java/org/apache/iceberg/types/TypeUtil.java
tprelle: i just renamed it to doesNeedToKeepInnerStruct
Because it's more if it's a struct and if all ids need to be filtered we need keep all. So we remove all ids from filtered ids set

PR #4601, opened 2022-04-21
[CORE] - Add MetadataUpdateParser and Jackson SerDe with implementation covering UpgradeFormatVersion

kbendick:
We need to be able to serialize and deserialize instances of `MetadataUpdate`.

Because `MetadataUpdate` is an interface that has multiple implementations, I've added a serializer / deserializer that handles `MetadataUpdate` and then reads or write the correct value based on the `action` field, as defined in the open-api spec.

I've only implemented  and tested`MetadataUpdate#UpgradeFormatVersion`, and have laid the foundation for the remaining classes which I will handle in a follow up PR.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
@@ -107,7 +107,7 @@ public class MetadataUpdateParser {
       case ASSIGN_UUID:
         throw new UnsupportedOperationException("Not Implemented: MetadataUpdate#toJson for AssignUUID");
       case UPGRADE_FORMAT_VERSION:
-        writeUpgradeFormatVersion(metadataUpdate, generator);
+        writeUpgradeFormatVersion((MetadataUpdate.UpgradeFormatVersion) metadataUpdate, generator);
         break;
       default:
         throw new IllegalArgumentException(
@@ -149,9 +149,9 @@ public class MetadataUpdateParser {
   }
 
   // Write all fields specific to UpgradeFormatVersion
-  private static void writeUpgradeFormatVersion(MetadataUpdate metadataUpdate, JsonGenerator gen) throws IOException {
-    MetadataUpdate.UpgradeFormatVersion upgradeFormatVersion = (MetadataUpdate.UpgradeFormatVersion) metadataUpdate;
-    gen.writeNumberField(FORMAT_VERSION, upgradeFormatVersion.formatVersion());
+  private static void writeUpgradeFormatVersion(MetadataUpdate.UpgradeFormatVersion update, JsonGenerator gen)
+      throws IOException {
+    gen.writeNumberField(FORMAT_VERSION, update.formatVersion());
   }
 
   private static MetadataUpdate readAsUpgradeFormatVersion(JsonNode node) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -0,0 +1,163 @@
...
+    Preconditions.checkArgument(jsonNode != null && jsonNode.isObject(),
+        "Cannot parse metadata update from non-object value: %s", jsonNode);
+    Preconditions.checkArgument(jsonNode.hasNonNull(ACTION), "Cannot parse metadata update. Missing field: action");
+    String action = JsonUtil.getString(ACTION, jsonNode).toLowerCase(Locale.ROOT);
+
+    switch (action) {
+      case ASSIGN_UUID:
+        throw new UnsupportedOperationException("Not implemented: AssignUUID");
+      case UPGRADE_FORMAT_VERSION:
+        return readAsUpgradeFormatVersion(jsonNode);
```
kbendick, line 145: I will add the implementations for the other classes in a follow up PR after verifying that this is acceptable for others.

danielcweeks (approved):
LGTM, but I'll give @rdblue the chance to comment.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -0,0 +1,160 @@
...
+      toJson(metadataUpdate, generator);
+      generator.flush();
+      return writer.toString();
+    } catch (IOException e) {
+      throw new UncheckedIOException(String.format("Failed to write metadata update json for: %s", metadataUpdate), e);
+    }
+  }
+
+  public static void toJson(MetadataUpdate metadataUpdate, JsonGenerator generator) throws IOException {
+    String updateAction = ACTIONS.get(metadataUpdate.getClass());
```
rdblue, line 97: I think this should check whether an action was found and throw an `UnsupportedOperationException` if it is not.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -0,0 +1,160 @@
...
+        throw new UnsupportedOperationException("Not implemented: AssignUUID");
+      case UPGRADE_FORMAT_VERSION:
+        return readAsUpgradeFormatVersion(jsonNode);
+      default:
+        throw new UnsupportedOperationException(
+            String.format("Cannot convert metadata update action to json: %s", action));
+    }
+  }
+
+  // Write all fields specific to UpgradeFormatVersion
```
rdblue, line 149: Nit: This comment seems to duplicate what we know from the method name.

rdblue (approved):

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: It does that more or less via the switch statement's default. But I'll add an explicit UnsupportedOperationException on `null`.

rdblue:
Thanks! Looks like a great start.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
rdblue: Yeah, probably better to do it earlier, before passing it to the generator.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -0,0 +1,160 @@
...
+
+    // Write subclass specific fields
+    switch (updateAction) {
+      case ASSIGN_UUID:
+        throw new UnsupportedOperationException("Not Implemented: MetadataUpdate#toJson for AssignUUID");
+      case UPGRADE_FORMAT_VERSION:
+        writeUpgradeFormatVersion((MetadataUpdate.UpgradeFormatVersion) metadataUpdate, generator);
+        break;
+      default:
+        throw new IllegalArgumentException(
```
rdblue, line 111: Should this also be `UnsupportedOperationException`?

PR #4605, opened 2022-04-21
Support insert overwrite table for hive-mr engine.

southernriver:
Support insert overwrite table for hive-mr engine by setting runtime conf with `set iceberg.mr.write.is.overwrite = true` , otherwise we need to modify hive-module to adapt syntax.

### DIFF
--- a/mr/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/mr/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -296,11 +296,14 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
         dataFiles.forEach(overwrite::addFile);
         overwrite.commit();
         LOG.info("Overwrite commit took {} ms for table: {} with {} file(s)", System.currentTimeMillis() - startTime,
-          table, dataFiles.size());
+            table, dataFiles.size());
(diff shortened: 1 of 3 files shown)


mr/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -285,16 +288,33 @@ private void commitTable(FileIO io, ExecutorService executor, JobContext jobCont
...
+        ReplacePartitions overwrite = table.newReplacePartitions();
+        dataFiles.forEach(overwrite::addFile);
+        overwrite.commit();
+        LOG.info("Overwrite commit took {} ms for table: {} with {} file(s)", System.currentTimeMillis() - startTime,
+          table, dataFiles.size());
+      } else if (table.spec().isUnpartitioned()) {
+        table.newDelete().deleteFromRowFilter(Expressions.alwaysTrue()).commit();
+        LOG.info("Cleared table contents as part of empty overwrite for unpartitioned table. " +
+          "Commit took {} ms for table: {}", System.currentTimeMillis() - startTime, table);
+      }
```
chenjunjiedada, line 304: What is the behavior of using empty to overwrite a partitioned table? Should we print some log?

southernriver: Not do anything, I added some log.

PR #4606, opened 2022-04-21
docs:Add an example of CTAS with PARTITIONED BY (fix #3854)

### LINKED ISSUE #3854: Docs: Add an example of CTAS with PARTITIONED BY
The docs should include an example of using PARTITIONED BY in a CTAS statement (such as in [this ](https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/sql/TestCreateTableAsSelect.java#L45-L46)test). It may not be clear to a user that they can do this and would be useful to highlight in the docs that describe CTAS.

kyle-cx:
Add an example of CTAS with PARTITIONED BY, make it clear that we can use PARTITIONED BY in CTAS. @samredai 

image

fix https://github.com/apache/iceberg/issues/3854

### DIFF
--- a/docs/spark/spark-ddl.md
+++ b/docs/spark/spark-ddl.md
@@ -106,6 +106,8 @@ TBLPROPERTIES ('key'='value')
 AS SELECT ...
 ```
 
+The newly created table won't inherit the partition spec and table properties from the source table in SELECT, you can use PARTITIONED BY and TBLPROPERTIES in CTAS to declare partition spec and table properties for the new table.
+
 ## `REPLACE TABLE ... AS SELECT`
 
 Iceberg supports RTAS as an atomic operation when using a [`SparkCatalog`](../spark-configuration#catalog-configuration). RTAS is supported, but is not atomic when using [`SparkSessionCatalog`](../spark-configuration#replacing-the-session-catalog).
--- a/docs/spark/spark-ddl.md
+++ b/docs/spark/spark-ddl.md
@@ -98,6 +98,9 @@ CREATE TABLE prod.db.sample
 USING iceberg
 AS SELECT ...
 ```
+
+The newly created table won't inherit the partition spec and table properties from the source table in SELECT, you can use PARTITIONED BY and TBLPROPERTIES in CTAS to declare partition spec and table properties for the new table.
+
 ```sql
 CREATE TABLE prod.db.sample
 USING iceberg
@@ -105,8 +108,6 @@ PARTITIONED BY (part)
 TBLPROPERTIES ('key'='value')
 AS SELECT ...
 ```
-
-The newly created table won't inherit the partition spec and table properties from the source table in SELECT, you can use PARTITIONED BY and TBLPROPERTIES in CTAS to declare partition spec and table properties for the new table.
 
 ## `REPLACE TABLE ... AS SELECT`
 
--- a/docs/spark/spark-ddl.md
+++ b/docs/spark/spark-ddl.md
@@ -98,6 +98,13 @@ CREATE TABLE prod.db.sample
 USING iceberg
 AS SELECT ...
 ```
+```sql
+CREATE TABLE prod.db.sample
+USING iceberg
+PARTITIONED BY (part)
+TBLPROPERTIES ('key'='value')
+AS SELECT ...
+```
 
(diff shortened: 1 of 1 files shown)


samredai:
LGTM, thanks @kyle-cx !!

docs/spark/spark-ddl.md
```diff
@@ -98,6 +98,13 @@ CREATE TABLE prod.db.sample
 USING iceberg
 AS SELECT ...
 ```
+```sql
+CREATE TABLE prod.db.sample
+USING iceberg
+PARTITIONED BY (part)
+TBLPROPERTIES ('key'='value')
+AS SELECT ...
```
kbendick, line 106: I think it would make more sense to either:
1) Update the existing CTAS statement to use partitioning plus tblproperties.
2) Add a comment in between the two blocks indicating what's different between the two.

I would prefer (1) given that the CTAS syntax is not iceberg specific, this is just covering more Spark DML.

kyle-cx: hi @kbendick, thx for the advice. I made some changes base on ur suggestion, here is what it looks like now
cc @samredai 

image

docs/spark/spark-ddl.md
```diff
@@ -91,11 +91,15 @@ Supported transformations are:
 
 ## `CREATE TABLE ... AS SELECT`
 
-Iceberg supports CTAS as an atomic operation when using a [`SparkCatalog`](../spark-configuration#catalog-configuration). CTAS is supported, but is not atomic when using [`SparkSessionCatalog`](../spark-configuration#replacing-the-session-catalog).
+Iceberg supports CTAS as an atomic operation when using a [`SparkCatalog`](../spark-configuration#catalog-configuration). CTAS is supported, but is not atomic when using [`SparkSessionCatalog`](../spark-configuration#replacing-the-session-catalog).
```
rdblue, line 94: Can you revert the change to this line? Looks like this added trailing whitespace.

docs/spark/spark-ddl.md
```diff
@@ -91,11 +91,15 @@ Supported transformations are:
...
-Iceberg supports CTAS as an atomic operation when using a [`SparkCatalog`](../spark-configuration#catalog-configuration). CTAS is supported, but is not atomic when using [`SparkSessionCatalog`](../spark-configuration#replacing-the-session-catalog).
+Iceberg supports CTAS as an atomic operation when using a [`SparkCatalog`](../spark-configuration#catalog-configuration). CTAS is supported, but is not atomic when using [`SparkSessionCatalog`](../spark-configuration#replacing-the-session-catalog). 
+
+The newly created table won't inherit the partition spec and table properties from the source table in SELECT, you can use PARTITIONED BY and TBLPROPERTIES  in CTAS to declare partition spec and table properties for the new table.
 
 ```sql
 CREATE TABLE prod.db.sample
 USING iceberg
+PARTITIONED BY (part)
+TBLPROPERTIES ('key'='value')
```
rdblue, line 102: Documentation should start with simple examples and add more complexity if needed.

In this case, I think it is valid to add a `PARTITIONED BY` example, but not by modifying the first simple example. Please add the new paragraph and a separate example after the existing introduction to CTAS.

kyle-cx:
@rdblue thx for the guide, and here is what it looks like now

image

docs/spark/spark-ddl.md
```diff
@@ -98,6 +98,15 @@ CREATE TABLE prod.db.sample
 USING iceberg
 AS SELECT ...
 ```
+```sql
```
samredai, line 101: Can you move the paragraph below up here so it's before the sql snippet? Then I think this is good to merge 👍

kyle-cx: @samredai sure, here is what it looks like now

image

hililiwei:
LGTM

samredai (approved):
LGTM!

samredai:
@kyle-cx would you mind rebasing this so we can merge it in? The `spark-ddl.md` file has been moved up one directory but I think this should still rebase cleanly.

rdblue:
These changes were merged in #6020. Thanks to @samredai for porting the PR and thanks to @kyle-cx for contributing it!

PR #4608, opened 2022-04-21
Add RangeReadable interface for range base FileIO reads

danielcweeks:
This PR adds an optional interface, `RangeReadable`, which allows for positional, range-based reads to support object stores like S3 that prefer bounded reads.


api/src/main/java/org/apache/iceberg/io/RangeReadable.java
```diff
@@ -0,0 +1,60 @@
...
+  void readFully(long position, byte [] buffer, int offset, int length);
+
+  /**
+   * Fill the entire buffer with the contents of the input source starting
+   * at {@code position}.
+   *
+   * @param position start position of the read
+   * @param buffer target buffer to copy data
+   */
+  default void readFully(long position, byte [] buffer) {
```
kbendick, line 57: Nit: Looks like an extra space between `byte` and `[]`.

danielcweeks: Thanks, I've updated it.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -111,6 +114,19 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+  public void readFully(long position, byte[] buffer, int offset, int length) {
+    GetObjectRequest.Builder requestBuilder = GetObjectRequest.builder()
+        .bucket(location.bucket())
+        .key(location.key())
+        .range(String.format("bytes=%s-%s", position, position + length));
+
+    S3RequestUtil.configureEncryption(awsProperties, requestBuilder);
+
+    ResponseBytes<GetObjectResponse> response = s3.getObject(requestBuilder.build(), ResponseTransformer.toBytes());
+    System.arraycopy(response.asByteArrayUnsafe(), 0, buffer, offset, length);
```
kbendick, line 127: Should we be validating that `buffer` is big enough to hold the data from `offset` to `length`?

danielcweeks: `System.arraycopy` is going to be making that same check, so I feel like all we save is the request for an erroneous case.  Feels like unnecessary validation to me, but if others feel strongly, I can add a `Preconditions::checkPositionIndexes`.

danielcweeks:
@electrum I just an FYI, this adds the range based read and simplifies this path significantly as compared to AWS SDK v1 since much of the stream copy logic is handled by SDK v2.  This possibly save an array copy as well.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
rdblue: I would probably do this because we can catch the problem before making an S3 request. It's not a big deal, but if we can avoid an extra read operation that seems worth a simple check.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -111,6 +114,19 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) {
+    GetObjectRequest.Builder requestBuilder = GetObjectRequest.builder()
+        .bucket(location.bucket())
+        .key(location.key())
+        .range(String.format("bytes=%s-%s", position, position + length));
+
+    S3RequestUtil.configureEncryption(awsProperties, requestBuilder);
+
+    ResponseBytes<GetObjectResponse> response = s3.getObject(requestBuilder.build(), ResponseTransformer.toBytes());
```
rdblue, line 126: Looks like the `toBytes` transformer is really expensive:

```java
    /**
     * Reads and returns the rest of the given input stream as a byte array.
     * Caller is responsible for closing the given input stream.
     */
    public static byte[] toByteArray(InputStream is) throws IOException {
        try (ByteArrayOutputStream output = new ByteArrayOutputStream()) {
            byte[] b = new byte[BUFFER_SIZE];
            int n = 0;
            while ((n = is.read(b)) != -1) {
                output.write(b, 0, n);
            }
            return output.toByteArray();
        }
    }
```

While that implements the `readFully` logic for us, it doesn't seem worth creating a byte array output stream (and allocating arrays within it), allocating the temporary buffer per call, and copying the data twice, only to copy it again here.

I think we should take the `readFully` implementation from `IcebergDecoder` and update it to use offset/length:

```java
  /**
   * Reads a buffer from a stream, making multiple read calls if necessary.
   *
   * @param stream an InputStream to read from
   * @param bytes a buffer
   * @param offset starting offset in the buffer for the data
   * @param length length of bytes to copy from the input stream to the buffer
   * @return true if the buffer is complete, false otherwise (stream ended)
   * @throws IOException if there is an error while reading
   */
  @SuppressWarnings("checkstyle:InnerAssignment")
  private boolean readFully(InputStream stream, byte[] bytes, int offset, int length)
      throws IOException {
    int pos = offset;
    int bytesRead;
    while ((length - pos) > 0 &&
        (bytesRead = stream.read(bytes, pos, length - pos)) > 0) {
      pos += bytesRead;
    }
    return pos == length;
  }

```

api/src/main/java/org/apache/iceberg/io/RangeReadable.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Fill the provided buffer with the contents of the input source starting
+   * at {@code position} for the given {@code offset} and {@code length}.
+   *
+   * @param position start position of the read
+   * @param buffer target buffer to copy data
+   * @param offset offset in the buffer to copy the data
+   * @param length size of the read
+   */
+  void readFully(long position, byte[] buffer, int offset, int length);
```
rdblue, line 48: Using the `readFully` implementation I pasted below requires throwing `IOException` from these methods.

api/src/main/java/org/apache/iceberg/io/RangeReadable.java
```diff
@@ -0,0 +1,60 @@
...
+ * implementation.
+ *
+ * If the implementation is also a {@link SeekableInputStream}, the position
+ * of the stream is not required to be updated based on the positional reads
+ * performed by this interface.  Usage of {@link SeekableInputStream} should
+ * always seek to the appropriate position for {@link java.io.InputStream}
+ * based reads.
+ *
+ */
+public interface RangeReadable extends AutoCloseable {
```
rdblue, line 37: We usually extend `Closeable` rather than `AutoCloseable`. That's because most places check for `Closeable` and we would rather always pass the check. Plus, `InputStream` implements `Closeable`.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -111,6 +114,19 @@ public int read(byte[] b, int off, int len) throws IOException {
     return bytesRead;
   }
 
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) {
+    GetObjectRequest.Builder requestBuilder = GetObjectRequest.builder()
+        .bucket(location.bucket())
+        .key(location.key())
+        .range(String.format("bytes=%s-%s", position, position + length));
```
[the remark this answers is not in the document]
danielcweeks: Nice catch, this was somewhat silently handled by the way we were handling the response.  Turns out S3 also allows you to specify a range larger than the object but only returning the up to the end of the object.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
danielcweeks: Ok, I pulled this into an `IOUtil` class, but I also made it conform to other implementations of readFully (i.e. we shouldn't be returning a boolean, but rather throwing IOException in the event of reaching EOF).  I put this in the `org.apache.iceberg.io` package, but could also live in `org.apache.iceberg.util` with some of the other utils.  I don't have a strong preference.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
danielcweeks: I added checks since we're now not using the System.arraycopy, but I did take a look at many other implementations of readFully (hadoop, trino, commons) and none of them perform these checks, so we're being extra cautious here.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -111,6 +113,35 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+    String range = String.format("bytes=%s-%s", position, position + length - 1);
+
+    IOUtil.readFully(readRange(range), buffer, offset, length);
+  }
+
+  @Override
+  public void readTail(byte[] buffer, int offset, int length) throws IOException {
+    Preconditions.checkPositionIndexes(offset, offset + length, buffer.length);
+
+    String range = String.format("bytes=-%s", length);
```
danielcweeks, line 129: @rdblue and @electrum I tested this and S3 does respect the range read from the end of file, but there are a few things to note:

- if you try to read more than the object size, it will return just the size of the object (no error, just fewer bytes)
- that will not work with how we implement readFully, but there's a way to be more flexible with reading the file
- the s3 mock we use for testing however does not support this type of read, so we'll need another approach

core/src/main/java/org/apache/iceberg/io/IOUtil.java
```diff
@@ -0,0 +1,52 @@
...
+  @SuppressWarnings("checkstyle:InnerAssignment")
+  public static void readFully(
+      InputStream stream, byte[] bytes, int offset, int length) throws IOException {
+    int pos = offset;
+    int bytesRead;
+    while ((length - pos) > 0 && (bytesRead = stream.read(bytes, pos, length - pos)) > 0) {
+      pos += bytesRead;
+    }
+
+    if (pos != length) {
```
danielcweeks, line 48: @rdblue I updated this to be more consistent with other implementations since with the readFully from IcebergDecoder has two failure modes (exception and return false).

api/src/main/java/org/apache/iceberg/io/RangeReadable.java
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Fill the provided buffer with the contents of the input source starting
+   * at {@code position} for the given {@code offset} and {@code length}.
+   *
+   * @param position start position of the read
+   * @param buffer target buffer to copy data
+   * @param offset offset in the buffer to copy the data
+   * @param length size of the read
+   */
+  void readFully(long position, byte[] buffer, int offset, int length) throws IOException;
```
openinx, line 51: What will happen if the expected `length` exceed the object max file size starting from the given `position` ?  Will S3 SDK throw an IOException, or just read the remaining bytes silently ? 

If it's the later case, then I think we'd better to return the real read size in this method.

openinx: Okay,  looks like we are throwing the exception: https://github.com/apache/iceberg/pull/4608/files#diff-5a63855b4791f694674e376ee7e6c3186c7a9783704699fbfe70954fa7ce6659R41-R43

core/src/main/java/org/apache/iceberg/io/IOUtil.java
```diff
@@ -0,0 +1,67 @@
...
+   * @throws IOException if there is an error while reading
+   */
+  @SuppressWarnings("checkstyle:InnerAssignment")
+  public static int readToEnd(InputStream stream, byte[] bytes, int offset, int length) throws IOException {
+    int pos = offset;
+    int bytesRead = 0;
+
+    while ((length - bytesRead) > 0 && (bytesRead = stream.read(bytes, pos, length - bytesRead)) > 0) {
+      pos += bytesRead;
+    }
```
openinx, line 63: Is this loop correct ? 

If we have a stream with 1000 bytes, and the offset = 0,  length = 100,  assume we will read 10 bytes for each `stream.read(...)`. Then:

1. For the first read,  everything works fine, and the `pos` advanced to 10; 
2. For the second read,    it will try to read by the `stream.read(bytes, 10, 100-10) `.  Then the `pos` advanced to  20; 
3. For the third read, it will try to read by the `stream.read(bytes, 20, 100-10)`.  In fact,  we only need to read the remaining 80 bytes but we are passing the `100-10=90` bytes ?  That should be a problem....

rdblue: This was my bad. I found the wrong `readFully` implementation and didn't update it correctly.

I'm opening a PR against this branch with a better implementation and thorough tests for the method.

rdblue: Opened https://github.com/danielcweeks/iceberg/pull/1 to improve this. @danielcweeks can you take a look?

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
danielcweeks: Just to follow up on this, readTail now returns actual bytes read.

core/src/main/java/org/apache/iceberg/io/IOUtil.java
danielcweeks: Great catch @openinx and thanks for the update @rdblue.  I merged that implementation in.

rdblue (approved):

rdblue:
Thanks, @danielcweeks!

PR #4609, opened 2022-04-21
Add a generic Accessor class

samredai:
This adds a generic accessor class and adds tests that use a test `FooStruct` and `FooAccessor`.

### DIFF
--- a/python/src/iceberg/expressions/base.py
+++ b/python/src/iceberg/expressions/base.py
@@ -18,6 +18,7 @@ from abc import ABC, abstractmethod
 from enum import Enum, auto
 from typing import Any, Generic, TypeVar
 
+from iceberg.files import StructProtocol
 from iceberg.types import IcebergType
 
 T = TypeVar("T")
@@ -126,12 +127,8 @@ class Literal(Generic[T], ABC):
         return self.value >= other.value
 
 
-class Accessor(Generic[T], ABC):
-    """An accessor for a specific position in a container of type T
-
-    Args:
-        Generic[T]: The type of container that can be accessed using this Accessor (The container must implement the StructProtocol)
-    """
+class Accessor:
+    """An accessor for a specific position in a container that implements the StructProtocol"""
 
     def __init__(self, position: int, iceberg_type: IcebergType):
         self._position = position
@@ -147,14 +144,13 @@ class Accessor(Generic[T], ABC):
         """The IcebergType of the value at the accessor position"""
         return self._value_type
 
-    @abstractmethod
-    def get(self, container: T) -> Any:
-        """Returns the value at self.position in a container of type T
+    def get(self, container: StructProtocol) -> Any:
+        """Returns the value at self.position in `container`
 
         Args:
-            container(T): A container to access at position `self.position`
+            container(StructProtocol): A container to access at position `self.position`
 
         Returns:
             Any: The value at position `self.position` in the container
         """
-        ...
+        return container.get(self.position)
--- a/python/src/iceberg/expressions/base.py
+++ b/python/src/iceberg/expressions/base.py
@@ -19,7 +19,6 @@ from enum import Enum, auto
 from typing import Any, Generic, TypeVar
 
 from iceberg.files import StructProtocol
-from iceberg.types import IcebergType
 
 T = TypeVar("T")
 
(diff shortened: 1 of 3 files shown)


python/src/iceberg/expressions/base.py
```diff
@@ -122,3 +124,37 @@ def __le__(self, other):
 
     def __ge__(self, other):
         return self.value >= other.value
+
+
+class Accessor(Generic[T], ABC):
+    """An accessor for a specific position in a container of type T
```
rdblue, line 130: Do we need `T`? We parameterize that type in Java, but because types are more rigid. In Python, we can probably use the `StructProtocol` instead because a class can "implement" the protocol just by having those methods.

python/src/iceberg/expressions/base.py
```diff
@@ -122,3 +124,37 @@ def __le__(self, other):
...
+    def get(self, container: T) -> Any:
+        """Returns the value at self.position in a container of type T
+
+        Args:
+            container(T): A container to access at position `self.position`
+
+        Returns:
+            Any: The value at position `self.position` in the container
+        """
+        ...
```
rdblue, line 160: If this used the protocol, I think you could just implement it here by calling `container.get(self.position)`

python/tests/expressions/test_expressions_base.py
```diff
@@ -15,9 +15,42 @@
...
+    BooleanType,
+    DecimalType,
+    FloatType,
+    IntegerType,
+    StringType,
+    UUIDType,
+)
+
+
+class FooStruct:
```
rdblue, line 36: This implements `StructProtocol` right?

python/src/iceberg/expressions/base.py
samredai: You're right, I think I was under the impression that we'd have implementations of this class but this is a class that will be instantiated and used directly right? Something like:

```py
container = FooStruct()
container.set(0, "foo")

value = Accessor(position=0, iceberg_type=StringType).get(container)
print(value)  # "foo"
```
Does that look right?

python/src/iceberg/expressions/base.py
samredai: Good point, I think this is also related to my [misunderstanding](https://github.com/apache/iceberg/pull/4609#discussion_r855654244) of the abstraction

python/tests/expressions/test_expressions_base.py
samredai: Yes I didn't subclass it since it's not required but it would be more helpful to include it here. I'll update this 👍

python/src/iceberg/expressions/base.py
rdblue: It's really close. In Java, we pass in the JVM class for type safety since this returns the class that you pass in. In Python, there's no need. So you don't need the `iceberg_type` parameter at all when creating it. Otherwise it looks good.

samredai: Changed `Accessor` to a concrete class. 👍

python/tests/expressions/test_expressions_base.py
rdblue: Actually, I think it's a good idea not to include it since that's why we used a Protocol, right?

python/src/iceberg/expressions/base.py
samredai: Fixed!

python/src/iceberg/expressions/base.py
```diff
@@ -122,3 +125,32 @@ def __le__(self, other):
...
     def __ge__(self, other):
         return self.value >= other.value
+
+
+class Accessor:
+    """An accessor for a specific position in a container that implements the StructProtocol"""
+
+    def __init__(self, position: int, iceberg_type: IcebergType):
+        self._position = position
+        self._iceberg_type = iceberg_type
```
rdblue, line 135: Yeah, I didn't notice this before, but I'd probably remove the type. No need for it.

samredai: Removed! Thanks

python/tests/expressions/test_expressions_base.py
samredai: Ok I switched it back 👍

dramaticlly:
Accessor as a class looks like a really cool way to access the index based element in a struct like container

But i am curious if we can share some more concrete example to understand how it can/shall be used in iceberg domain? 

Guessing from module.class structure, looks like `StructProtocol` is sitting in `files.py` and accessor is probably mostly used in expressions?

rdblue:
@dramaticlly, accessors are created by a schema visitor and then used when binding an expression to a schema. The bound term (variable of a predicate) basically wraps an accessor to get the data value for evaluation. Then the predicate gets the value of the term (via accessor) and compares it to the value of the literal.

rdblue (approved):

PR #4610, opened 2022-04-22
Nessie: Support Namespace properties

nastra:
This requires https://github.com/apache/iceberg/pull/4594 in order for the failing tests to pass, thus I'll rebase the PR once #4595 is in

### DIFF
--- a/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
+++ b/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
@@ -223,16 +223,16 @@ public class NessieCatalog extends BaseMetastoreCatalog implements AutoCloseable
 
   @Override
   public boolean setProperties(Namespace namespace, Map<String, String> properties) {
-    throw new UnsupportedOperationException(
-        "Cannot set properties for namespace '" + namespace +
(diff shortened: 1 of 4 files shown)


nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -223,16 +223,16 @@ public boolean dropNamespace(Namespace namespace) throws NamespaceNotEmptyExcept
 
   @Override
   public boolean setProperties(Namespace namespace, Map<String, String> properties) {
-    throw new UnsupportedOperationException(
-        "Cannot set properties for namespace '" + namespace +
-            "': setProperties is not supported by the NessieCatalog");
+    client.setProperties(namespace, properties);
```
RussellSpitzer, line 226: I would probably just have the client method return a boolean and then force the true in there rather than at this level in the Catalog. But that's just my opinion, up to you

RussellSpitzer: On this note we should probably raise another Issue to specify what the return of this method is expected to do.

RussellSpitzer (approved):
Looks good to me, let me know if you want to make more modifications or if you want a merge now.

PR #4611, opened 2022-04-22
Fix Objects.requireNonNull usages

findepi:
Per project code style, Preconditions.checkNotNull should be used
instead.

### DIFF
--- a/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
+++ b/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
@@ -21,7 +21,6 @@ package org.apache.iceberg.nessie;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import org.apache.hadoop.conf.Configurable;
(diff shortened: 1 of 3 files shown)


findepi:
> Looks good to me, is this something we have in our Static Analysis rules?

i don't think so. That's something reviewers point out during review process.

RussellSpitzer:
> > Looks good to me, is this something we have in our Static Analysis rules?
> 
> i don't think so. That's something reviewers point out during review process.

We should probably add that as well then, we can have checkstyle automatically fail any code which adds it like we do with usages of unshaded guava.

findepi:
Good idea, let's do this. I don't plan on doing this within this PR though.

RussellSpitzer:
Uh looks like we already had some :) 
```
Error: eckstyle] [ERROR] /home/runner/work/iceberg/iceberg/nessie/src/main/java/org/apache/iceberg/nessie/UpdateableReference.java:53: Use Preconditions.checkNotNull(Object, String). [RegexpSinglelineJava]
```

RussellSpitzer:
https://github.com/apache/iceberg/issues/4613

RussellSpitzer:
Just waiting on tests here

kbendick (approved):
lgtm

PR #4612, opened 2022-04-22
Core:Optimize Error msg in validateDeletedManifests (fix #3466)

### LINKED ISSUE #3466: "Manifest is missing" ValidationException when there have Concurrent applications to rewrite manifests
we concurrently run a lots of sql-shell to overwrite different day's data in iceberg table, and every application end with "CALL spark_catalog.system.rewrite_manifests(table => 'dwm.tableA', use_caching => false)". The applications will rewrite the tableA's manifests concurrently , and throw ValidationException("Manifest is missing: xxx") in validateDeletedManifests method.
`		 diagnostics: User class threw exception: org.apache.spark.SparkException: org.apache.iceberg.exceptions.ValidationException: Manifest is missing: oss://xgimi-data/apps/spark/warehouse/ods.db/screen_event_log_hi/metadata/52c1e98f-02a5-4ce3-ae05-d7382a6a68c2-m2.avro
		at org.apache.iceberg.BaseRewriteManifests.lambda$validateDeletedManifests$7(BaseRewriteManifests.java:261)
		at java.util.Optional.ifPresent(Optional.java:159)
		at org.apache.iceberg.BaseRewriteManifests.validateDeletedManifests(BaseRewriteManifests.java:260)
		at org.apache.iceberg.BaseRewriteManifests.apply(BaseRewriteManifests.java:169)
		at org.apache.iceberg.SnapshotProducer.apply(SnapshotProducer.java:163)
		at org.apache.iceberg.BaseRewriteManifests.apply(BaseRewriteManifests.java:53)
		at org.apache.iceberg.SnapshotProducer.lambda$commit$2(SnapshotProducer.java:276)
		at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:404)
		at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:213)
		at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:197)
		at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:189)
		at org.apache.iceberg.SnapshotProducer.commit(SnapshotProducer.java:275)
		at org.apache.iceberg.BaseRewriteManifests.commit(BaseRewriteManifests.java:53)
		at org.apache.iceberg.actions.BaseSnapshotUpdateAction.commit(BaseSnapshotUpdateAction.java:40)
		at org.apache.iceberg.actions.RewriteManifestsAction.replaceManifests(RewriteManifestsAction.java:309)
		at org.apache.iceberg.actions.RewriteManifestsAction.execute(RewriteManifestsAction.java:196)`

kyle-cx91:
Error msg "Manifest is missing" in BaseRewriteManifests.validateDeletedManifests()  is correct but a bit confusing, I optimize it to make user easier found out what happened. and now it looks like:
```
"Cannot apply RewriteManifests result, " +
                  "since some manifests being deleted have already been removed in current snapshot or not exist" +
                  " at all. current-snapshot-id:%d, missing-manifest-path:%s"
```

fix https://github.com/apache/iceberg/issues/3466
follow the suggestion from https://github.com/apache/iceberg/issues/3466#issuecomment-1104404998
cc: @RussellSpitzer

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java
+++ b/core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java
@@ -165,7 +165,7 @@ public class BaseRewriteManifests extends SnapshotProducer<RewriteManifests> imp
     List<ManifestFile> currentManifests = base.currentSnapshot().dataManifests();
     Set<ManifestFile> currentManifestSet = ImmutableSet.copyOf(currentManifests);
 
-    validateDeletedManifests(currentManifestSet);
+    validateDeletedManifests(currentManifestSet, base.currentSnapshot().snapshotId());
 
     if (requiresRewrite(currentManifestSet)) {
       performRewrite(currentManifests);
@@ -251,13 +251,15 @@ public class BaseRewriteManifests extends SnapshotProducer<RewriteManifests> imp
     }
   }
 
-  private void validateDeletedManifests(Set<ManifestFile> currentManifests) {
+  private void validateDeletedManifests(Set<ManifestFile> currentManifests, long currentSnapshotId) {
     // directly deleted manifests must be still present in the current snapshot
     deletedManifests.stream()
         .filter(manifest -> !currentManifests.contains(manifest))
         .findAny()
         .ifPresent(manifest -> {
-          throw new ValidationException("Manifest is missing: %s", manifest.path());
+          throw new ValidationException("Cannot apply RewriteManifests result, " +
+                  "since some manifests being deleted have already been removed in current snapshot or not exist" +
+                  " at all. current-snapshot-id:%d, missing-manifest-path:%s", currentSnapshotId, manifest.path());
         });
   }
 
--- a/core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java
+++ b/core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java
@@ -165,7 +165,7 @@ public class BaseRewriteManifests extends SnapshotProducer<RewriteManifests> imp
     List<ManifestFile> currentManifests = base.currentSnapshot().dataManifests();
     Set<ManifestFile> currentManifestSet = ImmutableSet.copyOf(currentManifests);
 
-    validateDeletedManifests(currentManifestSet, base.currentSnapshot().snapshotId());
+    validateDeletedManifests(currentManifestSet);
 
     if (requiresRewrite(currentManifestSet)) {
       performRewrite(currentManifests);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java
```diff
@@ -251,13 +251,15 @@ private void performRewrite(List<ManifestFile> currentManifests) {
...
 
-  private void validateDeletedManifests(Set<ManifestFile> currentManifests) {
+  private void validateDeletedManifests(Set<ManifestFile> currentManifests, long currentSnapshotId) {
     // directly deleted manifests must be still present in the current snapshot
     deletedManifests.stream()
         .filter(manifest -> !currentManifests.contains(manifest))
         .findAny()
         .ifPresent(manifest -> {
-          throw new ValidationException("Manifest is missing: %s", manifest.path());
+          throw new ValidationException("Cannot apply RewriteManifests result, " +
```
RussellSpitzer, line 260: I think in this message we want to make clear that this is not a corrupt table issue and that it is most likely the result of a concurrent manifest operation. Maybe something like

"Cannot commit RewriteManifests; manifests that existed at the beginning of this rewrite have already been removed by another operation. Manifest %s, which would be replaced by this operation, has already been removed."?

I'm not sure the snapshot ID really helps unless you also pass the snapshot that the operation started on, but even with that information I'm not sure a User can do much.

flyrain: Agreed with @RussellSpitzer. The action `RewriteManifests` doesn't need users to input any snapshot id. A user couldn't do much with the snapshot id.

kyle-cx: @RussellSpitzer thx for the guide, I have already remove the snapshot id and use the msg u suggested.

core/src/main/java/org/apache/iceberg/BaseRewriteManifests.java
```diff
@@ -257,7 +257,10 @@ private void validateDeletedManifests(Set<ManifestFile> currentManifests) {
         .filter(manifest -> !currentManifests.contains(manifest))
         .findAny()
         .ifPresent(manifest -> {
-          throw new ValidationException("Manifest is missing: %s", manifest.path());
+          throw new ValidationException("Cannot commit RewriteManifests; " +
+                  "manifests that existed at the beginning of this rewrite have already been removed by another " +
+                  "operation. Manifest %s, which would be replaced by this operation, has already been removed.",
+                  manifest.path());
```
rdblue, line 263: I think this is too long to be an effective error message. I think your intent is to tell the end user that something changed. But there isn't anything that user can do besides retry the operation. I think the short message works fine, but you could also extend it to "Manifest is missing: %s (table changed concurrently)". That's more clear without being confusing or overly long.

RussellSpitzer: I think "Manifest is missing" is a very confusing message. From my talks with folks who have hit this as well as other Iceberg dev's, the gut reaction to this message is that the table has become corrupted. After all why would "manifest is missing" be an issue for an operation which is removing manifests? Having a missing manifest also generally means that the table is no longer stable so users are on high alert.

To further confound things, the manifest probably still exists and a user can find it, which leads them to ask "Why would Iceberg think the file is missing when it is clearly on disk?". Users often let me know in conjunction that Iceberg says a manifest is missing but they can see it on disk

In my opinion a good error message has to explain what went wrong, why it went wrong, and how to respond (if such a response is possible and non-obvious).

At the moment the error message at best explains the "why it went wrong" but without enough context to really let the user even know what part of the rewrite actually broke.

If I was going to order information here the most important concept to get across is 
1. "What" - The commit has failed (not writing, reading or deleting manifests)
2. "Why" - The change we prepared no longer cleanly applies because a manifest we thought existed is no longer current
3. "Remediation" - Retry the operation if desired

So if we want to trim it down to the essentials
```
"Cannot commit, table has been concurrently modified and manifest %s is no longer part of the table"
```
Although even knowing which manifest is missing is not even that valid so we could remove that too.

I actually have issues with most of our Validation messages in MergingSnapshotProducer for the similar reason. Users are often confused by "Found conflicting files that can contain records matching true". But we can save that for future PR's

PR #4614, opened 2022-04-22
Spark: Spark 2.4 supports overwriting by expression

gustavoatt:
## Summary

Spark 2.4 supports overwriting partition by expression. Similarly to how in Spark 3.x we can do it via the dataframe API:

```java
df
  .write
  .mode("iceberg")
  .overwrite($ds === "2021-01-01")
  .save($TABLE)
```

It can now be done through the write option `overwrite-partitions`:

```java
df
  .write
  .mode("iceberg")
  .option("overwrite-partitions", "ds='2021-01-01'")
  .save($TABLE).
```

## Background

This work is motivated due to the lack of support of `INSERT OVERWRITE PARTITION($VAL) ...` on Spark 2. As of now there was no way of replacing a partition when multiple partition specs are present on a table.

More details can be found on this [slack thread](https://apache-iceberg.slack.com/archives/C025PH0G1D4/p1649869204622389)


spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -1415,6 +1415,89 @@ public void testRemoveOrphanFilesActionSupport() throws InterruptedException {
...
+            .format("iceberg")
+            .mode("overwrite")
+            .option("overwrite-partitions", "id=1")
+            .save(loadLocation(tableIdentifier));
+
+    result = spark.read()
+            .format("iceberg")
+            .load(loadLocation(tableIdentifier))
+            .as(Encoders.bean(SimpleRecord.class))
+            .collectAsList();
```
rdblue, line 1448: Can you fix the formatting of this file? Continuation indents should be 4 spaces, not 8.

spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/source/Writer.java
```diff
@@ -173,6 +185,38 @@ private void replacePartitions(WriterCommitMessage[] messages) {
...
+    final Expression overwriteExpression = parseOverwritePartitionFilter(overwriteFilter);
+    Iterable<DataFile> files = files(messages);
+
+    if (!files.iterator().hasNext()) {
+      LOG.info("Static overwrite is empty, skipping commit");
+      return;
+    }
+
+    OverwriteFiles overwriteFiles = table.newOverwrite()
+            .overwriteByRowFilter(overwriteExpression);
```
rdblue, line 198: Indentation is off here as well.

spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/source/Writer.java
```diff
@@ -66,10 +71,13 @@
 class Writer implements DataSourceWriter {
   private static final Logger LOG = LoggerFactory.getLogger(Writer.class);
 
+  private final SparkSession spark;
   private final JavaSparkContext sparkContext;
   private final Table table;
   private final FileFormat format;
   private final boolean replacePartitions;
+
```
rdblue, line 79: Nit: unnecessary whitespace.

spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/source/Writer.java
```diff
@@ -112,8 +122,10 @@ public DataWriterFactory<InternalRow> createWriterFactory() {
 
   @Override
   public void commit(WriterCommitMessage[] messages) {
-    if (replacePartitions) {
+    if (replacePartitions && overwriteFilter == null) {
       replacePartitions(messages);
+    } else if (replacePartitions) {
```
rdblue, line 127: Shouldn't this be `if (overwriteFilter != null)`? What happens if this isn't using overwrite mode but you give it an overwrite filter? Should it fail?

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -1415,6 +1415,89 @@ public void testRemoveOrphanFilesActionSupport() throws InterruptedException {
...
+
+    List<Row> insertRecords = Lists.newArrayList(
+            RowFactory.create("2021-01-01", 100)
+    );
+    spark.createDataFrame(insertRecords, originalSparkSchema)
+            .select("ds", "data")
+            .write()
+            .format("iceberg")
+            .option("overwrite-partitions", "ds='2021-01-01'")
+            .mode(SaveMode.Overwrite)
```
rdblue, line 1490: We prefer using the `String` version of `mode` so that we don't need to import and use Scala classes.

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -1415,6 +1415,89 @@ public void testRemoveOrphanFilesActionSupport() throws InterruptedException {
...
+            .save(loadLocation(identifier));
+
+    // When reading we should only see two rows still
+    readInputDf = spark.read()
+            .format("iceberg")
+            .load(loadLocation(identifier));
+    Assert.assertEquals(2, readInputDf.count());
```
rdblue, line 1497: It would be better to assert that the records match the expected ones.

spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/source/Writer.java
gustavoatt: It should, but the `overwriteFilter != null` condition is always true at this point, since otherwise it would entered the branch at line 125. Added the condition for better readability.

Good point about what happens when the overwrite mode is set to `append` but we give an overwrite filter. I throw an exception here when that happens. Should I throw here or rather on `IcebergSource` just before creating the writer?

ajantha-bhat:
Closing this PR as spark-2.4 module is removed from master code.

PR #4617, opened 2022-04-22
Style: Add checkstyle rule to prevent using Objects.requireNonNull

### LINKED ISSUE #4613: Add Checkstyle Rule for Objects.notNull
We prefer Preconditions.notNull in the project (cc: @findepi ) so we should make it a failing rule in the checkstyle tests. Like we ban usages of new HashMap() here

https://github.com/apache/iceberg/blob/master/.baseline/checkstyle/checkstyle.xml#L54-L57

tal66:
Hi, this closes #4613.

running the checks on this simple example, :
```
    public static void main(String[] args)  {
        String str = null;

        // not allowed
        Objects.nonNull(str);
        Objects.nonNull(Integer.valueOf(123));
        Objects.requireNonNull(str);
        Objects.requireNonNullElse(str, "123");
        Objects.requireNonNullElseGet(str, () -> "1");

        // allowed
        Objects.isNull(str);
    }
```
generates: `<error line="..." severity="error" message="Prefer using Preconditions.checkNotNull(Object, String)." source=...` for the 'not allowed' lines above.

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -337,6 +337,10 @@
             <property name="format" value="Preconditions\.checkNotNull\((?!.*,)([^()]*(\(([^()]*|\(([^()]*|\([^()]*\))*\))*\)[^()]*)*)\)"/>
             <property name="message" value="Use Preconditions.checkNotNull(Object, String)."/>
         </module>
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="Objects\.(require|)[Nn]onNull.*\(.*\)"/>
+            <property name="message" value="Prefer using Preconditions.checkNotNull(Object, String)."/>
+        </module>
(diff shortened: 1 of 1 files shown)


kbendick (approved):
I tried this locally with `Objects.requireNonNull(obj)` and `Objects.requireNonNull(obj, "error message)` and both of them failed checkstyle.

LGTM provided tests are passing. Thanks @tal66!

singhpk234 (approved):
LGTM as well, Thanks @tal66

rdblue (approved):

rdblue:
Thanks, @tal66!

nastra:
just curious, why wouldn't we want to allow something like `if (Objects.nonNull(x))`? The checkstyle message suggests to use `Preconditions.checkNotNull` instead, but that's not really the same. I can see why we'd want to use `Preconditions.checkNotNull` rather than `Objects.requireNonNull` but not why we'd prevent usage of `Objects.nonNull`, since it's just returning a `boolean`

PR #4619, opened 2022-04-23
Add BoundReference class

samredai:
This adds a `BoundReference` class that takes a field and an accessor and allows you to evaluate a struct (a row object that abides by the `StructProtocol`) to retrieve the value at the referenced field's position.

For some more context, expressions are initially built with references (`x` in x = 1`) that are not bound to an actual field. The binding of the reference to the actual field `x` will return a `BoundReference` instance. The instance includes an `eval` method that uses the accessor to get the field at that position, given a row object.

### DIFF
--- a/python/src/iceberg/expressions/base.py
+++ b/python/src/iceberg/expressions/base.py
@@ -20,7 +20,7 @@ from functools import reduce
 from typing import Any, Generic, TypeVar
 
 from iceberg.files import StructProtocol
-from iceberg.types import Singleton
+from iceberg.types import IcebergType, NestedField, Singleton
 
 T = TypeVar("T")
 
@@ -287,3 +287,61 @@ class Accessor:
             Any: The value at position `self.position` in the container
         """
         return container.get(self.position)
+
+
+class BoundReference:
+    """A reference bound to a field with an accessor for acquiring the field's value
+
+    Args:
+        field (NestedField): A referenced field in an Iceberg schema
+        accessor (Accessor): An Accessor object to access the value at the field's position
+    """
+
+    def __init__(self, field: NestedField, accessor: Accessor):
+        self._field = field
+        self._accessor = accessor
+
+    def __str__(self):
+        return f"ref(id={self.field_id})"
+
+    def __repr__(self):
+        return f"BoundReference(field={repr(self.field)}, accessor={repr(self.accessor)})"
+
+    @property
+    def field(self) -> NestedField:
+        """The referenced field"""
+        return self._field
+
+    @property
+    def field_id(self) -> int:
+        """The ID referenced field"""
+        return self._field.field_id
+
+    @property
+    def field_type(self) -> IcebergType:
+        """The type of the referenced field"""
+        return self._field.type
+
+    @property
+    def accessor(self) -> Accessor:
+        """The accessor for retrieving the value at the referenced field's position"""
+        return self._accessor
+
+    def __eq__(self, other) -> bool:
+        if id(self) == id(other):
+            return True
+        elif other is None or not isinstance(other, BoundReference):
+            return False
+
+        return self.field_id == other.field_id and self.field_type == other.field_type
+
+    def eval(self, struct: StructProtocol) -> Any:
+        """Returns the value at the referenced field's position in an object that abides by the StructProtocol
+
+        Args:
+            struct (StructProtocol): A row object that abides by the StructProtocol and returns values given a position
+
+        Returns:
+            Any: The value at the referenced field's position in `struct`
+        """
+        return self.accessor.get(struct)
--- a/python/src/iceberg/expressions/base.py
+++ b/python/src/iceberg/expressions/base.py
@@ -20,7 +20,7 @@ from functools import reduce
 from typing import Any, Generic, TypeVar
 
 from iceberg.files import StructProtocol
-from iceberg.types import IcebergType, NestedField, Singleton
+from iceberg.types import NestedField, Singleton
 
 T = TypeVar("T")
 
@@ -272,6 +272,12 @@ class Accessor:
     def __init__(self, position: int):
         self._position = position
 
+    def __str__(self):
+        return f"Accessor(position={self._position})"
+
+    def __repr__(self):
+        return f"Accessor(position={self._position})"
+
     @property
     def position(self):
         """The position in the container to access"""
@@ -290,7 +296,7 @@ class Accessor:
 
 
 class BoundReference:
-    """A reference bound to a field with an accessor for acquiring the field's value
+    """A reference bound to a field in a schema
 
     Args:
         field (NestedField): A referenced field in an Iceberg schema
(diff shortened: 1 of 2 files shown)


samredai:
cc: @dramaticlly @CircArgs @jun-he @dhruv-pratap

python/src/iceberg/expressions/base.py
```diff
@@ -147,3 +148,61 @@ def get(self, container: StructProtocol) -> Any:
...
+    @property
+    def accessor(self) -> Accessor:
+        """The accessor for retrieving the value at the referenced field's position"""
+        return self._accessor
+
+    def __eq__(self, other) -> bool:
+        if id(self) == id(other):
+            return True
+        elif other is None or not isinstance(other, BoundReference):
+            return False
```
CircArgs, line 195: If you try to compare to another BoundReference here with a different `id` then doesn't this return `None`?: Misses the first branch, it's not None but is  BoundReference -> No more branches, function returns None.

not sure if this is desired to just boil down to `self is other` (I think equivalent to the first branch) e.g. the default `__eq__` if I'm not mistaken, in which case we don't even need to implement this

samredai: In the path you described it would fall to the return statement at the end of the function:
```py
        return self.field_id == other.field_id and self.field_type == other.field_type
```
I don't yet have full context around all scenarios where we'd compare against a `BoundReference` but the if/else here is just to get the easy cases (True is the id is the same, False if `other` isn't even a `BoundReference`). In the case where the ids differ but `other` is a `BoundReference`, it checks if the field being bound is the same ID and Type.

One thing I'm noticing though is that `elif other is None or not isinstance(other, BoundReference)` is redundant and can be changed to just `elif not isinstance(other, BoundReference)`. Updated it!

samredai: I just realized your point about how using `is` is the equivalent of the first branch. Updated!

rdblue:
Can you rebase this? Looks like there are conflicts.

python/src/iceberg/expressions/base.py
```diff
@@ -147,3 +148,61 @@ def get(self, container: StructProtocol) -> Any:
             Any: The value at position `self.position` in the container
         """
         return container.get(self.position)
+
+
+class BoundReference:
+    """A reference bound to a field with an accessor for acquiring the field's value
```
rdblue, line 154: Accessors are internal, so there is no need to mention them in docs. This doesn't necessarily need to use an accessor to work.

python/src/iceberg/expressions/base.py
```diff
@@ -147,3 +148,61 @@ def get(self, container: StructProtocol) -> Any:
...
+
+    def __init__(self, field: NestedField, accessor: Accessor):
+        self._field = field
+        self._accessor = accessor
+
+    def __str__(self):
+        return f"ref(id={self.field_id})"
+
+    def __repr__(self):
+        return f"BoundReference(field={repr(self.field)}, accessor={repr(self.accessor)})"
```
rdblue, line 169: Accessor doesn't have a repr. We should either add it or not use it here.

python/src/iceberg/expressions/base.py
```diff
@@ -147,3 +148,61 @@ def get(self, container: StructProtocol) -> Any:
...
+        """The ID referenced field"""
+        return self._field.field_id
+
+    @property
+    def field_type(self) -> IcebergType:
+        """The type of the referenced field"""
+        return self._field.type
+
+    @property
+    def accessor(self) -> Accessor:
```
rdblue, line 187: Why expose all of these properties? It seems simpler just to expose `eval`.

python/src/iceberg/expressions/base.py
```diff
@@ -147,3 +148,61 @@ def get(self, container: StructProtocol) -> Any:
...
+        """The accessor for retrieving the value at the referenced field's position"""
+        return self._accessor
+
+    def __eq__(self, other) -> bool:
+        if self is other:
+            return True
+        elif not isinstance(other, BoundReference):
+            return False
+
+        return self.field_id == other.field_id and self.field_type == other.field_type
```
rdblue, line 197: I don't think I agree with this definition of equality. I think that equality comes down to the accessor (position) and the type if we were to implement it... But we may want to be more strict and include the entire field, not just field ID.

We purposely don't implement equality in the JVM implementation because it's weird to define what "equals" means for an expression: is it that the expression produces the same result? Or is it the exact same expression including the referenced field? If it is the latter, isn't that restrictive because we don't really need the field in this class, just the type and the accessor?

Rather than answer those questions, we test using the behavior and the exposed API fields -- basically just `field_type` and `eval`.

python/src/iceberg/expressions/base.py
samredai: Thanks for catching this, I added a repr to Accessor.

python/src/iceberg/expressions/base.py
samredai: I updated this description to "A reference bound to a field in a schema" but I'm not sure I fully understand what you mean by:
> This doesn't necessarily need to use an accessor to work.

Can you describe how this would be used without an accessor? I know this class won't be exposed at the top level API but I'd imagine a developer working in this area of the code would find a description of the relationship to accessors as helpful.

python/src/iceberg/expressions/base.py
samredai: Updated this to just expose `eval` and `field`. Should I remove the `field` property as well?

python/src/iceberg/expressions/base.py
samredai: I removed `__eq__` for now. We can add it later if it's required.

python/tests/expressions/test_expressions_base.py
```diff
@@ -217,3 +217,43 @@ def test_accessor_base_class(foo_struct):
...
+    assert str(bound_ref) == f"BoundReference(field={repr(field)}, accessor={repr(position1_accessor)})"
+    assert repr(bound_ref) == f"BoundReference(field={repr(field)}, accessor={repr(position1_accessor)})"
+
+
+def test_bound_reference_field_property():
+    """Test str and repr of BoundReference"""
+    field = NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False)
+    position1_accessor = base.Accessor(position=1)
+    bound_ref = base.BoundReference(field=field, accessor=position1_accessor)
+    assert bound_ref.field == NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False)
```
rdblue, line 236: I'm not sure I'd expose `field`, but it should be fine for now.

rdblue (approved):

python/src/iceberg/expressions/base.py
rdblue: I probably would. I think the relevant information is the value type that is produced, not the field info.

python/src/iceberg/expressions/base.py
rdblue: I don't think it would work without an accessor right now. But you don't need to tie the implementation to an accessor. That's an internal detail.

rdblue:
Thanks, @samredai!

PR #4620, opened 2022-04-23
Checkstyle: Add Checkstyle Rule to prevent Map<StructLike, ...> and Set<StructLike>

### LINKED ISSUE #4616: Add Checkstyle Rule to prevent Map<StructLike, ...> and Set<StructLike>
StructLike does not force the proper implementation of `hashCode` or `equals` and because of this is cannot be used in Sets or Maps safely. If the wrong implementation gets placed inside these containers they will not behave as expected.

Users should instead use

[StructLikeMap](https://github.com/apache/iceberg/blob/30639ef316a899480a3d0c88f8ea6498d9a21306/core/src/main/java/org/apache/iceberg/util/StructLikeMap.java#L32) or [StructLikeSet](https://github.com/apache/iceberg/blob/30639ef316a899480a3d0c88f8ea6498d9a21306/core/src/main/java/org/apache/iceberg/util/StructLikeSet.java#L33)

Another option is to use the concrete StructLike that particular map or set requires if applicable.

Checkstyle rules can be added like

https://github.com/apache/iceberg/blob/master/.baseline/checkstyle/checkstyle.xml#L54-L57

felixYyu:
this PR is close #4616

### DIFF
--- a/core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
+++ b/core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
@@ -55,7 +55,7 @@ abstract class FanoutWriter<T, R> implements PartitioningWriter<T, R> {
   }
 
   private FileWriter<T, R> writer(PartitionSpec spec, StructLike partition) {
-    StructLikeMap<FileWriter<T, R>> specWriters = writers.computeIfAbsent(
+    Map<StructLike, FileWriter<T, R>> specWriters = writers.computeIfAbsent(
         spec.specId(),
         id -> StructLikeMap.create(spec.partitionType()));
     FileWriter<T, R> writer = specWriters.get(partition);
--- a/core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
+++ b/core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
@@ -55,7 +55,7 @@ abstract class FanoutWriter<T, R> implements PartitioningWriter<T, R> {
   }
 
   private FileWriter<T, R> writer(PartitionSpec spec, StructLike partition) {
-    Map<StructLike, FileWriter<T, R>> specWriters = writers.computeIfAbsent(
+    StructLikeMap<FileWriter<T, R>> specWriters = writers.computeIfAbsent(
         spec.specId(),
         id -> StructLikeMap.create(spec.partitionType()));
     FileWriter<T, R> writer = specWriters.get(partition);
@@ -79,7 +79,7 @@ abstract class FanoutWriter<T, R> implements PartitioningWriter<T, R> {
   }
 
   private void closeWriters() throws IOException {
-    for (Map<StructLike, FileWriter<T, R>> specWriters : writers.values()) {
+    for (StructLikeMap<FileWriter<T, R>> specWriters : writers.values()) {
       for (FileWriter<T, R> writer : specWriters.values()) {
         writer.close();
         addResult(writer.result());
--- a/core/src/main/java/org/apache/iceberg/util/PartitionSet.java
+++ b/core/src/main/java/org/apache/iceberg/util/PartitionSet.java
@@ -87,7 +87,7 @@ public class PartitionSet implements Set<Pair<Integer, StructLike>> {
   }
 
   public boolean add(int specId, StructLike struct) {
-    StructLikeSet partitionSet = partitionSetById.computeIfAbsent(specId,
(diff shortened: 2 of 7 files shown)


felixYyu:
cc: @RussellSpitzer Can you take a look if you have time?

core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
```diff
@@ -55,7 +55,7 @@ public void write(T row, PartitionSpec spec, StructLike partition) {
   }
 
   private FileWriter<T, R> writer(PartitionSpec spec, StructLike partition) {
-    Map<StructLike, FileWriter<T, R>> specWriters = writers.computeIfAbsent(
+    StructLikeMap<FileWriter<T, R>> specWriters = writers.computeIfAbsent(
```
RussellSpitzer, line 58: :+1: I think this was fine before just because the actual implementation that uses this always returns a StructLikeMap (See line 60).

core/src/main/java/org/apache/iceberg/util/PartitionSet.java
```diff
@@ -87,7 +87,7 @@ public boolean add(Pair<Integer, StructLike> pair) {
   }
 
   public boolean add(int specId, StructLike struct) {
-    Set<StructLike> partitionSet = partitionSetById.computeIfAbsent(specId,
+    StructLikeSet partitionSet = partitionSetById.computeIfAbsent(specId,
```
RussellSpitzer, line 90: Ok these were also safe before since we were always using StructLikeSet under the hood

core/src/main/java/org/apache/iceberg/util/StructLikeMap.java
```diff
@@ -104,7 +104,7 @@ public void clear() {
   }
 
   @Override
-  public Set<StructLike> keySet() {
```
RussellSpitzer, line 107: I'm not sure we can change these since we'll be breaking an api that is more or less public .... I need to think on this and the StructLikeSet methods ... 

I'm pretty sure we don't want to break the api, but we also don't want to encourage folks using raw Set's and Maps of StructLike but in this case they would be safe since we know they came from StructLikeMap and StructLikeSet ....

felixYyu: Sometimes can use the original, sometimes can't，Then rule `<property name="format" value="^Map&lt;StructLike, .*&gt;"/>` may not be able to cover the whole.

RussellSpitzer: Yes, but in that case we put an exception for this signature. All checkstyle rules allow for overrides for particular lines, we don't need to have the check style rule be perfect.

core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
felixYyu: reverted

core/src/main/java/org/apache/iceberg/util/PartitionSet.java
felixYyu: reverted

core/src/main/java/org/apache/iceberg/util/StructLikeMap.java
felixYyu: ok, I think we revert that first, If has problem can use SuppressWarnings annotation.

core/src/main/java/org/apache/iceberg/io/FanoutWriter.java
RussellSpitzer: I don't think you should revert your change here. I just was checking to make sure it wasn't bugged

core/src/main/java/org/apache/iceberg/util/PartitionSet.java
RussellSpitzer: Same here, I was just checking to make sure that there wasn't a bug here we needed to investigate

felixYyu:
Could you take a look again, Thanks @RussellSpitzer

core/src/main/java/org/apache/iceberg/util/StructLikeMap.java
RussellSpitzer: This is the change I think we need to revert, I get worried about changing the public api of this class. I think just suppressing the rule for this usage is fine

felixYyu: ok

PR #4623, opened 2022-04-24
Core: Add ByteBufferInputStream and implementations

rdblue:
This copies an abstract class, `ByteBufferInputStream`, and its two implementations, `SingleBufferInputStream` and `MultiBufferInputStream` from Parquet. This is to support the in-memory streams that are proposed for tests (#4534) and for adding caching to `FileIO` (#4518).

This also copies the tests and updates the LICENSE file.


danielcweeks (approved):

rdblue:
Thanks @danielcweeks!

core/src/main/java/org/apache/iceberg/io/MultiBufferInputStream.java
```diff
@@ -0,0 +1,345 @@
...
+    nextBuffer();
+  }
+
+  @Override
+  public long getPos() {
+    return position;
+  }
+
+  public void seek(long newPosition) throws IOException {
+    if (newPosition > length) {
```
steveloughran, line 70: might want a check for a -ve position here

rdblue: Good point

Issue #4251, opened 2022-03-02
Proposal: Support partition with transform for Flink SQL


yittg:
For now, only the `identity` partition transform is supported with flink SQL. To support other generic transforms, it is expected that Flink SQL supports defining partitions with transform. It depends on the support of Flink engine and the corresponding transformers. Here proposes another way to support partition with transform by composing the computed columns and partition list in flink.

The basic idea is that defining the partition field as a computed column in flink, with the mapping from the expression in Flink to the transform Iceberg (or vice versa), then use the column in partition list.

The only concern is how to map the expression from Flink to Iceberg. We can use user defined function (UDF) to resolve it.

> The built in functions in Flink may not be enough to cover all partition transform in Iceberg.

Iceberg can provide all corresponding UDFs for all partition transforms from catalog.

For creating table, if a computed column is used in partition list and can be mapped as a partition transform, then Iceberg will save it as a partition field with the following rule:

- the column name will be used as the partition field name;
  
- the parameter column name will be used as the source field name;
  

> other kinds of computed columns are still not supported.

For loading table, a partition field will be mapped as a computed column and one element in partition list with the following rule:

- the partition field name will be used as the column name;
  
- the transform will be mapped as a UDF function invoking expression for the column expression;
  
- the type of partition field will be mapped as the type of the column.

kbendick:
Question wrt reading data:

Normal partition resolution rules will still apply, so nothing special wiill need to happen there from the users perspective?

If so, then something like https://github.com/apache/iceberg/issues/5000 is something I've been interested in for a long time.

Having to name the partition fields is possibly somewhat a pain though, as ideally users use the standard naming. However, there's no oblligation to use the standard naming (as partition fields can be renamed from their normal source + _ transform naming.

But because parttition fields can already be renamed, I think we can map them via the `PartitionSchema` and `PartitionField` itself (and not need the UDF). Maybe I'm mistaken though.

But is there a way we can do this without having to supply the given name? Is there a function catalog or something that can be applied in Flink SQL?

kbendick:
I do overall think this feature is important.

Can you provide a user side example of your idea @yittg?  I want to make sure I understand it fully.

yittg:
> Can you provide a user side example of your idea @yittg? I want to make sure I understand it fully.
This feature is not required for my personal scenario. In our platform, the metadata is not managed by Flink SQL, so I have no example about this though. 

This issue was primarily intended to initiate a discussion with some preliminary idea.

[...]

PR #4246, opened 2022-03-01
Flink: Add support for ResolvedSchema

hililiwei:
Some tasks in our production environment have been migrated to Schema/ResolvedSchema. This PR is mainly used to add support for ResolvedSchema.
1. Add the following `project` to the source:
```java
public Builder project(ResolvedSchema schema) {
  this.projectedResolvedSchema = schema;
  return this;
}
```
2. Add the following `tableSchema` to the sink:
```java
public Builder tableSchema(ResolvedSchema newTableSchema) {
  this.resolvedTableSchema = newTableSchema;
  return this;
}
```
3. FlinkDynamicTableFactory replaces CatalogTable with ResolvedCatalogTable and uses it to get `ResolvedSchema`, then pass `ResolvedSchema` in `IcebergTableSource`and `IcebergTableSink`. 
4. Add a tool class for ResolvedSchema: FlinkResolvedSchemaUtil. Convert the flink table schema to apache iceberg schema.
5. Add simple test cases.
https://github.com/apache/iceberg/blob/a08dff6a780f9d6154564f70c7d660f091dd5ec3/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/sink/TestFlinkIcebergSink.java#L203
https://github.com/apache/iceberg/blob/a08dff6a780f9d6154564f70c7d660f091dd5ec3/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkInputFormat.java#L132

PR #3681, opened 2021-12-07
Add watermark in flink table

hameizi:
If we add watermark in table we can deal data by flink with eventtime, and then we can count stream window with watermark and join with 'version table'(can works in combination https://github.com/apache/iceberg/pull/3095) what flink define.
In this PR the information of watermark is stored in table properties so that it will not affect original scheam of iceberg.
Below properties will be added in watermark table:
'watermark.0.rowtime'='testTime', 
'watermark.0.strategy.data-type'='TIMESTAMP(3)', 
'watermark.0.strategy.expr'='`testTime` - INTERVAL \'5\' SECOND'

After this PR we can create table set watermark like below sql:
 CREATE TABLE catalog.database.table(
    id BIGINT COMMENT 'unique id',
    data STRING, 
    testTime TIMESTAMP(3),
    PRIMARY KEY(id) NOT ENFORCED,
    WATERMARK FOR testTime AS testTime - INTERVAL '5' SECOND
) 
with ('format-version' = '2'
);

PR #2265, opened 2021-02-24
Flink : add computed column support for flink

zhangjun0x01:
related to #2262

Use computed column has many advantages , flink had supported the function , We can refer to [the design document of flink ](https://cwiki.apache.org/confluence/display/FLINK/FLIP-70%3A+Flink+SQL+Computed+Column+Design) ,I don't know whether spark supports this feature 。

- Virtual generated columns can be used as a way to simplify and unify queries. A complicated condition can be defined as a generated column and referred to from multiple queries on the table to ensure that all of them use exactly the same condition.

- Stored generated columns can be used as a materialized cache for complicated conditions that are costly to calculate on the fly.

- Generated columns can simulate functional indexes: Use a generated column to define a functional expression and index it. This can be useful for working with columns of types that cannot be indexed directly, such as JSON columns.

I add the support for iceberg.The main modifications include：

1.  add computed column  expression for iceberg field and  stored the expression in iceberg  metadata. The expression is column-level properties, similar to field doc.
2. when we write data to iceberg table ,the computed column is totally ignored.
3. When we read the iceberg table, the flink engine will automatically generate the value of the computed column for us, and then return it to the user together with the value of the physical column.

@rdblue @stevenzwu @openinx could you help  me review it ,thanks.

PR #4625, opened 2022-04-25
Flink: support watermark and computed columns

wuwenchi:
We hope iceberg can support watermark and computed columns.
The specific implementation details are as follows:

# Background and Motivation

There is a temporal table in iceberg, which needs to be used for real-time join with other tables, so a watermark needs to be defined. However, the watermark has requirements for the attributes of the fields, and the fields representing time in the source table may not meet this requirement, so it is necessary to convert the time fields of the source table by recalculating or calling functions.

**Why not directly use the iceberg connector table supported by flink?**

1. All flink tables are stored in memory. When the task ends, the table disappears. If you want to use it next time, you need to create it again.
2. In the scenes we use, a temporal table needs to be joined with multiple tables, and the attributes of the watermark are only related to the temporal table, so the watermark attributes of these tasks are consistent, so we hope to directly associate these attributes with Temporal table binding, when you use this table again in the future, you don't need to go to other tasks to check the watermark attribute setting parameters of the table.

# Goal

1. Iceberg supports flink-sql to set watermark and computed columns.
2. The watermark attribute and computed column attribute can be modified by modifying the table properties.

# Proposal

### Example

```sql
CREATE TABLE tl (
  id INT, 
  id2 AS id * 2, 
  f1 AS TO_TIMESTAMP(FROM_UNIXTIME(id*3)),
  t1 TIMESTAMP(6), 
  t2 AS cast(t1 AS TIMESTAMP(3)), 
  watermark FOR t2 AS t2 - INTERVAL '5' SECOND
);
```

Then the table properties are saved as:

```sql
flink.computed-columns.id2 = `id` * 2
flink.computed-columns.f1 = TO_TIMESTAMP(FROM_UNIXTIME(`id` * 3))
flink.computed-columns.t2 = CAST(`t1` AS TIMESTAMP(3))
flink.watermark.t2 = `t2` - INTERVAL '5' SECOND
```

### key format

fixed prefix + field name:

- fixed prefix for watermark: `flink.watermark.`
- fixed prefix  for computed columns: `flink.computed-columns.`

### value format

defined expression from user.

## The way of addition, deletion and modification

1、DDL of flink-sql, only supports adding

```sql
CREATE TABLE `hive_catalog`.`default`.`sample` (
  id INT, 
  id2 AS id * 2, 
  f1 AS TO_TIMESTAMP(FROM_UNIXTIME(id*3)),
  t1 TIMESTAMP(6), 
  t2 AS cast(t1 AS TIMESTAMP(3)), 
  watermark FOR t2 AS t2 - INTERVAL '5' SECOND
);
```

2、Syntax of table properties

- add or update

```sql
ALTER TABLE `hive_catalog`.`default`.`sample` SET (
	'flink.computed-columns.id2'='id*3'
)
```

- delete

```sql
ALTER TABLE `hive_catalog`.`default`.`sample` RESET (
	'flink.computed-columns.id2'
)
```

### add table process

1. If there is a defined computed column in the table, save its expression to the table property.
2. If there is a defined watermark in the table, save its expression to the table property.

### alter table properties process

1. Merge the modified properties with the original properties to get the updated properties of the table.

2. Generate the table of flink from the merged table properties and the schema of the original table, and verify the schema. (To prevent errors in expression, such as writing a non-existing function or column name in the expression of the computed column.)

3. If the verification in the previous step is successful, submit this property modification, 
   if not, report exception and do not make the final submission.

### get table process

1. Generate the flink table by combining the current table properties with the table schema, and verify the schema. (To prevent the schema of the table from being modified by other engines, resulting in an error in the expression of the calculated column. For example, in the expression of the computed column: `id` * 3, but then the `id` column was deleted using spark, and the corresponding property for computed column was not deleted.)

2. If the verification is successful, the table is returned. 
   If it is unsuccessful, the computed columns and watermarks in the table properties are ignored, and the original table physical schema is returned directly.

### DIFF
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
@@ -378,7 +378,7 @@ public class FlinkCatalog extends AbstractCatalog {
       throws CatalogException, TableAlreadyExistException {
     validateFlinkTable(table);
 
-    Schema icebergSchema = FlinkSchemaUtil.convert(table.getSchema());
+    Schema icebergSchema = FlinkSchemaUtil.convert(((ResolvedCatalogTable) table).getResolvedSchema());
     PartitionSpec spec = toPartitionSpec(((CatalogTable) table).getPartitionKeys(), icebergSchema);
 
     ImmutableMap.Builder<String, String> properties = ImmutableMap.builder();
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
@@ -37,8 +37,9 @@ import org.apache.flink.table.catalog.CatalogFunction;
 import org.apache.flink.table.catalog.CatalogPartition;
 import org.apache.flink.table.catalog.CatalogPartitionSpec;
 import org.apache.flink.table.catalog.CatalogTable;
-import org.apache.flink.table.catalog.CatalogTableImpl;
 import org.apache.flink.table.catalog.ObjectPath;
+import org.apache.flink.table.catalog.ResolvedCatalogTable;
+import org.apache.flink.table.catalog.ResolvedSchema;
 import org.apache.flink.table.catalog.exceptions.CatalogException;
 import org.apache.flink.table.catalog.exceptions.DatabaseAlreadyExistException;
 import org.apache.flink.table.catalog.exceptions.DatabaseNotEmptyException;
@@ -69,7 +70,6 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.NamespaceNotEmptyException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
-import org.apache.iceberg.flink.util.FlinkCompatibilityUtil;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
@@ -77,6 +77,8 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A Flink Catalog implementation that wraps an Iceberg {@link Catalog}.
@@ -89,6 +91,8 @@ import org.apache.iceberg.relocated.com.google.common.collect.Sets;
  * partition of Flink.
  */
 public class FlinkCatalog extends AbstractCatalog {
+
+  private static final Logger LOG = LoggerFactory.getLogger(FlinkCatalog.class);
 
   private final CatalogLoader catalogLoader;
   private final Catalog icebergCatalog;
@@ -387,6 +391,11 @@ public class FlinkCatalog extends AbstractCatalog {
       }
     }
 
+    // update properties from schema
+    Map<String, String> updatedProperties = FlinkSchemaUtil.generateTablePropertiesFromResolvedSchema(
+            ((ResolvedCatalogTable) table).getResolvedSchema());
+    properties.putAll(updatedProperties);
+
     try {
       icebergCatalog.createTable(
           toIdentifier(tablePath),
@@ -398,6 +407,19 @@ public class FlinkCatalog extends AbstractCatalog {
       if (!ignoreIfExists) {
         throw new TableAlreadyExistException(getName(), tablePath, e);
       }
+    }
+  }
+
+  private static void validateSchemaAndPartition(CatalogTable ct1, CatalogTable ct2) {
+    org.apache.flink.table.api.Schema ts1 = ct1.getUnresolvedSchema();
+    org.apache.flink.table.api.Schema ts2 = ct2.getUnresolvedSchema();
+
+    if (!Objects.equals(ts1, ts2)) {
+      throw new UnsupportedOperationException("Altering schema is not supported yet.");
+    }
+
+    if (!ct1.getPartitionKeys().equals(ct2.getPartitionKeys())) {
+      throw new UnsupportedOperationException("Altering partition keys is not supported yet.");
     }
   }
 
@@ -447,7 +469,7 @@ public class FlinkCatalog extends AbstractCatalog {
 
     // For current Flink Catalog API, support for adding/removing/renaming columns cannot be done by comparing
     // CatalogTable instances, unless the Flink schema contains Iceberg column IDs.
-    validateTableSchemaAndPartition(table, (CatalogTable) newTable);
+    validateSchemaAndPartition(table, (CatalogTable) newTable);
 
     Map<String, String> oldProperties = table.getOptions();
     Map<String, String> setProperties = Maps.newHashMap();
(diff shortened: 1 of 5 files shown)


wuwenchi:
relevant pr: #2265, #3381  
Could you help review it? Thanks!   @openinx  @rdblue  @hameizi   @zhangjun0x01  @stevenzwu

wuwenchi:
@kbendick   Thanks for your reply!

> I believe Flink 1.15 was released, so I am going to rebase my Flink 1.15 support PR to the release branch and we should have 1.15 support in the next few days.

Is this feature already supported on the flink 1.15 branch you are working on now? Or should we wait until flink1.15 is branched out and then support this function through this PR?

> that use partition transforms other than identity transform

I don't quite understand this... Do you mean that the specified partition columns are not within the identity columns range?

Zhangg7723:
nice job

wuwenchi:
@kbendick  Can we consider this way to implement partition transforms #4251 @yittg ?

stevenzwu:
I have a uber question regarding storing those properties into Iceberg table properties.

Watermark config is very job specific. Different Flink jobs may have different watermark assignment.
```
flink.watermark.t2 = `t2` - INTERVAL '5' SECOND
```

There are some discussions in PR #3681

wuwenchi:
@stevenzwu Thanks for your reply!

We have also considered this issue. 
But in fact, when our users are using, more scenarios are that a table needs to be joined with many other tables, and each association may be a separate task, so the watermak attributes of these tasks are actually the same.
If we don't have this feature, users have to look at the watermark property of other tasks to create a new task.

So my thoughts are:   For the same table
1. If you need different watermark attributes, you can use the flink connector to achieve it, create a connector for each task, and then specify different watermarks.
2. If you need the same watermark attributes, you can use the iceberg table directly.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkSchemaUtil.java
```diff
@@ -52,35 +63,69 @@
  */
 public class FlinkSchemaUtil {
 
+  public static final String FLINK_PREFIX = "flink.";
+
+  public static final String COMPUTED_COLUMNS = "computed-columns.";
```
hililiwei, line 68: use `computed-column` ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkSchemaUtil.java
```diff
@@ -172,4 +217,110 @@ public static TableSchema toSchema(Schema schema) {
...
+   *
+   * @param table flink unresolved schema to convert.
+   * @return Flink ResolvedSchema.
+   */
+  public static ResolvedSchema convertToResolvedSchema(CatalogTable table) {
+    StreamExecutionEnvironment env =  StreamExecutionEnvironment.getExecutionEnvironment(new Configuration());
+    StreamTableEnvironment streamTableEnvironment = StreamTableEnvironment.create(env);
+    CatalogManager catalogManager = ((TableEnvironmentImpl) streamTableEnvironment).getCatalogManager();
+    SchemaResolver schemaResolver = catalogManager.getSchemaResolver();
+    return table.getUnresolvedSchema().resolve(schemaResolver);
```
hililiwei, line 278:
```suggestion
    Configuration configuration = ExecutionEnvironment.getExecutionEnvironment().getConfiguration();
    TableEnvironment tableEnvironment = TableEnvironment.create(configuration);
    Preconditions.checkArgument(table.getDescription().isPresent(), "Illegal table.");
    return tableEnvironment.from(table.getDescription().get()).getResolvedSchema();
```

Based on your comment(https://github.com/apache/iceberg/pull/4246#issuecomment-1110614217), I took a look at our previous proposal and made some suggestions, but it may not be optimal.
Here, I think using 'TableEnvironment' directly is a better choice. It has better commonality, whether streaming, batch, or otherwise, and hides underlying differences.
Let's see if somebody else has a better solution.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/FlinkTestBase.java
```diff
@@ -78,7 +79,9 @@ protected TableEnvironment getTableEnv() {
               .build();
 
           TableEnvironment env = TableEnvironment.create(settings);
-          env.getConfig().getConfiguration().set(FlinkConfigOptions.TABLE_EXEC_ICEBERG_INFER_SOURCE_PARALLELISM, false);
+          env.getConfig().getConfiguration()
+              .set(FlinkConfigOptions.TABLE_EXEC_ICEBERG_INFER_SOURCE_PARALLELISM, false)
+              .set(TableConfigOptions.LOCAL_TIME_ZONE, "UTC");
```
hililiwei, line 84: good

hililiwei (commented):
Great job. This feature is useful for a lot of work, and I think we should put some thought into 'Schema' and 'ResoveSchema'. I look forward to more feature support after we finish discussion on 'ResoveSchema' and etc.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -374,7 +378,7 @@ void createIcebergTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
       throws CatalogException, TableAlreadyExistException {
     validateFlinkTable(table);
 
-    Schema icebergSchema = FlinkSchemaUtil.convert(table.getSchema());
+    Schema icebergSchema = FlinkSchemaUtil.convert(((ResolvedCatalogTable) table).getResolvedSchema());
```
stevenzwu, line 381: is it always safe to type cast to `ResolvedCatalogTable`?

Zhangg7723:
nice pr

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkSchemaUtil.java
wuwenchi: This method is indeed better!
But the description is taken from the table's comment:
```
    @Override
    public Optional<String> getDescription() {
        return Optional.of(getComment());
    }
```

The comment does not necessarily exist, and if it exists, it is not necessarily the path of the table, so `tableEnvironment.from` may not be able to get the table...

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkSchemaUtil.java
wuwenchi: yes, 's' is redundant, I will fix it.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: yes, in this interface, we always get `ResolvedCatalogTable`

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkSchemaUtil.java
hililiwei: Yes, so if we go this way, well, we might need to add a comment here.
https://github.com/apache/iceberg/blob/1300b8ac758e7644861388ac1e90f9ec9bd7e353/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java#L610-L614

wuwenchi: The `tableEnvironment` here is a newly created environment with only default_catalog and default_database, but this table actually exists in the catalog in the flink environment, so I think `tableEnvironment.from` should not find the required table. I also actually tested it and couldn't find it.

hililiwei: It's a bit tricky. I'll debug it locally to see if there's a better way.

kbendick (commented):
Thanks @wuwenchi for working on this!

I too would like to see watermark support and computed column support, particularly so that we can have partition transforms.

I have some concern, like @stevenzwu mentioned, about using table properties for things that might be very job-specific.

For example, I know that @chenjunjiedada has recommended that we don’t use the `write.upsert-enabled` config and instead set that per job. While we can’t immediately deprecate that configuration (and may never), I agree with his advice and I’d be hesitant to add more table properties that should be job-level properties.

My other concern is using `flink.` in table property names. While it might be the case today that these features are only supported by Flink (such as `write.upsert-enabled` configuration has been), we should ideally name things without regard for any specific engine as hopefully those engines will support such features in the future (or we can help add support for them).

Overall, I do really want to commend you on a job well done for this. I think it likely needs some changes, and might benefit from some kind of design discussion first, but truly thank you for making this PR as I’m sure much of this code will likely be useful for however this winds up getting designed (be it this way or another). And if this code is used it should definitely be credited.

I’m interested in exploring some of @yittg’s ideas from https://github.com/apache/iceberg/issues/4251  and combining some of them with your ideas, such as this updated idea of yours here  https://github.com/apache/iceberg/issues/5000

hililiwei:
As for computed columns, we try to support it in Iceberg itself, so that not only Flink, but other engines can also use it, but this change is quite large, just for reference https://github.com/apache/iceberg/pull/4994.

Issue #4604, opened 2022-04-21
Orc : Bug when adding a inner struct field as partition field 


tprelle:
When you you add a inner struct field as partition field without any transformation on an Iceberg orc table, you can not select anymore only this field because the as is a partition, iceberg do not read it from the file as it's inside a struct OrcValueReaders it's not able to reconstruct a constant struct.

It's working when you select also another field from the struct because as OrcValueReaders is able to read from the file using the another field reader

```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:659)
	at java.util.ArrayList.get(ArrayList.java:435)
	at org.apache.iceberg.orc.OrcValueReaders$StructReader.<init>(OrcValueReaders.java:161)
	at org.apache.iceberg.spark.data.SparkOrcValueReaders$StructReader.<init>(SparkOrcValueReaders.java:143)
	at org.apache.iceberg.spark.data.SparkOrcValueReaders.struct(SparkOrcValueReaders.java:70)
	at org.apache.iceberg.spark.data.SparkOrcReader$ReadBuilder.record(SparkOrcReader.java:75)
	at org.apache.iceberg.spark.data.SparkOrcReader$ReadBuilder.record(SparkOrcReader.java:65)
	at org.apache.iceberg.orc.OrcSchemaWithTypeVisitor.visitRecord(OrcSchemaWithTypeVisitor.java:71)
	at org.apache.iceberg.orc.OrcSchemaWithTypeVisitor.visit(OrcSchemaWithTypeVisitor.java:38)
	at org.apache.iceberg.orc.OrcSchemaWithTypeVisitor.visit(OrcSchemaWithTypeVisitor.java:32)
	at org.apache.iceberg.spark.data.SparkOrcReader.<init>(SparkOrcReader.java:52)
	at org.apache.iceberg.spark.source.RowDataReader.lambda$newOrcIterable$2(RowDataReader.java:163)
	at org.apache.iceberg.orc.OrcIterable.iterator(OrcIterable.java:108)
	at org.apache.iceberg.orc.OrcIterable.iterator(OrcIterable.java:45)
	at org.apache.iceberg.util.Filter.lambda$filter$0(Filter.java:35)
	at org.apache.iceberg.io.CloseableIterable$2.iterator(CloseableIterable.java:73)
	at org.apache.iceberg.spark.source.RowDataReader.open(RowDataReader.java:77)
[...]

PR #4627, opened 2022-04-25
Parquet: Fixes get null values for the nested field partition column

### LINKED ISSUE #4626: Get null values for for the nested field partition column
Code to reproduce:
```scala
scala> case class Data(id: Long, name: String)
scala> case class Item(id: Long, data: Data)

scala> spark.sql("create table local.db.test_table(id long, data struct<id: long, name: string>) using iceberg partitioned by (identity(data.name))").show()
scala> spark.range(5).map(i => Item(i, Data(i, s"${i}"))).writeTo("local.db.test_table").append()

scala> spark.sql("select data.name from local.db.test_table").show()
+----+
|name|
+----+
|null|
|null|
|null|
|null|
|null|
+----+
```

ConeyLiu:
We use `ConstantReader` for the partition column, and the `ConstantReader` field `column` is `NullReader.NULL_COLUMN`. When the `ConstantReader` or it's parent(when the parent has the only the constant children) wrapped into `OptionReader`, the `OptionReader` will always return null values because the following code:
```java
@Override
public T read(T reuse) {
  if (column.currentDefinitionLevel() > definitionLevel) {  // the `ConstantReader.currentDefinitionLevel` is always 0
    return reader.read(reuse);
  }

  for (TripleIterator<?> child : children) {
    child.nextNull();
  }

  return null;
}
```

Closes #4626

### DIFF
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
@@ -92,6 +92,7 @@ public class FlinkParquetReaders {
       // match the expected struct's order
       Map<Integer, ParquetValueReader<?>> readersById = Maps.newHashMap();
       Map<Integer, Type> typesById = Maps.newHashMap();
+      Map<Integer, Integer> maxDefinitionLevelsById = Maps.newHashMap();
       List<Type> fields = struct.getFields();
       for (int i = 0; i < fields.size(); i += 1) {
         Type fieldType = fields.get(i);
@@ -101,6 +102,9 @@ public class FlinkParquetReaders {
             int id = fieldType.getId().intValue();
             readersById.put(id, ParquetValueReaders.option(fieldType, fieldD, fieldReaders.get(i)));
             typesById.put(id, fieldType);
+            if (idToConstant.containsKey(id)) {
+              maxDefinitionLevelsById.put(id, fieldD);
+            }
           }
         }
       }
@@ -110,15 +114,14 @@ public class FlinkParquetReaders {
       List<ParquetValueReader<?>> reorderedFields =
           Lists.newArrayListWithExpectedSize(expectedFields.size());
       List<Type> types = Lists.newArrayListWithExpectedSize(expectedFields.size());
+      // Inferring MaxDefinitionLevel from parent field
+      int inferredMaxDefinitionLevel = type.getMaxDefinitionLevel(currentPath());
       for (Types.NestedField field : expectedFields) {
         int id = field.fieldId();
         if (idToConstant.containsKey(id)) {
           // containsKey is used because the constant may be null
-
-          // We use the max definition level of the parent node to infer the max definition level of the constant field
-          // in case of we could not find the given parquet field with typesById.
-          int fieldD = type.getMaxDefinitionLevel(currentPath());
-          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id), fieldD));
+          int fieldMaxDefinitionLevel = maxDefinitionLevelsById.getOrDefault(id, inferredMaxDefinitionLevel);
+          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id), fieldMaxDefinitionLevel));
           types.add(null);
         } else if (id == MetadataColumns.ROW_POSITION.fieldId()) {
           reorderedFields.add(ParquetValueReaders.position());
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
@@ -114,7 +114,11 @@ public class FlinkParquetReaders {
         int id = field.fieldId();
         if (idToConstant.containsKey(id)) {
           // containsKey is used because the constant may be null
-          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id)));
+
+          // We use the max definition level of the parent node to infer the max definition level of the constant field
+          // in case of we could not find the given parquet field with typesById.
+          int fieldD = type.getMaxDefinitionLevel(currentPath());
+          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id), fieldD));
           types.add(null);
         } else if (id == MetadataColumns.ROW_POSITION.fieldId()) {
           reorderedFields.add(ParquetValueReaders.position());
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
@@ -92,7 +92,6 @@ public class FlinkParquetReaders {
       // match the expected struct's order
       Map<Integer, ParquetValueReader<?>> readersById = Maps.newHashMap();
       Map<Integer, Type> typesById = Maps.newHashMap();
-      Map<Integer, Integer> maxDefinitionLevelsById = Maps.newHashMap();
       List<Type> fields = struct.getFields();
       for (int i = 0; i < fields.size(); i += 1) {
         Type fieldType = fields.get(i);
@@ -102,9 +101,6 @@ public class FlinkParquetReaders {
             int id = fieldType.getId().intValue();
             readersById.put(id, ParquetValueReaders.option(fieldType, fieldD, fieldReaders.get(i)));
             typesById.put(id, fieldType);
-            if (idToConstant.containsKey(id)) {
-              maxDefinitionLevelsById.put(id, fieldD);
-            }
           }
         }
       }
@@ -114,14 +110,11 @@ public class FlinkParquetReaders {
       List<ParquetValueReader<?>> reorderedFields =
           Lists.newArrayListWithExpectedSize(expectedFields.size());
       List<Type> types = Lists.newArrayListWithExpectedSize(expectedFields.size());
-      // Inferring MaxDefinitionLevel from parent field
-      int inferredMaxDefinitionLevel = type.getMaxDefinitionLevel(currentPath());
       for (Types.NestedField field : expectedFields) {
         int id = field.fieldId();
         if (idToConstant.containsKey(id)) {
           // containsKey is used because the constant may be null
-          int fieldMaxDefinitionLevel = maxDefinitionLevelsById.getOrDefault(id, inferredMaxDefinitionLevel);
-          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id), fieldMaxDefinitionLevel));
+          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id)));
           types.add(null);
         } else if (id == MetadataColumns.ROW_POSITION.fieldId()) {
           reorderedFields.add(ParquetValueReaders.position());
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkInputFormat.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkInputFormat.java
@@ -21,15 +21,12 @@ package org.apache.iceberg.flink.source;
 import static org.apache.iceberg.types.Types.NestedField.required;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import org.apache.flink.table.api.DataTypes;
 import org.apache.flink.table.api.TableSchema;
 import org.apache.flink.table.types.logical.RowType;
 import org.apache.flink.types.Row;
-import org.apache.iceberg.FileFormat;
-import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.catalog.TableIdentifier;
@@ -37,11 +34,9 @@ import org.apache.iceberg.data.GenericAppenderHelper;
 import org.apache.iceberg.data.RandomGenericData;
 import org.apache.iceberg.data.Record;
 import org.apache.iceberg.flink.FlinkSchemaUtil;
-import org.apache.iceberg.flink.TestFixtures;
 import org.apache.iceberg.flink.TestHelpers;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.types.Types;
-import org.junit.Assume;
 import org.junit.Test;
 
 /** Test {@link FlinkInputFormat}. */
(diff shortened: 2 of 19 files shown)


flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkInputFormat.java
```diff
@@ -126,6 +131,48 @@ public void testBasicProjection() throws IOException {
     TestHelpers.assertRows(result, expected);
   }
 
+  @Test
+  public void testReadPartitionColumn() throws Exception {
+    Assume.assumeTrue("Temporary skip ORC", FileFormat.ORC != fileFormat);
```
ConeyLiu, line 136: Temporary skip ORC test because #4599 is working on fixing it.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestPartitionValues.java
```diff
@@ -432,4 +436,47 @@ public void testPartitionedByNestedString() throws Exception {
 
     Assert.assertEquals("Number of rows should match", rows.size(), actual.size());
   }
+
+  @Test
+  public void testReadPartitionColumn() throws Exception {
+    Assume.assumeTrue("Temporary skip ORC", !"orc".equals(format));
```
ConeyLiu, line 442: Same as here too, temporary skip ORC test because https://github.com/apache/iceberg/pull/4599 is working on fixing it.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
```diff
@@ -116,7 +116,11 @@ public ParquetValueReader<RowData> struct(Types.StructType expected, GroupType s
         int id = field.fieldId();
         if (idToConstant.containsKey(id)) {
           // containsKey is used because the constant may be null
-          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id)));
+
+          // We use the max definition level of the parent node to infer the max definition level of the constant field
```
kbendick, line 120: Nit: typically we avoid the use of pronouns (we, etc) in comments.

Also, does it make sense to first get the result of `idToConstant` and then check if the max definition is needed? This code is definitely in the "hot path" for Flink, so unnecessary looks ups should be avoided if at all possible imo.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/StructRecord.java
```diff
@@ -0,0 +1,70 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.source;
+
+import org.apache.iceberg.relocated.com.google.common.base.Objects;
+
+public class StructRecord {
```
kbendick, line 24: Nit: As these two classes (StructRecord and InnerRecord) seem to only be used as java beans during tests, could you possibly name them something a bit more specific / something that makes it more clear that these are for test purposes?

The file `StructRecord` on its own really seems like it is part of the framework and could easily be confusing for people (e.g. it was confusing for me).

Some of the Spark tests use `LogRecord`, as well as some use `SimpleRecord`. Admittedly `SimpleRecord` might have the same naming implications, but `StructRecord` is so close to existing concepts and classes like `StructReader` for me.

Maybe `TwoLevelRecord` or something might be more appropriate? I can see how `StructRecord` seems similar to `SimpleRecord` (though I personally wouldn't have used `SimpleRecord` either), but `StructRecord` is _super_ similar in name to existing Iceberg concepts.

Even just `ComplexRecord` or `NestedRecord` or even `TestStructRecord` would be a lot better in my opinion.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/InnerRecord.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    InnerRecord that = (InnerRecord) o;
+    return innerId == that.innerId && Objects.equal(innerName, that.innerName);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(innerId, innerName);
+  }
```
kbendick, line 69: The other beans that are used during testing override the `toString` method, which winds up being really helpful when tests fail.

Could you please implement `toString` for this class and the other test data class? You can use `MoreObjects.toStringHelper` or just do it manually.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
```diff
@@ -116,7 +116,11 @@ public ParquetValueReader<RowData> struct(Types.StructType expected, GroupType s
...
+          // in case of we could not find the given parquet field with typesById.
+          int fieldD = type.getMaxDefinitionLevel(currentPath());
```
kbendick, line 122: nit: potential typo? `fieldD` reads oddly to me. Is that supposed to be fieldDepth or fieldId? I've been thinking it was `fieldId`, but now I'm not sure.

If it is `fieldId`, my follow up question is whether or not that's the most appropriate name for it given that it's an inferred max definition level.

If it's something else and `D` is an acronym, could you please name the field something more specific?

kbendick, line 122: Based on the comment, we only need `fieldD` if the `typesById` function look up fails

But if we look at lines 101-107 of this file, it looks like very similar logic already exists. Would it be more appropriate to move this logic up closer to lines 101-107, where `readersById` and `typesById` are constructed and there's already a call to `type.getMaxDefinitionLevel`?  Perhaps in an `else` block, so that the `typesById` mapping is always complete?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
ConeyLiu: Updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/InnerRecord.java
ConeyLiu: Updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/StructRecord.java
ConeyLiu: Updated

ConeyLiu:
Thanks, @kbendick for the detailed code reviewing. Comments have been addressed. Please take another look when you are free.

kbendick (commented):
Hey @ConeyLiu! Thanks for the PR. Sorry for the delay in reviewing it, this seems important.

I tested this locally and you are absolutely correct, this is an issue.

Could you rebase this off of latest master @ConeyLiu? There have been changes in `TestFlinkInputFormat` class it seems (and we're on 1.15 now, technically, but that's not too urgent).

cc @rdblue

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
```diff
@@ -112,11 +116,14 @@ public ParquetValueReader<RowData> struct(Types.StructType expected, GroupType s
       List<ParquetValueReader<?>> reorderedFields = Lists.newArrayListWithExpectedSize(
           expectedFields.size());
       List<Type> types = Lists.newArrayListWithExpectedSize(expectedFields.size());
+      // Inferring MaxDefinitionLevel from parent field
```
kbendick, line 119: Nit: Does this comment still apply since we have added to the `maxDefinitionLevelsById` map above?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
```diff
@@ -112,11 +116,14 @@ public ParquetValueReader<RowData> struct(Types.StructType expected, GroupType s
...
+      int inferredMaxDefinitionLevel = type.getMaxDefinitionLevel(currentPath());
```
kbendick, line 120: Nit: Now that we're putting the field depth into the map `maxDefinitionLevelsById`, and we have the same check for `idToConstant.containsKey(id)`, do we need to have this fallback?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestPartitionValues.java
```diff
@@ -432,4 +436,48 @@ public void testPartitionedByNestedString() throws Exception {
...
+    table.updateProperties().set(TableProperties.DEFAULT_FILE_FORMAT, format).commit();
+
+    // write into iceberg
+    MapFunction<Long, ComplexRecord> func =
+        value -> new ComplexRecord(value, new NestedRecord(value, "name_" + value));
+    spark.range(0, 10, 1, 1).map(func, Encoders.bean(ComplexRecord.class))
+            .write()
+            .format("iceberg")
+            .mode(SaveMode.Append)
+            .save(baseLocation);
```
rdblue, line 468: Can you fix indentation for these lines?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
ConeyLiu: @kbendick, we could not update the `maxDefinitionLevelsById ` if the `fieldType.getId()` is null, you could see it at L101.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestPartitionValues.java
ConeyLiu: Updated

ConeyLiu:
Thanks, @kbendick @rdblue for the review. The comments have been addressed.

ConeyLiu:
Hi @kbendick, is there any more advice for this?

parquet/src/main/java/org/apache/iceberg/parquet/ParquetValueReaders.java
```diff
@@ -116,9 +120,45 @@ public void setPageSource(PageReadStore pageStore, long rowPosition) {
...
+          return false;
+        }
+
+        @Override
+        public Object next() {
+          return null;
+        }
+      };
+
+      this.children = ImmutableList.of(column);
```
szehon-ho, line 161: Trying to understand how this works, but this previously is empty list, but we need it here?

szehon-ho (commented):
Nice observation and fix.  Though Im not a parquet expert, maybe worth to have some others look at it as well, @RussellSpitzer @aokolnychyi ?

parquet/src/main/java/org/apache/iceberg/parquet/ParquetValueReaders.java
szehon-ho: Never mind, I guess its like that in other readers, bit confusing.

pvary:
@ConeyLiu: Thanks for the finding and the fix.
May I ask you to put the fix in the main branches for Spark (3.3) and Flink (1.16) first, and then with another PR we can backport to all of the relevant branches.

Do we have the same issue for ORC and Avro too?

parquet/src/main/java/org/apache/iceberg/data/parquet/BaseParquetReaders.java
```diff
@@ -149,11 +153,14 @@ public ParquetValueReader<?> struct(Types.StructType expected, GroupType struct,
...
           expectedFields.size());
       List<Type> types = Lists.newArrayListWithExpectedSize(expectedFields.size());
+      // Inferring MaxDefinitionLevel from parent field
+      int inferredMaxDefinitionLevel = type.getMaxDefinitionLevel(currentPath());
       for (Types.NestedField field : expectedFields) {
         int id = field.fieldId();
         if (idToConstant.containsKey(id)) {
           // containsKey is used because the constant may be null
-          reorderedFields.add(ParquetValueReaders.constant(idToConstant.get(id)));
+          int fieldMaxDefinitionLevel = maxDefinitionLevelsById.getOrDefault(id, inferredMaxDefinitionLevel);
```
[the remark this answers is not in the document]
ConeyLiu: The max definition level of the current column path is: `type.getMaxDefinitionLevel(currentPath()) - 1`, the children should be `type.getMaxDefinitionLevel(currentPath())`. So the `fieldMaxDefinitionLevel ` is the inferred children max definition level, I have updated the comments. And it is used only when we could not find the value from `maxDefinitionLevelsById`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetValueReaders.java
ConeyLiu: Yeah, this is just keeping the same as the original code.

ConeyLiu:
Thanks @szehon-ho @pvary for the review.

> May I ask you to put the fix in the main branches for Spark (3.3) and Flink (1.16) first, and then with another PR we can backport to all of the relevant branches.

Updated to Spark 3.3 and Flink 1.16, the Pig part is kept.

> Do we have the same issue for ORC and Avro too?

Avro is OK, while ORC has a similar problem at here https://github.com/apache/iceberg/issues/4604.

pvary (approved):

parquet/src/main/java/org/apache/iceberg/data/parquet/BaseParquetReaders.java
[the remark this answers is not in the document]
szehon-ho: Nit: I still think the name 'inferred' is a bit confusing, it indicates to me that it's the one that will be chosen.  Will it be better to call it 'defaultMaxDefintionLevel' or 'parentMaxDefinitionLevel'? 

Also just curious, what is the example of a case where we need this default value?  I think i tried to walk through an example and find the expected field is always in the actual struct?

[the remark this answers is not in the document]
ConeyLiu: Changed to `defaultMaxDefinitionLevel`

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/data/SparkParquetReaders.java
```diff
@@ -148,6 +149,9 @@ public ParquetValueReader<?> struct(
           int id = fieldType.getId().intValue();
```
ConeyLiu, line 149: @szehon-ho here the `fieldType.getId()` could be null, I guess this is for a compatible purpose. So I add the `int defaultMaxDefinitionLevel = type.getMaxDefinitionLevel(currentPath());` in the following code in case we get a null value from `maxDefinitionLevelsById`.

szehon-ho: Thanks!  Yea that's ok to me, just in my understanding I'm not entirely sure when it is ever null.  

Sorry would you be able to change the comment as well to "Defaulting to parent max definition level" or something like that?  Otherwise patch looks good to me

szehon-ho (approved):
It's a small nit so I'll just approve it

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/data/SparkParquetReaders.java
ConeyLiu: Updated

szehon-ho: Thanks, looks good to me.

ConeyLiu:
Thanks, @szehon-ho for merging this and all for the review. Will submit a backport PR to other spark/flink versions.

PR #3457, opened 2021-11-03
Spark: Improve performance of expire snapshot by not double-scanning retained Snapshots

szehon-ho:
Expire snapshots can take a long time for large tables with millions of files and thousands of snapshots/manifests.

One cause is the calculation of files to be deleted.  The current algorithm is:

-  Find the reachability graph of all snapshots before expiration
-  Find reachability graph of all snapshots after expiration
-  Subtract the second from the first, these are files to delete
 
But this explores every retained snapshot twice.  Example: any periodic expire-snapshot job that expires 1 snapshot needs to explore all n-1 snapshots twice.

Proposal:

- Find reachability graph of all snapshots after expiration
- Find reachability graph of expired snapshots (if only a few expired, should be much smaller set)
- Subtract the first from the second, these are files to delete

Implementation: For expired-snapshot scan, change the original spark query of metadata tables to custom spark jobs that only explore from expired snapshot(s).

Note: The new expired-snapshot scan duplicates manifestList scan logic to handle "write.manifest-lists.enabled"="false" flag, but unfortunately the functionality seems broken without this change and so not possible to test currently.  Added a test for demonstration purpose.

PR #2518, opened 2021-04-26
Spark: Remove deletefiles when expiring snapshots.

coolderli:
When expiring snapshots, the deletefiles should be deleted.

PR #2510, opened 2021-04-25
Core: Add content field to ManifestTable Schema.

coolderli:
We need to add the content field to the schema of AllManifestsTable, so that we can get the type of manifest for expiration processing.

PR #4629, opened 2022-04-25
Spark: Add positional and equality delete file count to ExpireSnapshot results

szehon-ho:
Currently both the ExpireSnapshot SparkAction and Procedure return 'deletedDataFileCount' but this includes delete files.

Now return deletedPositionalDeleteFiles and deletedEqualityDeleteFiles as separate values, and take them out of deletedDataFiles.


dramaticlly:
from what I can tell, all Spark share the similar interface, so looks like Spark3.2 is updated with new results but [Spark 2.4](https://github.com/apache/iceberg/blob/master/spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java#L274) is still unhappy per CI failure 

```
/home/runner/work/iceberg/iceberg/spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java:274: error: constructor BaseExpireSnapshotsActionResult in class org.apache.iceberg.actions.BaseExpireSnapshotsActionResult cannot be applied to given types;

    return new BaseExpireSnapshotsActionResult(dataFileCount.get(), manifestCount.get(), manifestListCount.get());
           ^
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -243,23 +246,28 @@ private BaseExpireSnapshotsActionResult deleteFiles(Iterator<Row> expired) {
...
-              break;
-            case MANIFEST:
-              manifestCount.incrementAndGet();
-              LOG.debug("Deleted Manifest: {}", file);
-              break;
-            case MANIFEST_LIST:
-              manifestListCount.incrementAndGet();
-              LOG.debug("Deleted Manifest List: {}", file);
-              break;
+          if (type.equalsIgnoreCase(FileContent.DATA.name())) {
```
dramaticlly, line 249: nit: would it be better to switch the position such as 
`FileContent.DATA.name().equalsIgnoreCase(type)`

to provide better guarantee if type is null?

core/src/main/java/org/apache/iceberg/actions/BaseExpireSnapshotsActionResult.java
```diff
@@ -22,13 +22,19 @@
 public class BaseExpireSnapshotsActionResult implements ExpireSnapshots.Result {
 
   private final long deletedDataFilesCount;
+  private final long deletedPosDeleteFilesCount;
+  private final long deletedEqDeleteFilesCount;
   private final long deletedManifestsCount;
   private final long deletedManifestListsCount;
 
   public BaseExpireSnapshotsActionResult(long deletedDataFilesCount,
+                                         long deletedPosDeleteFilesCount,
```
dramaticlly, line 31: It's also used in Spark 2.4 code https://github.com/apache/iceberg/blob/master/spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java#L274

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -243,23 +246,28 @@ private BaseExpireSnapshotsActionResult deleteFiles(Iterator<Row> expired) {
...
+            eqDeleteFileCount.incrementAndGet();
+            LOG.trace("Deleted Equality Delete File: {}", file);
+          } else if (type.equalsIgnoreCase(MANIFEST)) {
+            manifestCount.incrementAndGet();
+            LOG.debug("Deleted Manifest: {}", file);
+          } else if (type.equalsIgnoreCase(MANIFEST_LIST)) {
+            manifestListCount.incrementAndGet();
+            LOG.debug("Deleted Manifest List: {}", file);
+          } else {
+            throw new ValidationException("Illegal Type: %s", type);
```
aokolnychyi, line 265: nit: `Illegal file type` to be more precise?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -176,16 +184,55 @@ protected Dataset<Row> loadMetadataTable(Table table, MetadataTableType type) {
...
+
+    public String getType() {
+      return type;
+    }
+
+    public void setType(String type) {
+      this.type = type;
+    }
+
+    public static ContentFileTypeBean toContentFilePathBean(ContentFile contentFile) {
```
aokolnychyi, line 231: nit: `ContentFile<?>`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -176,16 +184,55 @@ protected Dataset<Row> loadMetadataTable(Table table, MetadataTableType type) {
...
 
     @Override
-    public Iterator<String> call(ManifestFileBean manifest) {
-      return new ClosingIterator<>(ManifestFiles.readPaths(manifest, io.getValue()).iterator());
+    public Iterator<ContentFileTypeBean> call(ManifestFileBean manifest) {
+      switch (manifest.content()) {
+        case DATA:
+          return CloseableIterator.transform(
+              ManifestFiles.read(manifest, table.getValue().io(), table.getValue().specs()).iterator(),
+              ContentFileTypeBean::toContentFilePathBean);
```
aokolnychyi, line 200: Question: can we use a wrapper instead of creating a bean for every entry? I am a little worried about the number of generated instances and GC.

Would something like this work?

```
private CloseableIterator<ContentFileTypeBean> entries(ManifestFileBean manifest) {
  ContentFileTypeBean contentFileType = new ContentFileTypeBean();

  switch (manifest.content()) {
    case DATA:
      return CloseableIterator.transform(
          ...,
          contentFileType::wrap);
    case DELETES:
      return CloseableIterator.transform(
          ...,
          contentFileType::wrap);
    default:
      throw new IllegalArgumentException("Unsupported manifest content type:" + manifest.content());
  }
}
```

```
  public static class ContentFileTypeBean implements Serializable {
    private String file;
    private String type;

    public ContentFileTypeBean wrap(ContentFile<?> contentFile) {
      this.file = contentFile.path().toString();
      this.type = contentFile.content().toString();
      return this;
    }
  ...
}
```

core/src/main/java/org/apache/iceberg/actions/BaseExpireSnapshotsActionResult.java
aokolnychyi: We don't have to implement the same functionality for older versions in this PR.
Maybe, just keep the old constructor for older Spark versions?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: Hm a little confused how this will work, In the end, won't the iterator.next() always the same instance with same file/type which is incorrect?

api/src/main/java/org/apache/iceberg/actions/ExpireSnapshots.java
```diff
@@ -101,6 +101,16 @@ interface Result {
      */
     long deletedDataFilesCount();
 
+    /**
+     * Returns the number of deleted equality delete files.
+     */
+    long deletedEqDeleteFileCount();
```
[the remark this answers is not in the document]
szehon-ho: Makes sense, changed.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/ExpireSnapshotsProcedure.java
```diff
@@ -52,6 +52,8 @@ public class ExpireSnapshotsProcedure extends BaseProcedure {
 
   private static final StructType OUTPUT_TYPE = new StructType(new StructField[]{
       new StructField("deleted_data_files_count", DataTypes.LongType, true, Metadata.empty()),
+      new StructField("deleted_positional_data_files_count", DataTypes.LongType, true, Metadata.empty()),
```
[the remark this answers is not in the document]
szehon-ho: Whoops, good catch

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -176,16 +184,55 @@ protected Dataset<Row> loadMetadataTable(Table table, MetadataTableType type) {
...
 
-    ReadManifest(Broadcast<FileIO> io) {
-      this.io = io;
+    ReadManifest(Broadcast<Table> table) {
+      this.table = table;
     }
 
     @Override
-    public Iterator<String> call(ManifestFileBean manifest) {
-      return new ClosingIterator<>(ManifestFiles.readPaths(manifest, io.getValue()).iterator());
```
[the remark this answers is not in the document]
szehon-ho: Good catch

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: I got rid of the Bean and replaced with Scala Tuple2, makes it a bit simpler to not have the boilerplate code, but still doesn't get rid of the new instance per row.  (Still not too comfortable to replace with one single wrapper instance).

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -122,18 +124,24 @@ protected Table newStaticTable(TableMetadata metadata, FileIO io) {
...
         .repartition(spark.sessionState().conf().numShufflePartitions()) // avoid adaptive execution combining tasks
         .as(Encoders.bean(ManifestFileBean.class));
 
-    return allManifests.flatMap(new ReadManifest(ioBroadcast), Encoders.STRING()).toDF(FILE_PATH);
+    return allManifests.flatMap(new ReadManifest(tableBroadcast), Encoders.bean(ContentFileTypeBean.class))
+        .toDF(FILE_PATH, FILE_TYPE);
+  }
+
+  // builds a DF of delete and data file paths by reading all manifests
+  protected Dataset<Row> buildValidContentFileDF(Table table) {
```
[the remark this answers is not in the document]
szehon-ho: It was to preserve the backward compatiblity of the other actions, which didn't care much about the type

core/src/main/java/org/apache/iceberg/actions/BaseExpireSnapshotsActionResult.java
szehon-ho: Kept old constructor for older Spark versions (deleteFileCounts initialized to 0)

szehon-ho: Kept old constructor for older Spark versions (deleteFileCounts initialized to 0)

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestExpireSnapshotsAction.java
```diff
@@ -967,7 +990,77 @@ public void testExpireOlderThanWithRollbackAndMergedManifests() {
...
 
-    checkExpirationResults(1, 1, 1, result);
+    checkExpirationResults(1, 0, 0, 1, 1, result);
+  }
+
+  @Test
+  public void testExpireOlderThanWithDeleteFile() {
+    table.updateProperties()
+        .set(TableProperties.FORMAT_VERSION, "2")
+        .set(TableProperties.MANIFEST_MERGE_ENABLED, "false")
```
rajarshisarkar, line 1000: [Question] Any specific reason to disable the manifest merge?

szehon-ho: Yea as this test checks the exact list of files removed for ExpireSnapshots, I thought it's simpler to get the exact list if we don't have to deal with manifests auto-merging after some time, as then, I just need to get manifests from the latest snapshot.  

Though in this case, I tried and it seems not necessary as the auto-merge does not kick in (default is set high).  I was thinking to keep it in case the algorithm changes and manifests start merging, failing this test, but I am open if it's better to remove it.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -176,16 +186,35 @@ protected Dataset<Row> loadMetadataTable(Table table, MetadataTableType type) {
...
     @Override
-    public Iterator<String> call(ManifestFileBean manifest) {
-      return new ClosingIterator<>(ManifestFiles.readPaths(manifest, io.getValue()).iterator());
+    public Iterator<Tuple2<String, String>> call(ManifestFileBean manifest) {
+      return new ClosingIterator(entries(manifest));
+    }
+
+    public CloseableIterator<Tuple2<String, String>> entries(ManifestFileBean manifest) {
+      switch (manifest.content()) {
+        case DATA:
```
zinking, line 203: isn't this always DATA  in current imiplementation?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -243,23 +246,30 @@ private BaseExpireSnapshotsActionResult deleteFiles(Iterator<Row> expired) {
...
-              manifestCount.incrementAndGet();
-              LOG.debug("Deleted Manifest: {}", file);
-              break;
-            case MANIFEST_LIST:
-              manifestListCount.incrementAndGet();
-              LOG.debug("Deleted Manifest List: {}", file);
-              break;
+
+          if (FileContent.DATA.name().equalsIgnoreCase(type)) {
+            dataFileCount.incrementAndGet();
```
zinking, line 251: there must be away to use those enums properly.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -176,16 +186,35 @@ protected Dataset<Row> loadMetadataTable(Table table, MetadataTableType type) {
...
+          return CloseableIterator.transform(
+              ManifestFiles.read(manifest, table.getValue().io(), table.getValue().specs()).iterator(),
```
zinking, line 205: this is not necessary, only path is needed.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestExpireSnapshotsAction.java
rajarshisarkar: I see, makes sense to me.

rajarshisarkar: I see, makes sense to me to keep the manifest merge as false here.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: What do you mean?  There are now manifest with two kind of ManifestContent, and both will get read by expireSnapshot scan to find the files to be deleted.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: I am not sure I got the comment, but this needs to get the FileContent of the file to get the total number in the end, right?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
zinking: ManifestFileBean.content is always returning DATA looking from src

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
zinking: in the other pr, it was readPath, did this method open the data file ?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: Good catch, this seems it is working by chance, ManifestFiles.read() only checks to make sure content=Data, but doesn't use the content anywhere else in this path.  The file returned can still be delete or data file.

Not sure if it's worth to implement manifests.content field (I guess that's what was done in the original pr: #2518).   if there is not much use now, maybe we can do it in a separately and here just add a new method to ManifestFiles.files() that returns a Pair of type/path, similar to ManifestFiles.paths().  @aokolnychyi any preference?

szehon-ho: Yea after thinking, another option is to get replace  this altogether with a query on 'all_files' table, which should be coming soon.  It's a cleaner way as the complexity is kept within the metadata table, though requires to wait until that one.

api/src/main/java/org/apache/iceberg/actions/ExpireSnapshots.java
```diff
@@ -101,6 +101,16 @@ interface Result {
...
 
+    /**
+     * Returns the number of deleted equality delete files.
+     */
+    long deletedEqualityDeleteFilesCount();
+
+    /**
+     * Returns the number of deleted positional delete files.
+     */
+    long deletedPositionalDeleteFilesCount();
```
aokolnychyi, line 112: I think we use `positionDeleteFiles`, not `positionalDeleteFiles`. Better be consistent.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
aokolnychyi: I am afraid we don't project `content` in manifests-related metadata tables but we should probably fix that. I discovered that in another use case. I can submit a PR for that.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/ExpireSnapshotsProcedure.java
```diff
@@ -52,6 +52,8 @@ public class ExpireSnapshotsProcedure extends BaseProcedure {
 
   private static final StructType OUTPUT_TYPE = new StructType(new StructField[]{
       new StructField("deleted_data_files_count", DataTypes.LongType, true, Metadata.empty()),
+      new StructField("deleted_positional_delete_files_count", DataTypes.LongType, true, Metadata.empty()),
```
aokolnychyi, line 55: This should also be `position` vs `positional`.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -243,23 +246,30 @@ private BaseExpireSnapshotsActionResult deleteFiles(Iterator<Row> expired) {
...
+            LOG.trace("Deleted Data File: {}", file);
+          } else if (FileContent.POSITION_DELETES.name().equalsIgnoreCase(type)) {
+            posDeleteFileCount.incrementAndGet();
+            LOG.trace("Deleted Positional Delete File: {}", file);
+          } else if (FileContent.EQUALITY_DELETES.name().equalsIgnoreCase(type)) {
+            eqDeleteFileCount.incrementAndGet();
+            LOG.trace("Deleted Equality Delete File: {}", file);
+          } else if (type.equalsIgnoreCase(MANIFEST)) {
```
aokolnychyi, line 259: nit: switch type and `MANIFEST` to follow the new pattern?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -243,23 +246,30 @@ private BaseExpireSnapshotsActionResult deleteFiles(Iterator<Row> expired) {
...
+            manifestCount.incrementAndGet();
+            LOG.debug("Deleted Manifest: {}", file);
+          } else if (type.equalsIgnoreCase(MANIFEST_LIST)) {
```
aokolnychyi, line 262: nit: same here

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
aokolnychyi: I think it is a mix of enums and other String constants. Seems fine to me.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: For reference, related pr : https://github.com/apache/iceberg/pull/2510

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: Done, thanks for suggestion

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
szehon-ho: Ref: performance, yea I realized that all_entries table is not parallel in reading manifest-lists while commenting on #3457, sorry forgot to update here.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
szehon-ho: whoops missed these.

szehon-ho:
@aokolnychyi resolved all the comments, save the one about not having access to field ManifestFile 'content'.  I mean this still works as-is (just the ManifestContent=DELETES path is never taken though still returning delete files), but I can wait until that one goes in, either way.

szehon-ho:
@aokolnychyi when you have a chance, this now correctly utilizes 'content' field of manifest during the computation of valid files.

.palantir/revapi.yml
```diff
@@ -27,6 +27,12 @@ acceptedBreaks:
     - code: "java.method.addedToInterface"
       new: "method ThisT org.apache.iceberg.SnapshotUpdate<ThisT>::scanManifestsWith(java.util.concurrent.ExecutorService)"
       justification: "Accept all changes prior to introducing API compatibility checks"
+    - code: "java.method.addedToInterface"
```
aokolnychyi, line 30: I am okay with this given other changes, upcoming 1.0 and that it is mostly Iceberg itself that provides implementations of this interface. Even if someone has custom actions, they probably reuse the provided result implementation. 

If other folks are concerned, we could default the new methods.

.palantir/revapi.yml
```diff
@@ -27,6 +27,12 @@ acceptedBreaks:
...
+      new: "method long org.apache.iceberg.actions.ExpireSnapshots.Result::deletedEqualityDeleteFilesCount()"
+      justification: "Adding method to interface"
```
aokolnychyi, line 32: nit: Shall we be a little bit more specific?

api/src/main/java/org/apache/iceberg/ManifestContent.java
```diff
@@ -35,4 +35,12 @@ public enum ManifestContent {
   public int id() {
     return id;
   }
+
+  public static ManifestContent fromId(int id) {
+    switch (id) {
```
aokolnychyi, line 40: nit: would it be safer to throw an exception if `id` is unknown just in case?

core/src/main/java/org/apache/iceberg/actions/BaseExpireSnapshotsActionResult.java
```diff
@@ -22,13 +22,29 @@
...
     this.deletedDataFilesCount = deletedDataFilesCount;
+    this.deletedPosDeleteFilesCount = 0;
+    this.deletedEqDeleteFilesCount = 0;
+    this.deletedManifestsCount = deletedManifestsCount;
+    this.deletedManifestListsCount = deletedManifestListsCount;
+  }
+
+  public BaseExpireSnapshotsActionResult(long deletedDataFilesCount,
+                                         long deletedPosDeleteFilesCount,
+                                         long deletedEqDeleteFileCount,
```
aokolnychyi, line 42: nit: `File` -> `Files`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -122,18 +128,29 @@ protected Table newStaticTable(TableMetadata metadata, FileIO io) {
     return new BaseTable(ops, metadataFileLocation);
   }
 
-  // builds a DF of delete and data file locations by reading all manifests
-  protected Dataset<Row> buildValidContentFileDF(Table table) {
+  // builds a DF of delete and data file path and type by reading all manifests
+  protected Dataset<Row> buildValidContentFileWithTypeDF(Table table) {
     JavaSparkContext context = JavaSparkContext.fromSparkContext(spark.sparkContext());
-    Broadcast<FileIO> ioBroadcast = context.broadcast(SparkUtil.serializableFileIO(table));
+    Broadcast<Table> tableBroadcast = context.broadcast(SerializableTable.copyOf(table));
```
aokolnychyi, line 134: Not directly related to this PR but should we just use `sparkContext` defined as the instance variable instead of creating a new wrapper in this method?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSparkAction.java
```diff
@@ -122,18 +128,29 @@ protected Table newStaticTable(TableMetadata metadata, FileIO io) {
...
 
     Dataset<ManifestFileBean> allManifests = loadMetadataTable(table, ALL_MANIFESTS)
-        .selectExpr("path", "length", "partition_spec_id as partitionSpecId", "added_snapshot_id as addedSnapshotId")
+        .selectExpr("content",
```
aokolnychyi, line 137: nit: you could probably move `content` on a new line as well.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestExpireSnapshotsAction.java
```diff
@@ -122,13 +140,18 @@ private Long rightAfterSnapshot(long snapshotId) {
     return end;
   }
 
-  private void checkExpirationResults(long expectedDatafiles, long expectedManifestsDeleted,
-      long expectedManifestListsDeleted, ExpireSnapshots.Result results) {
+  private void checkExpirationResults(long expectedDatafiles, long expectedPosDeleteFiles, long expectedEqDeleteFiles,
```
aokolnychyi, line 143: Thanks for fixing the indentation!

aokolnychyi (approved):
LGTM. I had a few mostly optional comments.

szehon-ho:
Thanks @aokolnychyi , @zinking , @dramaticlly , @rajarshisarkar for detailed reviews

PR #4630, opened 2022-04-25
Only delete will throw ValidationException instead of IllegalArgumentException

Andre9580:
`SparkTables metadata only delete` should throw `ValidationException` instead of `IllegalArgumentException` based on @[szehon-ho](https://github.com/szehon-ho) comment.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
@@ -281,7 +281,7 @@ public class SparkTable implements org.apache.spark.sql.connector.catalog.Table,
   }
 
   @Override
-  public void deleteWhere(Filter[] filters) {
+  public void deleteWhere(Filter[] filters) throws ValidationException {
     Expression deleteExpr = SparkFilters.convert(filters);
 
     if (deleteExpr == Expressions.alwaysFalse()) {
@@ -289,14 +289,10 @@ public class SparkTable implements org.apache.spark.sql.connector.catalog.Table,
       return;
     }
 
-    try {
-      icebergTable.newDelete()
-          .set("spark.app.id", sparkSession().sparkContext().applicationId())
-          .deleteFromRowFilter(deleteExpr)
-          .commit();
-    } catch (ValidationException e) {
-      throw new ValidationException(e, "Failed to cleanly delete data files matching: %s", deleteExpr);
-    }
+    icebergTable.newDelete()
+            .set("spark.app.id", sparkSession().sparkContext().applicationId())
+            .deleteFromRowFilter(deleteExpr)
+            .commit();
   }
 
   @Override
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
@@ -281,7 +281,7 @@ public class SparkTable implements org.apache.spark.sql.connector.catalog.Table,
   }
 
   @Override
-  public void deleteWhere(Filter[] filters) throws ValidationException {
+  public void deleteWhere(Filter[] filters) {
     Expression deleteExpr = SparkFilters.convert(filters);
 
     if (deleteExpr == Expressions.alwaysFalse()) {
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
@@ -289,9 +289,9 @@ public class SparkTable implements org.apache.spark.sql.connector.catalog.Table,
     }
 
     icebergTable.newDelete()
-            .set("spark.app.id", sparkSession().sparkContext().applicationId())
-            .deleteFromRowFilter(deleteExpr)
-            .commit();
+        .set("spark.app.id", sparkSession().sparkContext().applicationId())
+        .deleteFromRowFilter(deleteExpr)
+        .commit();
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java
```diff
@@ -295,7 +295,7 @@ public void deleteWhere(Filter[] filters) {
           .deleteFromRowFilter(deleteExpr)
           .commit();
     } catch (ValidationException e) {
-      throw new IllegalArgumentException("Failed to cleanly delete data files matching: " + deleteExpr, e);
+      throw new ValidationException(e, "Failed to cleanly delete data files matching: %s", deleteExpr);
```
szehon-ho, line 298: Thanks, I mean, can we remove catch block entirely?  (It should throw the ValidationException directly from the method, no need to catch and re-throw)

Andre9580: @szehon-ho just to confirm you want the `ValidationException` directly thrown by the method `deleteWhere`?
Like this:
```
@Override
  public void deleteWhere(Filter[] filters) throws ValidationException {
    Expression deleteExpr = SparkFilters.convert(filters);

    if (deleteExpr == Expressions.alwaysFalse()) {
      LOG.info("Skipping the delete operation as the condition is always false");
      return;
    }

    icebergTable.newDelete()
            .set("spark.app.id", sparkSession().sparkContext().applicationId())
            .deleteFromRowFilter(deleteExpr)
            .commit();
  }
```

szehon-ho: Yea that's right, I think we can even remove  it from method's throw clause, as it's a RuntimeException.

Andre9580: Updated the code, however regarding "we can even remove it from method's throw clause" I am not sure to follow you, if you can explain this it would be great since I thought we actually wanted to throw a `ValidationException`.

szehon-ho: So I mean, we don't need the declaration 'throws ValidationException' =>

``` public void deleteWhere(Filter[] filters) { ```

We will still throw it from the internal method, but don't need to declare it as it's a RuntimeException.

Comment (Andre9580) [reply spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java]: Got it, that makes sense, thank you for the explanation, removed it.

File: spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java [hunk 2]
```diff
@@ -289,14 +288,10 @@ public void deleteWhere(Filter[] filters) {
...
-    try {
-      icebergTable.newDelete()
-          .set("spark.app.id", sparkSession().sparkContext().applicationId())
-          .deleteFromRowFilter(deleteExpr)
-          .commit();
-    } catch (ValidationException e) {
-      throw new IllegalArgumentException("Failed to cleanly delete data files matching: " + deleteExpr, e);
-    }
+    icebergTable.newDelete()
+            .set("spark.app.id", sparkSession().sparkContext().applicationId())
```
szehon-ho, line 292: Can we keep the same indent level (4 spaces)?

Andre9580: Added the code style XML to my IDE, it's fixed now.

szehon-ho (approved):

rajarshisarkar (approved):
Looks good to me. Thanks, @Andre9580!

Andre9580:
Thanks both of you for the review, @[szehon-ho](https://github.com/szehon-ho) your comment makes sense to me, I can update the other Spark version as well later today :)

szehon-ho (approved):

Andre9580:
It seems like a test `org.apache.iceberg.spark.sql.TestDeleteFrom` is failing for `Spark 3.0` although the test is successful for Spark `3.1` and `3.2`.

```
org.apache.iceberg.spark.sql.TestDeleteFrom > testDeleteFromUnpartitionedTable[catalogName = testhadoop, implementation = org.apache.iceberg.spark.SparkCatalog, config = {type=hadoop}] FAILED
    java.lang.AssertionError: [Should not delete when not all rows of a file match the filter] 
    Expecting actual throwable to be an instance of:
      java.lang.IllegalArgumentException
    but was:
      org.apache.iceberg.exceptions.ValidationException: Cannot delete file where some, but not all, rows match filter ref(name="id") < 2: file:/var/folders/_y/3bcnyg2s5hvfrbvr715c06n40000gn/T/warehouse5192576849229022400.tmp/default/table/data/00000-6-dbcb782f-4f45-4758-97e1-e6635f5fab97-00001.parquet
    	at org.apache.iceberg.exceptions.ValidationException.check(ValidationException.java:50)
    	at org.apache.iceberg.ManifestFilterManager.manifestHasDeletedFiles(ManifestFilterManager.java:365)
    	at org.apache.iceberg.ManifestFilterManager.filterManifest(ManifestFilterManager.java:307)
    	...(9 remaining lines not displayed - this can be changed with Assertions.setMaxStackTraceElementsDisplayed)
        at org.apache.iceberg.AssertHelpers.assertThrows(AssertHelpers.java:47)
        at org.apache.iceberg.spark.sql.TestDeleteFrom.testDeleteFromUnpartitionedTable(TestDeleteFrom.java:65)
```

szehon-ho:
Yea that makes sense, I think we should fix the test.

rajarshisarkar:
Yeah, we need to change the expected exception in the test which is failing:

https://github.com/apache/iceberg/blob/216a085b603859927c43b280d2b4033ac2b2e61e/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/sql/TestDeleteFrom.java#L65-L67

Andre9580:
Fixed the failing tests, thanks for guidance @[rajarshisarkar](https://github.com/rajarshisarkar).

szehon-ho (approved):

szehon-ho:
Thanks @Andre9580 for the contribution and @rajarshisarkar for extra review

PR #4632, opened 2022-04-26
Core - Implement MetadataUpdateParser for AddSchema, SetCurrentSchema, and SetDefaultPartitionSpec

kbendick:
This covers a few more parsers for MetadataUpdate:
-  AddSchema
- SetCurrentSchema
- SetDefaultPartitionSpec

Given that this one parser covers several classes, I am breaking this up into multiple PRs to make it easier to review.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
@@ -185,21 +185,13 @@ public class MetadataUpdateParser {
       case SET_DEFAULT_PARTITION_SPEC:
         return readAsSetDefaultPartitionSpec(jsonNode);
       case ADD_PARTITION_SPEC:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case ADD_SORT_ORDER:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case SET_DEFAULT_SORT_ORDER:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case ADD_SNAPSHOT:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case REMOVE_SNAPSHOTS:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case SET_SNAPSHOT_REF:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case SET_PROPERTIES:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case REMOVE_PROPERTIES:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       case SET_LOCATION:
         throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       default:
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
@@ -56,6 +56,19 @@ public class MetadataUpdateParser {
   // UpgradeFormatVersion
   private static final String FORMAT_VERSION = "format-version";
 
+  // AddSchema
+  private static final String SCHEMA = "schema";
+  private static final String LAST_COLUMN_ID = "last-column-id";
+
+  // SetCurrentSchema
+  private static final String SCHEMA_ID = "schema-id";
+
+  // AddPartitionSpec
+  private static final String SPEC = "spec";
+
+  // SetDefaultPartitionSpec
+  private static final String SPEC_ID = "spec-id";
+
   private static final Map<Class<? extends MetadataUpdate>, String> ACTIONS = ImmutableMap
       .<Class<? extends MetadataUpdate>, String>builder()
       .put(MetadataUpdate.AssignUUID.class, ASSIGN_UUID)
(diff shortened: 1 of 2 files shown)


kbendick:
I left `AddPartitionSpec` unimplemented because of concerns about processing incoming JSON that does not have field-id set when parsed, which would be assigned default field ids (starting at 1000). Want to double check any potential consequences or realities of that happening.

 I'm not even sure that could happen / needs to be accounted for, but after having that thought I figured it would be better to open the `AddPartitionSpec` SetDe in a separate PR given that smaller PRs are generally better anyway. I'll open that one shortly.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -96,17 +109,42 @@ public static String toJson(MetadataUpdate metadataUpdate, boolean pretty) {
...
+      case SET_CURRENT_SCHEMA:
+        writeAsSetCurrentSchema((MetadataUpdate.SetCurrentSchema) metadataUpdate, generator);
         break;
+      case SET_DEFAULT_PARTITION_SPEC:
+        writeAsSetDefaultPartitionSpec((MetadataUpdate.SetDefaultPartitionSpec) metadataUpdate, generator);
+        break;
+      case ADD_PARTITION_SPEC:
+        throw new UnsupportedOperationException("Not Implemented: MetadataUpdate#toJson for " + updateAction);
+      case ADD_SORT_ORDER:
+        throw new UnsupportedOperationException("Not Implemented: MetadataUpdate#toJson for " + updateAction);
```
kbendick, line 133: Originally this used case fall through to get the process of listing all cases into the source code, but error prone complains when switch cases fall through (rightly so).

kbendick:
cc @rdblue

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -140,21 +170,73 @@ public static MetadataUpdate fromJson(JsonNode jsonNode) {
...
+        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
       default:
         throw new UnsupportedOperationException(
             String.format("Cannot convert metadata update action to json: %s", action));
     }
   }
 
-  // Write all fields specific to UpgradeFormatVersion
-  private static void writeUpgradeFormatVersion(MetadataUpdate.UpgradeFormatVersion update, JsonGenerator gen)
+  private static void writeAsUpgradeFormatVersion(MetadataUpdate.UpgradeFormatVersion update, JsonGenerator gen)
```
kbendick, line 195: Updated the function name to `writeAs*` to match the existing `readAs*`.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -60,15 +82,119 @@ public void testUpgradeFormatVersionFromJson() {
...
+  @Test
+  public void testSetDefaultPartitionSpecFromJson() {
+    String action = SET_DEFAULT_PARTITION_SPEC;
+    int specId = 4;
+    String json = String.format("{\"action\":\"%s\",\"spec-id\":%d}", action, specId);
+    MetadataUpdate.SetDefaultPartitionSpec expected = new MetadataUpdate.SetDefaultPartitionSpec(specId);
+    assertEquals(action, expected, MetadataUpdateParser.fromJson(json));
+  }
+
   public void assertEquals(String action, MetadataUpdate expectedUpdate, MetadataUpdate actualUpdate) {
```
kbendick, line 145: We could just have each unit test call the correct `assertEquals` function based on the type.

However, this tests comparing two `MetadataUpdate` objects first and foremost. As we'll mostly be working with `MetadataUpdate` and not the specific subtype, I went with this.

I am open to suggestions here 🙂

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -140,21 +170,73 @@ public static MetadataUpdate fromJson(JsonNode jsonNode) {
...
+        return readAsAddSchema(jsonNode);
+      case SET_CURRENT_SCHEMA:
+        return readAsSetCurrentSchema(jsonNode);
+      case SET_DEFAULT_PARTITION_SPEC:
+        return readAsSetDefaultPartitionSpec(jsonNode);
+      case ADD_PARTITION_SPEC:
+      case ADD_SORT_ORDER:
+      case SET_DEFAULT_SORT_ORDER:
+      case ADD_SNAPSHOT:
+      case REMOVE_SNAPSHOTS:
```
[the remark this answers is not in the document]
kbendick: So I had the same thought, but in the spec, the action is defined as `remove-snapshots`.

If you look at the implementation, it immediately converts the single snapshot ID to a set (which is what the spec defines).

So it might make more sense to eventually rename the class `RemoveSnapshots` and/or add a constructor that takes in a collection / iterable. That's what I intend to do in one of my follow up PRs that implements the serde for this class.

Good catch though!

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
rdblue: I don't think the names make sense with `As`. I think it's okay if it doesn't quite match. I'd opt for names that you can read easily.

rdblue: I think it would also be fine to drop `As` from the read side if you want them to match. `readAddSchema` makes sense to me.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -56,6 +56,19 @@ private MetadataUpdateParser() {
...
 
+  // AddSchema
+  private static final String SCHEMA = "schema";
+  private static final String LAST_COLUMN_ID = "last-column-id";
+
+  // SetCurrentSchema
+  private static final String SCHEMA_ID = "schema-id";
+
+  // AddPartitionSpec
+  private static final String SPEC = "spec";
```
rdblue, line 67: Did you intend to add `AddPartitionSpec` to this PR?

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -140,21 +170,73 @@ public static MetadataUpdate fromJson(JsonNode jsonNode) {
...
     int formatVersion = JsonUtil.getInt(FORMAT_VERSION, node);
     return new MetadataUpdate.UpgradeFormatVersion(formatVersion);
   }
+
+  private static MetadataUpdate readAsAddSchema(JsonNode node) {
+    Preconditions.checkArgument(node.hasNonNull(SCHEMA),
+        "Cannot parse missing field: schema");
+    JsonNode schemaNode = node.get(SCHEMA);
+    Preconditions.checkArgument(schemaNode.isObject(),
+        "Invalid type for schema field. Expected object");
```
rdblue, line 226: I think this is already checked in `SchemaParser.fromJson`. It reads the type and then makes sure it is a struct, which should provide a good error message.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -21,11 +21,33 @@
...
 import java.util.List;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.types.Types;
+import org.assertj.core.api.Assertions;
 import org.junit.Assert;
 import org.junit.Test;
 
 public class TestMetadataUpdateParser {
 
+  // MetadataUpdate actions
```
rdblue, line 31: Can you make the other references package-private so you don't have to copy them? Or is this for testing purposes so that you ensure you're not altering tests by altering the constants? (I probably wouldn't worry about that)

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
rdblue: This looks reasonable to me.

rdblue (approved):

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Originally I had it in this PR, but I realized I had some concerns around `fieldId`, so I took it out to allow that conversation to be had in a more isolated place.

Seems I forgot to remove its fields. My bad!

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Ok. I'll drop `As` from the read side. Thanks for the feedback.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Ah good point. I'll defer to the parser for consistency.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: package-private would be fine. The consideration around not altering constants was not something I'd thought of.

PR #4553, opened 2022-04-13
Add Support For Flink 1.15

kbendick:
The newest Flink release, Flink 1.15, is being voted on right now. Currently, the latest release candidate is Flink 1.15-rc4, which is what's being used in this PR.

As we are deprecating Flink 1.12 in https://github.com/apache/iceberg/pull/4551, I've begun the process of migrating the 1.14 code to Flink 1.15.

We are using some deprecated APIs, so this includes the minimum changes needed to make things work.

The (current) diff between 1.14 and 1.15 is located here: https://gist.github.com/kbendick/8265987b7386dc2738b09d7977abe41b

PR #4635, opened 2022-04-26
Flink - Fix flakey, order-dependent FlinkTableSource tests for Flink 1.14

kbendick:
When upgrading our Flink code for the upcoming Flink 1.15 release in https://github.com/apache/iceberg/pull/4553, it was discovered that there is a failing test: https://github.com/apache/iceberg/pull/4553#issuecomment-1107688576

Looking closer, the expected and actual results are the same, just in a different order.

Because we are not testing the ordering of records returned from the source here, using list comparison doesn't make sense and we can just use set comparison (with an initial size check) for any query that returns more than 1 record.

I have updated all queries in this test to use set comparison semantics as list comparison is inherently flakey.

I will port this over to Flink 1.15 in https://github.com/apache/iceberg/pull/4553 as that's where tests are failing (though we don't have any guarantee that they will pass at present in 1.14 and it's simply implementation specific that they do).

cc @openinx @yittg

### DIFF
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java
@@ -136,7 +136,7 @@ public class TestFlinkTableSource extends FlinkTestBase {
         Row.of(2, "b", 20.0),
         Row.of(3, null, 30.0)
     );
-    assertSameElements(result, expectedRecords);
+    assertSameElements(expectedRecords, result);
     Assert.assertEquals("Should not push down a filter", Expressions.alwaysTrue(), lastScanEvent.filter());
   }
 
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java
@@ -271,7 +271,7 @@ public class TestFlinkTableSource extends FlinkTestBase {
         Row.of(1, "iceberg", 10.0),
         Row.of(2, "b", 20.0)
     );
-    assertSameElements(expectedGT, resultGT);;
+    assertSameElements(expectedGT, resultGT);
 
     Assert.assertEquals("Should create only one scan", 1, scanEventCount);
     Assert.assertEquals("Should contain the push down filter", expectedFilter, lastScanEvent.filter().toString());
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java
@@ -289,7 +289,7 @@ public class TestFlinkTableSource extends FlinkTestBase {
         Row.of(2, "b", 20.0),
         Row.of(3, null, 30.0)
     );
-    assertSameElements(resultGTE, expectedGTE);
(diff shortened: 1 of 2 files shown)


danielcweeks:
This is one of those cases where AssertJ could help with convenience checks like `assertThat(actual).hasSameElementsAs(expected)`, but since we're not using that library here yet, this seems like a reasonable approach. Alternatively, you could do something like the following wrapped in a utility function:

```
assertTrue(expected.size() == actual.size() && expected.containsAll(actual) && actual.containsAll(expected));
``

**Comment by kbendick** [comment]
Yeah adding a convenience function makes sense.

**Comment by kbendick** [comment]
Actually, it seems we do have the `AssertJ` library available for this subproject, so I'll just use that unless there's opposition to it.

**Comment by kbendick** [comment]
Added a utility function `assertSameElements` into `FlinkTestBase`.

In the interest of keeping the diff minimal, I didn't remove existing assertions on the size of the result (although that is checked by `assertSameElements`.

I also didn't update any of the results with only one element. If we want to do that, we can use `Collections.singleton` to reuse the comparison.

Let me know if removing the (redundant) size checks is a good idea. I think so because in the case of a failure, we'll see the incorrect records instead of just the incorrect size but I opted to be conservative and change less code.

File: flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkTableSource.java [hunk 1]
```diff
@@ -119,7 +119,7 @@ public void testLimitPushDown() {
         Row.of(2, "b", 20.0),
         Row.of(3, null, 30.0)
     );
-    Assert.assertEquals("Should produce the expected records", expectedList, resultExceed);
+    assertSameElements(expectedList, resultExceed);
```
yittg, line 122: How about keeping this description by `#describeAs(...)`

kbendick: For AssertJ, the assertion message is already pretty good. I updated one test so it would fail and here's the output I got. I think it's pretty clean / self explanatory.

```
Expecting:
  [+I[1, iceberg, 10.0], +I[2, b, 20.0]]
to contain exactly in any order:
  [+I[1, iceberg, 10.0]]
but the following elements were unexpected:
  [+I[2, b, 20.0]]
```

With `.as` before the `containsExactlyInAnyOrderElementsOf`, here's the output. I don't personally think it provides much value.

```
java.lang.AssertionError: [should contain the same elements] 
Expecting:
  [+I[1, iceberg, 10.0], +I[2, b, 20.0]]
to contain exactly in any order:
  [+I[1, iceberg, 10.0]]
but the following elements were unexpected:
  [+I[2, b, 20.0]]
```

yittg: I think we should make the assertSameElements open(can be optional) for description.
For the case you mentioned, i think it didn't provide much value because the statement itself, not for all other descriptions.
WDYT @kbendick ?

kbendick: So you mean add a description argument like `assertSameElements(String errorMessage, List<Row> exepected, List<Row> actual)`?

I'm happy to add that as well as keep the original signature. I can't think of a place in the current file that it would be useful as they're all the same scenarios, but I'm happy to add it.

kbendick: Added that signature in should anybody want to use it @yittg.

PR #4636, opened 2022-04-26
Core - Use singleton RESTObjectMapper in REST RequestResponseTestBase class

kbendick:
There is now a global singleton object mapper for the `RESCatalog`, which already has the necessary serializers and deserializers registered.

The rest / response objects are all tested via the `RequestResponseTestBase` class, which currently repeats the object mapper instantiation and initialization logic that the `RESTObjectMapper` is ultimately responsible for.

We should be testing the rest objects using the correct object mapper (which should also reduce allocations during testing and other minor advantages).

cc @rdblue @danielcweeks

### DIFF
--- a/core/src/test/java/org/apache/iceberg/rest/responses/TestCreateNamespaceResponse.java
+++ b/core/src/test/java/org/apache/iceberg/rest/responses/TestCreateNamespaceResponse.java
@@ -101,7 +101,8 @@ public class TestCreateNamespaceResponse extends RequestResponseTestBase<CreateN
     String jsonMisspelledKeys = "{\"namepsace\":[\"accounting\",\"tax\"],\"propertiezzzz\":{\"owner\":\"Hank\"}}";
     AssertHelpers.assertThrows(
         "A JSON response with the keys spelled incorrectly should fail to deserialize and validate",
-        JsonProcessingException.class,
+        IllegalArgumentException.class,
+        "Invalid namespace: null",
         () -> deserialize(jsonMisspelledKeys)
     );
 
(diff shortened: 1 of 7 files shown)


danielcweeks (approved):
LGTM (just waiting on checks)

core/src/test/java/org/apache/iceberg/rest/responses/TestCreateNamespaceResponse.java
```diff
@@ -101,7 +101,8 @@ public void testDeserializeInvalidResponse() {
     String jsonMisspelledKeys = "{\"namepsace\":[\"accounting\",\"tax\"],\"propertiezzzz\":{\"owner\":\"Hank\"}}";
     AssertHelpers.assertThrows(
         "A JSON response with the keys spelled incorrectly should fail to deserialize and validate",
-        JsonProcessingException.class,
+        IllegalArgumentException.class,
+        "Invalid namespace: null",
```
kbendick, line 105: Because `RESTObjectMapper` will ignore unknown fields, the exception thrown in these tests where the keys are misspelled turn out to be more specific.

I've updated them to use the proper thrown exception type, as well as add part of the exception message to help clarify the tests purpose and intent.

core/src/test/java/org/apache/iceberg/rest/responses/TestRESTCatalogConfigResponse.java
```diff
@@ -156,14 +156,6 @@ public void testDeserializeInvalidResponse() {
         () -> deserialize(jsonOverridesHasWrongType)
     );
 
-    String jsonMisspelledKeys =
-        "{\"defaultzzzzzzz\":{\"warehouse\":\"s3://bucket/warehouse\"},\"overrrrrrrrides\":{\"clients\":\"5\"}}";
-    AssertHelpers.assertThrows(
-        "A JSON response with the keys spelled incorrectly should fail to deserialize and validate",
-        JsonProcessingException.class,
-        () -> deserialize(jsonMisspelledKeys)
-    );
```
kbendick, line 165: Because the RESTObjectMapper does not fail on unknown keys, this test is no longer valid.

We might consider adding a warning log in the config response's `validate` method to log if `null` was received for both (not an explicit empty collection, but null).

In practice, users should be using the config response class and its associated builder, which cannot build an incorrect payload.

PR #4641, opened 2022-04-27
Core - Add serde for TableMetadata for REST catalog

kbendick:
The REST catalog sends and receives `TableMetadata` objects as JSON.

There is an existing parser for `TableMetadata` that can convert metadata into JSON and also can parse JSON into `TableMetadata`.

Adds Jackson serializers to the `RESTSerializers` so that accessing tables via the REST catalog works.

There are existing tests for `TableMetadataParser`, and I intend to add further tests in a follow up PR around `LoadTableResponse`. This PR is intended so allow people to continue implementing REST catalogs and be able to work with TableMetadata, but more extensive coverage will come in a follow up.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableMetadataParser.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadataParser.java
@@ -155,7 +155,7 @@ public class TableMetadataParser {
     }
   }
 
-  private static void toJson(TableMetadata metadata, JsonGenerator generator) throws IOException {
+  public static void toJson(TableMetadata metadata, JsonGenerator generator) throws IOException {
     generator.writeStartObject();
 
     generator.writeNumberField(FORMAT_VERSION, metadata.formatVersion());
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/TableMetadataParser.java
```diff
@@ -301,6 +301,10 @@ static TableMetadata fromJson(FileIO io, InputFile file, JsonNode node) {
     return fromJson(io, file.location(), node);
   }
 
+  public static TableMetadata fromJson(JsonNode node) {
+    return fromJson(null, (String) null, node);
+  }
```
kbendick, line 306: This parser will return a TableMetadata that presently can raise an NPE if the changes of a particular snapshot are queried, as the FileIO object is not present.

Unfortunately, the needed `FileIO` is not available in the scope of the serializers used by Jackson with the `RESTCatalog`.

To get around this, in a follow up PR we will take the functions in `BaseSnapshot` that _do_ potentially depend on `FileIO` and have them take an explicit `FileIO` object as an argument.

This will allow for catalogs to pick the correct FileIO when needed, and removes the need for this class to carry around as much state (which it passes to children in its snapshot log etc).

But if the TableMetadata doesn't have a call to access added files, deleted files, or added / deleted manifests, or if there is no Snapshot log, then the object is correct and useable just fine.

rdblue (approved):

rdblue:
This looks like a good first step to me. We'll have to fix the FileIO problem, but eventually we don't want to be passing FileIO into TableMetadata anyway. Everything should work fine and we can fix up the FileIO in a follow-up.

PR #4649, opened 2022-04-27
Flink 1.14: Implement LookupTableSource for IcebergTableSource

hailin0:
Please help review this change

reference: [flink-jdbc-lookup-table](https://github.com/apache/flink/blob/release-1.14.0/flink-connectors/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/table/JdbcRowDataLookupFunction.java)

@rdblue


bundled-guava/src/main/java/org/apache/iceberg/GuavaClasses.java
```diff
@@ -94,6 +96,8 @@ public class GuavaClasses {
     Iterables.class.getName();
     CountingOutputStream.class.getName();
     Suppliers.class.getName();
+    Cache.class.getName();
+    CacheBuilder.class.getName();
```
kbendick, line 100: Usually we use the [`Caffeine` caches](https://github.com/ben-manes/caffeine) instead (which mostly have the same API as Guava's).

Is there a reason that `Caffeine` won't work here?

flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
```diff
@@ -142,6 +143,12 @@ public Set<ConfigOption<?>> optionalOptions() {
     Set<ConfigOption<?>> options = Sets.newHashSet();
     options.add(CATALOG_DATABASE);
     options.add(CATALOG_TABLE);
+    options.add(FlinkLookupOptions.LOOKUP_CACHE_MAX_ROWS);
+    options.add(FlinkLookupOptions.LOOKUP_CACHE_TTL);
+    options.add(FlinkLookupOptions.LOOKUP_CACHE_IGNORE_EMPTY);
+    options.add(FlinkLookupOptions.LOOKUP_MAX_RETRIES);
+    options.add(FlinkLookupOptions.LOOKUP_BASE_RETRY_BACKOFF_MILLS);
+    options.add(FlinkLookupOptions.LOOKUP_MAX_RETRY_BACKOFF_MILLS);
```
kbendick, line 151: Usually, we implement new features against just the latest version (in this case, Flink 1.14).

This makes reviewing easier as there isn't code duplication and comments aren't split across two sets of files.

Then, after everything is reviewed and merged, we back port to previous versions.

Is it possible to keep this just to Flink 1.14 for now?

kbendick (commented):
Thanks for this PR @hailin0!

I have some general concerns about performance for `LookupFunction` reading row by row against the table. But I don't think that should stop us from exploring the problem necessarily.

I left some notes about sticking to the latest Flink version during review, and also using `Caffeine` cache instead of the Guava caches.

bundled-guava/src/main/java/org/apache/iceberg/GuavaClasses.java
hailin0: Accepted, please check again

flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
hailin0: Accepted, please check again

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
```diff
@@ -142,6 +143,12 @@ public Set<ConfigOption<?>> optionalOptions() {
     Set<ConfigOption<?>> options = Sets.newHashSet();
     options.add(CATALOG_DATABASE);
     options.add(CATALOG_TABLE);
+    options.add(FlinkLookupOptions.LOOKUP_CACHE_MAX_ROWS);
+    options.add(FlinkLookupOptions.LOOKUP_CACHE_TTL);
+    options.add(FlinkLookupOptions.LOOKUP_CACHE_IGNORE_EMPTY);
+    options.add(FlinkLookupOptions.LOOKUP_MAX_RETRIES);
+    options.add(FlinkLookupOptions.LOOKUP_BASE_RETRY_BACKOFF_MILLS);
+    options.add(FlinkLookupOptions.LOOKUP_MAX_RETRY_BACKOFF_MILLS);
```
rdblue, line 151: Can we separate caching from the lookup source implementation?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/IcebergTableSource.java
```diff
@@ -47,7 +55,8 @@
  * Flink Iceberg table source.
  */
 public class IcebergTableSource
-    implements ScanTableSource, SupportsProjectionPushDown, SupportsFilterPushDown, SupportsLimitPushDown {
+    implements ScanTableSource, LookupTableSource, SupportsProjectionPushDown, SupportsFilterPushDown,
```
stevenzwu, line 58: It is probably common to load Iceberg as `ScanTableSource`.  But it is unclear how Iceberg fits as a lookup source. @hailin0 can you provide some context on the actual use cases?

I thought LookupTableSource is meant for external storage that supports fast point query (like HBase/JDBC). `HBaseDynamicTableSource` also supports async lookup function.

Here is the Javadoc from `LookupTableSource`
```
 * A {@link DynamicTableSource} that looks up rows of an external storage system by one or more keys
 * during runtime.
 *
 * <p>Compared to {@link ScanTableSource}, the source does not have to read the entire table and can
 * lazily fetch individual values from a (possibly continuously changing) external table when
 * necessary.
```

docs/flink/flink-connector.md
```diff
@@ -141,5 +147,16 @@ SELECT * FROM flink_table;
 +----+------+
 3 rows in set
 ```
+## Lookup Join
```
samredai, line 150: Shouldn't this be next to the Flink SQL content? I don't like that all of Flink SQL is covered in a page called "Getting Started" but I think that section should be broken into two: "SQL" and "DataStream", in which case this will go in the "SQL" section.

docs/flink/flink-connector.md
```diff
@@ -141,5 +147,16 @@ SELECT * FROM flink_table;
...
+
+Iceberg connector can be used in temporal join as a [lookup source](https://ci.apache.org/projects/flink/flink-docs-release-1.14/docs/dev/table/sql/queries/joins/#lookup-join) (aka. dimension table). Currently, only sync lookup mode is supported.
```
samredai, line 152: Can we include an Iceberg specific sql snippet here?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
hailin0: sorry, i don't understand what you mean.
FlinkLookupFunction#eval method has no return value, how to cache function return value?

https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-hbase-base/src/main/java/org/apache/flink/connector/hbase/source/HBaseRowDataLookupFunction.java#L96
https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/table/JdbcRowDataLookupFunction.java#L146

docs/flink/flink-connector.md
hailin0: please check again

hailin0:
> Hi @hailin0, thanks for your contribution, we all have some concerns about performance and I have a friendly question, have you ever done a test or benchmark for this PR? I am curious about the performance of the point query for Iceberg. AFAIK, Iceberg has some general characteristics with Hive, such that loading all data in memory to accelerate the point query is more reasonable, just like the implementation of Hive Lookup Join, and the community is driving the `[hash lookup join](https://issues.apache.org/jira/browse/FLINK-23687)`, which will bring more benefits, WDYS?

I agree to add the option loading all data in memory

PR #4651, opened 2022-04-27
Core: Optimize to delete the oldest metadata file after commit

southernriver:
This is to address 2 issues:

- Just construct `Set` of  `removedPreviousMetadataFiles` After  `write.metadata.delete-after-commit.enabled = true`
- Delete the function of `deleteRemovedMetadataFiles `  in `BaseMetastoreTableOperations` and `HadoopTableOperations`,  and then implement it in interface of `TableOperations`

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableOperations.java
+++ b/core/src/main/java/org/apache/iceberg/TableOperations.java
@@ -19,15 +19,11 @@
 
 package org.apache.iceberg;
 
-import java.util.Set;
 import java.util.UUID;
 import org.apache.iceberg.encryption.EncryptionManager;
 import org.apache.iceberg.encryption.PlaintextEncryptionManager;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.LocationProvider;
-import org.apache.iceberg.relocated.com.google.common.collect.Sets;
-import org.apache.iceberg.util.Tasks;
-import org.slf4j.LoggerFactory;
 
 /**
  * SPI interface to abstract table metadata access and updates.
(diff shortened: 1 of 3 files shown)


szehon-ho:
I think the first change makes sense, the second one of raising it to the TableOperations base class may not make sense, I am not sure if all TableOperations in the future (particularly REST) will use this mechanism of metadata.json files?

kbendick:
I agree with @szehon-ho's assessment.

core/src/main/java/org/apache/iceberg/TableOperations.java
```diff
@@ -120,4 +124,31 @@ default long newSnapshotId() {
...
+        TableProperties.METADATA_DELETE_AFTER_COMMIT_ENABLED,
+        TableProperties.METADATA_DELETE_AFTER_COMMIT_ENABLED_DEFAULT);
+
+    if (deleteAfterCommit) {
+      Set<TableMetadata.MetadataLogEntry> removedPreviousMetadataFiles = Sets.newHashSet(base.previousFiles());
+      removedPreviousMetadataFiles.removeAll(metadata.previousFiles());
+      Tasks.foreach(removedPreviousMetadataFiles)
+          .noRetry().suppressFailureWhenFinished()
+          .onFailure((previousMetadataFile, exc) ->
+              LoggerFactory.getLogger(TableOperations.class).warn("Delete failed for previous metadata file: {}",
```
kbendick, line 148: Nit: I don't think this move is necessary, but if it does happen, we shouldn't be instantiating the logger every time we call the log function.

southernriver:
@szehon-ho @kbendick  Thanks for the review, code has been updated.

szehon-ho (approved):

singhpk234 (approved):

szehon-ho:
Thanks @southernriver for commit, @kbendick  and @singhpk234 for extra review

PR #4653, opened 2022-04-27
Core - Rename test class for ConfigResponse to match updated name.

kbendick:
The class previously called `RESTCatalogConfigResponse` was renamed to `ConfigResponse` in https://github.com/apache/iceberg/pull/4554

However, the corresponding test class was not renamed.


rajarshisarkar (approved):
Looks good to me. Thanks, @kbendick!

PR #4658, opened 2022-04-28
Flink: Support drop non-empty namespace with CASCADE

Zephyr589:
`DROP NAMESPACE [IF EXISTS] {ns} [CASCADE]` does not work with Iceberg on Flink now.
``` 
Flink SQL> drop database test_db CASCADE;
[ERROR] Could not execute SQL statement. Reason:
org.apache.hadoop.hive.metastore.api.InvalidOperationException: Database test_db is not empty. One or more tables exist.
```
I've added support for `CASCADE` and there are tests to verify the new behavior.
Please help review this change @openinx @kbendick


flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -229,6 +229,20 @@ public void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade
...
       try {
+        if (cascade) {
+          List<ObjectPath> deleteTablePaths = icebergCatalog.listTables(toNamespace(name)).stream()
+                  .map(s -> ObjectPath.fromString(s.toString()))
+                  .collect(Collectors.toList());
+          deleteTablePaths.forEach(
+                  objectPath -> {
+                    try {
+                      dropTable(objectPath, true);
+                    } catch (TableNotExistException ex) {
```
hililiwei, line 240: If we encounter other exception while dropping the table, should we continue to drop other tables, or should we abort the drop operation?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -229,6 +229,20 @@ public void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade
       throws DatabaseNotExistException, DatabaseNotEmptyException, CatalogException {
     if (asNamespaceCatalog != null) {
       try {
+        if (cascade) {
+          List<ObjectPath> deleteTablePaths = icebergCatalog.listTables(toNamespace(name)).stream()
```
hililiwei, line 233: Consider parallel operations using `Tasks` here?

hililiwei (commented):
Thank you for your contribution. 😄

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -229,6 +229,20 @@ public void dropDatabase(String name, boolean ignoreIfNotExists, boolean cascade
...
+                  .map(s -> ObjectPath.fromString(s.toString()))
```
hililiwei, line 234: A more meaningful name is recommended.

Zephyr589:
@hililiwei Thank you for your review. I am very glad to receive your advice, I have make some changes according to your suggestion.

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
Zephyr589: We can ignore it, and drop other tables.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestFlinkCatalogDatabase.java
```diff
@@ -273,4 +273,40 @@ public void testHadoopNotSupportMeta() {
...
+    Assert.assertTrue("Namespace should still exist after a failed drop",
+            validationNamespaceCatalog.namespaceExists(icebergNamespace));
+    Assert.assertTrue("Table should still exist after a failed drop",
+            validationCatalog.tableExists(TableIdentifier.of(icebergNamespace, "tl")));
+
+    sql("DROP DATABASE %s CASCADE", flinkDatabase);
+
+    Assert.assertFalse("Namespace should not exist after deleting with CASCADE",
+            validationNamespaceCatalog.namespaceExists(icebergNamespace));
+    Assert.assertFalse("Namespace should not exist after deleting its parent namespace with CASCADE",
```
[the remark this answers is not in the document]
Zephyr589: Yes

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
hililiwei: Is it possible to get an exception other than `TableNotExistException`?

Zephyr589: Since I made these changes based on the way Flink implements it, I think we should not change it.
image

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
Zephyr589: If droping a database and droping a table under it happen at the same time, it will make no influence.
If droping a database and creating a table under it happen at the same time, it will fail, and we just need to retry it.

hililiwei: I mean, Can we delete tables in parallel via `org.apache.iceberg.util.Tasks` ?

flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
hililiwei: We can certainly be consistent with it, I have no objection. I just want to confirm our deletion policy. For now, when encounter an exception other than A, it will abort the deletion, right?

PR #4660, opened 2022-04-28
Core: Make snapshot summary return default values

homar:
Fixes: https://github.com/apache/iceberg/pull/4660

Currently snapshot summary, which is a map, contains values only if they were affected by changes in that snapshot.
It'd be good to return default values, like 0s, for most of the fields. 
This is needed in trino in order to  reliably provide number of deleted rows to the user, also when delete happens to delete no files.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/SnapshotSummary.java
+++ b/core/src/main/java/org/apache/iceberg/SnapshotSummary.java
@@ -185,9 +185,9 @@ public class SnapshotSummary {
       builder.putAll(properties);
 
       metrics.addTo(builder);
-      setIf(deletedDuplicateFiles > 0, builder, DELETED_DUPLICATE_FILES, deletedDuplicateFiles);
+      builder.put(DELETED_DUPLICATE_FILES, String.valueOf(deletedDuplicateFiles));
       Set<String> changedPartitions = partitionMetrics.keySet();
-      setIf(trustPartitionMetrics, builder, CHANGED_PARTITION_COUNT_PROP, changedPartitions.size());
+      builder.put(CHANGED_PARTITION_COUNT_PROP, String.valueOf(changedPartitions.size()));
 
       if (trustPartitionMetrics && changedPartitions.size() <= maxChangedPartitionsForSummaries) {
         setIf(changedPartitions.size() > 0, builder, PARTITION_SUMMARY_PROP, "true");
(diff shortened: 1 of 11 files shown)


findepi:
@danielcweeks can you approve the CI run?

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
```diff
@@ -181,14 +181,14 @@ public Map<String, String> build() {
       builder.putAll(properties);
 
       metrics.addTo(builder);
-      setIf(deletedDuplicateFiles > 0, builder, DELETED_DUPLICATE_FILES, deletedDuplicateFiles);
+      SnapshotSummary.set(builder, DELETED_DUPLICATE_FILES, deletedDuplicateFiles);
```
singhpk234, line 184: [minor] can call set here as we are in same class `SnapshotSummary`
```suggestion
      set(builder, DELETED_DUPLICATE_FILES, deletedDuplicateFiles);
```

homar: Unfortunately non static set is defined and has different signature so SnapshotSummary is necessary
```
public void set(String property, String value) {
      properties.put(property, value);
    }
```

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
```diff
@@ -181,14 +181,14 @@ public Map<String, String> build() {
...
       Set<String> changedPartitions = partitionMetrics.keySet();
-      setIf(trustPartitionMetrics, builder, CHANGED_PARTITION_COUNT_PROP, changedPartitions.size());
+      SnapshotSummary.set(builder, CHANGED_PARTITION_COUNT_PROP, changedPartitions.size());
 
       if (trustPartitionMetrics && changedPartitions.size() <= maxChangedPartitionsForSummaries) {
-        setIf(changedPartitions.size() > 0, builder, PARTITION_SUMMARY_PROP, "true");
+        SnapshotSummary.set(builder, PARTITION_SUMMARY_PROP, "true");
```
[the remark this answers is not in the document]
homar: yep that makes sense, thanks

rdblue:
@homar, can you please update the description with information about what you're doing in this PR and why it is needed?

rdblue:
@homar, looks like there's a commit conflict with the addition of `added-eq-delete-files` and `added-pos-delete-files`. Can you rebase?

homar:
@rdblue done

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
```diff
@@ -354,4 +354,8 @@ private static void setIf(boolean expression, ImmutableMap.Builder<String, Strin
       builder.put(property, String.valueOf(value));
     }
   }
+
+  private static void set(ImmutableMap.Builder<String, String> builder, String property, Object value) {
```
rdblue, line 358: What's the value of this method instead of `builder.set(property, value)`?

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
rdblue: I think it probably makes more sense to just call `set` directly. There's no need for a static method anymore.

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
homar: That's a good point. Thanks

rdblue:
Thanks for the update, @homar.

I'm debating whether this is a good idea. This adds quite a bit more content to each snapshot in metadata, especially when collecting partition summaries. The rationale for this is:

> This is needed in trino in order to reliably provide number of deleted rows to the user, also when delete happens to delete no files.

I'm not sure if that's a good reason to add this. If you want to rely on these summary properties, why not default to 0 if the property isn't there? Seems like the argument against that is the narrow case where the property is modified on the Iceberg side without anyone noticing, but I don't see a reason to modify these properties so the problem is very unlikely to occur.

In addition, I wouldn't expect summary properties to be used to report back. Typically, I'd recommend checking the metadata tree for things like number of deleted records. Although if you're happy with these for user reporting then it should be fine. It is just a summary and not going to the source of truth (the metadata tree).

homar:
@rdblue thanks for your input.
Regarding metadata tree, do you mean that the code below would be better/more precise than using snapshot summary ? 
```
icebergTable.currentSnapshot().dataManifests().stream()
.map(dm -> dm.deletedRowsCount()).reduce(0L, (a,b) -> a+b)
```

rdblue:
@homar, yeah. That's probably a better way if you don't want to rely on a special property name.

homar:
@rdblue I may be missing something but using metadata tree is not very precise/not very feasible.
Consider following scenario(I will use SQL for readability):
```
1. CREATE TABLE orders (orderkey bigint,  orderstatus varchar);
2. INSERT INTO orders VALUES(1, 'ok')
3. INSERT INTO orders VALUES(2, 'invalid')
4. INSERT INTO orders VALUES(3, 'ok')
5. Do some OPTIMIZE/COMPACT or something that will merge these 3 data files into 1 data file
6. DELETE FROM orders WHERE orderkey = 2; // this will create position delete 
7. DELETE FROM orders WHERE EXISTS(SELECT 1) // delete everything - so 2 rows with orderkey 1 and 3
```
if you now check dataManifests from currentSnapshot it will show that 3 rows were deleted. 
When in fact only 2 were deleted in the last operation and 1 in the previous one. 
And I don't see a way from the data manifest to figure out if some rows were deleted in the current snapshot or if the deletions were merged from the previous position deletes/equality deletes

Issue #4661, opened 2022-04-28
Metadata table queries fail if a partition column was reused in V2


szlta:
As already discussed in https://github.com/apache/iceberg/pull/3411#discussion_r821845093 this special case will fail to create a metadata schema with the following error:
```
PartitionSpec initialSpec = PartitionSpec.builderFor(SCHEMA)
    .identity("data")
    .build();
TestTables.TestTable table = TestTables.create(tableDir, "test", SCHEMA, initialSpec, V2_FORMAT_VERSION);

table.updateSpec()
    .removeField("data")
    .commit();

table.updateSpec()
    .addField("data")
    .addField("id")
    .commit();

struct<1000: data: optional string, 1001: data: optional string, 1002: id: optional int>
```
gives:
`Invalid schema: multiple fields for name partition.data: 1000 and 1001 `

It's due to how `Partitioning$partitionType` method collects all specs and creates a union partition type. For V2 tables reused partition columns don't get renamed, they just get a new id.

d4r3topk:
This issue has not been resolved yet for V2 API. 

I have a table where one of the columns is "region" and "region" is also part of the partition spec.


I run a query with spark sql - `INSERT OVERWRITE db_name.table_name PARTITION(region = 'us-east-1') SELECT * FROM data`. This leads to Error: `An error occurred while calling o190.sql. Invalid schema: multiple fields for name region: 16 and 17`

PR #4662, opened 2022-04-28
Core: Metadata table queries fail if a partition column was reused in V2

szlta:
First go at fixing #4661.
The proposed solution described at https://github.com/apache/iceberg/pull/3411#discussion_r823692312 seems very hardly feasible, so this would be a quick fix to just avoid `Invalid schema` errors thrown in any metadata query where the a partition col was reused.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java
+++ b/core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java
@@ -103,13 +103,19 @@ class BaseUpdatePartitionSpec implements UpdatePartitionSpec {
     this.lastAssignedPartitionId = lastAssignedPartitionId;
   }
 
+  private int assignFieldId() {
+    this.lastAssignedPartitionId += 1;
+    return lastAssignedPartitionId;
+  }
+
   /**
-   * Calculates a field ID to the newly added PartitionField.
-   * New field ID is assigned in every case for V1 tables, but for V2 we try to recycle a former ID if possible.
+   * In V2 it searches for a similar partition field in historical partition specs. Tries to match on source field
+   * ID, transform type and target name (optional). If not found or in V1 cases it creates a new PartitionField.
    * @param sourceTransform pair of source ID and transform for this PartitionField addition
-   * @return the calculated field ID
+   * @param name target partition field name, if specified
+   * @return the recycled or newly created partition field
    */
-  private int assignFieldId(Pair<Integer, Transform<?, ?>> sourceTransform) {
+  private PartitionField recycleOrCreatePartitionField(Pair<Integer, Transform<?, ?>> sourceTransform, String name) {
     if (formatVersion == 2 && base != null) {
       int sourceId = sourceTransform.first();
       Transform<?, ?> transform = sourceTransform.second();
@@ -121,13 +127,14 @@ class BaseUpdatePartitionSpec implements UpdatePartitionSpec {
 
       for (PartitionField field : allHistoricalFields) {
         if (field.sourceId() == sourceId && field.transform().equals(transform)) {
-          return field.fieldId();
+          // if target name is specified then consider it too, otherwise not
+          if (name == null || field.name().equals(name)) {
+            return field;
+          }
         }
       }
     }
-
-    this.lastAssignedPartitionId += 1;
-    return lastAssignedPartitionId;
+    return new PartitionField(sourceTransform.first(), assignFieldId(), name, sourceTransform.second());
   }
 
   @Override
@@ -179,8 +186,7 @@ class BaseUpdatePartitionSpec implements UpdatePartitionSpec {
     Preconditions.checkArgument(added == null,
         "Cannot add duplicate partition field %s=%s, already added: %s", name, term, added);
 
-    PartitionField newField = new PartitionField(
-        sourceTransform.first(), assignFieldId(sourceTransform), name, sourceTransform.second());
+    PartitionField newField = recycleOrCreatePartitionField(sourceTransform, name);
     if (newField.name() == null) {
       String partitionName = PartitionSpecVisitor.visit(schema, newField, PartitionNameGenerator.INSTANCE);
       newField = new PartitionField(newField.sourceId(), newField.fieldId(), partitionName, newField.transform());
--- a/core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java
+++ b/core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java
@@ -106,8 +106,8 @@ class BaseUpdatePartitionSpec implements UpdatePartitionSpec {
   /**
    * Calculates a field ID to the newly added PartitionField.
    * New field ID is assigned in every case for V1 tables, but for V2 we try to recycle a former ID if possible.
-   * @param sourceTransform - pair of source ID and transform for this PartitionField addition
-   * @return - the calculated field ID
+   * @param sourceTransform pair of source ID and transform for this PartitionField addition
+   * @return the calculated field ID
    */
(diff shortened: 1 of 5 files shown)


szehon-ho:
I still need to ramp a bit on this, but I thought in that discussion we are going to collapse all of the same partition fields (if some are re-used) and they are uniquely identified with specid?  


example for reference:
```
PartitionSpec initialSpec = PartitionSpec.builderFor(SCHEMA)
    .identity("data")
    .build();
TestTables.TestTable table = TestTables.create(tableDir, "test", SCHEMA, initialSpec, V2_FORMAT_VERSION);

table.updateSpec()
    .removeField("data")
    .commit();

table.updateSpec()
    .addField("data")
    .addField("id")
    .commit();

struct<1000: data: optional string, 1001: data: optional string, 1002: id: optional int>
```

I am thinking the code does try to re-use old partition values in both metadata (ManifestEvaluator) and data filtering, right?  (Let me know if assumption is incorrect).  Example here, we would hopefully use  "data" partition values written by the table's initial spec, for queries after re-adding "data" to the spec.   So then may make sense to collapse the fields together in metadata table?  Welcome any thoughts from @rdblue , @aokolnychyi who have more background in the initial discussion.

szehon-ho:
Yea I understand, my gist of the earlier conversation was that the most optimal solution is to try to re-use the field-id if its the equivalent transformation as a partition field before?  I think the thing to check  would be whether the regular metadata filtering is affected or not (today it seems to work,  via logic in Projection filter being able to resolve it)

szlta:
@szehon-ho I think we were probably talking about the same thing for generating no new ID for a PartitionField that was already seen. I have amended my change based on this. I looked into this question in some detail and have found that this is actually kind of already happening when we are removing part of the PartitionFields:

- we start with
`spec0: 1000: identity(a), 1001: identity(b)`
- then if update the spec so that only `b` remains, `a` doesn't get reassigned a new fieldID, although we do create a new spec:
`spec1: 1000: identity(a)`
- now if we re-add `b` and `c`, then `b` would get assigned ID 1002 and `c` would get ID 1003
- I propose we look through in the past specs to see if `b` as partition field already participated (based on transform function and source IDs) and recycle the old ID, so we end up with:
`spec2: 1000: identity(a), 1001: identity(b), 1002: identity(c)`

Looking at the tests this should work, there's only one failure I get now in `TestMetadataTableScans.testFilesTableScanWithDroppedPartition`:
```
Expected :struct<1000: data_bucket: optional int, 1001: data_bucket_16: optional int, 1002: data_trunc_2: optional string>
Actual   :struct<1000: data_bucket_16: optional int, 1001: data_trunc_2: optional string>
```

This made me think though.. it seems like we have two different ways of generating the partition field column name:
https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java#L460 generates `data_bucket_16`, while
https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L484 generates `data_bucket`. Does anyone know why this discrepancy exists? My gut feeling is that the former would be the correct way.

core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java
```diff
@@ -103,7 +103,29 @@ class BaseUpdatePartitionSpec implements UpdatePartitionSpec {
     this.lastAssignedPartitionId = lastAssignedPartitionId;
   }
 
-  private int assignFieldId() {
+  /**
+   * Calculates a field ID to the newly added PartitionField.
+   * New field ID is assigned in every case for V1 tables, but for V2 we try to recycle a former ID if possible.
+   * @param sourceTransform - pair of source ID and transform for this PartitionField addition
```
szehon-ho, line 109: Nit: I don't see many method with '-', maybe we can remove it.

core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java
```diff
@@ -103,7 +103,29 @@ class BaseUpdatePartitionSpec implements UpdatePartitionSpec {
...
+   * @return - the calculated field ID
+   */
+  private int assignFieldId(Pair<Integer, Transform<?, ?>> sourceTransform) {
+    if (formatVersion == 2 && base != null) {
+      int sourceId = sourceTransform.first();
+      Transform<?, ?> transform = sourceTransform.second();
+
+      Set<PartitionField> allHistoricalFields = Sets.newHashSet();
+      for (PartitionSpec partitionSpecpec : base.specs()) {
```
szehon-ho, line 118: Typo 'partitionSpecpec'

szehon-ho (commented):
This change looks like it's the right direction for me.

Is there any way to validate the data scan is correct filtering out the right files in this scenario (remove and add back partition filter), maybe by adding a test in "TestDataTableScan"?

In regards to your question, yes I agree the former seems the correct name to me and seems we should try to align it, but I don't really have much context and  maybe @aokolnychyi , @rdblue  or @RussellSpitzer have more thoughts about that.

core/src/test/java/org/apache/iceberg/TestDataTableScan.java
```diff
@@ -130,4 +132,63 @@ public void testTableScanWithPlanExecutor() {
     Assert.assertEquals(2, Iterables.size(scan.planFiles()));
     Assert.assertTrue("Thread should be created in provided pool", planThreadsIndex.get() > 0);
   }
+
+  @Test
+  public void testReaddingPartitionField() throws Exception {
```
szehon-ho, line 137: typo: "readding" => "reading"

core/src/test/java/org/apache/iceberg/TestDataTableScan.java
```diff
@@ -130,4 +132,63 @@ public void testTableScanWithPlanExecutor() {
...
+    Assume.assumeTrue(formatVersion == 2);
```
szehon-ho, line 138: What happens with v1?

szehon-ho (commented):
This looks better to me, just some minor comments and some test failures.

core/src/test/java/org/apache/iceberg/TestDataTableScan.java
szlta: That was supposed to be re-adding, I made it camelCase now to be less ambiguous.

core/src/test/java/org/apache/iceberg/TestDataTableScan.java
szlta: The change itself only affects V2. Since in V1 the reused partition field names always get appended by the ID, the fix is a no-op there. Another reason for this assume() call is that expected file structure differs greatly in V1 due to the void transformations that deleted partition fields get converted to, the test code would be quite bloated :/

szlta: Right. Btw I spotted an unnecessary if statement in your example: https://github.com/apache/iceberg/blob/master/core/src/test/java/org/apache/iceberg/TestMetadataTableFilters.java#L385
I guess all these could be cleaned up in another PR.

szlta:
@aokolnychyi , @rdblue or @RussellSpitzer could you please share your opinion on https://github.com/apache/iceberg/pull/4662#pullrequestreview-974341996 aka on the discrepancy in generating PartitionField names?

pvary:
Since the decision on #4868 is to stick with the old naming convention of default partition names, I think this is the best solution here. Please call out if you disagree, otherwise I will merge this soon.

Thanks,
Peter

pvary:
Thanks for the fix @szlta and the @szehon-ho and @rdblue for the review!

PR #4663, opened 2022-04-28
Spark: Fix NPEs in spark value converter

edgarRd:
This PR adds tests for `SparkValueConverter` across spark versions and fixed several NPE instances when using `null` in a Spark row converting to an Iceberg record using different schemas, including:
- Spark 2.4 with Scala 2.11 fails to unwrap null values for maps and lists
- Spark 3.1 is affected by [SPARK-37654](https://issues.apache.org/jira/browse/SPARK-37654); this should be fixed in 3.1.3 but that requires a version upgrade
- All versions fail to handle null structs within a row


spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/SparkValueConverter.java
```diff
@@ -58,20 +58,32 @@ public static Object convert(Type type, Object object) {
...
+        try {
+          for (Map.Entry<?, ?> entry : map.entrySet()) {
+            convertedMap.put(
+                convert(type.asMapType().keyType(), entry.getKey()),
+                convert(type.asMapType().valueType(), entry.getValue()));
+          }
+          return convertedMap;
+        } catch (NullPointerException npe) {
+          // Scala 2.11 fix: Catch NPE as internal value could be null and scala wrapper does not
+          // evaluate until iteration.
```
rdblue, line 84: Can you explain this a bit more? I don't understand what is happening in Scala 2.11. Is it that the object returned by `map.entrySet()` is non-null, but calling `iterator` on it will result in NPE?

Is there an alternative to this? If the value returned by `Iterable.iterator()` is null, can we catch that directly rather than using a catch block?

spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/SparkValueConverter.java
```diff
@@ -95,6 +107,10 @@ public static Object convert(Type type, Object object) {
   }
 
   private static Record convert(Types.StructType struct, Row row) {
+    if (row == null) {
+      return null;
+    }
```
rdblue, line 112: This looks reasonable to me.

rdblue (approved):
Overall this looks fine, but I'd like to avoid catching NPE if possible.

spark/v2.4/spark/src/main/java/org/apache/iceberg/spark/SparkValueConverter.java
edgarRd: Unfortunately, I don't think we can access the underlying map that is `null`. The line throwing the NPE is https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/convert/Wrappers.scala#L184 - here the `underlying.iterator` throws NPE as underlying is `null`, similar to any other Map method call that uses `underlying` such as `Map#size()`.
In fact, `Map#entrySet` returns a valid anonymous set but when the `iterator()` method is called the NPE is raised.

I wonder if with reflection you can check for nullity on `underlying` but I think that's no better approach. Open to suggestions. Thanks!

rdblue (approved):

rdblue:
Thanks, @edgarRd! I'll mark this for inclusion in 0.13.2 as well.

PR #4667, opened 2022-04-29
Spark: Add explicit deps to parquet-column and parquet-hadoop deps in spark …

stevenzwu:
…3.x modules and exclude transitive parquet deps from spark-hive. This way it is easier for have consistent parquet versions across iceberg modules.

I am excluding 2.4 for this deps cleanup.


rdblue (approved):
Looks reasonable to me.

PR #4668, opened 2022-04-29
Core - Add MetadataUpdateParser implementation for AddPartitionSpec and AddSortOrder

kbendick:
Adds toJson / fromJson implementations for two more implementations of `MetadataUpdate`:
1. `AddPartitionSpec`
2. `AddSortOrder`

Following up on PR feedback from https://github.com/apache/iceberg/pull/4632, this PR also renames the `writeAs***` and `readAs***` by dorpping the `As` so that the functions read better. Also, the action constants have been made package-private so they can be shared with the tests.

I've also placed `AddPartitionSpec` into the switch statement based on its declaration order in the `MetadataUpdate` file. As it was skipped in the last PR, I moved it down so it's now moved back up.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
@@ -37,21 +37,21 @@ public class MetadataUpdateParser {
 
   private static final String ACTION = "action";
 
-  // action types
-  private static final String ASSIGN_UUID = "assign-uuid";
-  private static final String UPGRADE_FORMAT_VERSION = "upgrade-format-version";
-  private static final String ADD_SCHEMA = "add-schema";
-  private static final String SET_CURRENT_SCHEMA = "set-current-schema";
-  private static final String ADD_PARTITION_SPEC = "add-spec";
-  private static final String SET_DEFAULT_PARTITION_SPEC = "set-default-spec";
-  private static final String ADD_SORT_ORDER = "add-sort-order";
-  private static final String SET_DEFAULT_SORT_ORDER = "set-default-sort-order";
-  private static final String ADD_SNAPSHOT = "add-snapshot";
-  private static final String REMOVE_SNAPSHOTS = "remove-snapshots";
-  private static final String SET_SNAPSHOT_REF = "set-snapshot-ref";
-  private static final String SET_PROPERTIES = "set-properties";
-  private static final String REMOVE_PROPERTIES = "remove-properties";
-  private static final String SET_LOCATION = "set-location";
+  // action types - visible for testing
+  static final String ASSIGN_UUID = "assign-uuid";
+  static final String UPGRADE_FORMAT_VERSION = "upgrade-format-version";
+  static final String ADD_SCHEMA = "add-schema";
+  static final String SET_CURRENT_SCHEMA = "set-current-schema";
+  static final String ADD_PARTITION_SPEC = "add-spec";
+  static final String SET_DEFAULT_PARTITION_SPEC = "set-default-spec";
+  static final String ADD_SORT_ORDER = "add-sort-order";
+  static final String SET_DEFAULT_SORT_ORDER = "set-default-sort-order";
+  static final String ADD_SNAPSHOT = "add-snapshot";
+  static final String REMOVE_SNAPSHOTS = "remove-snapshots";
+  static final String SET_SNAPSHOT_REF = "set-snapshot-ref";
+  static final String SET_PROPERTIES = "set-properties";
+  static final String REMOVE_PROPERTIES = "remove-properties";
+  static final String SET_LOCATION = "set-location";
 
   // UpgradeFormatVersion
   private static final String FORMAT_VERSION = "format-version";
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -116,19 +119,23 @@ public static void toJson(MetadataUpdate metadataUpdate, JsonGenerator generator
...
+        writeSetCurrentSchema((MetadataUpdate.SetCurrentSchema) metadataUpdate, generator);
+        break;
+      case ADD_PARTITION_SPEC:
+        writeAddPartitionSpec((MetadataUpdate.AddPartitionSpec) metadataUpdate, generator);
         break;
       case SET_DEFAULT_PARTITION_SPEC:
-        writeAsSetDefaultPartitionSpec((MetadataUpdate.SetDefaultPartitionSpec) metadataUpdate, generator);
+        writeSetDefaultPartitionSpec((MetadataUpdate.SetDefaultPartitionSpec) metadataUpdate, generator);
         break;
-      case ADD_PARTITION_SPEC:
```
kbendick, line 130: This is where `ADD_PARTITION_SPEC` was moved up to match the declaration order.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -197,4 +315,55 @@ private static void assertEqualsSetDefaultPartitionSpec(
...
+        .forEachOrdered(i -> {
+            UnboundPartitionSpec.UnboundPartitionField expectedField = expected.spec().fields().get(i);
+            UnboundPartitionSpec.UnboundPartitionField actualField = actual.spec().fields().get(i);
+            Assert.assertTrue(
+                "Fields of the unbound partition spec should be the same",
+                Objects.equals(expectedField.partitionId(), actualField.partitionId()) &&
+                    expectedField.name().equals(actualField.name()) &&
+                    Objects.equals(expectedField.transformAsString(), actualField.transformAsString()) &&
+                    expectedField.sourceId() == actualField.sourceId());
+        });
```
kbendick, line 339: `UnboundPartitionField` does not implement equals or hashCode. It also doesn't actually inherit from `PartitionField`, so none of the methods for checking if they're the same can be used.

Given the complexity of this code, it might make more sense to simply implement `equals` for this class. Open to feedback on this.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -197,4 +315,55 @@ private static void assertEqualsSetDefaultPartitionSpec(
...
+    IntStream.range(0, expected.sortOrder().fields().size())
+        .forEachOrdered(i -> {
+            UnboundSortOrder.UnboundSortField expectedField = expected.sortOrder().fields().get(i);
+            UnboundSortOrder.UnboundSortField actualField = actual.sortOrder().fields().get(i);
+            Assert.assertTrue("Fields of the sort order should be the same",
+                expectedField.sourceId() == actualField.sourceId() &&
+                expectedField.nullOrder().equals(actualField.nullOrder()) &&
+                expectedField.direction().equals(actualField.direction()) &&
+                Objects.equals(expectedField.transformAsString(), actualField.transformAsString()));
+        });
```
kbendick, line 367: Same note as above about potnetially adding `equals` to `UnboundSortField`.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -197,4 +314,55 @@ private static void assertEqualsSetDefaultPartitionSpec(
...
+    Assert.assertEquals("Unbound partition specs should have the same number of fields",
+        expected.spec().fields().size(), actual.spec().fields().size());
+
+    // Sort the fields so that we can compare them field by field
+    expected.spec().fields().sort(Comparator.comparing(UnboundPartitionSpec.UnboundPartitionField::transformAsString));
+    actual.spec().fields().sort(Comparator.comparing(UnboundPartitionSpec.UnboundPartitionField::transformAsString));
+    IntStream.range(0, expected.spec().fields().size())
+        .forEachOrdered(i -> {
+          UnboundPartitionSpec.UnboundPartitionField expectedField = expected.spec().fields().get(i);
+          UnboundPartitionSpec.UnboundPartitionField actualField = actual.spec().fields().get(i);
```
rdblue, line 331: I don't think these need to be sorted, but otherwise this looks good.

rdblue, line 331: I don't think these need to be sorted, but otherwise this looks good.

rdblue (changes requested):
I think this looks good other than sorting the partition spec fields and the sort order fields.

kbendick:
> I think this looks good other than sorting the partition spec fields and the sort order fields.

Removed the sorting on fields in the unbound partition spec and sort order.

kbendick:
I removed the ordering of fields from the test. This should be good to go in maybe 10 more minutes once the last of the tests finish @rdblue.

rdblue (approved):

PR #4669, opened 2022-04-29
Core: Fix ConvertEqualityDeleteStrategy::options return type

liuml07:
After reading the [design doc](https://docs.google.com/document/d/1-EyKSfwd_W9iI5jrzAvomVw3w1mb_kayVNT7f2I-SUg) I think the `ConvertEqualityDeleteStrategy::options` should return `ConvertEqualityDeleteStrategy` instead of the `RewritePositionDeleteStrategy` type. I guess this was an omission unless I miss something.


szehon-ho (approved):
Yea looks like a typo.  Not sure if options() is needed now as the RewritePositionDeleteStrategy doesn't have it, but I guess it's fine as it's better than today.

szehon-ho:
THanks @liuml07 for the change

PR #4670, opened 2022-04-29
Flink: add job level upsert config for table sink

chenjunjiedada:
This adds the option to the Flink SQL job so that users could specify the write mode. 

We have a table property to indicate the upsert mode for table level, while I think that is not the exact case since we have rewrite jobs in most cases. Also, we might have multiple writers to write in different branches/partitions.  So I would suggest removing that property as well.

### DIFF
--- a/flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
+++ b/flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
@@ -42,6 +42,7 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.util.PropertyUtil;
 
 public class FlinkDynamicTableFactory implements DynamicTableSinkFactory, DynamicTableSourceFactory {
   static final String FACTORY_IDENTIFIER = "iceberg";
@@ -113,7 +114,11 @@ public class FlinkDynamicTableFactory implements DynamicTableSinkFactory, Dynami
           objectPath.getObjectName());
     }
 
-    return new IcebergTableSink(tableLoader, tableSchema);
+    boolean upsert = PropertyUtil.propertyAsBoolean(tableProps,
+        FlinkConfigOptions.TABLE_EXEC_ICEBERG_WRITE_UPSERT_ENABLED.key(),
+        FlinkConfigOptions.SINK_TABLE_WRITE_UPSERT_ENABLED_DEFAULT);
+
+    return new IcebergTableSink(tableLoader, tableSchema, upsert);
   }
 
   @Override
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.util.List;
 import org.apache.flink.types.Row;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.iceberg.AssertHelpers;
 import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.Snapshot;
@@ -133,7 +134,7 @@ public class TestChangeLogTable extends ChangeLogTableTestBase {
     );
 
     testSqlChangeLog(TABLE_NAME, ImmutableList.of("id"), inputRowsPerCheckpoint,
-        expectedRecordsPerCheckpoint);
+        expectedRecordsPerCheckpoint, false);
   }
 
   @Test
@@ -164,7 +165,7 @@ public class TestChangeLogTable extends ChangeLogTableTestBase {
         ImmutableList.of(insertRow(1, "aaa"), insertRow(1, "ccc"), insertRow(2, "aaa"), insertRow(2, "ccc"))
     );
 
-    testSqlChangeLog(TABLE_NAME, ImmutableList.of("data"), elementsPerCheckpoint, expectedRecords);
+    testSqlChangeLog(TABLE_NAME, ImmutableList.of("data"), elementsPerCheckpoint, expectedRecords, false);
   }
 
   @Test
@@ -194,7 +195,7 @@ public class TestChangeLogTable extends ChangeLogTableTestBase {
         ImmutableList.of(insertRow(1, "aaa"), insertRow(1, "ccc"), insertRow(2, "aaa"), insertRow(2, "bbb"))
     );
 
-    testSqlChangeLog(TABLE_NAME, ImmutableList.of("data", "id"), elementsPerCheckpoint, expectedRecords);
+    testSqlChangeLog(TABLE_NAME, ImmutableList.of("data", "id"), elementsPerCheckpoint, expectedRecords, false);
   }
 
   @Test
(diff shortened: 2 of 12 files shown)


rdblue:
@kbendick, can you review this?

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
```diff
@@ -235,17 +236,59 @@ public void testPureInsertOnIdKey() throws Exception {
         )
     );
 
-    testSqlChangeLog(TABLE_NAME, ImmutableList.of("data"), elementsPerCheckpoint, expectedRecords);
+    testSqlChangeLog(TABLE_NAME, ImmutableList.of("data"), elementsPerCheckpoint, expectedRecords, false);
+  }
+
+  @Test
+  public void testUpsertOnIdKey() throws Exception {
```
rdblue, line 243: What happens when the table doesn't have its identifier field IDs set?

flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
```diff
@@ -71,6 +72,12 @@ public class FlinkDynamicTableFactory implements DynamicTableSinkFactory, Dynami
...
 
+  private static final ConfigOption<Boolean> WRITE_UPSERT_ENABLED =
+      ConfigOptions.key("write-upsert-enabled")
+          .booleanType()
+          .defaultValue(false)
+          .withDescription("Whether to transform all INSERT/UPDATE_AFTER events to UPSERT.");
+
   // Flink 1.13.x change the return type from CatalogTable interface to ResolvedCatalogTable which extends the
   // CatalogTable. Here we use the dynamic method loading approach to avoid adding explicit CatalogTable or
   // ResolvedCatalogTable class into the iceberg-flink-runtime jar for compatibility purpose.
```
kbendick, line 83: Unrelated, but this dynamic function call can probably be removed now that we have code per major Flink version (unless this change was made in a patch release after 1.13.0).

I think @hililiwei has a PR that removes it iirc, but that PR also tackles much larger issues. @hililiwei can you possibly open a PR to just update the no-longer needed dynamic `getCatalogTable` call? I tried looking for the PR but couldn't find it.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import org.apache.flink.types.Row;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.iceberg.AssertHelpers;
```
stevenzwu, line 27: why is the unit test change not ported to 1.13?

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
chenjunjiedada: In `FlinkSink` class, `checkAndGetEqualityFieldIds` validates the equality fields. It throws an exception when building an upsert sink without being given a primary key.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
chenjunjiedada: I'm not sure whether we should keep developing on the latest version, so I was hesitant to add changes to the 1.13 version, not to mention we are developing 1.15 now. But there are some production CDC cases still base on Flink 1.13, let me port them anyway.

stevenzwu: I saw you ported the main src to 1.13 and not the unit test. Hence I was asking. I am fine if we don't update 1.13 at all.

kbendick: I'd be fine with not adding for 1.13 in this PR (and backporting it once review is done).

flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
```diff
@@ -126,7 +133,9 @@ public DynamicTableSink createDynamicTableSink(Context context) {
           objectPath.getObjectName());
     }
 
-    return new IcebergTableSink(tableLoader, tableSchema);
+    boolean upsert = PropertyUtil.propertyAsBoolean(tableProps, WRITE_UPSERT_ENABLED.key(), false);
```
kbendick, line 136: Can we add a constant for the default value here, similar to `TableProperties.UPSERT_ENABLED_DEFAULT`?

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
```diff
@@ -235,17 +236,59 @@ public void testPureInsertOnIdKey() throws Exception {
...
+    );
+
+    if (!partitioned) {
+      testSqlChangeLog(TABLE_NAME, ImmutableList.of("id"), elementsPerCheckpoint, expectedRecords, true);
+    } else {
+      AssertHelpers.assertThrows("Should be error because equality field columns don't include all partition keys",
+          IllegalStateException.class, "should be included in equality fields",
+          () -> {
+            testSqlChangeLog(TABLE_NAME, ImmutableList.of("id"), elementsPerCheckpoint, expectedRecords, true);
+            return null;
```
kbendick, line 278: Nit: I don't think that the `return null` is needed here.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
```diff
@@ -235,17 +236,59 @@ public void testPureInsertOnIdKey() throws Exception {
...
   }
 
-  private Table createTable(String tableName, List<String> key, boolean isPartitioned) {
+  private Table createTable(String tableName, List<String> key, boolean isPartitioned, boolean upsertEnabled) {
     String partitionByCause = isPartitioned ? "PARTITIONED BY (data)" : "";
-    sql("CREATE TABLE %s(id INT, data VARCHAR, PRIMARY KEY(%s) NOT ENFORCED) %s",
-        tableName, Joiner.on(',').join(key), partitionByCause);
+    sql("CREATE TABLE %s(id INT, data VARCHAR, PRIMARY KEY(%s) NOT ENFORCED) %s " +
+            "WITH ('write-upsert-enabled'='%s')",
+        tableName, Joiner.on(',').join(key), partitionByCause, upsertEnabled ? "true" : "false");
```
kbendick, line 291: It doesn't seem like we're testing if the job level config is being respected. Just that the table level configuration is being used now.

Can we add a test to enure that the job level config is respected even if the table's `write-upsert-enabled` isn't set (or is explicitly set to false assuming that we want job level configs to override)? If I missed the test that checks the new `ConfigOption`, let me know.

flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
chenjunjiedada: Will do.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
chenjunjiedada: The `write-upsert-enabled` is the option of the Flink logic table and the `write.upsert.enabled` is the property of the iceberg table. Right now in this unit test, it creates the table via Flink catalog and thus respects the job level option.

chenjunjiedada: Also, I would suggest removing table level configuration as I mentioned before since in real cases we should always have offline service to optimize tables which means two writers with different writing models.

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
chenjunjiedada: `AssertThrows` can assert `Runnable` and `Callable`. `Runnable` is used when we expect runtime exceptions and `Callable` is used when we expect checked exceptions. `return null` is used so the lambda is treated as a `callable`.  

If we remove `return null`, the lambda is treated as `Runnable` thus we have to use `try/catch` to catch the exception and wrap it into a runtime exception and unwrap outside.

stevenzwu (approved):

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/TestChangeLogTable.java
kbendick: Ok. That makes sense. Thanks for clarifying for me.

kbendick (approved):
+1 this LGTM.

I'm not sure if we can revert the table level configuration (as mentioned in the PR description) due to backwards compatibility concerns. But this looks good to me.

flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
hililiwei: you mean here?
https://github.com/apache/iceberg/pull/4246#discussion_r817304711

https://github.com/apache/iceberg/pull/4246

chenjunjiedada:
@kbendick, Do you want to merge this first or port 1.13 changes in?

kbendick:
> @kbendick, Do you want to merge this first or port 1.13 changes in?

I’m ok with it being merged now (that’s what we normally do). But I don’t have permissions to merge.

If you can get 1.13 working without issues, I’d suggest maybe putting it in this PR so that it’s merged sooner as well.

I’ll try to get somebody to merge this as soon as possible.

chenjunjiedada:
@kbendick， Now 1.13 is in. Do we still need port 1.15 as well?

chenjunjiedada:
@kbendick @rdblue , Can we include this in the 0.13.2 release?

chenjunjiedada:
@kbendick @rdblue , Port to Flink 1.15 now, PTAL.

chenjunjiedada:
Ping @kbendick @rdblue, another two weeks passed.

PR #4169, opened 2022-02-18
Add logging when optimizing delete to metadata-only-delete

szehon-ho:
It's not easy for users to know whether they are writing deletes that can be performed with just metadata, so adding this log line.

Example output from a unit test:
```
org.apache.spark.sql.execution.datasources.v2.OptimizeMetadataOnlyDeleteFromTable - Optimizing delete expression: IsNull(dep) as metadata delete
```

PR #4672, opened 2022-04-29
API: Add expression sanitizer, sanitize scan log messages

rdblue:
This adds an expression sanitizer and sanitizes filters logged by table scans. This is to avoid leaking IDs and other sensitive information in logs.

Literals are sanitized using the following rules:
* Integers and longs are converted to `(N-digit-int)`, where `N` is the number of digits in the number
* Floats and doubles are converted to `(N-digit-float)`, where N is the number of digits before the decimal point
* Dates, times, and timestamps are converted to `(date)`, `(time)`, or `(timestamp)`
* All other types are hashed to produce `(hash-%08h)` from the string representation

The result for string literals matches the result after the literals are bound and converted.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
@@ -1,0 +1,257 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.expressions;
+
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import org.apache.iceberg.transforms.Transform;
+import org.apache.iceberg.transforms.Transforms;
+import org.apache.iceberg.types.Types;
+
+/**
+ * Expression utility methods.
+ */
+public class ExpressionUtil {
+  private static final Transform<CharSequence, Integer> HASH_FUNC = Transforms
+      .bucket(Types.StringType.get(), Integer.MAX_VALUE);
+  private static final Pattern DATE = Pattern.compile("\\d\\d\\d\\d-\\d\\d-\\d\\d");
+  private static final Pattern TIME = Pattern.compile(
+      "\\d\\d:\\d\\d(:\\d\\d(.\\d{1,6})?)?");
+  private static final Pattern TIMESTAMP = Pattern.compile(
+      "\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d(:\\d\\d(.\\d{1,6})?)?([-+]\\d\\d:\\d\\d)?");
+
+  private ExpressionUtil() {
+  }
+
+  /**
+   * Produces an unbound {@link Expression} with the same structure, but with data values replaced by descriptions.
+   * <p>
+   * Numbers are replaced with magnitude and type, string-like values are replaced by hashes, and date/time values are
+   * replaced by the type.
+   *
+   * @param expr an Expression to sanitize
+   * @return a sanitized Expression
+   */
+  public static Expression sanitize(Expression expr) {
+    return ExpressionVisitors.visit(expr, ExpressionSanitizer.INSTANCE);
+  }
+
+  /**
+   * Produces a sanitized expression string with the same structure, but with data values replaced by descriptions.
+   * <p>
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * Produces a sanitized expression string with the same structure, but with data values replaced by descriptions.
+   * <p>
+   * Numbers are replaced with magnitude and type, string-like values are replaced by hashes, and date/time values are
+   * replaced by the type.
+   *
+   * @param expr an Expression to sanitize
+   * @return a sanitized expression string
+   */
+  public static String toSanitizedString(Expression expr) {
```
kbendick, line 65: As the specific needs for sanitization are often very business specific, would it make sense to leave this somehow as something that end users could override somewhat easily?

Possibly the `santitize` method itself could be abstract with this as the default implementation.

kbendick (approved):
This looks good to me overall, +1 from me once tests are passing.

I'm wondering if we need to be this aggressive about sanitizing dates and times. Possibly a small portion of it could be left in?

I do recognize that determining the level of sanitization required is usually a business specific issue, so it's likely better to over-sanitize than under-sanitize.

Is this something we should provide a configuration to enable / disable, possibly within the engines somehow? In many cases, these values are not sensitive, or not sensitive enough to be a concern for people's logs. And they might want to disable this for debugging purposes.

This could all be addressed in a follow up task, of course.

kbendick:
The test failure seems to be a transient issue.

If I close this PR though, I'm not sure if I can reopen it.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: We can improve this as needed, but I don't want to guess ahead of time what makes sense. People can always extend this later.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
```diff
@@ -0,0 +1,257 @@
...
+    return ExpressionVisitors.visit(expr, StringSanitizer.INSTANCE);
+  }
+
+  private static class ExpressionSanitizer extends ExpressionVisitors.ExpressionVisitor<Expression> {
+    private static final ExpressionSanitizer INSTANCE = new ExpressionSanitizer();
+
+    @Override
+    public Expression alwaysTrue() {
+      return Expressions.alwaysTrue();
```
szehon-ho, line 74: For another pr, but might be nice to have these boilerplate codes (and the string ones) in base class at some point to reduce chance of errors

szehon-ho (commented):
Thanks there is one more (that I added, that could be a bit sensitive):  https://github.com/apache/iceberg/pull/4169/files , if it can added to this as well?

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: The default is to return `null`, which I think is a reasonable thing to do. We'd probably have to change a lot of the implementations so I'm not sure if it is worth it.

rdblue:
@szehon-ho, I think we'll need to address the log in #4169 separately because that's entirely in Spark code and we don't have an Iceberg filter. We'd need to either add the ability to sanitize Spark filters (which I'm reluctant to add) or just allow it for Spark. This PR changes instances in core, which gets used in other engines as well.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
```diff
@@ -0,0 +1,257 @@
...
+    } else if (literal instanceof Literals.DateLiteral) {
+      return "(date)";
+    } else if (literal instanceof Literals.TimeLiteral) {
+      return "(time)";
+    } else if (literal instanceof Literals.TimestampLiteral) {
+      return "(timestamp)";
+    } else if (literal instanceof Literals.IntegerLiteral) {
+      return sanitizeNumber(((Literals.IntegerLiteral) literal).value(), "int");
+    } else if (literal instanceof Literals.LongLiteral) {
+      return sanitizeNumber(((Literals.LongLiteral) literal).value(), "int");
```
szehon-ho, line 237: why not long?  (and below, why not double?)

szehon-ho (commented):
Fair enough, maybe we can use SparkUtils redact methods for that, can discuss in another issue not to block this pr.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: Good catch! I used `int` or `float` in both cases because the type is determined by what you pass in to create the literal, not the actual type of data in a table (because the expression is unbound). Since the types are floating anyway, I didn't think it made much sense to be strict. What really matters is the magnitude of the value. If you have a `4-digit-int` it doesn't matter if it was passed in as a long or passed in as an int.

szehon-ho: I see, though is there any disadvantage to just putting the exact one being passed in ?  I took a look and seems Spark does some schema-based binding of expression before passing it to Iceberg, so a comparison of long column results in long literal being passed down to Iceberg scan.   Though understand its not a huge point.

rdblue: I think that would require more processing later to use these sanitized values. Right now, this sanitizes log messages, but if you wanted to start collecting sanitized filters and doing analysis on them, one of the first steps would be to collapse these, I think. That's why I went ahead and normalized.

rdblue:
Thanks, @szehon-ho!

PR #4673, opened 2022-04-29
Core: Fix table corruption from OOM during commit cleanup in Spark

### LINKED ISSUE #4666: Iceberg commit flow can abort and delete data files exposed to readers
During the WAP flow, we are seeing staged snapshots with aborted data files being committed to table active snapshots line. This leaves the table unreadable with scans failing with java.io.FileNotFoundException. 

If there is a fatal error in the system during the following portion of commit procedure Iceberg can end up promoting invalid snapshot to the active snapshots line: https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L328-L346

If there is a failure anywhere in the above snippet  iceberg aborts and deletes the data under that snapshot here https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L215

Any other process is now allowed to cherrypick this snapshot and read from the table and will fail with java.io.FileNotFoundException. The table becomes unreadable. 

**Update:** This applies to non-WAP flow as well. Even regular commit flow can end up with deleted data in master if fatal error occurs during above code snippet.

**Sequence of events to reproduce:**

T1: Create table with Write-Audit-Publish enabled 
T2: Write some data to it with a wap.id=B1 using Snapshot S1.  During the commit execution there is a fatal error anywhere during this code snippet https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L328-L346
T3: Above Commit failure triggers an abort and deletes data under S1. This leaves the Snapshot S1 in the staged snapshot list but hasn't been cherry-picked.
T4: A different worker tries to write data with wap.id=B1 using S2. 
T5: After validation during this worker filters table.snapshots() to find S1 which has the same wap.id and cherrypicks this to add to table.
T6: S1 gets published to active line. 
T7: Reading table fails with FNF since S1 has no data files.

rdblue:
Fixes #4666.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/SnapshotProducer.java
+++ b/core/src/main/java/org/apache/iceberg/SnapshotProducer.java
@@ -323,9 +323,9 @@ abstract class SnapshotProducer<ThisT> implements SnapshotUpdate<ThisT> {
       Exceptions.suppressAndThrow(e, this::cleanAll);
     }
 
-    LOG.info("Committed snapshot {} ({})", newSnapshotId.get(), getClass().getSimpleName());
-
     try {
+      LOG.info("Committed snapshot {} ({})", newSnapshotId.get(), getClass().getSimpleName());
+
       // at this point, the commit must have succeeded. after a refresh, the snapshot is loaded by
       // id in case another commit was added between this commit and the refresh.
       Snapshot saved = ops.refresh().snapshot(newSnapshotId.get());
@@ -343,11 +343,15 @@ abstract class SnapshotProducer<ThisT> implements SnapshotUpdate<ThisT> {
         LOG.warn("Failed to load committed snapshot, skipping manifest clean-up");
       }
 
-    } catch (RuntimeException e) {
-      LOG.warn("Failed to load committed table metadata, skipping manifest clean-up", e);
+    } catch (Throwable e) {
+      LOG.warn("Failed to load committed table metadata or during cleanup, skipping further cleanup", e);
     }
 
-    notifyListeners();
+    try {
+      notifyListeners();
+    } catch (Throwable e) {
+      LOG.warn("Failed to notify event listeners", e);
+    }
   }
 
   private void notifyListeners() {


dilipbiswal:
LGTM

RussellSpitzer (approved):
Lgtm

dilipbiswal:
@rdblue @RussellSpitzer 
Should we move https://github.com/apache/iceberg/blob/17500536d714d32b4e6dfc91dd737edc4046fed1/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L326 into the following try block as a safe measure ?

rdblue:
@dilipbiswal, done.

dilipbiswal:
Thanks a lot @rdblue

dilipbiswal (approved):

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -343,11 +343,15 @@ public void commit() {
         LOG.warn("Failed to load committed snapshot, skipping manifest clean-up");
       }
 
-    } catch (RuntimeException e) {
-      LOG.warn("Failed to load committed table metadata, skipping manifest clean-up", e);
+    } catch (Throwable e) {
```
stevenzwu, line 346: `cleanAll` is called in line 323 before this try-catch block. what is the further cleanup after this block? I only see `notifyListeners`.

stevenzwu: If we are talking about `OutOfMemoryError` happened during the first/commit try-catch block (line 286-324), then `cleanAll` won't be called either as `OutOfMemoryError` is not `RuntimeException`.

rdblue: `cleanAll` is called when the commit fails. If the commit fails, then there is no problem with Spark cleaning up the data files, or cleaning the metadata files that Iceberg creates.

stevenzwu: if there is an `OutOfMemoryError` throw from the try block (line 326-346), this change will catch and swallow it. So it skip further cleanup from line 333-339. I am not following how this fix the issue #4666 where committed data files were deleted.

RussellSpitzer: The issue was that a throwable not caught by the catch block here would be thrown to Spark, which would cause Spark to consider the commit failed and cause it to perform it's own abort.

So the old behavior was
1. Table Operations Commit (this can happen successfully)
2. While Cleaning up old files or theoretically while calling notify listeners an exception is thrown
3. The commit has been successful but the exception is re-thrown to the Spark Commit code
4. The Spark commit code sees the exception and executes it's abort method

So the fix is basically never ever throw an exception once the commit has occurred, regardless of what happens

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -343,11 +343,15 @@ public void commit() {
...
+      LOG.warn("Failed to load committed table metadata or during cleanup, skipping further cleanup", e);
```
RussellSpitzer, line 347: Failed loading committed table metadata or during cleanup

or

Failed to load committed table metadata or failed during cleanup

StefanXiepj:
This pr looks like copied from [Core: Skipping manifest clean-up for all Error or Exception. #4507](https://github.com/apache/iceberg/pull/4507) , @rdblue  why not continue on the original pr and copied them into new pr ? I think we should encourage more contributors to join us.

@dilipbiswal @RussellSpitzer @stevenzwu @electrum cc, pls

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
stevenzwu: @RussellSpitzer thanks a lot for the detailed explanation. Didn't realize it is Spark abort flow deleted the data files. 

I am uneasy with swallowing fatal errors (like OutOfMemoryError) though. Should Spark only catch `CommitFailedException` and perform abort?

RussellSpitzer: @stevenzwu It's not in our code, it's in Spark's code where it doesn't have a defined exception for either of those states. Apache Spark basically has this code
```java
try {
  DataSource.commit() 
  } catch (anything) {
  datasource.abort()
  }
```
[source](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/WriteToDataSourceV2Exec.scala#L392-L400)  


Spark doesn't know about any of our particular exceptions, it just wants to know if it's call to commit failed or not. In this particular instance the worry is that the commit did succeed and no exceptions were thrown but additional code ran after the commit (like our clean up code or notify listeners) which throws another exception. 

Spark itself cannot distinguish between an exception that happened before the actual commit and after the actual commit within our DataSource commit method. If we wanted to let engines handle this we could have our Iceberg SparkWrite try to handle this but I think that would probably be difficult to manage as well. I think it more sense to say that once committed, we suppress all exceptions. Our Commit method in SnapshotProducer only throws exceptions when the commit fails and in *no* other circumstances.

stevenzwu: Got it. I thought it was iceberg-spark.

When `CommitStateUnknownException` is thrown, commit may have actually succeeded in the backend. If Spark performs abort and delete data files in this case, it could corrupt table state. could this happen or it is taken care of by iceberg-spark?

Sorry for asking some newbie Spark questions.

RussellSpitzer: I think that may be possible... All of our tests at the moment just make sure that the Metadata.json file is not incorrectly removed. We probably should add an additional test to make sure data files in Spark are not removed. I would think we would need a catch [here](https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L199) 

```
    try {
      operation.commit(); // abort is automatically called if this fails
    } catch (CommitStateUnknownException commitStateUnknownException) {
      LOG.warn("Unknown Commit State", commitStateUnknownException);
    }
```

stevenzwu: Catching and swallowing the `CommitStateUnknownException` is also not ideal. If the commit actually failed, that would lead Spark to assume the commit is successful.

Ideally, Spark needs to know these commit results and then iceberg-spark can translate Iceberg commit result to Spark commit result.
* commit success. treat it as success
* commit failure. treat it as failure and perform the abort
* commit state unknown. treat it as failure but don't perform the abort

This discussion is probably outside the scope of this PR. This PR is fine. Basically, it swallows any exceptions from post-commit-success cleanup steps.

stevenzwu (approved):

kbendick (approved):
LGTM.

rdblue:
Thanks for reviewing this, everyone! Good to have it fixed.

PR #4674, opened 2022-04-30
Spark-3.2: Avoid duplicate computation of ALL_MANIFESTS metadata table for spark actions

ajantha-bhat:
ALL_MANIFESTS computation is a heavy operation when lot of snapshots exist as it involves reading of the the manifest list file for every snapshot.
We are currently having duplicate computation of the same in many places in spark actions.

This PR aims to improve the performance of spark actions and stored procedures by computing once and reusing it in other locations using dataset.persist().

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
@@ -120,6 +120,9 @@ public class BaseDeleteOrphanFilesSparkAction
   private Consumer<String> deleteFunc = defaultDelete;
   private ExecutorService deleteExecutorService = null;
 
+  private static final String USE_CACHING = "use-caching";
+  private static final boolean USE_CACHING_DEFAULT = true;
+
   public BaseDeleteOrphanFilesSparkAction(SparkSession spark, Table table) {
     super(spark);
 
@@ -209,7 +212,11 @@ public class BaseDeleteOrphanFilesSparkAction
   }
 
   private DeleteOrphanFiles.Result doExecute() {
-    Dataset<Row> allManifests = loadMetadataTable(table, ALL_MANIFESTS).persist();
+    Dataset<Row> allManifests = loadMetadataTable(table, ALL_MANIFESTS);
+    boolean useCaching = PropertyUtil.propertyAsBoolean(options(), USE_CACHING, USE_CACHING_DEFAULT);
+    if (useCaching) {
+      allManifests.persist();
+    }
     Dataset<Row> validContentFileDF = buildValidContentFileDF(table, allManifests);
     Dataset<Row> validMetadataFileDF = buildValidMetadataFileDF(table, allManifests);
     Dataset<Row> validFileDF = validContentFileDF.union(validMetadataFileDF);
@@ -223,7 +230,9 @@ public class BaseDeleteOrphanFilesSparkAction
     List<String> orphanFiles = actualFileDF.join(validFileDF, joinCond, "leftanti")
         .as(Encoders.STRING())
         .collectAsList();
-    allManifests.unpersist();
+    if (useCaching) {
+      allManifests.unpersist();
+    }
     Tasks.foreach(orphanFiles)
         .noRetry()
         .executeWith(deleteExecutorService)
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
@@ -68,6 +68,7 @@ import org.apache.spark.util.SerializableConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.iceberg.MetadataTableType.ALL_MANIFESTS;
 import static org.apache.iceberg.TableProperties.GC_ENABLED;
 import static org.apache.iceberg.TableProperties.GC_ENABLED_DEFAULT;
 
@@ -208,8 +209,9 @@ public class BaseDeleteOrphanFilesSparkAction
   }
 
   private DeleteOrphanFiles.Result doExecute() {
-    Dataset<Row> validContentFileDF = buildValidContentFileDF(table);
-    Dataset<Row> validMetadataFileDF = buildValidMetadataFileDF(table);
+    Dataset<Row> allManifests = loadMetadataTable(table, ALL_MANIFESTS).persist();
+    Dataset<Row> validContentFileDF = buildValidContentFileDF(table, allManifests);
+    Dataset<Row> validMetadataFileDF = buildValidMetadataFileDF(table, allManifests);
     Dataset<Row> validFileDF = validContentFileDF.union(validMetadataFileDF);
     Dataset<Row> actualFileDF = compareToFileList == null ? buildActualFileDF() : filteredCompareToFileList();
 
@@ -221,7 +223,7 @@ public class BaseDeleteOrphanFilesSparkAction
     List<String> orphanFiles = actualFileDF.join(validFileDF, joinCond, "leftanti")
         .as(Encoders.STRING())
         .collectAsList();
-
+    allManifests.unpersist();
     Tasks.foreach(orphanFiles)
         .noRetry()
         .executeWith(deleteExecutorService)
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
@@ -71,6 +71,10 @@ public class BaseDeleteReachableFilesSparkAction
   private ExecutorService deleteExecutorService = null;
   private FileIO io = new HadoopFileIO(spark().sessionState().newHadoopConf());
   private Dataset<Row> allManifests = null;
+  private boolean useCaching = true;
+
+  private static final String USE_CACHING = "use-caching";
+  private static final boolean USE_CACHING_DEFAULT = true;
 
   public BaseDeleteReachableFilesSparkAction(SparkSession spark, String metadataFileLocation) {
     super(spark);
@@ -124,13 +128,19 @@ public class BaseDeleteReachableFilesSparkAction
     } else {
       result = deleteFiles(reachableFileDF.collectAsList().iterator());
     }
-    allManifests.unpersist();
+    if (useCaching) {
+      allManifests.unpersist();
+    }
     return result;
   }
 
   private Dataset<Row> buildReachableFileDF(TableMetadata metadata) {
     Table staticTable = newStaticTable(metadata, io);
-    allManifests = loadMetadataTable(staticTable, ALL_MANIFESTS).persist();
+    allManifests = loadMetadataTable(staticTable, ALL_MANIFESTS);
+    useCaching = PropertyUtil.propertyAsBoolean(options(), USE_CACHING, USE_CACHING_DEFAULT);
+    if (useCaching) {
+      allManifests.persist();
+    }
     return withFileType(buildValidContentFileDF(staticTable, allManifests), CONTENT_FILE)
         .union(withFileType(buildManifestFileDF(allManifests), MANIFEST))
         .union(withFileType(buildManifestListDF(staticTable), MANIFEST_LIST))
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
@@ -42,6 +42,7 @@ import org.apache.spark.sql.SparkSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.iceberg.MetadataTableType.ALL_MANIFESTS;
 import static org.apache.iceberg.TableProperties.GC_ENABLED;
 import static org.apache.iceberg.TableProperties.GC_ENABLED_DEFAULT;
 
@@ -69,6 +70,7 @@ public class BaseDeleteReachableFilesSparkAction
   private Consumer<String> deleteFunc = defaultDelete;
   private ExecutorService deleteExecutorService = null;
   private FileIO io = new HadoopFileIO(spark().sessionState().newHadoopConf());
+  private Dataset<Row> allManifests = null;
 
   public BaseDeleteReachableFilesSparkAction(SparkSession spark, String metadataFileLocation) {
     super(spark);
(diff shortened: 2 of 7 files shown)


ajantha-bhat:
cc: @RussellSpitzer , @rdblue

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
```diff
@@ -125,8 +126,9 @@ private Result doExecute() {
 
   private Dataset<Row> buildReachableFileDF(TableMetadata metadata) {
     Table staticTable = newStaticTable(metadata, io);
-    return withFileType(buildValidContentFileDF(staticTable), CONTENT_FILE)
-        .union(withFileType(buildManifestFileDF(staticTable), MANIFEST))
+    Dataset<Row> allManifests = loadMetadataTable(staticTable, ALL_MANIFESTS);
+    return withFileType(buildValidContentFileDF(staticTable, allManifests), CONTENT_FILE)
+        .union(withFileType(buildManifestFileDF(allManifests), MANIFEST))
```
RussellSpitzer, line 131: How does this change the performance? Don't we have to compute all manifests in both locations here? Or Does changing to an object let Spark cache the relation?

ajantha-bhat: https://github.com/apache/iceberg/blob/434c5658b26ff06f11da261d1d49a84f19e267d7/core/src/main/java/org/apache/iceberg/AllManifestsTable.java#L118

For example in expire_snapshots, as the ALL_MANIFESTS metadata tables computed twice, it has to read manifest file list twice for every snapshot. Now only once ALL_MANIFESTS metadata table is created and it is reused.

ajantha-bhat: `loadMetadataTable` was called twice, now it is only once for ALL_MANIFESTS.

ajantha-bhat: > Or Does changing to an object let Spark cache the relation?

I thought compute of dataset will happen in the first action and the results are reused for both the steps.

ajantha-bhat: @RussellSpitzer : I got what you meant now.

To solve this problem only way I have now is to collect the dataset as list of rows and prepare new dataset from list of rows and use that dataset in both location. This way, reading of manifest list happens only once. WDYT ? any other solution?

ajantha-bhat: fixed and verified the scanning by adding and checking logs.

ajantha-bhat: It seems, individual test cases passes. But when whole suit is ran, persist() is causing failures for hadoop catalog. Not sure why. I might have to explore more on this. Probably I need some job completion listener to unpersist().

RussellSpitzer: Ah yes I should have been more clear, I was referring to the fact that the dataset would be recomputed on both lines. The "loadMetadataTable" function should be very fast but the actual planning should be expensive and that would require a cache of some kind. 

I'm a little worried in general about persisting things since I want to make sure we clean up our caches asap.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -220,6 +222,7 @@ private DeleteOrphanFiles.Result doExecute() {
     Column joinCond = nameEqual.and(actualContains);
     List<String> orphanFiles = actualFileDF.join(validFileDF, joinCond, "leftanti")
         .as(Encoders.STRING())
+        .unpersist()
```
RussellSpitzer, line 225: I do not think this will trigger the unpersist on "allManifests"

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -169,7 +170,7 @@ public Dataset<Row> expire() {
       Dataset<Row> validFiles = buildValidFileDF(ops.refresh());
 
       // determine expired files
-      this.expiredFiles = originalFiles.except(validFiles);
+      this.expiredFiles = originalFiles.except(validFiles).unpersist();
```
RussellSpitzer, line 173: Again I believe if you want to get rid of the cache I do think you need to call that the allManifests table

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
ajantha-bhat: > Very much dependent on the computation leading up to the cached df

Because it involves the IO operation, it will definitely help when there are hundreds or thousands of snapshots.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
ajantha-bhat: yeah, I figured that out, but didn't figured out the final solution. So, changed to draft :) 

But as you said, I need to call allManifests table again I guess.

ajantha-bhat: @RussellSpitzer , which is better? a cache or manually calling dataset.collect() on that allManifest df and building new dataset on top of those collected rows and reusing it on both the location?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
ajantha-bhat: @RussellSpitzer , which is better? a cache or manually calling dataset.collect() on that allManifest df and building new dataset on top of those collected rows and reusing it on both the location?

RussellSpitzer: @ajantha-bhat The problem is that persisting does another IO serialization of what should be basically the same amount of information but hopefully in a more readable form. Persist by default is a Memory and Disk based cached. You really need to test it out to be sure.

For Cache vs Collect. Cache is probably much better, pieces would be stored on executors and the IO would hopefully be mostly local, doing a Collect and and building a DF from it would essentially bring all data back to the driver serialize, then deserialize and send everything back out. This is always worse than cache/persist

szehon-ho:
@RussellSpitzer  pointed me to this, I had a pr is orthogonal to this, to avoid duplicate computation of all_reachable_files here https://github.com/apache/iceberg/pull/3457/files. To me that was the bigger time consumer (exploring all reachable files), though maybe I need to re-do that pr.

Anyway, agree with @RussellSpitzer that maybe cache is a better option than persist (great to see benchmarks for tables with huge snapshots)?  And also it would be better to have it be configurable, and be able to be uncached as soon as possible.

szehon-ho:
Although I had a thought in #3457 to implement IncrementalScan for all_files table (to pass in from/to snapshots to filter), and then refactor the expire_snapshot action to use that.  That would somewhat conflict with this change, as it doesn't seem clean to pass in cached manifest_file df to the all_files table scan, so we would pick one or another.

ajantha-bhat:
@szehon-ho :

> @RussellSpitzer pointed me to this, I had a pr is orthogonal to this, to avoid duplicate computation of all_reachable_files here https://github.com/apache/iceberg/pull/3457 To me that was the bigger time consumer (exploring all reachable files), though maybe I need to re-do that pr. Wasn't sure how much bottleneck getting all_manifests was.

yeah, scanning the all_manifest table twice was the major problem for me. 


> Anyway, agree with @RussellSpitzer that maybe cache is a better option than persist? It'd be great to see some numbers for tables with huge snapshots for these two options vs today, if possible. I think if , if we go with this approach, it should probably be 1) configurable , 2) able to be GC'ed sooner than later.

Sure,  I will make it configurable option to cache or not and get the performance report locally with large number of snapshots.  I will work on this over this weekend.

szehon-ho:
Also FYI @aokolnychyi , if you wanted to take a look as well

ajantha-bhat:
@RussellSpitzer, @szehon-ho, @aokolnychyi:  PR is ready for review. 

a) I tested with local file system and local test case in IDE, without cache (22ms), with cache (18ms) with 1000 manifest list files to read. 
Not so much difference, probably because it is local file system. As PR avoids reading manifest list files again (verified with logs), in S3 file system probably better difference can be seen with huge number of snapshots.

```
@Test
  public void testExpireSnapshotUsingNamedArgs() {
    sql("CREATE TABLE %s (id bigint NOT NULL, data string) USING iceberg", tableName);

    for (int i = 0; i < 1000; i++) {
      sql("INSERT INTO TABLE %s VALUES (1, 'a')", tableName);
    }

    Table table = validationCatalog.loadTable(tableIdent);

    Assert.assertEquals("Should be 1000 snapshots", 1000, Iterables.size(table.snapshots()));

    waitUntilAfter(table.currentSnapshot().timestampMillis());

    Timestamp currentTimestamp = Timestamp.from(Instant.ofEpochMilli(System.currentTimeMillis()));

    long time = System.currentTimeMillis();

    List<Object[]> output = sql(
        "CALL %s.system.expire_snapshots(" +
            "older_than => TIMESTAMP '%s'," +
            "table => '%s'," +
            "retain_last => 999, use_caching => false)",
        catalogName, currentTimestamp, tableIdent);

    System.out.println("time taken in ms: " + (System.currentTimeMillis() - time));
  }

  time taken in ms: 18478  -- with cache
  time taken in ms: 22589  -- without cache
```
  
**b) Also in the base code I found we use caching for `RewriteManifestsProcedure`, so I used the same pattern and syntax (keeping cache as default).**
Also found a bug in the base code and raised PR #4722 

c) I have some more plans for improving the `expire_snaphots` (different idea than https://github.com/apache/iceberg/pull/3457), I will work in the subsequent PR.

Idea is that add a new column "from_snapshot_id" while preparing the actual files, then filter out (NOT IN filter) the expired snapshot ids rows from persisted output (without scanning again) and then same logic of df.except() to find the expired files.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/ExpireSnapshotsProcedure.java
```diff
@@ -101,14 +103,18 @@ public InternalRow[] call(InternalRow args) {
         action.retainLast(retainLastNum);
       }
 
-      if (maxConcurrentDeletes != null && maxConcurrentDeletes > 0) {
```
ajantha-bhat, line 104: maxConcurrentDeletes > 0 is already checked in the `preconditions` above.

szehon-ho:
> c) I have some more plans for improving the expire_snaphots (different idea than https://github.com/apache/iceberg/pull/3457), I will work in the subsequent PR.

> Idea is that add a new column "from_snapshot_id" while preparing the actual files, then filter out (NOT IN filter) the expired snapshot ids rows from persisted output (without scanning again) and then same logic of df.except() to find the expired files.

The problem I think is that there's not many utilities to project anything other than partition filter to do the filtering .  I spent some time to look again, and tried to use time-travel which is effectively snapshot-filtering in  https://github.com/apache/iceberg/pull/4736 but unfortunately it didn't work as manifest table does not support it.  You can take a look if that also makes sense.

Anyway look forward to working together on this.

ajantha-bhat:
@RussellSpitzer , @szehon-ho : What you guys think about this PR?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -221,7 +230,9 @@ private DeleteOrphanFiles.Result doExecute() {
     List<String> orphanFiles = actualFileDF.join(validFileDF, joinCond, "leftanti")
         .as(Encoders.STRING())
         .collectAsList();
-
+    if (useCaching) {
```
RussellSpitzer, line 233: this needs to be in a finally block, otherwise if we have an error in the collect as list there is a possibility that we do not uncache. For services running these actions that would be a problem.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseExpireSnapshotsSparkAction.java
```diff
@@ -147,7 +153,13 @@ public BaseExpireSnapshotsSparkAction deleteWith(Consumer<String> newDeleteFunc)
   public Dataset<Row> expire() {
     if (expiredFiles == null) {
       // fetch metadata before expiration
-      Dataset<Row> originalFiles = buildValidFileDF(ops.current());
+      Table staticTableBefore = newStaticTable(ops.current(), table.io());
+      allManifestsBefore = loadMetadataTable(staticTableBefore, ALL_MANIFESTS);
+      useCaching = PropertyUtil.propertyAsBoolean(options(), USE_CACHING, USE_CACHING_DEFAULT);
+      if (useCaching) {
+        allManifestsBefore.persist();
```
RussellSpitzer, line 160: I think we would want to do the "memory only" cache here (and in all the other usages) ... but I'm not sure

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/ExpireSnapshotsProcedure.java
RussellSpitzer: It's best to leave changes like this for a separate cleanup pr

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
```diff
@@ -116,17 +122,27 @@ private Result doExecute() {
...
+    Result result;
     if (streamResults) {
-      return deleteFiles(reachableFileDF.toLocalIterator());
+      result = deleteFiles(reachableFileDF.toLocalIterator());
     } else {
-      return deleteFiles(reachableFileDF.collectAsList().iterator());
+      result = deleteFiles(reachableFileDF.collectAsList().iterator());
+    }
+    if (useCaching) {
+      allManifests.unpersist();
```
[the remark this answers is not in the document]
szehon-ho: Yea if we go for this approach, can we do something like in https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteManifestsSparkAction.java#L237. (withReusableDS) ?

ajantha-bhat:
closing as this is stale and we have improved the performance by other PRs

PR #4677, opened 2022-05-01
Core: add the file count of specific deletes in the snapshot summary

chenjunjiedada:
This adds more info for deletes in the snapshot summary.

In the case of V2 tables, users may have offline service to compact deletes.  For example, in our production, we analyze the snapshot summary from the commit event to determine whether to start major compaction or minor compaction. The detailed summary info about deletes can help us calculate better compaction timing and resource.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/SnapshotSummary.java
+++ b/core/src/main/java/org/apache/iceberg/SnapshotSummary.java
@@ -31,6 +31,10 @@ public class SnapshotSummary {
   public static final String DELETED_FILES_PROP = "deleted-data-files";
   public static final String TOTAL_DATA_FILES_PROP = "total-data-files";
   public static final String ADDED_DELETE_FILES_PROP = "added-delete-files";
+  public static final String ADD_EQ_DELETE_FILES_PROP = "added-eq-delete-files";
+  public static final String REMOVED_EQ_DELETE_FILES_PROP = "removed-eq-delete-files";
+  public static final String ADD_POS_DELETE_FILES_PROP = "added-pos-delete-files";
+  public static final String REMOVED_POS_DELETE_FILES_PROP = "removed-pos-delete-files";
   public static final String REMOVED_DELETE_FILES_PROP = "removed-delete-files";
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/SnapshotSummary.java
```diff
@@ -236,6 +248,10 @@ void clear() {
     void addTo(ImmutableMap.Builder<String, String> builder) {
       setIf(addedFiles > 0, builder, ADDED_FILES_PROP, addedFiles);
       setIf(removedFiles > 0, builder, DELETED_FILES_PROP, removedFiles);
+      setIf(addedEqDeleteFiles > 0, builder, ADD_EQ_DELETE_FILES_PROP, addedEqDeleteFiles);
+      setIf(removedEqDeleteFiles > 0, builder, REMOVED_EQ_DELETE_FILES_PROP, removedEqDeleteFiles);
+      setIf(addedPosDeleteFiles > 0, builder, ADD_POS_DELETE_FILES_PROP, addedPosDeleteFiles);
+      setIf(removedPosDeleteFiles > 0, builder, REMOVED_POS_DELETE_FILES_PROP, removedPosDeleteFiles);
       setIf(addedDeleteFiles > 0, builder, ADDED_DELETE_FILES_PROP, addedDeleteFiles);
       setIf(removedDeleteFiles > 0, builder, REMOVED_DELETE_FILES_PROP, removedDeleteFiles);
```
singhpk234, line 256: [question] since now we are tracking added / removed delete files at a more granular level (i.e pos delete added / removed & eq delete added / removed delete files) these metrics can be re-created using them, should we remove or deprecate them from snapshot summary ?

rdblue: Deprecating things is quite a bit of work for little gain here. I'd just continue to update them.

rdblue (approved):

rdblue:
Thanks, @chenjunjiedada!

PR #4678, opened 2022-05-02
Add a skeleton for the BuildPositionAccessors visitor

samredai:
This adds a skeleton visitor for the `BuildPositionAccessors` visitor and the following `# TODO` in the struct component of the visitor:
```
# TODO: Populate the `self._index` dictionary where the key is the field ID and the value is an accessor for that field.
#   The equivalent java logic can be found here: https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/Accessors.java#L213-L230
```
The main reason for adding this now is to unblock the work on expressions visitors (for example PR #4679)

cc: @rdblue @emkornfield @dramaticlly @CircArgs @jun-he @dhruv-pratap

### DIFF
--- a/python/src/iceberg/schema.py
+++ b/python/src/iceberg/schema.py
@@ -526,7 +526,7 @@ class _BuildPositionAccessors(SchemaVisitor[Dict[int, "Accessor"]]):
         return self._index
 
 
-def build_position_accessors(schema_or_type) -> Dict[int, "Accessor"]:
+def build_position_accessors(schema_or_type: Schema | IcebergType) -> Dict[int, Accessor]:
     """Generate an index of field IDs to schema position accessors
 
     Args:
--- a/python/src/iceberg/schema.py
+++ b/python/src/iceberg/schema.py
@@ -19,7 +19,10 @@ from __future__ import annotations
 
 import sys
 from abc import ABC, abstractmethod
-from typing import Dict, Generic, Iterable, List, TypeVar
+from typing import TYPE_CHECKING, Dict, Generic, Iterable, List, TypeVar
+
+if TYPE_CHECKING:
+    from iceberg.expressions.base import Accessor
 
 if sys.version_info >= (3, 8):
     from functools import singledispatch  # pragma: no cover
@@ -54,6 +57,7 @@ class Schema:
         self._name_to_id_lower: Dict[str, int] = {}  # Should be accessed through self._lazy_name_to_id_lower()
         self._id_to_field: Dict[int, NestedField] = {}  # Should be accessed through self._lazy_id_to_field()
         self._id_to_name: Dict[int, str] = {}  # Should be accessed through self._lazy_id_to_name()
+        self._id_to_accessor: Dict[int, "Accessor"] = {}  # Should be accessed through self._lazy_id_to_accessor()
 
     def __str__(self):
         return "table {\n" + "\n".join(["  " + str(field) for field in self.columns]) + "\n}"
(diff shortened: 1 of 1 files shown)


python/src/iceberg/schema.py
```diff
@@ -19,7 +19,10 @@
 
 import sys
 from abc import ABC, abstractmethod
-from typing import Dict, Generic, Iterable, List, TypeVar
+from typing import TYPE_CHECKING, Dict, Generic, Iterable, List, TypeVar
+
+if TYPE_CHECKING:
+    from iceberg.expressions.base import Accessor
```
samredai, line 25: `if TYPE_CHECKING` here is to avoid circular imports when using the class as a typehint (see docs [here](https://docs.python.org/3/library/typing.html#constant)). In summary, it's true when static type-checking is happening by mypy and false at runtime (avoiding any circular imports at runtime).

Fokko: I was curious about the circular import. Removing the guard doesn't give an error on my side. The guard doesn't do any harm, but it would be better to avoid any potential circular imports at all.

python/src/iceberg/schema.py
```diff
@@ -471,3 +495,44 @@ def index_name_by_id(schema_or_type) -> Dict[int, str]:
...
+    def map(
+        self, map_type: MapType, key_result: Dict[int, "Accessor"], value_result: Dict[int, "Accessor"]
+    ) -> Dict[int, "Accessor"]:
+        return self._index
+
+    def primitive(self, primitive: PrimitiveType) -> Dict[int, "Accessor"]:
+        return self._index
+
+
+def build_position_accessors(schema_or_type) -> Dict[int, "Accessor"]:
```
Fokko, line 529:
```suggestion
def build_position_accessors(schema_or_type: Schema | IcebergType) -> Dict[int, Accessor]:
```

python/src/iceberg/schema.py
```diff
@@ -471,3 +495,44 @@ def index_name_by_id(schema_or_type) -> Dict[int, str]:
...
     return indexer.by_id()
+
+
+class _BuildPositionAccessors(SchemaVisitor[Dict[int, "Accessor"]]):
+    """A schema visitor for generating a field ID to accessor index"""
+
+    def __init__(self) -> None:
+        self._index: Dict[int, "Accessor"] = {}
+
+    def schema(self, schema, result: Dict[int, "Accessor"]) -> Dict[int, "Accessor"]:
```
Fokko, line 506: If we defer the evaluation of the annotations:
```python
from __future__ import annotations
```
Then we can just write
```suggestion
    def schema(self, schema, result: Dict[int, "Accessor"]) -> Dict[int, Accessor]:
```

python/src/iceberg/schema.py
samredai: Sorry I forgot to mention that the circular import happens when PR #4679 is combined with this since that adds an import from schema in `expressions/base.py` so this is preemptive, specifically for [this](https://github.com/apache/iceberg/pull/4679/files#diff-47510d45242fa16200be79774cb6177b3084a1d363e15fde90940b3949a9eabaR23) line.

Fokko: At that makes sense, thanks!

Fokko: At that makes sense, thanks!

python/src/iceberg/schema.py
samredai: Nice! We already have that import at the top of this file. I was getting `NameError: name 'Accessor' is not defined` but after your comment and playing with this a bit I realize it was only [this](https://github.com/apache/iceberg/pull/4678/files#diff-03d2c14d19e23634780ee3ccba4a37a5b5a06b0acd1d7c6a5e2a9f7c60da0c1bR500) single line that was causing it:
```py
class _BuildPositionAccessors(SchemaVisitor[Dict[int, "Accessor"]]):
```
I wonder if it's because that's a type for a generic? I just left that one there and changed `"Accessor"` to `Accessor` everywhere else, thanks @Fokko !

rdblue:
Seems reasonable to me. Do you want to merge this right away, or are you planning to update it with an implementation?

samredai:
@rdblue this is good to merge now and then I'll rebase PR #4679 which should get that in a good state to merge. Thanks!

rdblue (approved):

PR #4680, opened 2022-05-02
Fix `META-INF/services` in shaded jars

snazy:
None of the shaded jars properly relocates the file names nor the contents of the services files.

For example, the file `META-INF/services/com.fasterxml.jackson.core.ObjectCodec` in a Spark runtime jar should have the name `META-INF/services/org.apache.iceberg.shaded.com.fasterxml.jackson.core.ObjectCodec` and the contents should contain `org.apache.iceberg.shaded.com.fasterxml.jackson.databind.ObjectMapper`, not
`com.fasterxml.jackson.databind.ObjectMapper`.

The Gradle shadow plugin has a builtin functionality to handle this - just need to be enabled.

### DIFF
--- a/flink/v1.13/build.gradle
+++ b/flink/v1.13/build.gradle
@@ -166,6 +166,8 @@ project(":iceberg-flink:iceberg-flink-runtime-${flinkMajorVersion}") {
     relocate 'io.airlift', 'org.apache.iceberg.shaded.io.airlift'
     relocate 'org.threeten.extra', 'org.apache.iceberg.shaded.org.threeten.extra'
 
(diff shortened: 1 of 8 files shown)


rdblue:
@snazy, what is broken by this? Iceberg purposely avoids using things like Jackson annotations because we don't control the final classpath and want to avoid problems that come from exactly this kind of issue. What does this fix that requires it to be ported to 0.13.x?

rdblue:
I'm going to close this since I don't think it is needed. Feel free to reopen if you want to clarify. Thanks!

PR #4681, opened 2022-05-02
Core: Log the new metadata location in commit.

flyrain:
This pretty useful to figure out which version actually committed to the catalog from the log, especially when we debug a catalog consistency or locking issue.
cc @aokolnychyi @RussellSpitzer @szehon-ho @karuppayya


RussellSpitzer:
I'm not sure we want to force all future table operations to include a literal location (or have to return one). So we may want to have the logging in the operations (HiveTableOperations, HadoopTableOperations, ... ) themselves?

flyrain:
> I'm not sure we want to force all future table operations to include a literal location (or have to return one). So we may want to have the logging in the operations (HiveTableOperations, HadoopTableOperations, ... ) themselves?

Good point. But I think a table version string will be there even for a future catalog without a metatdata.json file. It can easily support that, with minor word change. like `Successfully committed to table {} with the new metadata location({})` -> `Successfully committed to table {} with the new version ({})`

Hi @kbendick, what do you think from the perspective of rest API catalog?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -311,6 +311,8 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     } finally {
       cleanupMetadataAndUnlock(commitStatus, newMetadataLocation, lockId, tableLevelMutex);
     }
+
+    LOG.info("Committed to table {} with the new metadata location {}", fullName, newMetadataLocation);
```
flyrain, line 315: Instead of putting this log in both line 278 and line 297, I put the log here.

kbendick:
> > I'm not sure we want to force all future table operations to include a literal location (or have to return one). So we may want to have the logging in the operations (HiveTableOperations, HadoopTableOperations, ... ) themselves?
> 
> Good point. But I think a table version string will be there even for a future catalog without a metatdata.json file. It can easily support that, with minor word change. like `Successfully committed to table {} with the new metadata location({})` -> `Successfully committed to table {} with the new version ({})`
> 
> Hi @kbendick, what do you think from the perspective of rest API catalog?

For the current REST catalog, a file should be loggable. That said, I do agree that it's probably best not to force catalogs to do so.

flyrain:
Hi @RussellSpitzer, @rdblue @kbendick, made the change only to `HiveTableOperation`. I can make change to other table operations you think it is necessary.

kbendick:
> Hi @RussellSpitzer, @rdblue @kbendick, made the change only to `HiveTableOperation`. I can make change to other table operations you think it is necessary.

I'm good with just adding the log to `HiveTableOperations` for now. I'd _consider_ adding the log to `HadoopTableOperations` as well, as a lot of testing takes place using that and so investigations might wind up using that.

But realistically, if people who use other table operations find need or value for this log, it can be added as a follow up (particularly by people who make more common use of those table operations).

flyrain:
Added the support for `HadoopTableOperations`. But it is less likely be confusing for Hadoop table since it always write with the latest version number as the file name. One of the pain point for Hive table is that, there are multiple metadata files in the metadata directory, and you don't know which one is generated by a succeeded job, which is not. Hadoop table doesn't have this issue, it can overwrite the file generated by the failed commit. 

It is still valuable though. You can connect the file with the job easily by looking at the log.

kbendick:
> Added the support for `HadoopTableOperations`. But it is less likely be confusing for Hadoop table since it always write with the latest version number as the file name. One of the pain point for Hive table is that, there are multiple metadata files in the metadata directory, and you don't know which one is generated by a succeeded job, which is not. Hadoop table doesn't have this issue, it can overwrite the file generated by the failed commit.
> 
> It is still valuable though. You can connect the file with the job easily by looking at the log.

Ah that makes sense. I'm good with this use case then. If other TableOperations decide it's needed, then we can add it.

kbendick (approved):
This looks good to me.

flyrain:
Thanks all for the review.

PR #2538, opened 2021-04-28
Core: Add delete marker metadata column

chenjunjiedada:
This adds a metadata column to indicate whether a row is deleted or not.  A delete marker column can be used when finding the deleted rows as we discussed in https://github.com/apache/iceberg/pull/2372, it can also be used to simplify the overall merge on read process.  In order to avoid overhead, the delete marker metadata column only projected when the delete files exist.

PR #3535, opened 2021-11-11
Core: Replace Set with Bitmap to make delete filtering simpler and faster

flyrain:
This PR replaces `Set<Long>` with `Roaring64Bitmap` to store the deleted positions for filtering. Bitmap use significant less memory than a set, here is a [benchmark](https://github.com/apache/iceberg/pull/3287#discussion_r733910082). With that, we don't have to use streaming filter in case of large amount of positions, we can keep all deleted positions in memory.
cc @aokolnychyi @rdblue @RussellSpitzer @karuppayya  @szehon-ho

PR #4683, opened 2022-05-02
Read deleted rows with metadata column IS_DELETED

flyrain:
Per the discussion in https://github.com/apache/iceberg/pull/4539, created this new PR. With the change in #2538 and this PR, we can read deleted rows from row-level deletes.
<s>Please note that there is no IS_DELETED column filter push down at this moment. This will be done as a followup PR.</s> Think a bit more, we may not need pushdown, especially for non-vectorized read. It shouldn't have any difference between Spark filters out rows or Iceberg filters out rows. We iterate through all rows anyway.

cc @aokolnychyi @RussellSpitzer @chenjunjiedada @stevenzwu @Reo-LEI @hameizi @singhpk234 @rajarshisarkar @kbendick @rdblue


data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -94,6 +96,23 @@ protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSc
...
+
+  private int getIndexOfIsDeletedMetadataColumn() {
+    List<Types.NestedField> icebergFields = requiredSchema.asStruct().fields();
+    for (int i = 0; i < icebergFields.size(); i++) {
+      if (icebergFields.get(i).fieldId() == MetadataColumns.IS_DELETED.fieldId()) {
+        return i;
+      }
+    }
+
+    return -1;
```
singhpk234, line 111: [question] should we add this helper (which now takes fieldId and returns it's column position) in [Schema](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/Schema.java), Your thoughts ?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
```diff
@@ -195,5 +195,10 @@ protected StructLike asStructLike(InternalRow row) {
     protected InputFile getInputFile(String location) {
       return RowDataReader.this.getInputFile(location);
     }
+
+    @Override
+    protected void markRowDeleted(InternalRow row) {
+      row.setBoolean(columnIsDeletedIndex(), true);
+    }
```
singhpk234, line 202: [question] should we add a precondition here that `columnIsDeletedIndex` should not be -1 ? Your thoughts

presently we have made sure we will invoke markRowDelete only when hasColumnIsDeleted is true implying `columnIsDeletedIndex` can't be ideally -1.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -80,9 +88,9 @@ public TestSparkReaderDeletes(boolean vectorized) {
 
   @Parameterized.Parameters(name = "vectorized = {0}")
   public static Object[][] parameters() {
-    return new Object[][] {
-        new Object[] {false},
-        new Object[] {true}
+    return new Object[][]{
+        new Object[]{false},
+        new Object[]{true}
```
singhpk234, line 93: can revert it back if it's un-intentional

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -227,6 +251,37 @@ public void close() {
...
+        super(items, deletePositions);
+      }
+
+      @Override
+      protected boolean shouldKeep(T row) {
+        boolean isDeleted = !super.shouldKeep(row);
+        if (isDeleted) {
+          markDeleted.accept(row);
+        }
+        return true;
```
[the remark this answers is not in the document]
flyrain: Makes sense to add a comment. This part of implementation is where I'm not quite satisfied, but not sure how to make it better. I will try to refactor a bit.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
flyrain: Yeah, we don't need to check since the index cannot be -1 if hasColumnIsDeleted is true.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: We should, thanks for the suggestion.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
singhpk234: Ah I see, my IDE (using Intellij) here doesn't reformat this though, I am using this [iceberg-java-style](https://iceberg.apache.org/contribute/#setting-up-ide-and-code-style) . either ways, I think it should be fine if it's aligned to [iceberg-java-style](https://iceberg.apache.org/contribute/#setting-up-ide-and-code-style) :) !!!

Also just realized, we can re-write this to as well : 
```java
    return new Object[] {
        new Object[] {false, true};
    }
```

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -73,6 +74,16 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
     return filter.filter(rows);
   }
 
+  public static <T> CloseableIterable<T> marker(CloseableIterable<T> rows, Function<T, Long> rowToPosition,
```
chenjunjiedada, line 77: nit: `marker` is a noun,  I think here should use a verb. Maybe `filter` still fits since it is still filtering.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -185,16 +204,29 @@ private CloseableIterable<T> applyEqDeletes(CloseableIterable<T> records) {
         .reduce(Predicate::and)
         .orElse(t -> true);
 
-    Filter<T> remainingRowsFilter = new Filter<T>() {
+    if (hasColumnIsDeleted) {
+      return CloseableIterable.transform(records, record -> {
```
chenjunjiedada, line 208: I think these two can also be abstracted to `Deletes`
```java
  public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Predicate<T> pred, Consumer<T> consumer) {
    return CloseableIterable.transform(rows, row -> {
      if (!pred.test(row)) {
        consumer.accept(row);
      }
      return row;
    });
  }
```

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
flyrain: Right, we don't actually need a 2D array here.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: Nice suggestion. Made the change.

flyrain:
Do we still need class `EqualityDeleteRowReader`? The functionality can be archived by a filter on the `IS_DELETED` column? cc @chenjunjiedada

api/src/main/java/org/apache/iceberg/Schema.java
```diff
@@ -381,6 +381,22 @@ public String idToAlias(Integer fieldId) {
     return null;
   }
 
+  /**
+   * Returns the index of the given field id.
+   *
+   * @param fieldId a column id in this schema
+   * @return the index of the field in the schema, or -1 if one wasn't found
+   */
+  public int idToIndex(Integer fieldId) {
```
stevenzwu, line 390: `idToPosition`?  `position` might be more clear. 

Also we are using `Integer` arg type, which means it can be null. Probably following the pattern from the `idToAlias ` above. But I am wondering if the `fieldId` arg should be primitive `int` type. 

Also, should the return type be `Integer` and nullable just to be consistent with other methods in this class?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -94,6 +95,12 @@ protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSc
     this.eqDeletes = eqDeleteBuilder.build();
     this.requiredSchema = fileProjection(tableSchema, requestedSchema, posDeletes, eqDeletes);
     this.posAccessor = requiredSchema.accessorForField(MetadataColumns.ROW_POSITION.fieldId());
+    this.hasColumnIsDeleted = requestedSchema.findField(MetadataColumns.IS_DELETED.fieldId()) != null;
```
stevenzwu, line 98: nit: I saw a lot of places in Iceberg just use `null` to indicate the condition. should we use null for `columnIsDeletedIndex` too?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -94,6 +95,12 @@ protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSc
...
+    this.columnIsDeletedIndex = requestedSchema.idToIndex(MetadataColumns.IS_DELETED.fieldId());
```
stevenzwu, line 99: if we rename `idToIndex` to `idToPosition` as suggested in another comment, we can also rename this variable.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -227,6 +237,39 @@ public void close() {
     }
   }
 
+  private static class PositionStreamDeleteMarker<T> extends PositionStreamDeleteFilter<T> {
```
stevenzwu, line 240: as mentioned in the draft PR, this is not actually a filter as it always returns true. It is only to use the iteration part of  the filter. Maybe use `CloseableIterable#transform` instead for iteration?

chenjunjiedada:
@flyrain , I agree the implementation of `EqualityDeleteRowReader` can be replaced, but we are using the class for rewrite equality deletes. If we have the utilities to produce only equality deleted rows or position deleted rows, I'm OK to delete them.

api/src/main/java/org/apache/iceberg/Schema.java
flyrain: Thanks for the name suggestion. There are always performance concerns about non-primitive type. I'm kind of OK with both primitive type and non-primitive type here. In terms of consistency, these two public methods also return `int`. I will leave the return type to `int` unless we have a strong reason to use Integer.
```
public int schemaId()
public int highestFieldId()
```

flyrain: Even this place uses `int`, I'd prefer to use `int` for the parameter as well. It is unnecessary to do this boxing and unboxing.
https://github.com/apache/iceberg/blob/44c1d00b37f0d7fc2e978baad4f7a861a8335cf0/api/src/main/java/org/apache/iceberg/types/Types.java#L479

flyrain:
> @flyrain , I agree the implementation of `EqualityDeleteRowReader` can be replaced, but we are using the class for rewrite equality deletes. If we have the utilities to produce only equality deleted rows or position deleted rows, I'm OK to delete them.

We don't plan for that utilities in this PR. Let's not change anything about `EqualityDeleteRowReader`. Thanks for the input.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Yeah, I've replace `filter` with `transform` for all places except this one. This is trickier to change. Let me think a bit more.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -63,13 +65,23 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
     return equalityFilter.filter(rows);
   }
 
-  public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Function<T, Long> rowToPosition,
-                                                PositionDeleteIndex deleteSet) {
-    if (deleteSet.isEmpty()) {
-      return rows;
-    }
+  public static <T> CloseableIterable<T> markDeleted(CloseableIterable<T> rows, Predicate<T> isDeleted,
+                                                     Consumer<T> markDeleted) {
```
stevenzwu, line 69: nit: markDeleted -> deleteMarker

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -169,32 +176,23 @@ public CloseableIterable<T> findEqualityDeleteRows(CloseableIterable<T> records)
...
 
-    return remainingRowsFilter.filter(records);
+    if (hasColumnIsDeleted) {
+      return Deletes.markDeleted(records, isEqDeleted, this::markRowDeleted);
+    } else {
+      return Deletes.filter(records, isEqDeleted.negate());
+    }
   }
 
+  protected abstract void markRowDeleted(T item);
```
stevenzwu, line 194: provide a default impl here s for throwing `UnsupportedOperationException` so that we don't need to change too many extended classes?

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -63,13 +65,23 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
...
+    return CloseableIterable.transform(rows, row -> {
+      if (isDeleted.test(row)) {
+        markDeleted.accept(row);
+      }
+      return row;
+    });
+  }
 
-    PositionSetDeleteFilter<T> filter = new PositionSetDeleteFilter<>(rowToPosition, deleteSet);
+  public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Predicate<T> shouldKeep) {
```
stevenzwu, line 78: I like the new filter method API. only concern is for compatibility. this is a public static method from a public class. Probably very few users call this API directly. if we want to be safe, we can keep the old API and mark it as deprecated.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -67,6 +66,8 @@
   private final List<DeleteFile> eqDeletes;
   private final Schema requiredSchema;
   private final Accessor<StructLike> posAccessor;
+  private final boolean hasColumnIsDeleted;
+  private final int columnIsDeletedPosition;
```
stevenzwu, line 70: nit: isDeletedColumnPosition?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -72,46 +84,74 @@ public final ColumnarBatch read(ColumnarBatch reuse, int numRowsToRead) {
...
-    private int numRows;
     private ColumnarBatch columnarBatch;
+    private final int batchSize;
+    private int[] rowIdMapping; // the rowId mapping to skip deleted rows for all column vectors inside a batch
+    private boolean[] isDeleted; // the array to indicate if a row is deleted or not
 
-    ColumnBatchLoader(int numRowsToRead) {
-      initRowIdMapping(numRowsToRead);
-      loadDataToColumnBatch(numRowsToRead);
+    ColumnBatchLoader(int batchSize) {
```
stevenzwu, line 92: why renaming from `numRowsToRead` to `batchSize`? I have got review feedback before that `batchSize` is a little vague.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/IcebergArrowColumnVector.java
```diff
@@ -153,10 +153,15 @@ public ArrowColumnVector getChild(int ordinal) {
...
-  static ColumnVector forHolder(VectorHolder holder, int numRows) {
-    return holder.isDummy() ?
-        new ConstantColumnVector(Types.IntegerType.get(), numRows, ((ConstantVectorHolder) holder).getConstant()) :
-        new IcebergArrowColumnVector(holder);
+  static ColumnVector forHolder(VectorHolder holder, int numRows, boolean[] isDeleted) {
+    if (holder.isDummy()) {
+      if (holder instanceof VectorHolder.IsDeletedVectorHolder) {
+        return new DeletedMetaColumnVector(Types.BooleanType.get(), isDeleted);
+      }
+      return new ConstantColumnVector(Types.IntegerType.get(), numRows, ((ConstantVectorHolder) holder).getConstant());
```
stevenzwu, line 161: should we also check `instanceOf` ConstantVectorHolder in else-if, then for the unknown type we throw an exception?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +271,191 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+      }
+
+      StructLike structLike = new InternalRecordWrapper(PROJECTION_SCHEMA.asStruct()).wrap(record);
+      set.add(structLike);
+    }
+    return set;
+  }
+
+  @NotNull
+  private static List getRecordsWithDeletedColumn() {
```
stevenzwu, line 447: nit: Iceberg coding style doesn't use `get`. also return type probably can be a generic type

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +271,191 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+    df.collectAsList().forEach(row -> {
+      SparkStructLike rowWrapper = new SparkStructLike(projection);
+      actualDeleted.add(rowWrapper.wrap(row));
+    });
+
+    validate(expected, actualDeleted);
+  }
+
+  private static final Schema PROJECTION_SCHEMA = new Schema(
+      required(100, "id", Types.LongType.get()),
```
stevenzwu, line 408: curious why the field id is different with the schema in `DeleteReadTests` base class? field id doesn't matter here?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +271,191 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+      required(101, "data", Types.StringType.get()),
+      MetadataColumns.IS_DELETED
+  );
+
+  private void validate(StructLikeSet expected, StructLikeSet actual) {
```
stevenzwu, line 413: StructLikeSet supports `equals` impl. do we need this?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +271,191 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+    for (StructLike expectedRow : expected) {
+      Assert.assertTrue("Table should contain expected row", actual.contains(expectedRow));
+    }
+  }
+
+  protected static StructLikeSet expectedRowSet(int... idsToRemove) {
+    return expectedRowSet(false, false, idsToRemove);
+  }
+
+  protected static StructLikeSet expectedRowSet(boolean removeDeleted, boolean removeUndeleted, int... idsToRemove) {
```
stevenzwu, line 424: `removeUndeleted` was a little confusing name to me. Should we use a predicate to capture the filter for removeDelete or removeUndeleted?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +271,191 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+    List<Record> records = getRecordsWithDeletedColumn();
+
+    Set<Integer> deletedIds = Sets.newHashSet(ArrayUtil.toIntList(idsToRemove));
+    StructLikeSet set = StructLikeSet.create(PROJECTION_SCHEMA.asStruct());
+    for (Record record : records) {
+      if (deletedIds.contains(record.getField("id"))) {
+        if (removeDeleted) {
+          continue;
+        } else {
+          record.setField(MetadataColumns.IS_DELETED.name(), true);
```
stevenzwu, line 434: do we need to set is_deleted column to false for undeleted rows?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
stevenzwu: maybe it can be decoupled to two small methods: one for filter with predicate and one for set the is_deleted column to true.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: yeah, that's my original implementation, but you and @Reo-LEI suggested an abstract method in this comment, https://github.com/apache/iceberg/pull/4539#discussion_r854748246. I'm OK with either one. We can save some code with the default implementation, and an abstract method can enforce the implementation of the subclass, which is fine, since the default implementation also throws the same exception (UnsupportedOperationException). What do you think?

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: That's a good point. I had the same concern. However, we've discussed what should be considered as APIs in one of the community sync. we agreed that only consider the public things in the API module as APIs. cc @rdblue. With that, it should be OK to change this public method?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
flyrain: OK, I can change it back.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/IcebergArrowColumnVector.java
flyrain: It is safer that way. Although it throws anyway if the casting doesn't work.
```
(ConstantVectorHolder) holder
```

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
flyrain: It doesn't matter. I will use the same id.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
flyrain: It doesn't match when orders of rows are different so I'm using this
```
actual.contains(expectedRow)
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/IcebergArrowColumnVector.java
flyrain: Checked and refactored this part of code by adding a column vector builder.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
stevenzwu: yeah. maybe a default impl is a little better by saving some lines. sorry for going back and forth

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
stevenzwu: I thought the `equals` impl doesn't care about order either. it eventually calls the containsAll method 

```
  public boolean containsAll(Collection<?> objects) {
    if (objects != null) {
      return Iterables.all(objects, this::contains);
    }
    return false;
  }
```

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +271,189 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+  }
+
+  private static StructLikeSet expectedRowSet(int... idsToRemove) {
+    return expectedRowSet(false, false, idsToRemove);
+  }
+
+  private static StructLikeSet expectedRowSet(boolean removeDeleted, boolean removeUndeleted, int... idsToRemove) {
+    List<Record> records = recordsWithDeletedColumn();
+
+    Set<Integer> deletedIds = Sets.newHashSet(ArrayUtil.toIntList(idsToRemove));
```
stevenzwu, line 427: nit: remove the extra empty lines before and after. and maybe move this line as the first line

api/src/main/java/org/apache/iceberg/Schema.java
stevenzwu: > Even this place uses `int`, I'd prefer to use `int` for the parameter as well. It is unnecessary to do this boxing and unboxing.
> 
> https://github.com/apache/iceberg/blob/44c1d00b37f0d7fc2e978baad4f7a861a8335cf0/api/src/main/java/org/apache/iceberg/types/Types.java#L479

That is different. For NestedField, fieldId will never be null. hence it returns primitive int, which makes sense. But here, we may not find the field with the provided id, we are returning a special value `-1`. If we follow the style of other find methods in the `Schema` class, we can see `aliasToId` returns null if no matching field is found.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -63,13 +65,23 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
     return equalityFilter.filter(rows);
   }
 
-  public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Function<T, Long> rowToPosition,
-                                                PositionDeleteIndex deleteSet) {
-    if (deleteSet.isEmpty()) {
-      return rows;
-    }
+  public static <T> CloseableIterable<T> markDeleted(CloseableIterable<T> rows, Predicate<T> isDeleted,
```
szehon-ho, line 68: I think this API does a lot of things, can it be like:
``` CloseableIterable<T> markDeleted(CloseableIterable<T> rows, Consumer<T> deleteMarker) ```

 just markDeleted and then have the user call:
``` Deletes.markDeleted(filtered, marker)```

File: core/src/main/java/org/apache/iceberg/deletes/Deletes.java [hunk 24]
```diff
@@ -63,13 +65,23 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
...
+    return CloseableIterable.transform(rows, row -> {
+      if (isDeleted.test(row)) {
+        deleteMarker.accept(row);
+      }
+      return row;
+    });
+  }
 
-    PositionSetDeleteFilter<T> filter = new PositionSetDeleteFilter<>(rowToPosition, deleteSet);
+  public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Predicate<T> shouldKeep) {
```
szehon-ho, line 78: Not sure I see the point of this API, it seems equivalent to CloseableIterable.filter() ?

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -227,6 +237,39 @@ public void close() {
...
+      this.markDeleted = markDeleted;
+    }
+
+    @Override
+    protected PositionFilterIterator createPosDeleteIterator(CloseableIterator<T> items,
+                                                             CloseableIterator<Long> deletePosIterator) {
+      return new PositionDeleteMarkerIterator(items, deletePosIterator);
+    }
+
+    private class PositionDeleteMarkerIterator extends PositionFilterIterator {
```
szehon-ho, line 255: I probably need to read fully , but I also feel on the same line that we can have the caller use functional composition, instead of trying to have one master iterator that does everything?  ie,

``` filter.filter(rows).transform(deleteMarker)```

It would be cleaner and avoid having this complex iterator.

**Review by szehon-ho — COMMENTED** [review]
It's a long pr but I put some initial thoughts, can look more afterwards.

File: api/src/main/java/org/apache/iceberg/Schema.java [hunk 26]
```diff
@@ -381,6 +381,22 @@ public String idToAlias(Integer fieldId) {
...
   }
 
+  /**
+   * Returns the index of the given field id.
+   *
+   * @param fieldId a column id in this schema
+   * @return the index of the field in the schema, or -1 if one wasn't found
+   */
+  public int idToPosition(int fieldId) {
+    for (int i = 0; i < struct.fields().size(); i++) {
```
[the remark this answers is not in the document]
flyrain: yeah, I don't have to have this method in that case, thanks for the suggestion! I can do 
```
    this.columnIsDeletedPosition = requestedSchema.columns().indexOf(MetadataColumns.IS_DELETED.fieldId());
```

api/src/main/java/org/apache/iceberg/Schema.java
flyrain: Thanks for the comment. With @szehon-ho's suggestion https://github.com/apache/iceberg/pull/4683#discussion_r869717071, I removed the the new method.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
flyrain: You are right, the order doesn't matter. The error I hit before is the type mismatch here, https://github.com/apache/iceberg/blob/30639ef316a899480a3d0c88f8ea6498d9a21306/core/src/main/java/org/apache/iceberg/util/StructLikeSet.java#L166
I've fixed it in the new commit.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Not sure I understand you correctly. We cannot filter anything here, right? We keep all rows while mark the deleted rows.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: They should be the same. Thanks for the reminder.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: A boolean variable should be expressive.

flyrain:
Per discussion with @aokolnychyi @RussellSpitzer @szehon-ho, only keep the non-vectorized read in this PR, will file a vectorized version as a followup.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Fixed it in a new commit.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Yeah, as I said, we couldn't combine filter and deleteMarker. The current way is the simplest way I tried. Maybe there is a better way to handle this. I'm open to ideas.

aokolnychyi:
Thanks, @flyrain. I think working on non-vectorized and vectorized paths separately is a good idea.
Let me take a look at this PR.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -94,6 +95,12 @@ protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSc
     this.eqDeletes = eqDeleteBuilder.build();
     this.requiredSchema = fileProjection(tableSchema, requestedSchema, posDeletes, eqDeletes);
     this.posAccessor = requiredSchema.accessorForField(MetadataColumns.ROW_POSITION.fieldId());
+    this.hasColumnIsDeleted = requestedSchema.findField(MetadataColumns.IS_DELETED.fieldId()) != null;
+    this.columnIsDeletedPosition = requestedSchema.columns().indexOf(MetadataColumns.IS_DELETED);
```
aokolnychyi, line 99: Should we use `requiredSchema` or `requestedSchema`?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,10 +226,15 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
 
     // if there are fewer deletes than a reasonable number to keep in memory, use a set
     if (posDeletes.stream().mapToLong(DeleteFile::recordCount).sum() < setFilterThreshold) {
-      return Deletes.filter(records, this::pos, Deletes.toPositionIndex(filePath, deletes));
+      Predicate<T> isInDeleteSet = record -> Deletes.toPositionIndex(filePath, deletes).isDeleted(pos(record));
```
aokolnychyi, line 229: Won't we call `Deletes$toPositionIndex` for every row now?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -169,30 +176,23 @@ public CloseableIterable<T> findEqualityDeleteRows(CloseableIterable<T> records)
...
-    return remainingRowsFilter.filter(records);
+    if (hasColumnIsDeleted) {
+      return Deletes.markDeleted(records, isEqDeleted, this::markRowDeleted);
+    } else {
+      return CloseableIterable.filter(records, isEqDeleted.negate());
+    }
+  }
+
+  protected void markRowDeleted(T item) {
+    throw new UnsupportedOperationException("GenericDeleteFilter.markRowDeleted() is not supported");
```
aokolnychyi, line 195: What about `this.getClass().getName() + " does not implement markRowDeleted"` to cover all potential implementations?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -290,8 +295,6 @@ private static Schema fileProjection(Schema tableSchema, Schema requestedSchema,
       requiredIds.addAll(eqDelete.equalityFieldIds());
     }
 
-    requiredIds.add(MetadataColumns.IS_DELETED.fieldId());
```
aokolnychyi, line 293: Could you comment on why this was removed?

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
```diff
@@ -104,7 +103,6 @@ public class TestSparkParquetReadMetadataColumns {
       }
       row.update(1, UTF8String.fromString("str" + i));
       row.update(2, i);
-      row.update(3, false);
```
aokolnychyi, line 107: Why do we need these changes?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
aokolnychyi, line 99: This assumes query engines should be able to set a value in a row for a particular position. Seems reasonable but let us also think about alternatives.

aokolnychyi (commented):
I did one high-level pass.

We should also keep in mind performance. Do we have benchmarks that can be run before and after this change? If no, can we add them? I think we had some already but I doubt they cover this particular metadata column.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: We shouldn't add the metadata column if user doesn't ask a project or filter on it, right?  
It was added by #2538. Hi @chenjunjiedada, any idea why we need this line in #2538?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: I didn't see any difference between them. Although we got the following code in the method `fileProjection()`, I don't think the code takes effect since the method `fileProjection()` is to get the columns which are required by equality deletes and the pos metadata column if pos delete is there. Anyway, I admit it's safer to use `requiredSchema` for the future change. Will make the change.
```
    if (missingIds.contains(MetadataColumns.IS_DELETED.fieldId())) {
      columns.add(MetadataColumns.IS_DELETED);
    }
```

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -63,14 +65,14 @@ public static <T> CloseableIterable<T> filter(CloseableIterable<T> rows, Functio
...
-    if (deleteSet.isEmpty()) {
-      return rows;
-    }
-
-    PositionSetDeleteFilter<T> filter = new PositionSetDeleteFilter<>(rowToPosition, deleteSet);
-    return filter.filter(rows);
+  public static <T> CloseableIterable<T> markDeleted(CloseableIterable<T> rows, Predicate<T> isDeleted,
+                                                     Consumer<T> deleteMarker) {
+    return CloseableIterable.transform(rows, row -> {
+      if (isDeleted.test(row)) {
```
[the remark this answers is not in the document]
flyrain: My point is we have to check each row inevitably for non-vectorized read. The vectorized read is a different story. For example, we can generate the isDelete boolean array directly from the pos deletes. 
Of course there is no way to avoid per-row check for eq deletes, neither with vectorized read, nor non-vectorized read.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: I'm not sure about the other ways. A accessor-like?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: Good catch! Thanks. Fixed it in the new commit.

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
flyrain: They are broken by the vectorized changes. I added back them since we are going to handle the vectorized version in another PR.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
chenjunjiedada: We already project an extra position column, right? It depends on whether the projected column is useful. The projected  `_delete` column can be used for converting delete as I wrote if I recall. The CDC changes may also need it for joining, right?

flyrain:
Here are benchmarks of equality deletes and position deletes for a 10M-rows data file with different percentage(from 0 to 100%) of deleted rows.
The perf of read without is_deleted column are the same with the read with "_deleted = false", which is expected. The grey line is for read of deleted rows. It outputs more and more rows when percentage of deletes increases. The perf chart also makes sense, even though there are room to optimize when there are just a few rows deleted.

Screen Shot 2022-05-17 at 9 58 58 PM

Screen Shot 2022-05-17 at 10 01 04 PM

spark/v3.2/spark/src/jmh/java/org/apache/iceberg/spark/source/IcebergSourceDeleteBenchmark.java
```diff
@@ -103,6 +103,30 @@ public void readIcebergVectorized() {
     });
   }
 
+  @Benchmark
```
flyrain, line 106: Added two benchmarks @aokolnychyi

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: We project the pos column only if there are pos deletes as the following code shows, which makes sense, since we need it for filtering pos deletes. 
```
    if (!posDeletes.isEmpty()) {
      requiredIds.add(MetadataColumns.ROW_POSITION.fieldId());
    }
```
Here is my thought on Is_deleted column, it presents only if the front end(e.g. spark read) asked for it. For example, in case of CDC, we put it filter like this to read deleted rows. Here is the code from my CDC draft PR #4539.
```
    Dataset<Row> scanDF = spark().read().format("iceberg")
        .option(SparkReadOptions.FILE_SCAN_TASK_SET_ID, groupID)
        .load(table.name())
        .filter(functions.column(MetadataColumns.IS_DELETED.name()).equalTo(true));
```

chenjunjiedada: Sounds good to me!

flyrain:
Thanks @aokolnychyi, @stevenzwu, @szehon-ho @chenjunjiedada. Resolved all comments. Please take a look again when you have a chance.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
szehon-ho: One idea is composition:

Make a class DeleteIterator(Iterator<T> items, Iterator<T> deletePosIterator), which has method isDeleted() (the inverse of today's PositionFilterIterator's shouldKeep).

Then both PosFilterIterator and PosDeleteIterator compose it:
```
PosFilterIterator(DeleteIterator<T> deleteIterator) extend FilterIterator {
   boolean shouldKeep(row) {
     return !deleteIterator.isDeleted(row) 
   }

PositionDeleteMarkerIterator(DeleteIterator<T> deleteIterator, Consumer<T> markDeleted) {
   T next() {
      if (deleteIterator.isDeleted(row)) {
         markDeleted.apply(row)
      }
      return deleteIterator.next()
   }
```

Hope that's correct.  I think overhead wise, it should be similar today to the extra super method call?  Also fyi if @aokolnychyi  has any thought on it.

flyrain: Thanks for the suggestion. Made the change. We don't have to create the new class for the marker. I use the `CloseableIterator.transform` for the marker, and an anonymous class of FilterIterator for the filter. Added a test as well.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +146,81 @@ protected boolean shouldKeep(T row) {
...
-        super(items);
-        this.deletePosIterator = deletePositions;
+      // consume delete positions until the next is past the current position
+      boolean isDeleted = currentPos == nextDeletePos;
+      while (deletePosIterator.hasNext() && nextDeletePos <= currentPos) {
         this.nextDeletePos = deletePosIterator.next();
+        if (!isDeleted && currentPos == nextDeletePos) {
+          // if any delete position matches the current position
+          isDeleted = true;
+        }
```
flyrain, line 191: This logic isn't necessary. Not sure why it is there at the beginning. The only chance to go into the while loop is that there are duplicated pos deletes. For example, these positions are deleted, 0L, 0L, 1L, 3L, 3L. It is not necessary to set anything in that case, line 185 already check it. To provide more context, both row position and pos deletes are sorted. Check testPositionStreamRowFilterWithDuplicates for details.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Refactored

core/src/test/java/org/apache/iceberg/deletes/TestPositionFilter.java
```diff
@@ -151,11 +176,11 @@ public void testPositionStreamRowFilterWithRowGaps() {
         Row.of(6L, "g")
     ));
 
-    CloseableIterable<Long> deletes = CloseableIterable.withNoopClose(Lists.newArrayList(0L, 3L, 4L, 7L, 9L));
+    CloseableIterable<Long> deletes = CloseableIterable.withNoopClose(Lists.newArrayList(0L, 2L, 3L, 4L, 7L, 9L));
```
flyrain, line 179: This change is to cover the code path line 188-191 in class `Deletes`

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,10 +226,16 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
 
     // if there are fewer deletes than a reasonable number to keep in memory, use a set
     if (posDeletes.stream().mapToLong(DeleteFile::recordCount).sum() < setFilterThreshold) {
-      return Deletes.filter(records, this::pos, Deletes.toPositionIndex(filePath, deletes));
+      PositionDeleteIndex positionIndex = Deletes.toPositionIndex(filePath, deletes);
```
szehon-ho, line 229: Extract this to a separate variable on top, and use it in outside the else block?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,10 +226,16 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
...
+      Predicate<T> isInDeleteSet = record -> positionIndex.isDeleted(pos(record));
+      return hasColumnIsDeleted ?
+          Deletes.markDeleted(records, isInDeleteSet, this::markRowDeleted) :
```
szehon-ho, line 232: Just personal preference, but maybe we can push the pos function call down to Deletes (as we do it for streaming, but not the non-streaming)


```
PositionDeleteIndex positionDeleteIndex = Deletes.deletePositions(filePath, deletes);
   return hasColumnIsDeleted ?
      Deletes.markDeleted(records, this::pos, positionDeleteIndex, this::markDeleted) : 
      Deletes.filterDeleted(records, this::pos, positionDeleteIndex);
}

where Deletes.markDeleted is just a wrapper that makes the Predicate.

```

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+
+    table.newRowDelta()
+        .addDeletes(posDeletes.first())
+        .validateDataFilesExist(posDeletes.second())
+        .commit();
+
+    StructLikeSet expected = expectedRowSet(29, 43, 61, 89);
+    StructLikeSet actual = rowSet(tableName, PROJECTION_SCHEMA.asStruct(), "id", "data", "_deleted");
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
```
szehon-ho, line 302: Why not just assertEquals?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+        table, Files.localOutput(temp.newFile()), TestHelpers.Row.of(0), dataDeletes, deleteRowSchema);
+
+    table.newRowDelta()
+        .addDeletes(eqDeletes)
+        .commit();
+
+    StructLikeSet expected = expectedRowSet(29, 89, 122);
+    StructLikeSet actual = rowSet(tableName, PROJECTION_SCHEMA.asStruct(), "id", "data", "_deleted");
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
```
szehon-ho, line 330: Same, assertEquals

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+    table.newRowDelta()
+        .addDeletes(eqDeletes)
+        .addDeletes(posDeletes.first())
+        .validateDataFilesExist(posDeletes.second())
+        .commit();
+
+    StructLikeSet expected = expectedRowSet(29, 89, 121, 122);
+    StructLikeSet actual = rowSet(tableName, PROJECTION_SCHEMA.asStruct(), "id", "data", "_deleted");
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
```
szehon-ho, line 367: Same

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+
+    StructLikeSet expected = expectedRowSet(29, 89, 122);
+    StructLikeSet actual = rowSet(tableName, PROJECTION_SCHEMA.asStruct(), "id", "data", "_deleted");
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
+  }
+
+  @Test
+  public void testMixedPosAndEqDeletesWithDeletedColumn() throws IOException {
+    if (vectorized) {
```
szehon-ho, line 335: How about just Assume?  So it makes it properly skipped on JUnit then.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+
+    StructLikeSet expected = expectedRowSet(29, 43, 61, 89);
+    StructLikeSet actual = rowSet(tableName, PROJECTION_SCHEMA.asStruct(), "id", "data", "_deleted");
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
+  }
+
+  @Test
+  public void testEqualityDeleteWithDeletedColumn() throws IOException {
+    if (vectorized) {
```
szehon-ho, line 307: Same, Assume

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
 
     Assert.assertEquals("Table should contain expected rows", expected, actual);
   }
+
+  @Test
+  public void testPosDeletesWithDeletedColumn() throws IOException {
+    if (vectorized) {
```
szehon-ho, line 279: Assume

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+        .select("id", "data", "_deleted")
+        .filter("_deleted = true");
+
+    StructLikeSet actualDeleted = StructLikeSet.create(projection);
+    df.collectAsList().forEach(row -> {
+      SparkStructLike rowWrapper = new SparkStructLike(projection);
+      actualDeleted.add(rowWrapper.wrap(row));
+    });
+
+    Assert.assertTrue("Table should contain expected row", actualDeleted.equals(expectedDeleted));
```
szehon-ho, line 424: assertEquals

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+        .filter("_deleted = false");
+
+    Types.StructType projection = PROJECTION_SCHEMA.asStruct();
+    StructLikeSet actual = StructLikeSet.create(projection);
+    df.collectAsList().forEach(row -> {
+      SparkStructLike rowWrapper = new SparkStructLike(projection);
+      actual.add(rowWrapper.wrap(row));
+    });
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
```
szehon-ho, line 407: assertEquals

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+
+    Pair<DeleteFile, CharSequenceSet> posDeletes = FileHelpers.writeDeleteFile(
+        table, Files.localOutput(temp.newFile()), TestHelpers.Row.of(0), deletes);
+
+    table.newRowDelta()
+        .addDeletes(posDeletes.first())
+        .validateDataFilesExist(posDeletes.second())
+        .commit();
+
+    // get undeleted rows
```
szehon-ho, line 391: Nit: comment may make more sense if directly above the spark.read() line instead of line to make expectedRowSet.  Or can remove it , I think.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+
+    StructLikeSet expected = expectedRowSet(29, 89, 121, 122);
+    StructLikeSet actual = rowSet(tableName, PROJECTION_SCHEMA.asStruct(), "id", "data", "_deleted");
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
+  }
+
+  @Test
+  public void testFilterOnDeletedMetadataColumn() throws IOException {
+    if (vectorized) {
```
szehon-ho, line 372: Assume

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+    StructLikeSet actual = StructLikeSet.create(projection);
+    df.collectAsList().forEach(row -> {
+      SparkStructLike rowWrapper = new SparkStructLike(projection);
+      actual.add(rowWrapper.wrap(row));
+    });
+
+    Assert.assertTrue("Table should contain expected row", actual.equals(expected));
+
+    // get deleted rows
+    StructLikeSet expectedDeleted = expectedRowSet(false, true, 29, 43, 61, 89);
```
szehon-ho, line 410: Nit: can put inline comment on true/false, like: https://iceberg.apache.org/contribute/#boolean-arguments

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -262,4 +273,199 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
...
+      required(1, "id", Types.IntegerType.get()),
+      required(2, "data", Types.StringType.get()),
+      MetadataColumns.IS_DELETED
+  );
+
+  private static StructLikeSet expectedRowSet(int... idsToRemove) {
+    return expectedRowSet(false, false, idsToRemove);
+  }
+
+  private static StructLikeSet expectedRowSet(boolean removeDeleted, boolean removeUndeleted, int... idsToRemove) {
```
szehon-ho, line 437: Should it be nonDeleted instead of unDeleted?  (unDeleted sounds like it was deleted and then put back)

szehon-ho (changes requested):
Looks better to me, some more minor comments.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +146,81 @@ protected boolean shouldKeep(T row) {
...
       addCloseable(iter);
+      addCloseable(deletePosIterator);
 
       return iter;
     }
 
-    private class PositionFilterIterator extends FilterIterator<T> {
-      private final CloseableIterator<Long> deletePosIterator;
-      private long nextDeletePos;
+    boolean isDeleted(T row) {
```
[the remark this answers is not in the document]
flyrain: We intentionally changed the method name `deleted` to `isDeleted` in this PR https://github.com/apache/iceberg/pull/3535#discussion_r748893494. `deleted` is more concise, `isDeleted` is a bit easier to grab. I'm OK with either one. What do you think?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: Not sure I got you, but `positionIndex` is used only in this block.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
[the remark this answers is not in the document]
szehon-ho: OK , that's fine then.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
szehon-ho: Yea sorry you can ignore this one, I mis-read it.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
flyrain: `Deletes.markDeleted` is used by eq delete as well. I'd prefer to keep it as is. I made a new method `Deletes.filterDeleted` to make it more symmetric.

szehon-ho: Hm i didnt see markDeleted used anywhere else in code, maybe I missed it?  

I thought it would be more symmetric to have Deletes class take in PositionDeleteIndex (given it returns a PositionDeleteIndex) and Function<T, Long> rowToPosition (given other methods take it), but just style preference.

szehon-ho (approved):
Looks good to me

aokolnychyi:
Let me take another look too. Sorry for the delay.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +150,91 @@ protected boolean shouldKeep(T row) {
...
-        // consume delete positions until the next is past the current position
-        boolean keep = currentPos != nextDeletePos;
-        while (deletePosIterator.hasNext() && nextDeletePos <= currentPos) {
-          this.nextDeletePos = deletePosIterator.next();
-          if (keep && currentPos == nextDeletePos) {
-            // if any delete position matches the current position, discard
-            keep = false;
-          }
+    @Override
+    protected CloseableIterator createPosDeleteIterator(CloseableIterator<T> items) {
```
aokolnychyi, line 211: nit: raw type

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +150,91 @@ protected boolean shouldKeep(T row) {
...
-      @Override
-      protected boolean shouldKeep(T row) {
-        long currentPos = extractPos.apply(row);
-        if (currentPos < nextDeletePos) {
-          return true;
-        }
+      return isDeleted;
+    }
+
+    protected abstract CloseableIterator createPosDeleteIterator(CloseableIterator<T> items);
```
aokolnychyi, line 201: nit: raw type

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +150,91 @@ protected boolean shouldKeep(T row) {
...
 
-      @Override
-      public void close() {
-        super.close();
-        try {
-          deletePosIterator.close();
-        } catch (IOException e) {
-          throw new UncheckedIOException("Failed to close delete positions iterator", e);
+    @Override
+    protected CloseableIterator createPosDeleteIterator(CloseableIterator<T> items) {
```
aokolnychyi, line 231: nit: raw type

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
```diff
@@ -195,5 +195,10 @@ protected StructLike asStructLike(InternalRow row) {
     protected InputFile getInputFile(String location) {
       return RowDataReader.this.getInputFile(location);
     }
+
+    @Override
+    protected void markRowDeleted(InternalRow row) {
+      row.setBoolean(columnIsDeletedPosition(), true);
```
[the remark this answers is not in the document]
flyrain: Yes. That's the assumption. Agreed that we'd better not have that assumption. There will be a bit more expensive since we need to set every row now. I'd guess the perf impact is minor though. Will make the change.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
aokolnychyi, line 201: Do we really create a position delete iterator? Don't we iterate over records here?

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
aokolnychyi: The logic here is tricky but seems correct to me.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +150,91 @@ protected boolean shouldKeep(T row) {
...
         iter = rows.iterator();
-        try {
-          deletePosIterator.close();
-        } catch (IOException e) {
-          throw new UncheckedIOException("Failed to close delete positions iterator", e);
-        }
       }
 
       addCloseable(iter);
+      addCloseable(deletePosIterator);
```
aokolnychyi, line 177: Did we leak resources before?

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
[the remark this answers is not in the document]
RussellSpitzer: Is the worry here that branch prediction is gonna be bad here? I would assume that speculative branch execution would probably make ay optimizations we do here very sensitive to the amount of deletes. Anyway I'm on team leave it simple for now.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -67,6 +66,8 @@
   private final List<DeleteFile> eqDeletes;
   private final Schema requiredSchema;
   private final Accessor<StructLike> posAccessor;
+  private final boolean hasColumnIsDeleted;
```
aokolnychyi, line 69: nit: `hasIsDeletedColumn`?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
aokolnychyi, line 70: nit: `isDeletedColumnPosition`?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -94,6 +95,12 @@ protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSc
     this.eqDeletes = eqDeleteBuilder.build();
     this.requiredSchema = fileProjection(tableSchema, requestedSchema, posDeletes, eqDeletes);
     this.posAccessor = requiredSchema.accessorForField(MetadataColumns.ROW_POSITION.fieldId());
+    this.hasColumnIsDeleted = requiredSchema.findField(MetadataColumns.IS_DELETED.fieldId()) != null;
+    this.columnIsDeletedPosition = requiredSchema.columns().indexOf(MetadataColumns.IS_DELETED);
+  }
+
+  protected int columnIsDeletedPosition() {
```
aokolnychyi, line 102: nit: `isDeletedColumnPosition`

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: No, it was closed in the nested class `PositionFilterIterator` before, which is removed in this PR.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: We do iterate the pos delete records, for example, line 191, `this.nextDeletePos = deletePosIterator.next();`.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,10 +222,20 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
...
+      return createDeleteIterable(records, isDeleted);
     }
 
-    return Deletes.streamingFilter(records, this::pos, Deletes.deletePositions(filePath, deletes));
+    return hasColumnIsDeleted ?
+        Deletes.streamingMarker(records, this::pos, Deletes.deletePositions(filePath, deletes), this::markRowDeleted) :
+        Deletes.streamingFilter(records, this::pos, Deletes.deletePositions(filePath, deletes));
+  }
+
+  private CloseableIterable<T> createDeleteIterable(CloseableIterable<T> records, Predicate<T> isDeleted) {
```
aokolnychyi, line 235: Is there a better name? Is it an iterable of remaining rows?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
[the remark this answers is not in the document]
aokolnychyi: We should test the perf implication but we will remove the if statement so it should be either on par or faster.

spark/v3.2/spark/src/jmh/java/org/apache/iceberg/spark/source/IcebergSourceDeleteBenchmark.java
aokolnychyi: Thanks for adding these.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
aokolnychyi: We do iterate over position deletes but I think this iterate is for remaining data records, no?

aokolnychyi:
This looks close to me. I'd switch to an explicit call to set the `_deleted` value to avoid any assumptions about default values.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +150,91 @@ protected boolean shouldKeep(T row) {
...
+        return false;
+      }
 
-      protected PositionFilterIterator(CloseableIterator<T> items, CloseableIterator<Long> deletePositions) {
-        super(items);
-        this.deletePosIterator = deletePositions;
+      // consume delete positions until the next is past the current position
+      boolean isDeleted = currentPos == nextDeletePos;
+      while (deletePosIterator.hasNext() && nextDeletePos <= currentPos) {
         this.nextDeletePos = deletePosIterator.next();
```
RussellSpitzer, line 191: This is now the only place in the code which refers to it as "this.nextDeletePos"

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
```diff
@@ -137,93 +150,91 @@ protected boolean shouldKeep(T row) {
...
+    boolean isDeleted(T row) {
+      long currentPos = rowToPosition.apply(row);
+      if (currentPos < nextDeletePos) {
+        return false;
+      }
 
-      protected PositionFilterIterator(CloseableIterator<T> items, CloseableIterator<Long> deletePositions) {
-        super(items);
-        this.deletePosIterator = deletePositions;
+      // consume delete positions until the next is past the current position
```
RussellSpitzer, line 188: Not sure if this is any simpler but we can remove one if statement, not sure if this is more clear.

```java
      // Consume nextDeletePos till past currentPos, if currentPos equals any consumed nextDeletePos the current row has been deleted
      boolean isDeleted = currentPos == nextDeletePos;
      while (deletePosIterator.hasNext() && nextDeletePos <= currentPos) {
        this.nextDeletePos = deletePosIterator.next();
        isDeleted |= currentPos == nextDeletePos
      }
    return isDeleted;
  }
  ```

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
rdblue: I'd say it's a bit more than a nit! Iceberg PRs should always use the type system.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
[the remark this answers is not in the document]
flyrain: Discussed with @aokolnychyi offline a bit. I've tested out that we actually set the default value explicitly here, https://github.com/apache/iceberg/blob/664439344aefc60836fd519a21c57c0406c3f96c/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/SparkParquetReaders.java#L170. So it is not an assumption. Would prefer to keep it as is.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Fixed it in a new commit.

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
flyrain: Discussed offline with @aokolnychyi, I've changed the name to `applyDelete`

core/src/main/java/org/apache/iceberg/deletes/Deletes.java
aokolnychyi: Let's keep it as-is to avoid last-minute changes in this tricky place.

aokolnychyi (approved):

aokolnychyi:
Thanks, @flyrain! Great to have this done. Thanks everyone who reviewed!

flyrain:
Thank @aokolnychyi! Thanks everyone for the review.

PR #4685, opened 2022-05-03
Python: First version of the Accessor

Fokko:
First version of the visitor going over the struct to construct the accessors.


python/src/iceberg/schema.py
```diff
@@ -290,6 +292,8 @@ def _(obj: StructType, visitor: SchemaVisitor[T]) -> T:
         visitor.before_field(field)
         try:
             result = visit(field.type, visitor)
+        except Exception:
+            logger.exception("Unable to visit the field in the schema")
```
rdblue, line 296: Should this also log which field wasn't handled? Something like `f"Unable to visit field: {field}"`?

python/src/iceberg/schema.py
```diff
@@ -503,21 +507,34 @@ class _BuildPositionAccessors(SchemaVisitor[Dict[int, "Accessor"]]):
...
+        return field_results
+
+    def struct(self, struct: StructType, field_results: List[Dict[int, Accessor]]) -> Dict[int, Accessor]:
+        fields = struct.fields
+        for idx, field in enumerate(fields):
+            result = field_results[idx]
+            if result:
+                self._index.update(result)
+            else:
+                self._index[field.field_id] = Accessor(field.field_id)
```
rdblue, line 520: Doesn't this need to create the accessor based on `idx` because the accessor uses position to locate a field?

rdblue: Yeah, it looks like this is creating accessors based on the field ID, but what we need is to create accessors that use the current schema's positions and return a map of them by field ID.

For example, this schema:
```
table {
  2: int id,
  1: string data,
  3: struct location {
    5: float lat,
    6: float long
  }
}
```

Should produce the following accessor map:

```
1 -> Accessor(position=1)
2 -> Accessor(position=0)
3 -> Accessor(position=2)
5 -> Accessor(position=2, inner=Accessor(position=0))
6 -> Accessor(position=2, inner=Accessor(position=1))
```

That way when an expression uses, `location.lat`, the accessor will call `row.get(2).get(0)` to return the value.

python/src/iceberg/expressions/base.py
```diff
@@ -278,6 +278,9 @@ def __str__(self):
     def __repr__(self):
         return f"Accessor(position={self._position})"
 
+    def __eq__(self, other):
+        return other and self._position == other.position
```
samredai, line 282: We don't manipulate the `self._position` property so this is just a nitpick, but should we make this `self.position` just so it goes through that path?

samredai:
@Fokko thanks for catching this! It makes sense to me and I don't think there are other areas that make post-order traversal a requirement.

python/tests/test_schema.py
```diff
@@ -347,3 +349,17 @@ def test_schema_find_type(table_schema_simple):
         == table_schema_simple.find_type("BAZ", case_sensitive=False)
         == BooleanType()
     )
+
+
+def test_build_position_accessors(table_schema_nested):
```
Fokko, line 354: As you can see here, the list/maps are still to be implemented :)

samredai (approved):

python/src/iceberg/schema.py
Fokko: I've changed the code to re-raise the exception. I had a couple of time during debugging that it silently swallowed the exception, which didn't really help in the process of debugging :)

python/tests/test_schema.py
Fokko: Also fixed!

python/src/iceberg/schema.py
```diff
@@ -279,53 +319,74 @@ def visit(obj, visitor: SchemaVisitor[T]) -> T:
...
     finally:
         visitor.after_map_key(obj.key)
 
+    visitor.field(obj.value)
     visitor.before_map_value(obj.value)
     try:
-        value_result = visit(obj.value.type, visitor)
+        visit(obj.value.type, visitor)
+    except Exception as e:
+        logger.exception(f"Unable to visit the map value type: {obj.value}")
```
rdblue, line 384: Do we need to log exceptions at every level? Why not propagate like normal to avoid too many logs?

A heavily nested object with an exception would result in _a lot_ of logs, plus the exception getting handled and logged at a higher level.

python/src/iceberg/schema.py
```diff
@@ -262,10 +267,45 @@ def primitive(self, primitive: PrimitiveType) -> T:
...
+    def get(self, container: StructProtocol) -> Any:
+        """Returns the value at self.position in `container`
+
+        Args:
+            container(StructProtocol): A container to access at position `self.position`
+
+        Returns:
+            Any: The value at position `self.position` in the container
+        """
+        return container.get(self.position)
```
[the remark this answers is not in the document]
Fokko: Great catch. Just updated the code and added a test.

python/src/iceberg/schema.py
```diff
@@ -262,10 +267,45 @@ def primitive(self, primitive: PrimitiveType) -> T:
...
+
+
 @singledispatch
 def visit(obj, visitor: SchemaVisitor[T]) -> T:
     """A generic function for applying a schema visitor to any point within a schema
 
+    The function traverses the schema in pre-order fashion
```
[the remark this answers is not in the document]
Fokko: For building the Accessors, the pre-order is much easier since we can just wrap the field in the `inner` function. Also, I think it is much easier for most people to reason because it follows the pattern of a recursive function.

I was looking at the existing `_IndexByName` visitor, and it looks like we're actually using pre-order traversal as well. But we misuse the before and after hooks: https://github.com/apache/iceberg/blob/1bb526264b22372bf94a0d28ccf626da93ea6014/python/src/iceberg/schema.py#L391-L410

I could implement a similar pattern for creating the accessors. WDYT @samredai ?

python/src/iceberg/schema.py
Fokko: This was in the original code, but it would be best to remove this. If something fails while traversing the schema, we don't want to continue and just return a part of the schema. In that case we just want to throw an exception

python/tests/conftest.py
```diff
@@ -19,7 +19,7 @@
 
 import pytest
 
-from iceberg import schema
+from iceberg.schema import Schema
```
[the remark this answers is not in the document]
samredai: I think it matters more for the files containing actual tests (for example `schema.py` should be tested in a file named `test_schema.py` which should import the module using `from iceberg import schema`). That way when you're looking at an individual test, it's almost always a safe assumption that the objects that start with `schema.` are what's being tested.

That doesn't necessarily apply here in `conftest.py` which is mostly a collection of global fixtures so I think either is fine.

python/src/iceberg/schema.py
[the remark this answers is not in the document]
rdblue: The purpose of the visitor pattern is to separate the logic that traverses the tree structure from the logic that operates on it. Moving to pre-order conflicts with that goal because you have to inspect the inner type using `instanceof`. It think it looks more like recursive method that traverses the schema because it has to redo the work of the visitor.

Iceberg uses the pattern of using a visitor to build a tree structure quite a bit, and to do that you almost always want to use post-order so that you can wrap what was returned by visiting the children of a node. There are other ways to do it, but you end up keeping state in the visitor or needing to pass state down the tree. I think it's cleaner and easier to work with using the existing pattern. But even if that's not convincing, I'm concerned about having to rewrite all of the Java visitors to use different logic, rather than being able to see what they're doing and port the same logic. That's going to be quite a bit harder.

python/src/iceberg/schema.py
```diff
@@ -498,26 +565,76 @@ def index_name_by_id(schema_or_type) -> Dict[int, str]:
...
-    def struct(self, struct, result: List[Dict[int, Accessor]]) -> Dict[int, Accessor]:
-        # TODO: Populate the `self._index` dictionary where the key is the field ID and the value is an accessor for that field.
-        #   The equivalent java logic can be found here: https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/Accessors.java#L213-L230
+    def struct(self, struct: StructType) -> Dict[int, Accessor]:
         return self._index
 
-    def field(self, field: NestedField, result: Dict[int, Accessor]) -> Dict[int, Accessor]:
+    def field(self, field: NestedField) -> Dict[int, Accessor]:
+        field_type = field.type
+        if isinstance(field_type, StructType):
```
rdblue, line 613: I don't think it is necessary to duplicate the traversal logic. This can be done in `struct` before `field` is handled. Same for `map` and `list`.

python/src/iceberg/schema.py
```diff
@@ -498,26 +565,76 @@ def index_name_by_id(schema_or_type) -> Dict[int, str]:
...
+            self._parents[field_type.element.field_id] = field.field_id
+
+        parent = 0
+        if field.field_id in self._parents:
+            parent = self._parents[field.field_id]
+            self._index[field.field_id] = self._index[parent].with_inner(Accessor(position=self._pos[parent]))
+        else:
+            self._index[field.field_id] = Accessor(position=self._pos[parent])
+
+        self._pos[parent] = self._pos[parent] + 1
```
rdblue, line 630: I see that this is reconstructing the position in the nested type based on the number of calls to `field`. This seems a bit hard to follow, at least to me. I prefer the more straightforward approach in the Java version, where position is handled with a local variable, `i`, in `struct`:

```java
      for (int i = 0; i < fieldResults.size(); i += 1) {
        Map<Integer, Accessor<StructLike>> result = fieldResults.get(i);
        if (result != null) {
          for (Map.Entry<Integer, Accessor<StructLike>> entry : result.entrySet()) {
            accessors.put(entry.getKey(), newAccessor(i, field.isOptional(), entry.getValue()));
          }
        } else {
          accessors.put(field.fieldId(), newAccessor(i, field.type()));
        }
      }
```

python/src/iceberg/schema.py
```diff
@@ -279,53 +326,54 @@ def visit(obj, visitor: SchemaVisitor[T]) -> T:
...
-    return visitor.list(obj, result)
+    return result
 
 
 @visit.register(MapType)
 def _(obj: MapType, visitor: SchemaVisitor[T]) -> T:
     """Visit a MapType with a concrete SchemaVisitor"""
+    result = visitor.map(obj)
+
+    visitor.field(obj.key)
```
rdblue, line 366: Changing the behavior of `field` is probably going to cause problems later when translating between Java and Python implementations. `field` isn't called for map key, map value, or list element in any of the Java visitors.

python/src/iceberg/schema.py
```diff
@@ -262,10 +267,52 @@ def primitive(self, primitive: PrimitiveType) -> T:
...
+    position: int
+    inner: Optional["Accessor"] = None
+
+    def __str__(self):
+        return f"Accessor(position={self.position},inner={self.inner})"
+
+    def __repr__(self):
+        return self.__str__()
+
+    def with_inner(self, accessor: "Accessor") -> "Accessor":
```
rdblue, line 283: Is there a better name for this method? It sounds like it should replace this accessor's `inner` field, but it actually sets the leaf `inner`. Maybe `with_inner_most`?

python/src/iceberg/files.py
```diff
@@ -46,8 +46,10 @@ class FileFormat(Enum):
 class StructProtocol(Protocol):  # pragma: no cover
     """A generic protocol used by accessors to get and set at positions of an object"""
 
+    @abstractmethod
```
rdblue, line 49: Is this not implied by `Protocol`?

python/tests/conftest.py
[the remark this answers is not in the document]
rdblue: Unless there is a reason to touch this file, I'd probably prefer to revert this change.

python/src/iceberg/schema.py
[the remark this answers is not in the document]
samredai: I agree that pre-order feels more intuitive, especially since the index gets built in ascending field ID order. That being said, it's not worth it if there are gotchas that we're missing here.

> ...and I think makes the visitor you're building here more complicated because it can't rely on traversal to create the child accessors it needs.

This is a great point--I'm noticing now that the visit method for NestedField here is becoming responsible for much than just visiting the NestedField. I couldn't think of a post-order solution for creating the wrapped accessor after the inner accessor is visited, but I completely missed what we're doing there in `_IndexByName`. I think mimicking that is a good idea and keeping something like a `self._inner` property on the `_BuildPositionAccessors` class.

python/src/iceberg/schema.py
Fokko: We don't need this one anymore :)

python/src/iceberg/schema.py
Fokko: I've reverted this 👍🏻

python/src/iceberg/schema.py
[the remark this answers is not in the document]
Fokko: @rdblue I've seen the light, thank you. Just reverted it to post-order traversal. It took me a while to wrap my head around it, but I like the fact that there is no state in the visitor itself. This makes is much cleaner.

python/tests/test_schema.py
```diff
@@ -347,3 +351,80 @@ def test_schema_find_type(table_schema_simple):
...
+    iceberg_schema = schema.Schema(
+        NestedField(
+            field_id=4,
+            name="qux",
+            field_type=ListType(element_id=5, element_type=StringType(), element_is_optional=True),
+            is_optional=True,
+        ),
+        schema_id=1,
+    )
+    accessors = build_position_accessors(iceberg_schema)
```
Fokko, line 383: @rdblue You mentioned that in the Java implementation there is no accessor for the field list. I've changed this on the Python side as well to have the same behavior.

python/src/iceberg/schema.py
```diff
@@ -40,6 +41,8 @@
 
 T = TypeVar("T")
 
+logger = logging.getLogger(__name__)
```
rdblue, line 44: Looks like this is unused. Do we need to add it?

python/src/iceberg/schema.py
```diff
@@ -301,11 +336,10 @@ def _(obj: StructType, visitor: SchemaVisitor[T]) -> T:
 @visit.register(ListType)
 def _(obj: ListType, visitor: SchemaVisitor[T]) -> T:
     """Visit a ListType with a concrete SchemaVisitor"""
+
     visitor.before_list_element(obj.element)
-    try:
-        result = visit(obj.element.type, visitor)
-    finally:
```
rdblue, line 307: Why remove the try/finally? It seems safer to have it, although I'll admit that it would be weird for an exception to be thrown during operation.

python/src/iceberg/schema.py
```diff
@@ -497,31 +530,78 @@ def index_name_by_id(schema_or_type) -> Dict[int, str]:
...
 
-    def map(self, map_type: MapType, key_result: Dict[int, Accessor], value_result: Dict[int, Accessor]) -> Dict[int, Accessor]:
-        return self._index
+    def struct(self, struct: StructType, field_results: List[Dict[Position, Accessor]]) -> Dict[Position, Accessor]:
+        result = {}
 
-    def primitive(self, primitive: PrimitiveType) -> Dict[int, Accessor]:
-        return self._index
+        for position, field in enumerate(struct.fields):
+            if field_results[position]:
```
rdblue, line 579: Since `list`, `map`, and `primitive` all return a map, is the `else` case used? It is used in the JVM version for all 3 cases, but I think here it is probably just used for the `primitive` case because the accessor map is empty, right?

I think that is correct (if `{}` uses the `else` branch) but I don't think there's a need to return accessors for fields inside maps or lists because we don't really know how to handle repeated elements. We could extend this later so that a repeated layer returns a tuple of the values, but right now we don't really need that.

python/src/iceberg/schema.py
Fokko: No, let's remove it!

python/src/iceberg/schema.py
Fokko: This was very confusing when debugging, and I don't think we should swallow (at least not the broad Exception) those in silence. When visiting an element caused an error, it would actually crash on something else because the state was broken.

python/src/iceberg/schema.py
Fokko: This is Python:
```python
➜  python3
Python 3.9.12 (main, Mar 26 2022, 15:44:31) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> if {}:
...   print('if')
... else:
...   print('else')
... 
else
```

> I think that is correct (if `{}` uses the else branch) but I don't think there's a need to return accessors for fields inside maps or lists because we don't really know how to handle repeated elements. We could extend this later so that a repeated layer returns a tuple of the values, but right now we don't really need that.

So you only want to support structs (including nested structs), but discard support for lists and maps?

python/tests/test_schema.py
Fokko: Correct me if I'm wrong here. But I believe that the `list` itself isn't being returned (id 4 is missing).

python/tests/test_schema.py
```diff
@@ -347,3 +351,80 @@ def test_schema_find_type(table_schema_simple):
...
+    assert accessors == {
+        7: Accessor(position=0, inner=Accessor(position=0, inner=None)),
+        8: Accessor(position=0, inner=Accessor(position=1, inner=None)),
+        9: Accessor(position=0, inner=Accessor(position=1, inner=Accessor(position=0, inner=None))),
+        10: Accessor(position=0, inner=Accessor(position=1, inner=Accessor(position=1, inner=None))),
+    }
+
+
+class TestStructProtocol(StructProtocol):
```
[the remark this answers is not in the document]
Fokko: I agree, but that would be outside of the scope of this PR. @samredai WDYT?

python/tests/test_schema.py
```diff
@@ -347,3 +351,80 @@ def test_schema_find_type(table_schema_simple):
...
+        ),
+        schema_id=1,
+    )
+    accessors = build_position_accessors(iceberg_schema)
+    assert accessors == {
+        7: Accessor(position=0, inner=Accessor(position=0, inner=None)),
+        8: Accessor(position=0, inner=Accessor(position=1, inner=None)),
+        9: Accessor(position=0, inner=Accessor(position=1, inner=Accessor(position=0, inner=None))),
+        10: Accessor(position=0, inner=Accessor(position=1, inner=Accessor(position=1, inner=None))),
+    }
```
Fokko, line 413: @rdblue this is an example for a map. I would expect id 7 to be in there as it is the key to the map. id 8 doesn't make much sense, since it will point at another map. id 9 and 10 should be in there since they are primitives.

python/src/iceberg/schema.py
rdblue: > So you only want to support structs (including nested structs), but discard support for lists and maps?

Yes, because maps and lists aren't really supported. An accessor that is produced will be able to load, for example, a map by position, but the inner accessor for key will load position 0. What does it mean to call `m.get(0)`? It seems to me that we need a different sort of accessor for repeated elements like map keys, map values, and list elements. One that gets the collection and then calls any inner accessor on each item in the collection. We can do that, but Java doesn't actually use it yet so there's not much point to implementing it over here.

python/tests/test_schema.py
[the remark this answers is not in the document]
rdblue: I mean just this test class: `TestStructProtocol` -> `TestStruct`.

python/src/iceberg/schema.py
Fokko: @rdblue Interesting case, and I see your point, just updated the code and extended the test cases 👍🏻

rdblue (approved):

rdblue:
Looks great! Thanks, @Fokko!

python/src/iceberg/schema.py
```diff
@@ -262,10 +262,44 @@ def primitive(self, primitive: PrimitiveType) -> T:
...
+    """An accessor for a specific position in a container that implements the StructProtocol"""
+
+    position: int
+    inner: Optional["Accessor"] = None
+
+    def __str__(self):
+        return f"Accessor(position={self.position},inner={self.inner})"
+
+    def __repr__(self):
+        return self.__str__()
```
rdblue, line 276: Following up on this: should we use `@dataclass(repr=True, ...)` instead of defining this?

PR #4691, opened 2022-05-04
Static Analysis: Mixed Mutability Return Type

felixYyu:
This method returns both mutable and immutable collections or maps from different paths, I think it should `return ImmutableList.of();`

### DIFF
--- a/core/src/main/java/org/apache/iceberg/io/MultiBufferInputStream.java
+++ b/core/src/main/java/org/apache/iceberg/io/MultiBufferInputStream.java
@@ -25,6 +25,7 @@ import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Fokko (commented):
Thanks for looking into this @felixYyu 

Looks like the `Collections.emptyList()` also returns an immutable list: https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#emptyList()

felixYyu:
Static Analysis waring, I think it better ImmutableList.of() @Fokko ```
iceberg\core\src\main\java\org\apache\iceberg\io\MultiBufferInputStream.java:179: [MixedMutabilityReturnType] This method returns both mutable and immutable collections or maps from different paths. This may be confusing for users of the method.
  public List<ByteBuffer> sliceBuffers(long len) throws EOFException {
                          ^
    (see https://errorprone.info/bugpattern/MixedMutabilityReturnType)
  Did you mean 'return ImmutableList.of();'?
```

PR #4693, opened 2022-05-04
Core - Add Parser for UpdateRequirement

kbendick:
We need a parser for `UpdateTableRequest.UpdateRequirement` class.

- Adds `UpdateRequirementParser`
- Adds json serializer / deserializer to the RESTSerializers class
- Adds tests for all of the individual implementations of `UpdateRequirement`, similar to `MetadataUpdate` tests.

Open questions:
- Some of the field names don't necessarily line up with the spec.
  - I've noted them on the PR. I was initially going to change them, but they are intertwined with some other classes (namely `MetadataUpdate`) and would also possibly cause shadowing issues in some cases.
  - The biggest source of confusion is `sortOrderId` vs `default-write-order-id`, though looking at the spec, the corresponding `MetadataUpdate` field is just called `order-id` so maybe it's not a concern.

EDIT:
For the naming, we've decided to keep the specs field name of `ref`, but update `default-write-order-id` and all references in `UpdateRequirement` to use `sort-order-id` instead of `write-order-id`.

There's an associated field in `MetadataUpdate` that's just called `order-id` which I will update in a different PR.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
@@ -46,9 +46,9 @@ public class UpdateRequirementParser {
   // AssertTableUUID
   private static final String UUID = "uuid";
 
-  // AssertRefSnapshotID  // field `name` is ref in the spec
-  private static final String REF = "ref";  // renamed from `name` to match spec
-  private static final String SNAPSHOT_ID = "snapshot-id";  // Long
+  // AssertRefSnapshotID
+  private static final String REF = "ref";
+  private static final String SNAPSHOT_ID = "snapshot-id";
 
   // AssertLastAssignedFieldId
   private static final String LAST_ASSIGNED_FIELD_ID = "last-assigned-field-id";  // int
--- a/core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
@@ -1,0 +1,255 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.iceberg.rest.requests;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonNode;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.UncheckedIOException;
+import java.util.Locale;
+import java.util.Map;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.rest.requests.UpdateTableRequest.UpdateRequirement;
+import org.apache.iceberg.util.JsonUtil;
+
+public class UpdateRequirementParser {
+
+  private UpdateRequirementParser() {
+  }
+
+  private static final String TYPE = "type";
+
+  // assertion types
+  static final String ASSERT_TABLE_UUID = "assert-table-uuid";
+  static final String ASSERT_TABLE_DOES_NOT_EXIST = "assert-create";
+  static final String ASSERT_REF_SNAPSHOT_ID = "assert-ref-snapshot-id";
+  static final String ASSERT_LAST_ASSIGNED_FIELD_ID = "assert-last-assigned-field-id";
+  static final String ASSERT_CURRENT_SCHEMA_ID = "assert-current-schema-id";
+  static final String ASSERT_LAST_ASSIGNED_PARTITION_ID = "assert-last-assigned-partition-id";
+  static final String ASSERT_DEFAULT_SPEC_ID = "assert-default-spec-id`";
+  static final String ASSERT_DEFAULT_SORT_ORDER_ID = "assert-default-write-order-id";
+
+  // AssertTableUUID
+  private static final String UUID = "uuid";  // String
+
+  // AssertTableDoesNotExist - Doesn't have any fields - just checks for null of current TableMetadata.
+  //                           Maybe `base`
+
+  // AssertRefSnapshotID  // field `name` is ref in the spec
+  private static final String REF = "ref";  // renamed from `name` to match spec
+  private static final String SNAPSHOT_ID = "snapshot-id";  // Long
+
+  // AssertLastAssignedFieldId
+  private static final String LAST_ASSIGNED_FIELD_ID = "last-assigned-field-id";  // int
+
+  // AssertCurrentSchemaID  // field is current-schema-id in spec
+  private static final String SCHEMA_ID = "schema-id";  // int
+
+  // AssertLastAssignedPartitionId
+  private static final String LAST_ASSIGNED_PARTITION_ID = "last-assigned-partition-id"; // int
+
+  // AssertDefaultSpecID  // field is default-spec-id in spec
+  private static final String SPEC_ID = "spec-id";  // int
+
+  // AssertDefaultSortOrderID  // field is default-write-order-id in spec
+  private static final String SORT_ORDER_ID = "sort-order-id";  // int - some of these should be long though?
+
+  private static final Map<Class<? extends UpdateTableRequest.UpdateRequirement>, String> TYPES = ImmutableMap
+      .<Class<? extends UpdateTableRequest.UpdateRequirement>, String>builder()
+      .put(UpdateRequirement.AssertTableUUID.class, ASSERT_TABLE_UUID)
+      .put(UpdateRequirement.AssertTableDoesNotExist.class, ASSERT_TABLE_DOES_NOT_EXIST)
+      .put(UpdateRequirement.AssertRefSnapshotID.class, ASSERT_REF_SNAPSHOT_ID)
+      .put(UpdateRequirement.AssertLastAssignedFieldId.class, ASSERT_LAST_ASSIGNED_FIELD_ID)
+      .put(UpdateRequirement.AssertCurrentSchemaID.class, ASSERT_CURRENT_SCHEMA_ID)
+      .put(UpdateRequirement.AssertLastAssignedPartitionId.class, ASSERT_LAST_ASSIGNED_PARTITION_ID)
+      .put(UpdateRequirement.AssertDefaultSpecID.class, ASSERT_DEFAULT_SPEC_ID)
+      .put(UpdateRequirement.AssertDefaultSortOrderID.class, ASSERT_DEFAULT_SORT_ORDER_ID)
+      .build();
+
+  public static String toJson(UpdateRequirement updateRequirement) {
+    return toJson(updateRequirement, false);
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
```diff
@@ -0,0 +1,264 @@
...
+  static final String ASSERT_DEFAULT_SPEC_ID = "assert-default-spec-id";
+  static final String ASSERT_DEFAULT_SORT_ORDER_ID = "assert-default-write-order-id";
+
+  // AssertTableUUID
+  private static final String UUID = "uuid";
+
+  // AssertRefSnapshotID
+  // TODO - This is called `ref` in the spec.
+  //   https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L1359-L1360
+  private static final String NAME = "name";
```
kbendick, line 52: This field, `name`, is the name of the branch or tag. In the spec it's referred to as `ref`, but using `ref` could potentially cause confusion with `TableMetadata.ref` in this code here: https://github.com/apache/iceberg/blob/a19ac7835c8e251fd6a042e056f29ee63262adec/core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java#L261-L297

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
```diff
@@ -0,0 +1,264 @@
...
+  // AssertLastAssignedPartitionId
+  private static final String LAST_ASSIGNED_PARTITION_ID = "last-assigned-partition-id";
+
+  // AssertDefaultSpecID
+  private static final String SPEC_ID = "default-spec-id";
+
+  // AssertDefaultSortOrderID
+  // TODO - Is currently referred to as default-write-order-id in spec but class and comments use sort-order. Need to
+  //  update it in class or in spec
+  private static final String SORT_ORDER_ID = "default-write-order-id";
```
kbendick, line 70: This doesn't necessarily need to be updated in the class or the spec, as we do use `sort-order-id` and `write-order-id` somewhat interchangeably.

I hesitated to update the class, because the associated `MetadataUpdate` class uses `order-id` in the spec, but the classes are all named `SortOrder`. https://github.com/apache/iceberg/blob/a19ac7835c8e251fd6a042e056f29ee63262adec/open-api/rest-catalog-open-api.yaml#L1231-L1240

Happy to update either one. We could also leave these two as not implemented, review the easier parts of this PR, and then add a second PR to handle it in isolation.

kbendick: Happy to update either one. We could also leave these two as not implemented, review the easier parts of this PR, and then add a second PR to handle it in isolation.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
kbendick: If we made this `ref`, we'd just need to rename the current variable `ref` in the class's validate method to `baseRefType` or something: https://github.com/apache/iceberg/blob/a19ac7835c8e251fd6a042e056f29ee63262adec/core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java#L279-L281

open-api/rest-catalog-open-api.yaml
```diff
@@ -1361,7 +1361,8 @@ components:
         uuid:
           type: string
         snapshot-id:
-          type: integer
+          type: interger
+          format: int64
```
kbendick, line 1365: This is a long, so updated the doc to reflect that.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
rdblue: How about `ref-name`?

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
rdblue: I'd be fine renaming this to `assert-default-sort-order-id` and `default-sort-order-id` so that it matches `SetDefaultSortOrder` with `sort-order-id`. We'd need to update the spec.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
rdblue: I'm not sure that this actually needs to change. This parser is the only place where this name is used, so it doesn't really conflict or cause confusion if this is "ref". The `name` field of the assertion could still be set as "ref" in the serialization.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
```diff
@@ -0,0 +1,264 @@
...
+    return new UpdateRequirement.AssertTableDoesNotExist();
+  }
+
+  private static UpdateRequirement readAssertTableUUID(JsonNode node) {
+    String uuid = JsonUtil.getString(UUID, node);
+    return new UpdateRequirement.AssertTableUUID(uuid);
+  }
+
+  private static UpdateRequirement readAssertRefSnapshotId(JsonNode node) {
+    String name = JsonUtil.getStringOrNull(NAME, node);
```
rdblue, line 235: Isn't `ref` (or `name`) required? I would expect this to not allow `null`.

core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateRequirementParser.java
```diff
@@ -0,0 +1,312 @@
...
+    String name = "snapshot-name";
+    Long snapshotId = 1L;
+    String json = String.format("{\"type\":\"%s\",\"name\":\"%s\",\"snapshot-id\":%d}",
+        requirementType, name, snapshotId);
+    UpdateRequirement expected = new UpdateRequirement.AssertRefSnapshotID(name, snapshotId);
+    assertEquals(requirementType, expected, UpdateRequirementParser.fromJson(json));
+  }
+
+  @Test
+  public void testAssertRefSnapshotIdFromJson() {
```
rdblue, line 94: This needs to validate that `snapshot-id` serializes correctly when it is null.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
```diff
@@ -0,0 +1,264 @@
...
+
+  private static void writeAssertTableUUID(UpdateRequirement.AssertTableUUID requirement, JsonGenerator gen)
+      throws IOException {
+    gen.writeStringField(UUID, requirement.uuid());
+  }
+
+  private static void writeAssertRefSnapshotId(UpdateRequirement.AssertRefSnapshotID requirement, JsonGenerator gen)
+      throws IOException {
+    gen.writeStringField(NAME, requirement.refName());
+    gen.writeNumberField(SNAPSHOT_ID, requirement.snapshotId());
```
rdblue, line 196: I think this should test whether the required snapshot ID is null and explicitly write `null` if it is.

rdblue: (This is the only one that can be null)

rdblue (changes requested):
Thanks, @kbendick! This is really close. I think that the only blocker is the `null` handling for `snapshot-id` for `assert-ref-snapshot-id`.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
kbendick: Looking at the class definition, there are `null` checks with an else statement. That's what I based it on:

https://github.com/apache/iceberg/blob/a19ac7835c8e251fd6a042e056f29ee63262adec/core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java#L279-L297

kbendick: Oh wait I read that wrong from when I renamed it to `ref` temporarily. You're right.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
kbendick: I though this would write `null` if the `Long` value were null. I'll write a test to ensure it though.

core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateRequirementParser.java
kbendick: Added the tests. There seems to be a bug in `JsonUtils.getLongOrNull`. Should I open a separate PR for that for the benefit of people who cherry-pick?

The check should be `node.hasNonNull`. I'm going to open a separate PR because there's also issues with the int one I think. 

https://github.com/apache/iceberg/blob/4ae2002bd46bf8e1c20db03cffc6319237e4d74a/core/src/main/java/org/apache/iceberg/util/JsonUtil.java#L71-L79_

kbendick:
It looks like `null` handling is incorrect in `JsonUtil::getLongOrNull`. I have a PR to update it that should likely go in first (though the same changes for `getLongOrNull` are reflected in this PR,  but not the changes for `getIntOrNull`): https://github.com/apache/iceberg/pull/4696

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
kbendick: You're right. `writeNumberField` doesn't handle null properly.

Updated and added a test.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
kbendick: Yeah. I'll just make the `NAME` field as `ref` in the parser.

core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java
kbendick: Ok. I'm somewhat partial to `write-order-id` as the condition is about how the data is written (and not how the data is necessarily read), but `sort-order-id` does match much more of the existing code.

core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateRequirementParser.java
kbendick: Oh and added a test for explicit null.

kbendick:
These tests failed because of a flakey Flink 1.15 test. Several tests in that file needed to be updated for Flink 1.15 as they were order dependent, and Flink no longer is returning the records in a defined ordering - and we have no ordering guarantees per the spec unless we have an `ORDER BY` clause _in the select statement_.

This PR should fix the issue: https://github.com/apache/iceberg/pull/4697

open-api/rest-catalog-open-api.yaml
```diff
@@ -1233,9 +1233,9 @@ components:
         - $ref: '#/components/schemas/BaseUpdate'
         - type: object
           required:
-            - order-id
+            - sort-order-id
           properties:
-            order-id:
+            sort-order-id:
```
kbendick, line 1238: I can update the corresponding fields for `MetadataUpdate` in another PR.

I'm opening a PR for the last `MetadataUpdate` parsers today.

kbendick: I am moving this change to https://github.com/apache/iceberg/pull/4716, as that's where the associated parser is being written.

kbendick:
I've rebased this now that https://github.com/apache/iceberg/pull/4697 has been merged.

rdblue:
Thanks, @kbendick!

PR #4695, opened 2022-05-04
AWS: Add Cache for LakeFormation Credentials

xiaoxuandev:


aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -135,24 +141,57 @@ private LakeFormationClient lakeFormation() {
...
     private String tableArn;
+    private final long expiryLeadTimeForRefreshInMins;
+
+    private LakeFormationTemporaryCredentials cachedCredentials;
+    private final Cache<String, LakeFormationTemporaryCredentials> cache;
 
     LakeFormationCredentialsProvider(LakeFormationClient lakeFormationClient, String tableArn) {
       this.client = lakeFormationClient;
       this.tableArn = tableArn;
+      this.expiryLeadTimeForRefreshInMins = 1;
```
jackye1995, line 152: this should be a config passed in from aws properties

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -135,24 +141,57 @@ private LakeFormationClient lakeFormation() {
...
+      this.cache = Caffeine.newBuilder()
+          .initialCapacity(1)
+          .recordStats()
```
jackye1995, line 155: is record stats needed?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
jackye1995: also we should have units in milliseconds, not minutes

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -135,24 +141,57 @@ private LakeFormationClient lakeFormation() {
...
+            GetTemporaryGlueTableCredentialsRequest.builder()
+                .tableArn(key)
+                // Now only two permission types (COLUMN_PERMISSION and CELL_FILTER_PERMISSION) are supported
+                // and Iceberg only supports COLUMN_PERMISSION at this time
+                .supportedPermissionTypes(PermissionType.COLUMN_PERMISSION)
+                .build();
+        GetTemporaryGlueTableCredentialsResponse response =
+            client.getTemporaryGlueTableCredentials(getTemporaryGlueTableCredentialsRequest);
+
+        AwsSessionCredentials awsSessionCredentials = AwsSessionCredentials.create(response.accessKeyId(),
```
jackye1995, line 177: nit: newline after `create(`

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -135,24 +141,57 @@ private LakeFormationClient lakeFormation() {
...
+    private final long expiryLeadTimeForRefreshInMins;
+
+    private LakeFormationTemporaryCredentials cachedCredentials;
+    private final Cache<String, LakeFormationTemporaryCredentials> cache;
 
     LakeFormationCredentialsProvider(LakeFormationClient lakeFormationClient, String tableArn) {
       this.client = lakeFormationClient;
       this.tableArn = tableArn;
+      this.expiryLeadTimeForRefreshInMins = 1;
+      this.cache = Caffeine.newBuilder()
```
jackye1995, line 153: So this cache is basically used only for its locking feature. This should work, but wondering if there is any better and simpler way.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationTemporaryCredentials.java
```diff
@@ -0,0 +1,42 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.aws.lakeformation;
+
+import java.time.Instant;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+
+public class LakeFormationTemporaryCredentials {
```
jackye1995, line 25: can be moved as an inner class

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -135,24 +141,57 @@ private LakeFormationClient lakeFormation() {
...
-              // and Iceberg only supports COLUMN_PERMISSION at this time
-              .supportedPermissionTypes(PermissionType.COLUMN_PERMISSION)
-              .build();
-      GetTemporaryGlueTableCredentialsResponse response =
-          client.getTemporaryGlueTableCredentials(getTemporaryGlueTableCredentialsRequest);
-      return AwsSessionCredentials.create(response.accessKeyId(), response.secretAccessKey(), response.sessionToken());
+      LakeFormationTemporaryCredentials credentials = cache.getIfPresent(tableArn);
+      if (credentials != null && shouldRefresh(credentials)) {
+        cache.invalidate(tableArn);
+      }
```
singhpk234, line 164: [question] should we use expireAfter(Expiry) for variable TTL based on record instead, while creating the Caffeine obj itself
- can ref https://github.com/ben-manes/caffeine/wiki/Eviction#time-based

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
xiaoxuandev: we are also utilizing cache expiration feature here.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
xiaoxuandev: done. seconds should be enough?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
xiaoxuandev: done. thanks!

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
xiaoxuandev: That's a good point! changed. thanks!

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -132,27 +145,90 @@ private LakeFormationClient lakeFormation() {
...
+            .create(response.accessKeyId(), response.secretAccessKey(), response.sessionToken());
+
+        cachedCredentials = new LakeFormationTemporaryCredentials(awsSessionCredentials, response.expiration());
+        return cachedCredentials;
+      });
+
+      return credentials.credentials();
+    }
+
+    private long getExpiryDuration(LakeFormationTemporaryCredentials value) {
```
amogh-jahagirdar, line 227: In Iceberg, please avoid using getter https://github.com/apache/iceberg/blob/master/CONTRIBUTING.md#style

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -357,6 +357,13 @@ public class AwsProperties implements Serializable {
    */
   public static final String LAKE_FORMATION_DB_NAME = "lakeformation.db-name";
 
+  /**
+   * Used by {@link LakeFormationAwsClientFactory}.
+   * The expiry lead time of the lakeformation credentials in seconds, default to 60 seconds.
+   */
+  public static final String LAKE_FORMATION_EXPIRY_LEAD_SEC = "lakeformation.expiry-lead-time";
+  public static final int LAKE_FORMATION_EXPIRY_LEAD_SEC_DEFAULT = 60;
```
jackye1995, line 365: prefer millisecond as unit, long as data type, `-millis` suffix in config

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -132,27 +145,90 @@ private LakeFormationClient lakeFormation() {
...
-          GetTemporaryGlueTableCredentialsRequest.builder()
-              .tableArn(tableArn)
-              // Now only two permission types (COLUMN_PERMISSION and CELL_FILTER_PERMISSION) are supported
-              // and Iceberg only supports COLUMN_PERMISSION at this time
-              .supportedPermissionTypes(PermissionType.COLUMN_PERMISSION)
-              .build();
-      GetTemporaryGlueTableCredentialsResponse response =
-          client.getTemporaryGlueTableCredentials(getTemporaryGlueTableCredentialsRequest);
-      return AwsSessionCredentials.create(response.accessKeyId(), response.secretAccessKey(), response.sessionToken());
+      LakeFormationTemporaryCredentials credentials = cache.get(tableArn, key -> {
```
jackye1995, line 206: nit: can we make the `key -> { ... }` a helper method?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -132,27 +145,90 @@ private LakeFormationClient lakeFormation() {
...
-    LakeFormationCredentialsProvider(LakeFormationClient lakeFormationClient, String tableArn) {
+    LakeFormationCredentialsProvider(LakeFormationClient lakeFormationClient,
+        String tableArn, long expiryLeadTimeForRefreshInSecs) {
       this.client = lakeFormationClient;
       this.tableArn = tableArn;
+      this.expiryLeadTimeForRefreshInSecs = expiryLeadTimeForRefreshInSecs;
+      this.cache = Caffeine.newBuilder()
+          .expireAfter(new Expiry<String, LakeFormationTemporaryCredentials>() {
+            @Override
+            public long expireAfterCreate(@NonNull String key,
```
jackye1995, line 183: nit: if we cannot fit all arguments on the same line, let's put one argument per line, like:

```
public long expireAfterCreate(
  @NonNull String key,
  @NonNull LakeFormationTemporaryCredentials value, 
  long currentTime) {
  ...
}
```

jackye1995: also I don't think we need these `@NonNull` and `@NonNegative` decorators?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -132,27 +145,90 @@ private LakeFormationClient lakeFormation() {
...
   static class LakeFormationCredentialsProvider implements AwsCredentialsProvider {
     private LakeFormationClient client;
     private String tableArn;
+    private final long expiryLeadTimeForRefreshInSecs;
+
+    private LakeFormationTemporaryCredentials cachedCredentials;
+    private final Cache<String, LakeFormationTemporaryCredentials> cache;
 
-    LakeFormationCredentialsProvider(LakeFormationClient lakeFormationClient, String tableArn) {
+    LakeFormationCredentialsProvider(LakeFormationClient lakeFormationClient,
```
jackye1995, line 175: nit: similar comment as below for putting arguments on newlines

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -401,6 +417,23 @@ public class AwsProperties implements Serializable {
...
+  public static final int LAKE_FORMATION_MAX_CACHE_SIZE_DEFAULT = 100;
+
+  /**
+   * Used by {@link LakeFormationAwsClientFactory}. The expiry time of the cached lakeformation
+   * credentials in milliseconds, default to 30 minutes.
+   */
+  public static final String LAKE_FORMATION_CACHE_EXPIRY_MILLIS =
+      "lakeformation.cache-expiry-millis";
+
+  public static final long LAKE_FORMATION_CACHE_EXPIRY_MILLIS_DEFAULT = 30 * 60000;
```
jackye1995, line 435: nit: 30 * 60 * 1000

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -401,6 +417,23 @@ public class AwsProperties implements Serializable {
    */
   public static final String LAKE_FORMATION_DB_NAME = "lakeformation.db-name";
 
+  /**
+   * Used by {@link LakeFormationAwsClientFactory}. The maximum size of the lakeformation
+   * credentials cache, default to 100.
+   */
+  public static final String LAKE_FORMATION_MAX_CACHE_SIZE = "lakeformation.max-cache-size";
```
jackye1995, line 424: nit: `cache.max-size` and `cache.expiry-millis`?

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -209,18 +209,27 @@ void initialize(
...
+              tableIdentifier, awsProperties.glueCatalogSkipNameValidation()));
+      if (awsProperties.glueLakeFormationEnabled()) {
+        try {
+          glue.getTable(
+              GetTableRequest.builder()
+                  .databaseName(tableIdentifier.namespace().toString())
+                  .name(tableIdentifier.name())
+                  .build());
+          awsProperties.setS3PrefetchClientEnabled(true);
+        } catch (EntityNotFoundException ignored) {
```
jackye1995, line 229: I think we want to catch all possible exceptions here to avoid this path failing, and also log warning

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
jackye1995: actually, I notice we have a config `CatalogProperties.CACHE_EXPIRATION_INTERVAL_MS`, let's use the same name `lakeformation.cache.expiration-interval-ms`

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationTemporaryCredentials.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.aws.lakeformation;
+
+import java.time.Instant;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+
+public class LakeFormationTemporaryCredentials {
```
jackye1995, line 24: let's examine every class to see (1) what class could be package protected instead of `public`, (2) what class needs to implement `Serializable`. We should do a serialization check to make sure a `AwsClientFactory` with this credential is still serializable.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
amogh-jahagirdar: or TimeUnit.MINUTES.toMillis(30);

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -339,6 +339,9 @@ public void initialize(Map<String, String> props) {
         this.credential = ((CredentialSupplier) clientFactory).getCredential();
       }
       this.s3 = clientFactory::s3;
+      if (awsProperties.s3PrefetchClientEnabled()) {
+        client();
+      }
```
amogh-jahagirdar, line 344: Sorry not fully getting why is this needed?

amogh-jahagirdar: Oh nvm, I got it now

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationTemporaryCredentials.java
xiaoxuandev: Done. Added unit tests for LakeFormationAwsClientFactory serialization check.

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/TestLakeFormationMetadataOperations.java
```diff
@@ -165,20 +165,20 @@ public void testCreateTableSuccess() {
...
       grantTablePrivileges(testDbName, testTableName, Permission.DELETE, Permission.DROP);
-      glueCatalogPrivilegedRole.dropTable(TableIdentifier.of(testDbName, testTableName), false);
+      glueCatalogPrivilegedRole.dropTable(TableIdentifier.of(testDbName, testTableName), true);
       lfRegisterPathRoleDeleteDb(testDbName);
     }
   }
 
   @Test
-  public void testDropTableSuccessWhenPurgeIsFalse() {
+  public void testDropTableSuccessWhenPurgeIsTrue() {
```
jackye1995, line 174: nit: can we just say `testPurgeTableSuccess`?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -389,6 +390,21 @@ public class AwsProperties implements Serializable {
    */
   @Deprecated public static final boolean CLIENT_ENABLE_ETAG_CHECK_DEFAULT = false;
 
+  /**
+   * This flag controls whether the S3 client will be initialized during the S3FileIO
+   * initialization, instead of normal lazy initialization upon use. This is needed to deal with a
```
jackye1995, line 395: I don't think we need to explain too much about the use case for LF, this is a generic S3FileIO config. Maybe something like "This is for cases that the credentials to use might change and needs to be pre-loaded"

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -401,6 +417,23 @@ public class AwsProperties implements Serializable {
    */
   public static final String LAKE_FORMATION_DB_NAME = "lakeformation.db-name";
 
+  /**
+   * Used by {@link LakeFormationAwsClientFactory}. The maximum size of the lakeformation
+   * credentials cache, default to 100.
+   */
+  public static final String LAKE_FORMATION_CACHE_MAX_SIZE = "lakeformation.cache-max-size";
```
jackye1995, line 424: nit: `cache.max-size`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -209,18 +210,28 @@ void initialize(
...
+        try {
+          glue.getTable(
+              GetTableRequest.builder()
+                  .databaseName(tableIdentifier.namespace().toString())
+                  .name(tableIdentifier.name())
+                  .build());
+          awsProperties.setS3PrefetchClientEnabled(true);
+        } catch (GlueException e) {
+          awsProperties.setS3PrefetchClientEnabled(false);
+          LOG.warn("Suppressing failure when getting Glue table", e);
```
jackye1995, line 232: logging warning already means suppressing failure, no need to say that again. Maybe `"Fail to get Glue table information, cannot determine S3 prefetch strategy"`?

Also this will hit error very often, especially for new table creation. We should do debug log level.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
```diff
@@ -0,0 +1,151 @@
...
+      return true;
+    }
+
+    if (credentials
+        .sessionCredentialsExpiration()
+        .minus(expiryLeadTimeForRefresh)
+        .isBefore(clock.instant())) {
+      return true;
+    }
+    return false;
```
amogh-jahagirdar, line 58: Nit: maybe this is cleaner, I'm not super opinionated considering it's a longer statement. 

`return credentials
        .sessionCredentialsExpiration()
        .minus(expiryLeadTimeForRefresh)
        .isBefore(clock.instant())`

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -209,18 +210,28 @@ void initialize(
...
+      if (awsProperties.glueLakeFormationEnabled()) {
+        try {
+          glue.getTable(
+              GetTableRequest.builder()
+                  .databaseName(tableIdentifier.namespace().toString())
+                  .name(tableIdentifier.name())
+                  .build());
+          awsProperties.setS3PrefetchClientEnabled(true);
+        } catch (GlueException e) {
+          awsProperties.setS3PrefetchClientEnabled(false);
```
jackye1995, line 231: this is the default, no need to set again

aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java
```diff
@@ -209,18 +210,28 @@ void initialize(
...
+          IcebergToGlueConverter.getTableName(
+              tableIdentifier, awsProperties.glueCatalogSkipNameValidation()));
+      if (awsProperties.glueLakeFormationEnabled()) {
+        try {
+          glue.getTable(
+              GetTableRequest.builder()
+                  .databaseName(tableIdentifier.namespace().toString())
+                  .name(tableIdentifier.name())
+                  .build());
+          awsProperties.setS3PrefetchClientEnabled(true);
```
jackye1995, line 229: I think we need to update `tableSpecificCatalogProperties` instead of setting `awsProperties`, because that is what is used to initialize the file IO

aws/src/main/java/org/apache/iceberg/aws/lakeformation/AssumeRoleRequestIdentity.java
```diff
@@ -0,0 +1,102 @@
...
+    this.roleArn = roleArn;
+    this.sessionName = sessionName;
+    this.externalId = externalId;
+    this.tags = tags;
+    this.region = region;
+    this.timeout = timeout;
+    this.lakeFormationClient = lakeFormationClient;
+  }
+
+  protected String roleArn() {
```
jackye1995, line 52: these methods can be public, as long as the class is protected we are good

aws/src/main/java/org/apache/iceberg/aws/lakeformation/AssumeRoleRequestIdentity.java
```diff
@@ -0,0 +1,102 @@
...
+  protected LakeFormationClient lakeFormationClient() {
+    return lakeFormationClient;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof AssumeRoleRequestIdentity)) {
```
jackye1995, line 85: nit: newline after `if`, `for`, `while`, `try` blocks

aws/src/main/java/org/apache/iceberg/aws/lakeformation/AssumeRoleRequestIdentity.java
```diff
@@ -0,0 +1,102 @@
...
+class AssumeRoleRequestIdentity {
+  private String roleArn;
+  private String sessionName;
+  private String externalId;
+  private Set<Tag> tags;
+  private String region;
+  private int timeout;
+  private LakeFormationClient lakeFormationClient;
+
+  AssumeRoleRequestIdentity(
```
jackye1995, line 35: can be public

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCacheFactory.java
```diff
@@ -0,0 +1,75 @@
...
+    return Optional.ofNullable(instance);
+  }
+
+  synchronized LakeFormationCredentialsCache buildCache(
+      AssumeRoleRequestIdentity assumeRoleRequestIdentity) {
+    try {
+      LakeFormationCredentialsCache cache =
+          caches.get(
+              assumeRoleRequestIdentity,
+              new Function<AssumeRoleRequestIdentity, LakeFormationCredentialsCache>() {
```
jackye1995, line 60: this can be `() -> {}` shorthand

aws/src/test/java/org/apache/iceberg/aws/TestAwsClientFactories.java
```diff
@@ -56,6 +58,17 @@ public void testLoadCustom() {
         "should load custom class", AwsClientFactories.from(properties) instanceof CustomFactory);
   }
 
+  @Test
+  public void testLoadLakeFormationAwsClientFactory() {
+    AwsClientFactory awsClientFactory = new LakeFormationAwsClientFactory();
+
+    byte[] data = SerializationUtil.serializeToBytes(awsClientFactory);
+    AwsClientFactory expectedAwsClientFactory = SerializationUtil.deserializeFromBytes(data);
+    Assert.assertTrue(
```
jackye1995, line 67: this is guaranteed if serialization roundtrip succeeds. We just need to run L65-66 and make sure it does not fail.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationTemporaryCredentials.java
```diff
@@ -0,0 +1,40 @@
...
+import java.time.Instant;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+
+class LakeFormationTemporaryCredentials {
+  private final AwsCredentials credentials;
+  private final Instant sessionCredentialsExpiration;
+
+  LakeFormationTemporaryCredentials(AwsCredentials credentials, Instant expiration) {
+    this.credentials = credentials;
+    this.sessionCredentialsExpiration = Instant.ofEpochMilli(expiration.toEpochMilli());
```
jackye1995, line 30: why do we need to create a copy?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
```diff
@@ -0,0 +1,151 @@
...
+  public void cleanUp() {
+    cache.cleanUp();
+  }
+
+  @Override
+  public @NonNull Policy<String, LakeFormationTemporaryCredentials> policy() {
+    return null;
+  }
+
+  public void refreshIfNecessary(String key) {
```
jackye1995, line 91: This can be a part of the logic inside `get`? I don't see it called in other places

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
```diff
@@ -0,0 +1,151 @@
...
+class LakeFormationCredentialsCache
+    implements LoadingCache<String, LakeFormationTemporaryCredentials> {
+  private final LoadingCache<String, LakeFormationTemporaryCredentials> cache;
+  private final Clock clock;
+  private final Duration expiryLeadTimeForRefresh;
+
+  LakeFormationCredentialsCache(LoadingCache<String, LakeFormationTemporaryCredentials> cache) {
+    this.cache = Preconditions.checkNotNull(cache, "Cache cannot be null");
+    this.clock = Clock.systemUTC();
+    this.expiryLeadTimeForRefresh = Duration.ofMillis(TimeUnit.MINUTES.toMillis(1));
```
jackye1995, line 44: this should be a config

jackye1995:
The caching logic is quite complex, I would suggest separate the S3 prefetch client feature to a separated PR if possible.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/CachingLakeFormationCredentialsProvider.java
```diff
@@ -0,0 +1,48 @@
...
+
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+
+class CachingLakeFormationCredentialsProvider implements AwsCredentialsProvider {
+  private final LakeFormationCredentialsCache cache;
+  private final String tableArn;
+  private final AssumeRoleRequestIdentity assumeRoleRequestIdentity;
+
+  CachingLakeFormationCredentialsProvider(
```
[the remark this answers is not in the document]
xiaoxuandev: Checkstyle complains it being redundant if keep it public.

aws/src/test/java/org/apache/iceberg/aws/TestAwsClientFactories.java
xiaoxuandev: So you suggest to remove this assertion? I feel it's better to still verify something.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
xiaoxuandev: done. thanks!

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCacheFactory.java
xiaoxuandev: done. thanks!

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
xiaoxuandev: Fixed, thanks!

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
xiaoxuandev: done. thanks!

aws/src/integration/java/org/apache/iceberg/aws/lakeformation/TestLakeFormationMetadataOperations.java
```diff
@@ -165,20 +165,20 @@ public void testCreateTableSuccess() {
...
     } finally {
       grantTablePrivileges(testDbName, testTableName, Permission.DELETE, Permission.DROP);
-      glueCatalogPrivilegedRole.dropTable(TableIdentifier.of(testDbName, testTableName), false);
+      glueCatalogPrivilegedRole.dropTable(TableIdentifier.of(testDbName, testTableName), true);
       lfRegisterPathRoleDeleteDb(testDbName);
     }
   }
 
   @Test
-  public void testDropTableSuccessWhenPurgeIsFalse() {
```
jackye1995, line 174: I think we need test for both `testDropTableSuccess` and `testPurgeTableSuccess`

aws/src/test/java/org/apache/iceberg/aws/TestAwsClientFactories.java
jackye1995: I see, yeah that works for me

aws/src/main/java/org/apache/iceberg/aws/lakeformation/AssumeRoleRequestIdentity.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+package org.apache.iceberg.aws.lakeformation;
+
+import java.util.Objects;
+import java.util.Set;
+import software.amazon.awssdk.services.lakeformation.LakeFormationClient;
+import software.amazon.awssdk.services.sts.model.Tag;
+
+class AssumeRoleRequestIdentity {
+  private String roleArn;
```
jackye1995, line 27: these variables can all be final

aws/src/main/java/org/apache/iceberg/aws/lakeformation/AssumeRoleRequestIdentity.java
```diff
@@ -0,0 +1,104 @@
...
+ * under the License.
+ */
+package org.apache.iceberg.aws.lakeformation;
+
+import java.util.Objects;
+import java.util.Set;
+import software.amazon.awssdk.services.lakeformation.LakeFormationClient;
+import software.amazon.awssdk.services.sts.model.Tag;
+
+class AssumeRoleRequestIdentity {
```
jackye1995, line 26: this name seems to be not describing what it does, looks like it's describing a LakeFormation caller, so maybe just `LakeFormationIdentity`? or `LakeFormationCallerIdentity`?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -50,13 +45,17 @@
...
 
   private String dbName;
   private String tableName;
   private String glueCatalogId;
   private String glueAccountId;
+  private int maxCacheSize;
+  private long cacheExpirationInMillis;
+
+  private long cacheExpiryLeadTimeInMillis;
+  private AssumeRoleRequestIdentity assumeRoleRequestIdentity;
```
jackye1995, line 58: nit: can use shorter name `identity`

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -410,6 +410,32 @@ public class AwsProperties implements Serializable {
...
+  public static final String LAKE_FORMATION_CACHE_EXPIRATION_INTERVAL_MS =
+      "lakeformation.cache.expiration-interval-ms";
+
+  public static final long LAKE_FORMATION_CACHE_EXPIRATION_INTERVAL_MS_DEFAULT = 30 * 60 * 1000;
+
+  /**
+   * Used by {@link LakeFormationAwsClientFactory}. The expiry lead time of the lakeformation
+   * credentials cache, default to 1 minute.
+   */
+  public static final String LAKE_FORMATION_CACHE_EXPIRY_LEAD_TIME_MS =
```
jackye1995, line 434: the name and config name is not matching, `REFRESH_LEAD_TIME_MS`?

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationAwsClientFactory.java
```diff
@@ -50,13 +45,17 @@
...
 public class LakeFormationAwsClientFactory extends AssumeRoleAwsClientFactory {
-
   public static final String LF_AUTHORIZED_CALLER = "LakeFormationAuthorizedCaller";
 
   private String dbName;
   private String tableName;
   private String glueCatalogId;
   private String glueAccountId;
+  private int maxCacheSize;
+  private long cacheExpirationInMillis;
```
jackye1995, line 55: nit: `cacheExpirationInMillis` and `cacheExpiryLeadTimeInMillis` should match the config name, like `cacheRefreshLeadTimeMs`

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCache.java
```diff
@@ -0,0 +1,146 @@
...
+    return cache.asMap();
+  }
+
+  @Override
+  public void cleanUp() {
+    cache.cleanUp();
+  }
+
+  @Override
+  public @NonNull Policy<String, LakeFormationTemporaryCredentials> policy() {
```
jackye1995, line 89: are these decorators needed? You return null anyway for a non-null method

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCacheFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+class LakeFormationCredentialsCacheFactory {
+  private final Cache<LakeFormationIdentity, LakeFormationCredentialsCache> caches;
+
+  private static LakeFormationCredentialsCacheFactory instance;
```
jackye1995, line 30: static variable should be before instance variable, and have capitalized variable name.

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationCredentialsCacheLoader.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public LakeFormationTemporaryCredentials load(String key) {
+    GetTemporaryGlueTableCredentialsRequest getTemporaryGlueTableCredentialsRequest =
+        GetTemporaryGlueTableCredentialsRequest.builder()
+            .tableArn(key)
+            // Now only two permission types (COLUMN_PERMISSION and CELL_FILTER_PERMISSION) are
+            // supported
+            // and Iceberg only supports COLUMN_PERMISSION at this time
```
jackye1995, line 43: nit: can be at the same line as L42

aws/src/main/java/org/apache/iceberg/aws/lakeformation/LakeFormationIdentity.java
```diff
@@ -0,0 +1,104 @@
...
+
+  public String region() {
+    return region;
+  }
+
+  public int timeout() {
+    return timeout;
+  }
+
+  public LakeFormationClient lakeFormationClient() {
```
jackye1995, line 76: nit: can just be `client()`

PR #4696, opened 2022-05-05
Core: Fix invalid null handling in JsonUtils::getIntOrNull ::getLongOrNull

kbendick:
Using `JsonUtils::getLongOrNull` in https://github.com/apache/iceberg/pull/4693, it seems that the functions `getIntOrNull` and `getLongOrNull` do not properly handle an explicit null.

As you can see in the `Preconditions` check, after getting the property, if the JsonNode of the nullable Integer or Long is explicitly `null`, the Precondition check will fail.

cc @rdblue @RussellSpitzer


danielcweeks (approved):

rdblue:
+1

Good catch, @kbendick!

PR #4699, opened 2022-05-05
Flink: improve unit tests for sink

liuml07:
Initially I found the `TestIcebergFilesCommitter.java` always use the parquet file format for delete files. I think why not test with the parameterized file format as the data file does. Then I found the Flink sink tests maintain and pass variable `tablePath` unnecessarily. It can be replaced with `table` when comparing results, and can be replaced with `table.location()` when the absolute path is required.

I think the fix is straightforward enough so I consolidate the changes in this PR:
- Make sure the committer test use parameterized file format
- Simply test variables by removing `tablePath`

### DIFF
--- a/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestIcebergFilesCommitter.java
+++ b/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestIcebergFilesCommitter.java
@@ -74,7 +74,6 @@ import static org.apache.iceberg.flink.sink.ManifestOutputFileFactory.FLINK_MANI
 public class TestIcebergFilesCommitter extends TableTestBase {
   private static final Configuration CONF = new Configuration();
 
-  private String tablePath;
   private File flinkManifestFolder;
 
   private final FileFormat format;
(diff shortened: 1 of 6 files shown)


flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestIcebergFilesCommitter.java
```diff
@@ -743,14 +740,13 @@ public void testCommitTwoCheckpointsInSingleTxn() throws Exception {
 
   private DeleteFile writeEqDeleteFile(FileAppenderFactory<RowData> appenderFactory,
                                        String filename, List<RowData> deletes) throws IOException {
-    return SimpleDataUtil.writeEqDeleteFile(table, FileFormat.PARQUET, tablePath, filename, appenderFactory, deletes);
```
liuml07, line 746: This should not be always `PARQUET` but `format`.

stevenzwu (approved):

hililiwei (approved):

rdblue (approved):

PR #4700, opened 2022-05-05
Nessie: Bump Nessie to 0.29.0 / Use ref.hash parameter to read data at given hash

nastra:
This makes sure that the `NessieCatalog` reads data at the given
`ref.hash` (`NessieConfigConstants.CONF_NESSIE_REF_HASH`) when it's provided. If `ref.hash` is `null`, then this means
that data should be read from whatever the latest `HEAD` is.


versions.props
```diff
@@ -24,7 +24,7 @@ javax.xml.bind:jaxb-api = 2.3.1
 javax.activation:activation = 1.1.1
 org.glassfish.jaxb:jaxb-runtime = 2.3.3
 software.amazon.awssdk:* = 2.17.131
-org.projectnessie:* = 0.28.0
+org.projectnessie:* = 0.28.1-SNAPSHOT
```
nastra, line 27: this needs to be updated to 0.29.0 once it's released

nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java
```diff
@@ -87,7 +87,9 @@ public void refresh() throws NessieNotFoundException {
   }
 
   public NessieIcebergClient withReference(String requestedRef, String hash) {
-    if (null == requestedRef) {
+    if (null == requestedRef ||
+        getRef().getReference().getName().equals(requestedRef) ||
+        getRef().getHash().equals(hash)) {
```
snazy, line 92: Sure that you want equals on ref-name _OR_ hash? I.e. still return `this`, if the branch name doesn't change?

nastra: at this point yes, because Trino will make heavier use of this method and we don't want to do another REST call to fetch the branch if it didn't change

ajantha-bhat: > at this point yes, because Trino will make heavier use of this method and we don't want to do another REST call to fetch the branch if it didn't change

but functionally it is wrong in case of "drop a reference and create new reference with same name but different hash/source"

Can we make trino pass the hash as well ?

ajantha-bhat (approved):
Please add the version bump info in PR description or header

RussellSpitzer (approved):
Lgtm

PR #4579, opened 2022-04-18
Spark 3.x: Add option to introduce ordering of RewriteFileGroup

rajarshisarkar:
This PR backports https://github.com/apache/iceberg/pull/4377 to Spark 3.1 and Spark 3.0. 

The PR also includes the changes discussed here: https://github.com/apache/iceberg/pull/4377#discussion_r851419052 

---
cc: @RussellSpitzer @aokolnychyi

PR #4701, opened 2022-05-05
Spark 3.x: Support rewrite data files with starting sequence number

rajarshisarkar:
I noticed that `use-starting-sequence-number` option is missing while developing https://github.com/apache/iceberg/pull/4377 and https://github.com/apache/iceberg/pull/4579. Adding support for the same in this PR.

When `use-starting-sequence-number` is enabled then compaction should use the sequence number of the snapshot at compaction start time for new data files, instead of using the sequence number of the newly produced snapshot. This avoids commit conflicts with updates that add newer equality deletes at a higher sequence number.

---

cc: @RussellSpitzer @jackye1995 @szehon-ho


RussellSpitzer (approved):

RussellSpitzer:
Thanks @rajarshisarkar !

PR #4702, opened 2022-05-05
Avoid possible NPE in Writer close

felixYyu:
cc: @rdblue @RussellSpitzer

### DIFF
--- a/core/src/main/java/org/apache/iceberg/ManifestWriter.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestWriter.java
@@ -178,8 +178,10 @@ public abstract class ManifestWriter<F extends ContentFile<F>> implements FileAp
 
   @Override
   public void close() throws IOException {
-    this.closed = true;
-    writer.close();
+    if (writer != null) {
+      writer.close();
+      this.closed = true;
+    }
   }
 
   static class V2Writer extends ManifestWriter<DataFile> {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/ManifestWriter.java
```diff
@@ -178,8 +178,10 @@ public ManifestFile toManifestFile() {
 
   @Override
   public void close() throws IOException {
-    this.closed = true;
-    writer.close();
+    if (writer != null) {
+      writer.close();
+      this.closed = true;
```
kbendick, line 183: Nit: This should probably still set `this.closed = true` even if the writer is null.

kbendick (commented):
Is it possible for any of these writers to actually be `null`? It seems like they're all created in their respective constructors but maybe you've encountered something I haven't.

felixYyu:
Refer to other writing methods, Possible NPE risks, I think we should try to avoid it.
e.g. AvroFileAppender
```
@Override
    if (writer != null) {
      writer.close();
      this.writer = null;
      isClosed = true;
    }
  }
```

hililiwei (commented):
As @kbendick says, it's unlikely that it will be Null (at least I haven't encountered it yet), but for serious reasons, I'm in favor of this PR.

rdblue:
These aren't going to be null because they're set in the constructor and don't change. I don't see much value in changes like this.

felixYyu:
@rdblue 

> These aren't going to be null because they're set in the constructor and don't change. I don't see much value in changes like this.

OK, I'll close this PR.

PR #4704, opened 2022-05-05
Spark: Support parallel delete files when abort

ConeyLiu:
This patch adds support for parallel delete files when commit aborting.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
@@ -43,8 +43,4 @@ public class SparkSQLProperties {
   // Controls whether to check the order of fields during writes
   public static final String CHECK_ORDERING = "spark.sql.iceberg.check-ordering";
   public static final boolean CHECK_ORDERING_DEFAULT = true;
-
-  // Control whether to parallel delete files when abort
-  public static final String DELETE_FILES_PARALLEL_WHEN_ABORT = "spark.sql.iceberg.delete-files-parallel-when-abort";
-  public static final boolean DELETE_FILES_PARALLEL_WHEN_ABORT_DEFAULT = true;
 }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
@@ -43,4 +43,8 @@ public class SparkSQLProperties {
   // Controls whether to check the order of fields during writes
   public static final String CHECK_ORDERING = "spark.sql.iceberg.check-ordering";
   public static final boolean CHECK_ORDERING_DEFAULT = true;
+
+  // Control whether to parallel delete files when abort
+  public static final String DELETE_FILES_PARALLEL_WHEN_ABORT = "spark.sql.iceberg.delete-files-parallel-when-abort";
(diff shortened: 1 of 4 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
```diff
@@ -43,4 +43,8 @@ private SparkSQLProperties() {
   // Controls whether to check the order of fields during writes
   public static final String CHECK_ORDERING = "spark.sql.iceberg.check-ordering";
   public static final boolean CHECK_ORDERING_DEFAULT = true;
+
+  // Control whether to parallel delete files when abort
+  public static final String DELETE_FILES_PARALLEL_WHEN_ABORT = "spark.sql.iceberg.delete-files-parallel-when-abort";
```
rdblue, line 48: The name "delete-files-parallel-when-abort" is very long. How about "parallel-abort-enabled"?

Does this even need to be an option? Why would anyone turn it off?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -204,6 +205,7 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
   private void abort(WriterCommitMessage[] messages) {
     Map<String, String> props = table.properties();
     Tasks.foreach(files(messages))
+        .executeWith(writeConf.deleteFilesParallelWhenAbort() ? ThreadPools.getWorkerPool() : null)
```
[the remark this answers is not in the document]
rdblue: I'd probably use the worker pool for now. No need to extend it until someone needs it.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
ConeyLiu: @rdblue I remove the option. Previously I want to have the ability to keep the behavior as before.

ConeyLiu:
Thanks, @singhpk234 @rdblue for the reviewing, the code has been updated. Please take another look.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -213,6 +214,7 @@ private void abort(WriterCommitMessage[] messages) {
     if (cleanupOnAbort) {
       Map<String, String> props = table.properties();
       Tasks.foreach(files(messages))
+          .executeWith(ThreadPools.getWorkerPool())
```
rdblue, line 217: @RussellSpitzer, what do you think about this change?

RussellSpitzer: Seems fine to me, although I really wish we had a more discrete pool for things like this. I find that our worker pool is hard to configure for end users.

RussellSpitzer (approved):

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
rdblue: Should we add a Spark-managed pool? We have the ability to plug in the pool used in most places now, thanks to @yittg.

rdblue:
Thanks, @ConeyLiu!

ConeyLiu:
Thanks all for the review.

Issue #3245, opened 2021-10-08
[Python] support iceberg base catalog in python library


jun-he:
Seems we can use ABC for this.

samredai:
Here are some additional notes/details/questions to guide this issue. I've also included a diagram of the java implementation.

- Catalog abstraction
  - Operations
    - List tables
    - Create a table
    - Drop a table
    - Replace table
    - Load table (returns an instance of the main table class that users will interact with)
    - Rename a table
    - Check that a table exists (or maybe this is done in load table. Python users will be familiar with `try catalog.load_table() except MissingTableException`)
- Table identifier (maybe not needed?)
  - In python, a table identifier could just be a string and the main Table class itself could also serve as the table identifier. `Table(“foo.bar.baz”, catalog=Catalog(...))` or `Table(namespace=”foo”, database=”bar”, table=”baz”, catalog=Catalog(...))`. Loading the table from the catalog would then be `Table.load()`.
  - “Loading a table” from the catalog could easily be confused with loading the actual data. Since we’ll have both of these functionalities in a single library, should we have a more explicit class name here so it’s not confused with whatever the particular in-memory table format is that’s being used, i.e. pyarrow.Table, pandas.DataFrame, numpy.ndarray, etc.? Some options:
    - IcebergTable
    - TableMetadata
    - TableInfo
- Namespace: Does this need to be a class? Can the `catalog.listNamespaces` method just take a string as an argument?
- Should the JDBC catalog implementation be ranked higher in priority than the hive catalog?

Java implementation:

![catalog](https://user-images.githubusercontent.com/43911210/152832960-77afbccd-02f3-456e-8863-bd8988d02afe.png)

cccs-eric:
@samredai Is there a plan to support a filesystem-based catalog?  Local filesystem, AWS S3 buckets or Azure datalake?  Or that use case would be cover by `Table.load(locationUri)` assuming you know the location URI?

dhruv-pratap:
@jun-he Can you please assign this task to me?

I had a discussion with @samredai and @cabhishek on this task today, and we feel it is a nice candidate work-stream to kick-start  in parallel to all the others as catalog would be a nice convergence point for all the other work-streams to integrate into and test. 

To summarize our discussion, we would be breaking this larger effort into smaller chunks as listed below:

- [ ] Base Catalog with just table interfaces
- [ ] In-memory catalog implementation (Dictionary based)
- [ ] Add Transaction interfaces to Base Catalog
[...]

Issue #3227, opened 2021-10-06
[Python] support table in iceberg python library


jun-he:

xinbinhuang:
@jun-he I want to take a stab at this if it doesn't have dependencies on other issues. Can you assign it to me?

jun-he:
SG, thanks for the help. BTW, table depends on the partition spec and schema, which are ongoing work.

xinbinhuang:
> SG, thanks for the help. BTW, table depends on the partition spec and schema, which are ongoing work.

Understood, I'm keeping an eye on those

dhruv-pratap:
@xinbinhuang I have some bandwidth available right now and I was thinking of picking up this issue. Are you cool if I can get going on this?

xinbinhuang:
@dhruv-pratap feel free to pick it up!

dhruv-pratap:
This tracker is essentially an epic and needs to be broken down further into smaller pieces. We need to break this down into the below pieces to begin with:

1. Define a `TableBuilder` API to create a Table without `Transaction` to begin with.
2. Create a table using `TableBuilder` API without `Transaction`.
 a. With table identifier and schema
 b. With PartitionSpec 
 c. With SortOrder
 d. With Location
 e. With Properties
3. Load a table which returns the `Table` instance with the current table spec.
4. Define `Transaction` API for performing multiple updates to a table in a transaction.
5. Extend the `TableBuilder` API to start create/replace transactions.
6. Create a table in transaction using `Transaction` API
7. Replace a table in transaction using `Transaction` API
8. Create or Replace a table in transaction using `Transaction` API

Class Diagrams:
![TableBuilder](https://user-images.githubusercontent.com/11399509/169559989-40519b91-f423-41cc-add6-212c07d2e2f1.png)

Fokko:
I'll pick this up if that's okay. I'm working on the REST Catalog, and need to implement this anyway

dhruv-pratap:
Hi @Fokko that should be okay, please go ahead. Apologies I couldn't get to this as some work related stuff came up at my end. I'll sync up with you once I circle back to this to see if there are any left overs from this item that I can pick-up. Thanks!

Fokko:
@dhruv-pratap 👍🏻  I think we can close this, we have a table structure: https://github.com/apache/iceberg/blob/master/python/pyiceberg/table/__init__.py#L32

PR #4706, opened 2022-05-05
Python: Support iceberg base catalog in python library (#3245)

dhruv-pratap:
**READY FOR REVIEW.** The current PR has changes for the base Catalog with table and namespace APIs. It also has a concrete In-Memory Catalog implementation, that is Dictionary based, for testing purposes. It also helps us validate our Catalog interface and see if it holds firm in action. 

The PR also includes placeholders for Table and PartitionSpec, and minutely touches issue #3227 i.e., support table in iceberg python library, and #3228 i.e. support partition spec in iceberg python library, as the catalog management is dependent on those classes.


python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+        :param schema: Table's schema
+        :param partition_spec: A partition spec for the table
+        :param location: a location for the table; Optional Keyword Argument
+        :param properties: a string dictionary of table properties; Optional Keyword Argument
+        :return: the replaced table instance
+        :raises TableNotFoundError: If a table with the name does not exist
+        """
+
+
+class NamespacedCatalog(Catalog):
```
samredai, line 133: Do we need to have this as a subclass? I know the JDBC catalog has been updated recently to include namespaces. If the ideal implementation of a catalog should support namespaces, then I'd vote for adding these abstractmethods to `Catalog(ABC)` and an implementation that does not support namespaces will have to explicitly override it and raise some kind of error.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+        schema: Schema,
+        partition_spec: PartitionSpec,
+        *,
+        location: Optional[str] = None,
+        properties: Optional[dict] = None
+    ) -> Table:
+        """
+        Create a table
+
+        :param name: Table's name. Fully classified table name, if it is a namespaced catalog.
```
samredai, line 67: For consistency with the other docstrings, can you switch this to [google style](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) instead of sphinx style? (So using `Args:`)

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+
+        :return: list of table names in the catalog.
+        """
+
+    @abstractmethod
+    def create_table(
+        self,
+        name: str,
+        schema: Schema,
+        partition_spec: PartitionSpec,
```
samredai, line 59: This should be optional to account for unpartitioned table creation, right? Actually thinking about this more, information like name, schema, and partition spec should be contained in an instance of `Table`--would it be better here if these table related methods just took a `Table` instance?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+        List tables under the given namespace in the catalog. If namespace not provided, will list all tables in the
+        catalog.
+
+        :param namespace: the namespace to search
+        :return: list of table names under this namespace.
+        :raises NamespaceNotFoundError: If no such namespace exist
+        """
+
+    @abstractmethod
+    def list_namespaces(self, namespace: Optional[str] = None) -> list:
```
samredai, line 172: I'm not sure if this is from the java client but I can't think of an example of when you would provide a namespace here. I might be missing something but I'm imagining that `catalog.list_namespaces()` would always be called with no args and return all namespaces in the catalog.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        List namespaces from the given namespace. If not given, list top-level namespaces from the catalog.
+
+        :param namespace: given namespace
+        :return: a List of namespace names
+        """
+
+    @abstractmethod
+    def get_namespace_metadata(self, namespace: str) -> dict:
```
samredai, line 181: I'm not familiar with namespace metadata, is this general metadata that's persisted in the catalog?

python/src/iceberg/catalog/base.py
dhruv-pratap: Tell me about it :) I was torn on this one as well as I too strongly felt that `Catalog(ABC)` and `NamespacedCatalog(Catalog)` could be just one. I could not find even a single catalog implementation in the Java API that did NOT support namespaces. All of them do, so it seemed like a redundant extensibility being offered.

python/src/iceberg/catalog/base.py
dhruv-pratap: Yep, that was in the Java client. I think it to narrow dow your list under a certain level. But I'll get rid of it since it sounds like it is not a prominent use-case, if needed the API consumer user can always list all namespaces and filter on namespace levels.

python/src/iceberg/catalog/base.py
dhruv-pratap: It is different from catalog metadata. Some (or maybe all) catalogs maintain metadata at a namespace level as well. For instance, JDBC catalog maintains it's namespace level metadata in the table [`iceberg_namespace_properties`](https://github.com/apache/iceberg/blob/96a46c5caa841777d6fff62a17f09dcc6929603d/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L81).

python/src/iceberg/catalog/base.py
dhruv-pratap: Agreed, they should ideally be taking a `TableBuilder` instance which has the table spec and the method returns  the created 'Table' instance. In my mental model, `TableBuilder` is a spec that has all the information required to create a 'Table', and 'Table' is something that holds the existing table state. Are you thinking on similar lines as well?

python/src/iceberg/catalog/base.py
dhruv-pratap: Ack, will do.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def properties(self) -> dict:
+        return self._properties
+
+    @abstractmethod
+    def list_tables(self) -> list:
```
Fokko, line 47:
```suggestion
    def list_tables(self) -> list[Table]:
```

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        List tables in the catalog.
+
+        :return: list of table names in the catalog.
+        """
+
+    @abstractmethod
+    def create_table(
+        self,
+        name: str,
```
Fokko, line 57: Instead of having a name and a separate schema, we could also accept a path `Tuple[str, ...]`.

Fokko (commented):
Great start!

python/src/iceberg/catalog/base.py
dhruv-pratap: Thanks for the suggestion Fokko! Although the built-in collection typing, as suggested above, only works in Python 3.9 and above. But I see where you are going with this, so let me return a `typing.List` instead. Ex: `List[Table]`

python/src/iceberg/catalog/base.py
dhruv-pratap: Added all the namespace methods to just `Catalog(ABC)`. Resolving this.

python/src/iceberg/catalog/base.py
dhruv-pratap: Added collection typing for all catalog APIs now. Thanks!

python/src/iceberg/catalog/base.py
kbendick: In Java, we use a dedicated Unpartitioned spec.

But I agree that this should be optional and then the builder (or this function… either way) should default to Unpartitioned if this is None (implicitly using the Unpartitioned spec).

python/src/iceberg/catalog/base.py
dhruv-pratap: Removed the argument in the recent commit.

python/src/iceberg/catalog/base.py
dhruv-pratap: Added a frozen `TableSpec` class in a recent commit to hold user provided table specifications to create or replace a table. Also, I made partition_spec optional in there.

python/src/iceberg/catalog/base.py
Fokko: Love it, thanks!

python/src/iceberg/table/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+from abc import ABC
+from typing import Dict
+
+from attrs import Factory, field, frozen
+
+from iceberg.schema import Schema
+
+
+@frozen(kw_only=True)
```
samredai, line 28: Since we're not using runtime type validation, we can use dataclasses here which is basically a slimmed down version of attrs that's been added to the standard library (might even be by the same author although I'm not 100% sure).
```py
from dataclasses import dataclass

@dataclass(kw_only=True)
class TableSpec:
    ...
```

python/src/iceberg/table/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+    TODO: Validators to be added
+    """
+
+    _namespace: str
+    _name: str
+    _schema: Schema
+    _location: str = field()
+    _partition_spec: PartitionSpec = field()
+    _properties: Dict[str, str] = Factory(Dict[str, str])
```
samredai, line 53: Is the leading underscore needed here since freezing the properties is handled by attrs/dataclasses?

dhruv-pratap: You're right. It's not actually required. Will remove it in the next commit. Thanks!

python/src/iceberg/table/base.py
dhruv-pratap: I did start with that initially but thought we would eventually need to add validators for every attribute later on to validate the table spec. That's why went with attrs. I have added a TODO in the code as well to signify the same.

python/src/iceberg/catalog/in_memory.py
```diff
@@ -0,0 +1,106 @@
...
+#  specific language governing permissions and limitations
+#  under the License.
+
+from typing import Dict, List, Optional
+
+from iceberg.catalog.base import AlreadyExistsError, Catalog, TableNotFoundError
+from iceberg.table.base import Table, TableSpec
+
+
+class InMemoryCatalog(Catalog):
```
samredai, line 24: Thanks for adding this! This will be great for demos/quickstarts/tests. If it's not too much would you mind breaking this out into a separate PR? I think there's value in getting just the catalog interface merged in asap so different implementations can be worked on in parallel.

dhruv-pratap: I am done with the In-memory catalog implementation and have committed it within this PR. I think it is important to test the Catalog interface with a concrete implementation to see if it holds firm, and will also help the reviewer visualize the interface in action better. I'll mark this PR ready for review now :)

python/setup.cfg
```diff
@@ -43,6 +43,7 @@ package_dir =
 packages = find:
 python_requires = >=3.7
 install_requires =
+    attrs
```
kbendick, line 46: Is there an alternative in the standard library we can use?

The fewer libraries we add as a dependency, the more we can allow users to pick their own versions and not have to worry about conflicts.

dhruv-pratap: There is Python built-in [dataclasses support](https://docs.python.org/3/library/dataclasses.html), which essentially was influenced by the [attrs community](https://www.attrs.org/en/stable/why.html) but it always seems to lack behind in terms of features. Unfortunately, pre-Python 3.10 dataclasses isn't as powerful as attrs. For instance, it doesn't support "kw_only" out of the box and requires us to write custom decorators to support that.

I did talk briefly about this with Sam in [this conversation thread](https://github.com/apache/iceberg/pull/4706#discussion_r867216868).

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,222 @@
...
+    @abstractmethod
+    def list_namespaces(self) -> List[str]:
+        """List namespaces from the given namespace. If not given, list top-level namespaces from the catalog.
+
+        Returns:
+            List[str]: a List of namespace string
+        """
+
+    @abstractmethod
+    def get_namespace_metadata(self, namespace: str) -> Dict[str, str]:
```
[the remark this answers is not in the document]
dhruv-pratap: Ack. Will change in the next commit.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,222 @@
...
+
+        Returns:
+            Table: the created table instance
+
+        Raises:
+            AlreadyExistsError: If a table with the name already exists
+        """
+
+    @abstractmethod
+    def table(self, namespace: str, name: str) -> Table:
```
danielcweeks, line 59: `namespace` here and throughout the other signatures should probably be a tuple (`Tuple[str, ...]`) as opposed to a string.  After talking with @fokko, I think that's probably a more concise representation of the namespace without having to worry about parsing levels.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,222 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def properties(self) -> Dict[str, str]:
+        return self._properties
+
+    @abstractmethod
+    def create_table(self, table_spec: TableSpec) -> Table:
```
danielcweeks, line 45: I feel like there's an open question as to whether we want to use a wrapper like `TableSpec`, but even if we do, we should separate namespace from the table spec and make that a tuple parameter like the other signatures to be consistent.

python/src/iceberg/catalog/in_memory.py
danielcweeks: The in memory implementation should only exist in the test srcs so that it does not get confused / used for a real implementation.  Also echo Sam's comment that splitting the PR will make it easier to get the catalog in.  We can always follow up with changes if necessary.

python/setup.cfg
danielcweeks: While the dataclasses support may lag, we've had a lot of pushback around library creep.  I don't feel like the added value warrants bringing in a library in this case.  This may feel overly conservative, but with Iceberg being as low-level as it will be in many frameworks moving forward, we really want to keep the dependencies as slim as possible.

python/src/iceberg/catalog/base.py
dhruv-pratap: As touched in the other thread, I've removed `TableSpec`, and now take in exploded keyword arguments.

python/src/iceberg/catalog/base.py
dhruv-pratap: I've removed the `TableSpec` wrapper class in a recent commit and reverted back to taking in exploded keyword arguments. I've made the `partition_spec` optional in the signature now.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,251 @@
...
+        Args:
+            namespace: A Tuple of namespace levels. Ex: ('com','org','dept')
+            metadata: a dictionary of properties for the given namespace
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the name does not exist in the namespace
+        """
+
+
+class TableNotFoundError(Exception):
```
samredai, line 226: This is the first time we'll be adding custom exceptions. Since these will probably be used in other parts of the library, I think we should create a file `src/iceberg/exceptions.py` at the root and define these there.

Also, (mostly opinion here), I'd rather not override `__init__` just to provide a default message. I've actually found that default messages make debugging harder and I'd rather we simply define the exception types and raise more explicit messages based on the context where they're raise.

_exceptions.py_
```py
class TableNotFoundError(Exception):
    """Raised when a referenced table cannot be found"""
    pass
```
Then we can require providing a more detailed message whenever it's raised:

_catalog/base.py_
```py
raise TableNotFoundError(f"Cannot load table {table}, not found in catalog {catalog}")
```
```py
raise TableNotFoundError(f"Cannot drop table {table}, not found in catalog {catalog}")
```

dhruv-pratap: Done. Removed the custom messaging for errors and moved them over to `src/iceberg/exceptions.py` as suggested

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def set_namespace_metadata(self, namespace: Tuple[str, ...], metadata: Dict[str, str]) -> None:
+        if namespace in self.__namespaces:
+            self.__namespaces[namespace] = metadata
+        else:
+            raise NamespaceNotFoundError(f"Namespace {namespace} not found in the catalog")
+
+
+@pytest.fixture
+def catalog() -> InMemoryCatalog:
```
samredai, line 198: We may want to use this fixture in other test files so I would suggest moving it to `conftest.py`.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,444 @@
...
+    return InMemoryCatalog("test.in.memory.catalog", {"test.key": "test.value"})
+
+
+@pytest.fixture
+def table_spec() -> InMemoryTable:
```
samredai, line 203: Since we're exploding the arguments to the catalog methods now, do we need this fixture or the `InMemoryTable` class above? At first glance it may imply that catalog functionality requires you provide `Table` instances. Maybe we can just use a file-level constant that's just a dictionary.
```py
EXAMPLE_TABLE1 = {
  "namespace": ("com", "organization", "department"),
  "name": "my_table",
  "schema": Schema(schema_id=1),
  "location": "protocol://some/location",
  "partition_spec": PartitionSpec(),
  "properties": {"key1": "value1", "key2": "value2"}
}
```

dhruv-pratap: As suggested above, I've removed the table_spec fixture and just using constants for these test values in my test now.

I've removed the `InMemoryTable` implementation of `Table(ABC)` and trimmed down the InMemoryCatalog implementation to not use it and just use a bland `Table` instance. The only reason `InMemoryTable` existed was because I didn't want to add any table attributes to `Table(ABC)` in this PR to keep it as slim as possible. I had to sacrifice some of the `InMemoryCatalog` functionality around `rename_table()` and `replace_table()` but it still does a decent job of testing the base catalog interface and we can always add that back once we have a base `Table` implementation with all the attributes.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+    @property
+    def properties(self) -> Dict[str, str]:
+        return self._properties
+
+    @abstractmethod
+    def create_table(
+        self,
+        *,
+        namespace: Tuple[str, ...],
+        name: str,
```
rdblue, line 50: Why separate namespace and name? Can this be a single `Tuple[str]`?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Returns:
+            Table: the created table instance
+
+        Raises:
+            AlreadyExistsError: If a table with the name already exists
+        """
+
+    @abstractmethod
+    def table(self, namespace: Tuple[str, ...], name: str) -> Table:
```
rdblue, line 74: This should be called `load_table`.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+        """Loads the table's metadata and returns the table instance.
+
+        You can also use this method to check for table existence using 'try catalog.table() except TableNotFoundError'
+        Note: This method does not load table's data in any form.
```
rdblue, line 78: I think this is confusing. I believe you're trying to say that this doesn't scan data stored in the table, which would be more clear.

python/src/iceberg/catalog/base.py
rdblue: Also, what about allowing people to pass `str` here, which would be parsed by splitting on `.`? That makes simple cases easier. And if someone needs more complicated parsing, they can use the `Tuple[str]` form.

python/src/iceberg/table/base.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Table(ABC):
+    """Placeholder for Table managed by the Catalog that points to the current Table Metadata.
+
+    To be implemented by https://github.com/apache/iceberg/issues/3227
+    """
+
+
+class PartitionSpec:
```
rdblue, line 30: Should this create a `SortOrder` placeholder as well?

python/src/iceberg/exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+
+class TableNotFoundError(Exception):
```
rdblue, line 19: SQL Alchemy and Iceberg Java use `NoSuchTable` rather than `TableNotFound`. I think it would make sense to use the existing naming convention.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Returns:
+            Table: the table instance with its metadata
+
+        Raises:
+            TableNotFoundError: If a table with the name does not exist
+        """
+
+    @abstractmethod
+    def drop_table(self, namespace: Tuple[str, ...], name: str, purge: bool = True) -> None:
```
rdblue, line 92: I don't think it's a good idea to default `purge` to `True` because that will remove the ability to un-delete tables in a lot of cases where the caller doesn't actually need to purge immediately. We had to do it in Java because it wasn't originally an option, but here it can be required. Alternatively, we could split this into `purge_table` and `drop_table` if you think it is cleaner to have multiple methods.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+        Args:
+            namespace: The namespace to be dropped. Tuple of namespace levels. Ex: ('com','org','dept')
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the name does not exist in the namespace
+            NamespaceNotEmptyError: If the namespace is not empty
+        """
+
+    @abstractmethod
+    def list_tables(self, namespace: Optional[Tuple[str, ...]] = None) -> List[Tuple[Tuple[str, ...], str]]:
```
rdblue, line 174: Is this return type correct?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @abstractmethod
+    def drop_namespace(self, namespace: Tuple[str, ...]) -> None:
+        """Drop a namespace.
+
+        Args:
+            namespace: The namespace to be dropped. Tuple of namespace levels. Ex: ('com','org','dept')
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the name does not exist in the namespace
```
rdblue, line 169: I don't think you need "in the namespace" at the end. What namespace is that referring to?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+    @abstractmethod
+    def list_namespaces(self) -> List[Tuple[str, ...]]:
+        """List namespaces from the given namespace. If not given, list top-level namespaces from the catalog.
+
+        Returns:
+            List[Tuple[str, ...]]: a List of namespace, where each element is a Tuple of namespace levels. Ex: ('com','org','dept')
+        """
+
+    @abstractmethod
+    def load_namespace_metadata(self, namespace: Tuple[str, ...]) -> Dict[str, str]:
```
rdblue, line 198: How about renaming this to `load_namespace` instead? Do we need the `_metadata` part?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Returns:
+            Dict[str, str]: a dictionary of properties for the given namespace
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the name does not exist in the namespace
+        """
+
+    @abstractmethod
+    def set_namespace_metadata(self, namespace: Tuple[str, ...], metadata: Dict[str, str]) -> None:
```
rdblue, line 212: This needs to be very clear about behavior. The verb `set` can be interpreted 2 ways: set the whole metadata, or set the given properties. In most parts of the API, we prefer to be very clear and use separate args or methods for setting properties or removing properties.

I think this should do the same. How about changing this to `update_namespace` or `update_namespace_metadata` and accepting both `updates: Dict[str, str]` to set and `removals: Set[str]` to remove?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+        self,
+        *,
+        namespace: Tuple[str, ...],
+        name: str,
+        schema: Schema,
+        location: Optional[str] = None,
+        partition_spec: Optional[PartitionSpec] = None,
+        properties: Optional[Dict[str, str]] = None,
+    ) -> Table:
+        """Starts a transaction and replaces the table with the provided spec.
```
rdblue, line 132: It doesn't look like this starts a transaction because the return type is a `Table`. For now, how about removing this method until we want to add transactions? We can get a lot done with just `create_table` and `drop_table`. Transactions aren't really ready to be added yet anyway.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,223 @@
...
+        return self._name
+
+    @property
+    def properties(self) -> Dict[str, str]:
+        return self._properties
+
+    @abstractmethod
+    def create_table(
+        self,
+        *,
```
rdblue, line 48: What is your rationale for this? I'd probably just allow whatever the caller prefers.

samredai: +1, I can't think of a reason why we would want to stop someone from doing `create_table(("foodb"), "footable", schema)` and force them to do `create_table(namespace=("foodb"), name="footable", schema=schema)`. Both seem valid to me.

python/src/iceberg/catalog/base.py
samredai: +1, I would make this type `Union[Tuple[str], str]`

dhruv-pratap: _Why separate namespace and name?_ 
I feel like since they are two different concepts we should keep them explicitly separate, rather that it being implicit.

_Re: namespace type_
This was recommendation from @danielcweeks and @Fokko from [an earlier thread](https://github.com/apache/iceberg/pull/4706#discussion_r869626281). We have gone from `str` to `Tuple[str, ...] ` to now `Union[Tuple[str, ...], str]`. I think there is conversation to be had :)

python/src/iceberg/catalog/base.py
dhruv-pratap: Ack, will make the change. But the thinking here was the user can use method chaining here and the name `load_table` didn't quite fit in there.

`
catalog
 .table('my_table')
 .rename(...)
`

python/src/iceberg/catalog/base.py
dhruv-pratap: Ack. Will change

python/src/iceberg/table/base.py
dhruv-pratap: Intentionally left out to make this PR as slim as possible. SortOrder is an indirect dependency of the catalog interface.

python/src/iceberg/exceptions.py
dhruv-pratap: Ack. Will change.

python/src/iceberg/catalog/base.py
dhruv-pratap: I like the idea to split it to make it explicit, will make the change.

python/src/iceberg/catalog/base.py
dhruv-pratap: Yes :) It essentially returns a list of tuple of table-namespace and table-name.

python/src/iceberg/catalog/base.py
dhruv-pratap: Ack. Copy-pasta error. Will amend.

python/src/iceberg/catalog/base.py
dhruv-pratap: I like `update_namespace` accepting both `updates` and `removals`, will make the arguments optional as well.

python/src/iceberg/catalog/base.py
dhruv-pratap: Ack. Will remove this API for now.

python/src/iceberg/catalog/base.py
dhruv-pratap: IMO the model should reflect that "A catalog has namespaces that has tables". So `Catalog` --> `Namespace` --> `Table`

I thought about this several times when modeling it but didn't have the context on why it wasn't done this way in the Java API. The fact that we don't allow namespaces to be dropped that are not empty is a code smell that it should be modeled that way. What do you think?

python/src/iceberg/catalog/base.py
rdblue: Let's start by considering whether to support loading the table by `str`.

I think that's a good idea because that's going to cover most use cases. Do you agree, @dhruv-pratap?

python/src/iceberg/catalog/base.py
rdblue: Understood

python/src/iceberg/catalog/base.py
rdblue: Yeah, I think we should allow callers to choose. You make a good point about internal APIs, where this is probably more relevant. But forcing style on library callers seems unnecessary to me.

python/src/iceberg/catalog/base.py
dhruv-pratap: I personally prefer to have my data types reflect what the end user see so I like 'str' and that's why I started with it in the first place. 

But just so that I understand it right, when you say load table using 'str' do you mean using just a single fully classified table name 'str'? Or,  a 'str' namespace and 'str' name?

python/src/iceberg/catalog/base.py
dhruv-pratap: Makes sense. Will remove the enforcement.

python/src/iceberg/catalog/base.py
rdblue: @dhruv-pratap, I agree that this should support strings for table names, like `catalog.load_table("examples.nyc_taxi_yellow")`.

> do you mean using just a single fully classified table name 'str'? Or, a 'str' namespace and 'str' name?

I don't think it makes sense to force the caller to parse if we are going to split on `.`. If Iceberg can handle splitting the namespace then it can also split the table name out. That also aligns with how people will naturally use this, like Spark's `writeTo("db.table")` or `spark.read("examples.nyc_taxi_yellow")`. That means we support a string table identifier.

We also know that we need to support `Tuple[str]`, for cases like SQL engines where `.` can be included in an identifier by escaping and quoting.

Supporting both `str` and `Tuple[str]` means that we should not have a separate `namespace` argument. Otherwise, there are ambiguous (or not obvious) cases, like `catalog.load_table(namespace=("a", "b"), name="c.d")`.

Bringing all that together, the catalog methods should accept `identifier: str | Tuple[str]`. If the identifier is a string, it is split into a tuple on `.`. If it is a tuple, it is used as-is. That's simple and understandable, while supporting all of the cases that we need.

python/src/iceberg/catalog/base.py
rdblue: Changing methods to identify tables by `identifier: str | Tuple[str]` instead of by `namespace` and `name` simplifies this. It can just be `List[Tuple[Str]]` rather than having another tuple of namespace and name.

You may also consider making this a bit more clear by defining type aliases, like @Fokko did in #4685 for `Position`. For example:

```python
Identifier = Tuple[str]

def load_table(self, ident: str | Identifier, ...) -> Table:
    ...

def list_tables(self, namespace: Optional[Identifier] = None) -> List[Identifier]:
    ...
```

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/exceptions.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Fixed in a recent commit. Marking it resolved.

python/src/iceberg/table/base.py
dhruv-pratap: Marking it resolved as this will be implemented by [#3227](https://github.com/apache/iceberg/issues/3227)

python/src/iceberg/catalog/base.py
dhruv-pratap: API is now renamed to `load_namespace` in a recent commit.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Raises:
+            AlreadyExistsError: If a table with the name already exists
+        """
+
+    @abstractmethod
+    def load_table(self, identifier: Union[str, Identifier]) -> Table:
+        """Loads the table's metadata and returns the table instance.
+
+        You can also use this method to check for table existence using 'try catalog.table() except TableNotFoundError'
```
Fokko, line 77: We could also add a convenience method for this:

```python
def table_exists(self, identifier: Union[str, Identifier]) -> bool:
    try:
        self.load_table(identifier)
        return True
    except:
        return False
```

We could implement this in the

Fokko (approved):
Looking good 👍🏻  The only thing I'm not sure about is if we should both accept strings _and_ tuples for identifying tables/namespaces etc or just only tuples of strings.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+        identifier: Union[str, Identifier],
+        schema: Schema,
+        location: Optional[str] = None,
+        partition_spec: Optional[PartitionSpec] = None,
+        properties: Optional[Metadata] = None,
+    ) -> Table:
+        """Create a table
+
+        Args:
+            identifier: Fully classified table name with its namespace. If the identifier is a string, it is split into a tuple on '.'. If it is a tuple, it is used as-is.
```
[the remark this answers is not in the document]
dhruv-pratap: We did discuss how we would be handling it for such cases over [here](https://github.com/apache/iceberg/pull/4706#discussion_r873059425). CC @rdblue 

The idea is for such special use cases we expect the user to pass us a tuple so that it can be used as-is.

python/src/iceberg/catalog/base.py
dhruv-pratap: We could, but most of the usages I found for this API in the internal Java implementation uses it as a check to throw the exception AlreadyExistsError.  Maybe we can add it later if we really feel we need it. What do you think?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+#  under the License.
+
+from abc import ABC, abstractmethod
+from typing import Dict, List, Optional, Set, Tuple, Union
+
+from iceberg.schema import Schema
+from iceberg.table.base import PartitionSpec, Table
+
+Identifier = Tuple[str, ...]
+Metadata = Dict[str, str]
```
rdblue, line 25: `Metadata` is definitely an overloaded word. In most places, we prefer "properties" instead. How about `Properties`?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @abstractmethod
+    def load_table(self, identifier: Union[str, Identifier]) -> Table:
+        """Loads the table's metadata and returns the table instance.
+
+        You can also use this method to check for table existence using 'try catalog.table() except TableNotFoundError'
+        Note: This method doesn't scan data stored in the table.
+
+        Args:
+            identifier: Fully classified table name with its namespace. If the identifier is a string, it is split into a tuple on '.'. If it is a tuple, it is used as-is.
```
rdblue, line 81: Rather than copying this everywhere, how about using a simple description and clarifying how it is handled in the class documentation?

python/src/iceberg/catalog/base.py
rdblue, line 60: Normally, I think the term is "fully qualified". If you use "fully qualified" then there is no need to say "with its namespace" because that's redundant.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+    @abstractmethod
+    def list_namespaces(self) -> List[Identifier]:
+        """List namespaces from the given namespace. If not given, list top-level namespaces from the catalog.
+
+        Returns:
+            List[Identifier]: a List of namespace, where each element is a Tuple of namespace levels. Ex: ('com','org','dept')
+        """
+
+    @abstractmethod
+    def load_namespace(self, namespace: Union[str, Identifier]) -> Metadata:
```
rdblue, line 176: If this isn't returning a `Namespace` object, then it should still be named `load_namespace_metadata`. The option to remove the `_metadata` assumed that it would return a `Namespace` object rather than a properties dict.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the given name does not exist
+        """
+
+    @abstractmethod
+    def update_namespace_metadata(
+        self, namespace: Union[str, Identifier], removals: Optional[Set[str]] = None, updates: Optional[Metadata] = None
+    ) -> None:
+        """Removes provided metadata keys and then updates metadata for a namespace.
```
rdblue, line 193: This should not specify an order of operations. It is an error if `removals` and `updates` have overlapping keys. There is no required ordering behavior.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Note: No errors are raised if a provided removal key is not found.
+
+        Args:
+            namespace: If the namespace is a string, it is split into a tuple on '.'. If it is a tuple, it is used as-is.
+            removals: a set of metadata keys that need to be removed. Optional Argument.
+            updates: a dictionary of properties to be updated for the given namespace. Optional Argument.
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the given name does not exist
```
rdblue, line 203: Can you add `ValueError` for when `removals` and `updates` have an overlapping property name?

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,418 @@
...
+
+def test_create_table(catalog: InMemoryCatalog):
+    table = catalog.create_table(
+        identifier=TEST_TABLE_IDENTIFIER,
+        schema=TEST_TABLE_SCHEMA,
+        location=TEST_TABLE_LOCATION,
+        partition_spec=TEST_TABLE_PARTITION_SPEC,
+        properties=TEST_TABLE_PROPERTIES,
+    )
+    assert catalog.load_table(TEST_TABLE_IDENTIFIER) == table
```
rdblue, line 230: When creating a table, I think this should validate that the table looks as expected by validating the schema, partition spec, location, and properties.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,418 @@
...
+    # Given
+    given_catalog_has_a_table(catalog)
+    # When
+    with pytest.raises(AlreadyExistsError):
+        catalog.create_table(
+            identifier=TEST_TABLE_IDENTIFIER,
+            schema=TEST_TABLE_SCHEMA,
+            location=TEST_TABLE_LOCATION,
+            partition_spec=TEST_TABLE_PARTITION_SPEC,
+            properties=TEST_TABLE_PROPERTIES,
```
rdblue, line 243: The extra arguments aren't needed.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,418 @@
...
+    # Given
+    given_table = given_catalog_has_a_table(catalog)
+    # When
+    table = catalog.load_table(TEST_TABLE_IDENTIFIER)
+    # Then
+    assert table == given_table
+
+
+def test_table_raises_error_on_table_not_found(catalog: InMemoryCatalog):
+    with pytest.raises(NoSuchTableError):
```
rdblue, line 257: Can you also validate the error message? It should include the table identifier ("Table does not exist: db.table") so that there is context.

python/src/iceberg/catalog/base.py
kbendick: +1. Properties is also the name of the field and is codified in the REST OpenAPI spec as well.

dhruv-pratap: @rdblue Given this, should I rename the namespace methods to `load_namespace_properties()` `update_namespace_properties()` to have the same verbiage?

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+class Catalog(ABC):
+    """Base Catalog for table operations like - create, drop, load, list and others.
+
+    Attributes:
+        name(str): Name of the catalog
+        properties(Dict[str, str]): Catalog properties
+    """
+
+    def __init__(self, name: str, properties: Metadata):
```
Fokko, line 36: We could also generate this `__init__` very easily using the `@dataclass` by adding the decorator and defining the properties in the class itself:
```
    name: str
    properties: Metadata
```

This will also autogenerate the getters below:
```
    @property
    def name(self) -> str:
        return self._name

    @property
    def properties(self) -> Metadata:
        return self._properties
```

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,418 @@
...
+from iceberg.table.base import PartitionSpec, Table
+
+
+class InMemoryCatalog(Catalog):
+    """An in-memory catalog implementation for testing purposes."""
+
+    __tables: Dict[Identifier, Table]
+    __namespaces: Dict[Identifier, Metadata]
+
+    def __init__(self, name: str, properties: Metadata):
```
Fokko, line 39: If we go with a `@dataclass`, you can easily remove this init, and replace the internal fields with:
```
    __tables: Dict[Identifier, Table]  = field(default_factory=dict)
    __namespaces: Dict[Identifier, Metadata] = field(default_factory=dict)

```

Fokko (commented):
The errors in the CI are on spelling. You need to add the missing words (such as mypy which don't exist in the default US dictionary) in `python/spellcheck-dictionary.txt`

python/src/iceberg/catalog/base.py
dhruv-pratap: @Fokko I am a little on the fence with having our interfaces (abstract classes) marked as `@dataclass`.  I'm not entirely sure if `ABC` and `@dataclass` play along nicely. Also, it forces all sub-classes to be annotated with `@dataclass` as well.

python/src/iceberg/catalog/base.py
dhruv-pratap: Addressed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Addressed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Addressed in a recent commit. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Addressed in a recent commit. Marking it resolved.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,418 @@
...
+        partition_spec=TEST_TABLE_PARTITION_SPEC,
+        properties=TEST_TABLE_PROPERTIES,
+    )
+
+
+def test_namespace_from_tuple():
+    # Given
+    identifier = ("com", "organization", "department", "my_table")
+    # When
+    namespace_from = InMemoryCatalog.namespace_from(identifier)
```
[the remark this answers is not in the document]
dhruv-pratap: Pulled up. Marking it resolved.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,418 @@
...
+        catalog.create_table(
+            identifier=TEST_TABLE_IDENTIFIER,
+            schema=TEST_TABLE_SCHEMA,
+            location=TEST_TABLE_LOCATION,
+            partition_spec=TEST_TABLE_PARTITION_SPEC,
+            properties=TEST_TABLE_PROPERTIES,
+        )
+
+
+def test_table(catalog: InMemoryCatalog):
```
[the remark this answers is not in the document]
dhruv-pratap: Addressed. Marking it resolved.

python/tests/catalog/test_base.py
dhruv-pratap: Addressed. Marking it resolved.

python/src/iceberg/catalog/base.py
dhruv-pratap: Addressed. Marking it resolved.

dhruv-pratap:
Sorry for the mess above ^. Had to do a rebase from master to pull in recent PRs to add `namespace` to `spellcheck-dictionary.txt` for the CI checks to pass.

python/tests/catalog/test_base.py
rdblue: Adding them in #3227 is okay. FYI, you may want to look at the tests we use to [validate catalogs](https://github.com/apache/iceberg/blob/master/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java) in the Java implementation. Many of the cases are the same to make sure you're getting all of the checks.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,410 @@
...
+    # Given
+    given_table = given_catalog_has_a_table(catalog)
+    # When
+    table = catalog.load_table(TEST_TABLE_IDENTIFIER)
+    # Then
+    assert table == given_table
+
+
+def test_table_raises_error_on_table_not_found(catalog: InMemoryCatalog):
+    with pytest.raises(NoSuchTableError, match="Table my_table not found"):
```
rdblue, line 243: Error messages like this can be confusing if the name is invalid: `Table  not found`. Instead, we prefer to make it clear that the name was included by using a message like `Table does not exist: my_table`. Can you update the error message?

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,410 @@
...
+    # Given
+    given_table = given_catalog_has_a_table(catalog)
+
+    # When
+    new_table = "new.namespace.new_table"
+    table = catalog.rename_table(TEST_TABLE_IDENTIFIER, new_table)
+
+    # Then
+    assert table
+    assert table is not given_table
```
rdblue, line 282: Isn't this just asserting that the table isn't the same instance? Is that useful?

rdblue:
I'm ready to commit this once the error message format is fixed.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,410 @@
...
+        partition_spec: Optional[PartitionSpec] = None,
+        properties: Optional[Properties] = None,
+    ) -> Table:
+
+        identifier = Catalog.identifier_to_tuple(identifier)
+        namespace = Catalog.namespace_from(identifier)
+        name = Catalog.table_name_from(identifier)
+
+        if identifier in self.__tables:
+            raise AlreadyExistsError(f"Table {name} already exists in namespace {namespace}")
```
rdblue, line 58: This error message should also use the same format that I noted a minute ago: `Table already exists: {identifier}`

python/tests/catalog/test_base.py
dhruv-pratap: @rdblue In the earlier commits I was asserting if the table name has changed, but since I had rollback any changes to the `Table` interface in this PR I resorted to checking just that the table is changing in some form post renaming. I will eventually be changing this to assert table name change once we add attributes back to `Table` in #3227

dhruv-pratap: Nevermind. I worked around it for the time being. Since the `Table` class isn't frozen yet I added an attribute `identifier` post creation in the `InMemoryCatalog` and asserting on `identifier` renaming for now. This obviously would fail, and rightly so, once we make `Table` immutable/frozen.

python/tests/catalog/test_base.py
dhruv-pratap: Addressed in a recent commit. Marking it resolved.

rdblue (approved):

dhruv-pratap:
@rdblue Are we good to merge?

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,401 @@
...
+    NoSuchTableError,
+)
+from iceberg.schema import Schema
+from iceberg.table.base import PartitionSpec, Table
+
+
+class InMemoryCatalog(Catalog):
+    """An in-memory catalog implementation for testing purposes."""
+
+    __tables: Dict[Identifier, Table]
```
kbendick, line 36: Nit / question: I know that the usage of double underscore somewhat mangles function names to make them "private". Does that apply to class fields as well?

I've always preferred single underscore for this reason, but I'm not a huge pythonista and this is a test class. So more just for my own knowledge as well as to better understand the current "best practices".

samredai: I noticed this too. In some ways I think the double underscore here signals "These properties don't normally exist in a non-testing Catalog implementation" but I think it'd be clearer to just directly describe that in the docstring and use property names without any underscores. You're right in that the double underscore would mangle the names, but only for a child class, something like `Foo(InMemoryCatalog)`, which isn't done anywhere here. Python would then make the property name `_InMemoryCatalog__tables` for instances of `Foo`.

I came to the same conclusion as you, this is a test class so it doesn't matter as much.

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,247 @@
...
+
+    The catalog namespace APIs follow a similar convention wherein they also accept a namespace identifier that can be a string
+    or tuple of strings.
+
+    Attributes:
+        name(str): Name of the catalog
+        properties(Properties): Catalog properties
+    """
+
+    def __init__(self, name: str, properties: Properties):
```
Fokko, line 42: I'll continue my quest for data classes here as well. Instead of creating the getters/setters, we could also let the dataclasses do this:
```python
@dataclass
class Catalog(ABC):

    name: str =  field()
    name: Properties =  field(default_factory=dict)
```
This will automatically generate the getters and setters.

python/tests/catalog/test_base.py
Fokko: I think a single underscore is more appropriate here. It is an internal variable of the `InMemoryCatalog`.

If we want to set this as a ClassVar, then we also have to annotation for it:
```suggestion
    _tables: ClassVar[Dict[Identifier, Table]]
```
But I'm not sure if we want to do that since we would like to instantiate a fresh InMemoryCatalog for the different tests.

python/src/iceberg/catalog/base.py
Fokko: If we want to go full blown pythonic in the future, we could also validate these using Pydantic validators: https://pydantic-docs.helpmanual.io/usage/validators/

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,401 @@
...
+        table.identifier = to_identifier
+        self.__tables[to_identifier] = table
+        return table
+
+    def create_namespace(self, namespace: Union[str, Identifier], properties: Optional[Properties] = None) -> None:
+        namespace = Catalog.identifier_to_tuple(namespace)
+        if namespace in self.__namespaces:
+            raise AlreadyExistsError(f"Namespace already exists: {namespace}")
+        else:
+            self.__namespaces[namespace] = properties if properties else {}
```
Fokko, line 105: Less is more:
```suggestion
            self.__namespaces[namespace] = properties or {}
```

python/src/iceberg/catalog/base.py
```diff
@@ -0,0 +1,247 @@
...
+            removals: Set of property keys that need to be removed. Optional Argument.
+            updates: Properties to be updated for the given namespace. Optional Argument.
+
+        Raises:
+            NamespaceNotFoundError: If a namespace with the given name does not exist
+            ValueError: If removals and updates have overlapping keys.
+        """
+
+    @staticmethod
+    def identifier_to_tuple(identifier: Union[str, Identifier]) -> Identifier:
```
Fokko, line 212: Naming things is one of the hardest things in Computer Science, but I find this one a bit awkward since:
```python
Identifier = Tuple[str, ...]
```
So it is actually already a tuple. How about:
```suggestion
    def to_identifier(identifier: Union[str, Identifier]) -> Identifier:
```

Fokko (commented):
Few last remarks, looks great in general!

python/src/iceberg/catalog/base.py
rdblue: I think using `@dataclass` is a good idea, but we can move on here and refactor it later.

python/tests/catalog/test_base.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+TEST_TABLE_IDENTIFIER = ("com", "organization", "department", "my_table")
+TEST_TABLE_NAMESPACE = ("com", "organization", "department")
+TEST_TABLE_NAME = "my_table"
+TEST_TABLE_SCHEMA = Schema(schema_id=1)
+TEST_TABLE_LOCATION = "protocol://some/location"
+TEST_TABLE_PARTITION_SPEC = PartitionSpec()
+TEST_TABLE_PROPERTIES = {"key1": "value1", "key2": "value2"}
+NO_SUCH_TABLE_ERROR = "Table does not exist: \\('com', 'organization', 'department', 'my_table'\\)"
```
rdblue, line 157: These should use string names, not tuples.

python/tests/conftest.py
```diff
@@ -112,3 +113,8 @@ def table_schema_nested():
 @pytest.fixture(scope="session", autouse=True)
 def foo_struct():
     return FooStruct()
+
+
+@pytest.fixture
+def catalog() -> InMemoryCatalog:
+    return InMemoryCatalog("test.in.memory.catalog", {"test.key": "test.value"})
```
rdblue, line 120: `test.key` isn't used anywhere, nor is `test.in.memory.catalog`. Can you add tests for properties and name?

rdblue (approved):

rdblue:
There are still a few things to fix, but I think we can clean up things like moving to `@dataclass` separately to unblock other work.

PR #4707, opened 2022-05-05
Doc : Document time travel via SPARK SQL

singhpk234:
At present `AS OF` grammar is not supported in SPARK-SQL (was added in Spark 3.3 via [SPARK-37219](https://issues.apache.org/jira/browse/SPARK-37219)).

On digging more in the code found that there was a work-around adopted by the community until this grammar is supported in SPARK itself, by using `at_timestamp_`, `snapshot_id_` 
https://github.com/apache/iceberg/blob/bf582ebf68ea05be26f38d786584d474aebe048b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java#L545-L555

Presently the docs stated `Time travel is not yet supported by Spark's SQL syntax.` which can be worked around with the functionalities present. This PR attempts to document that.

--- 

cc @rdblue @jackye1995  @RussellSpitzer

### DIFF
--- a/docs/spark/spark-queries.md
+++ b/docs/spark/spark-queries.md
@@ -116,7 +116,19 @@ ignored. Do not use `table` when attempting to time-travel or use other options.
 in [Spark 3.1 - SPARK-32592](https://issues.apache.org/jira/browse/SPARK-32592).
 {{< /hint >}}
 
-Time travel is not yet supported by Spark's SQL syntax.
+Time travel is not yet fully supported by Spark's SQL syntax. (ref : [Spark 3.3 - SPARK-37219](https://issues.apache.org/jira/browse/SPARK-37219)
+for `AS OF` syntax support)
+
+As an alternative we can use `at_timestamp_`, `snapshot_id_` as prefix of timestamp, snapshot_id respectively along with
+the table name to time travel.
+
+```sql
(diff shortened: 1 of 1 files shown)


singhpk234:
Looks like this was intended for internal use and we don't want the users to actually use it. 

refering  : https://github.com/apache/iceberg/pull/3269#issuecomment-961520437

singhpk234:
@puchengy, I was referring to this subpart of https://github.com/apache/iceberg/pull/3269#issuecomment-961520437 where @rdblue suggested :
> We don't have to document that we support time travel through table identifiers so that we can remove it later and replace it with the AS OF syntax

Hence I closed this PR considering it not being documented is intentional. 

> Do you mean this workaround is not guaranteed to be supported and might be dropped in the future?

As per my understanding of the comment yes, hence not documenting this could avoid users to make dependency on it.

Now `AS OF` Syntax is present since [3.3](https://issues.apache.org/jira/browse/SPARK-37219), Though I am not sure of the plans of spark community to backport it back to lower spark versions like 3.0, 3.1, 3.2. Also not sure if it will be dropped from iceberg once they are supported in lower version or in 3.3. Maybe the  authors / reviewers of the pr's have deeper insights on this.

puchengy:
@singhpk234 got it, thanks for your reponse.

PR #4708, opened 2022-05-05
API: Add an action to generate table change set

flyrain:
Extracted the interface change from the draft PR #4539. According the MVP design and recent community discussion, we are going to create an action interface first for Change Data Capture(CDC).

cc @aokolnychyi @RussellSpitzer @szehon-ho @jackye1995 @kbendick @karuppayya @chenjunjiedada @stevenzwu @rdblue @Reo-LEI @hameizi @singhpk234 @rajarshisarkar

### DIFF
--- a/api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
+++ b/api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.actions;
+
+public interface GenerateChangeSet extends Action<GenerateChangeSet, GenerateChangeSet.Result> {
+  /**
+   * Emit changed data set by a snapshot id.
+   *
+   * @param snapshotId id of the snapshot to generate changed data
+   * @return this for method chaining
+   */
+  GenerateChangeSet forSnapshot(long snapshotId);
+
+  /**
+   * Emit changed data set for the current snapshot.
+   *
+   * @return this for method chaining
+   */
+  GenerateChangeSet forCurrentSnapshot();
+
+  /**
+   * Emit changed data from a particular snapshot(exclusive).
+   *
+   * @param fromSnapshotId id of the start snapshot
+   * @return this for method chaining
+   */
+  GenerateChangeSet afterSnapshot(long fromSnapshotId);
+
+  /**
+   * Emit change data set from the start snapshot (exclusive) to the end snapshot (inclusive).
+   *
+   * @param fromSnapshotId id of the start snapshot
+   * @param toSnapshotId   id of the end snapshot
+   * @return this for method chaining
+   */
+  GenerateChangeSet betweenSnapshots(long fromSnapshotId, long toSnapshotId);
+
+  /**
+   * The action result that contains a dataset of changed rows.
+   */
+  interface Result<T> {
+    /**
+     * Returns the change set.
+     */
+    T changeSet();
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
+++ b/api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
@@ -19,6 +19,9 @@
 
 package org.apache.iceberg.actions;
 
+/**
+ * An action that generates a change data set from snapshots.
+ */
 public interface GenerateChangeSet extends Action<GenerateChangeSet, GenerateChangeSet.Result> {
   /**
    * Emit changed data set by a snapshot id.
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+   * @return this for method chaining
+   */
+  GenerateChangeSet forSnapshot(long snapshotId);
+
+  /**
+   * Emit changed data set for the current snapshot.
+   *
+   * @return this for method chaining
+   */
+  GenerateChangeSet forCurrentSnapshot();
```
flyrain, line 36: The original discussion to remove this API is here, https://github.com/apache/iceberg/pull/4539#discussion_r855718913. I think it is still useful as a shortcut of `forSnapshot(table.currentSnapshot().snapshotId())`, like a syntactic sugar of a programming language. It is not worth if it causes any confusion, which I don't think it will. I'm open to any suggestion.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Emit changed data from a particular snapshot(exclusive).
+   *
+   * @param fromSnapshotId id of the start snapshot
+   * @return this for method chaining
+   */
+  GenerateChangeSet afterSnapshot(long fromSnapshotId);
+
+  /**
+   * Emit change data set from the start snapshot (exclusive) to the end snapshot (inclusive).
```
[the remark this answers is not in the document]
flyrain: Hi @singhpk234, I meant to use `change` as a noun here. So it is a table change, which is presented as a data set, instead of a data set of the table which has been changed.

rdblue:
@flyrain, is this really an "Action"? What if we had a separate API for helpers like this that wasn't called "actions"? Then we wouldn't need to add and remove things from actions. I'd prefer to keep it separate, though I'm open to hearing the arguments for mixing this in.

flyrain:
@rdblue, @aokolnychyi and me have planned a separated PR for the CDC scan API, which will be more general, can be used across engines.

> What if we had a separate API for helpers like this that wasn't called "actions"?

Is the helper the CDC scan API, or an abstract(interface) of the CDC action and the CDC scan API? I think the later makes more sense, which I image is a pretty thin layer. Is it necessary?

aokolnychyi:
Let me take a look. Sorry for the delay on my side.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.actions;
+
+public interface GenerateChangeSet extends Action<GenerateChangeSet, GenerateChangeSet.Result> {
```
aokolnychyi, line 22: Shall we add a few sentences about the purpose of this action similar to other actions?

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
aokolnychyi: I am not sure how frequently folks will be calling this. Is there a particular use case? I thought folks would usually have a starting point and would be interested to see changes since that.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+   */
+  GenerateChangeSet forCurrentSnapshot();
+
+  /**
+   * Emit changed data from a particular snapshot(exclusive).
+   *
+   * @param fromSnapshotId id of the start snapshot
+   * @return this for method chaining
+   */
+  GenerateChangeSet afterSnapshot(long fromSnapshotId);
```
aokolnychyi, line 44: This naming seems good to me but I wonder whether we should follow whatever was recently added in #4580. One thing I like about that API is using explicit inclusive/exclusive words whenever configuring the start snapshot.

```
GenerateChangeSet fromSnapshotInclusive(long fromSnapshotId);
GenerateChangeSet fromSnapshotExclusive(long fromSnapshotId);
GenerateChangeSet toSnapshot(long toSnapshotId);
```

Thoughts, @flyrain @stevenzwu @szehon-ho @rdblue @RussellSpitzer?

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
```diff
@@ -0,0 +1,64 @@
...
+   * @param fromSnapshotId id of the start snapshot
+   * @param toSnapshotId   id of the end snapshot
+   * @return this for method chaining
+   */
+  GenerateChangeSet betweenSnapshots(long fromSnapshotId, long toSnapshotId);
+
+  /**
+   * The action result that contains a dataset of changed rows.
+   */
+  interface Result<T> {
```
aokolnychyi, line 58: I personally don't think the action API should be limited to table maintenance. When we added it, the idea was to provide solutions for common problems that require a query engine. Initially, those happened to be mostly related to table maintenance. However, I wouldn't mind actions for other purposes.

We all agree to share the planning part across engines via a dedicated scan API. However, we still need to build an engine-specific representation somewhere. Of course, we could have a utility but I am not a big fan of exposing utilities to users. The action API is much more user-facing and requires us to think about the proper interface and keep the compatibility. We had many issues with Spark utils in the past.

One thing about using the action API for CDC that concerns me is the need to parameterize the result. Our action API was engine agnostic so far but we need to return Spark `Dataset` here. I am not sure how big of a deal it is.

Thoughts, @flyrain @rdblue @RussellSpitzer @szehon-ho @stevenzwu?

@rdblue @stevenzwu, I think both of you had concerns about using an action. Could you elaborate a little bit? Given that we will have a common Scan API to share the planning logic.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
flyrain: I like the idea to use the same name as https://github.com/apache/iceberg/pull/4580, will make the change if there is no objection.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
aokolnychyi: Another alternative we discussed is using a new metadata table so that users can point to it to load changes. What is our long-term plan?

stevenzwu: @aokolnychyi My question would be why not exposing the CDC read (change set) via a source (in Spark or other engines). I remember @rdblue mentioned some problem with Spark source for that purpose.

aokolnychyi: @stevenzwu, do you mean a custom source for CDC records or just extending the main data source integration? Spark, unlike Flink, does not have an API for CDC which can be used by data sources. We would want to add that to Spark eventually but it will be a major effort and will take tons of time.

stevenzwu: thx for explaining. I guess that is the problem Ryan mentioned. Spark source API doesn't support CDC.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
flyrain: Probably not so frequent. I will remove this. We can always add it back if this is useful.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
flyrain: Made the change in the new commit.

api/src/main/java/org/apache/iceberg/actions/GenerateChangeSet.java
rdblue: > I think both of you had concerns about using an action. Could you elaborate a little bit?

I don't have a problem with using Actions for other tasks, but I think those other tasks should be actions taken on the table. It seems odd to me to have an "action" be something that essentially builds a view of a table that is evaluated lazily. In this case, the action doesn't really _do_ anything. It is a convenience.

That's why the return value doesn't really fit. Normally, you get back a summary of what was done, but in this case nothing was done and you get back the resulting dataframe.

The other good point is what Steven raised. I would like to make this a source so that you can use it as natively in Spark as possible. We can use the same approach as time travel and metadata tables. If you load `db.table.change_log`, then you get the CDC view of the table. And you can use time travel selectors (from-snapshot-id, to-snapshot-id, etc) to select the range of time. I'd definitely prefer this more native approach.

szehon-ho: Sorry to chime in late, but I like metadata table way as well, as was discussing with @flyrain, though I might be missing some context.

To me it's a better UX (user can use SQL to query it).  And code wise it may make the integration with other engine like Trino easier as there's some common interfaces exposed across metadata tables they may take advantage of, though I admit that's a bit theoretical at this point.

aokolnychyi:
PR #4870 is one way to design the CDC scan. Let me know what everybody thinks.

flyrain:
Close this. PR #4870 is the way to go.

PR #4709, opened 2022-05-05
Spark: Document that metadata tables support time travel

liuml07:
I think the current spark-queries doc is unclear about metadata tables can be inspected with time travel feature. This PR documents that with a sample query.


format/spec.md
```diff
@@ -456,9 +456,9 @@ The column metrics maps are used when filtering to select both data and delete f
 
 The manifest entry fields are used to keep track of the snapshot in which files were added or logically deleted. The `data_file` struct is nested inside of the manifest entry so that it can be easily passed to job planning without the manifest entry fields.
 
-When a file is added to the dataset, it’s manifest entry should store the snapshot ID in which the file was added and set status to 1 (added).
```
liuml07, line 459: The character `’` seems non-English character. It should be `its` instead of `it's` anyway so I fix this as a trivial change in this pull request.

docs/spark/spark-queries.md
```diff
@@ -332,4 +332,15 @@ Metadata tables can be loaded in Spark 2.4 or Spark 3 using the DataFrameReader
...
+You can also inspect Iceberg metadata tables with the time travel feature:
+
+```scala
+// get table's all data files and each data file's metadata at snapshot-id 7277403863961056344
+spark.read
+        .format("iceberg")
+        .option("snapshot-id", 7277403863961056344L)
+        .load("db.table.files")
+        .show()
+```
```
singhpk234, line 346: [question] There is a dedicated `TimeTravel` section above, should we move this there ? Your thoughts ?

liuml07: Yes adding to that section also works. But my concern was that, the `TimeTravel` section is before metadata tables section. By then, readers may not understand what the table name `db.table.files` is for.

hililiwei: +1 for show it here.

liuml07:
I have rebased this to avoid conflicts. Not sure if I can get a review? @kbendick

docs/spark-queries.md
```diff
@@ -394,3 +394,14 @@ spark.read.format("iceberg").load("db.table.files").show(truncate = false)
 // Hadoop path table
 spark.read.format("iceberg").load("hdfs://nn:8020/path/to/table#files").show(truncate = false)
 ```
+
+You can also inspect Iceberg metadata tables with the time travel feature:
```
szehon-ho, line 398: 1. I think we tend to avoid second person 'you' in the docs, and just state that something can be done.

2. Also what do you think about making a separate section (Time Travel with Metadata Tables) and showing both the DF and the SQL way ?  Otherwise people may assume its only possible via dataframe.

3.  Also this is unrelated but I realized that this section (Inspecting with DataFrames) has the wrong heading level, it should be under the level of (Inspecting Tables), as that level is about all metadata tables.  Can you help fix it as well?

docs/spark-queries.md
```diff
@@ -394,3 +394,22 @@ spark.read.format("iceberg").load("db.table.files").show(truncate = false)
...
 spark.read.format("iceberg").load("hdfs://nn:8020/path/to/table#files").show(truncate = false)
 ```
+
+### Time Travel with Metadata Tables
+
+To inspect a tables's metadata with the time travel feature:
+
+```sql
+-- get the table's file manifests at timestamp Sep 20, 2021 08:00:00
+SELECT * FROM prod.db.table.manifests TIMESTAMP AS OF '2021-09-20 08:00:00';
```
szehon-ho, line 404: Nit: can we remove the end semicolon for consistency?

docs/spark-queries.md
```diff
@@ -394,3 +394,22 @@ spark.read.format("iceberg").load("db.table.files").show(truncate = false)
...
+
+-- get the table's partitions with snapshot id 10963874102873L
+SELECT * FROM prod.db.table.partitions VERSION AS OF 10963874102873;
+```
+
+Metadata tables can also be inspected with time travel using the DataFrameReader API:
+
+```scala
+// get table's data files and each data file's metadata at snapshot-id 10963874102873
+spark.read.format("iceberg").option("snapshot-id", 10963874102873L).load("db.table.files").show()
```
szehon-ho, line 414: Can we remove show()?  I know the other one had it, but just realized its a bit unnecessary (this doc should just explain how to load the metadata table as dataframe, just like the previous section explain how to load the data table as dataframe, not necessarily how to show it)

docs/spark-queries.md
```diff
@@ -394,3 +394,22 @@ spark.read.format("iceberg").load("db.table.files").show(truncate = false)
...
+SELECT * FROM prod.db.table.manifests TIMESTAMP AS OF '2021-09-20 08:00:00';
+
+-- get the table's partitions with snapshot id 10963874102873L
+SELECT * FROM prod.db.table.partitions VERSION AS OF 10963874102873;
+```
+
+Metadata tables can also be inspected with time travel using the DataFrameReader API:
+
+```scala
+// get table's data files and each data file's metadata at snapshot-id 10963874102873
```
szehon-ho, line 413: What do you think simplifying it, // Load table's file metadata at snapshot-id... as dataframe

(Also files table may be delete files as well, hence removing 'data')

docs/spark-queries.md
```diff
@@ -394,3 +394,22 @@ spark.read.format("iceberg").load("db.table.files").show(truncate = false)
 // Hadoop path table
 spark.read.format("iceberg").load("hdfs://nn:8020/path/to/table#files").show(truncate = false)
```
szehon-ho, line 395: Sorry its unrelated, but could you also remove the show (truncate=true) here?  Should have added the comment in the previous pr.  This doc should just explain how to load the dataframe, not necessarily how to show it.

szehon-ho (commented):
Thanks for the update, looks much better.  Some nits and then I think it should be good to go

docs/spark-queries.md
liuml07: I see some places end with this `;` and some other places do not in this doc. I do not have specific preference for example code. But it seems a bit clearer for a code block with multiple SQL statement (like here). Do you think it's a good idea to make all places end with `;` instead?

docs/spark-queries.md
liuml07: Yeah this makes perfect sense - `show()` is totally irrelevant and also we do not attach result examples here.

docs/spark-queries.md
szehon-ho: You are right, I didn't notice it.

szehon-ho (approved):
Looks good to me, thanks for all the fixes, including unrelated ones.

docs/spark-queries.md
liuml07: I make all other SQL example end with `;` to make this style consistent and embarrassingly clear that the statement is complete.

szehon-ho (approved):

szehon-ho:
Thanks @liuml07 for the change and patience !

PR #4710, opened 2022-05-06
API: Add equals and hashCode to Schema

jzhuge:
Add `equals` method to Schema. It leverages the existing `sameSchema` method.

This will make it easy to compare 2 schemas.

Currently `Schema.toString()` is used to compare 2 schemas at a few places.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/Schema.java
+++ b/api/src/main/java/org/apache/iceberg/Schema.java
@@ -484,4 +484,21 @@ public class Schema implements Serializable {
             .map(this::identifierFieldToString)
             .collect(Collectors.toList())));
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    return sameSchema((Schema) o);
+  }
+
+  @Override
+  public int hashCode() {
+    return 31 *  struct.hashCode() + Arrays.hashCode(identifierFieldIds);
+  }
 }


kbendick (commented):
Thanks @jzhuge.

Can you point out the places where the `toString` method is used to compare?

`sameSchema` ignores the `schemaId` so using it for `equals` _might_ have unintended consequences down the road if we need to check if two schemas are exactly the same.

jzhuge:
https://github.com/apache/iceberg/blob/f68d8d426661efc0d7e5686fe833b573b74eadab/core/src/test/java/org/apache/iceberg/hadoop/TestHadoopCatalog.java#L69
https://github.com/apache/iceberg/blob/d2bb30f863d7276097c39754b9501ba27403b73c/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogTable.java#L87

jzhuge:
Yeah, we might need an exact match down the road.

rdblue:
@jzhuge, what does it mean for two schemas to be equal?

We didn't implement this originally because it isn't clear what the right answer is. If two schemas are structurally equal but have different identity columns, are they equal? Not implementing equals avoids the problem. Is there some reason why you need `equals` and `hashCode` that makes it a good time to solve this problem?

jzhuge:
Ah, didn't know this is deliberate. My only goal is to make test code cleaner, especially not to use `Schema.toString`.

I guest test code can use `sameSchema` or just compare `asStruct` outputs.

jzhuge:
Rejecting this PR in a few days if no further comment.

PR #4714, opened 2022-05-06
[AWS]Add AwsKmsClient for table encryption KMS client implementation

JiJiTang:
Aws KMS client implementation:

- Support data key generation using AWS KMS, configurable via property `ENABLE_KEY_GENERATION`

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsKmsClient.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsKmsClient.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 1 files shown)


ggershinsky:
cc @jackye1995 @RussellSpitzer @flyrain

aws/src/main/java/org/apache/iceberg/aws/AwsKmsClient.java
```diff
@@ -0,0 +1,101 @@
...
+import software.amazon.awssdk.services.kms.model.GenerateDataKeyRequest;
+import software.amazon.awssdk.services.kms.model.GenerateDataKeyResponse;
+
+public class AwsKmsClient implements KmsClient {
+  /*
+   * property for enabling data key generation using AWS KMS service, by default it's enabled,
+   * please set this property to "false" to disable key generation by AWS KMS server.
+   */
+  public static final String ENABLE_KEY_GENERATION = "kms.client.aws.key.generation.enabled";
+  // property for specifying data key generation key spec: AES_256 or AES_128
```
flyrain, line 41: Nit: suggest an empty line above the comment.

flyrain (approved):
LGTM

JiJiTang:
@danielcweeks , thanks VM for the good comments.  Please find my comments as the following:

1. Default AWS KMS client created by default AWSClientFactory relies on normal AWS credential provider which works out of box for typical AWS environment without the need of customoisation.
2. I didn't use `AwsClientFactories.from`  because  AWSClientFactory relies on catalog properties for initialisation but Iceberg Encryption KMS client `initialise` method only brings table properties not catalog properties in the scope.

IMHO, support key generation or not should be per table configuration and configured via table properties, not sure we should configure via AWSClientFactory, cc @ggershinsky for more inputs.

ggershinsky:
Yep, a KMS client doesn't receive catalog properties (which makes sense, because a KMS and a Catalog are two unrelated services). It can be configured via table properties and via custom channels. If the default AWS KMS client works out of box for a typical environment, then maybe we can start without custom channels; allowing users to override this as needed.

ggershinsky:
not any users, but org security experts, see eg our earlier docs at https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#kms-client . Let us think a bit more about this PR.

PR #4716, opened 2022-05-06
[CORE] - Add remaining parsers for MetadataUpdate implementations

kbendick:
Adds the remaining parsers for metadata update subclasses.

Opening this as a draft so I can compare and because some field renaming in a tangentially related PR: https://github.com/apache/iceberg/pull/4693

I'll probably move the field renaming over here if they don't get merged in over there before.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdate.java
@@ -233,16 +233,16 @@ public interface MetadataUpdate extends Serializable {
   }
 
   class SetSnapshotRef implements MetadataUpdate {
-    private final String name;
+    private final String refName;
     private final Long snapshotId;
     private final SnapshotRefType type;
     private Integer minSnapshotsToKeep;
     private Long maxSnapshotAgeMs;
     private Long maxRefAgeMs;
 
-    public SetSnapshotRef(String name, Long snapshotId, SnapshotRefType type, Integer minSnapshotsToKeep,
+    public SetSnapshotRef(String refName, Long snapshotId, SnapshotRefType type, Integer minSnapshotsToKeep,
                           Long maxSnapshotAgeMs, Long maxRefAgeMs) {
-      this.name = name;
+      this.refName = refName;
       this.snapshotId = snapshotId;
       this.type = type;
       this.minSnapshotsToKeep = minSnapshotsToKeep;
@@ -251,7 +251,7 @@ public interface MetadataUpdate extends Serializable {
     }
 
     public String name() {
-      return name;
+      return refName;
     }
 
     public String type() {
@@ -281,7 +281,7 @@ public interface MetadataUpdate extends Serializable {
           .maxSnapshotAgeMs(maxSnapshotAgeMs)
           .maxRefAgeMs(maxRefAgeMs)
           .build();
-      metadataBuilder.setRef(name, ref);
+      metadataBuilder.setRef(refName, ref);
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
@@ -26,8 +26,11 @@ import java.io.StringWriter;
 import java.io.UncheckedIOException;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.util.JsonUtil;
 
 public class MetadataUpdateParser {
@@ -53,6 +56,9 @@ public class MetadataUpdateParser {
   static final String REMOVE_PROPERTIES = "remove-properties";
   static final String SET_LOCATION = "set-location";
 
+  // AssignUUID
+  private static final String UUID = "uuid";
+
   // UpgradeFormatVersion
   private static final String FORMAT_VERSION = "format-version";
 
@@ -71,6 +77,32 @@ public class MetadataUpdateParser {
 
   // AddSortOrder
   private static final String SORT_ORDER = "sort-order";
+
+  // SetDefaultSortOrder
+  private static final String SORT_ORDER_ID = "sort-order-id";
+
+  // AddSnapshot
+  private static final String SNAPSHOT = "snapshot";
+
+  // RemoveSnapshot
+  private static final String SNAPSHOT_IDS = "snapshot-ids";
+
+  // SetSnapshotRef
+  private static final String REF_NAME = "ref-name";
+  private static final String SNAPSHOT_ID = "snapshot-id";
+  private static final String TYPE = "type";
+  private static final String MIN_SNAPSHOTS_TO_KEEP = "min-snapshots-to-keep";
+  private static final String MAX_SNAPSHOT_AGE_MS = "max-snapshot-age-ms";
+  private static final String MAX_REF_AGE_MS = "max-ref-age-ms";
+
+  // SetProperties
+  private static final String UPDATED = "updated";
+
+  // RemoveProperties
+  private static final String REMOVED = "removed";
+
+  // SetLocation
+  private static final String LOCATION = "location";
 
   private static final Map<Class<? extends MetadataUpdate>, String> ACTIONS = ImmutableMap
       .<Class<? extends MetadataUpdate>, String>builder()
@@ -112,12 +144,18 @@ public class MetadataUpdateParser {
   public static void toJson(MetadataUpdate metadataUpdate, JsonGenerator generator) throws IOException {
     String updateAction = ACTIONS.get(metadataUpdate.getClass());
 
+    // Provide better exception message than the NPE thrown by writing null for the update action, which is required
+    Preconditions.checkArgument(updateAction != null,
+        "Cannot convert metadata update to json. Unrecognized metadata update type: {}",
+        metadataUpdate.getClass().getName());
+
     generator.writeStartObject();
     generator.writeStringField(ACTION, updateAction);
 
     switch (updateAction) {
       case ASSIGN_UUID:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdate#toJson for AssignUUID");
+        writeAssignUUID((MetadataUpdate.AssignUUID) metadataUpdate, generator);
+        break;
       case UPGRADE_FORMAT_VERSION:
         writeUpgradeFormatVersion((MetadataUpdate.UpgradeFormatVersion) metadataUpdate, generator);
         break;
@@ -137,13 +175,26 @@ public class MetadataUpdateParser {
         writeAddSortOrder((MetadataUpdate.AddSortOrder) metadataUpdate, generator);
         break;
       case SET_DEFAULT_SORT_ORDER:
+        writeSetDefaultSortOrder((MetadataUpdate.SetDefaultSortOrder) metadataUpdate, generator);
+        break;
       case ADD_SNAPSHOT:
+        writeAddSnapshot((MetadataUpdate.AddSnapshot) metadataUpdate, generator);
+        break;
       case REMOVE_SNAPSHOTS:
+        writeRemoveSnapshots((MetadataUpdate.RemoveSnapshot) metadataUpdate, generator);
+        break;
       case SET_SNAPSHOT_REF:
+        writeSetSnapshotRef((MetadataUpdate.SetSnapshotRef) metadataUpdate, generator);
+        break;
       case SET_PROPERTIES:
+        writeSetProperties((MetadataUpdate.SetProperties) metadataUpdate, generator);
+        break;
       case REMOVE_PROPERTIES:
+        writeRemoveProperties((MetadataUpdate.RemoveProperties) metadataUpdate, generator);
+        break;
       case SET_LOCATION:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdate#toJson for " + updateAction);
+        writeSetLocation((MetadataUpdate.SetLocation) metadataUpdate, generator);
+        break;
       default:
         throw new IllegalArgumentException(
             String.format("Cannot convert metadata update to json. Unrecognized action: %s", updateAction));
@@ -174,7 +225,7 @@ public class MetadataUpdateParser {
 
     switch (action) {
       case ASSIGN_UUID:
-        throw new UnsupportedOperationException("Not implemented: AssignUUID");
+        return readAssignUUID(jsonNode);
       case UPGRADE_FORMAT_VERSION:
         return readUpgradeFormatVersion(jsonNode);
       case ADD_SCHEMA:
@@ -188,17 +239,27 @@ public class MetadataUpdateParser {
       case ADD_SORT_ORDER:
         return readAddSortOrder(jsonNode);
       case SET_DEFAULT_SORT_ORDER:
+        return readSetDefaultSortOrder(jsonNode);
       case ADD_SNAPSHOT:
+        return readAddSnapshot(jsonNode);
       case REMOVE_SNAPSHOTS:
+        return readRemoveSnapshots(jsonNode);
       case SET_SNAPSHOT_REF:
+        return readSetSnapshotRef(jsonNode);
       case SET_PROPERTIES:
+        return readSetProperties(jsonNode);
       case REMOVE_PROPERTIES:
+        return readRemoveProperties(jsonNode);
       case SET_LOCATION:
-        throw new UnsupportedOperationException("Not Implemented: MetadataUpdatefromJson for " + action);
+        return readSetLocation(jsonNode);
       default:
         throw new UnsupportedOperationException(
             String.format("Cannot convert metadata update action to json: %s", action));
     }
+  }
+
+  private static void writeAssignUUID(MetadataUpdate.AssignUUID update, JsonGenerator gen) throws IOException {
+    gen.writeStringField(UUID, update.uuid());
   }
 
   private static void writeUpgradeFormatVersion(MetadataUpdate.UpgradeFormatVersion update, JsonGenerator gen)
@@ -232,6 +293,55 @@ public class MetadataUpdateParser {
       throws IOException {
     gen.writeFieldName(SORT_ORDER);
     SortOrderParser.toJson(update.sortOrder(), gen);
+  }
+
+  private static void writeSetDefaultSortOrder(MetadataUpdate.SetDefaultSortOrder update, JsonGenerator gen)
+      throws IOException {
+    gen.writeNumberField(SORT_ORDER_ID, update.sortOrderId());
+  }
+
+  private static void writeAddSnapshot(MetadataUpdate.AddSnapshot update, JsonGenerator gen) throws IOException {
+    gen.writeFieldName(SNAPSHOT);
+    SnapshotParser.toJson(update.snapshot(), gen);
+  }
+
+  // TODO - Reconcile the spec's set-based removal with the current class implementation that only handles one value.
+  private static void writeRemoveSnapshots(MetadataUpdate.RemoveSnapshot update, JsonGenerator gen) throws IOException {
+    gen.writeArrayFieldStart(SNAPSHOT_IDS);
+    for (long snapshotId : ImmutableSet.of(update.snapshotId())) {
+      gen.writeNumber(snapshotId);
+    }
+    gen.writeEndArray();
+  }
+
+  private static void writeSetSnapshotRef(MetadataUpdate.SetSnapshotRef update, JsonGenerator gen) throws IOException {
+    gen.writeStringField(REF_NAME, update.name());
+    gen.writeNumberField(SNAPSHOT_ID, update.snapshotId());
+    gen.writeStringField(TYPE, update.type());
+    JsonUtil.writeIntegerFieldIf(
+        update.minSnapshotsToKeep() != null, MIN_SNAPSHOTS_TO_KEEP, update.minSnapshotsToKeep(), gen);
+    JsonUtil.writeLongFieldIf(update.maxSnapshotAgeMs() != null, MAX_SNAPSHOT_AGE_MS, update.maxSnapshotAgeMs(), gen);
+    JsonUtil.writeLongFieldIf(update.maxRefAgeMs() != null, MAX_REF_AGE_MS, update.maxRefAgeMs(), gen);
+  }
+
+  private static void writeSetProperties(MetadataUpdate.SetProperties update, JsonGenerator gen) throws IOException {
+    gen.writeFieldName(UPDATED);
+    gen.writeObject(update.updated());
+  }
+
+  private static void writeRemoveProperties(MetadataUpdate.RemoveProperties update, JsonGenerator gen)
+      throws IOException {
+    gen.writeFieldName(REMOVED);
+    gen.writeObject(update.removed());
+  }
+
+  private static void writeSetLocation(MetadataUpdate.SetLocation update, JsonGenerator gen) throws IOException {
+    gen.writeStringField(LOCATION, update.location());
+  }
+
+  private static MetadataUpdate readAssignUUID(JsonNode node) {
+    String uuid = JsonUtil.getString(UUID, node);
+    return new MetadataUpdate.AssignUUID(uuid);
   }
 
   private static MetadataUpdate readUpgradeFormatVersion(JsonNode node) {
@@ -270,6 +380,51 @@ public class MetadataUpdateParser {
     JsonNode sortOrderNode = node.get(SORT_ORDER);
     UnboundSortOrder sortOrder = SortOrderParser.fromJson(sortOrderNode);
     return new MetadataUpdate.AddSortOrder(sortOrder);
+  }
+
+  private static MetadataUpdate readSetDefaultSortOrder(JsonNode node) {
+    int sortOrderId = JsonUtil.getInt(SORT_ORDER_ID, node);
+    return new MetadataUpdate.SetDefaultSortOrder(sortOrderId);
+  }
+
+  private static MetadataUpdate readAddSnapshot(JsonNode node) {
+    Preconditions.checkArgument(node.has(SNAPSHOT), "Cannot parse missing field: snapshot");
+    Snapshot snapshot = SnapshotParser.fromJson(null, node.get(SNAPSHOT));
+    return new MetadataUpdate.AddSnapshot(snapshot);
+  }
+
+  private static MetadataUpdate readRemoveSnapshots(JsonNode node) {
+    Set<Long> snapshotIds = JsonUtil.getLongSetOrNull(SNAPSHOT_IDS, node);
+    Preconditions.checkArgument(snapshotIds != null && snapshotIds.size() == 1,
+        "Invalid set of snapshot ids to remove. Expected one value but received: %s", snapshotIds);
+    Long snapshotId = Iterables.getOnlyElement(snapshotIds);
+    return new MetadataUpdate.RemoveSnapshot(snapshotId);
+  }
+
+  private static MetadataUpdate readSetSnapshotRef(JsonNode node) {
+    String refName = JsonUtil.getString(REF_NAME, node);
+    long snapshotId = JsonUtil.getLong(SNAPSHOT_ID, node);
+    SnapshotRefType type = SnapshotRefType.valueOf(JsonUtil.getString(TYPE, node).toUpperCase(Locale.ENGLISH));
+    Integer minSnapshotsToKeep = JsonUtil.getIntOrNull(MIN_SNAPSHOTS_TO_KEEP, node);
+    Long maxSnapshotAgeMs = JsonUtil.getLongOrNull(MAX_SNAPSHOT_AGE_MS, node);
+    Long maxRefAgeMs = JsonUtil.getLongOrNull(MAX_REF_AGE_MS, node);
+    return new MetadataUpdate.SetSnapshotRef(
+        refName, snapshotId, type, minSnapshotsToKeep, maxSnapshotAgeMs, maxRefAgeMs);
+  }
+
+  private static MetadataUpdate readSetProperties(JsonNode node) {
+    Map<String, String> updated = JsonUtil.getStringMap(UPDATED, node);
+    return new MetadataUpdate.SetProperties(updated);
+  }
+
+  private static MetadataUpdate readRemoveProperties(JsonNode node) {
+    Set<String> removed = JsonUtil.getStringSet(REMOVED, node);
+    return new MetadataUpdate.RemoveProperties(removed);
+  }
+
+  private static MetadataUpdate readSetLocation(JsonNode node) {
+    String location = JsonUtil.getString(LOCATION, node);
+    return new MetadataUpdate.SetLocation(location);
   }
 }
 
--- a/core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
+++ b/core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
@@ -20,13 +20,20 @@
 package org.apache.iceberg;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.IntStream;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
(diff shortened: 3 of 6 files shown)


core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -72,6 +75,20 @@ private MetadataUpdateParser() {
   // AddSortOrder
   private static final String SORT_ORDER = "sort-order";
 
+  // SetDefaultSortOrder
+  private static final String SORT_ORDER_ID = "sort-order-id";
```
kbendick, line 79: This field is called `order-id` in the spec.

Due to discussion in #4693  , we've decided to align the names `order-id` and `write-order-id` on `sort-order-id` (and `default-sort-order-id`) to match the existing code.

kbendick:
Given this is a much larger change set than I'd expected, and one or two of the classes need to be aligned with the spec, I'm going to use this PR as a template and open a series of PRs.

I'll add tasks to that to make it easier to track but the general goal is:
1) Full shell of all needed functions, so individual changes are easy to make.
2) Add implementations and tests for the simplest classes
3) Add implementations and tests for any classes that are simple except for needing to update `JsonUtil`
4) Open a PR for classes that require a closer look, like `RemoveSnapshots` which allows for a set of snapshotIds to remove per the spec, but the implementation only accepts one value.

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
```diff
@@ -234,6 +234,7 @@ public void applyTo(TableMetadata.Builder metadataBuilder) {
 
   class SetSnapshotRef implements MetadataUpdate {
     private final String name;
+    // TODO - this can be primitive.
     private final Long snapshotId;
```
kbendick, line 238: TODO / possible follow up - This can be made primitive

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
```diff
@@ -250,6 +251,8 @@ public SetSnapshotRef(String name, Long snapshotId, SnapshotRefType type, Intege
       this.maxRefAgeMs = maxRefAgeMs;
     }
 
+    // TODO - Consider renaming to `refName`, as the JSON is called `ref`. This also matches the equivalent
+    //  UpdateRequirement.
     public String name() {
```
kbendick, line 256: TODO / Possible Follow Up - Consider renaming to `refName`, as the JSON is called `ref`. This also matches the equivalent UpdateRequirement.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -234,6 +290,58 @@ private static void writeAddSortOrder(MetadataUpdate.AddSortOrder update, JsonGe
...
+    gen.writeStringField(REF_NAME, update.name());
+    gen.writeNumberField(SNAPSHOT_ID, update.snapshotId());
+    gen.writeStringField(TYPE, update.type());
+    // TODO - Should we explicitly set the null values?
+    JsonUtil.writeIntegerFieldIf(update.minSnapshotsToKeep() != null,
+        MIN_SNAPSHOTS_TO_KEEP, update.minSnapshotsToKeep(), gen);
+    JsonUtil.writeLongFieldIf(update.maxSnapshotAgeMs() != null,
+        MAX_SNAPSHOT_AGE_MS, update.maxSnapshotAgeMs(), gen);
+    JsonUtil.writeLongFieldIf(update.maxRefAgeMs() != null, MAX_REF_AGE_MS,
+        update.maxRefAgeMs(), gen);
```
kbendick, line 322: Question: Should we explicitly set the `null` values or not?

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +379,55 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
+  private static MetadataUpdate readSetDefaultSortOrder(JsonNode node) {
+    int sortOrderId = JsonUtil.getInt(SORT_ORDER_ID, node);
+    return new MetadataUpdate.SetDefaultSortOrder(sortOrderId);
+  }
+
+  // TODO - Return to this when FileIO isn't required.
+  private static MetadataUpdate readAddSnapshot(JsonNode node) {
+    Snapshot snapshot = SnapshotParser.fromJson(null, node);
+    return new MetadataUpdate.AddSnapshot(snapshot);
+  }
```
kbendick, line 391: We'll need to return to this when `FileIO` can be plumbed through.

Happy to leave it out of this PR for now.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +379,55 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
+    Long maxRefAgeMs = JsonUtil.getLongOrNull(MAX_REF_AGE_MS, node);
+    return new MetadataUpdate.SetSnapshotRef(refName, snapshotId, type, minSnapshotsToKeep, maxSnapshotAgeMs, maxRefAgeMs);
+  }
+
+  private static MetadataUpdate readSetProperties(JsonNode node) {
+    // TODO - Instead of this we could just call removeProperties, which would avoid the need for null fields.
+    //   but that's probably not intuitive for the user.
+    Map<String, String> updated = JsonUtil.getStringMapWithNullValues(UPDATED, node);
+    return new MetadataUpdate.SetProperties(updated);
+  }
```
kbendick, line 418: The introduction of `JsonUtil::getStringMapWithNullValues` was added to support setting properties to null.

This could also be achieved via `removeProperties`, but that might not be intuitive for the user.

Happy to split this part into a separate PR.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -350,9 +485,51 @@ private static void assertEqualsAddSortOrder(
...
+    Assert.assertEquals("Sort order id should be the same", expected.sortOrderId(), actual.sortOrderId());
+  }
+
+  // TODO - Come back to this when FileIO is injected. If this class extends TableTestBase, the validateSnapshot
+  //  methods can be reused. Also, this needs to handle V1 vs V2 for assertions. Or casting to BaseSnapshot,
+  //  BaseSnapshot::equals can be used.
+  private static void assertEqualsAddSnapshot(
+      MetadataUpdate.AddSnapshot expected, MetadataUpdate.AddSnapshot actual) {
+    Assert.fail("MetadataUpdate equality checking for AddSnapshot is not implemented yet");
+  }
```
kbendick, line 505: TODO / Follow up - Come back to this when FileIO is injected. If this class extends TableTestBase, the validateSnapshot methods can be reused. Also, this possibly needs to handle the difference between V1 vs V2 for assertions. Or casting to BaseSnapshot, BaseSnapshot::equals can be used.

Happy to remove this in this PR and handle `EqualsAddSnapshot` in another PR.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -350,9 +485,51 @@ private static void assertEqualsAddSortOrder(
...
+  private static void assertEqualsRemoveSnapshots(
+      MetadataUpdate.RemoveSnapshot expected, MetadataUpdate.RemoveSnapshot actual) {
+    Assert.assertEquals("Snapshots to remove should be the same", expected.snapshotId(), actual.snapshotId());
+  }
+
+  // See TODOs in MetadataUpdate.SetSnapshotRef.
+  private static void assertEqualsSetSnapshotRef(
+      MetadataUpdate.SetSnapshotRef expected, MetadataUpdate.SetSnapshotRef actual) {
+    Assert.fail("MetadataUpdate equality checking for SetSnapshotRef is not implemented yet");
+  }
```
kbendick, line 516: See the notes in `MetadataUpdate.SetSnapshotRef` about possible things to update.

Happy to remove this from this PR and place it elsewhere.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -251,11 +274,122 @@ public void testAddSortOrderFromJson() {
...
+  @Test
+  public void testSetDefaultSortOrderFromJson() {
+    String action = MetadataUpdateParser.SET_DEFAULT_SORT_ORDER;
+    int sortOrderId = 2;
+    String json = String.format("{\"action\":\"%s\",\"sort-order-id\":%d}", action, sortOrderId);
+    MetadataUpdate expected = new MetadataUpdate.SetDefaultSortOrder(sortOrderId);
+    assertEquals(action, expected, MetadataUpdateParser.fromJson(json));
+  }
+
+  // TODO - AddSnapshot tests.
```
kbendick, line 296: Should I implement this in another PR instead? We still need to plumb through `FileIO` to `BaseSnapshot` so I think probably so.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -251,11 +274,122 @@ public void testAddSortOrderFromJson() {
...
+
+  // TODO - Tests that handle multiple snapshot ids.
+  @Test
+  public void testRemoveSnapshotsFromJson() {
```
kbendick, line 300: Currently, the spec allows for `RemoveSnapshots` to have multiple snapshots. But the code only allows for one.

Should we implement in another PR, or move forward with just 1 allowed temporarily?

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -234,6 +290,57 @@ private static void writeAddSortOrder(MetadataUpdate.AddSortOrder update, JsonGe
...
+    gen.writeNumberField(SORT_ORDER_ID, update.sortOrderId());
+  }
+
+  private static void writeAddSnapshot(MetadataUpdate.AddSnapshot update, JsonGenerator gen) throws IOException {
+    gen.writeFieldName(SNAPSHOT);
+    SnapshotParser.toJson(update.snapshot(), gen);
+  }
+
+  // TODO - Reconcile the spec's set-based removal with the current class implementation that only handles one value.
+  private static void writeRemoveSnapshots(MetadataUpdate.RemoveSnapshot update, JsonGenerator gen) throws IOException {
```
kbendick, line 304: Same note about multiple snapshots being allowed in the code. Although for now this should be fine.

I'll hold off on editing this until more input re: following up `RemoveSnapshots` in another PR or simply only supporting one value is determined.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -350,9 +502,52 @@ private static void assertEqualsAddSortOrder(
           UnboundSortOrder.UnboundSortField actualField = actual.sortOrder().fields().get(i);
           Assert.assertTrue("Fields of the sort order should be the same",
               expectedField.sourceId() == actualField.sourceId() &&
-              expectedField.nullOrder().equals(actualField.nullOrder()) &&
-              expectedField.direction().equals(actualField.direction()) &&
-              Objects.equals(expectedField.transformAsString(), actualField.transformAsString()));
+                  expectedField.nullOrder().equals(actualField.nullOrder()) &&
+                  expectedField.direction().equals(actualField.direction()) &&
+                  Objects.equals(expectedField.transformAsString(), actualField.transformAsString()));
```
rdblue, line 507: Accidental whitespace change?

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
rdblue: I think they should be omitted.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -72,6 +78,32 @@ private MetadataUpdateParser() {
...
+  private static final String SORT_ORDER_ID = "sort-order-id";
+
+  // AddSnapshot
+  private static final String SNAPSHOT = "snapshot";
+
+  // RemoveSnapshot
+  private static final String SNAPSHOT_IDS = "snapshot-ids";
+
+  // SetSnapshotRef
+  private static final String REF_NAME = "ref";
```
rdblue, line 91: I'd support updating this to `ref-name` here and in the spec.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +378,51 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
+
+  private static MetadataUpdate readAddSnapshot(JsonNode node) {
+    Snapshot snapshot = SnapshotParser.fromJson(null, node);
+    return new MetadataUpdate.AddSnapshot(snapshot);
+  }
+
+  private static MetadataUpdate readRemoveSnapshots(JsonNode node) {
+    Set<Long> snapshotIds = JsonUtil.getLongSetOrNull(SNAPSHOT_IDS, node);
+    Preconditions.checkArgument(snapshotIds != null && snapshotIds.size() == 1,
+        "Invalid set of snapshot ids to remove. Expected one value but received: %o", snapshotIds);
```
rdblue, line 395: Is `%o` correct? I would expect `%s` or maybe `%d`.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +378,51 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
+    return new MetadataUpdate.SetDefaultSortOrder(sortOrderId);
+  }
+
+  private static MetadataUpdate readAddSnapshot(JsonNode node) {
+    Snapshot snapshot = SnapshotParser.fromJson(null, node);
+    return new MetadataUpdate.AddSnapshot(snapshot);
+  }
+
+  private static MetadataUpdate readRemoveSnapshots(JsonNode node) {
+    Set<Long> snapshotIds = JsonUtil.getLongSetOrNull(SNAPSHOT_IDS, node);
```
rdblue, line 393: When could this be null? I think this is correct (so this comment is minor) but I'd normally expect to use `getLogSet` that handles the null case internally. This is fine because there's a good error message.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +378,51 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
+    SnapshotRefType type = SnapshotRefType.valueOf(JsonUtil.getString(TYPE, node).toUpperCase(Locale.ENGLISH));
+    Integer minSnapshotsToKeep = JsonUtil.getIntOrNull(MIN_SNAPSHOTS_TO_KEEP, node);
+    Long maxSnapshotAgeMs = JsonUtil.getLongOrNull(MAX_SNAPSHOT_AGE_MS, node);
+    Long maxRefAgeMs = JsonUtil.getLongOrNull(MAX_REF_AGE_MS, node);
+    return new MetadataUpdate.SetSnapshotRef(
+        refName, snapshotId, type, minSnapshotsToKeep, maxSnapshotAgeMs, maxRefAgeMs);
+  }
+
+  private static MetadataUpdate readSetProperties(JsonNode node) {
+    Map<String, String> updated = JsonUtil.getStringMapWithNullValues(UPDATED, node);
```
rdblue, line 412: Values can be `null`? Is that allowed in a normal table?

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
rdblue: Ah, I think the right way to remove properties at this level is to use the `RemovePropertiesUpdate` from the spec.

Otherwise, this is ambiguous because someone may be trying to set a property to null, instead of removing it. That's what I thought you were trying to support with my comment above. But I don't think that we allow setting properties to null.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +378,51 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
+    return new MetadataUpdate.SetProperties(updated);
+  }
+
+  private static MetadataUpdate readRemoveProperties(JsonNode node) {
+    Preconditions.checkArgument(node.hasNonNull(REMOVED), "Missing field: removed");
+    JsonNode removedNode = node.get(REMOVED);
+    String[] removed = JsonUtil.getStringArray(removedNode);
```
rdblue, line 419: What about using `getStringSet`? There's a `getLongSetOrNull` method.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
rdblue: No, I don't think that the `FileIO` should not affect the test done here. We just need to make sure that the `Snapshot` info matches.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
rdblue: Move forward with just 1 allowed, temporarily. I think in a follow up PR, we should support multiple snapshots in `RemoveSnapshot` (renamed to `RemoveSnapshots`).

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -251,11 +274,120 @@ public void testAddSortOrderFromJson() {
...
+    MetadataUpdate expected = new MetadataUpdate.RemoveSnapshot(snapshotIds[0]);
+    assertEquals(action, expected, MetadataUpdateParser.fromJson(json));
+  }
+
+  // TODO - Tests that handle multiple snapshot ids.
+  @Test
+  public void testRemoveSnapshotsToJson() {
+    String action = MetadataUpdateParser.REMOVE_SNAPSHOTS;
+    long[] snapshotIds = { 2L };
+    String expected = String.format("{\"action\":\"%s\",\"snapshot-ids\":%s}", action, Arrays.toString(snapshotIds));
```
rdblue, line 313: I think it would be better to just embed `[ 2 ]` rather than relying on `Arrays.toString`.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -251,11 +274,120 @@ public void testAddSortOrderFromJson() {
...
+
+  // TODO - SetSnapshotRef tests.
+
+  @Test
+  public void testSetPropertiesFromJson() {
+    String action = MetadataUpdateParser.SET_PROPERTIES;
+    Map<String, String> props = Maps.newHashMap();
+    props.put("prop1", "val1");
+    props.put("prop2", null);
+    String propsMap = "{\"prop1\":\"val1\",\"prop2\":null}";
```
rdblue, line 328: Probably don't need a null test.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
rdblue: I think this should check that the snapshot ID, manifest list location, and summary map are equal.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
rdblue: I don't see the note you're referring to. The parser looks like it supports this, so I think it can be implemented. This shouldn't be affected by whether or not the optional fields are serialized using `null` or are omitted. Though you will want tests for reading all cases (omitted, null, non-null) and make sure they match here.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -271,5 +378,51 @@ private static MetadataUpdate readAddSortOrder(JsonNode node) {
...
     return new MetadataUpdate.AddSortOrder(sortOrder);
   }
+
+  private static MetadataUpdate readSetDefaultSortOrder(JsonNode node) {
+    int sortOrderId = JsonUtil.getInt(SORT_ORDER_ID, node);
+    return new MetadataUpdate.SetDefaultSortOrder(sortOrderId);
+  }
+
+  private static MetadataUpdate readAddSnapshot(JsonNode node) {
+    Snapshot snapshot = SnapshotParser.fromJson(null, node);
```
rdblue, line 388: Looks like this is slightly wrong. The node passed in is the update, but it should pass `node.get(SNAPSHOT)` here instead.

rdblue: This seems to be working:

```
    Preconditions.checkArgument(node.has(SNAPSHOT), "Cannot parse missing field: snapshot");
    Snapshot snapshot = SnapshotParser.fromJson(null, node.get(SNAPSHOT));
```

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: The line for `MIN_SNAPSHOTS_TO_KEEP` can't fit on one line. Everything else seems like it can.

Updating.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: You're correct. This should never be null (as evidence by the null check below).

I added the `getLongSetOrNull` function essentially because it mirrored the `getIntegerSetOrNull` function and still needed a preconditions check here for the size (given we don't currently support multiple values).

I can update to `getLongSet` given it's added in this PR, but the Precondition check will still be needed.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: I just checked in JShell and using `%s` directly worked. Will update.

```
jshell> String.format("Format Set<Integer> with o: %o", s);
|  Exception java.util.IllegalFormatConversionException: o != java.util.ImmutableCollections$SetN
|        at Formatter$FormatSpecifier.failConversion (Formatter.java:4426)
|        at Formatter$FormatSpecifier.printInteger (Formatter.java:2938)
|        at Formatter$FormatSpecifier.print (Formatter.java:2892)
|        at Formatter.format (Formatter.java:2673)
|        at Formatter.format (Formatter.java:2609)
|        at String.format (String.java:2897)
|        at (#7:1)

jshell> String.format("Format Set<Integer> with s: %s", s);
$10 ==> "Format Set<Integer> with s: [1, 2, 3]"
```

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: +1. Do we also want to change the corresponding update requirement field? We had the same concern there.

https://github.com/apache/iceberg/blob/3586e140dc3fbe0b0414464497be3ccd4ca0927d/core/src/main/java/org/apache/iceberg/rest/requests/UpdateRequirementParser.java#L55

kbendick: Here's the associated class `UpdateRequirement.AsserrRefSnapshotId`: https://github.com/apache/iceberg/blob/3586e140dc3fbe0b0414464497be3ccd4ca0927d/core/src/main/java/org/apache/iceberg/rest/requests/UpdateTableRequest.java#L261-L272

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: No, they shouldn't be null. However, I could see somebody using this API to unset values given that it's referred to as `updated` in the field name.

Users should really use `RemoveProperties` for that purpose so happy to remove this to reduce complexity.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: +1. That's what I was trying to support, though as you mentioned people should be using `RemoveProperties`.

So this might be giving them just enough rope to hang themselves.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Yeah happy to add that. I might move some of these changes to another PR so that we can also add testing for `JsonUtil` class, but for now I'll continue to stage the work here (and possibly we'll leave it here).

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
```diff
@@ -233,25 +233,25 @@ public void applyTo(TableMetadata.Builder metadataBuilder) {
   }
 
   class SetSnapshotRef implements MetadataUpdate {
-    private final String name;
+    private final String refName;
```
kbendick, line 236: Given we updated the spec to use `ref-name`, I also changed the field name and the method name.

I can do this in a separate PR (as well as update the spec in that same PR) or I can leave the class's code as is for now. WDYT? Changing it here does admittedly make the PR's scope somewhat broad for what it touches.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: Yeah good call 👍

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: Changed to ImmutableMap and added a test for throwing when reading `null`.

Not allowing `null` as a value _before converting to json_ seems like something that the `MetadataUpdate.SetProperties` class should be responsible for. So I only tested in `fromJson`.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -251,11 +274,138 @@ public void testAddSortOrderFromJson() {
...
+    String json = String.format("{\"action\":\"%s\",\"updated\":%s}", action, propsMap);
+    AssertHelpers.assertThrows(
+        "Parsing updates from SetProperties with a property set to null should throw",
+        IllegalArgumentException.class,
+        "Cannot parse prop2 to a string value: null",
```
kbendick, line 346: This is the error message thrown, although combined with the full stack trace I think it should be clear as we'll have `MetadataUpdateParser.fromJson` in it.

We might want to move some of the validation logic into `MetadataUpdate.SetProperties` though, particularly when the Java object is being created (so when the client would be sending the payload using `toJson` instead).

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -72,6 +79,32 @@ private MetadataUpdateParser() {
...
+  private static final String SORT_ORDER_ID = "sort-order-id";
+
+  // AddSnapshot
+  private static final String SNAPSHOT = "snapshot";
+
+  // RemoveSnapshot
+  private static final String SNAPSHOT_IDS = "snapshot-ids";
+
+  // SetSnapshotRef
+  private static final String REF_NAME = "ref-name";
```
kbendick, line 92: We updated this field from `ref` to `ref-name` in the spec.

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
kbendick: I did update this to `refName`, as we updated the spec to refer to this field as `ref-name` instead of simply `name`.

I can revert the changes to the code though.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Updated to put as many on one line as possible.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: I think that's sufficient.

What about the snapshots schema id, operation, and the timestamp? Possibly overkill as the three you mentioned (especially the combination of snapshot ID and manifest list location) could arguably be considered sufficient.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: reverted.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: I updated this to `ref-name`, as well as updated the spec.

I also updated the code to use `refName` as the field in `SetSnapshotRef` as well as the accessor (to match `AssertRefSnapshotId`. Please let me know if you think the changes for the accessor are too much for this PR.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Updated to use `%s` as demonstrated in jshell.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: I added `getStringSet` and changed to use that. We should introduce a test class for `JsonUtil`, but I can take care of that in a follow up PR or open an issue for it as it could be a good issue for a new contributor.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: Updated as mentioned.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: I have added all of those tests. I basically copied the tests from `TestSnapshotRefParser` and updated them for this, with versions with versions that read all cases.

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
```diff
@@ -233,25 +233,25 @@ public void applyTo(TableMetadata.Builder metadataBuilder) {
...
       this.snapshotId = snapshotId;
       this.type = type;
       this.minSnapshotsToKeep = minSnapshotsToKeep;
       this.maxSnapshotAgeMs = maxSnapshotAgeMs;
       this.maxRefAgeMs = maxRefAgeMs;
     }
 
-    public String name() {
-      return name;
+    public String refName() {
```
rdblue, line 253: Probably better not to rename this, since it requires touching two extra files.

rdblue (approved):

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
kbendick: Reverted.

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
kbendick: Change tot he code has been reverted.

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
kbendick: This TODO is no longer written in the code, but doesn't show as outdated because the comment is on the line below.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
kbendick: We actually won't need to return to it for when FileIO is plumbed through for testing purposes, but we will want to consider it still in case.

I'm working on that PR now.

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: I added timestamp, schema id and parent id as well as the ones you stated. That should be more than sufficient to verify that the existing parser is working (which we reuse).

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
kbendick: Yeah I moved forward with just 1 allowed. Would you like me to change the class name to be singular for now? Currently the singular constructor class is `RemovePartitions`.

We can either accept one value and then collect it into a set and expose two constructors or make two classes.

rdblue (approved):

Issue #4631, opened 2022-04-25
Python: PartitionSpec Construction


dramaticlly:
From both issue https://github.com/apache/iceberg/issues/3228#issuecomment-1095661430 and pull request https://github.com/apache/iceberg/pull/3407#discussion_r738976688, there's some discussion on how do we build the immutable PartitionSpec with proper validation against every PartitionField 

- in java we leveraged the builder pattern
- in python there's no perfect substitution, some proposed to look into  `PartitionSpec` dunder init.

In this issue, I want to highlight the list of validation we need to enforce before we can construct a valid PartitionSpec. 

Considering its attribute, the PartitionSpec has
- spec_id (int) as unique identifier
- schema (Schema) as data table schema
- list of PartitionField
- lastAssignedPartitionFieldId (int): from offset of 1000, increment 1 for each partitionField

In the meantime, the PartitionField is defined by 
- field_id (int) as unique identifier
- source_id (int), the source id found in table schema
- name (str), human-readable identifier
- transform(Transform), the transformation from source column in data table to partition values, include identity, bucket etc

Validation we want to enforce when build the partitionSpec
- for each partitionField in `__init__` parameter, we want to make sure
  - [`findSourceColumn`](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L399), source column in table schema can be found by referencing its name
  - [`checkAndAddPartitionName`](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L364),  the new name is neither null nor empty, not duplicated
  - [`checkForRedundantPartitions`](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L385), there's no collision on PartitionField in PartitionSpec if we group by tuple of source column id and transform
  - [`checkCompatibility`](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L539), lastly we need to make sure source column in table schema have a type such that its type is not null, and is of iceberg primitive type, also the registered transform can transform the given type

dramaticlly:
Since there's some sporadic discussion in both PR and slack (which I cannot see it history unfortunately), so I want to collect the opinion of yours to see if above list of validations above are enough to construct the PartitionSpec in python without the `builder`

CC @rdblue  @samredai  @jun-he

samredai:
[...]

PR #4717, opened 2022-05-06
Python: Add PartitionSpec

dramaticlly:
2nd step of complete the https://github.com/apache/iceberg/issues/3228

This change include the PartitionSpec but not include its builder part as I want to keep the changelist as small as possible. As suggested in https://github.com/apache/iceberg/issues/4631, the construction of PartitionSpec shall rely on its Builder class with proper checks and such logic shall not be duplicated in its dunder init methods.

```python
from iceberg.schema import Schema
from iceberg.transforms import bucket
from iceberg.types import BooleanType, IntegerType, NestedField, StringType
from iceberg.table.partitioning import PartitionSpec, PartitionField

table_schema = Schema(
        NestedField(field_id=1, name="foo", field_type=StringType(), is_optional=False),
        NestedField(field_id=2, name="bar", field_type=IntegerType(), is_optional=True),
        NestedField(field_id=3, name="baz", field_type=BooleanType(), is_optional=False),
        schema_id=1,
        identifier_field_ids=[1],
    )
bucket_transform = bucket(IntegerType(), 100)
foo_field = PartitionField(source_id=1, field_id=1001, transform=bucket_transform, name="foo_bucket")
partition_spec = PartitionSpec(schema=table_schema, spec_id=0, fields=(foo_field,), last_assigned_field_id=1001)

>>> partition_spec
PartitionSpec: [
  1001: foo_bucket: bucket[100](1)
]
``` 

I will come up with follow up on PartitionSpecBuilder class

CC @samredai @rdblue @dhruv-pratap

### DIFF
--- a/python/src/iceberg/table/partitioning.py
+++ b/python/src/iceberg/table/partitioning.py
@@ -14,10 +14,12 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from typing import Dict, List, Tuple
+from typing import Dict, Iterable, List, Tuple
 
 from iceberg.schema import Schema
 from iceberg.transforms import Transform
+
+_PARTITION_DATA_ID_START: int = 1000
 
 
 class PartitionField:
@@ -83,15 +85,13 @@ class PartitionSpec:
         last_assigned_field_id(int): auto-increment partition field id starting from PARTITION_DATA_ID_START
     """
 
-    PARTITION_DATA_ID_START: int = 1000
-
-    def __init__(self, schema: Schema, spec_id: int, fields: Tuple[PartitionField], last_assigned_field_id: int):
+    def __init__(self, schema: Schema, spec_id: int, fields: Iterable[PartitionField], last_assigned_field_id: int):
         self._schema = schema
         self._spec_id = spec_id
-        self._fields = fields
+        self._fields = tuple(fields)
         self._last_assigned_field_id = last_assigned_field_id
         # derived
-        self.fields_by_source_id: Dict[int, List[PartitionField]] = {}
+        self._fields_by_source_id: Dict[int, List[PartitionField]] = {}
 
     @property
     def schema(self) -> Schema:
@@ -102,7 +102,7 @@ class PartitionSpec:
         return self._spec_id
 
     @property
-    def fields(self) -> Tuple[PartitionField]:
+    def fields(self) -> Tuple[PartitionField, ...]:
         return self._fields
 
     @property
@@ -131,28 +131,26 @@ class PartitionSpec:
     def is_unpartitioned(self) -> bool:
         return len(self.fields) < 1
 
-    def get_fields_by_source_id(self, filed_id: int) -> List[PartitionField]:
-        if not self.fields_by_source_id:
+    def fields_by_source_id(self, field_id: int) -> List[PartitionField]:
+        if not self._fields_by_source_id:
             for partition_field in self.fields:
                 source_column = self.schema.find_column_name(partition_field.source_id)
                 if not source_column:
                     raise ValueError(f"Cannot find source column: {partition_field.source_id}")
-                existing = self.fields_by_source_id.get(partition_field.source_id, [])
+                existing = self._fields_by_source_id.get(partition_field.source_id, [])
                 existing.append(partition_field)
-                self.fields_by_source_id[partition_field.source_id] = existing
-        return self.fields_by_source_id.get(filed_id)  # type: ignore
+                self._fields_by_source_id[partition_field.source_id] = existing
+        return self._fields_by_source_id[field_id]
 
     def compatible_with(self, other) -> bool:
         """
         Returns true if this partition spec is equivalent to the other, with partition field_id ignored.
         That is, if both specs have the same number of fields, field order, field name, source column ids, and transforms.
         """
-        if self.__eq__(other):
+        if self == other:
             return True
-
         if len(self.fields) != len(other.fields):
             return False
-
         for index in range(len(self.fields)):
             this_field = self.fields[index]
             that_field = other.fields[index]
@@ -162,5 +160,4 @@ class PartitionSpec:
                 or this_field.name != that_field.name
             ):
                 return False
-
         return True
--- a/python/src/iceberg/table/partitioning.py
+++ b/python/src/iceberg/table/partitioning.py
@@ -14,6 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from typing import Dict, List, Tuple
+
+from iceberg.schema import Schema
 from iceberg.transforms import Transform
 
 
@@ -64,3 +67,100 @@ class PartitionField:
 
     def __repr__(self):
         return f"PartitionField(field_id={self.field_id}, name={self.name}, transform={repr(self.transform)}, source_id={self.source_id})"
+
+    def __hash__(self):
+        return hash((self.source_id, self.field_id, self.name, self.transform))
+
+
+class PartitionSpec:
+    """
+    PartitionSpec capture the transformation from table data to partition values
+
+    Attributes:
+        schema(Schema): the schema of data table
+        spec_id(int): any change to PartitionSpec will produce a new specId
+        fields(List[PartitionField): list of partition fields to produce partition values
+        last_assigned_field_id(int): auto-increment partition field id starting from PARTITION_DATA_ID_START
+    """
+
+    PARTITION_DATA_ID_START: int = 1000
+
+    def __init__(self, schema: Schema, spec_id: int, fields: Tuple[PartitionField], last_assigned_field_id: int):
+        self._schema = schema
+        self._spec_id = spec_id
+        self._fields = fields
+        self._last_assigned_field_id = last_assigned_field_id
+        # derived
+        self.fields_by_source_id: Dict[int, List[PartitionField]] = {}
+
+    @property
+    def schema(self) -> Schema:
+        return self._schema
+
+    @property
+    def spec_id(self) -> int:
+        return self._spec_id
+
+    @property
+    def fields(self) -> Tuple[PartitionField]:
+        return self._fields
+
+    @property
+    def last_assigned_field_id(self) -> int:
+        return self._last_assigned_field_id
+
+    def __eq__(self, other):
+        return self.spec_id == other.spec_id and self.fields == other.fields
+
+    def __str__(self):
+        if self.is_unpartitioned():
+            return "[]"
+        else:
+            delimiter = "\n  "
+            partition_fields_in_str = (str(partition_field) for partition_field in self.fields)
+            head = f"[{delimiter}"
+            tail = f"\n]"
+            return f"{head}{delimiter.join(partition_fields_in_str)}{tail}"
+
+    def __repr__(self):
+        return f"PartitionSpec: {str(self)}"
+
+    def __hash__(self):
+        return hash((self.spec_id, self.fields))
+
+    def is_unpartitioned(self) -> bool:
+        return len(self.fields) < 1
+
+    def get_fields_by_source_id(self, filed_id: int) -> List[PartitionField]:
+        if not self.fields_by_source_id:
+            for partition_field in self.fields:
+                source_column = self.schema.find_column_name(partition_field.source_id)
+                if not source_column:
+                    raise ValueError(f"Cannot find source column: {partition_field.source_id}")
+                existing = self.fields_by_source_id.get(partition_field.source_id, [])
+                existing.append(partition_field)
+                self.fields_by_source_id[partition_field.source_id] = existing
+        return self.fields_by_source_id.get(filed_id)  # type: ignore
+
+    def compatible_with(self, other) -> bool:
+        """
+        Returns true if this partition spec is equivalent to the other, with partition field_id ignored.
+        That is, if both specs have the same number of fields, field order, field name, source column ids, and transforms.
+        """
+        if self.__eq__(other):
+            return True
+
+        if len(self.fields) != len(other.fields):
+            return False
+
+        for index in range(len(self.fields)):
+            this_field = self.fields[index]
+            that_field = other.fields[index]
+            if (
+                this_field.source_id != that_field.source_id
+                or this_field.transform != that_field.transform
+                or this_field.name != that_field.name
+            ):
+                return False
+
+        return True
--- a/python/src/iceberg/table/partitioning.py
+++ b/python/src/iceberg/table/partitioning.py
@@ -14,13 +14,16 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from typing import Dict, Iterable, List, Tuple
+from dataclasses import dataclass, field
+from typing import Dict, List, Tuple
 
 from iceberg.schema import Schema
 from iceberg.transforms import Transform
 
 _PARTITION_DATA_ID_START: int = 1000
 
+
+@dataclass(frozen=True)
 class PartitionField:
     """
     PartitionField is a single element with name and unique id,
@@ -33,43 +36,16 @@ class PartitionField:
         name(str): The name of this partition field
     """
 
-    def __init__(self, source_id: int, field_id: int, transform: Transform, name: str):
-        self._source_id = source_id
-        self._field_id = field_id
-        self._transform = transform
-        self._name = name
-
-    @property
-    def source_id(self) -> int:
-        return self._source_id
-
-    @property
-    def field_id(self) -> int:
-        return self._field_id
-
-    @property
-    def name(self) -> str:
-        return self._name
-
-    @property
-    def transform(self) -> Transform:
-        return self._transform
-
-    def __eq__(self, other):
-        return (
-            self.field_id == other.field_id
-            and self.source_id == other.source_id
-            and self.name == other.name
-            and self.transform == other.transform
-        )
+    source_id: int
+    field_id: int
+    transform: Transform
+    name: str
 
     def __str__(self):
         return f"{self.field_id}: {self.name}: {self.transform}({self.source_id})"
 
-    def __repr__(self):
-        return f"PartitionField(field_id={self.field_id}, name={self.name}, transform={repr(self.transform)}, source_id={self.source_id})"
 
-
+@dataclass(eq=False)
 class PartitionSpec:
     """
     PartitionSpec capture the transformation from table data to partition values
@@ -81,34 +57,32 @@ class PartitionSpec:
         last_assigned_field_id(int): auto-increment partition field id starting from PARTITION_DATA_ID_START
     """
 
-    def __init__(self, schema: Schema, spec_id: int, fields: Iterable[PartitionField], last_assigned_field_id: int):
-        self._schema = schema
-        self._spec_id = spec_id
-        self._fields = tuple(fields)
-        self._last_assigned_field_id = last_assigned_field_id
-        # derived
-        self._fields_by_source_id: Dict[int, List[PartitionField]] = {}
+    schema: Schema
+    spec_id: int
+    fields: Tuple[PartitionField]
+    last_assigned_field_id: int
+    source_id_to_fields_map: Dict[int, List[PartitionField]] = field(init=False)
 
-    @property
-    def schema(self) -> Schema:
-        return self._schema
-
-    @property
-    def spec_id(self) -> int:
-        return self._spec_id
-
-    @property
-    def fields(self) -> Tuple[PartitionField, ...]:
-        return self._fields
-
-    @property
-    def last_assigned_field_id(self) -> int:
-        return self._last_assigned_field_id
+    def __post_init__(self):
+        self.source_id_to_fields_map = dict()
+        for partition_field in self.fields:
+            source_column = self.schema.find_column_name(partition_field.source_id)
+            if not source_column:
+                raise ValueError(f"Cannot find source column: {partition_field.source_id}")
+            existing = self.source_id_to_fields_map.get(partition_field.source_id, [])
+            existing.append(partition_field)
+            self.source_id_to_fields_map[partition_field.source_id] = existing
 
     def __eq__(self, other):
+        """
+        Equality check on spec_id and partition fields only
+        """
         return self.spec_id == other.spec_id and self.fields == other.fields
 
     def __str__(self):
+        """
+        PartitionSpec str method highlight the partition field only
+        """
         result_str = "["
         for partition_field in self.fields:
             result_str += f"\n  {str(partition_field)}"
@@ -117,22 +91,11 @@ class PartitionSpec:
         result_str += "]"
         return result_str
 
-    def __repr__(self):
-        return f"PartitionSpec: {str(self)}"
-
     def is_unpartitioned(self) -> bool:
         return len(self.fields) < 1
 
     def fields_by_source_id(self, field_id: int) -> List[PartitionField]:
-        if not self._fields_by_source_id:
-            for partition_field in self.fields:
-                source_column = self.schema.find_column_name(partition_field.source_id)
-                if not source_column:
-                    raise ValueError(f"Cannot find source column: {partition_field.source_id}")
-                existing = self._fields_by_source_id.get(partition_field.source_id, [])
-                existing.append(partition_field)
-                self._fields_by_source_id[partition_field.source_id] = existing
-        return self._fields_by_source_id[field_id]
+        return self.source_id_to_fields_map[field_id]
 
     def compatible_with(self, other: "PartitionSpec") -> bool:
         """
@@ -140,7 +103,8 @@ class PartitionSpec:
         That is, if both specs have the same number of fields, field order, field name, source column ids, and transforms.
         """
         return all(
-            this_field.source_id == that_field.source_id and this_field.transform == that_field.transform and this_field.name == that_field.name
-            for this_field, that_field
-            in zip(self.fields, other.fields)
+            this_field.source_id == that_field.source_id
+            and this_field.transform == that_field.transform
+            and this_field.name == that_field.name
+            for this_field, that_field in zip(self.fields, other.fields)
         )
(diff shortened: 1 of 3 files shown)


python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
 
     def __repr__(self):
         return f"PartitionField(field_id={self.field_id}, name={self.name}, transform={repr(self.transform)}, source_id={self.source_id})"
+
+    def __hash__(self):
+        return hash((self.source_id, self.field_id, self.name, self.transform))
```
rdblue, line 72: Why is `__hash__` needed? Are partition fields used as map keys?

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+    PartitionSpec capture the transformation from table data to partition values
+
+    Attributes:
+        schema(Schema): the schema of data table
+        spec_id(int): any change to PartitionSpec will produce a new specId
+        fields(List[PartitionField): list of partition fields to produce partition values
+        last_assigned_field_id(int): auto-increment partition field id starting from PARTITION_DATA_ID_START
+    """
+
+    PARTITION_DATA_ID_START: int = 1000
```
rdblue, line 86: This isn't used and probably doesn't belong in the class later. Can you make it a top-level constant and prefix it with `_`?

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+
+    def __init__(self, schema: Schema, spec_id: int, fields: Tuple[PartitionField], last_assigned_field_id: int):
```
rdblue, line 88: Should `fields` be `Iterable` rather than being so strict? This could easily convert to a tuple.

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+        return self.spec_id == other.spec_id and self.fields == other.fields
+
+    def __str__(self):
+        if self.is_unpartitioned():
+            return "[]"
+        else:
+            delimiter = "\n  "
+            partition_fields_in_str = (str(partition_field) for partition_field in self.fields)
+            head = f"[{delimiter}"
+            tail = f"\n]"
```
rdblue, line 122: Why not inline these?

python/src/iceberg/table/partitioning.py
dramaticlly: yeah I added back because it's being used in the hash of the PartitionSpec at line 129

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+    def __repr__(self):
+        return f"PartitionSpec: {str(self)}"
+
+    def __hash__(self):
+        return hash((self.spec_id, self.fields))
+
+    def is_unpartitioned(self) -> bool:
+        return len(self.fields) < 1
+
+    def get_fields_by_source_id(self, filed_id: int) -> List[PartitionField]:
```
rdblue, line 134: How about `find_field_by_source`?

rdblue, line 134: Typo: `filed` should be `field`.

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+                existing.append(partition_field)
+                self.fields_by_source_id[partition_field.source_id] = existing
+        return self.fields_by_source_id.get(filed_id)  # type: ignore
+
+    def compatible_with(self, other) -> bool:
+        """
+        Returns true if this partition spec is equivalent to the other, with partition field_id ignored.
+        That is, if both specs have the same number of fields, field order, field name, source column ids, and transforms.
+        """
+        if self.__eq__(other):
```
rdblue, line 150: We usually don't call these methods directly.

python/src/iceberg/table/partitioning.py
dramaticlly: Sure I can rename to `find_field_by_source`

python/src/iceberg/table/partitioning.py
dramaticlly: I will update to `if self == other`

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+    def get_fields_by_source_id(self, filed_id: int) -> List[PartitionField]:
+        if not self.fields_by_source_id:
+            for partition_field in self.fields:
+                source_column = self.schema.find_column_name(partition_field.source_id)
+                if not source_column:
+                    raise ValueError(f"Cannot find source column: {partition_field.source_id}")
+                existing = self.fields_by_source_id.get(partition_field.source_id, [])
+                existing.append(partition_field)
+                self.fields_by_source_id[partition_field.source_id] = existing
+        return self.fields_by_source_id.get(filed_id)  # type: ignore
```
rdblue, line 143: Why is `type: ignore` needed? Can you get type checking working by adding annotations?

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +67,100 @@ def __str__(self):
...
+
+    @property
+    def last_assigned_field_id(self) -> int:
+        return self._last_assigned_field_id
+
+    def __eq__(self, other):
+        return self.spec_id == other.spec_id and self.fields == other.fields
+
+    def __str__(self):
+        if self.is_unpartitioned():
```
rdblue, line 116: Rather than special casing `is_unpartitioned`, I think this should just construct the inner string differently. The Java implementation doesn't use special cases.

python/src/iceberg/table/partitioning.py
dramaticlly: I think this relate to mypy hint when `dict.get(field_id)` will return `Optional[List[PartitionField]]` instead of `List[PartitionField]`

do you prefer `self.fields_by_source_id[field_id]` or adjust the return type hint to be optional?

python/src/iceberg/table/partitioning.py
dramaticlly: I think it's a bit nasty to construct it exactly like what Java implementation does, so the special casing is essentially simplified from https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L299. The unpartitioned differ from rest with 1 less `\n` right before the `]`

so here's my alternative way to construct, not sure how do you like it compare to what I have right now.



```python
    def __str__(self):
        result_str = "["
        for partition_field in self.fields:
            result_str += f"\n  {str(partition_field)}"
        if self.is_unpartitioned():
            result_str += "]"
        else:
            result_str += "\n]"
        return result_str
```

Python `string.join(Iterable)` does not have a way to attach head and tail, so I figured it might be easier to see it this way as in `f"{head}{delimiter.join(partition_fields_in_str)}{tail}"`.

But I can certainly change quickly if above is what you prefer

python/src/iceberg/table/partitioning.py
dramaticlly: Python `string.join(Iterable)` does not have a way to attach head and tail, so I figured it might be easier to see it this way as in `f"{head}{delimiter.join(partition_fields_in_str)}{tail}"`.

But I can certainly change quickly if above is what you prefer

python/src/iceberg/table/partitioning.py
dramaticlly: updated to `self.fields_by_source_id[field_id]` and removed ignore type override

python/src/iceberg/table/partitioning.py
dhruv-pratap: IMO, and this is just me, I feel like this is too much code for just str and repr. I would just take what `dataclass` or `attrs` will give me, even for the special unpartitioned case.

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +69,95 @@ def __str__(self):
...
+
+    def fields_by_source_id(self, field_id: int) -> List[PartitionField]:
+        if not self._fields_by_source_id:
+            for partition_field in self.fields:
+                source_column = self.schema.find_column_name(partition_field.source_id)
+                if not source_column:
+                    raise ValueError(f"Cannot find source column: {partition_field.source_id}")
+                existing = self._fields_by_source_id.get(partition_field.source_id, [])
+                existing.append(partition_field)
+                self._fields_by_source_id[partition_field.source_id] = existing
```
dhruv-pratap, line 142: I feel this field value should be derived in `__init__` , or `__post_init__` if you are using `@dataclass` or `@attrs`. Reason being it validates the correctness of the object state, and does should raise `ValueError` as soon as it is created. This seems too late to do raise that error.

dhruv-pratap (changes requested):

python/src/iceberg/table/partitioning.py
rdblue: > I think it's a bit nasty to construct it exactly like what Java implementation does

I don't think it is necessary to construct it exactly like Java. But it should not have a special case here, since `is_unpartitioned` can include specs with `void` transforms. This should show all of the transforms and not special case.

python/src/iceberg/table/partitioning.py
rdblue: > because it's being used in the hash of the PartitionSpec at line 129

Why does `PartitionSpec` define `__hash__`? Will spec be used as a key in a dict or map? I doubt it so I'd probably remove both.

python/src/iceberg/table/partitioning.py
dramaticlly: >I don't think it is necessary to construct it exactly like Java. 

got it, thanks for clarify

> is_unpartitioned can include specs with void transforms

But I am a bit behind on this. From what I can tell, the void transform can never be unpartitioned per existing java code https://github.com/apache/iceberg/blob/32841a273b6dda897f6efece4ebb178f8ad54719/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L97-L103. And if there's one void transform added to such PartitionSpec, then I assume there's one partition field with voidTransform as its Transform, which is not `unpartitioned` like below 

```
> PartitionSpec ps = PartitionSpec.builderFor(SCHEMA).alwaysNull("ts").build();
> System.out.println(ps.toString());
[
  1000: ts_null: void(5)
]

> System.out.println(ps.isUnpartitioned());
false
```

python/src/iceberg/table/partitioning.py
```diff
@@ -14,8 +14,13 @@
...
 # under the License.
+from typing import Dict, Iterable, List, Tuple
+
+from iceberg.schema import Schema
 from iceberg.transforms import Transform
 
+_PARTITION_DATA_ID_START: int = 1000
+
 
 class PartitionField:
```
Fokko, line 25: Instead of implementing the `__eq__` and `__hash__` we could leverage the dataclass library. If we set `eq=True` and `frozen=True` (which makes it immutable, which is also nice), then we get __hash__ automatically:

If `eq` and `frozen` are both true, by default [dataclass()](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) will generate a `__hash__()` method for you. If eq is true and frozen is false, `__hash__()` will be set to `None`, marking it unhashable (which it is, since it is mutable). If `eq` is false, `__hash__()` will be left untouched meaning the `__hash__()` method of the superclass will be used (if the superclass is [object](https://docs.python.org/3/library/functions.html#object), this means it will fall back to id-based hashing).

More information here: https://docs.python.org/3/library/dataclasses.html

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +69,95 @@ def __str__(self):
...
+    PartitionSpec capture the transformation from table data to partition values
+
+    Attributes:
+        schema(Schema): the schema of data table
+        spec_id(int): any change to PartitionSpec will produce a new specId
+        fields(List[PartitionField): list of partition fields to produce partition values
+        last_assigned_field_id(int): auto-increment partition field id starting from PARTITION_DATA_ID_START
+    """
+
+    def __init__(self, schema: Schema, spec_id: int, fields: Iterable[PartitionField], last_assigned_field_id: int):
```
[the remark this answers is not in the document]
dramaticlly: hey @dhruv-pratap , I specifically mentioned that this is second part of rolling out partitioning, we will have a follow up PR for PartitionSpec builder as it shall be the right way to construct the PartitionSpec. I leave it outside scope of this PR to it doesn't get too large

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +69,95 @@ def __str__(self):
...
+            for partition_field in self.fields:
+                source_column = self.schema.find_column_name(partition_field.source_id)
+                if not source_column:
+                    raise ValueError(f"Cannot find source column: {partition_field.source_id}")
+                existing = self._fields_by_source_id.get(partition_field.source_id, [])
+                existing.append(partition_field)
+                self._fields_by_source_id[partition_field.source_id] = existing
+        return self._fields_by_source_id[field_id]
+
+    def compatible_with(self, other) -> bool:
```
Fokko, line 138:
```suggestion
    def compatible_with(self, other: "PartitionSpec") -> bool:
```

python/src/iceberg/table/partitioning.py
```diff
@@ -64,3 +69,95 @@ def __str__(self):
...
+        """
+        Returns true if this partition spec is equivalent to the other, with partition field_id ignored.
+        That is, if both specs have the same number of fields, field order, field name, source column ids, and transforms.
+        """
+        if self == other:
+            return True
+        if len(self.fields) != len(other.fields):
+            return False
+        for index in range(len(self.fields)):
```
Fokko, line 147: I would write this as:
```python
return all(
    this_field.source_id == that_field.source_id and this_field.transform == that_field.transform and this_field.name == that_field.name
    for this_field, that_field
    in zip(self.fields], other.fields)
)
```
Or rely on the `eq` method :)

Fokko (commented):
Thanks for working on this @dramaticlly 👍🏻  A few comments

python/src/iceberg/table/partitioning.py
dramaticlly: Thanks @Fokko , I think your suggestion looks clean, but want to make sure we dont need `]` after self.fields, so I guess it's probably typo and expect 

```python
in zip(self.fields, other.fields)
```

On the other side, I think this method differs from `eq` just because we can relax on `field_id` for equality, everything else is the same

python/src/iceberg/table/partitioning.py
dramaticlly: yeah I agree the PartitionField is an immutable class after construction so dataclass with both eq and frozen sounds fair to me. 

On the other side, I think the biggest benefit of the dataclass is the `__post_init__` method which allow for java-like builderPattern equivalent processing when we build the PartitionSpec. There's collection of validations need to happen and I am discussing with @samredai  in https://github.com/apache/iceberg/issues/4631#issuecomment-1113632408. 

From what I can tell, we will need a `PartitionSpecBuilder` class with convenient way to construct the PartitionSpec, but we also want to make sure avoid duplicate the builder logic in an overly complex init method for PartitionSpec

samredai: @dramaticlly that's an interesting idea I haven't thought of. A big argument for using the builder pattern was that we wanted `PartitionSpec` to be immutable, which would require us to include a ton of validation logic (everything that would be in a builder) in the `__init__` method. If I understand your suggestion, using `__post_init__` would allow us to have a typical init method, but then include the builder-type validation logic in the `__post_init__` which would fail the initialization of any invalid `PartitionSpec`.

rdblue: Is the plan to use `@dataclass`? I like that idea, but I won't hold off on reviewing if we want to get it in like this first.

dramaticlly:
Updated the PR with what dataclass can offer for PartitionField and PartitionSpec. The scope of this change stay the same and need a following PR to add convenient Builder for PartitionSpec

Appreciate the review, if you can take a look that would be great! @rdblue @samredai @Fokko

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,76 @@ class PartitionField:
...
+        result_str = "["
+        for partition_field in self.fields:
+            result_str += f"\n  {str(partition_field)}"
+        if len(self.fields) > 0:
+            result_str += "\n"
+        result_str += "]"
+        return result_str
 
-    def __repr__(self):
-        return f"PartitionField(field_id={self.field_id}, name={self.name}, transform={repr(self.transform)}, source_id={self.source_id})"
```
dramaticlly, line 66: FYI using default`__repr__` from dataclass now also include the derived attribute such as `source_id_to_fields_map` into consideration, please let me know if anyone feels strongly to override and keep the original behaviour

python/src/iceberg/table/partitioning.py
samredai: I really like this idea of wrapping the classes generated from the OpenAPI spec. The class naming here might be tricky. Module namespacing allows us to re-use the same name if we want, something like:
```py
from iceberg.openapi import rest_catalog

class PartitionField(rest_catalog.PartitionField):
    ...
```
We shouldn't expect users to import from the openapi module directly so we shouldn't need to worry about naming conflicts, right? Maybe we should name it `_openapi` just to be super clear about that.

kbendick: I'll hold off on commenting too much until I have a chance to look into the `pydantic` project as well as look at the other PR.

My first ask would be how many dependencies are we bringing in if we add `pydantic`? I know that some folks were concerned about adding very many external python dependencies, so as not to conflict with their own dependencies, but if the benefit is very large I'm not personally opposed to it (I believe it was somebody / some group from Netflix that originally requested we keep the number of required dependencies down).

But validation, either via a library or via a common pattern we settle on, is something that would be very beneficial.

samredai: Great point, I just checked the [dependencies](https://github.com/samuelcolvin/pydantic/blob/master/setup.py#L132) for pydantic and the good news is that all it requires is `typing-extensions` which is probably just for some python 3.7 backports. That will even probably get dropped at some point when they no longer support 3.7

python/src/iceberg/table/partitioning.py
Fokko: You can easily restore the original behavior by changing:
```python
    source_id_to_fields_map: Dict[int, List[PartitionField]] = field(init=False)
```
into:
```
    source_id_to_fields_map: Dict[int, List[PartitionField]] = field(init=False, repr=False)
```

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,76 @@ class PartitionField:
...
-        )
+        """
+        Equality check on spec_id and partition fields only
+        """
+        return self.spec_id == other.spec_id and self.fields == other.fields
 
     def __str__(self):
-        return f"{self.field_id}: {self.name}: {self.transform}({self.source_id})"
+        """
+        PartitionSpec str method highlight the partition field only
```
rdblue, line 85: This isn't really a docstring. It is more of a comment to explain the logic in the method. If you want to add it as a docstring, then I think it should be real docs (like "Produce a human-readable string representation of PartitionSpec") and have this as an additional note.

python/tests/table/test_partitioning.py
```diff
@@ -29,8 +30,37 @@ def test_partition_field_init():
     assert partition_field.transform == bucket_transform
     assert partition_field.name == "id"
     assert partition_field == partition_field
+    print(str(partition_field))
+    print("repr")
+    print(repr(partition_field))
```
rdblue, line 35: Can you remove all print statements from the tests?

python/src/iceberg/table/partitioning.py
dramaticlly: Thanks @rdblue , updated per your suggestion

python/tests/table/test_partitioning.py
dramaticlly: Thank you, I was testing and forgot about this, removed

python/src/iceberg/table/partitioning.py
Fokko: I would not consider this blocking for now.

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,79 @@ class PartitionField:
...
-        return (
-            self.field_id == other.field_id
-            and self.source_id == other.source_id
-            and self.name == other.name
-            and self.transform == other.transform
-        )
+        """
+        Equality check on spec_id and partition fields only
+        """
+        return self.spec_id == other.spec_id and self.fields == other.fields
```
Fokko, line 81: Nit: It would be good to check if the other is `is not None`, maybe also `isinstanceof(other, PartitionSpec)` otherwise the `other.spec_id` will/could throw an exception

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,79 @@ class PartitionField:
...
+    PartitionSpec capture the transformation from table data to partition values
 
-    @property
-    def name(self) -> str:
-        return self._name
+    Attributes:
+        schema(Schema): the schema of data table
+        spec_id(int): any change to PartitionSpec will produce a new specId
+        fields(List[PartitionField): list of partition fields to produce partition values
+        last_assigned_field_id(int): auto-increment partition field id starting from PARTITION_DATA_ID_START
```
Fokko, line 57: Would be great if we can also add an example: https://github.com/apache/iceberg/blob/master/python/src/iceberg/types.py#L83-L87 this will also test the __str__ method as the examples are executed as tests as well 👍🏻

Fokko (approved):
Some small comments, but apart from that it looks good to me.

Note that it is still a stub, for example `last_assigned_field_id` isn't being used (yet).

python/src/iceberg/table/partitioning.py
dramaticlly: Thanks Fokko, I think this is where I want to leave out the example as I intended to construct the PartitionSpec via [dedicated builder](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L337) (which is not included in this PR), as that's the desired way to construct the PartitionSpec with convenient transform helper method and equipped with validation. I can include the example here, what do you think?

dramaticlly:
Thanks @Fokko for reviewing, appreciate one more pass on this @rdblue and @samredai

python/src/iceberg/table/partitioning.py
samredai: I think what we want to do here if `other` is not an instance of `PartitionSpec` is just return False. So that `PartitionSpec(...) == "foo"` returns `False` instead of raising an error. Same thing for cases where `other` is `None`.

dramaticlly: Thank you @samredai , updated

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,81 @@ class PartitionField:
...
+        return f"{self.field_id}: {self.name}: {self.transform}({self.source_id})"
+
 
-    @property
-    def source_id(self) -> int:
-        return self._source_id
+@dataclass(eq=False, frozen=True)
+class PartitionSpec:
+    """
+    PartitionSpec capture the transformation from table data to partition values
```
samredai, line 51: nit: `s/capture/captures`

samredai (approved):
This LGTM, thanks @dramaticlly!

python/src/iceberg/table/partitioning.py
dramaticlly: Thanks, fixed it

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,81 @@ class PartitionField:
...
+        result_str += "]"
+        return result_str
+
+    def is_unpartitioned(self) -> bool:
+        return len(self.fields) < 1
+
+    def fields_by_source_id(self, field_id: int) -> List[PartitionField]:
+        return self.source_id_to_fields_map[field_id]
+
+    def compatible_with(self, other: "PartitionSpec") -> bool:
```
Fokko, line 107: We should also assert the number of fields. The following test is passing:
```python
def test_partition_compatible_with(table_schema_simple: Schema):
    bucket_transform = bucket(IntegerType(), 4)
    field1 = PartitionField(3, 100, bucket_transform, "id")
    field2 = PartitionField(3, 102, bucket_transform, "id")
    lhs = PartitionSpec(table_schema_simple, 0, (field1,), 1001)
    rhs = PartitionSpec(table_schema_simple, 0, (field1, field2), 1001)
    assert lhs.compatible_with(rhs)
```
image

We also do this on the Java side: https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L211

The Python zip will only zip the number of fields that are in the shortest list.

python/src/iceberg/table/partitioning.py
```diff
@@ -29,38 +36,81 @@ class PartitionField:
...
+            Only include list of partition fields in the PartitionSpec's string representation
+        """
+        result_str = "["
+        if self.fields:
+            result_str += "\n  " + "\n  ".join([str(field) for field in self.fields]) + "\n"
+        result_str += "]"
+        return result_str
+
+    def is_unpartitioned(self) -> bool:
+        return len(self.fields) < 1
```
Fokko, line 102: supernit: We could shorten this to:
```suggestion
        return not self.fields
```

Fokko (approved):
I think there is a bug in the `compatible_with`, apart from that it looks good!

python/src/iceberg/table/partitioning.py
dramaticlly: Thanks @Fokko , updated

python/src/iceberg/table/partitioning.py
dramaticlly: Thanks for catching it for me, I forgot about this when move from loop into zip. Added eq and length check back. Your proposed unit test also looks helpful, so I added in the latest commit as well!

dramaticlly:
Thank you Fokko and Sam for reviewing, @rdblue can you take another look?

rdblue (approved):

rdblue:
Looks great. Thanks, @dramaticlly! I merged this.

Issue #4718, opened 2022-05-07
Iceberg 0.13 with Spark 3.2 - list partitions query always need partition.date and partition.hour columns in the result


abmo-x:
Steps to reproduce:

1) create a Iceberg table using Hive metastore with date, hour partition columns
2) Insert few records
3) try to list partitions for the table created above. 

Select queries only succeed if date AND hour columns are used. If either column is skipped or both are skipped then query fails.

This is not reproducible with spark 3.0 and Iceberg 0.11

see queries below:

```
**spark-sql>** select partition from spark_catalog.monitoring.test.partitions;
{"date":"2020-01-02","hour":"10"}
{"date":"2020-01-02","hour":"00"}
{"date":"2020-01-01","hour":"00"}
{"date":"2020-01-01","hour":"10"}
Time taken: 0.191 seconds, Fetched 4 row(s)


**spark-sql>** select partition.date, partition.hour from spark_catalog.monitoring.test.partitions;
2020-01-02	10
2020-01-02	00
2020-01-01	00
2020-01-01	10
Time taken: 0.198 seconds, Fetched 4 row(s)

**spark-sql>** select file_count from spark_catalog.monitoring.test.partitions; 
22/05/06 16:12:25 ERROR SparkSQLDriver: Failed in [select file_count from spark_catalog.monitoring.test.partitions]
java.lang.IllegalArgumentException: Cannot find source column: partition.date
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:217) ~[ ]
	at org.apache.iceberg.PartitionSpec$Builder.findSourceColumn(PartitionSpec.java:374) ~[ ]
	at org.apache.iceberg.PartitionSpec$Builder.identity(PartitionSpec.java:379) ~[ ]
	at org.apache.iceberg.BaseMetadataTable.lambda$transformSpec$0(BaseMetadataTable.java:68) ~[ ]
	at org.apache.iceberg.relocated.com.google.common.collect.ImmutableList.forEach(ImmutableList.java:405) ~[ ]
	at org.apache.iceberg.BaseMetadataTable.transformSpec(BaseMetadataTable.java:68) ~[ ]
	at org.apache.iceberg.PartitionsTable.planFiles(PartitionsTable.java:114) ~[ ]
	at org.apache.iceberg.PartitionsTable.partitions(PartitionsTable.java:97) ~[ ]
[...]

PR #4720, opened 2022-05-07
Core: Fix query failure when using projection on top of partitions metadata table

singhpk234:
Closes https://github.com/apache/iceberg/issues/4718

As per my understanding, we should use scan.tableSchema() rather than scan.schema() for transformSpec when planFiles() for partitions metaData table. 

Hence we faced :
> 22/05/06 16:12:25 ERROR SparkSQLDriver: Failed in [select file_count from spark_catalog.monitoring.test.partitions]
java.lang.IllegalArgumentException: Cannot find source column: partition.date

as present in reported ticket

Though in master we will get : 

>Cannot find source column: 1000
java.lang.NullPointerException: Cannot find source column: 1000

---

Testing Done : 
Added UT fails without the change

cc @szehon-ho @RussellSpitzer

### DIFF
--- a/core/src/main/java/org/apache/iceberg/PartitionsTable.java
+++ b/core/src/main/java/org/apache/iceberg/PartitionsTable.java
@@ -152,7 +152,7 @@ public class PartitionsTable extends BaseMetadataTable {
 
     LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
       PartitionSpec spec = table.specs().get(specId);
-      PartitionSpec transformedSpec = transformSpec(scan.schema(), spec);
+      PartitionSpec transformedSpec = transformSpec(scan.tableSchema(), spec);
       return ManifestEvaluator.forRowFilter(scan.filter(), transformedSpec, caseSensitive);
     });
 
--- a/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMetadataTables.java
+++ b/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMetadataTables.java
@@ -171,6 +171,13 @@ public class TestMetadataTables extends SparkExtensionsTestBase {
     Assert.assertEquals("Metadata table should return one data file", 1, actualDataFiles.size());
     TestHelpers.assertEqualsSafe(filesTableSchema.asStruct(), expectedDataFiles.get(0), actualDataFiles.get(0));
 
+    List<Row> actualPartitionsWithProjection =
+        spark.sql("SELECT file_count FROM " + tableName + ".partitions ").collectAsList();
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/iceberg/PartitionsTable.java
```diff
@@ -152,7 +152,7 @@ static CloseableIterable<FileScanTask> planFiles(StaticTableScan scan) {
 
     LoadingCache<Integer, ManifestEvaluator> evalCache = Caffeine.newBuilder().build(specId -> {
       PartitionSpec spec = table.specs().get(specId);
-      PartitionSpec transformedSpec = transformSpec(scan.schema(), spec);
+      PartitionSpec transformedSpec = transformSpec(scan.tableSchema(), spec);
```
RussellSpitzer, line 155: So in this case "Schema" is the projected schema which is missing all the partition columns and "tableSchema" has the full partitionSpec. LGTM

RussellSpitzer (approved):

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMetadataTables.java
```diff
@@ -171,6 +171,13 @@ public void testPartitionedTable() throws Exception {
     Assert.assertEquals("Metadata table should return one data file", 1, actualDataFiles.size());
     TestHelpers.assertEqualsSafe(filesTableSchema.asStruct(), expectedDataFiles.get(0), actualDataFiles.get(0));
 
+    List<Row> actualPartitionsWithProjection =
```
RussellSpitzer, line 174: While I like this test I believe we should also add one to https://github.com/apache/iceberg/blob/4ae2002bd46bf8e1c20db03cffc6319237e4d74a/core/src/test/java/org/apache/iceberg/TestMetadataTableScans.java

To make sure we have one in the core module.

RussellSpitzer: @singhpk234  ^

singhpk234: +1 added a UT for the same, thanks @RussellSpitzer !!!

abmo-x:
Thanks @singhpk234 for a quick turn around, appreciate it!

abmo-x (approved):

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMetadataTables.java
RussellSpitzer: Thanks! Otherwise we wouldn't have an automated test running on this with "Core" only changes.

szehon-ho:
Arg, sorry about the initial bug, thanks @singhpk234 for the fix and all for review

PR #4721, opened 2022-05-07
API: Introduce endWiths Predicate.

southernriver:
Some users have requested prefix matching or endsWith. And Push-Down for Spark/Flink is follow-up.
cc @rdblue @aokolnychyi @sujithjay @kbendick

### DIFF
--- a/api/src/main/java/org/apache/iceberg/transforms/Truncate.java
+++ b/api/src/main/java/org/apache/iceberg/transforms/Truncate.java
@@ -298,6 +298,7 @@ abstract class Truncate<T> implements Transform<T, T> {
             return ProjectionUtil.truncateArray(name, pred, this);
 
           case NOT_STARTS_WITH:
+          case NOT_ENDS_WITH:
             if (pred.literal().value().length() < width()) {
               return Expressions.predicate(pred.op(), name, pred.literal().value());
             } else if (pred.literal().value().length() == width()) {
@@ -338,6 +339,7 @@ abstract class Truncate<T> implements Transform<T, T> {
             return null;
 
           case NOT_STARTS_WITH:
+          case NOT_ENDS_WITH:
             if (pred.literal().value().length() < width()) {
               return Expressions.predicate(pred.op(), name, pred.literal().value());
             } else if (pred.literal().value().length() == width()) {
--- a/api/src/main/java/org/apache/iceberg/transforms/Truncate.java
+++ b/api/src/main/java/org/apache/iceberg/transforms/Truncate.java
@@ -288,6 +288,7 @@ abstract class Truncate<T> implements Transform<T, T> {
         BoundLiteralPredicate<CharSequence> pred = predicate.asLiteralPredicate();
         switch (pred.op()) {
           case STARTS_WITH:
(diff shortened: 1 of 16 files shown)


api/src/main/java/org/apache/iceberg/transforms/Truncate.java
```diff
@@ -288,6 +288,7 @@ public UnboundPredicate<CharSequence> project(String name,
         BoundLiteralPredicate<CharSequence> pred = predicate.asLiteralPredicate();
         switch (pred.op()) {
           case STARTS_WITH:
+          case ENDS_WITH:
```
kbendick, line 291: I'm not sure that the `STARTS_WITH` truncate logic can be reused entirely.

The `ProjectionUtil.truncateArray` initially seems suspect to me. As we're concerned with the end of the string, truncating seems like not the best option.

For the cases where `width` is less than or equal to the value's length, these two should be ok.

The safest thing to do (which wouldn't be optimal in terms of filter pushdown) would be simply to return `null` instead of truncating the predicate, as we are concerned with the end of the string, and truncation will have removed that (making it not comparable for certain levels of file pruning).

kbendick (commented):
This is a great start @southernriver.

I only took an initial pass over this, so I'll need to look further before I can comment on some of the details. This is getting deep into the core library usage.

I'm not sure if we can use `ENDS_WITH` for truncated values unless the truncation width is longer than the value being compared to (as mentioned in my comment).

But this will still allow for more push down in many situations that aren't covered presently at all.

I wrote a lot more tests than were eventually included for the `NOT_STARTS_WITH` work. I can try to find them for you as a reference guide if you'd like. I'd consider implementing the `ENDS_WITH` logic before the `NOT_ENDS_WITH` logic, as the double negation presents some... mental overhead. And normally we deal with the starts of the strings (e.g. for truncation).

I'm not sure how much of this work you may or may not have already been using in a fork somewhere.

But given the existence of `StringEndsWith` in Spark, this is something that could be made use of in at least some situations.

I'll be following along and will do a deeper dive next week, but please feel free to reach out if you have any questions or need any guidance or anything. 👍

rdblue:
@southernriver, when can Iceberg metadata be used to evaluate `endsWith`? The only case I can think of is when evaluating for identity partitioned data values, but that's an extremely narrow use case. I don't think it is worth the complexity of adding more predicates unless Iceberg metadata can be used to skip more data.

kbendick:
> @southernriver, when can Iceberg metadata be used to evaluate `endsWith`? The only case I can think of is when evaluating for identity partitioned data values, but that's an extremely narrow use case. I don't think it is worth the complexity of adding more predicates unless Iceberg metadata can be used to skip more data.

Yeah I do agree. It would have to be identity partitions and the metrics would need to be stored longer than the standard truncation width. I pulled this down and played with it and it’s a very narrow use case.

There are cases where it can have benefit, but we should weigh the complexity of the implementation over the value added. The use cases that will see a performance benefit from this are very narrow, but as it stands now I think  the added complexity could lead to incorrect results potentially. Like the truncation right now etc.

PR #4725, opened 2022-05-09
Docs: add types supported by truncate

wuwenchi:
In fact, truncate also supports the binary type:

```java
abstract class Truncate<T> implements Transform<T, T> {
  @SuppressWarnings("unchecked")
  static <T> Truncate<T> get(Type type, int width) {
    Preconditions.checkArgument(width > 0,
        "Invalid truncate width: %s (must be > 0)", width);

    switch (type.typeId()) {
      case INTEGER:
        return (Truncate<T>) new TruncateInteger(width);
      case LONG:
        return (Truncate<T>) new TruncateLong(width);
      case DECIMAL:
        return (Truncate<T>) new TruncateDecimal(width);
      case STRING:
        return (Truncate<T>) new TruncateString(width);
      case BINARY:
        return (Truncate<T>) new TruncateByteBuffer(width);
      default:
            "Cannot truncate type: " + type);
    }
  }
```

so, add `binary` type, from:
```
| truncate[W] | Value truncated to width `W` (see below)  | `int`, `long`, `decimal`, `string`                              
```
to
```
| truncate[W] | Value truncated to width `W` (see below) | `int`, `long`, `decimal`, `string`, `binary`                             
```


singhpk234 (approved):

wuwenchi:
Can you help check it ? Thx! @kbendick

PR #4728, opened 2022-05-09
Python: Remove mypy error

Fokko:
Mypy complaints about the config file:

```
No [mypy] section in config file
```

Also enabled some additional checks on unreachable code

### DIFF
--- a/python/src/iceberg/transforms.py
+++ b/python/src/iceberg/transforms.py
@@ -64,10 +64,10 @@ class Transform(ABC, Generic[S, T]):
     def __str__(self):
         return self._transform_string
 
-    def __call__(self, value: S) -> Optional[T]:
+    def __call__(self, value: Optional[S]) -> Optional[T]:
         return self.apply(value)
 
-    def apply(self, value: S) -> Optional[T]:
+    def apply(self, value: Optional[S]) -> Optional[T]:
         ...
 
     def can_transform(self, source: IcebergType) -> bool:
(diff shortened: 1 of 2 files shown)


python/src/iceberg/transforms.py
```diff
@@ -120,9 +120,6 @@ def hash(self, value: S) -> int:
         raise NotImplementedError()
 
     def apply(self, value: S) -> Optional[int]:
-        if value is None:
```
Fokko, line 123: Since we have `S`, it can never be `None`, otherwise, the signature should be `Optional[S]`. If we want, we can still check if it `None`, `""` or `[]` using:
```python
if not value:
```

Fokko: Okay, looking at the tests, we _might_ expect a `None` there. I've changed the code the other way around.

python/src/iceberg/transforms.py
```diff
@@ -83,10 +83,8 @@ def preserves_order(self) -> bool:
     def satisfies_order_of(self, other) -> bool:
         return self == other
 
-    def to_human_string(self, value) -> str:
-        if value is None:
-            return "null"
-        return str(value)
+    def to_human_string(self, value: Optional[S]) -> str:
+        return str(value) if value is not None else "null"
```
rdblue, line 87: Is this change necessary for this PR? We generally try to avoid code churn that isn't making functional changes.

python/src/iceberg/transforms.py
rdblue: Yeah, transforms can be applied to optional columns so this should be `Optional[S]`.

rdblue (approved):

rdblue:
Thanks, @Fokko!

PR #4730, opened 2022-05-09
Python: Add spellcheck to the CI

Fokko:
Simple spellcheck so we don't have to spend any mental energy on the spelling


python/src/iceberg/table/partitioning.py
```diff
@@ -24,7 +24,7 @@ class PartitionField:
 
     Attributes:
         source_id(int): The source column id of table's schema
-        field_id(int): The partition field id across all the table metadata's partition specs
+        field_id(int): The partition field id across all the table partition metadata specs
```
dramaticlly, line 27: nit: I think [partition spec](https://iceberg.apache.org/spec/#partition-specs) here refer to a single entity in iceberg jargon, so I feel `partition metadata specs` does not confer the same meaning

python/spellcheck-dictionary.txt
```diff
@@ -0,0 +1,33 @@
+accessor
```
dramaticlly, line 1: just curious, do we need to manually maintain the private dictionary going forward?

dramaticlly: also I am curious, do we need both `python/spellcheck-dictionary.txt` and top level `spellcheck-dictionary.txt`?

python/src/iceberg/table/partitioning.py
Fokko: I've updated it to `The partition field id across all the table partition specs metadata` WDYT?

python/spellcheck-dictionary.txt
Fokko: > just curious, do we need to manually maintain the private dictionary going forward?

Yes, once there is special jargon that's outside of the English dictionary, we need to add this.

> also I am curious, do we need both python/spellcheck-dictionary.txt and top level spellcheck-dictionary.txt?

Yes, in that case, we would have another dictionary, but a better option would be to move it to the top level. Or in `dev/` to keep the top-level nice and clean.

dev/.rat-excludes
```diff
@@ -25,4 +25,5 @@ package-list
 sitemap.xml
 derby.log
 .python-version
+spellcheck-dictionary.txt
```
rdblue, line 28: There's no way to add comments to this file?

python/src/iceberg/table/partitioning.py
```diff
@@ -24,7 +24,7 @@ class PartitionField:
 
     Attributes:
         source_id(int): The source column id of table's schema
-        field_id(int): The partition field id across all the table metadata's partition specs
+        field_id(int): The partition field id across all the table partition specs' metadata
```
rdblue, line 27: This changes the meaning. "Table metadata" refers to all of the metadata, and "table metadata's partition specs" is the set of partition specs in table metadata. "Partition specs' metadata" is narrower, the metadata across partition specs.

To fix this, I think you can just omit "metadata": "across all the table's partition specs".

python/tests/test_schema.py
```diff
@@ -228,7 +228,7 @@ def test_schema_find_field_by_id_raise_on_unknown_field(table_schema_simple):
 
 
 def test_schema_find_field_type_by_id(table_schema_simple):
-    """Test retrieving a columns's type using its field ID"""
+    """Test retrieving a column type using its field ID"""
```
rdblue, line 231: "column's"? I think the possessive form should still be used.

python/tox.ini
```diff
@@ -52,10 +52,13 @@ deps =
     black
     isort
     autoflake
+    pylint
+    pyenchant
 commands =
     autoflake -r --check --ignore-init-module-imports --remove-all-unused-imports src tests
     isort --profile black --check-only src tests
     black --line-length 130 --check --diff src tests
+    pylint --disable all --enable spelling --spelling-dict en_US --spelling-private-dict-file spellcheck-dictionary.txt en_US src tests
```
kbendick, line 61: Question / Nit: Do we need to specify `en_US` twice? Specifically for the 2nd argument after `--spelling-private-dict-file`.

Fokko: Great catch @kbendick! That's just me being messy with copy and paste!

python/tests/test_schema.py
Fokko: Updated!

python/src/iceberg/table/partitioning.py
Fokko: Love it, less is more 👍🏻

rdblue (approved):

rdblue:
Thanks, @Fokko!

PR #4733, opened 2022-05-09
Core: remove reference files for BaseEqualityDeltaWriter

chenjunjiedada:
Flink upsert commit doesn't validate the conflict of data files that position deletes referred, thus it doesn't have to keep referred data files of position deletes in memory anymore.


rdblue:
I don't think this is a good idea. While Flink may not need the referenced files set, the writer that you're modifying here lives in core and could be used by other modules. If this were entirely in Flink I think it would be fine, but it probably isn't a good idea to modify this class.

chenjunjiedada:
Sounds reasonable to me. Let me close this.

PR #4573, opened 2022-04-16
ORC: Upgrade to 1.7.4

williamhyun:

PR #4734, opened 2022-05-09
ORC: Upgrade orc version to use the built-in estimate memory method

hililiwei:

Because ORC has been updated, we can call method `estimateMemory` directly instead of reflection.

## Brief change log

1. Use  memory evaluation methods in `Writer`

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcFileAppender.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcFileAppender.java
@@ -31,7 +31,6 @@ import org.apache.hadoop.fs.Path;
 import org.apache.iceberg.Metrics;
 import org.apache.iceberg.MetricsConfig;
 import org.apache.iceberg.Schema;
-import org.apache.iceberg.common.DynFields;
 import org.apache.iceberg.exceptions.RuntimeIOException;
 import org.apache.iceberg.hadoop.HadoopOutputFile;
 import org.apache.iceberg.io.FileAppender;
@@ -43,7 +42,6 @@ import org.apache.orc.Reader;
 import org.apache.orc.StripeInformation;
 import org.apache.orc.TypeDescription;
 import org.apache.orc.Writer;
-import org.apache.orc.impl.writer.TreeWriter;
 import org.apache.orc.storage.ql.exec.vector.VectorizedRowBatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -57,7 +55,6 @@ class OrcFileAppender<D> implements FileAppender<D> {
   private final int batchSize;
   private final OutputFile file;
   private final Writer writer;
-  private final TreeWriter treeWriter;
   private final VectorizedRowBatch batch;
   private final int avgRowByteSize;
   private final OrcRowWriter<D> valueWriter;
@@ -92,8 +89,6 @@ class OrcFileAppender<D> implements FileAppender<D> {
     options.setSchema(orcSchema);
     this.writer = newOrcWriter(file, options, metadata);
 
-    // TODO: Turn to access the estimateMemorySize directly after ORC 1.7.4 released with https://github.com/apache/orc/pull/1057.
-    this.treeWriter =  treeWriterHiddenInORC();
     this.valueWriter = newOrcRowWriter(schema, orcSchema, createWriterFunc);
   }
 
@@ -123,10 +118,7 @@ class OrcFileAppender<D> implements FileAppender<D> {
       return file.toInputFile().getLength();
     }
 
-    Preconditions.checkNotNull(treeWriter,
-        "Cannot estimate length of file being written as the ORC writer's internal writer is not present");
-
-    long estimateMemory = treeWriter.estimateMemory();
+    long estimateMemory = writer.estimateMemory();
 
     long dataLength = 0;
     try {
@@ -192,11 +184,5 @@ class OrcFileAppender<D> implements FileAppender<D> {
                                                      BiFunction<Schema, TypeDescription, OrcRowWriter<?>>
                                                          createWriterFunc) {
     return (OrcRowWriter<D>) createWriterFunc.apply(schema, orcSchema);
-  }
-
-  private TreeWriter treeWriterHiddenInORC() {
-    DynFields.BoundField<TreeWriter> treeWriterFiled =
-        DynFields.builder().hiddenImpl(writer.getClass(), "treeWriter").build(writer);
-    return treeWriterFiled.get();
   }
 }


data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java
```diff
@@ -297,7 +297,9 @@ public void testNoNulls() {
   }
 
   @Test
-  public void testIsNaN() {
+  public void testIsNaNForParquet() {
```
kbendick, line 300: Nit: You don't need to change the test name, because the `Assume` will take care of it and signal that in the tests.


For people that maintain forks and cherry-pick individual changes, updating the test name will probably cause them more of headache than is worth it.

Also, I would consider using `Assume.assumeTrue("Parquet metrics do not track NaN", format == FileFormat.PARQUET)` unless this should also possibly run someday for Avro.

data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java
```diff
@@ -311,6 +313,14 @@ public void testIsNaN() {
     Assert.assertFalse("Should skip: all null column will not contain nan value", shouldRead);
   }
 
+  @Test
+  public void testIsNaNForORC() {
+    Assume.assumeFalse("Not for Parquet files", format == FileFormat.PARQUET);
```
kbendick, line 318: Same note about the `Assume` statement.

I would just put `Assume.assumeTrue(format == FileFormat.ORC)` unless you think that Avro should also be included here.

Many of the `Assume` statements don't seem to give a reason phrase at all.

data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java
kbendick, line 300: EDIT - I see now why the name was changed.

My suggestion, to avoid modifying the name / making two tests that are named specifically for their file format, that you consider making two separate functions and  calling out to each situationally.

Like
```java
public void testIsNaN() {
  Assume.assumeTrue("Avro files do not have row group statistics", format != FileFormat.AVRO);
  
  if (format == FileFormat.ORC) {
    testIsNanForOrc();
  } else {
    testIsNanForParquet();
  }
}

private void testIsNanForOrc() { .... }

private void testIsNanForParquet() { ... }
```

This way, the same test (by name) will be run for each, but the correct assertions will be handled.

hililiwei:
Hi @kbendick, thank you for your timely review.
In the latest revision, I adopted the split scheme for subsequent review.
1. Keep the original method for the common part of the test.
2. separated the different tests(for no_nans) into two methods, one for ORC and one for Parquet, named `testIsNanOrc` and `testIsNanParquet`.

data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java
```diff
@@ -298,17 +298,32 @@ public void testNoNulls() {
...
+  }
+
+  private void testIsNanParquet() {
+    boolean shouldRead = shouldRead(isNaN("no_nans"));
+    Assert.assertTrue("Should read: NaN counts are not tracked in Parquet metrics", shouldRead);
+  }
+
+  private void testIsNanOrc() {
+    boolean shouldRead = shouldRead(isNaN("no_nans"));
+    Assert.assertFalse("Should skip: no nans will not contain nan value", shouldRead);
```
kbendick, line 326: Good work @hililiwei! I like this a lot more.

I think we might be ok without the `testIsNan${format}` helper methods. We can just inline within the test as most of the logic is shared.

So something like the following, now that I see that so much of the logic can be shared.
```java
shouldRead = shouldRead(isNaN("no_nans"));
if (format == FileFormat.ORC) {
  Assert.assertFalse("Should skip: no nans will not contain nan value", shouldRead);
} else {
  Assert.assertTrue("Should read: NaN counts are not tracked in Parquet metrics", shouldRead);
}
```

kbendick (commented):
Great work @hililiwei!

This is almost there for me. Just one more adjustment on my end.

data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java
kbendick: For the ORC assertion message, it might be more clear if we're a little more descriptive, given that we've stated that the `NaN counts are not tricked in orc metrics` already.

Something like `Should skip: The existence of NaN in a column is tracked in ORC metrics`. WDYT?

hililiwei: > Good work @hililiwei! I like this a lot more.
> 
> I think we might be ok without the `testIsNan${format}` helper methods. We can just inline within the test as most of the logic is shared.
> 

good!

kbendick (approved):
This looks good to me.

I'm wondering if we can split the upgrade from 1.7.3 -> 1.7.4 out and whether or not that would be beneficial. I don't think it adds benefit, and could possibly make cherry-picking harder (although some people might not have ever cherry-picked the original patch).

Unless anybody objects, I think it's probably safer to do both in one pass so that people who upgrade to 1.7.4 don't accidentally continue to use the reflection accessed hidden methods.

cc @RussellSpitzer @szehon-ho and @dongjoon-hyun for your input on splitting the version upgrade into two PRs (as I know you maintain a fork) and/or final approval. 👍

dongjoon-hyun (changes requested):
Hi, All.

I believe that the following PR should be merged first. @williamhyun was the Apache ORC PMC member and the release manager to make this happen after all testing. And, he made a PR first.

- #4573

Thank you for pinging me, @kbendick . The above PR is focusing on only upgrading the dependency.

dongjoon-hyun:
cc @openinx because he was reviewed there and gave some comments on that.

data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java
```diff
@@ -298,17 +298,23 @@ public void testNoNulls() {
...
+    Assert.assertTrue(String.format("Should read: NaN counts are not tracked in %s metrics", format), shouldRead);
 
     shouldRead = shouldRead(isNaN("all_nulls"));
     Assert.assertFalse("Should skip: all null column will not contain nan value", shouldRead);
+
+    shouldRead = shouldRead(isNaN("no_nans"));
+    if (format == FileFormat.ORC) {
+      Assert.assertFalse("Should skip: no nans will not contain nan value", shouldRead);
+    } else {
+      Assert.assertTrue("Should read: NaN counts are not tracked in Parquet metrics", shouldRead);
```
openinx, line 316: Why this PR will affect the parquet test case ?

openinx (commented):
Almost looks good to me, I just left one question !

RussellSpitzer:
Merged the Version bump https://github.com/apache/iceberg/pull/4573 here, The difference in the test case is that previously assumed all file formats would not be able to skip the row group, now with the new ORC version ORC can skip the row group.

kbendick:
Hi @hililiwei! Can you update the PR description, now that this PR is not updating the ORC version?

kbendick:
Given that the tests were also updated in #4573, I think that this is good to go. Rerequesting review from @dongjoon-hyun (to clear the impression that further changes are expected). Please correct me if I'm wrong.

Corresponding test adjustment: https://github.com/apache/iceberg/blob/04871b79e9aba02e9534498fa035b627bcb9af6e/data/src/test/java/org/apache/iceberg/data/TestMetricsRowGroupFilter.java#L299-L321

hililiwei:
> update

 updated

hililiwei:
cc @openinx  @RussellSpitzer 😄

pvary (approved):

RussellSpitzer (approved):
Do we shade in the ORC Version? Just checking because this could potentially break folks who are using the older versions of ORC at runtime right?

openinx:
Looks good to me !  Thanks all for the reviewing !

pvary:
> Do we shade in the ORC Version? Just checking because this could potentially break folks who are using the older versions of ORC at runtime right?

I think we do not shade things. We had to do that in Hive code to avoid clashing versions. We also struggle with Parquet and Avro. It might worth considering shading our own versions of file formats. But it is for another PR.

kbendick:
> > Do we shade in the ORC Version? Just checking because this could potentially break folks who are using the older versions of ORC at runtime right?
> 
> I think we do not shade things. We had to do that in Hive code to avoid clashing versions. We also struggle with Parquet and Avro. It might worth considering shading our own versions of file formats. But it is for another PR.

We do shade parquet. If you open up a spark repl with the iceberg-spark-runtime it’s under `org.apache.iceberg.relocated` I believe.

pvary:
> > > Do we shade in the ORC Version? Just checking because this could potentially break folks who are using the older versions of ORC at runtime right?
> > 
> > 
> > I think we do not shade things. We had to do that in Hive code to avoid clashing versions. We also struggle with Parquet and Avro. It might worth considering shading our own versions of file formats. But it is for another PR.
> 
> We do shade parquet. If you open up a spark repl with the iceberg-spark-runtime it’s under `org.apache.iceberg.relocated.org.apache.parquet` I believe. If the others aren’t there (which I believe they are), we should look into adding them.

Thanks for the pointer @kbendick. So the issue is solved for all the runtime jars (found the same code for Flink and Hive too). So it is just the direct users of the core jars who has to do that themselves.

RussellSpitzer:
If that's the case let's be sure to note this in the 0.14 release notes

hililiwei:
Thanks @openinx  and all for the reviewing ! 

@RussellSpitzer Do we need to add shadows in core modules?  If needed, I can try to rise a PR. 😄

PR #4736, opened 2022-05-10
WIP: Improve performance of expire snapshot by not double-scanning non-expired manifests

szehon-ho:
Variation of https://github.com/apache/iceberg/pull/3457 that attempts to use time-travel on Manifest table to do the snapshot filtering.  Demonstration, as it seems Manifest table does not support time travel yet.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/Spark3Util.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/Spark3Util.java
@@ -632,6 +632,22 @@ public class Spark3Util {
   }
 
   /**
+   * Returns a metadata table as a Dataset based on the given Iceberg table.
+   *
+   * @param spark SparkSession where the Dataset will be created
+   * @param table an Iceberg table
+   * @param type the type of metadata table
+   * @param snapshotId snapshot id
+   * @return a Dataset that will read the metadata table
+   */
+  private static Dataset<Row> loadMetadataTable(SparkSession spark, org.apache.iceberg.Table table,
+                                                MetadataTableType type, Long snapshotId) {
+    Table metadataTable = new SparkTable(MetadataTableUtils.createMetadataTableInstance(table, type), snapshotId,
+        false);
+    return Dataset.ofRows(spark, DataSourceV2Relation.create(metadataTable, Some.empty(), Some.empty()));
+  }
+
+  /**
    * Returns an Iceberg Table by its name from a Spark V2 Catalog. If cache is enabled in {@link SparkCatalog},
    * the {@link TableOperations} of the table may be stale, please refresh the table to get the latest one.
    *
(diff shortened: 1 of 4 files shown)


aokolnychyi:
I think this is an interesting idea. Let me summarize how I understand it.

Right now, we compute a diff between the reachability sets before and after snapshot expiry. Whenever we build the reachability set before the expiry, we read manifests of all snapshots and that seems suboptimal. The assumption is that it is sufficient to just build the reachability set for expired snapshots and compare that to the reachability set after the snapshot expiry. Did I get that right?

I guess one way to implement (and I believe this is what this PR tries to do) is to load the `FILES` metadata table for each expired snapshot and union `DataFrame`s to produce the reachability set. One potential problem with that is that we will load the manifest list for every expired snapshot on the driver, which can become a bottleneck if we expire a lot of snapshots. I've seen such cases.

An alternative idea is to add some sort of `snapshot-ids` option to `ALL_MANIFESTS` metadata table and read only manifest lists for snapshots whose ID is in that list. That way, we won't read manifest lists and manifests added after the last expired snapshot. That's already an optimization. However, we can go even further and remove still live manifests before opening them, which could give even a bigger performance boost. Suppose an expired snapshot references `manifest-1` but we know it is still live. In that case, we can skip opening it up.

Does that seem reasonable?

@szehon-ho @RussellSpitzer @flyrain @rdblue @kbendick

aokolnychyi:
@RussellSpitzer, could you cross-check the idea as you worked on this part in the past?

rdblue:
> Do you refer to point 5 or 6 as optional?

You're right: it's 5 that is optional. My mistake.

RussellSpitzer:
The manifest skipping idea sounds good to me, basically lets us cut down on the "except" cost as well.

kbendick:
Thanks for tagging me Anton.

I think this is a good idea as well.

I won’t repeat others but I have some questions and possibly a few additional ideas but I need to go through some theoretical cases on paper first.

> One potential problem with that is that we will load the manifest list for every expired snapshot on the driver, which can become a bottleneck if we expire a lot of snapshots. I've seen such cases.

Two thoughts:

1) We should add an event / metric describing this replanning work. Could be used as a signal to perform table maintenance.
2) We might be able to track a metric to determine if we should do this initial replanning work in a distributed manner.

szehon-ho:
Hi @aokolnychyi @rdblue , update on this discussion (see main changes in #3457).  I  implemented and was playing around with these optimizations in calculating the delete candidate files:  (skipping valid reference_snapshot_ids, and skipping valid manifests).  But I find actually that this skip : 
```all_manifest table.reference_snapshot_id in  set(expired snapshots ids)``` makes a positive difference.  The result is the same as what I saw initially, reduce time of Spark job 30-40% in the case of expiring 1 snapshot out of many snapshots.  

As each manifest in all-manifest table is now marked with snapshot id, I am thinking the first filter already kind of filters out any current reachable manifests.   So implementing the skipping of valid manifests actually makes the performance worse, as it adds one more Spark read of all_manifest table and a Spark join for little gain.  Unless there is a scenario I am not thinking of.

PR #4737, opened 2022-05-10
delete unnecessary switch case judgement

flashJd:
private static RowData convert(Types.StructType struct, Record record) invoke private static Object convert(Type type, Object object), the latter has already use a switch case to handle STRUCT,LIST,MAP，so I just remove the unnecessary judgement  in the caller method.

### DIFF
--- a/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/RowDataConverter.java
+++ b/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/RowDataConverter.java
@@ -62,6 +62,7 @@ public class RowDataConverter {
     List<Types.NestedField> fields = struct.fields();
     for (int i = 0; i < fields.size(); i += 1) {
       Types.NestedField field = fields.get(i);
+
       Type fieldType = field.type();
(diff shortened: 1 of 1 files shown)


flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/RowDataConverter.java
```diff
@@ -62,22 +62,8 @@ private static RowData convert(Types.StructType struct, Record record) {
     List<Types.NestedField> fields = struct.fields();
     for (int i = 0; i < fields.size(); i += 1) {
       Types.NestedField field = fields.get(i);
-
```
rdblue, line 65: Please revert this unnecessary whitespace change.

flashJd: I have reverted the unnecessary whitespace change.

flashJd: @rdblue anything do I need to modify?

rdblue:
Thanks, @flashJd!

PR #4537, opened 2022-04-11
Add reader and writer for Puffin, indexes and stats file format

findepi:
Format documentation: https://github.com/apache/iceberg/pull/4944

PR #4741, opened 2022-05-10
Expose table statistics in Table API

findepi:
Follows https://github.com/apache/iceberg/pull/4945  and https://github.com/apache/iceberg/pull/5021


findepi:
@rdblue i have this split. https://github.com/apache/iceberg/pull/4537 has already been reviewed.

core/src/main/java/org/apache/iceberg/SnapshotParser.java
```diff
@@ -147,21 +185,69 @@ static Snapshot fromJson(FileIO io, JsonNode node) {
 
     Integer schemaId = JsonUtil.getIntOrNull(SCHEMA_ID, node);
 
+    List<StatisticsFile> statistics = null;
```
zinking, line 188: not sure how this project deals with null, how about init this to empty collection and void null checking as possible?

core/src/main/java/org/apache/iceberg/GenericStatisticsFile.java
```diff
@@ -0,0 +1,76 @@
...
+      Map<String, Set<Set<Integer>>> statistics) {
+    this.location = Preconditions.checkNotNull(location, "location is null");
+    this.fileSizeInBytes = fileSizeInBytes;
+    this.fileFooterSizeInBytes = fileFooterSizeInBytes;
+    this.sequenceNumber = sequenceNumber;
+    Preconditions.checkNotNull(statistics, "statistics is null");
+    this.statistics = statistics.entrySet().stream()
+        .collect(ImmutableMap.toImmutableMap(
+            Map.Entry::getKey,
+            entry -> entry.getValue().stream().map(ImmutableSet::copyOf).collect(ImmutableSet.toImmutableSet())));
```
zinking, line 49: maybe this line deserves a method name to explain a bit.

core/src/main/java/org/apache/iceberg/SnapshotParser.java
```diff
@@ -147,21 +185,69 @@ static Snapshot fromJson(FileIO io, JsonNode node) {
...
+  private static StatisticsFile statisticsFileFromJson(JsonNode json) {
+    Preconditions.checkArgument(json != null && !json.isNull() && json.isObject(),
+        "Cannot parse statistics from non-object value: %s", json);
+
+    String location = JsonUtil.getString(LOCATION, json);
+    long fileSizeInBytes = JsonUtil.getLong(FILE_SIZE_IN_BYTES, json);
+    long fileFooterSizeInBytes = JsonUtil.getLong(FILE_FOOTER_SIZE_IN_BYTES, json);
+    long sequenceNumber = JsonUtil.getLong(SEQUENCE_NUMBER, json);
+    JsonNode statisticsNode = json.get(STATISTICS);
+    Preconditions.checkArgument(statisticsNode != null && !statisticsNode.isNull() && statisticsNode.isObject(),
```
zinking, line 221: is this for avoiding some json library, what about just parse the class using json first ?

core/src/main/java/org/apache/iceberg/SnapshotParser.java
findepi: @zinking i agree with you and what you suggest is how i write code myself. However, i think i am following project rules here.

core/src/main/java/org/apache/iceberg/SnapshotParser.java
findepi: Yes, Iceberg doesn't use annotation-based JSON parsing/generation. a project design decision

alexjo2144:
We need to remember to keep `remove_orphan_files` from deleting Statistics files

findepi:
> We need to remember to keep `remove_orphan_files` from deleting Statistics files

Thanks for reminding about this here. It's on my todo list, see eg https://github.com/trinodb/trino/pull/12317#discussion_r876044928

findepi:
> We need to remember to keep `remove_orphan_files` from deleting Statistics files

Fixed & test added.

core/src/main/java/org/apache/iceberg/GenericStatisticsFile.java
findepi: refactored

findepi:
> Conflicting files
.palantir/revapi.yml
core/src/main/java/org/apache/iceberg/BaseSnapshot.java

ops.

api/src/main/java/org/apache/iceberg/StatisticsFile.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.iceberg;
+
+import java.util.List;
+
+public interface StatisticsFile {
+  String location();
+
+  long fileSizeInBytes();
```
zinking, line 27: I just find that parquet provides un-compressed size in footer as well. is it possible to include this in the stats,  un-compressed size could be used in more accurate stats calculation.

findepi: this just for reading the stats file itself; it's not stats about the parquet file, or data inside the data file

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
```diff
@@ -325,7 +325,7 @@ public void testAddSnapshotToJson() {
 
     Snapshot snapshot = new BaseSnapshot(null, snapshotId, parentId, System.currentTimeMillis(),
         DataOperations.REPLACE, ImmutableMap.of("files-added", "4", "files-deleted", "100"),
-        schemaId, manifests);
+        schemaId, manifests, null);
```
rdblue, line 328: Rather than making so many changes to add `null` to tests, can you add a package-private constructor that defaults this?

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
```diff
@@ -64,6 +64,7 @@ private static class InMemorySeekableInputStream extends SeekableInputStream {
 
     private final int length;
     private final ByteArrayInputStream delegate;
+    private boolean closed = false;
```
rdblue, line 67: Since these changes are mostly unrelated, can you separate them into their own PR?

core/src/main/java/org/apache/iceberg/puffin/Blob.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.puffin;
```
rdblue, line 20: Looks like this includes a PR that's been merged. Can you rebase to remove the extra changes?

api/src/main/java/org/apache/iceberg/UpdateTableStatistics.java
```diff
@@ -0,0 +1,44 @@
...
+   * Delete a statistics file path from the underlying table.
+   * <p>
+   * To remove a statistics file from the table, this path must equal a path in the table's metadata. Paths
+   * that are different but equivalent will not be removed. For example, file:/path/file.avro is
+   * equivalent to file:///path/file.avro, but would not remove the latter path from the table.
+   *
+   * @param path a fully-qualified file path to remove from the table
+   * @return this for method chaining
+   */
+  UpdateTableStatistics deleteStatisticsFile(CharSequence path);
```
rdblue, line 36: I think it would be fine for this to be `replaceStatistics` (and possibly validate the latest stats file has not changed) or `setStatistics`. I don't really see the issue that you raised in [the other thread](https://github.com/apache/iceberg/pull/4945#discussion_r894207600).

Either way, a caller updating statistics will do some calculation on the table and write a statistics file. Whether that stats file is based on the last doesn't really affect this API.

api/src/main/java/org/apache/iceberg/Transaction.java
```diff
@@ -137,6 +137,14 @@ default AppendFiles newFastAppend() {
    */
   DeleteFiles newDelete();
 
+  /**
+   * Create a new {@link UpdateTableStatistics update table statistics API} to add or remove statistics
+   * files in this table.
+   *
+   * @return a new {@link UpdateTableStatistics}
+   */
+  UpdateTableStatistics newUpdateTableStatistics();
```
rdblue, line 146: I think this should be called `updateTableStatistics`. No need for "new" because "update" is a verb.

core/src/test/java/org/apache/iceberg/io/InMemoryInputFile.java
findepi: sure: https://github.com/apache/iceberg/pull/5019

core/src/main/java/org/apache/iceberg/puffin/Blob.java
findepi: > Looks like this includes a PR that's been merged. Can you rebase to remove the extra changes?

This PR is based on https://github.com/apache/iceberg/pull/4537. I don't think the latter is merged yet. Was it supposed to be?

core/src/test/java/org/apache/iceberg/TestMetadataUpdateParser.java
findepi: sure, adding

api/src/main/java/org/apache/iceberg/Transaction.java
findepi: I guess i named this after newReplacePartitions or newOverwrite ("replace" and "overwrite" are also verbs).
Sure, will remove "new" here.

> Can you add the API to `Table` and `Transaction` in a separate PR?

Sure.
Should come together with or separate from BaseTransaction changes?
Should come together with or separate from BaseSnapshot changes?
Should come together with or separate from remove_orphan_files changes?

findepi: Extracted https://github.com/apache/iceberg/pull/5021. Let me know if I got this right.

rdblue: I think we prefer not to add the implementation changes. Just the API addition with a default that throws an UnsupportedOperationException.

api/src/main/java/org/apache/iceberg/UpdateTableStatistics.java
findepi: > 1. There's nothing preventing the stats from being in the same file. Even if a writer doesn't understand some of the blobs, it can still preserve them in the same file where it is adding new ones. There isn't a need for separate files.

agreed



> 2\. Most of the time, writers will understand the blobs produced by other writers. The aim is to standardize the stats and index content. Situations where Ryan's stats and Piotr's stats are completely different should be rare.

they can be of different types: one can be a sketch, the other a bloom filter, or a histogram

> While it is more work for the writer, this ensures that stats are maintained. Tracking multiple files makes it so we need to worry about pruning those files.

agreed

> One problem with this is that I don't think we are tracking freshness metadata in the Puffin file BlobMetadata, only for the file itself,

Yes.

> I think we should add snapshot-id and sequence-number to blob metadata.

we can add them as standard properties in the BlobMetadata.properties field.
This would also propagate to the metadata properties we have here in the table, and thus would let the reader determine whether the stats are "fresh enough" for their needs, without reading the Puffin footer.

WDYT?

findepi:
Updated to the current state of https://github.com/apache/iceberg/pull/5021

findepi:
Rebased after #5021 has been merged to make Conflicts disappear.

findepi:
@rdblue please let me know if you have any comments.

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -308,6 +308,13 @@ default UpdateStatistics updateStatistics() {
   /** Returns a {@link LocationProvider} to provide locations for new data files. */
   LocationProvider locationProvider();
 
+  /**
+   * Returns the current statistics files for the table
+   *
+   * @return the current statistics files for the table
+   */
+  List<StatisticsFile> statisticsFiles();
```
rdblue, line 316: Do we want to expose these directly? What about instead adding an API for finding statistics? That way consumers don't need to decide which one to use themselves.

api/src/main/java/org/apache/iceberg/UpdateStatistics.java
```diff
@@ -18,8 +18,10 @@
  */
 package org.apache.iceberg;
 
+import java.util.List;
+
 /** API for updating statistics files in a table. */
-public interface UpdateStatistics extends PendingUpdate<Snapshot> {
+public interface UpdateStatistics extends PendingUpdate<List<StatisticsFile>> {
```
rdblue, line 24: Good catch here.

core/src/main/java/org/apache/iceberg/GenericBlobMetadata.java
```diff
@@ -26,7 +27,17 @@
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 
-public class GenericBlobMetadata implements BlobMetadata {
+public class GenericBlobMetadata implements BlobMetadata, Serializable {
+
+  public static BlobMetadata from(org.apache.iceberg.puffin.BlobMetadata puffinMetadata) {
```
rdblue, line 32: Why don't we make it so that the Puffin `BlobMetadata` implements `BlobMetadata`? Then this would be unnecessary.

core/src/main/java/org/apache/iceberg/GenericStatisticsFile.java
```diff
@@ -18,13 +18,14 @@
...
 
+import java.io.Serializable;
 import java.util.List;
 import java.util.Objects;
 import java.util.StringJoiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 
-public class GenericStatisticsFile implements StatisticsFile {
+public class GenericStatisticsFile implements StatisticsFile, Serializable {
```
rdblue, line 28: Why does this need to be serializable?

rdblue: Oh, I see. For `SerializableTable`. I don't think that we need to actually serialize these. Instead, you can delegate to `lazyTable()` like some of the other methods. We don't expect serialized versions to use this since it is for planning on the coordinator or driver.

core/src/main/java/org/apache/iceberg/GenericBlobMetadata.java
```diff
@@ -26,7 +27,17 @@
...
+    return new GenericBlobMetadata(
+        puffinMetadata.type(),
+        puffinMetadata.snapshotId(),
+        puffinMetadata.sequenceNumber(),
+        puffinMetadata.inputFields(),
+        puffinMetadata.properties());
```
rdblue, line 38: If you want to implement `Serializable` then this needs to be wrapped in a `SerializableMap`.

core/src/main/java/org/apache/iceberg/GenericBlobMetadata.java
findepi: Puffin BlobMetadata contains all the blob metadata information from the Puffin footer.
In particular, it contains more fields, like offset and length, which are not carried over to the table metadata.

We can make Puffin `BlobMetadata` implement the `BlobMetadata`, but
- we don't want to serialize (as in `Serializable`) the Puffin BlobMetadata, as we're not interested in sending over those additional fields (I know these are small today, and _maybe_ are never big in the future)
- there is a challenge with equality semantics. `GenericBlobMetadata` implements equals, but with more complicated class hierarchy it wouldn't be useful anymore.

Let me know what you think

api/src/main/java/org/apache/iceberg/Table.java
findepi: > Do we want to expose these directly? 

Yes, these should be exposed, just like we expose other information about table (files, their min/maxes, etc.) _and_ we provide convenience functionality on top of them (like `planFiles`).

> What about instead adding an API for finding statistics?

Good idea. What would you like to see here?

core/src/main/java/org/apache/iceberg/GenericBlobMetadata.java
findepi: > we don't want to serialize (as in Serializable) the Puffin BlobMetadata

i see https://github.com/apache/iceberg/pull/4741#discussion_r966204993 now

core/src/main/java/org/apache/iceberg/ReachableFileUtil.java
```diff
@@ -113,4 +113,18 @@ public static List<String> manifestListLocations(Table table) {
...
+   *
+   * @param table table for which statistics files needs to be listed
+   * @return the location of statistics files
+   */
+  public static List<String> statisticsFilesLocations(Table table) {
+    List<String> statisticsFilesLocations = Lists.newArrayList();
+    for (StatisticsFile statisticsFile : table.statisticsFiles()) {
+      statisticsFilesLocations.add(statisticsFile.path());
+    }
+    return statisticsFilesLocations;
```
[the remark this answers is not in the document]
findepi: Adding. I was just trying to be consistent with the rest of this file

findepi:
Per https://github.com/apache/iceberg/pull/4741#discussion_r966200811, I've shrunk the scope of this PR
and extracted https://github.com/apache/iceberg/pull/5794 and also https://github.com/apache/iceberg/pull/5795 .

rdblue:
Thanks, @findepi! I'll give this another look.

findepi:
Thanks @rdblue . https://github.com/apache/iceberg/pull/5799 might be the easiest to start from.

findepi:
@rdblue this is ready to review.

rdblue (approved):

rdblue:
Thanks, @findepi!

findepi:
Thank you for the merge!

PR #4742, opened 2022-05-10
Python: Convert Avro to Iceberg schema

Fokko:
Converts an Avro schema into an Iceberg one.

Looked at https://github.com/apache/iceberg/blob/master/python_legacy/iceberg/core/avro/avro_to_iceberg.py but I decided to rewrite it to make it more pythonic.


python/src/iceberg/schema.py
```diff
@@ -67,6 +67,17 @@ def __repr__(self):
...
+    def __eq__(self, other) -> bool:
+        if not other:
+            return False
+
+        columns = list(self.columns)
+
+        if len(columns) != len(other.columns):
+            return False
+
+        return all([lhs == rhs for lhs, rhs in zip(columns, other.columns)])
```
kbendick, line 79: Question: Should we be comparing the `NestedField` based on id as well? I _believe_ we're ok to just assume that they're in order, but good to check. Here's the Java equals for `NestedField`:

https://github.com/apache/iceberg/blob/06cdfa3f7c4ebd26766bc67100ba5d30b6bf3940/api/src/main/java/org/apache/iceberg/types/Types.java#L502-L521

python/src/iceberg/schema.py
```diff
@@ -67,6 +67,17 @@ def __repr__(self):
             f"Schema(fields={repr(self.columns)}, schema_id={self.schema_id}, identifier_field_ids={self.identifier_field_ids})"
         )
 
+    def __eq__(self, other) -> bool:
```
kbendick, line 70: I think we also need to check the identifier fields. Java's `Schema::isSameSchema` does so.

https://github.com/apache/iceberg/blob/06cdfa3f7c4ebd26766bc67100ba5d30b6bf3940/api/src/main/java/org/apache/iceberg/Schema.java#L444-L452

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            avro_types = type_union
+
+        is_optional = "null" in avro_types
+
+        # Filter the null value, so we know the actual type
+        avro_types = list(filter(lambda t: t != "null", avro_types))
+
+        if len(avro_types) != 1:
+            raise ValueError("Support for unions is yet to be implemented")
```
rdblue, line 144: This is outside the Iceberg spec (although #4242 is proposing a way to handle it in Java).

I think this can just state that the type is invalid because it is a non-option union.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+        Args:
+            type_union: The field, can be a string 'str', list ['null', 'str'], or dict {"type": 'str'}
+
+        Returns:
+            A tuple containing the type and nullability
+
+        """
+        avro_types: Union[Dict, List]
+        if isinstance(type_union, str):
+            avro_types = [type_union]
```
rdblue, line 134: Why not just return `(type_union, False)` here directly?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+            >>> avro_schema == iceberg_schema
+            True
+
+        Args:
+            field:
+
+        Returns:
+
+        """
+        field_id = field["field-id"]
```
rdblue, line 231: Should we throw a better error than `KeyError` if there is no field ID? I think that would be nice since this ID is specific to Iceberg.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+                field_type=AvroSchemaConversion.PRIMITIVE_FIELD_TYPE_MAP[avro_type],
+                is_optional=is_optional,
+                doc=field_doc,
+            )
+        elif avro_type == "record":
+            return NestedField(
+                field_id=field_id, name=field_name, field_type=self._parse_record(field), is_optional=is_optional, doc=field_doc
+            )
+        elif avro_type == "array":
+            inner_type, _ = self._resolve_union(field["type"])
```
rdblue, line 252: It seems a little strange to me that this has to call `_resolve_union` twice because the first call replaces `avro_type` with the inner string. Why not keep the original `avro_type` around and use it here rather than calling this a second time?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+            assert isinstance(inner_type, dict)
```
rdblue, line 253: If this fails, do we get a good error message? I generally prefer to throw a `ValueError` with a good message if there's a chance anyone would hit this.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+            inner_type["items"]["field-id"] = inner_type["element-id"]
```
rdblue, line 254: I don't think there is a guarantee that `"items"` returns a `dict`. The example from the Avro spec is `{ "type": "array", "items": "string", "default": [] }` so we know `"items"` can be a string.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+            inner_field = self._parse_field(inner_type["items"])
```
rdblue, line 255: I think it would be cleaner to separate type conversion from field conversion. It's a bit awkward to me that this always converts the type to a field and then gets its type back out. Plus, this hits the same problem as above, where `inner_type["items"]` may not be a dict.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+        "boolean": BooleanType(),
+        "bytes": BinaryType(),
+        "date": DateType(),
+        "double": DoubleType(),
+        "float": FloatType(),
+        "int": IntegerType(),
+        "long": LongType(),
+        "string": StringType(),
+        "time-millis": TimeType(),
+        "timestamp-millis": TimestampType(),
```
rdblue, line 55: For date, time, timestamp, and decimal I think you'll need to look at the logical type annotation. For fixed, you'll have to add a branch to the type conversion since you need the length.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+                name=field_name,
+                field_type=AvroSchemaConversion.PRIMITIVE_FIELD_TYPE_MAP[avro_type],
+                is_optional=is_optional,
+                doc=field_doc,
+            )
+        elif avro_type == "record":
+            return NestedField(
+                field_id=field_id, name=field_name, field_type=self._parse_record(field), is_optional=is_optional, doc=field_doc
+            )
+        elif avro_type == "array":
```
rdblue, line 251: I think you'll have to check for a logical type annotation here that signals a map with a non-string key.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,274 @@
...
+            )
+        elif avro_type == "map":
+            return NestedField(
+                field_id=field_id,
+                name=field_name,
+                field_type=MapType(
+                    key_id=field["key-id"],
+                    key_type=StringType(),
+                    value_id=field["value-id"],
+                    value_type=self._parse_field(field["values"]),
```
rdblue, line 268: I think there are two problems here. First, this returns a field so you probably want to get the `type` from it. Second, `field["values"]` could be a string, which would cause conversion to a field to fail.

python/src/iceberg/schema.py
Fokko: Great point @kbendick. Now https://github.com/apache/iceberg/pull/4767 has been merged, the eq methods are generated automatically :)

python/src/iceberg/utils/schema_conversion.py
Fokko: Good to know, I'll update the error! Thanks!

python/src/iceberg/utils/schema_conversion.py
Fokko: Good one, I'll update it

python/src/iceberg/schema.py
Fokko: This is now also implemented using the autogenerated `__eq__` of the dataclass 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Good one. I was thinking as a potential follow-up is setting id's just sequentially when traversing the schema if the id's aren't missing. This should be a conscious choice then 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Ah, this is a relic from the refactoring. Thanks for catching this!

python/src/iceberg/utils/schema_conversion.py
Fokko: That's correct, it can also be a primitive 👍🏻  I've created test cases and working on a fix

python/src/iceberg/utils/schema_conversion.py
Fokko: Thanks for pointing this out. There is actually a bug in the array conversion. In the test cases, you can see that the ListType is missing, I'll push a fix tomorrow morning.

python/src/iceberg/types.py
```diff
@@ -146,7 +146,7 @@ class NestedField(IcebergType):
     name: str = field()
     field_type: IcebergType = field()
     is_optional: bool = field(default=True)
-    doc: Optional[str] = field(default=None, repr=False)
+    doc: Optional[str] = field(default=None, repr=False, compare=False, hash=False)
```
Fokko, line 149: @rdblue do we want to compare the docs as well? I know that for Avro we only compare the fields that are part of the canonical representation. I caught this in a test where my example didn't have the docs 🤔

python/src/iceberg/utils/schema_conversion.py
Fokko: Done! Added a test for the date, and split the logical types out on their own map.

python/src/iceberg/utils/schema_conversion.py
Fokko: I've refactored this and added tests for the permutations:
- `"str"`
- `["str", "null"]`
- `["null", { record }]`

python/src/iceberg/utils/schema_conversion.py
Fokko: Added with tests

python/src/iceberg/utils/schema_conversion.py
Fokko: Fixed both of them 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Great catch, thanks! Just added it with a test

Fokko:
@rdblue @kbendick I've updated the PR with the comments addressed 👍🏻

rdblue:
Thanks! I'll take another look.

python/src/iceberg/schema.py
rdblue: Does this `__eq__` implementation override the autogenerated one? If we're using a generated one, I don't understand why this would still be needed.

python/src/iceberg/types.py
rdblue: We compare docs in the Java implementation. That's because equality for types is that the type matches exactly, including documentation.

The question about whether a field is "equal" if the doc string is different is the reason why we don't implement `equals` in Java's `Schema` class. We could do the same here if you want.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,371 @@
...
+        "date": DateType(),
+        "time-millis": TimeType(),
+        "timestamp-millis": TimestampType(),
+        "uuid": UUIDType(),
+    }
+
+    def avro_to_iceberg(self, avro_schema: Dict[str, Any]) -> Schema:
+        """Converts an Apache Avro into an Apache Iceberg schema equivalent
+
+        This expects to have field id's to be encoded in the Avro schema::
```
rdblue, line 70: You may want to mention in the docs that this handles string types, like "long".

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,371 @@
...
+        Args:
+            avro_schema (Dict[str, Any]): The JSON decoded Avro schema
+
+        Returns:
+            Equivalent Iceberg schema
+        """
+        fields = self._parse_record(avro_schema)
+        return Schema(*fields.fields, schema_id=1)
+
+    def _parse_record(self, avro_field: Dict[str, Any]) -> StructType:
```
rdblue, line 113: Nit: `avro_field` is a misleading name because it's actually a record.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,371 @@
...
+        fields = [self._parse_field(field) for field in avro_field["fields"]]
```
rdblue, line 114: Should this check that the Avro schema's `type` is `record`? That seems safer than assuming that `fields` is a key.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,371 @@
...
+        return StructType(*fields)  # type: ignore
+
+    def _resolve_union(self, type_union: Union[Dict, List, str]) -> Tuple[Union[str, Dict[str, Any]], bool]:
```
rdblue, line 117: Minor: now that we're using 3.8, we can use the `|` operator for `Union` right?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,371 @@
...
+        if isinstance(type_union, str):
+            # It is a primitive and required
+            return type_union, False
+        elif isinstance(type_union, dict):
+            # It is a context and required
+            return type_union, False
+        else:
+            avro_types = type_union
+
+        is_optional = "null" in avro_types
```
rdblue, line 150: You could also compare the size of `avro_types` after filtering with `avro_types` before filtering.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,371 @@
...
+        Args:
+            field: The Avro field
+            field_id: Ability to override the field_id when it is provided from up in the tree (in the case of a list or map)
+
+        Returns:
+            The equivalent IcebergType
+        """
+        # In the case of a primitive field
+        if isinstance(field, str):
+            return AvroSchemaConversion.PRIMITIVE_FIELD_TYPE_MAP[field]
```
rdblue, line 259: I find the logic in these methods hard to follow and I think it's because the cases are not cleanly separated by method. This is a good example. This method is handling an Avro field, but this is checking if the field is not a field and is instead a primitive type. That should never happen in a schema so it raises questions about when this method is called.

This is one reason why we use the visitor pattern elsewhere. Keeping the logic to traverse a schema separate from the logic to actually do something with it is useful, but it also keeps you using a consistent and focused pattern to construct this logic: here's how to convert a record, here's how to convert a field, here's how to convert a map, etc.

Since this isn't handling an Avro Schema class, I wasn't originally going to suggest it, but I think this would be cleaner and easier to review/maintain if it were structured around Avro's schema model:

```python
Schema = 
def _convert_schema(schema: str | Dict[str, Any]):
    if isinstance(schema, str):
        return AvroSchemaConversion.PRIMITIVE_FIELD_TYPE_MAP[schema]
    elif isinstance(schema, dict):
        type_name = schema["type"]
        if type_name == "record":
            return _convert_record(schema)
        elif type_name == "union":
            ...
        elif type_name == "map":
            ...
        elif type_name == "array":
            ...
        else:
            logical_type = schema.get("logicalType")
            if logicalType:
                ...
            else:
                return AvroSchemaConversion.PRIMITIVE_FIELD_TYPE_MAP[type_name]
    else:
        raise ValueError(f"Cannot convert invalid schema: {schema}")

def _convert_record(schema: Dict[str, Any]):
    ... # calls _convert_field

def _convert_field(field: Dict[str, Any]):
    ...
```

python/src/iceberg/schema.py
Fokko: Yes, that's true. I forgot to remove this one, thanks!

python/src/iceberg/types.py
Fokko: Let's include it here for now. I'll revert this change and update the tests 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Yes, let me bulk update this in another PR. There is a nice tool called pyupgrade that will enforce this: https://github.com/asottile/pyupgrade#pep-604-typing-rewrites

python/src/iceberg/utils/schema_conversion.py
Fokko: But then a `['string', 'bytes']` union could also slip through

python/src/iceberg/utils/schema_conversion.py
Fokko: Good one, thanks!

python/src/iceberg/utils/schema_conversion.py
rdblue: The check for number of non-null branches would catch that. But `['string', 'bytes']` does not contain null, which is what you want here.

python/src/iceberg/schema.py
Fokko: I was a bit too quick on this one. Schema doesn't have an autogenerated one, but the types do have one now. So we still need the eq on the schema

python/src/iceberg/utils/schema_conversion.py
Fokko: You're right! I've updated the code

python/src/iceberg/utils/schema_conversion.py
Fokko: Yes, that's much cleaner indeed. The method grew a bit over time, but I really like the suggestion of decoupling the field and type. I've updated the code.

python/src/iceberg/schema.py
rdblue: In that case, I think we'll need to check the identifier field IDs like Kyle suggested

Fokko: Ah, right! Sorry about that

python/src/iceberg/schema.py
```diff
@@ -62,8 +62,23 @@ def __repr__(self):
...
+            return False
+
+        identifier_field_ids_is_equal = self.identifier_field_ids == other.identifier_field_ids
+        schema_is_equal = all([lhs == rhs for lhs, rhs in zip(self.columns, other.columns)])
+
+        return identifier_field_ids_is_equal and schema_is_equal
+
     @property
-    def columns(self) -> Iterable[NestedField]:
+    def columns(self) -> Tuple[NestedField]:
```
rdblue, line 81: Should this be `Tuple[NestedField, ...]` since it isn't a single field in the tuple?

From the [typing.Tuple docs](https://docs.python.org/3/library/typing.html#typing.Tuple):
> To specify a variable-length tuple of homogeneous type, use literal ellipsis, e.g. Tuple[int, ...]

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+    def _convert_field(self, field: Dict[str, Any]) -> NestedField:
+        """
+        Converts an Avro field into an Iceberg equivalent field
+        Args:
+            field: The Avro field
+
+        Returns:
+            The Iceberg equivalent field
+        """
+        assert "field-id" in field, "Missing field-id in the Avro field, this is required for converting it to an Iceberg schema"
```
rdblue, line 216: What is thrown when using `assert` rather than `raise ValueError`?

Also, we try to be more direct with error messages by starting with what was attempted, then giving context for what went wrong. It's also assumed that if an exception is raised then some requirement wasn't met so we'd typically leave out the "this is required ..." part. How about `f"Cannot convert field, missing field-id: {field}"`?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+
+        is_optional = "null" in avro_types
+
+        if len(avro_types) > 2:
+            raise TypeError("Non-optional types aren't part of the Iceberg specification")
+
+        # Filter the null value and return the type
+        return list(filter(lambda t: t != "null", avro_types))[0], is_optional
+
+    def _resolve_inner_type(self, avro_type: Dict[str, Any], inner_field_name: str, id_field: str) -> Tuple[IcebergType, bool]:
```
rdblue, line 155: Nit: since this is used by array and map conversion, it would make more sense to have it near those methods.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+            inner_field_name: For example 'items' in case of an array
+            id_field: Contains the element or key/value id, for example element-id, to be copied to the field-id
+
+        Returns:
+            The inner type and the nullability
+        """
+        plain_type, element_is_optional = self._resolve_union(avro_type[inner_field_name])
+        if isinstance(plain_type, dict):
+            # We need the element-id downstream
+            plain_type["field-id"] = avro_type[id_field]
```
rdblue, line 170: I strongly prefer not modifying the incoming data like this is doing. I think `avro_type` and parts of that type should be handled as though they are immutable.

Also, I don't think that this works. `_convert_schema` is going to return a type. For primitive types, there is no way to pass in the `field-id` property, and in all cases the `field-id` is going to be ignored because Iceberg types don't carry field IDs.

I think the simple solution to both problems is to handle inner types in the map and array methods directly, rather than using a common method.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+        Args:
+            record_type: The record type itself
+
+        Returns:
+        """
+        return StructType(*[self._convert_field(field) for field in record_type["fields"]])
+
+    def _convert_array_type(self, array_type: Dict[str, Any]) -> ListType:
+        element_type, element_is_optional = self._resolve_inner_type(array_type, "items", "element-id")
+        return ListType(element_id=array_type["element-id"], element_type=element_type, element_is_optional=element_is_optional)
```
rdblue, line 281: `_convert_field` uses an assertion to ensure that `field-id` is present. Should this do the same?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+            # Avro only supports string keys
+            key_type=StringType(),
+            value_id=map_type["value-id"],
+            value_type=self._convert_schema(value_type),
+            value_is_optional=value_is_optional,
+        )
+
+    def _convert_logical_type(self, avro_logical_type: Dict[str, Any]) -> IcebergType:
+        """
+        When a logical type is found, we'll resolve it here. For the decimal and map
```
rdblue, line 321: Style: Avoid the use of personal pronouns, like "I" and "we". Those make docs longer and don't add clarity. Here, you can use "Convert a schema with a logical type annotation", which is more direct.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+                if type_identifier == "record":
+                    return self._convert_record_type(avro_type)
+                elif type_identifier == "array":
+                    return self._convert_array_type(avro_type)
+                elif type_identifier == "map":
+                    return self._convert_map_type(avro_type)
+                elif type_identifier == "fixed":
+                    return self._convert_fixed_type(avro_type)
+                else:
+                    raise ValueError(f"Unknown type: {avro_type}")
```
rdblue, line 203: It's uncommon, but it's valid for a primitive to be wrapped in a dict, like `{"type": "array", "items": {"type": "double", "additional": "property"}}`. That should be handled here. I think you can check whether `type_identifier` is in the primitive map.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+        Raises:
+            ValueError: When the logical type is unknown
+        """
+        logical_type = avro_logical_type["logicalType"]
+        if logical_type == "decimal":
+            return self._convert_logical_decimal_type(avro_logical_type)
+        elif logical_type == "map":
+            return self._convert_logical_map_type(avro_logical_type)
+        elif logical_type in LOGICAL_FIELD_TYPE_MAPPING:
+            return LOGICAL_FIELD_TYPE_MAPPING[logical_type]
```
rdblue, line 350: Should we also validate the underlying type? The Java library does this automatically, but I don't think there's any validation in python that the actual type for a `date` is `int`.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+    "float": FloatType(),
+    "int": IntegerType(),
+    "long": LongType(),
+    "string": StringType(),
+}
+
+LOGICAL_FIELD_TYPE_MAPPING: Dict[str, PrimitiveType] = {
+    "date": DateType(),
+    "time-millis": TimeType(),
+    "timestamp-millis": TimestampType(),
```
rdblue, line 63: Iceberg will only write `time-micros` and `timestamp-micros` because at least microsecond precision is required by the SQL spec. The mappings for millis are correct, but we should also be able to handle micros.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+            True
+
+        .. _Apache Iceberg specification:
+            https://iceberg.apache.org/spec/#appendix-a-format-specific-requirements
+
+        Returns:
+            The logical map
+        """
+        fields = avro_type["items"]["fields"]
+        assert len(fields) == 2, f"Expected two fields in the logical map, but got: {fields}"
```
rdblue, line 420: How about `Invalid key-value pair schema: {avro_type["items"]}`?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+        key = self._convert_field(fields[0])
+        value = self._convert_field(fields[1])
```
rdblue, line 422: This is probably okay, but I don't think that there's a guarantee that the fields are in order. It would be good to check the field names.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,451 @@
...
+logger = logging.getLogger(__name__)
+
+PRIMITIVE_FIELD_TYPE_MAPPING: Dict[str, PrimitiveType] = {
+    "boolean": BooleanType(),
+    "bytes": BinaryType(),
+    "double": DoubleType(),
+    "float": FloatType(),
+    "int": IntegerType(),
+    "long": LongType(),
+    "string": StringType(),
```
rdblue, line 57: `"enum"` is also converted to `StringType()`.

python/src/iceberg/utils/schema_conversion.py
Fokko: Good one, I've added a check

python/src/iceberg/schema.py
Fokko: Yes, good one. I know that we're going a bit back and forth on this annotation, but I've double-checked using the inspector, and it is a Tuple at runtime 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Great point. I'm not against assert's personally, but there is a battle in the Python community if you should use them for checks like these. Technically you can ignore them using a flag `python -O`, and they are meant for developers. I've changed them to a ValueError to be in line with the rest of the exceptions (which I like).

I've updated the message to your suggestion. I think it is good to have the field in there 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Good one, moved it 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Seems fair, I've added the check 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Good point, thanks!

python/src/iceberg/utils/schema_conversion.py
Fokko: Good one, haven't seen that myself. Added a test and logic for handling this

python/src/iceberg/utils/schema_conversion.py
Fokko: Replaced it by:
```python
key = self._convert_field(list(filter(lambda f: f["name"] == "key", fields))[0])
value = self._convert_field(list(filter(lambda f: f["name"] == "value", fields))[0])
```

python/src/iceberg/utils/schema_conversion.py
Fokko: We can. I've added logic to lookup the logical/physical combination 👍🏻

python/src/iceberg/utils/schema_conversion.py
Fokko: Hey Ryan, I agree. I don't think we really need the method at all. I've removed it and just resolved the type instead of the field. Still some old code dangling around.

python/src/iceberg/utils/schema_conversion.py
Fokko: I've added the mapping, but I think we need to correct the scale when reading the data. Analog to the Java impl: https://github.com/apache/iceberg/blob/90225d6c9413016d611e2ce5eff37db1bc1b4fc5/core/src/main/java/org/apache/iceberg/avro/GenericAvroReader.java#L137-L140

rdblue: Yes, exactly. We'll need to be aware of whether it is in millis or micros when reading the file!

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,454 @@
...
+            >>> expected == actual
+            True
+
+        Args:
+            record_type: The record type itself
+
+        Returns:
+        """
+        if record_type["type"] != "record":
+            raise ValueError(f"Expected type, got: {record_type}")
```
rdblue, line 268: Expected record?

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,454 @@
...
+            ValueError: When the logical type is unknown
+        """
+        logical_type = avro_logical_type["logicalType"]
+        physical_type = avro_logical_type["type"]
+        if logical_type == "decimal":
+            return self._convert_logical_decimal_type(avro_logical_type)
+        elif logical_type == "map":
+            return self._convert_logical_map_type(avro_logical_type)
+        elif (logical_type, physical_type) in LOGICAL_FIELD_TYPE_MAPPING:
+            return LOGICAL_FIELD_TYPE_MAPPING[(logical_type, physical_type)]
```
rdblue, line 352: I just also realized that this doesn't check for the `adjustToUtc` flag that we use to distinguish between `TimestampType()` and `TimestamptzType()`. We'll have to add that.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,454 @@
...
+            ...     precision=19,
+            ...     scale=25
+            ... )
+            >>> actual == expected
+            True
+
+        Returns:
+            A Iceberg DecimalType
+        """
+        return DecimalType(precision=avro_type["precision"], scale=avro_type["scale"])
```
rdblue, line 380: Since you're validating physical type above, you may want to validate that the physical type here is fixed, long, or int.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -0,0 +1,454 @@
...
+        .. _Apache Iceberg specification:
+            https://iceberg.apache.org/spec/#appendix-a-format-specific-requirements
+
+        Returns:
+            The logical map
+        """
+        fields = avro_type["items"]["fields"]
+        if len(fields) != 2:
+            raise ValueError(f'Invalid key-value pair schema: {avro_type["items"]}')
+        key = self._convert_field(list(filter(lambda f: f["name"] == "key", fields))[0])
```
rdblue, line 424: Looks good. Too bad it is more complicated, but thanks for fixing this.

rdblue (approved):

rdblue:
There are a few minor things to fix, but overall I think this looks good. I'm going to commit it to unblock reader work and we can follow up with minor updates. Thanks, @Fokko!

PR #4744, opened 2022-05-11
Core: Incremental append scan impl

stevenzwu:
this is a follow-up to PR #4580. will need to rebase this branch when 4580 is merged

### DIFF
--- a/api/src/main/java/org/apache/iceberg/Scan.java
+++ b/api/src/main/java/org/apache/iceberg/Scan.java
@@ -29,7 +29,7 @@ import org.apache.iceberg.io.CloseableIterable;
  * Scan objects are immutable and can be shared between threads. Refinement methods, like
  * {@link #select(Collection)} and {@link #filter(Expression)}, create new TableScan instances.
  */
-interface Scan<T extends Scan<T>> {
+public interface Scan<T extends Scan<T>> {
   /**
    * Create a new scan from this scan's configuration that will override the {@link Table}'s behavior based
    * on the incoming pair. Unknown properties will be ignored.
--- a/api/src/main/java/org/apache/iceberg/Scan.java
+++ b/api/src/main/java/org/apache/iceberg/Scan.java
@@ -133,4 +133,19 @@ interface Scan<T extends Scan<T>> {
    * @return an Iterable of tasks for this scan
    */
   CloseableIterable<CombinedScanTask> planTasks();
+
+  /**
+   * Returns the target split size for this scan.
+   */
+  long targetSplitSize();
+
+  /**
+   * Returns the split lookback for this scan.
+   */
+  int splitLookback();
+
+  /**
+   * Returns the split open file cost for this scan.
+   */
+  long splitOpenFileCost();
 }
--- a/api/src/main/java/org/apache/iceberg/Scan.java
+++ b/api/src/main/java/org/apache/iceberg/Scan.java
@@ -85,6 +85,13 @@ interface Scan<T extends Scan<T>> {
   T filter(Expression expr);
 
   /**
+   * Create a new scan from this that applies data filtering to files but not to rows in those files.
+   *
+   * @return a new scan based on this that does not filter rows in files.
+   */
+  T ignoreResiduals();
+
+  /**
    * Create a new scan to use a particular executor to plan. The default worker pool will be
    * used by default.
    *
@@ -92,6 +99,18 @@ interface Scan<T extends Scan<T>> {
    * @return a table scan that uses the provided executor to access manifests
    */
   T planWith(ExecutorService executorService);
+
+  /**
+   * Returns this scan's projection {@link Schema}.
+   * <p>
+   * If the projection schema was set directly using {@link #project(Schema)}, returns that schema.
+   * <p>
+   * If the projection schema was set by calling {@link #select(Collection)}, returns a projection
+   * schema that includes the selected data fields and any fields used in the filter expression.
+   *
+   * @return this scan's projection schema
+   */
+  Schema schema();
 
   /**
    * Plan the {@link FileScanTask files} that will be read by this scan.
--- a/api/src/main/java/org/apache/iceberg/events/IncrementalScanEvent.java
+++ b/api/src/main/java/org/apache/iceberg/events/IncrementalScanEvent.java
@@ -63,7 +63,7 @@ public final class IncrementalScanEvent {
     return projection;
   }
 
-  public boolean fromSnapshotInclusive() {
+  public boolean isFromSnapshotInclusive() {
     return fromSnapshotInclusive;
   }
 }
--- a/api/src/main/java/org/apache/iceberg/events/IncrementalScanEvent.java
+++ b/api/src/main/java/org/apache/iceberg/events/IncrementalScanEvent.java
@@ -31,14 +31,16 @@ public final class IncrementalScanEvent {
   private final long toSnapshotId;
   private final Expression filter;
   private final Schema projection;
+  private final boolean fromSnapshotInclusive;
 
   public IncrementalScanEvent(String tableName, long fromSnapshotId, long toSnapshotId, Expression filter,
-                              Schema projection) {
+                              Schema projection, boolean fromSnapshotInclusive) {
     this.tableName = tableName;
     this.fromSnapshotId = fromSnapshotId;
     this.toSnapshotId = toSnapshotId;
     this.filter = filter;
     this.projection = projection;
+    this.fromSnapshotInclusive = fromSnapshotInclusive;
   }
 
   public String tableName() {
@@ -59,5 +61,9 @@ public final class IncrementalScanEvent {
 
   public Schema projection() {
     return projection;
+  }
+
+  public boolean fromSnapshotInclusive() {
+    return fromSnapshotInclusive;
   }
 }
--- a/api/src/test/java/org/apache/iceberg/TestHelpers.java
+++ b/api/src/test/java/org/apache/iceberg/TestHelpers.java
@@ -42,6 +42,17 @@ public class TestHelpers {
   private TestHelpers() {
   }
 
+  /**
+   * Wait in a tight check loop until system clock is past {@code timestampMillis}
+   */
+  public static long waitUntilAfter(long timestampMillis) {
+    long current = System.currentTimeMillis();
+    while (current <= timestampMillis) {
+      current = System.currentTimeMillis();
+    }
+    return current;
+  }
+
   public static <T> T assertAndUnwrap(Expression expr, Class<T> expected) {
     Assert.assertTrue("Expression should have expected type: " + expected,
         expected.isInstance(expr));
--- a/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
+++ b/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
@@ -53,14 +53,7 @@ class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implemen
     Snapshot startingSnapshotInclusive = table().snapshot(fromSnapshotId);
     Preconditions.checkArgument(startingSnapshotInclusive != null,
         "Cannot find the starting snapshot: %d", fromSnapshotId);
-    // To achieve inclusive behavior, we pass the parent snapshot id to the TableScanContext.
-    // If parent id is null, leave the TableScanContext#fromSnapshotId as null.
-    // This way, TableScanContext#fromSnapshotId always capture the exclusive behavior.
-    if (startingSnapshotInclusive.parentId() != null) {
-      return newRefinedScan(tableOps(), table(), schema(),
-          context().fromSnapshotId(startingSnapshotInclusive.parentId()));
-    }
-    return this;
+    return newRefinedScan(tableOps(), table(), schema(), context().fromSnapshotIdInclusive(fromSnapshotId));
   }
 
   @Override
@@ -79,26 +72,50 @@ class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implemen
 
   @Override
   public CloseableIterable<FileScanTask> planFiles() {
-    // Return an empty iterable when table is empty
-    if (table().currentSnapshot() == null) {
-      return CloseableIterable.empty();
+    long toSnapshotIdInclusive;
+    if (context().toSnapshotId() != null) {
+      toSnapshotIdInclusive = context().toSnapshotId();
+    } else if (table().currentSnapshot() != null) {
+      toSnapshotIdInclusive = table().currentSnapshot().snapshotId();
+    } else {
+      throw new IllegalArgumentException("End snapshot is not set and table has no current snapshot");
     }
 
-    long toSnapshotIdInclusive = context().toSnapshotId() != null ?
-        context().toSnapshotId() : table().currentSnapshot().snapshotId();
+
     if (context().fromSnapshotId() != null) {
       Preconditions.checkArgument(
           SnapshotUtil.isAncestorOf(table(), toSnapshotIdInclusive, context().fromSnapshotId()),
-          "Starting snapshot (exclusive) %d is not an ancestor of end snapshot %d",
+          "Starting snapshot (%s) %s is not an ancestor of end snapshot %s",
+          context().fromSnapshotInclusive() ? "inclusive" : "exclusive",
           context().fromSnapshotId(), toSnapshotIdInclusive);
     }
 
+    Long fromSnapshotIdExclusive = context().fromSnapshotId();
+    if (context().fromSnapshotId() != null && context().fromSnapshotInclusive()) {
+      // for inclusive behavior fromSnapshotIdExclusive is set to the parent snapshot id, which can be null.
+      // appendsBetween handles null fromSnapshotIdExclusive properly by finding the ancestors of end snapshot.
+      fromSnapshotIdExclusive = table().snapshot(context().fromSnapshotId()).parentId();
+    }
+
+    notifyIncrementalScanEvent(table(), context(), toSnapshotIdInclusive);
+
     // appendsBetween handles null fromSnapshotId (exclusive) properly
-    List<Snapshot> snapshots = appendsBetween(table(), context().fromSnapshotId(), toSnapshotIdInclusive);
+    List<Snapshot> snapshots = appendsBetween(table(), fromSnapshotIdExclusive, toSnapshotIdInclusive);
     if (snapshots.isEmpty()) {
       return CloseableIterable.empty();
     }
 
+    return planFilesFromSnapshots(snapshots);
+  }
+
+  @Override
+  public CloseableIterable<CombinedScanTask> planTasks() {
+    CloseableIterable<FileScanTask> fileScanTasks = planFiles();
+    CloseableIterable<FileScanTask> splitFiles = TableScanUtil.splitFiles(fileScanTasks, targetSplitSize());
+    return TableScanUtil.planTasks(splitFiles, targetSplitSize(), splitLookback(), splitOpenFileCost());
+  }
+
+  private CloseableIterable<FileScanTask> planFilesFromSnapshots(List<Snapshot> snapshots) {
     Set<Long> snapshotIds = Sets.newHashSet(Iterables.transform(snapshots, Snapshot::snapshotId));
     Set<ManifestFile> manifests = FluentIterable
         .from(snapshots)
@@ -121,11 +138,6 @@ class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implemen
       manifestGroup = manifestGroup.ignoreResiduals();
     }
 
-    long fromSnapshotId = resolveFromSnapshotIdForListenerEvent(
-        table(), context().fromSnapshotId(), toSnapshotIdInclusive);
-    Listeners.notifyAll(new IncrementalScanEvent(table().name(), fromSnapshotId,
-        toSnapshotIdInclusive, context().rowFilter(), table().schema()));
-
     if (manifests.size() > 1 &&
         (DataTableScan.PLAN_SCANS_WITH_WORKER_POOL || context().planWithCustomizedExecutor())) {
       manifestGroup = manifestGroup.planWith(context().planExecutor());
@@ -134,28 +146,21 @@ class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implemen
     return manifestGroup.planFiles();
   }
 
-  @Override
-  public CloseableIterable<CombinedScanTask> planTasks() {
-    CloseableIterable<FileScanTask> fileScanTasks = planFiles();
-    CloseableIterable<FileScanTask> splitFiles = TableScanUtil.splitFiles(fileScanTasks, targetSplitSize());
-    return TableScanUtil.planTasks(splitFiles, targetSplitSize(), splitLookback(), splitOpenFileCost());
-  }
-
-  /**
-   * Resolve nullable fromSnapshotIdExclusive to a primitive long type,
-   * as {@link IncrementalScanEvent} only accepts primitive long type.
-   */
-  private static long resolveFromSnapshotIdForListenerEvent(
-      Table table, Long fromSnapshotIdExclusive, long toSnapshotIdInclusive) {
-    if (fromSnapshotIdExclusive != null) {
-      return fromSnapshotIdExclusive;
+  private static void notifyIncrementalScanEvent(
+      Table table, TableScanContext context, long toSnapshotIdInclusive) {
+    long fromSnapshotId;
+    boolean fromSnapshotInclusive;
+    if (context.fromSnapshotId() != null) {
+      fromSnapshotId = context.fromSnapshotId();
+      fromSnapshotInclusive = context.fromSnapshotInclusive();
     } else {
-      // Resolve to the oldest ancestor of toSnapshotIdInclusive.
-      // This doesn't accurately capture the exclusive behavior of fromSnapshotId.
-      // Alternative is a special value (like 0L) to indicate null fromSnapshotIdExclusive.
       Snapshot oldestAncestorSnapshot = SnapshotUtil.oldestAncestor(toSnapshotIdInclusive, table::snapshot);
-      return oldestAncestorSnapshot != null ? oldestAncestorSnapshot.snapshotId() : toSnapshotIdInclusive;
+      fromSnapshotId = oldestAncestorSnapshot.snapshotId();
+      fromSnapshotInclusive = true;
     }
+
+    Listeners.notifyAll(new IncrementalScanEvent(table.name(), fromSnapshotId,
+        toSnapshotIdInclusive, context.rowFilter(), table.schema(), fromSnapshotInclusive));
   }
 
   private static List<Snapshot> appendsBetween(Table table, Long fromSnapshotIdExclusive, long toSnapshotIdInclusive) {
--- a/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
+++ b/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
@@ -101,7 +101,7 @@ class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implemen
         // validate the parent snapshot id an ancestor of toSnapshotId
         Preconditions.checkArgument(
             SnapshotUtil.isParentAncestorOf(table(), toSnapshotIdInclusive, context().fromSnapshotId()),
-            "Starting snapshot (exclusive) %s is not an parent ancestor of end snapshot %s",
+            "Starting snapshot (exclusive) %s is not a parent ancestor of end snapshot %s",
             context().fromSnapshotId(), toSnapshotIdInclusive);
       }
     }
--- a/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
+++ b/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
@@ -29,7 +29,6 @@ import org.apache.iceberg.relocated.com.google.common.collect.FluentIterable;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
-import org.apache.iceberg.util.PropertyUtil;
 import org.apache.iceberg.util.SnapshotUtil;
 import org.apache.iceberg.util.TableScanUtil;
 
@@ -169,26 +168,5 @@ class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implemen
     }
 
     return snapshots;
-  }
-
-  private long targetSplitSize() {
-    long tableValue = tableOps().current().propertyAsLong(
-        TableProperties.SPLIT_SIZE,
-        TableProperties.SPLIT_SIZE_DEFAULT);
-    return PropertyUtil.propertyAsLong(context().options(), TableProperties.SPLIT_SIZE, tableValue);
-  }
-
-  private int splitLookback() {
-    int tableValue = tableOps().current().propertyAsInt(
-        TableProperties.SPLIT_LOOKBACK,
-        TableProperties.SPLIT_LOOKBACK_DEFAULT);
-    return PropertyUtil.propertyAsInt(context().options(), TableProperties.SPLIT_LOOKBACK, tableValue);
-  }
-
-  private long splitOpenFileCost() {
-    long tableValue = tableOps().current().propertyAsLong(
-        TableProperties.SPLIT_OPEN_FILE_COST,
-        TableProperties.SPLIT_OPEN_FILE_COST_DEFAULT);
-    return PropertyUtil.propertyAsLong(context().options(), TableProperties.SPLIT_OPEN_FILE_COST, tableValue);
   }
 }
--- a/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
+++ b/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
@@ -1,0 +1,194 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.util.List;
+import java.util.Set;
+import org.apache.iceberg.events.IncrementalScanEvent;
+import org.apache.iceberg.events.Listeners;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.FluentIterable;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.util.PropertyUtil;
+import org.apache.iceberg.util.SnapshotUtil;
+import org.apache.iceberg.util.TableScanUtil;
+
+class BaseIncrementalAppendScan extends BaseScan<IncrementalAppendScan> implements IncrementalAppendScan {
+
+  BaseIncrementalAppendScan(TableOperations ops, Table table) {
+    this(ops, table, table.schema(), new TableScanContext());
+  }
+
+  BaseIncrementalAppendScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
+    super(ops, table, schema, context);
+  }
+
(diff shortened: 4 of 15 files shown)


core/src/test/java/org/apache/iceberg/TestBaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,121 @@
...
+        .createBranch("branch", snapshotA.snapshotId())
+        .commit();
+
+    table.newFastAppend().appendFile(FILE_B).commit();
+    Snapshot snapshotB = table.currentSnapshot();
+    table.newFastAppend().appendFile(FILE_C).commit();
+    Snapshot snapshotC = table.currentSnapshot();
+
+    // sleep a milli second so that snapshotD will certainly have a different timestamp
+    Thread.sleep(1);
```
rdblue, line 95: We typically use `waitUntilAfter(snapshotC.timestampMillis())` to avoid `Thread.sleep`. It does basically the same thing, but ensures that you have a different timestamp.

core/src/test/java/org/apache/iceberg/TestBaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Test
+  public void testMultipleRootSnapshots() throws Exception {
+    table.newFastAppend().appendFile(FILE_A).commit();
+    Snapshot snapshotA = table.currentSnapshot();
+
+    table.manageSnapshots()
+        .createBranch("branch", snapshotA.snapshotId())
+        .commit();
```
rdblue, line 87: Remove the branch. Add:

```java
long expireA = waitUntilAfter(snapshotA.timestampMillis());
```

Then use `table.newFastAppend().appendFile(FILE_D).stageOnly().commit()` to create the unrelated commit for D. And then create B and C.

And expire A with:

```java
table.expireSnapshots().expireOlderThan(expireA).commit();
```

api/src/test/java/org/apache/iceberg/TestHelpers.java
```diff
@@ -42,6 +42,17 @@ public class TestHelpers {
   private TestHelpers() {
   }
 
+  /**
+   * Wait in a tight check loop until system clock is past {@code timestampMillis}
+   */
+  public static long waitUntilAfter(long timestampMillis) {
```
stevenzwu, line 48: This is copied from `SparkTestBase`

core/src/test/java/org/apache/iceberg/ScanTestBase.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import static org.apache.iceberg.types.Types.NestedField.required;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public abstract class ScanTestBase<T extends Scan<T>> extends TableTestBase {
```
stevenzwu, line 39: This is refactored out of `TestDataTableScan`

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -114,4 +133,19 @@
    * @return an Iterable of tasks for this scan
    */
   CloseableIterable<CombinedScanTask> planTasks();
+
+  /**
+   * Returns the target split size for this scan.
+   */
+  long targetSplitSize();
```
stevenzwu, line 140: I moved those 3 split methods so that we can share the implementations in `BaseScan` class. 

Alternatively, we can move them into a static util class. Would love to get some feedback here.
```
public class ScanUtil {
  public static long targetSplitSize(TableOperations ops, TableScanContext context) {
    long tableValue = ops.current().propertyAsLong(
        TableProperties.SPLIT_SIZE,
        TableProperties.SPLIT_SIZE_DEFAULT);
    return PropertyUtil.propertyAsLong(context.options(), TableProperties.SPLIT_SIZE, tableValue);
  }
}
```

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -84,6 +84,13 @@
    */
   T filter(Expression expr);
 
+  /**
+   * Create a new scan from this that applies data filtering to files but not to rows in those files.
+   *
+   * @return a new scan based on this that does not filter rows in files.
+   */
+  T ignoreResiduals();
```
stevenzwu, line 92: this is missed in the last API PR

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -93,6 +100,18 @@
...
+   * Returns this scan's projection {@link Schema}.
+   * <p>
+   * If the projection schema was set directly using {@link #project(Schema)}, returns that schema.
+   * <p>
+   * If the projection schema was set by calling {@link #select(Collection)}, returns a projection
+   * schema that includes the selected data fields and any fields used in the filter expression.
+   *
+   * @return this scan's projection schema
+   */
+  Schema schema();
```
stevenzwu, line 113: This is moved from `TableScan` to base `Scan` interface as it is used by `ScanTestBase`.

hameizi:
@stevenzwu This PR is for "append" snapshot, so for V2 table format we will add like "BaseIncrementalOverwriteScan"?

stevenzwu:
@hameizi there could be a new `IncrementalChangelogScan` interface and an implementation for the more complex CDC like read.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,172 @@
...
+        table(), context().fromSnapshotId(), toSnapshotIdInclusive);
+    Listeners.notifyAll(new IncrementalScanEvent(table().name(), fromSnapshotId,
+        toSnapshotIdInclusive, context().rowFilter(), table().schema()));
+
+    if (manifests.size() > 1 &&
+        (DataTableScan.PLAN_SCANS_WITH_WORKER_POOL || context().planWithCustomizedExecutor())) {
+      manifestGroup = manifestGroup.planWith(context().planExecutor());
+    }
+
+    return manifestGroup.planFiles();
```
hameizi, line 134: `manifestGroup.planFiles()` will apply all delete-file whose seq number greater than datafile even though the delete file is not belong to the snapshots we scan, is it the intended effect?

stevenzwu: This impl works with appends only table. For row-level deletes, we would need another `IncrementalChangelogScan` interface and implementation.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,172 @@
...
+  private static long resolveFromSnapshotIdForListenerEvent(
+      Table table, Long fromSnapshotIdExclusive, long toSnapshotIdInclusive) {
+    if (fromSnapshotIdExclusive != null) {
+      return fromSnapshotIdExclusive;
+    } else {
+      // Resolve to the oldest ancestor of toSnapshotIdInclusive.
+      // This doesn't accurately capture the exclusive behavior of fromSnapshotId.
+      // Alternative is a special value (like 0L) to indicate null fromSnapshotIdExclusive.
+      Snapshot oldestAncestorSnapshot = SnapshotUtil.oldestAncestor(toSnapshotIdInclusive, table::snapshot);
+      return oldestAncestorSnapshot != null ? oldestAncestorSnapshot.snapshotId() : toSnapshotIdInclusive;
```
rdblue, line 157: Can we also add an inclusive/exclusive flag to the event so that we are correctly labelling this case?

rdblue, line 157: Should we check that `oldestAncestorSnapshot.parentId() == fromSnapshotIdExclusive`?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,172 @@
...
+
+  @Override
+  public CloseableIterable<FileScanTask> planFiles() {
+    // Return an empty iterable when table is empty
+    if (table().currentSnapshot() == null) {
+      return CloseableIterable.empty();
+    }
+
+    long toSnapshotIdInclusive = context().toSnapshotId() != null ?
+        context().toSnapshotId() : table().currentSnapshot().snapshotId();
```
rdblue, line 88: With the `currentSnapshot()` check removed above, you'll need to update this:

```java
long toSnapshotIdInclusive;
if (context.toSnapshotId() != null) {
  toSnapshotIdInclusive = context.toSnapshotId();
} else if (table.currentSnapshot() != null) {
  toSnapshotIdInclusive = table.currentSnapshot().snapshotId();
} else {
  // throw exception
}
```

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,172 @@
...
+    if (context().fromSnapshotId() != null) {
+      Preconditions.checkArgument(
+          SnapshotUtil.isAncestorOf(table(), toSnapshotIdInclusive, context().fromSnapshotId()),
+          "Starting snapshot (exclusive) %d is not an ancestor of end snapshot %d",
+          context().fromSnapshotId(), toSnapshotIdInclusive);
+    }
```
rdblue, line 94: If the `fromSnapshotId` is null then we need to validate that the oldest ancestor's parent is null.

rdblue: Here's a case I'm thinking of:

```
 A - B - C <-- current
 D - E <-- branch
```

If D's parent is null and the `toSnapshotId` is C. If the from snapshot is D, then the exclusive one will be null, D's parent. But D is not in C's ancestors.

rdblue: Probably need to add an inclusive/exclusive flag to context so that we keep track of the requested starting snapshot, D.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -107,6 +107,19 @@ public static Snapshot oldestAncestor(Table table) {
     return lastSnapshot;
   }
 
+  /**
+   * Traverses the history and finds the oldest ancestor of snapshotId.
+   * @return null if there is no current snapshot in the table, else the oldest Snapshot.
```
rdblue, line 112: Also needs a `@param` for `snapshotId` and `lookup`.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -107,6 +107,19 @@ public static Snapshot oldestAncestor(Table table) {
...
+   */
+  public static Snapshot oldestAncestor(long snapshotId, Function<Long, Snapshot> lookup) {
+    Snapshot lastSnapshot = null;
+    for (Snapshot snapshot : ancestorsOf(snapshotId, lookup)) {
```
rdblue, line 116: Be aware that this includes the snapshot for `snapshotId`.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -107,6 +107,19 @@ public static Snapshot oldestAncestor(Table table) {
     return lastSnapshot;
   }
 
+  /**
+   * Traverses the history and finds the oldest ancestor of snapshotId.
```
rdblue, line 111: Note that this may return `snapshotId` if that snapshot is the oldest known snapshot (its parent is null or has been expired).

core/src/test/java/org/apache/iceberg/TestBaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Test
+  public void testMultipleRootSnapshots() throws Exception {
+    table.newFastAppend().appendFile(FILE_A).commit();
+    long snapshotAId = table.currentSnapshot().snapshotId();
+    long expireTimestampSnapshotA = TestHelpers.waitUntilAfter(table.currentSnapshot().timestampMillis());
+
+    // append file B in a staged branch
+    table.newFastAppend().appendFile(FILE_B).stageOnly().commit();
```
rdblue, line 90: You can also call the operation's `apply()` method to get the snapshot ID early.

```java
AppendFiles append = table.newFastAppend()...;
Snapshot snapshotWithB = append.apply();
append.commit();
```

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,172 @@
...
+    if (table().currentSnapshot() == null) {
+      return CloseableIterable.empty();
+    }
+
+    long toSnapshotIdInclusive = context().toSnapshotId() != null ?
+        context().toSnapshotId() : table().currentSnapshot().snapshotId();
+    if (context().fromSnapshotId() != null) {
+      Preconditions.checkArgument(
+          SnapshotUtil.isAncestorOf(table(), toSnapshotIdInclusive, context().fromSnapshotId()),
+          "Starting snapshot (exclusive) %d is not an ancestor of end snapshot %d",
```
rdblue, line 92: I think %s would fill in the values.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -53,14 +53,7 @@ public IncrementalAppendScan fromSnapshotInclusive(long fromSnapshotId) {
...
         "Cannot find the starting snapshot: %d", fromSnapshotId);
-    // To achieve inclusive behavior, we pass the parent snapshot id to the TableScanContext.
-    // If parent id is null, leave the TableScanContext#fromSnapshotId as null.
-    // This way, TableScanContext#fromSnapshotId always capture the exclusive behavior.
-    if (startingSnapshotInclusive.parentId() != null) {
-      return newRefinedScan(tableOps(), table(), schema(),
-          context().fromSnapshotId(startingSnapshotInclusive.parentId()));
-    }
-    return this;
+    return newRefinedScan(tableOps(), table(), schema(), context().fromSnapshotIdInclusive(fromSnapshotId));
```
stevenzwu, line 56: this was a bug that was caught when adding a new test coverage

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -79,26 +72,50 @@ public IncrementalAppendScan toSnapshot(long toSnapshotId) {
...
+    notifyIncrementalScanEvent(table(), context(), toSnapshotIdInclusive);
+
     // appendsBetween handles null fromSnapshotId (exclusive) properly
-    List<Snapshot> snapshots = appendsBetween(table(), context().fromSnapshotId(), toSnapshotIdInclusive);
+    List<Snapshot> snapshots = appendsBetween(table(), fromSnapshotIdExclusive, toSnapshotIdInclusive);
     if (snapshots.isEmpty()) {
       return CloseableIterable.empty();
     }
 
+    return planFilesFromSnapshots(snapshots);
```
stevenzwu, line 108: this method was extracted to avoid checkstyle error
```
[ant:checkstyle] [ERROR] /Users/stevenwu/ws/iceberg/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java:73:3: Cyclomatic Complexity is 13 (max allowed is 12). [CyclomaticComplexity]
```

api/src/main/java/org/apache/iceberg/events/IncrementalScanEvent.java
```diff
@@ -60,4 +62,8 @@ public Expression filter() {
   public Schema projection() {
     return projection;
   }
+
+  public boolean fromSnapshotInclusive() {
```
rdblue, line 66: Would it be more clear for this to be `isFromSnapshotInclusive`?

stevenzwu: agree. updated

aokolnychyi:
I'd be interested to take a look today as well.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+
+  @Override
+  public CloseableIterable<FileScanTask> planFiles() {
+    long toSnapshotIdInclusive;
+    if (context().toSnapshotId() != null) {
+      toSnapshotIdInclusive = context().toSnapshotId();
+    } else if (table().currentSnapshot() != null) {
+      toSnapshotIdInclusive = table().currentSnapshot().snapshotId();
+    } else {
+      throw new IllegalArgumentException("End snapshot is not set and table has no current snapshot");
```
rdblue, line 81: I'm not sure this is the right behavior. If you try to scan an empty table, then it should return 0 tasks instead of failing.

I think the case where this should fail is when the from snapshot ID is set and there is no to snapshot.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+
+    if (context().fromSnapshotId() != null) {
+      Preconditions.checkArgument(
+          SnapshotUtil.isAncestorOf(table(), toSnapshotIdInclusive, context().fromSnapshotId()),
```
rdblue, line 87: This doesn't handle the exclusive case by checking parent ID, so I think you need to loop over the snapshots and check parent ID.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+          "Starting snapshot (%s) %s is not an ancestor of end snapshot %s",
+          context().fromSnapshotInclusive() ? "inclusive" : "exclusive",
+          context().fromSnapshotId(), toSnapshotIdInclusive);
+    }
+
+    Long fromSnapshotIdExclusive = context().fromSnapshotId();
```
rdblue, line 93: This calls `fromSnapshotId` in a lot of places. Can you store that in a variable?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+    if (context().fromSnapshotId() != null && context().fromSnapshotInclusive()) {
+      // for inclusive behavior fromSnapshotIdExclusive is set to the parent snapshot id, which can be null.
+      // appendsBetween handles null fromSnapshotIdExclusive properly by finding the ancestors of end snapshot.
+      fromSnapshotIdExclusive = table().snapshot(context().fromSnapshotId()).parentId();
+    }
+
+    notifyIncrementalScanEvent(table(), context(), toSnapshotIdInclusive);
```
rdblue, line 100: I think this should either pass the information for the event, or pass the context, but not both. It's unclear where information should come from and we've already done the analysis to know the starting snapshot here.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private static void notifyIncrementalScanEvent(
+      Table table, TableScanContext context, long toSnapshotIdInclusive) {
+    long fromSnapshotId;
+    boolean fromSnapshotInclusive;
+    if (context.fromSnapshotId() != null) {
+      fromSnapshotId = context.fromSnapshotId();
+      fromSnapshotInclusive = context.fromSnapshotInclusive();
+    } else {
+      Snapshot oldestAncestorSnapshot = SnapshotUtil.oldestAncestor(toSnapshotIdInclusive, table::snapshot);
```
rdblue, line 157: Looks like this is redoing work that has already been done in `planFiles`.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
rdblue: t0 : A - B - C
t1 : B - C
scan( A, C, exclusive )

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+  @Override
+  public IncrementalAppendScan toSnapshot(long toSnapshotId) {
+    Preconditions.checkArgument(table().snapshot(toSnapshotId) != null,
+        "Cannot find end snapshot: %d", toSnapshotId);
+    return newRefinedScan(tableOps(), table(), schema(),  context().toSnapshotId(toSnapshotId));
+  }
+
+  @Override
+  public CloseableIterable<FileScanTask> planFiles() {
+    long toSnapshotIdInclusive;
```
rdblue, line 75: Add `Long fromSnapshotId = context().fromSnapshotId()`

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+      fromSnapshotInclusive = true;
+    }
+
+    Listeners.notifyAll(new IncrementalScanEvent(table.name(), fromSnapshotId,
+        toSnapshotIdInclusive, context.rowFilter(), table.schema(), fromSnapshotInclusive));
+  }
+
+  private static List<Snapshot> appendsBetween(Table table, Long fromSnapshotIdExclusive, long toSnapshotIdInclusive) {
+    List<Snapshot> snapshots = Lists.newArrayList();
+    for (Snapshot snapshot : SnapshotUtil.ancestorsBetween(
```
rdblue, line 170: Comment: this does not validation because `planFiles` validated the from snapshot and to snapshot range.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @Override
+  public CloseableIterable<CombinedScanTask> planTasks() {
+    CloseableIterable<FileScanTask> fileScanTasks = planFiles();
+    CloseableIterable<FileScanTask> splitFiles = TableScanUtil.splitFiles(fileScanTasks, targetSplitSize());
+    return TableScanUtil.planTasks(splitFiles, targetSplitSize(), splitLookback(), splitOpenFileCost());
+  }
+
+  private CloseableIterable<FileScanTask> planFilesFromSnapshots(List<Snapshot> snapshots) {
```
rdblue, line 120: Can we have a more specific name, like `appendFilesFromSnapshots`?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: good point. added a unit test to cover this scenario

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  @Override
+  protected IncrementalAppendScan newRefinedScan(
+      TableOperations newOps, Table newTable, Schema newSchema, TableScanContext newContext) {
+    return new BaseIncrementalAppendScan(newOps, newTable, newSchema, newContext);
+  }
+
+  @Override
+  public IncrementalAppendScan fromSnapshotInclusive(long fromSnapshotId) {
```
aokolnychyi, line 52: How do we handle schema evolution in incremental scans? Do we always pick the current schema? Do we test it anywhere?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,179 @@
...
+    this(ops, table, table.schema(), new TableScanContext());
+  }
+
+  BaseIncrementalAppendScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
+    super(ops, table, schema, context);
+  }
+
+  @Override
+  protected IncrementalAppendScan newRefinedScan(
+      TableOperations newOps, Table newTable, Schema newSchema, TableScanContext newContext) {
```
aokolnychyi, line 47: nit: arg formatting

core/src/main/java/org/apache/iceberg/TableScanContext.java
```diff
@@ -168,7 +172,17 @@ Long fromSnapshotId() {
   TableScanContext fromSnapshotId(long id) {
     return new TableScanContext(snapshotId, rowFilter, ignoreResiduals,
         caseSensitive, colStats, projectedSchema, selectedColumns, options, id, toSnapshotId,
-        planExecutor);
+        planExecutor, fromSnapshotInclusive);
+  }
+
+  TableScanContext fromSnapshotIdInclusive(long id) {
```
aokolnychyi, line 178: Shall we rename the existing methods (`fromSnapshotId`) to indicate they configure the start snapshot exclusive?
We may also set `fromSnapshotInclusive` explicitly in existing methods.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,179 @@
...
+      toSnapshotIdInclusive = context().toSnapshotId();
+    } else if (table().currentSnapshot() != null) {
+      toSnapshotIdInclusive = table().currentSnapshot().snapshotId();
+    } else if (context().fromSnapshotId() != null) {
+      throw new IllegalArgumentException(
+          "Invalid config: end snapshot is not set, start snapshot is set, and table has no current snapshot");
+    } else {
+      // It is an empty table (no current snapshot). Both from and to snapshots aren't set either.
+      // In this case, listener notification is also skipped
+      return CloseableIterable.empty();
```
aokolnychyi, line 85: It is not very obvious to have a return statement for the whole method nested in a block that initializes the end snapshot.

Have you considered a separate method for computing the end snapshot, a call to `Preconditions` and a separate if statement for an early return?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    // fromSnapshotIdExclusive can be null. appendsBetween handles null fromSnapshotIdExclusive properly
+    // by finding the oldest ancestor of end snapshot.
+    Long fromSnapshotIdExclusive = context().fromSnapshotId();
+    if (context().fromSnapshotId() != null) {
```
aokolnychyi, line 91: Looks like this block accesses the same variables from `context` quite a few times. Would it make sense to define temp vars? It may make it a little bit easier to read as there will be fewer method calls.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,179 @@
...
+        // validate the parent snapshot id an ancestor of toSnapshotId
+        Preconditions.checkArgument(
+            SnapshotUtil.isParentAncestorOf(table(), toSnapshotIdInclusive, context().fromSnapshotId()),
+            "Starting snapshot (exclusive) %s is not a parent ancestor of end snapshot %s",
+            context().fromSnapshotId(), toSnapshotIdInclusive);
+      }
+    }
+
+    if (fromSnapshotIdExclusive != null) {
+      Listeners.notifyAll(new IncrementalScanEvent(table().name(), fromSnapshotIdExclusive,
```
aokolnychyi, line 110: nit: The purpose of the true/false flags is not clear without looking into the code. Would it make sense to offer static methods like `fromSnapshotInclusive` and `fromSnapshotExclusive` in `IncrementalScanEvent`?

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -45,70 +38,46 @@
...
   }
 
-  @SuppressWarnings("checkstyle:HiddenField")
-  protected abstract TableScan newRefinedScan(TableOperations ops, Table table, Schema schema,
-                                              TableScanContext context);
-
   protected abstract CloseableIterable<FileScanTask> doPlanFiles();
 
   @Override
   public Table table() {
```
aokolnychyi, line 79: Do we still need this implementation here? Can't we consume from the parent directly?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -77,35 +76,52 @@ public CloseableIterable<FileScanTask> planFiles() {
...
-      throw new IllegalArgumentException("End snapshot is not set and table has no current snapshot");
+      // It is an empty table (no current snapshot). Both from and to snapshots aren't set either.
+      // In this case, listener notification is also skipped
+      return CloseableIterable.empty();
     }
 
-
+    // fromSnapshotIdExclusive can be null. appendsBetween handles null fromSnapshotIdExclusive properly
+    // by finding the oldest ancestor of end snapshot.
+    Long fromSnapshotIdExclusive = context().fromSnapshotId();
```
rdblue, line 90: Can you refactor out all the calls to `context().fromSnapshotId()`? I think it would be more readable to have a variable.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -0,0 +1,179 @@
...
+    Set<Long> snapshotIds = Sets.newHashSet(Iterables.transform(snapshots, Snapshot::snapshotId));
+    Set<ManifestFile> manifests = FluentIterable
+        .from(snapshots)
+        .transformAndConcat(Snapshot::dataManifests)
+        .filter(manifestFile -> snapshotIds.contains(manifestFile.snapshotId()))
+        .toSet();
+
+    ManifestGroup manifestGroup = new ManifestGroup(tableOps().io(), manifests)
+        .caseSensitive(context().caseSensitive())
+        .select(context().returnColumnStats() ? DataTableScan.SCAN_WITH_STATS_COLUMNS : DataTableScan.SCAN_COLUMNS)
```
aokolnychyi, line 144: This block also accesses a lot of information on `context()`. There were some helper methods in `BaseTableScan`. What about moving them to `BaseScan` and using here?

```
  protected Long snapshotId() {
    return context().snapshotId();
  }

  protected boolean colStats() {
    return context().returnColumnStats();
  }

  protected boolean shouldIgnoreResiduals() {
    return context().ignoreResiduals();
  }

  @Override
  public Expression filter() {
    return context().rowFilter();
  }

  @Override
  public boolean isCaseSensitive() {
    return context().caseSensitive();
  }

  protected ExecutorService planExecutor() {
    return context().planExecutor();
  }
```

rdblue (approved):

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -45,70 +38,46 @@
...
-    return context.returnColumnStats();
+    return context().returnColumnStats();
   }
 
   protected boolean shouldIgnoreResiduals() {
-    return context.ignoreResiduals();
+    return context().ignoreResiduals();
   }
 
   protected Collection<String> selectedColumns() {
```
aokolnychyi, line 64: Is this still being used?

rdblue:
Thanks, @stevenzwu!

aokolnychyi:
@rdblue @stevenzwu, I had some minor comments about styling in a few places. They don't affect correctness but it would be great to fix some of those in a follow-up.

rdblue:
@aokolnychyi, Sorry, I didn't realize you were reviewing at the same time. I think a follow-up is a good idea, I had a comment as well. I just wanted to get this in to unblock the rest since I think it's correct.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: Good question. I think it is outside the scope of the scan/source. 

Let's use the more common streaming source (kafka with Avro record). The problem of schema evolution is outside the scope of Kafka source. If the schema is evolved in a backward compatible, the consumers won't break. If the schema evolution is incompatible, then consumers should be notified and the breaking change need to be coordinated. If there is a Avro schema registry, notification service can be provided by schema registry or some other control plane.

Now coming to Iceberg streaming source, I would assume similar expectation. maybe some control plane service can notify consumers of table schema change. Then consumers can take proper action if needed

core/src/main/java/org/apache/iceberg/TableScanContext.java
stevenzwu: agree. since this is a non-public class, it is safe to rename the old setter of `fromSnapshotId` to `fromSnapshotIdExclusive`

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: sounds good. will make the change

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: @aokolnychyi agree with your comment that boolean flag is not clear without looking into the code. Just static methods like `fromSnapshotInclusive` and `fromSnapshotExclusive` is also a little odd to me. Do you think we should provide a builder class instead? Then in the builder class, we can provide methods like `fromSnapshotInclusive` and `fromSnapshotExclusive` methods.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
stevenzwu: yeah. `BaseScan#table()` is a protected method. this is a public method.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
stevenzwu: doesn't looks like it is used based on code search. will remove it in the follow up

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: I am not sure adding a bunch help methods like these makes the code simpler

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: addressed in PR #4886

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
aokolnychyi: As a source, which schema do we expect Iceberg to use? What if I scan for appends and some earlier snapshots doesn't have a column? Or what if earlier snapshots have an extra column? It is a tricky use case but I just want to understand our take on this and what we expect from Iceberg and what we expect from the user.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
aokolnychyi: I mentioned separate methods as we usually prefer a named method to give some meaning to a boolean flag but I agree there are quite some args in this case so such it is questionable. Builder sounds good but this point is minor. Up to you. I think we can leave as-is and don't worry about it.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
aokolnychyi: These have been already defines in `BaseTableScan` for most common methods as we have like 10+ scans so I was just thinking about moving them to `BaseScan`. Up to you, though. I don't feel strongly.

stevenzwu: I missed that `BaseTableScan` already has many helper methods like these. I am ok to move the protected helper method to `BaseScan`.  I am not opinionated either way.. A couple of the helper methods (like `filter` and `isCaseSensitive`) from `BaseTableScan` are public and have to stay.

@rdblue do you have any preference?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: I like the builder slightly better. but as you said, this is minor. We can leave as-is for now.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: In Kafka source case, users can pin the read Avro schema. As long as  schema evolution is backward compatible (like an extra column), this is fine. If users don't pin read schema, only Avro GenericRecord is possible with a schema registry. Kafka record contains a schema id or fingerprint that consumers can use to look up schema from registry. SpecificRecord implicitly pinned the read schema to the schema used to generate the SpecificRecord code. 

Similarly with Flink Iceberg source (as an example), users can set read/projected schema. Then the Flink reader should emit `RowData` with fields only from the pinned read schema. if users don't pin the read schema, current table schema is used as read schema. Read schema stays the same until the next time the Flink job is redeployed and then read schema is refreshed to the table schema again. 

If a mandatory field is deleted (incompatible schema change), then consumer apps may be broken if the deleted column is accessed. 

Here is probably a more tricky scenario that would fail
- users don't pin read schema
- a new required column is added to Iceberg table schema
- read scanned some old snapshots/data where is new required column is not present

Either users should pin read schema or users need to make sure not to scan old data where current table is incompatible. I would think adding a required column is also an incompatible schema evolution as it is incompatible to older data.

cc @rdblue @kbendick @Fokko for more inputs

aokolnychyi: Thanks for the context. I think the current behavior in the incremental scan matches the expectation.

> I would think adding a required column is also an incompatible schema evolution

That's true. Iceberg doesn't allow adding required columns by default unless the user explicitly acknowledged it would not break older data files.

PR #4747, opened 2022-05-11
Core: Fix data loss when duplicate snapshot-id is allocated

ajantha-bhat:
There is astronomically low probability of allocating same snapshot id. 
But as we are converting 128 bit UUID to 64 bit uuid, snapshot id is not as unique as UUID.

If the duplicate snapshot ids are allocated, Iceberg doesn't handle it and it leads to the data loss.

TODO: This PR only reproduces issue. Yet to fix.
I am not sure whether to throw exception during commit or do metadata retry with new snapshot id.


core/src/main/java/org/apache/iceberg/SnapshotIdGeneratorUtil.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.util.UUID;
+
+public final class SnapshotIdGeneratorUtil {
```
ajantha-bhat, line 24: I just moved snapshot id generation logic to static method, so that I can use MockStatic. 

Normal object level mock or spy is impossible to use for an end to end SQL testcase as `Table Operations` objects keeps on changing. Hence, static using static mock.

rdblue:
Looks like what's happening is that the snapshot is never added. It hits this check: https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/TableMetadata.java#L984-L985

I think we can fix this by changing `SnapshotProducer`:

```diff
+ while (snapshotId == null || ops.current().snapshot(snapshotId) != null) {
- if (snapshotId != null) {
    this.snapshotId = ops.newSnapshotId();
  }
```

ajantha-bhat:
> I think we can fix this by changing SnapshotProducer:

@rdblue: Fixed it as per your suggestion. I too expected this kind of check initially.

Should we do this check at the commit time and throw exception? Now, concurrent table operation getting (both are uncommitted) getting the same snapshot id is not handled I guess.

ajantha-bhat:
> Now, concurrent table operation getting (both are uncommitted) getting the same snapshot id is not handled I guess.

Handled concurrent scenario by throwing validation exception in TableMetadata preparation.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -980,11 +980,15 @@ public Builder addSortOrder(SortOrder order) {
...
     public Builder addSnapshot(Snapshot snapshot) {
-      if (snapshot == null || snapshotsById.containsKey(snapshot.snapshotId())) {
+      if (snapshot == null) {
         // change is a noop
         return this;
       }
 
+      ValidationException.check(!snapshotsById.containsKey(snapshot.snapshotId()),
+          "Snapshot already exist for the id: %s",
+          snapshot.snapshotId());
```
rdblue, line 990: @RussellSpitzer, what do you think about this? Any situations where you think this would cause an issue because we add the same snapshot and rely on the operation being idempotent?

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -980,11 +980,15 @@ public Builder addSortOrder(SortOrder order) {
...
 
     public Builder addSnapshot(Snapshot snapshot) {
-      if (snapshot == null || snapshotsById.containsKey(snapshot.snapshotId())) {
+      if (snapshot == null) {
         // change is a noop
         return this;
       }
 
+      ValidationException.check(!snapshotsById.containsKey(snapshot.snapshotId()),
+          "Snapshot already exist for the id: %s",
```
rdblue, line 989: `exist` -> `exists`

RussellSpitzer:
I still think the probability of collision here is probably worse than serial lottery wins statistically, but this should be a relatively cheap check so It's probably fine to do

ajantha-bhat:
@rdblue, @RussellSpitzer:  Can we merge this PR?

ajantha-bhat:
@rdblue , @RussellSpitzer : Ping...

rdblue (approved):

PR #4748, opened 2022-05-11
Core: optimize planning for position delete matching

chenjunjiedada:
In the case of the Flink upsert job, the position delete contains the same sequence number as its referred data files. In terms of that, the planning could be optimized by checking the equality of sequence numbers.

The situation may not fit other cases, so this PR adds a table property.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -340,4 +340,8 @@ public class TableProperties {
 
   public static final String UPSERT_ENABLED = "write.upsert.enabled";
   public static final boolean UPSERT_ENABLED_DEFAULT = false;
+
+  public static final String POS_DELETE_HAS_SAME_SEQ_WITH_REFS_ENABLED =
+      "write.pos-delete-has-same-seq-with-refs.enabled";
+  public static final boolean POS_DELETE_HAS_SAME_SEQ_WITH_REFS_ENABLED_DEFAULT = false;
 }
--- a/core/src/main/java/org/apache/iceberg/DataTableScan.java
+++ b/core/src/main/java/org/apache/iceberg/DataTableScan.java
@@ -86,6 +86,12 @@ public class DataTableScan extends BaseTableScan {
         .specsById(table().specs())
         .ignoreDeleted();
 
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -340,4 +340,8 @@ private TableProperties() {
 
   public static final String UPSERT_ENABLED = "write.upsert.enabled";
   public static final boolean UPSERT_ENABLED_DEFAULT = false;
+
+  public static final String POS_DELETE_HAS_SAME_SEQ_WITH_REFS_ENABLED =
+      "write.pos-delete-has-same-seq-with-refs.enabled";
```
chenjunjiedada, line 345: I haven't figured out a shorter name.

chenjunjiedada:
@aokolnychyi , Could you please take a look? Can spark merge into produces different sequence numbers in the MOR mode?

aokolnychyi:
@chenjunjiedada, what do you think is the root cause in this case? Is it that our matching logic is suboptimal and assigns delete files that actually don't apply (I believe we have partition + min/max stats on file path) or is the job planning slow as we need to open more manifests?

aokolnychyi:
> Can spark merge into produces different sequence numbers in the MOR mode?

Spark will write only position delete files and they will have the current table sequence number but apply to data files produced by previous snapshots.

chenjunjiedada:
@aokolnychyi , Flink writer only produces position deletes that apply the data files in the same snapshot. The current planning strategy doesn't take advantage of this assumption. Spark creates position deletes that apply to data files in the previous snapshot so we can also add an option to ignore deletes with the same sequence number as data files.

chenjunjiedada:
@aokolnychyi, After applying this option the matching deletes of a data file could be significantly reduced and thus improve the read performance. The unit test in this PR is an example.

core/src/main/java/org/apache/iceberg/TableProperties.java
hililiwei: Can we turn it on automatically by conditions like whether it's upsert or not?

chenjunjiedada:
Replaced by https://github.com/apache/iceberg/pull/5760.

PR #4751, opened 2022-05-12
Core: Add the API for multiple table locations.

jfz:
This new API expose multiple locations so that the data writer can choose the location.
Writing data to multiple locations is desired in some cases, e.g.
- better data locality when data is ingested from multiple data centers 
- avoid hotspotting to underlying storage system


jzhuge:
ObjectStoreLocationProvider has the feature to avoid throttling. Is it possible to add region awareness?

jfz:
> 

This may be a good idea, let me look into that

PR #4752, opened 2022-05-12
Flink: Get the CatalogTable directly instead of reflecting

hililiwei:
Get the `CatalogTable` directly instead of reflecting
It was originally part of #4246, but I take it out separately this time, because it has no correlation with #4246 and is easier to solve.

cc @kbendick


stevenzwu (approved):

wuwenchi (approved):

felixYyu:
This is more concise, I think if correct other version(v1.13 v1.14) should be also changed.

hililiwei:
> This is more concise, I think if correct other version(v1.13 v1.14) should be also changed.

Yes, agree with you. However, I prefer one PR to focus on issues in the main version and another PR to migrate changes to other versions.  Of course, this change is minor, I can also include the modifications for all versions. 😄

rdblue (approved):

kbendick:
Thanks @hililiwei!

PR #4758, opened 2022-05-12
Spark 3.2: Avoid reflection to load metadata tables in SparkTableUtil

aokolnychyi:
This logic removes the unnecessary reflection to load metadata tables in `SparkTableUtil` since we have separate modules for Spark versions.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
@@ -602,6 +602,17 @@ public class SparkTableUtil {
         .run(item -> io.deleteFile(item.path()));
   }
 
+  /**
+   * Loads a metadata table.
+   *
+   * @deprecated since 0.14.0, will be removed in 0.15.0;
+   *             use {@link #loadMetadataTable(SparkSession, Table, MetadataTableType)}.
+   */
+  @Deprecated
+  public static Dataset<Row> loadCatalogMetadataTable(SparkSession spark, Table table, MetadataTableType type) {
+    return loadMetadataTable(spark, table, type);
+  }
+
   public static Dataset<Row> loadMetadataTable(SparkSession spark, Table table, MetadataTableType type) {
     return loadMetadataTable(spark, table, type, ImmutableMap.of());
   }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
@@ -37,11 +37,11 @@ import org.apache.iceberg.ManifestFile;
 import org.apache.iceberg.ManifestFiles;
 import org.apache.iceberg.ManifestWriter;
 import org.apache.iceberg.MetadataTableType;
+import org.apache.iceberg.MetadataTableUtils;
 import org.apache.iceberg.MetricsConfig;
 import org.apache.iceberg.PartitionSpec;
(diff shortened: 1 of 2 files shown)


aokolnychyi:
cc @RussellSpitzer @flyrain @karuppayya @szehon-ho @rdblue

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -599,46 +602,15 @@ private static void deleteManifests(FileIO io, List<ManifestFile> manifests) {
...
-    } else if (tableName.startsWith("hive")) {
-      // Try loading by name as a Hive table without Catalog
-      return dataFrameReader.load(tableName.replaceFirst("hive\\.", "") + "." + type);
-    } else {
-      throw new IllegalArgumentException(String.format(
-          "Cannot find the metadata table for %s of type %s", tableName, type));
-    }
+  public static Dataset<Row> loadMetadataTable(SparkSession spark, Table table, MetadataTableType type,
+                                               Map<String, String> extraOptions) {
+    SparkTable metadataTable = new SparkTable(MetadataTableUtils.createMetadataTableInstance(table, type), false);
```
aokolnychyi, line 611: Looks like constructing `DataSourceV2Relation` directly is the easiest and safest option.
Let me know if I missed a use case where we still need the old code.

aokolnychyi: We had this code in that private method in `Spark3Util`. I just moved it.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -599,46 +602,15 @@ private static void deleteManifests(FileIO io, List<ManifestFile> manifests) {
         .run(item -> io.deleteFile(item.path()));
   }
 
-  // Attempt to use Spark3 Catalog resolution if available on the path
-  private static final DynMethods.UnboundMethod LOAD_METADATA_TABLE = DynMethods.builder("loadMetadataTable")
-      .hiddenImpl("org.apache.iceberg.spark.Spark3Util", SparkSession.class, Table.class, MetadataTableType.class)
-      .orNoop()
-      .build();
-
-  public static Dataset<Row> loadCatalogMetadataTable(SparkSession spark, Table table, MetadataTableType type) {
```
szehon-ho, line 608: Just flagging that this will remove public method, in case anyone may use this method (though I feel there should not be).

szehon-ho (approved):
I just saw this too, great, this is one benefit about splitting the Spark modules.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
aokolnychyi: Good point, I missed it. I wonder why we made it public, though. I'd probably remove the method but I don't mind deprecating and delegating to the one below if folks think this may impact anyone.

flyrain: Yeah, this is a grey area, in which these methods actually are APIs, but not considered as APIs.  Per our community sync, only the public things in the api module are considered as APIs. We need something(e.g. annotation) to mark these methods as APIs.

aokolnychyi: Sounds good to me.

PR #4759, opened 2022-05-12
Make sort strategy an idempotent action

khaitranq:
This PR will make sort rewrite action an idempotent operation, meaning we will skip sorting data if the data files are somewhat already sorted.

First we remove all files that have no overlap with other files as these files are considered as sorted. For the remaining files, rewrite will be applied if
1. There are a certain number of mis-sized data files or
2. Those data files do not have sortedness score (defined in this the below doc) good enough

We use threshold params to control these conditions.

Example: if the Sort strategy is used on a set of files which is ordered by column x and original has files File A (x: 0 - 50), File B ( x: 10 - 40), File C ( x: 30 - 60), File D ( x: 61 - 80), and File E ( x: 81 - 100), this Strategy will attempt to rewrite files A, B and C into File A' (x: 0-20), File B' (x: 21 - 40), File C' (x: 41 - 60) and keep File D and E unchanged.

Proposal doc for the feature:
https://docs.google.com/document/d/1rZUHljNsLn8JqsO5lYElst3F800T8OBnQ5fasUB-fp8/

### DIFF
--- a/core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
+++ b/core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
@@ -19,27 +19,57 @@
 
 package org.apache.iceberg.actions;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.iceberg.DataFile;
+import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.FluentIterable;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.apache.iceberg.util.BinPacking;
+import org.apache.iceberg.util.BinPacking.ListPacker;
+import org.apache.iceberg.util.PropertyUtil;
+import org.apache.iceberg.util.SortStrategyUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 
 /**
  * A rewrite strategy for data files which aims to reorder data with data files to optimally lay them out
  * in relation to a column. For example, if the Sort strategy is used on a set of files which is ordered
- * by column x and original has files File A (x: 0 - 50), File B ( x: 10 - 40) and File C ( x: 30 - 60),
- * this Strategy will attempt to rewrite those files into File A' (x: 0-20), File B' (x: 21 - 40),
- * File C' (x: 41 - 60).
+ * by column x and original has files File A (x: 0 - 50), File B ( x: 10 - 40), File C ( x: 30 - 60),
+ * File D ( x: 61 - 80), and File E ( x: 81 - 100), this Strategy will attempt to rewrite files A, B and C
+ * into File A' (x: 0-20), File B' (x: 21 - 40), File C' (x: 41 - 60) and keep File D and E unchanged.
  * <p>
- * Currently the there is no file overlap detection and we will rewrite all files if {@link SortStrategy#REWRITE_ALL}
- * is true (default: false). If this property is disabled any files that would be chosen by
- * {@link BinPackStrategy} will be rewrite candidates.
+ * When the {@link BinPackStrategy#REWRITE_ALL} flag is, all files are selected for rewrite. Otherwise,
+ * only unsorted files are selected. Rewrite will be applied to those selected files if
+ * 1. There are a certain number of mis-sized data files or
+ * 2. Those data files do not have sortedness score good enough.
  * <p>
- * In the future other algorithms for determining files to rewrite will be provided.
  */
 public abstract class SortStrategy extends BinPackStrategy {
+  private static final Logger LOG = LoggerFactory.getLogger(SortStrategy.class);
+  /**
+   * Rewrites if the ratio of mis-sized files to total files is over this threshold.
+   * The value should be between 0.0 and 1.0
+   */
+  public static final String MIS_SIZED_RATIO_THRESHOLD = "mis-sized-ratio-threshold";
+  public static final double MIS_SIZED_RATIO_THRESHOLD_DEFAULT = 0.05;
 
+  /**
+   * Rewrites if the sortedness score of given files is below this threshold.
+   * The value should be between 0.0 and 1.0
+   */
+  public static final String SORTEDNESS_SCORE_THRESHOLD = "sortedness-score-threshold";
+  public static final double SORTEDNESS_SCORE__THRESHOLD_DEFAULT = 0.95;
+
+  private double misSizedRatioThreshold;
+  private double sortednessScoreThreshold;
   private SortOrder sortOrder;
 
   /**
@@ -72,6 +102,14 @@ public abstract class SortStrategy extends BinPackStrategy {
   public RewriteStrategy options(Map<String, String> options) {
     super.options(options); // Also checks validity of BinPack options
 
+    misSizedRatioThreshold = PropertyUtil.propertyAsDouble(options,
+        MIS_SIZED_RATIO_THRESHOLD,
+        MIS_SIZED_RATIO_THRESHOLD_DEFAULT);
+
+    sortednessScoreThreshold = PropertyUtil.propertyAsDouble(options,
+        SORTEDNESS_SCORE_THRESHOLD,
+        SORTEDNESS_SCORE__THRESHOLD_DEFAULT);
+
     if (sortOrder == null) {
       sortOrder = table().sortOrder();
     }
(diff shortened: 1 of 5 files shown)


hililiwei:
Will this result in more and more files? Because it doesn't seem to merge small files at the target file size we configured.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -19,27 +19,57 @@
...
- * File C' (x: 41 - 60).
+ * by column x and original has files File A (x: 0 - 50), File B ( x: 10 - 40), File C ( x: 30 - 60),
+ * File D ( x: 61 - 80), and File E ( x: 81 - 100), this Strategy will attempt to rewrite files A, B and C
+ * into File A' (x: 0-20), File B' (x: 21 - 40), File C' (x: 41 - 60) and keep File D and E unchanged.
  * <p>
- * Currently the there is no file overlap detection and we will rewrite all files if {@link SortStrategy#REWRITE_ALL}
- * is true (default: false). If this property is disabled any files that would be chosen by
- * {@link BinPackStrategy} will be rewrite candidates.
+ * When the {@link BinPackStrategy#REWRITE_ALL} flag is, all files are selected for rewrite. Otherwise,
+ * only unsorted files are selected. Rewrite will be applied to those selected files if
```
RussellSpitzer, line 50: The sentences here appear to conflict.
Will only unsorted files be selected or will mis-sized data files be selected or both?

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -19,27 +19,57 @@
...
- * In the future other algorithms for determining files to rewrite will be provided.
  */
 public abstract class SortStrategy extends BinPackStrategy {
+  private static final Logger LOG = LoggerFactory.getLogger(SortStrategy.class);
+  /**
+   * Rewrites if the ratio of mis-sized files to total files is over this threshold.
+   * The value should be between 0.0 and 1.0
+   */
+  public static final String MIS_SIZED_RATIO_THRESHOLD = "mis-sized-ratio-threshold";
+  public static final double MIS_SIZED_RATIO_THRESHOLD_DEFAULT = 0.05;
```
RussellSpitzer, line 62: We've generally opposed using ratio's because they tend to have inconsistent behaviors. I'm also not sure why we would want to add this ratio. In my mind we would have either wanted to use the legacy size based threshold or the Sort based threshold. In both cases I believe the thresholds should be absolute.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -80,12 +118,55 @@ public RewriteStrategy options(Map<String, String> options) {
...
+  public Iterable<FileScanTask> selectFilesToRewrite(Iterable<FileScanTask> dataFiles) {
+    if (rewriteAll()) {
+      LOG.info("Table {} set to rewrite all data files", table().name());
+      return dataFiles;
+    } else {
+      // Remove files that are completely sorted.
+      // Example: File_A(1, 10), File_B(11, 25), File_C(15, 30), File_D(31, 40)
+      // Then only File_B and File_C are selected
+      Iterable<FileScanTask> selectedFiles = SortStrategyUtil.removeSortedFiles(dataFiles, sortOrder);
+      if (!haveGoodFileSizes(selectedFiles) || !areFilesSorted(selectedFiles)) {
```
RussellSpitzer, line 131: for both of these clauses I would prefer using positive assertions rather than negations.

In this case
If !(x && y) or better yet if ( notX && notY) where the functions notX and notY themselves are the requirements.

I'll mark this below but "haveGood" is not a very descriptive name

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -80,12 +118,55 @@ public RewriteStrategy options(Map<String, String> options) {
...
+        // Rewrite all selected files if they are mis-sized or have bad sortedness score
+        return selectedFiles;
+      }
+      return ImmutableList.of();
+    }
+  }
+
+  @Override
+  public Iterable<List<FileScanTask>> planFileGroups(Iterable<FileScanTask> dataFiles) {
+    ListPacker<FileScanTask> packer = new BinPacking.ListPacker<>(maxGroupSize(), 1, false);
```
RussellSpitzer, line 141: In We may have a slight issue here as this bin packing algorithm may group together files which are not actually adjacent. For example say I mark files

A, B, X, Y ,Z

Where A and B overlap and X, Y , Z overlap

And then make groups

A,Y,Z and B,X

Sorting and writing B,X will have no effect
And sorting and rewriting a,y,z will just create new files which potentially overlap with B and X

So in this case we probably need to sort our files or cluster them together based on common overlaps. I think we may be able to hold off on clustering till a future improvement but for now we should probably at least attempt to sort this list based on start offsets?

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -80,12 +118,55 @@ public RewriteStrategy options(Map<String, String> options) {
...
+
     Preconditions.checkArgument(!sortOrder.isUnsorted(),
         "Can't use %s when there is no sort order, either define table %s's sort order or set sort" +
             "order in the action",
         name(), table().name());
 
     SortOrder.checkCompatibility(sortOrder, table().schema());
   }
+
+  boolean haveGoodFileSizes(Iterable<FileScanTask> dataFiles) {
```
RussellSpitzer, line 160: I don't think we should check file size if we are also checking "sortedness" but if we are I think we should just use the binpack rules.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -80,12 +118,55 @@ public RewriteStrategy options(Map<String, String> options) {
...
+    int totalDataFiles = Iterables.size(dataFiles);
+    int misSizedDataFiles = Iterables.size(super.selectFilesToRewrite(dataFiles));
+    return (0.0 + misSizedDataFiles) / totalDataFiles <= misSizedRatioThreshold;
```
RussellSpitzer, line 163: Rather than adding 0.0 I think a cast as double would probably be clearer

core/src/main/java/org/apache/iceberg/util/SortStrategyUtil.java
```diff
@@ -0,0 +1,223 @@
...
+  public static Iterable<FileScanTask> removeSortedFiles(Iterable<FileScanTask> fileScans, SortOrder sortOrder) {
+    List<Integer> sortColIndices = getSortColumnIndices(sortOrder);
+    List<DataFile> dataFiles = FluentIterable.from(fileScans).stream()
+        .map(FileScanTask::file)
+        .collect(Collectors.toList());
+    return FluentIterable.from(fileScans)
+        .filter(f -> hasOverLap(f.file(), dataFiles, sortColIndices, sortOrder.schema()));
+  }
+
+  private static boolean hasOverLap(
```
RussellSpitzer, line 101: hasOverLap -> hasOverlap since "overlap" is a single word

core/src/main/java/org/apache/iceberg/util/SortStrategyUtil.java
```diff
@@ -0,0 +1,223 @@
...
+      DataFile file,
+      List<DataFile> dataFiles,
+      List<Integer> sortColIndices,
+      Schema sortSchema) {
+    for (DataFile f : dataFiles) {
```
RussellSpitzer, line 106: I generally like Streams.anyMatch() for things like this but that's my personal pref

core/src/main/java/org/apache/iceberg/util/SortStrategyUtil.java
```diff
@@ -0,0 +1,223 @@
...
+      }
+    }
+    return false;
+  }
+
+  static List<Integer> getSortColumnIndices(SortOrder sortOrder) {
+    List<Integer> sortColIndices = Lists.newArrayList();
+    for (SortField sortField : sortOrder.fields()) {
+      if (!sortField.transform().isIdentity()) {
+        LOG.info("Cannot measure sortedness score on derived sort column {}",
```
RussellSpitzer, line 118: Why is this the case? Shouldn't the stored metric values just be transformed already?

core/src/main/java/org/apache/iceberg/util/SortStrategyUtil.java
```diff
@@ -0,0 +1,223 @@
...
+    return 1.0 - (overlapScore - 1.0) / (files.size() - 1.0);
+  }
+
+  // Check if a data point (x[1], x[2],..., x[k]) has overlap with a data file that has lower bounds of
+  // (min[1], min[2],..., min[k]) and upper bounds of (max[1], max[2],..., max[k])
+  // An overlap happens if for j going from 1 to k:
+  // if x[j] < min[j] or x[j] > max[j] then return false
+  // else if min[j <= x[j] <= max[j] and min[j] != max[j] and  then return true
+  // else (meaning min[j] == x[j] == max[j]) continue with j + 1
+  private static boolean isOverlap(
```
RussellSpitzer, line 157: I think we probably could just compare the pairs outright like we do
https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java#L1335-L1376

We also probably don't need the manfiest evaluators I think...

core/src/main/java/org/apache/iceberg/util/SortStrategyUtil.java
```diff
@@ -0,0 +1,223 @@
...
+    return overLapCount;
+  }
+
+  private static double overlapScore(
+      List<ByteBuffer> point,
+      List<DataFile> files,
+      List<Integer> sortColIndices,
+      Schema sortSchema) {
+    double overlapScore = overlapCount(point, files, sortColIndices, sortSchema);
+    return 1.0 - (overlapScore - 1.0) / (files.size() - 1.0);
```
RussellSpitzer, line 148: I'm not sure I understand the rational for the normalization here.

core/src/main/java/org/apache/iceberg/util/SortStrategyUtil.java
RussellSpitzer: Also a little confusing that we have "hasOverlap" and "isOverlap", I think we may just want to have a 
```java
boolean overlaps(fileA, fileB, fieldIds) {
   return true if fileA overlaps fileB 
}
```

core/src/test/java/org/apache/iceberg/actions/TestSortStrategy.java
```diff
@@ -120,23 +114,51 @@ public void testSelectAll() {
...
+
+  @Test
+  public void testSortednessScoreThreshold() {
+    List<FileScanTask> expected = ImmutableList.<FileScanTask>builder()
+        .addAll(tasksForSortOrder(table.sortOrder().orderId(), 500, 500, 510, 495))
         .build();
 
+    // Note that the sortedness score of existing data files of the table is 0.0
     RewriteStrategy strategy = defaultSort().options(ImmutableMap.of(
+        SortStrategy.SORTEDNESS_SCORE_THRESHOLD, "0.01"));
```
RussellSpitzer, line 140: We should have some test examples where all the files are completely sorted and nothing happens, then we should have a test when all files overlap and are a sorted, and we should have one where only a subset of files are unsorted.

RussellSpitzer (changes requested):
Sorry that it took me so long to get to this. It's been some difficult weeks for me lately.

I have a few major focus areas on this PR that I think need work:
1. The "mis-sized" approach seems a bit odd to me, for that I think we should stick with the absolute thresholds that we are using in BinPackStrategy. Ie Files smaller than X greater than Y and what not. I also think that we should either be operating with size limits or sort limits but mixing the two seems to not make sense to me. I'm open to arguements that in conjunction the two could work better but I think my gut feeling is a mis sized file being mixed in with unsorted files may be counter productive. For example a small file which spans the entire sort range could mess up the sorting of a small range of values. Say a file with 100 rows with values (1 - 100) gets included in a sort which is operating on files with millions of rows covering all only with values (1 - 10). Our output would be potentially files spanning (1 - 100) giving us worse overlap than we started. I think for this PR we should make it optional, size based file selection or sortedness selection although we can probably deprecate "size based" in the future. If we do want to combine them I think it should just be an OR with the Binpack options.

2. I think we have a bit of work to do on the overlap code, I think currently it leans into using manifest evaluators but i'm not convinced that is required in this case. I think we are probably ok just comparing the values like they are compared in the TestRewriteDatafilesAction test.

3. I think we are missing some testing, including proving that the current operation is idempotent, we may just want to change the title since I think a true idempotent action may require a more sophisticated approach.

4. Normalization in the sortedness score, I'm not sure if this is necessary at the moment and I think it probably makes the idempotency much more difficult. I think if we are going to normalize we should probably have a good explanation on the expected behavior and rational behind normailzation. I would be more comfortable with just a strict threshold.

5. File grouping, currently we just binpack all files that we consider "unsorted" I think this may give us issues that I identified in my comments. For example unsorted files (1-10) (1-10) (40-50) (40-50) and sorted file (25-35) would produce a rewrite of (1-10)(1-10)(40-50)(40-50) which would potentially write file (1 - 50) or (1-25)(25-50) or whatnot, which would leave us with new overlapping files that we didn't have when we started. I think when we group files they must be grouped together with their overlapping neighbors and we start a new group whenever we have a file which does not overlap with the previous overlapping files. This may be ok to hold off on for a future PR though but I think we can do a quick pr here without too much special logic and get a good result so let's work on that as well if possible.

PR #4764, opened 2022-05-13
Core: Add content and delete file counts to manifest tables

aokolnychyi:
This PR adds content and delete file counts fields to `manifests` and `all_manifests` metadata tables.
Existing tests were adapted to cover the new functionality.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/AllManifestsTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllManifestsTable.java
@@ -44,6 +44,7 @@ import org.apache.iceberg.util.StructProjection;
  */
 public class AllManifestsTable extends BaseMetadataTable {
   private static final Schema MANIFEST_FILE_SCHEMA = new Schema(
+      Types.NestedField.required(14, "content", Types.IntegerType.get()),
       Types.NestedField.required(1, "path", Types.StringType.get()),
       Types.NestedField.required(2, "length", Types.LongType.get()),
       Types.NestedField.optional(3, "partition_spec_id", Types.IntegerType.get()),
@@ -51,6 +52,9 @@ public class AllManifestsTable extends BaseMetadataTable {
       Types.NestedField.optional(5, "added_data_files_count", Types.IntegerType.get()),
       Types.NestedField.optional(6, "existing_data_files_count", Types.IntegerType.get()),
       Types.NestedField.optional(7, "deleted_data_files_count", Types.IntegerType.get()),
+      Types.NestedField.required(15, "added_delete_files_count", Types.IntegerType.get()),
+      Types.NestedField.required(16, "existing_delete_files_count", Types.IntegerType.get()),
+      Types.NestedField.required(17, "deleted_delete_files_count", Types.IntegerType.get()),
       Types.NestedField.optional(8, "partition_summaries", Types.ListType.ofRequired(9, Types.StructType.of(
           Types.NestedField.required(10, "contains_null", Types.BooleanType.get()),
           Types.NestedField.required(11, "contains_nan", Types.BooleanType.get()),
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -44,13 +44,17 @@
  */
 public class AllManifestsTable extends BaseMetadataTable {
   private static final Schema MANIFEST_FILE_SCHEMA = new Schema(
+      Types.NestedField.required(14, "content", Types.IntegerType.get()),
```
aokolnychyi, line 47: Using fresh IDs but adding columns in places where they make sense instead of adding to the end.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -44,13 +44,17 @@
...
       Types.NestedField.required(1, "path", Types.StringType.get()),
       Types.NestedField.required(2, "length", Types.LongType.get()),
       Types.NestedField.optional(3, "partition_spec_id", Types.IntegerType.get()),
       Types.NestedField.optional(4, "added_snapshot_id", Types.LongType.get()),
       Types.NestedField.optional(5, "added_data_files_count", Types.IntegerType.get()),
       Types.NestedField.optional(6, "existing_data_files_count", Types.IntegerType.get()),
       Types.NestedField.optional(7, "deleted_data_files_count", Types.IntegerType.get()),
+      Types.NestedField.required(15, "added_delete_files_count", Types.IntegerType.get()),
```
aokolnychyi, line 55: If we did this initially, we'd probably just have one set of counts with a generic name but this is public and widely used.

aokolnychyi:
cc @RussellSpitzer @flyrain @karuppayya @szehon-ho @rdblue @kbendick

szehon-ho:
Looks good to me, maybe we should add a test, if it's not too tricky?

aokolnychyi:
Yeah, let me add one. I did not release we did not have a test for that.

aokolnychyi:
@szehon-ho, I added a test for Spark 3.2. I made older Spark versions compile but did not add the new test logic there.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -1457,4 +1514,28 @@ private void asMetadataRecord(GenericData.Record file) {
...
+
+  private DeleteFile writePositionDeletes(Table table, PartitionSpec spec, StructLike partition,
+                                          Iterable<PositionDelete<InternalRow>> deletes) {
+    PositionDeleteWriter<InternalRow> positionDeleteWriter = newPositionDeleteWriter(table, spec, partition);
+
+    try (PositionDeleteWriter<InternalRow> writer = positionDeleteWriter) {
+      for (PositionDelete<InternalRow> delete : deletes) {
+        writer.write(delete);
+      }
+    } catch (IOException e) {
```
szehon-ho, line 1535: For test method I usually think to declare  "throws Exception" instead of writing catches to reduce lines of code, but it may be just personal preference.

szehon-ho (approved):
Looks good to me, thanks for the test

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
aokolnychyi: Me too. In this case, the method definition will be a bit ugly as it won't fit on one line and I'd have to put throws on a separate line. It is also part of try with resources which is a bit of a special case.

szehon-ho: Ah makes sense.

PR #4767, opened 2022-05-13
Change types into dataclasses

Fokko:
Proposal to change the types into dataclasses.

This has several improvments:

- We can use the dataclasss field(repr=True) to include the fields in the representation, instead of building our own strings
- We can assign the types in the post_init when they are dynamic (List, Maps, Structs etc) , or just override them when they are static (Primitives)
- We don't have to implement any eq methods because they come for free
- The types are frozen, which is kind of nice since we re-use them
- The code is much more consise
- We can assign the min/max of the int/long/float as Final as of 3.8: https://peps.python.org/pep-0591/

My inspiration was the comment by Kyle:
https://github.com/apache/iceberg/pull/4742#discussion_r869494393

This would entail implementing eq, but why not use the generated one since we're comparing all the attributes :)

Would love to get you input

### DIFF
--- a/python/src/iceberg/types.py
+++ b/python/src/iceberg/types.py
@@ -203,13 +203,13 @@ class StructType(IcebergType):
     _instances: ClassVar[Dict[Tuple[NestedField, ...], "StructType"]] = {}
 
     def __new__(cls, *fields: NestedField, **kwargs):
-        if "fields" in kwargs:
+        if not fields and "fields" in kwargs:
             fields = kwargs["fields"]
         cls._instances[fields] = cls._instances.get(fields) or object.__new__(cls)
         return cls._instances[fields]
 
     def __init__(self, *fields: NestedField, **kwargs):
-        if "fields" in kwargs:
+        if not fields and "fields" in kwargs:
             fields = kwargs["fields"]
         object.__setattr__(self, "fields", fields)
 
--- a/python/src/iceberg/types.py
+++ b/python/src/iceberg/types.py
@@ -42,7 +42,7 @@ class Singleton:
         return cls._instance
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class IcebergType:
     """Base type for all Iceberg Types
 
@@ -75,7 +75,7 @@ class PrimitiveType(IcebergType):
     """
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class FixedType(PrimitiveType):
     """A fixed data type in Iceberg.
 
@@ -123,7 +123,7 @@ class DecimalType(PrimitiveType):
         object.__setattr__(self, "type_string", f"decimal({self.precision}, {self.scale})")
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class NestedField(IcebergType):
     """Represents a field of a struct, a map key, a map value, or a list element.
 
@@ -180,7 +180,7 @@ class NestedField(IcebergType):
         return self.field_type
 
 
-@dataclass(frozen=True, eq=True, repr=True, init=False)
+@dataclass(frozen=True, init=False)
 class StructType(IcebergType):
     """A struct type in Iceberg
 
@@ -209,7 +209,7 @@ class StructType(IcebergType):
         object.__setattr__(self, "type_string", f"struct<{', '.join(map(str, self.fields))}>")
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class ListType(IcebergType):
     """A list type in Iceberg
 
@@ -249,7 +249,7 @@ class ListType(IcebergType):
         )
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class MapType(IcebergType):
     """A map type in Iceberg
 
@@ -297,7 +297,7 @@ class MapType(IcebergType):
         )
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class BooleanType(PrimitiveType, Singleton):
     """A boolean data type in Iceberg can be represented using an instance of this class.
 
@@ -312,7 +312,7 @@ class BooleanType(PrimitiveType, Singleton):
     type_string = "boolean"
 
 
-@dataclass(frozen=True, eq=True, repr=True)
+@dataclass(frozen=True)
 class IntegerType(PrimitiveType, Singleton):
     """An Integer data type in Iceberg can be represented using an instance of this class. Integers in Iceberg are
     32-bit signed and can be promoted to Longs.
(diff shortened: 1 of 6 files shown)


python/src/iceberg/types.py
```diff
@@ -42,57 +42,63 @@ def __new__(cls, *args, **kwargs):
...
-    """Base class for all Iceberg Primitive Types"""
+    """Base class for all Iceberg Primitive Types
 
+    Example:
+        >>> str(PrimitiveType())
+        'PrimitiveType()'
+    """
 
+
+@dataclass(frozen=True, eq=True, repr=True)
```
danielcweeks, line 78: Aren't defaults for `eq` and `repr` `True`?  Is there a reason to be explicit about those two vs. relying on the defaults?

python/src/iceberg/types.py
```diff
@@ -148,56 +159,28 @@ def __new__(
...
-    def name(self) -> str:
-        return self._name
-
-    @property
-    def doc(self) -> Optional[str]:
-        return self._doc
-
+        return not self.is_optional
+
+    def __post_init__(self):
```
danielcweeks, line 166: This may be beyond the scope of this PR (since the existing behavior is the same), but is there a way to lazily create the type string?  It's probably not very concerning from a performance standpoint in most cases, but it's also not typically used.  Just wondering if using data class makes that easier.

python/src/iceberg/types.py
```diff
@@ -209,25 +192,24 @@ class StructType(IcebergType):
...
     """
 
-    _instances: Dict[Tuple[NestedField, ...], "StructType"] = {}
+    fields: List[NestedField] = field()
+
+    _instances: ClassVar[Dict[Tuple[NestedField, ...], "StructType"]] = {}
 
-    def __new__(cls, *fields: NestedField):
+    def __new__(cls, *fields: NestedField, **kwargs):
+        if "fields" in kwargs:
```
rdblue, line 200: Doesn't this need to check that `*fields` is empty?

rdblue:
I really like how much this removes! Nice work.

python/src/iceberg/types.py
```diff
@@ -42,57 +42,63 @@ def __new__(cls, *args, **kwargs):
...
-    def __init__(self, length: int):
-        if not self._initialized:
-            super().__init__(f"fixed[{length}]", f"FixedType(length={length})")
-            self._length = length
-
-    @property
-    def length(self) -> int:
-        return self._length
+    def __post_init__(self):
+        object.__setattr__(self, "type_string", f"fixed[{self.length}]")
```
samredai, line 98: Is it required to use setattr here because the class is frozen?

python/src/iceberg/types.py
```diff
@@ -148,56 +159,28 @@ def __new__(
...
+                if self.doc is None
+                else f" ({self.doc})"
+            ),
+        )
+
+    # Alias for field_type
     @property
     def type(self) -> IcebergType:
-        return self._type
+        return self.field_type
```
samredai, line 180: Should we just rename this property to "field_type"? It might require some small changes in other parts of the library but it's probably better to not clash with a keyword and field_type is the argument name anyway. We can also do this in a follow-up PR.

python/src/iceberg/types.py
```diff
@@ -209,25 +192,24 @@ class StructType(IcebergType):
...
-            )
-            self._fields = fields
-
-    @property
-    def fields(self) -> Tuple[NestedField, ...]:
-        return self._fields
+    def __init__(self, *fields: NestedField, **kwargs):
+        if "fields" in kwargs:
+            fields = kwargs["fields"]
+        object.__setattr__(self, "fields", fields)
```
samredai, line 208: This should be nested in the if statement right above it right? If there's no fields kwarg we can't set it.

samredai (approved):
This looks great, thanks @Fokko ! Left a few comments but overall LGTM 👍

python/src/iceberg/types.py
Fokko: The Python zen dictates: _Explicit is better than implicit_. But I agree that this might be too obvious. I'll remove eq and repr 👍🏻

python/src/iceberg/types.py
Fokko: Hey @danielcweeks Thanks for pointing this out. I agree that lazily initializing the string is much better. In terms of caching, I've added the `cached_property` to the StructType: https://github.com/apache/iceberg/pull/4767/commits/5ac86f44e8b81d2e7285f85207c1731c9860ddec#diff-7882bc96112705abb40e28d8f45ef911e1a3d43dd62c4987f020dc8738ee7407R216-R218

I don't think we should cache literals, but only the dynamically generated ones. WDYT?

python/src/iceberg/types.py
Fokko: There seems to be a subtle behavioral change that requires this. Without this logic, the following test is failing:
```python
assert str(type_var) == str(eval(repr(type_var)))
```

When we dig into this:
```python
>> repr(type_var)
StructType(
    fields=(
        NestedField(field_id=1, name='optional_field', field_type=IntegerType(), is_optional=True), 
        NestedField(field_id=2, name='required_field', field_type=FixedType(length=5), is_optional=False), 
        ...
    )
)
```
It looks like it tries to initialize it again using a keyword argument:
```python
>> eval(repr(type_var))
{TypeError}__new__() got an unexpected keyword argument 'fields'
```
Instead of variable unpacking.

Using this logic we give priority to the keyword argument:
```
StructField(
    NestedField(field_id=1, name='optional_field', field_type=IntegerType(), is_optional=True), 
    NestedField(field_id=2, name='required_field', field_type=FixedType(length=5), is_optional=False), 
)
or
StructField(
    fields=(
        NestedField(field_id=1, name='optional_field', field_type=IntegerType(), is_optional=True), 
        NestedField(field_id=2, name='required_field', field_type=FixedType(length=5), is_optional=False), 
    )
)
```
I've added an additional check to see if fields is empty.

python/src/iceberg/types.py
Fokko: Yes, that's exactly the reason :)

python/src/iceberg/types.py
Fokko: Yes, I had exactly the same thoughts. I would love to change this to `field_type` to avoid clashing with the internal `type`.

Fokko: Relative small change: https://github.com/apache/iceberg/pull/4767/commits/f4661656ad303239467888dc9913d5d4bd83014e :)

python/src/iceberg/types.py
Fokko: This logic allows you to both pass in the fields through args and kwargs:
```python
StructField(
    NestedField(field_id=1, name='optional_field', field_type=IntegerType(), is_optional=True), 
    NestedField(field_id=2, name='required_field', field_type=FixedType(length=5), is_optional=False), 
)
or
StructField(
    fields=(
        NestedField(field_id=1, name='optional_field', field_type=IntegerType(), is_optional=True), 
        NestedField(field_id=2, name='required_field', field_type=FixedType(length=5), is_optional=False), 
    )
)
```
This was a change required as explained in https://github.com/apache/iceberg/pull/4767#discussion_r873959425

samredai: Ah I see, makes sense!

python/src/iceberg/types.py
danielcweeks: That sounds great to me.

danielcweeks (approved):

Issue #3117, opened 2021-09-15
Flink:  Decouple the iceberg integration work from hadoop libraries


openinx:
Thanks for working on this. I wanted to draw attention to the issue mentioned above (Update FlinkCatalogFactory to implement the new Factory interface). I closed it as basically a duplicate of this one though.

The reason I opened that issue is that several users have reported difficulties in using Iceberg with Flink, specifically in configuring Hadoop.

In some environments, users might not have access to setup the class path properly for including Hadoop. The new [Context interface that provides access to the ClassLoader ](https://github.com/apache/flink/commit/4497e96b96724aee6637f19f64ffc8ba47a9b0ac#diff-d4dda8460ab068327897de5487517f43186faced18d3a121981e70f6ad9f827aR85) might help with that.

The ultimate goal is to decouple the need for hadoop to be in the environment when it's not actually needed (for example, when using `S3FileIO`). The thought was that upgrading to 1.13 (and the new FlinkCatalogFactory interface) will be helpful to provide the easier access to the ClassLoader.

In environments where hadoop `Configuration` isn't used, such as Ververica Platform or AWS Kinesis Data Analytics with `S3FileIO`, the hope is to remove the need for the Hadoop `Configuration` entirely.

For Ververica, users have reported a workaround but in Kinesis Data Analytics, it's a bit more difficult (as far as has been reported).

More information on the Hadoop `Configurable` concern can be found in this issue: https://github.com/apache/iceberg/issues/3044

Thanks again @zhangjun0x01!

_Originally posted by @kbendick in https://github.com/apache/iceberg/issues/2558#issuecomment-908597980_

xinbinhuang:
@openinx  I would like to work on this. Can you assign it to me?

openinx:
Of course, assigned to you @xinbinhuang !  Pls enjoy.

rdblue:
@openinx, @xinbinhuang, there's no need to assign issues to specific people. Assigning issues almost always prevents work from getting done instead of deconflicting. It is far more likely that someone has an issue assigned and then doesn't have time to work on it, compared to having two people work on an issue at the same time. If you want to work on something, just mention that you plan to do it or link to a PR, rather than waiting to be assigned.

xinbinhuang:
Makes sense! thanks for the info!

fi-ecavc:
[...]

PR #4183, opened 2022-02-22
Flink: Implement catalog Factory interface (#3117)

kainoa21:
Implements the catalog Factory interface while also attempting to decouple catalog loading from requiring hadoop dependencies (see [#3117](https://github.com/apache/iceberg/issues/3117)).

For now this attempts to leave all of the existing catalog creation interfaces in tact but swaps out `Configuration` for `Object` as a follow on the previous work in [#3590](https://github.com/apache/iceberg/pull/3590).  If we end up going this direction, some additional work will be necessary to ensure no loss of functionality for the existing Hadoop and Hive catalog implementations when using the new interface.

While testing these changes locally, I was able to load a non-hadoop based catalog in a Flink SQL environment without any hadoop jars.  Though I did still eventually hit an error while attempting to read parquet files due to parquet's [hadoop dependency](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java#L174).

PR #4768, opened 2022-05-14
[wip]: Decouple Hadoop Configuration from FlinkCategoryFactory

xinbinhuang:
fixed: #3117


kainoa21:
Thanks for this update @xinbinhuang.  I will attempt to do some testing in non-hadoop environments, although I feel like I am likely to run into the same issue that I ultimately hit while attempting to do a similar fix in [this PR](https://github.com/apache/iceberg/pull/4183/files#diff-07827679a9caad4c43e57c09ee1a1d553559856743af8b5f81d00e3220ca81ce) related to `HadoopFileIO` in `FlinkSource.java`?  

Also, was there a particular reason that you did not include updates for Flink 1.14?

xinbinhuang:
> Thanks for this update @xinbinhuang.  I will attempt to do some testing in non-hadoop environments, although I feel like I am likely to run into the same issue that I ultimately hit while attempting to do a similar fix in [this PR](https://github.com/apache/iceberg/pull/4183/files#diff-07827679a9caad4c43e57c09ee1a1d553559856743af8b5f81d00e3220ca81ce) related to `HadoopFileIO` in `FlinkSource.java`?  

Ah sorry, wasn't aware you were working on similar things when I started the PR. Maybe we can merge incorporate them together somehow? 

> Also, was there a particular reason that you did not include updates for Flink 1.14?

No, I started out in 1.15, but realized that KDA only support up to 1.13 currently. Since this fix benefits KDA the most, so I added 1.13 support as well.

xinbinhuang:
@kainoa21 I don't think the parquet error is necessarily an issue related to our fixes. We try to decouple Hadoop s.t. Flink Iceberg application that doesn't use Hadoop related things doesn't require Hadoop on the classpath i.e. Glue. Parquet depends on Hadoop so I would classify it as Hadoop related

dyrnq:
Stare

Issue #4749, opened 2022-05-11
Add sequenceNumber method for ContentFile interface


chenjunjiedada:
Now the `DeleteFile` and `DataFile` don't have `sequenceNumber` method to get their sequence number directly, so we have to read inherited metadata from `Manifest`. It is not convenient, for example, we use two arrays and `Pair` in the delete file index.  Besides, we may have some actions to deal with deletes in real production, such as cleaning up deletes that have no data file with a smaller sequence number.  It would benefit many places if we could add a method `sequenceNumber`.

This is easy to implement but it changes the spec a bit, but actually, the `DeleteFile` and `DataFile` do have the sequence number in the V2 format definition.

szehon-ho:
Do you mean just add it to the ContentFile API?  Maybe it can be done without changing the spec much, like https://github.com/apache/iceberg/pull/1317 ?

chenjunjiedada:
@szehon-ho, What I want is like this https://github.com/apache/iceberg/pull/4769.

PR #4769, opened 2022-05-14
Api: add sequenceNumber method

chenjunjiedada:
Currently, there is no straight way to get sequence numbers of `DeleteFile` and `DataFile`.  It is not convenient to use, for example, it uses two arrays and the `Pair` class to build the delete file index. Furthermore, we may have some actions which depend on comparing the sequence numbers between deletes and data files. This PR tries to add `sequenceNumber` method to `ContentFile` and set the sequence number when applying the inherited metadata thus both `DeleteFile` and `DataFile` can get sequence numbers directly.


core/src/main/java/org/apache/iceberg/BaseFile.java
```diff
@@ -205,6 +207,15 @@ void setSpecId(int specId) {
     this.partitionSpecId = specId;
   }
 
+  @Override
+  public Long sequenceNumber() {
+    return sequenceNumber;
+  }
+
+  void setSequenceNumber(Long seq) {
```
rdblue, line 215: Sequence number is tracked by the manifest entry, not the content file. And I don't think it is a good idea to make this mutable. What is the rationale for adding it here (not just a ref to another issue)?

chenjunjiedada: I was following the specId change mentioned in https://github.com/apache/iceberg/issues/4749.

rdblue: The spec ID isn't available in manifest entry, which is why we thought it was reasonable to add it. Since this is available in the manifest entry, I would prefer to use it from there. That is, unless this is blocking something and we _need_ to add it for some reason.

chenjunjiedada: This is motivated when writing https://github.com/apache/iceberg/pull/4748 which handles sequence numbers. And I think we should need this when cleaning the delete files in the future.  Right now, we don't clean the delete files after bin-packing.

gaborkaszab:
Just for the record, Apache Impala would also need this sequence number to be exposed through the Iceberg API. Would be nice to know the feasibility, and if the community is open to such a change.

chenjunjiedada:
Moved changes to https://github.com/apache/iceberg/pull/5760.

PR #4770, opened 2022-05-14
Core: Accept all API warnings in master

kbendick:
Adds a blanket acceptance for all existing breaks that occurred prior to the introduction of the API / ABI compatibility guarantee checking tool, `revapi`, in https://github.com/apache/iceberg/pull/4638

cc @rdblue

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -1,2 +1,87 @@
 versionOverrides:
   org.apache.iceberg:iceberg-api:release-base-0.13.0: "0.13.0"
+acceptedBreaks:
+  release-base-0.13.0:
+    org.apache.iceberg:iceberg-api:
+    - code: "java.class.defaultSerializationChanged"
+      old: "class org.apache.iceberg.PartitionSpec"
+      new: "class org.apache.iceberg.PartitionSpec"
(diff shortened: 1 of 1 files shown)


rdblue (approved):

rdblue:
This looks good since there are only 2 warning types: `defaultSerializationChanged` and `addedToInterface`. The serialization changes are for classes without `serialVersionUID`, but Iceberg does not guarantee any compatibility for serialized objects across versions. The interface additions are also fine because these interfaces are for internal implementations. Adding methods does not break binary compatibility for people using the interfaces.

kbendick:
> This looks good since there are only 2 warning types: `defaultSerializationChanged` and `addedToInterface`. The serialization changes are for classes without `serialVersionUID`, but Iceberg does not guarantee any compatibility for serialized objects across versions. The interface additions are also fine because these interfaces are for internal implementations. Adding methods does not break binary compatibility for people using the interfaces.

We can be more specific for the changes (provide a justification for each or for generally that class of error - like methods added to interfaces is something I hope to suppress). But for now this will unblock 😄

rdblue:
Thanks, @kbendick!

PR #4771, opened 2022-05-14
Core: Add OAuth2 to REST catalog spec

rdblue:
This adds an additional security scheme to the REST catalog spec for OAuth2.

This also adds the `v1/oauth/token` route that describes the [client credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) or [token exchange](https://datatracker.ietf.org/doc/html/rfc8693) flows.


open-api/rest-catalog-open-api.yaml
```diff
@@ -1435,12 +1463,66 @@ components:
...
+            - urn:ietf:params:oauth:grant-type:token-exchange
+        client_id:
+          type: string
+        client_secret:
+          type: string
+        subject_token:
+          type: string
+        subject_token_type:
+          type: string
+        scope:
```
danielcweeks, line 1493: I believe this should be plural (`scopes`)

open-api/rest-catalog-open-api.yaml
```diff
@@ -1435,12 +1463,66 @@ components:
...
+              expires_in:
+                type: integer
+                description: lifetime of the access token in seconds
+              issued_token_type:
+                type: string
+              token_type:
+                type: string
+              refresh_token:
+                type: string
+              scope:
```
danielcweeks, line 1523: Plural here as well.

open-api/rest-catalog-open-api.yaml
```diff
@@ -58,7 +58,9 @@ servers:
         default: ""
 # All routes are currently configured using an Authorization header.
 security:
+  - OAuth2: [catalog]
```
rdblue, line 61: I added just one scope for all catalog and config access. Alternatively, we could separate this into read and read/modify scopes to allow read-only access.

open-api/rest-catalog-open-api.yaml
```diff
@@ -1435,12 +1463,66 @@ components:
...
+        - Token exchange flow - https://datatracker.ietf.org/doc/html/rfc8693
+      type: object
+      required:
+        - grant_type
+      properties:
+        grant_type:
+          type: string
+          enum:
+            - client_credentials
+            - urn:ietf:params:oauth:grant-type:token-exchange
```
rdblue, line 1484: @bryanck, do you know if the `urn:ietf:..` prefix can be used with `client_credentials` as well? If so, then I'd probably change so these are uniform.

bryanck: Not that I'm aware of, what you have here looks correct to me, per the OAuth2 specs.

open-api/rest-catalog-open-api.yaml
```diff
@@ -120,6 +122,32 @@ paths:
...
+    post:
+      tags:
+        - Auth API
+      summary: Get a token using an OAuth2 flow
+      operationId: getToken
+      description:
+        Exchange credentials for a token using the OAuth2 client credentials flow or token exchange.
+      requestBody:
+        content:
+          application/json:
```
bryanck, line 136: This should be `application/x-www-form-urlencoded`

open-api/rest-catalog-open-api.yaml
```diff
@@ -120,6 +122,32 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/TokenRequest'
+      responses:
+        200:
+          $ref: '#/components/responses/TokenResponse'
+        400:
+          $ref: '#/components/responses/BadRequestErrorResponse'
+        401:
```
bryanck, line 144: The OAuth2 RFC also specifies what the error response body should look like. We'd need to add an `error` field containing one of the codes if we want to be strict about that.

bryanck: (Here's the relevant [section](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2) in the RFC)

open-api/rest-catalog-open-api.yaml
```diff
@@ -1435,12 +1463,66 @@ components:
...
+        grant_type:
+          type: string
+          enum:
+            - client_credentials
+            - urn:ietf:params:oauth:grant-type:token-exchange
+        client_id:
+          type: string
+        client_secret:
+          type: string
+        subject_token:
```
bryanck, line 1489: Just to note, the OAuth2 spec [recommends](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) passing client ID and secret in an `Authorization: Basic...` header, but passing in the request body is supported.

open-api/rest-catalog-open-api.yaml
kbendick, line 1484: Should we consider giving the prefix `urn:ietf:params:oauth:` special status, like via a reusable component? Maybe a schema element?

It can be used with `token-type`, `grant-type`, and many other things.

Maybe it could be  a prefix in an `OAuthParams` schema object that has various? various possible enums in it. Mostly thinking out loud.

open-api/rest-catalog-open-api.yaml
rdblue: I'll add this to the description. I added them here because I saw that they can be passed here. They're also optional. Documenting this route isn't necessarily required, but the Swagger/OpenAPI docs say that you _can_ document the token endpoint and I thought it would be good to do as much as we can.

core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
```diff
@@ -151,15 +151,15 @@ private <T> T execute(
...
-    addRequestHeaders(request);
 
-    if (requestBody != null) {
-      try {
-        StringEntity stringEntity = new StringEntity(mapper.writeValueAsString(requestBody));
-        request.setEntity(stringEntity);
-      } catch (JsonProcessingException e) {
-        throw new RESTException(e, "Failed to write request body: %s", requestBody);
-      }
+    if (requestBody instanceof Map) {
```
rdblue, line 155: This change updates `HTTPClient` so that a string map will be form encoded rather than JSON encoded. This minimizes the changes needed to support requests to the `v1/oauth/token` endpoint, but an alternative is to add another method, `RESTClient.formPost`, to send map data encoded as `application/x-www-form-urlencoded` instead of JSON.

@danielcweeks, what do you think?

open-api/rest-catalog-open-api.yaml
rdblue: Fixed!

open-api/rest-catalog-open-api.yaml
rdblue: I added an OAuthErrorResponse to use instead of the standard one.

open-api/rest-catalog-open-api.yaml
rdblue: Splitting it would make this more confusing, I think. There's not a lot to reuse here besides a prefix.

open-api/rest-catalog-open-api.yaml
rdblue: Added.

core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
rdblue: Actually, the request body in `post` must be a `RESTRequest`, so we'll have to add the `formPost` method. I think that's probably the right solution anyway.

rdblue: Added `postForm` for this.

open-api/rest-catalog-open-api.yaml
```diff
@@ -1435,12 +1463,102 @@ components:
...
+          description: Subject token type for token exchange request
+        scope:
+          type: string
 
   #############################
   # Reusable Response Objects #
   #############################
   responses:
 
+    OAuthTokenResponse:
```
bryanck, line 1515: This doesn't quite match the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) for granting a token (it looks like it matches the token exchange response though)

bryanck, line 1515: This doesn't quite match the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) for granting a token (it looks like it matches the token exchange response though)

rdblue: It's a superset of the fields in section 5.1 so that we can use it for both the client credentials and token exchange flows.

I just double-checked and it looks like the main problem is that `issued_token_type` is marked as required, even though it won't be present for client credentials. Is that what you were referring to? I'll make sure that only the fields that are required in both are actually required here. I'll also update each of these with a description so it is clear when they can or should be set.

Do you know when `token_type` and `issued_token_type` would both be used for token exchange? `issued_token_type` makes sense to me, but I'm not sure why `token_type` wouldn't simply be the one that was issued.

bryanck: Yeah, that's what I meant, that the required fields for the grant token response are only `access_token` and `token_type` in the spec. For token exchange, `issued_token_type` is also required. That's referring to the HTTP Authorization token type, e.g. most of the time set to `bearer`. Also, `expires_in` is recommended but optional (for both).

rdblue: Thanks for the clarification! I just assumed `issued_token_type` was the same thing as `token_type`, but that makes sense.

bryanck (approved):

open-api/rest-catalog-open-api.yaml
kbendick: Fair point.

open-api/rest-catalog-open-api.yaml
kbendick: I think the one scope is fine for now.

danielcweeks (approved):

PR #4772, opened 2022-05-15
Core: Add request headers to REST client

rdblue:
This adds a parameter to `RESTClient` to pass headers when making a request. This was originally part of #4575, but we can get it in more easily on its own.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
+++ b/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
@@ -59,13 +59,12 @@ public class HTTPClient implements RESTClient {
   private final String uri;
   private final CloseableHttpClient httpClient;
   private final ObjectMapper mapper;
-  private final Map<String, String> additionalHeaders;
+  private final Map<String, String> baseHeaders;
 
-  private HTTPClient(
-      String uri, CloseableHttpClient httpClient, Map<String, String> additionalHeaders) {
+  private HTTPClient(String uri, Map<String, String> baseHeaders) {
     this.uri = uri;
-    this.httpClient = httpClient != null ? httpClient : HttpClients.createDefault();
-    this.additionalHeaders = additionalHeaders != null ? additionalHeaders : ImmutableMap.of();
+    this.httpClient = HttpClients.createDefault();
+    this.baseHeaders = baseHeaders != null ? baseHeaders : ImmutableMap.of();
     this.mapper = RESTObjectMapper.mapper();
   }
 
@@ -143,7 +142,8 @@ public class HTTPClient implements RESTClient {
    * @return The response entity, parsed and converted to its type T
    */
   private <T> T execute(
-      Method method, String path, Object requestBody, Class<T> responseType, Consumer<ErrorResponse> errorHandler) {
+      Method method, String path, Object requestBody, Class<T> responseType, Map<String, String> headers,
+      Consumer<ErrorResponse> errorHandler) {
     if (path.startsWith("/")) {
       throw new RESTException(
           "Received a malformed path for a REST request: %s. Paths should not start with /", path);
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
```diff
@@ -59,13 +59,12 @@ public class HTTPClient implements RESTClient {
...
   private final CloseableHttpClient httpClient;
   private final ObjectMapper mapper;
-  private final Map<String, String> additionalHeaders;
+  private final Map<String, String> baseHeaders;
 
-  private HTTPClient(
-      String uri, CloseableHttpClient httpClient, Map<String, String> additionalHeaders) {
+  private HTTPClient(String uri, Map<String, String> baseHeaders) {
     this.uri = uri;
-    this.httpClient = httpClient != null ? httpClient : HttpClients.createDefault();
```
rdblue, line 67: The builder was never initializing `httpClient` so it was always null anyway. This removes the unnecessary argument. FYI @kbendick.

core/src/main/java/org/apache/iceberg/rest/RESTCatalog.java
```diff
@@ -96,6 +96,10 @@ public void initialize(String name, Map<String, String> props) {
     this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), properties, conf);
   }
 
+  protected Map<String, String> headers() {
+    return ImmutableMap.of();
```
rdblue, line 100: Right now, headers aren't used other than overriding this in tests to validate that they can be set and are passed through the catalog correctly.

core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java
```diff
@@ -63,4 +63,8 @@ public String table(TableIdentifier ident) {
         "v1", prefix, "namespaces", RESTUtil.encodeNamespace(ident.namespace()), "tables",
         RESTUtil.encodeString(ident.name()));
   }
+
+  public String rename() {
+    return SLASH.join("v1", prefix, "tables", "rename");
```
rdblue, line 68: The path was hard-coded for rename. I'm adding this here since this PR already modifies all the other lines to add headers. FYI @kbendick.

core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
```diff
@@ -195,33 +195,34 @@ private <T> T execute(
...
   @Override
-  public void head(String path, Consumer<ErrorResponse> errorHandler) {
-    execute(Method.HEAD, path, null, null, errorHandler);
+  public void head(String path, Map<String, String> headers, Consumer<ErrorResponse> errorHandler) {
+    execute(Method.HEAD, path, null, null, headers, errorHandler);
   }
 
   @Override
-  public <T extends RESTResponse> T get(String path, Class<T> responseType,
+  public <T extends RESTResponse> T get(String path, Class<T> responseType, Map<String, String> headers,
```
danielcweeks, line 203: One thing you might want to consider is wrapping some of these parameters (i.e. headers and errorHandler) in some sort of wrapper that we can extend to further scope requests within sessions.  Just thinking something on the order of:

```java
RequestContext {
    Supplier<Map<String, String>> requestHeaders() 
    void handleResponseHeaders(Consumer<Map<String,String>>)
    Consumer<ErrorResponse> errorHandler()
    }
```

I feel like it will be hard to extend the current pattern if we need to propagate information back or retry requests (e.g. the 419 case) without being able to capture the current state.

core/src/main/java/org/apache/iceberg/rest/RESTClient.java
```diff
@@ -28,13 +29,15 @@
  */
 public interface RESTClient extends Closeable {
```
danielcweeks, line 30: If we do decide to define a `RequestContext`, this might be a good place to add an inner interface.

rdblue:
Thanks for the review, @danielcweeks! I think the `RequestContext` is a great idea, but I'm not sure that we're going to need it so I'd probably hold off and do it later.

PR #4773, opened 2022-05-15
API: Add SessionCatalog interface and base class

rdblue:
This adds a `SessionCatalog` interface for engines that share catalog instances across sessions.

`SessionCatalog` is like `Catalog`, but passes a `SessionContext` object in each call. This session context can be used to carry identity and credentials, which are passed through to the catalog service. `SessionCatalog` does not extend `Catalog` to avoid bugs where context is accidentally not passed.

Because `SessionCatalog` does not implement `Catalog`, a base implementation class is also included in this PR that adapts `SessionCatalog` to `Catalog` using `withContext` and `asCatalog` methods. For example:

```java
  boolean exists = sessionCatalog.withContext(sessionContext, (Catalog catalog) -> {
    return catalog.tableExists(tableIdent);
  });
```

This was originally part of #4575, but is not a separate PR to avoid blocking other work.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
+++ b/api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
@@ -141,7 +141,7 @@ public interface SessionCatalog {
   Catalog.TableBuilder buildTable(SessionContext context, TableIdentifier ident, Schema schema);
 
   /**
-   * Register a table with if it does not exist.
+   * Register a table if it does not exist.
    *
    * @param context session context
    * @param ident a table identifier
--- a/api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
+++ b/api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
@@ -1,0 +1,323 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
```diff
@@ -0,0 +1,323 @@
...
+   *
+   * @param context session context
+   * @param ident a table identifier
+   * @param schema a schema
+   * @return the builder to create a table or start a create/replace transaction
+   */
+  Catalog.TableBuilder buildTable(SessionContext context, TableIdentifier ident, Schema schema);
+
+  /**
+   * Register a table with if it does not exist.
```
danielcweeks, line 144: ~~with~~

core/src/main/java/org/apache/iceberg/catalog/BaseSessionCatalog.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public Catalog asCatalog(SessionContext context) {
+    return catalogs.get(context.sessionId(), id -> new AsCatalog(context));
+  }
+
+  public <T> T withContext(SessionContext context, Function<Catalog, T> task) {
+    return task.apply(asCatalog(context));
+  }
+
+  public class AsCatalog implements Catalog, SupportsNamespaces {
```
danielcweeks, line 67: Can this inner class be static?  Caches historically have reference problems with non-static inner classes (though I don't recall the exact scenarios).

danielcweeks: I don't think this usage has the issues I was concerned about. Resolving.

danielcweeks (approved):
Very minor nit on comment text.  +1

rdblue:
Thanks, @danielcweeks! And to everyone that looked at this in #4575.

PR #3887, opened 2022-01-13
AWS: fix Iceberg to Glue schema conversion

natsukawa-kanou:
some bugs fixed:
- if identity transform, duplicated column name and will fail
- use full column name, `y.z` instead of `z`
- convert null to "unknown" to avoid null pointer error
- improve dedupe logic

@jackye1995 @yyanyy

PR #4775, opened 2022-05-16
[WIP] AWS: Add partition info to Glue

singhpk234:
At present partition info is not being populated to the Glue. 

From checking past discussions, it looks like we decided to re-visit how to model these.
-  https://github.com/apache/iceberg/pull/3887

In this PR, the partition column's name are derived from `{tranform}({colName})`,  I am open to other ways to representing this as well. 

TODO : check if this works well with LF

---

cc @jackye1995 @rajarshisarkar @amogh-jahagirdar @arminnajafi @xiaoxuandev @yyanyy @flyrain @szehon-ho @natsukawa-kanou


jackye1995:
Does partition key allow properties? The info was there in the conversion, and was then removed because partition spec can also be versioned and requires some sort of properties to track which ones are the latest.

PR #4776, opened 2022-05-16
Replace deprecated RuntimeIOException with UncheckedIOException

ajantha-bhat:
Depends on https://github.com/apache/iceberg/pull/6887

- [102 places] Replace deprecated `RuntimeIOException` with `UncheckedIOException` if the cause is already having an `IOException`.
- [9 places] Introduce `FileHandlingException` which extends `RuntimeException` for the cases without a cause for `RuntimeIOException`.
- Removed deprecated `RuntimeIOException` class

Fixes: https://github.com/apache/iceberg/issues/9208


api/src/main/java/org/apache/iceberg/Files.java
```diff
@@ -59,9 +59,9 @@ public PositionOutputStream create() {
       }
 
       if (!file.getParentFile().isDirectory() && !file.getParentFile().mkdirs()) {
-        throw new RuntimeIOException(
+        throw new UncheckedIOException(new IOException(String.format(
```
rdblue, line 62: Rather than creating an `IOException`, I think this should throw a different exception.

api/src/main/java/org/apache/iceberg/Files.java
```diff
@@ -24,8 +24,8 @@
 import java.io.RandomAccessFile;
 import java.nio.file.Paths;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
+import org.apache.iceberg.exceptions.FileHandlingException;
```
jbonofre, line 27: Don't we have an overlap between this PR and #6887 ? 
Maybe worth to define an order and gather both PRs in a single PR.

jbonofre (approved):
LGTM, but it seems with have a kind of overlap with #6887

api/src/main/java/org/apache/iceberg/Files.java
ajantha-bhat: Yeah. Description mentions that this PR depends on #6887

Idea is that, I will rebase this PR after #6887 is merged. 
Reviewer can easily review this PR as it will just blind replacement once I rebase.

jbonofre: Awesome, thanks :) It will be easier to review 👍

ajantha-bhat:
Closing as we concluded to undo deprecation.
Fixed by https://github.com/apache/iceberg/pull/5640

PR #4777, opened 2022-05-16
Remove deprecated table.rollback()

ajantha-bhat:
Remove deprecated `table.rollback()` and `RollbackToSnapshot`

TODO: update testcases for table.rollback()


kbendick:
How long have we had this deprecated @ajantha-bhat? Are you aware of any compatibility guarantees or any timeline we might be bound to before fully removnig this?

It looks as though this has been deprecated for a while, as the old interface is simply written in terms of the new one. But we don't want to break compatibility relatively quickly.

ajantha-bhat:
> How long have we had this deprecated @ajantha-bhat? Are you aware of any compatibility guarantees or any timeline we might be bound to before fully removnig this?

I am aware that we deprecate things and wait for another minor release (like things deprecated in 12.x can be removed in 13.x)

This one was deprecated in FEB 2020. So, I think we can safely remove it as we have alternate implementation.

ajantha-bhat:
Handled by https://github.com/apache/iceberg/pull/5639

PR #4784, opened 2022-05-16
Python: Set Python 3.8 as minimal version

Fokko:
Python 3.8 has some nice features, like:

- The Walrus operator := https://realpython.com/python-walrus-operator/
- Ability to cache properties: @cached_property https://docs.python.org/3/library/functools.html#functools.cached_property
- (this is a non-exhaustist list that we can append to over time)

3.7 has still support for a bit over a year: https://endoflife.date/python

I've added 3.10 to the CI as well. Not sure if we can deprecate 3.7 just yet since it is still being used a lot:

image

The Python version that downloads packages from PyPI

Pulled this from BigQuery: https://console.cloud.google.com/bigquery?p=bigquery-public-data&d=pypi&page=dataset

That being said, upgrading from 3.7 to 3.8 should be quite easy nowadays with minimal breaking changes on the Python side, and just swapping out a base docker image.


rdblue:
I'm fine with this change, since I think we still support 3.7 for the legacy API. It really comes down to what people in the community are using, but since this is the new Python library I don't think there are that many people using it yet.

danielcweeks:
I just merged #4767 which had a little logic around 3.7, which you might want to remove as part of this PR.

samredai:
I think this is ok. Security support will be dropped for 3.7 within a year which means in a few months, everyone will begin scrambling to get everything off of 3.7 anyway.

samredai (approved):

Fokko:
Thanks @danielcweeks for noticing. I've just updated the PR and removed all the Python 3.7 related stuff (including singledispatch). Less is more 💪🏻

kbendick (approved):
+1. And thank you for the graph of downloads! Very insightful.

kbendick:
Also, Dask requires 3.8+ so we're in good company.

PR #4417, opened 2022-03-28
Flink - Fix UPSERT delete file metadata [Backport to Flink 1.13]

kbendick:
Backports https://github.com/apache/iceberg/pull/4364 to Flink 1.13

Co-authored-by: Li Lliwei [hililiwei@gmail.com](mailto:hililiwei@gmail.com)

PR #4189, opened 2022-02-22
Flink 1.14: Fix the flaky testHashDistributeMode by ingesting all rows in one checkpoint cycle.

openinx:
This PR is trying to fix the flaky testHashDistributeMode unit test fundamentally.  The following are the explannation about the current fix.

1. https://github.com/apache/iceberg/pull/4117#issuecomment-1042701849
2. https://github.com/apache/iceberg/pull/4117#issuecomment-1042718844
3. https://github.com/apache/iceberg/issues/2575#issuecomment-1046845868

PR #3834, opened 2021-12-31
change delete logic

hameizi:
Previous delete logic is write all delete data in eq-delete file although there is same key in pos-delete file.  This PR change this logic to just write the delete data what is not exist in pos-delete file in eq-delete file.  And the old logic will send several delete but one insert of one primary key in one snapshot when we want reperform the data by streaming read, in some scene will cause error (like aggregete count). 
the following is difference between old delete logic and the new. 
table schema: 
```
int key;  (primary key) 
string data;
``` 
**old logic:** 
tx1: 
> insert (1,'aa'); 

tx2: 
>delete (1,'aa');   -->eq-delete file add (1,'aa') 
> insert(1,'bb'); 
>delete (1,'bb');   -->eq-delete file add (1,'bb') pos-delete file add(1,filepath) 
>insert(1,'cc') 

result: 
eq-delete file has (1,'aa'),(1,'bb') 
pos-delete file has (1,filepath) 
 
**new logic:**
tx1: 
> insert (1,'aa'); 

tx2: 
>delete (1,'aa');   -->eq-delete file add (1,'aa') 
> insert(1,'bb'); 
>delete (1,'bb');   --> pos-delete file add(1,filepath) 
>>insert(1,'cc') 

result: 
eq-delete file has (1,'aa') 
pos-delete file has (1,filepath) 
 
Actually the data (1,'bb') is unnecessary in eq-delete file, because when we call function applyPosdelete that (1,'bb') will be delete from result so there is not data match (1,'bb') when we call applyEqdelete.  In streaming read scene the old logic will send two delete (1,'aa'),(1,'bb') of key 1 but one insert (1,'cc') to down opeator, but new logic just send one delete and one insert.

PR #4786, opened 2022-05-16
[0.13] Flink upsert delete file metadata backports

nastra:
This backports #4417 / #3834 (because #4364 depends on it) / #4364 / #4189 (because CI was failing without this test fix) to the 0.13.x branch

### DIFF
--- a/flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/sink/BaseDeltaTaskWriter.java
+++ b/flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/sink/BaseDeltaTaskWriter.java
@@ -103,5 +103,10 @@ abstract class BaseDeltaTaskWriter extends BaseTaskWriter<RowData> {
     protected StructLike asStructLike(RowData data) {
       return wrapper.wrap(data);
     }
+
+    @Override
+    protected StructLike asStructLikeKey(RowData data) {
+      return wrapper.wrap(data);
+    }
   }
 }
--- a/core/src/main/java/org/apache/iceberg/io/BaseTaskWriter.java
+++ b/core/src/main/java/org/apache/iceberg/io/BaseTaskWriter.java
@@ -136,13 +136,16 @@ public abstract class BaseTaskWriter<T> implements TaskWriter<T> {
      *
      * @param key has the same columns with the equality fields.
      */
-    private void internalPosDelete(StructLike key) {
+    private boolean internalPosDelete(StructLike key) {
       PathOffset previous = insertedRowMap.remove(key);
 
       if (previous != null) {
         // TODO attach the previous row if has a positional-delete row schema in appender factory.
         posDeleteWriter.delete(previous.path, previous.rowOffset, null);
+        return true;
       }
+
+      return false;
     }
 
     /**
@@ -152,9 +155,9 @@ public abstract class BaseTaskWriter<T> implements TaskWriter<T> {
      * @param row the given row to delete.
      */
     public void delete(T row) throws IOException {
-      internalPosDelete(structProjection.wrap(asStructLike(row)));
-
-      eqDeleteWriter.write(row);
+      if (!internalPosDelete(structProjection.wrap(asStructLike(row)))) {
+        eqDeleteWriter.write(row);
+      }
     }
 
     /**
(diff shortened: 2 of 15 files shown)


nastra:
@kbendick could you please double-check if there are any other Flink changes that need to be backported?

flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/sink/BaseDeltaTaskWriter.java
```diff
@@ -103,5 +103,10 @@ protected class RowDataDeltaWriter extends BaseEqualityDeltaWriter {
     protected StructLike asStructLike(RowData data) {
       return wrapper.wrap(data);
     }
+
+    @Override
+    protected StructLike asStructLikeKey(RowData data) {
+      return wrapper.wrap(data);
```
nastra, line 109: @kbendick should this throw an exception like in https://github.com/apache/iceberg/blob/910f2712c7b731181c4a72fb6b8c66addeb76dd9/flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/sink/BaseDeltaTaskWriter.java#L109 or does it even matter what we return here for Flink 1.12?

kbendick: I would keep it this way upon initial inspection.

It needs to be implemented for the API and it will possibly get called. Keeping it the same as `asStructLike` seems safest for backwards compatibility.

rdblue (approved):

kbendick:
I'm not sure what's the deal with Flink 1.13 tests. Maybe the heap change needs to be merged in _first_? It seems though that the 1.13 tests are just... waiting.

image

rdblue:
The test issue looks like a github problem. I tried to cancel, but it looks like I can't right now. I'll check back in a few hours. It should time out.

nastra:
Yes the job is timing out but it's unclear why because the run itself doesn't show any output. Locally it helped to increase the Heap size for Gradle (since the build didn't progress as well locally)

kbendick:
So GH Actions use the settings in master for security purposes.

Mayhe merging in the heap increase _first_ will help? I doubt the tests are running with the heap increase if it’s in this PR. Just a guess but the tests didn’t complete without increased heap for me as well.

Also, since this is a release branch, feel free to make the heap slightly bigger and/or tune the forked JVM’s GC settings in the settings.gradle (or wherever).

But I think running with the heap settings already merged in will help the most. I think you can try it in your fork if you run the tests there.

kbendick:
Oh and the other thing would be to add class unloading into the test JVM settings. I helped tune the JVM for CI in Spark and class unloading, a fixed MaxMetaSpaceSize (to force class unloading), and somewhat larger heaps helped quite a bit for a situation that had gotten really out of hand-

Also my only meaningful contribution to the Spark repo and the credit got misattributed as I never added that email to my GitHub 😂

kbendick:
The only other issue I can think of is a GitHub issue with the JVM cache as the problem is sticky for Flink 1.13.

But the most recent attempt _did_ run somewhat so I don’t think that’s it. I think it’s just poor memory utilization / availability combined with some somewhat heavy Flink SQL upsert tests.

nastra:
I think I found the issue (which can easily be debugged locally by adding `--no-parallel --debug` to get `./gradlew -DsparkVersions= -DhiveVersions= -DflinkVersions=1.13 :iceberg-flink:iceberg-flink-1.13:check :iceberg-flink:iceberg-flink-runtime-1.13:check -Pquick=true -x javadoc --no-parallel --debug`). It looks like something got messed up during conflict resolution and `BaseDeltaTaskWriter#asStructLike(..)` was accidentally returning `keyWrapper.wrap(data)` and not `wrapper.wrap(data)`.
This then resulted in `TestChangeLogTable#testChangeLogOnDataKey()` to run indefinitely in a retry-loop because there was an underlying exception. 
For the future I think it would be good to have a timeout on tests as issues like these are quite difficult to debug on CI without modifying how the test is being executed (e.g. by adding `--debug` to the gradle task and such)

kbendick:
> I think I found the issue (which can easily be debugged locally by adding `--no-parallel --debug` to get `./gradlew -DsparkVersions= -DhiveVersions= -DflinkVersions=1.13 :iceberg-flink:iceberg-flink-1.13:check :iceberg-flink:iceberg-flink-runtime-1.13:check -Pquick=true -x javadoc --no-parallel --debug`). It looks like something got messed up during conflict resolution and `BaseDeltaTaskWriter#asStructLike(..)` was accidentally returning `keyWrapper.wrap(data)` and not `wrapper.wrap(data)`. This then resulted in `TestChangeLogTable#testChangeLogOnDataKey()` to run indefinitely in a retry-loop because there was an underlying exception. For the future I think it would be good to have a timeout on tests as issues like these are quite difficult to debug on CI without modifying how the test is being executed (e.g. by adding `--debug` to the gradle task and such)

It's possible there's somethign else we need to cherry-pick given the the underlying exception for that one test case. I'll run locally and we'll see.

PR #4792, opened 2022-05-17
Python: Add BotoFileIO, a FileIO that wraps boto3

samredai:
This adds a FileIO implementation that uses boto3. This lets you optionally provide a boto session to the FileIO instance, enabling some lower level AWS functionality like custom signing.


python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def __len__(self) -> int:
+        """Returns the total length of the file, in bytes"""
+        file_info = self._file_info()
+        return file_info.size
+
+    def exists(self) -> bool:
+        """Checks whether the location exists"""
+        try:
+            self._session.resource('s3').Bucket(self._bucket).Object(self._path).load()  # raises botocore.exceptions.ClientError with a 404 if it does not exist
```
Fokko, line 148: Would this also pull the content? That might be expensive

Fokko (commented):
Some small suggestions, but looks good in general.

How about testing? Would be awesome if we can spin up something like Minio for running some tests.

samredai:
@Fokko I was thinking the same thing. I'm working on a MinIO based test suite now and also will include some tests with mocked request/responses to test a custom signer.

python/src/iceberg/io/boto.py
joshuarobinson: According to this page, it's just a head: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Object.load

but maybe a "head_object()" call would be clearer?  (I too assumed load() would actually load the contents...)

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,294 @@
...
+        >>> data = input_stream.read()
+    
+    Example2:
+        >>> from iceberg.io.boto import BotoFileIO
+        >>> file_io = BotoFileIO()
+        >>> output_file = file_io.output_file("s3://foo/bar.txt")
+        >>> output_stream = output_file.create()
+        >>> output_stream.write(b'foo')
+    """
+    def __init__(self, session: Optional[Session] = None):
```
joshuarobinson, line 249: My understanding of the intention was the by passing the optional session, you can let the user to whatever magic configuration they needed (specifically endpoint_url).

But when I read [the documentation here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html), I get the impression that it's in fact the "client" or "resource" where this is configured, not the session.

maybe I misunderstand something?

python/src/iceberg/io/boto.py
samredai: Ah you're absolutely right. But since we use the session object directly to create resource clients we'll probably need to add an endpoint argument to the `BotoFileIO` init and then consistently pass it to `resource(...)` which also has an [endpoint_url](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.resource) argument. Let me update this, thanks @joshuarobinson!

python/src/iceberg/io/boto.py
samredai: Ah you're absolutely right. But since we use the session object directly to create resource clients we'll probably need to add an endpoint argument to the `BotoFileIO` init and then consistently pass it to `resource(...)` which also has an [endpoint_url](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.resource) argument. Let me update this, thanks @joshuarobinson!

joshuarobinson: maybe use a kwargs?

as a point of reference, this is how I initialize connections using s3fs:
`storage_options = {'endpoint_url': ENDPOINT_URL}
fs = fsspec.filesystem('s3', client_kwargs=storage_options)`

samredai: Love it, I added a **client_kwargs argument and pass it through to the resource

.github/workflows/python-ci.yml
```diff
@@ -50,5 +49,7 @@ jobs:
       run: |
         pip install -e .[dev]
         pip install -U tox-gh-actions
+    - name: Start MinIO instance
+      run: docker-compose -f python/docker-compose.ci.yml up -d
     - working-directory: ./python
-      run: tox
+      run: tox -- -m "marked or not marked"
```
samredai, line 55: I don't think we want the tests marked `s3` to run by default locally, so I added `-m "not s3"` in the tox.ini file. Here what I'm doing is overriding that setting for the CI.

I couldn't find a `-m` argument dedicated for directing pytest to run **all** tests, including unmarked tests, so I do a bit of a trick here: although there is no group of tests named `marked`, telling pytest to run "marked or not marked" results in all tests running.

python/README.md
```diff
@@ -38,6 +38,28 @@ Testing is done using tox. The config can be found in `tox.ini` within the pytho
...
+| Argument                | Default               | Description                                         |
+|-------------------------|-----------------------|-----------------------------------------------------|
+| --endpoint-url          | http://localhost:9000 | The S3 endpoint URL for tests marked as s3          |
+| --aws-access-key-id     | admin                 | The AWS access key ID for tests marked as s3        |
+| --aws-secret-access-key | password              | The AWS secret access key ID for tests marked as s3 |
+
+For example, to run `s3` tests against an object store running locally on port 4000 with an access key ID of `foo` and an secret access key as `bar`, you would use the following command.
+
+```
+tox -- -m "s3" --endpoint-url=http://localhost:4000 --aws-access-key-id=foo --aws-secret-access-key=bar
```
samredai, line 60: Using this, you could run a minio instance locally (or use an aws s3 bucket) and run the tests marked as `s3` against that.

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,294 @@
...
+            range_header = f"bytes={self._position}-{position_new -1}"
+            self.seek(offset=size, whence=1)
+
+        return self._s3_object.get(Range=range_header)["Body"].read()
+
+    def seek(self, offset: int, whence: Literal[0, 1, 2] = 0) -> None:
+        position_new = offset if whence == 0 else self._position + offset if whence == 1 else self._s3_object.content_length + offset if whence == 2 else None
+
+        if not position_new:
+            raise ValueError(f"Cannot seek to position {offset}, invalid whence: {whence}")
```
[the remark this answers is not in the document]
samredai: Done!

python/src/iceberg/io/base.py
```diff
@@ -44,7 +44,7 @@ class InputStream(Protocol):  # pragma: no cover
     def read(self, size: int) -> bytes:
```
Fokko, line 44: How about keeping it in line with BinaryIO?
```suggestion
    def read(self, n: int = -1) -> bytes:
```

python/src/iceberg/io/base.py
```diff
@@ -44,7 +44,7 @@ class InputStream(Protocol):  # pragma: no cover
...
         ...
 
-    def seek(self, offset: int, whence: int) -> None:
```
Fokko, line 47: This is now also in line with BinaryIO 👍🏻

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,290 @@
...
+            ):  # If more bytes are requested than exists, just read the entire file from the current position
+                return self.read(size=-1)
+
+            range_header = f"bytes={self._position}-{position_new -1}"
+            self.seek(offset=size, whence=1)
+
+        return self._s3_object.get(Range=range_header)["Body"].read()
+
+    def seek(self, offset: int, whence: int = 0) -> int:
+        if whence not in (0, 1, 2):
```
Fokko, line 67: Instead of using the magic numbers, we can also import them from `io`:
```python
from io import SEEK_SET, SEEK_CUR, SEEK_END 
```

```suggestion
        if whence not in {SEEK_SET, SEEK_CUR, SEEK_END}:
```

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError(f"Cannot seek to position {offset}, invalid whence: {whence}")
+
+        if whence == 0:
```
Fokko, line 70:
```suggestion
        if whence == SEEK_SET:
```

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,290 @@
...
+            self._position = offset
+        elif whence == 1:
```
Fokko, line 72:
```suggestion
        elif whence == SEEK_CUR:
```

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,290 @@
...
+            self._position = self._position + offset
+        elif whence == 2:
```
Fokko, line 74:
```suggestion
        elif whence == SEEK_END:
```

python/tox.ini
```diff
@@ -24,6 +24,7 @@ usedevelop = true
 extras = arrow
 deps =
     pip>=21.1
+    boto3
```
Fokko, line 27:
```suggestion
    boto3==1.23.10
```

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,290 @@
...
+        else:
+            position_new = self._position + size
+
+            if (
+                position_new >= self._s3_object.content_length
+            ):  # If more bytes are requested than exists, just read the entire file from the current position
+                return self.read(size=-1)
+
+            range_header = f"bytes={self._position}-{position_new -1}"
+            self.seek(offset=size, whence=1)
```
Fokko, line 62:
```suggestion
            self.seek(offset=size, whence=SEEK_CUR)
```

Fokko (approved):
Some bikeshedding, apart from that I like it a lot @samredai 👏🏻

python/src/iceberg/io/base.py
rdblue: What does `whence` do?

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,291 @@
...
+
+        This uses the `Range` argument when reading the S3 Object that allows setting a range of bytes to the headers of the request to S3.
+
+        Args:
+            n (int, optional): The number of bytes to read. Defaults to -1 which reads the entire file.
+
+        Returns:
+            bytes: The byte content of the file
+        """
+        if n == -1:  # Read the entire file from the current position
```
rdblue, line 51: Should any negative have this effect? Or should we raise ValueError if n < -1?

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,291 @@
...
+            range_header = f"bytes={self._position}-"
+            self.seek(offset=0, whence=2)
+        else:
+            position_new = self._position + n
```
rdblue, line 55: This is a problem if `n` is negative, right?

python/src/iceberg/io/base.py
```diff
@@ -41,10 +41,10 @@ class InputStream(Protocol):  # pragma: no cover
     implementation's `open(...)` method. These methods are a subset of IOBase/RawIOBase.
     """
 
-    def read(self, size: int) -> bytes:
+    def read(self, n: int = -1) -> bytes:
```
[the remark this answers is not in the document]
samredai: I used that to be consistent with the python IO built-ins. The value is described here: https://docs.python.org/3/library/io.html#io.RawIOBase.read

> Read up to size bytes from the object and return them. As a convenience, if size is unspecified or -1, all bytes until EOF are returned. Otherwise, only one system call is ever made. Fewer than size bytes may be returned if the operating system call returns fewer than size bytes.

python/src/iceberg/io/boto.py
samredai: Ah right, that case isn't accounted for here and will produce a strange range header. Let me update it to be `n <= -1`

python/src/iceberg/io/boto.py
samredai: If I change the first if block to `n <= -1`, this else should never get triggered when n is negative. Updating now!

python/README.md
rdblue: I think this is going to be out of date now that we're using pre-commit.

python/setup.cfg
```diff
@@ -45,6 +45,8 @@ python_requires = >=3.8
 install_requires =
     mmh3
 [options.extras_require]
+boto =
```
rdblue, line 48: This creates an optional profile, like `iceberg[boto]` right?

python/src/iceberg/io/base.py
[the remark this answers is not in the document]
rdblue: Sound good. We may want to consider adding docs that at least point to the canonical ones, but that's not something that should block this PR.

python/src/iceberg/io/base.py
rdblue: I see below.

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Args:
+            n (int, optional): The number of bytes to read. Defaults to -1 which reads the entire file.
+
+        Returns:
+            bytes: The byte content of the file
+        """
+        if n < 0:  # Read the entire file from the current position
+            range_header = f"bytes={self._position}-"
+            self.seek(offset=0, whence=2)
```
rdblue, line 53: Can we use constants for `whence`?

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,291 @@
...
+
+            if (
+                position_new >= self._s3_object.content_length
+            ):  # If more bytes are requested than exists, just read the entire file from the current position
+                return self.read(n=-1)
+
+            range_header = f"bytes={self._position}-{position_new -1}"
+            self.seek(offset=n, whence=SEEK_CUR)
+
+        return self._s3_object.get(Range=range_header)["Body"].read()
```
rdblue, line 65: Should we do any checking to ensure that Boto returned the expected number of bytes?

python/src/iceberg/io/boto.py
samredai: Fixed! Changed this to `whence=SEEK_END`.

python/setup.cfg
samredai: Yep, exactly

python/src/iceberg/io/boto.py
samredai: Fixed!

python/src/iceberg/io/boto.py
samredai: How would we handle the `-1` case where the entire file is requested? Or if EOF is reached before the requested number of bytes are retrieved. Were you thinking we'd just check `["ContentLength"]` for the case where more bytes than requested are returned and raise an error?

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,291 @@
...
+        return self._position
+
+    def tell(self) -> int:
+        return self._position
+
+    def closed(self) -> bool:
+        return False
+
+    def close(self) -> None:
+        pass
```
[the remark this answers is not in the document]
samredai: Yes that was my original thinking. See [this](https://github.com/apache/iceberg/pull/4792#discussion_r886314530) comment that described what I changed this to after your review.

python/src/iceberg/io/boto.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+
+    def __init__(self, location: str, session: Session, **client_kwargs):
+        parsed_location = urlparse(location)  # Create a ParseResult from the uri
+
+        if not parsed_location.scheme.startswith("s3"):
+            raise ValueError(f"Cannot create BotoInputFile, scheme not supported: {parsed_location.scheme}")
+
+        self._bucket = parsed_location.netloc
+        self._path = parsed_location.path.strip("/")
```
rdblue, line 146: This alters the path by removing any trailing `/` as well. Since this is an object and not a file system path, we don't want to do that (in S3, `a/` and `a` are distinct). Instead, I think you want to check for a starting `/` and remove only that:

```python
self.path = parsed_location.path[1:] if parsed_location.path.startswith('/') else parsed_location.path
```

rdblue:
@samredai, to review this, I started looking more into the boto3 API. It looks like the API that you're using isn't a streaming API, which is what we typically want so that we can avoid things like buffering whole files in memory before writing them with a single PUT. When I went looking more into how to use boto3 for streaming reads and streaming writes, I quickly ran into `smart_open`, which appears to do everything that we want.

I think you had a S3FileIO that used smart_open before. Is there a reason not to use that to wrap boto3 now? I think we would be able to avoid maintaining a lot of this code.

samredai:
@rdblue I agree this isn't optimal and I was planning to add the file-io you mentioned but renamed to `SmartOpenFileIO`. The main reason why I'm hoping to add this `BotoFileIO` is to enable using a custom AWS signer. I couldn't find a way to do this with PyArrow's s3 filesystem (cc: @emkornfield to confirm). I'll have to look into if this is possible with [smart_open](https://github.com/RaRe-Technologies/smart_open). If it is then I think it makes sense to go with that and close this PR.

rdblue:
@samredai, I saw that `smart_open` does allow you to pass your own S3 session or client. I think it should be possible.

samredai:
@rdblue Nice! I'll try that out and confirm everything works as expected. Then I'll open a `SmartOpenFileIO` PR and close this one.

emkornfield:
> The main reason why I'm hoping to add this BotoFileIO is to enable using a custom AWS signer. I couldn't find a way to do this with PyArrow's s3 filesystem (cc: @emkornfield to confirm).

Sorry for the late reply, this is beyond my area of expertise, so I'm not really sure but looking at the [C++ implementation](https://github.com/apache/arrow/blob/c72f84a48b4952796ab78a0c33b84a9fc8f893db/cpp/src/arrow/filesystem/s3fs.cc#L560) it seems that all references to signer are hard-coded.

samredai:
Closing since the same requirements are met by s3fs

PR #4795, opened 2022-05-17
expose the latest snapshot id committed within a thread

CodingCat:
this PR implements the functionality to expose the latest snapshot id within a thread. 

The implementation is motivated by our internal use cases where we have multiple threads submitting jobs with a shared SparkContext and we need to do some work with the latest snapshot within the thread as input 

I think the scenario is more pervasive than our own case, e.g. each notebook attached to the Databricks' notebook cluster is basically handled by a thread. In such an scenario, users may fall into some race condition to get the snapshot id committed by their own notebook with just `currentSnapshot().snapshotId`. Because currentSnapshot() will  trigger the refresh of metadata and may show the snapshot id committed by someone else in another concurrent thread eventually


CodingCat:
@kbendick, @jackye1995, @nastra, @aokolnychyi, @danielcweeks, @rdblue  would you mind giving a review? thanks!

kbendick:
Hi @CodingCat 

I’m trying to understand the situation you’re trying to solve for a bit more.

> Because currentSnapshot() will trigger the refresh of metadata and may show the snapshot id committed by someone else in another concurrent thread eventually

As mentioned on Slack, the metadata refresh on commit is to ensure that the state of the table is the same as it was when the write was prepared. This is how ACID compliance is achieved.

I’m not sure I understand what you’re trying to achieve. I know you’d like to expose the snapshotId as it was when the current thread (or let’s just say writer) prepared it’s write, ie prior to the commit. But what do you intend to do with that information?

> I think the scenario is more pervasive than our own case, e.g. each notebook attached to the Databricks' notebook cluster is basically handled by a thread. In such an scenario, users may fall into some race condition to get the snapshot id committed by their own notebook with just currentSnapshot().snapshotId

What catalog are you using? You mention Databricks, and most people I’ve encountered using Iceberg on Databricks are using the `HadoopCatalog`. Which should _not_ be used in a production environment as there’s no locking mechanism to keep the current snapshot updateable by only one writer at a time (be it across threads or across Spark applications).

It sounds like maybe you’re trying to get around the lack of a lock, but I worry that you’ll have conflicting writes and clobber the previous writers work.

What do you intend to do with this thread local snapshot Id (particularly once it becomes outdated via some other writer).

CodingCat:
Hi, @kbendick 

Thanks for looking at this! I found there are some confusion about what I want to achieve here, please check the inlined answer

> I’m not sure I understand what you’re trying to achieve. I understand you’d like to expose the snapshotId as it was when the current thread (or let’s just say writer) prepared its write, ie prior to the commit.

I am trying to expose the snapshot id *AFTER* the commit, you can check the change at [here](https://github.com/apache/iceberg/pull/4795/files#diff-3c27b7c7615fcf206c99e3e441c0621561c0a19ac3e9a26bb142329879425151R332
) 

the problem I am trying to resolve is from the following workflow

```
ThreadA committed 

ThreadB committed

ThreadA want to do something with the snapshot it just committed

```

as I said in the description, currently there is no way for ThreadA to do this, as long as it calls currentSnapshotId(), it will get the latest snapshotId generated by ThreadB because of the refreshing of metadata. this PR is just to expose this information to ThreadA

it will not break ACID, as if ThreadA committed another version, it will be still based on the latest snapshot id since I didn't change that part at all


> 
> > I think the scenario is more pervasive than our own case, e.g. each notebook attached to the Databricks' notebook cluster is basically handled by a thread. In such an scenario, users may fall into some race condition to get the snapshot id committed by their own notebook with just currentSnapshot().snapshotId
> 
> What catalog are you using? You mention Databricks, and most people I’ve encountered using Iceberg on Databricks are using the `HadoopCatalog`. Which should _not_ be used in a production environment as there’s no locking mechanism to keep the current snapshot updateable by only one writer at a time (be it across threads or across Spark applications).
> 
> It sounds like maybe you’re trying to get around the lack of a lock, but I worry that you’ll have conflicting writes and clobber the previous writers work.

I am using HadoopCatalog but not really getting around of lack of lock here (the lock problem was resolved by https://github.com/apache/iceberg/pull/3663 where you can use some DynamoDB based lockManager with HadoopTableOperation


> 
> What do you intend to do with this thread local snapshot Id (particularly once it becomes outdated via some other writer).

we have multiple scenarios needing this local snapshot id, the simplest example might be we have multiple threads committing to the same table, we want to logging this information for easier debugging (with information like which thread generated which version)

kbendick:
> I am using HadoopCatalog but not really getting around of lack of lock here (the lock problem was resolved by https://github.com/apache/iceberg/pull/3663 where you can use some DynamoDB based lockManager with HadoopTableOperation

Ok. That's good you're using a lock. I'm not sure how well exercised that code path is in practice, but you are using it. So that's good.

> we have multiple scenarios needing this local snapshot id, the simplest example might be we have multiple threads committing to the same table, we want to audit this information (with information like which thread generated which version)...

Is there another possible way to solve this problem? I do worry that exposing the snapshot ID this way will be dangerous for other users, who might use the (outdated) snapshot ID for the wrong purpose. I understand that _you_ are using it for purposes that sound like they don't violate any integrity concerns (based on what you said), but we have to think of the general case. Exposing it as simply the latest snapshot id in the thread leaves it open to the user to have to do the correct thing with the no longer current snapshot ID.

If you add the thread ID to the snapshot summary, you'd be able to query the `snapshots` table and see it there. Is that a possibility?

Another possibility I can think of is emitting an event on commit.

The TLDR of my concern is that while you might very well know what you're doing with this information, there will likely be others who come along and do not use it properly.

kbendick:
Actually, I think there's an existing way that the thread ID could be exposed in the snapshot summary. Look into the `SnapshotSummary.EXTRA_METADATA_PREFIX`. Basically, in the spark `options` used on write, you'd add `snapshot-property.${desired_summary_key}` = `${desired_value_for_key}`. And then on write, in the snapshot summary, `desired_summary_key = desired_value_for_key` will be writen.

CodingCat:
> Actually, I think there's an existing way that the thread ID could be exposed in the snapshot summary. Look into the `SnapshotSummary.EXTRA_METADATA_PREFIX`. Basically, in the spark `options` used on write, you'd add `snapshot-property.${desired_summary_key}` = `${desired_value_for_key}`. And then on write, in the snapshot summary, `desired_summary_key = desired_value_for_key` will be written.
> 
> See here where the options are added:
> 
> https://github.com/apache/iceberg/blob/feeb94b255b705382282e04fbeb1efdc3210d6cc/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L168-L175
> 
> And here where the key is defined:
> 
> https://github.com/apache/iceberg/blob/feeb94b255b705382282e04fbeb1efdc3210d6cc/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L60

well, the reason I didn't go into this obviously easier path is that we there is no way for me to add this snapshot tag when we do DELETE and UPSERT (or I didn't find it?)

CodingCat:
I am wondering what type of risk the user may run into when he got a local snapshot id? when working on this PR, my consideration is that as long as iceberg's ACID model doesn't break, users cannot use this snapshot id to do anything except read it...

rdblue:
@CodingCat, I agree with @kbendick that the right way to do this is to add metadata to the snapshot so that you can identify it later. Also, if you already have an operation that is going to commit a snapshot (a `SnapshotProducer`) then you can use `apply` to write the changes and return the uncommitted snapshot. Here's an example:

```java
AppendFiles append = table.newAppend()
...
long snapshotId = append.apply().snapshotId();
append.commit();
```

CodingCat:
ok, but could you share if there is any existing way for me to add metadata for DELETE and MERGE operation ? (where I have no way to do something like `df.write.option("snapshot-summary.key", "value")`?)

CodingCat:
after looking at the code again, i don't think there is any existing way for me to do this, 

I would like to work on a PR to implement something similar to https://github.com/delta-io/delta/blob/8974685f574394628cf9b5c5457093e7cbcd3d23/core/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L62-L66, where you can set something in the sessionstate.conf and then when we commit from SparkWrite take that information and set with SnapshotUpdate.set(k, v), how does it sound to you ?

CodingCat:
ping @kbendick @rdblue ?

CodingCat:
Hi, @rdblue , thanks for the reply

> The approach you linked to doesn't work because that is in the SQLConf, which is at the session level. You'd need a separate session for every writer thread. If that's what you'd like to do, then we can help you get that feature in. Is that what you want?

yes, this is pretty much what we did with delta lake, every thread creates a new SparkSession inherited from the parent one, and then add metadata in SQLConf...I am not a big fan of this approach since it imposed creation of a bunch of new SparkSessions across threads, that's also why I had the way in the current PR

> 
> I think that we may want to introduce a different way than using write properties to set additional metadata, because of the SQL problem. We could add a way to pass this information, like this:
> 
> ```java
>   public class CommitMetadata {
>     private static final ThreadLocal<Map<String, String>> COMMIT_PROPERTIES = ThreadLocal.withInitial(ImmutableMap::of);
> 
>     public static <R> R withCommitProperties(Map<String, String> properties, Callable<R> callable) throws Exception {
>       COMMIT_PROPERTIES.set(properties);
>       try {
>         return callable.call();
>       } finally {
>         COMMIT_PROPERTIES.set(ImmutableMap.of());
>       }
>     }
> 
>     static Map<String, String> commitProperties() {
>       return COMMIT_PROPERTIES.get();
>     }
>   }
> ```
> 

yeah, this looks neat! I can work on a PR for this

CodingCat:
just made a PR at https://github.com/apache/iceberg/pull/4956, closing this one

PR #4796, opened 2022-05-18
[DO NOT MERGE] Test that CI detects breaking changes

kbendick:
The `revapi` plugin runs on `gradle check`, which is already part of CI.

However, given that we run with several flags (particularly `-Pquick`), I'm opening this PR to quickly check if CI will detect this break or if we should add an additional github action.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/Accessor.java
+++ b/api/src/main/java/org/apache/iceberg/Accessor.java
@@ -23,6 +23,10 @@ import java.io.Serializable;
 import org.apache.iceberg.types.Type;
 
 public interface Accessor<T> extends Serializable {
+
+  // Apply a breaking change to ensure that it is reflected in our CI.
+  public static final long serialVersionUID = 42L;
+
   Object get(T container);
 
   Type type();
--- a/api/src/main/java/org/apache/iceberg/Accessor.java
+++ b/api/src/main/java/org/apache/iceberg/Accessor.java
@@ -23,10 +23,6 @@ import java.io.Serializable;
 import org.apache.iceberg.types.Type;
 
 public interface Accessor<T> extends Serializable {
-
-  // Apply a breaking change to ensure that it is reflected in our CI.
-  public static final long serialVersionUID = 42L;
-
   Object get(T container);
 
   Type type();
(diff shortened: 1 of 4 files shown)


kbendick:
Here's the breaking output
```
 ./gradlew revapi
> Task :iceberg-api:revapi FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':iceberg-api:revapi'.
> There were Java public API/ABI breaks reported by revapi:

  java.method.addedToInterface: Method was added to an interface.

  old: <none>
  new: method org.apache.iceberg.ManageSnapshots org.apache.iceberg.ManageSnapshots::someBreakingChange(java.lang.String, long)

  SOURCE: BREAKING, BINARY: NON_BREAKING, SEMANTIC: POTENTIALLY_BREAKING

  From old archive: <none>
  From new archive: iceberg-api-0.14.0-SNAPSHOT.jar

  If this is an acceptable break that will not harm your users, you can ignore it in future runs like so for:

    * Just this break:
        ./gradlew :iceberg-api:revapiAcceptBreak --justification "{why this break is ok}" \
          --code "java.method.addedToInterface" \
          --new "method org.apache.iceberg.ManageSnapshots org.apache.iceberg.ManageSnapshots::someBreakingChange(java.lang.String, long)"
    * All breaks in this project:
        ./gradlew :iceberg-api:revapiAcceptAllBreaks --justification "{why this break is ok}"
    * All breaks in all projects:
        ./gradlew revapiAcceptAllBreaks --justification "{why this break is ok}"
  ----------------------------------------------------------------------------------------------------


* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 6s
6 actionable tasks: 4 executed, 2 up-to-date
```

kbendick:
So it doesn't seem to be catching the failure, but the command ` ./gradlew -DflinkVersions=1.13,1.14,1.15 -DsparkVersions=2.4,3.0,3.1,3.2 -DhiveVersions=2,3 build -x test -x javadoc -x integrationTes` when run locally _does_ cause a failure. But it passes in CI.

Maybe it's due to caching in a previous step? I'm going to just add a workflow for revapi itself in the meantime.

kbendick:
Looking at the Java CI, I see
```
> Task :iceberg-api:revapiAnalyze SKIPPED
> Task :iceberg-api:revapi SKIPPED
```

kbendick:
Closed as we have a specific action for this and it is ensured to work.

PR #4797, opened 2022-05-18
Python - Fix tox errors for not calling super __init__ on a dataclass

kbendick:
Tox tests for python 3.8 are failing, as the `StructType` is now a `dataclass`, but its parent class is not a dataclass and it has `init=False`.

To get around this, I've called `super().__init__()` in the post init method, as suggested here: https://stackoverflow.com/questions/66827839/super-classs-init-is-not-called-using-dataclass

It's also possible we want to just set an ignore on this.

cc @Fokko @samredai

### DIFF
--- a/python/src/iceberg/types.py
+++ b/python/src/iceberg/types.py
@@ -203,7 +203,7 @@ class StructType(IcebergType):
         cls._instances[fields] = cls._instances.get(fields) or object.__new__(cls)
         return cls._instances[fields]
 
-    def __init__(self, *fields: NestedField, **kwargs):
+    def __init__(self, *fields: NestedField, **kwargs):  # pylint: disable=super-init-not-called
         if not fields and "fields" in kwargs:
             fields = kwargs["fields"]
         object.__setattr__(self, "fields", fields)


kbendick:
Here's some example breaking output:
```
 linters run-test: commands[3] | pylint src tests
  ************* Module iceberg.types
  src/iceberg/types.py:206:4: W0231: __init__ method from base class 'IcebergType' is not called (super-init-not-called)
```

python/src/iceberg/types.py
```diff
@@ -208,6 +208,10 @@ def __init__(self, *fields: NestedField, **kwargs):
             fields = kwargs["fields"]
         object.__setattr__(self, "fields", fields)
 
+    # This __post_init__ call is needed as IcebergType is not a dataclass
+    def __post_init__(self):
+        super().__init__()
```
samredai, line 213: What happens if we call this in the init method right above? Does that interfere with setting this to a frozen dataclass?

kbendick: Yes, I believe that interferes with setting this to a frozen dataclass.

See this SO (the answer at the bottom): https://stackoverflow.com/questions/66827839/super-classs-init-is-not-called-using-dataclass

As well as this example: https://github.com/ericvsmith/dataclasses/issues/142

kbendick: To be honest, I'm not actually sure. I _think_ we might be able to safely just ignore this, but this was the suggested route so I went with it.

kbendick: For now, let's just add a single ignore statement and then come back to this so that CI is passing in the master branch.

danielcweeks (approved):
Thanks @kbendick !

Fokko:
Thanks for fixing this @kbendick

python/src/iceberg/types.py
a-recknagel: Just in case it ever does come up, `IcebergType ` can be protected from having frozen-dataclass-children by using `object.__setattr__` for assignments instead of `=`, e.g. instead of `self.x = x` do `object.__setattr__(self, 'x', x)`.

PR #4798, opened 2022-05-18
Infra - Add an explicit workflow for API binary compatibility checks

kbendick:
The API / ABI compatibility checking tool, `revapi`, is supposed to run whenever we run `./gradlew build` or `./gradlew check`.

However, for some reason - likely due to caching, this is getting skipped as can be verified in this example PR: https://github.com/apache/iceberg/pull/4796

That PR shows that the output of `./gradlew -DflinkVersions=1.13,1.14,1.15 -DsparkVersions=2.4,3.0,3.1,3.2 -DhiveVersions=2,3 build -x test -x javadoc -x integrationTest`, which is run during `Java CI`, should run `revapi`. However, `revapi` is being skipped in that step per the log output.

For the moment, I'm adding an explicit `API Binary Compatibility` workflow to circumvent this issue.

cc @rdblue

### DIFF
--- a/.github/workflows/api-binary-compatibility.yml
+++ b/.github/workflows/api-binary-compatibility.yml
@@ -36,20 +36,16 @@ concurrency:
 jobs:
   revapi:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        jvm: [11]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v1
         with:
-          java-version: ${{ matrix.jvm }}
+          java-version: 11
       - uses: actions/cache@v2
         with:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
-      - run: echo -e "$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)\t$(hostname -f) $(hostname -s)" | sudo tee -a /etc/hosts
       - run: ./gradlew :iceberg-api:revapi
       - uses: actions/upload-artifact@v2
         if: failure()
--- a/.github/workflows/api-binary-compatibility.yml
+++ b/.github/workflows/api-binary-compatibility.yml
@@ -1,0 +1,61 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


.github/workflows/api-binary-compatibility.yml
```diff
@@ -0,0 +1,61 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  revapi:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jvm: [8, 11]
```
rdblue, line 41: Is this necessary? I don't think that binary compatibility changes that much across JVM versions. It should be fine to run with just 11.

.github/workflows/api-binary-compatibility.yml
```diff
@@ -0,0 +1,61 @@
...
+    env:
+      SPARK_LOCAL_IP: localhost
```
rdblue, line 43: This shouldn't run any Spark jobs. Is this needed?

.github/workflows/api-binary-compatibility.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.jvm }}
+      - uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - run: echo -e "$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)\t$(hostname -f) $(hostname -s)" | sudo tee -a /etc/hosts
```
rdblue, line 54: This looks like it was probably pasted from another workflow.

.github/workflows/api-binary-compatibility.yml
kbendick: No it's not needed.

.github/workflows/api-binary-compatibility.yml
kbendick: This is in pretty much _every_ workflow that we have that I've checked. But I don't think it's needed. Let's try without and see. We can always add it in.

rdblue (approved):

rdblue:
Thanks, @kbendick! Do we also need to remove the checks from running in another workflow?

PR #4799, opened 2022-05-18
Core, Spark: Use MultiThreading in abort and clean-up

singhpk234:
This PR attempts to use MultiThreading where ever possible, Some scenarios where could be benefitial : 
1. droptTable currently deletes (mainfest-files, mainfest-list, previous meta-data) in single thread, As per my understanding using MultiThreading can be beneficial here.
2. Spark Writer aborts can also be sped up using MT, thus making it faster.

once FileIO's such as S3FileIO which supports bulkDeletes are integrated to these flows, these ops could be even more optimized :) !!!

cc @rdblue @ConeyLiu

### DIFF
--- a/core/src/main/java/org/apache/iceberg/CatalogUtil.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogUtil.java
@@ -102,16 +102,19 @@ public class CatalogUtil {
     }
 
     Tasks.foreach(Iterables.transform(manifestsToDelete, ManifestFile::path))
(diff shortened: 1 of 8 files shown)


ConeyLiu:
Thanks, @singhpk234 for doing this. This looks fine to me.

amogh-jahagirdar:
LGTM as well @singhpk234

amogh-jahagirdar (approved):

jackye1995 (approved):
looks good to me!

PR #4687, opened 2022-05-03
Spark commit unknown handling

RussellSpitzer:
The issue here was pointed out by @stevenzwu when we were discussing the issue with incorrect aborts when non-runtime exceptions were thrown. He correctly noticed that the same sort of issue would appear if we threw a commit unknown exception and the Spark Writer aborted and deleted underlying data files.

Closes #4686

PR #4800, opened 2022-05-18
Spark: Extend commit unknown exception handling to SparkPositionDeltaWrite

singhpk234:
### About the change

Extend the handling for https://github.com/apache/iceberg/pull/4687 for SparkPositionDeltaWriter used in MOR.

### Testing Done 
Added an E2E UT as done in earlier PR.

---

cc @RussellSpitzer @aokolnychyi @flyrain

### DIFF
--- a/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMergeOnReadDelete.java
+++ b/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMergeOnReadDelete.java
@@ -19,9 +19,24 @@
 
 package org.apache.iceberg.spark.extensions;
 
+import java.io.IOException;
 import java.util.Map;
+import org.apache.iceberg.AssertHelpers;
+import org.apache.iceberg.RowDelta;
+import org.apache.iceberg.Table;
 import org.apache.iceberg.TableProperties;
+import org.apache.iceberg.exceptions.CommitStateUnknownException;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.spark.source.SparkTable;
+import org.apache.iceberg.spark.source.TestSparkCatalog;
+import org.apache.spark.SparkException;
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException;
+import org.junit.Test;
+
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
 
 public class TestMergeOnReadDelete extends TestDelete {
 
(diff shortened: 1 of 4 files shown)


RussellSpitzer:
@singhpk234 you probably saw how I wrote the test in the other suite, it is not pretty but I think you maybe be able to pull something off like that as well. Although this probably won't be easy unless it is coupled with a Spark version that natively supports the merge operations (3.3 I believe @aokolnychyi has it merged in) since at the moment we would need custom plan rules as well I think ...

 I'm getting close to thinking we need byteman or some other bytecode fault injector :(.

singhpk234:
>since at the moment we would need custom plan rules as well I think

+1, I was also thinking in these lines, we need here analyzer rules so that we could get a DSV2Relation which hold our mocked table object (which would throw our CommitUnknowEx when commit is called). The main problem as you highlighted is the same, it supported in SQL via our Extensions :(.

singhpk234:
@RussellSpitzer , I went one step above and used catalog to manipulate and return the spy table I want. Used static prop to manipulate to get the spy spark table instead :) !!!

RussellSpitzer:
@singhpk234 Thats a great solution! Maybe i'l rework my PR later to do the same thing!

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMergeOnReadDelete.java
```diff
@@ -37,4 +52,51 @@ protected Map<String, String> extraTableProperties() {
...
+    RowDelta spyNewRowDelta = spy(newRowDelta);
+    doAnswer(invocation -> {
+      newRowDelta.commit();
+      throw new CommitStateUnknownException(new RuntimeException("Datacenter on Fire"));
+    }).when(spyNewRowDelta).commit();
+
+    Table spyTable = spy(table);
+    when(spyTable.newRowDelta()).thenReturn(spyNewRowDelta);
+    SparkTable sparkTable = new SparkTable(spyTable, false);
+
```
RussellSpitzer, line 78: nit: double space

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkCatalog.java
```diff
@@ -43,4 +49,8 @@ public Table loadTable(Identifier ident) throws NoSuchTableException {
 
     return new SparkTable(table, false);
   }
+
+  public static void setDummyIcebergTbl(Table dummyTbl) {
```
RussellSpitzer, line 53: Probably not needed, but you could do the similar map of "Identifier -> Table" and just allow our tests to just insert new tables into that map and then request them on demand. Then have a method clear it out after each test. Just thinking about this for tests which run in multiple parameterizations (which will have the same static dummyIcebergTbl)

I know i'm biased since this is how I did it my other example, I like your placement in the Catalog here much more than mine though.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMergeOnReadDelete.java
```diff
@@ -37,4 +52,51 @@ protected Map<String, String> extraTableProperties() {
         TableProperties.DELETE_MODE, "merge-on-read"
     );
   }
+
+
```
flyrain, line 56: Nit: double space

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkCatalog.java
singhpk234: +1, Agree with you, present impl allows me to set only 1 tbl per UT, having a map and doing a look-up can allow me creating more than 1 tbl at given time :) !!!

Made the changes

stevenzwu (approved):

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java
```diff
@@ -224,7 +227,7 @@ private Expression conflictDetectionFilter(SparkBatchQueryScan queryScan) {
     @Override
     public void abort(WriterCommitMessage[] messages) {
       for (WriterCommitMessage message : messages) {
-        if (message != null) {
+        if (message != null && cleanupOnAbort) {
```
RussellSpitzer, line 230: minor nit here, but we can but this outside the loop

RussellSpitzer (approved):

flyrain (approved):

rdblue:
Thanks, @singhpk234!

PR #4806, opened 2022-05-18
Fix invalid Stream to Iterable conversion in decryption

findepi:
`Stream.iterator` is a one-shot operation, cannot be called multiple
times, so `stream::iterator` shouldn't be used as an implementation of
`Iterable`.

While we could easily change the `InputFilesDecryptor` to implement
`Iterable` correctly, it wouldn't convey the intention that the data is
read at most once. To convey that, the encryption API is changed from
accepting `Iterable` to accepting `Iterator`.

Similar to https://github.com/apache/iceberg/pull/4863

### DIFF
--- a/api/src/main/java/org/apache/iceberg/encryption/EncryptionManager.java
+++ b/api/src/main/java/org/apache/iceberg/encryption/EncryptionManager.java
@@ -20,9 +20,11 @@
 package org.apache.iceberg.encryption;
 
 import java.io.Serializable;
+import java.util.Iterator;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterators;
 
 /**
  * Module for encrypting and decrypting table data files.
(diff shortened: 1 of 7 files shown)


hililiwei (approved):

findepi:
@openinx is it going in the right direction?

danielcweeks:
@findepi Do you have more context on what the issue is here.  I may be reading this wrong but the new and old approaches appear to be equivalent.  Both approaches result in a function returning an iterator that is only traversed once.  If the intent is to allow for multiple iterations, `.collect(Collections.toList())` would be preferable to avoid recomputing.

findepi:
@danielcweeks 

The point is that this is not correct Java code
```java
Stream s = ...
Iterable i = s::iterator;
```

it compiles, because `Stream::iterator` method handle can be fit as an implementation of the `Iterable` interface, but it's not a correct implementation of that interface. `Iterable.iterator` method can be invoked multiple times, while `s.iterator` only once.

The change fixes the above, but i see also the downside: it's now possible to invoke the costly operation twice without code failing. Before the change, the Stream would probably report some exception on the second traversal.

> f the intent is to allow for multiple iterations, .collect(Collectors.toList()) would be preferable to avoid recomputing.

I think the interface intent was to avoid loading all data into memory at once. 

A yet different approach would be to let `encryption.decrypt` take `Iterator` (or `Stream`) instead of `Iterable.`
I think this one is actually best, but would prefer to have buy-in first before going that route.

danielcweeks:
@findepi Yeah, I agree that the existing implementation does not adhere to the `Iterable` interface contract. 

I just wasn't sure if this was breaking or just an improvement to address the contract.

I'd be fine with either the Iterator or Stream approaches as it's explicit that it'll be traversed only once.

findepi:
> I'd be fine with either the Iterator or Stream approaches as it's explicit that it'll be traversed only once.

@danielcweeks changed to Iterator.

PTAL

findepi:
@danielcweeks who can merge this?

api/src/main/java/org/apache/iceberg/encryption/EncryptionManager.java
```diff
@@ -47,11 +49,26 @@ public interface EncryptionManager extends Serializable {
    * By default this calls the single-file decryption method for each element in the iterator.
    * Implementations can override this for a variety of optimizations. For example, an
    * implementation can perform lookahead on the input iterator and fetch encryption keys in batch.
+   *
+   * @deprecated Use {@link #decrypt(Iterator)}.
    */
+  @Deprecated
```
rdblue, line 55: This is a public API. We shouldn't make changes to a public API just because it is being called incorrectly in some places.

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/ArrowReader.java
```diff
@@ -268,10 +268,10 @@ private static final class VectorizedCombinedScanIterator implements CloseableIt
           .map(entry -> EncryptedFiles.encryptedInput(io.newInputFile(entry.getKey()), entry.getValue()));
 
       // decrypt with the batch call to avoid multiple RPCs to a key server, if possible
-      Iterable<InputFile> decryptedFiles = encryptionManager.decrypt(encrypted::iterator);
+      Iterator<InputFile> decryptedFiles = encryptionManager.decrypt(encrypted.iterator());
```
rdblue, line 271: While this isn't technically correct, I think it is better not to change a public API. If you think this is a significant problem, then we have to options:
1. Collect `encrypted` into a list
2. Create a `OneTimeIterable` to wrap `Stream::iterator` instead

PR #4808, opened 2022-05-18
Core: Clear unused functions of HadoopTables

southernriver:
Clear unused functions of HadoopTables .


CodingCat:
since this was exposed as public APIs, how we assume no users use these stuffs?

southernriver:
> since this was exposed as public APIs, how we assume no users use these stuffs?

I thought there is no need to expose two identical duplicated functions, and keep one of them is ok.  

```
default Transaction newReplaceTableTransaction(
      TableIdentifier identifier,
      Schema schema,
      PartitionSpec spec,
      String location,
      Map<String, String> properties,
      boolean orCreate) 
```

CodingCat:
well, it is not about needed or not, but more about how we manage breaking changes

rdblue:
These are public, so I don't think we should remove them.

PR #3852, opened 2022-01-06
Core, Spark: Fallback when snapshot does not have schema id

wypoon:
In `SnapshotUtil`, when a snapshot does not have a schema id (written before schema id was added to snapshots), we fall back to reading each of the previous metadata files until we find one whose current snapshot id matches the snapshot id we seek, and read its schema from there.

We introduce a setting for testing purposes that makes the `SnapshotParser` write JSON without `schema-id` for snapshots. We add variants of existing tests for reading snapshots after schema evolution where the metadata is written without `schema-id` in the snapshots. The tests fail without the change in `SnapshotUtil`.

PR #4809, opened 2022-05-18
Core: Recover the schema by reading previous metadata files

southernriver:
If snapshot was created before Iceberg added schema id to snapshot，schemaId could be null,  we could recover the schema by reading previous metadata files

### DIFF
--- a/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
+++ b/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
@@ -307,7 +307,10 @@ public class SnapshotUtil {
       return schema;
     }
 
-    // TODO: recover the schema by reading previous metadata files
+    snapshot = table.snapshot(snapshot.parentId());
+    if (snapshot != null) {
+      return table.schemas().get(snapshot.schemaId());
+    }
     return table.schema();
   }
 


core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -307,7 +307,10 @@ public static Schema schemaFor(Table table, long snapshotId) {
       return schema;
     }
 
-    // TODO: recover the schema by reading previous metadata files
+    snapshot = table.snapshot(snapshot.parentId());
+    if (snapshot != null) {
+      return table.schemas().get(snapshot.schemaId());
```
kbendick, line 312: Is it possible that this can throw?

It would be great to add a test case that covers this situation.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -307,7 +307,10 @@ public static Schema schemaFor(Table table, long snapshotId) {
       return schema;
     }
 
-    // TODO: recover the schema by reading previous metadata files
+    snapshot = table.snapshot(snapshot.parentId());
```
singhpk234, line 310: [minor] should we rename this`parentSnapshot` ?

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
singhpk234, line 312: [question] can it be possible that even parentSnapshot is also lacking schemaId ? 

If this is true we can now return `null` earlier we were returning `table.schema()` instead. WDYT ?

wypoon:
I don't think this is the way to do it. If the snapshot is written before schema ids were introduced, then most likely its ancestors also do not have schema ids, unless different versions of Iceberg are in use concurrently, and one is of a version before schema ids were introduced.
I put up #3852 months ago to perform the recovery, and also added tests for the recovery. @kbendick can you please review?

PR #4812, opened 2022-05-18
Spark 3.2: Support reading position deletes

szehon-ho:
This exposes position deletes as a metadata table "position_deletes", with schema:  file, pos, row, partition

This will be useful when trying to implement "RewritePositionDeleteFiles", as we will read positional deletes from Spark and then write it.  It will also be useful to implement "RemoveDanglingDeleteFiles", ie removing delete files that no longer reference live data files.  It will also be generally useful to get more insights into positional delete files as a table via SQL.

Notes:

1. Design choice:  Why via metadata table?  Initially I tried to implement it as a Spark Read config, but  SparkCatalog.loadTable didnt support read configuration options to load SparkTable with an alternate schema.  So hence chose metadata table path.
2. Implementation: Most of the changes here are adding the concept of DeleteFileScanTask.  The FileScanTask is today bound to DataFiles (FileScanTask.file() returns DataFile) and we can't really change that as it's used in hundreds of places, so added a contentFile() method that returns the DataFile for DataFileScanTask or DeleteFile in case of DeleteFileScanTask.  To support position delete file scan, change code calls that need to scan delete files from FileScanTask.file() => FileScanTask.contentFile(), as all logics should work equally with both DeleteFile and DataFile.


.palantir/revapi.yml
```diff
@@ -27,6 +27,12 @@ acceptedBreaks:
     - code: "java.method.addedToInterface"
       new: "method ThisT org.apache.iceberg.SnapshotUpdate<ThisT>::scanManifestsWith(java.util.concurrent.ExecutorService)"
       justification: "Accept all changes prior to introducing API compatibility checks"
+    - code: "java.method.addedToInterface"
+      new: "method org.apache.iceberg.ContentFile<?> org.apache.iceberg.FileScanTask::contentFile()"
+      justification: "New method added to FileScanTask.  Should be not many implementations\
+        \ outside Iceberg, if so it's easy to fill in new method (especially if they\
+        \ inherit BaseFileScanTask).  This is easiest way to add DeleteFileScanTask,\
+        \ otherwise even bigger interface change is necessary"
```
kbendick, line 35: Side note: I'm looking for a way to suppress `addedToInterface` in addition to a few other non-breaking changes.

The way we release Iceberg, it's not really something that we need to worry about much.

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
```diff
@@ -100,167 +101,4 @@ public String toString() {
...
-    private final List<Long> offsets;
-    private final List<Long> splitSizes;
-    private final FileScanTask parentScanTask;
-    private int sizeIdx = 0;
-
-    OffsetsAwareTargetSplitSizeScanTaskIterator(List<Long> offsetList, FileScanTask parentScanTask) {
-      this.offsets = ImmutableList.copyOf(offsetList);
-      this.parentScanTask = parentScanTask;
-      this.splitSizes = Lists.newArrayListWithCapacity(offsets.size());
-      if (offsets.size() > 0) {
```
kbendick, line 118: Is this a situation that can be encountered? And if it can, do we need to ensure that `parentScanTask` is in fact empty?

.palantir/revapi.yml
szehon-ho: Yea I didn't know the new policy, thanks, good to know.   (was thinking to try adding a default too).

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
szehon-ho: I think sometimes it happens?  (This part is the original code, I just moved it)

szehon-ho:
FYI @aokolnychyi @RussellSpitzer @rdblue, if you also have time to take a look

aokolnychyi:
Let me take a look.

api/src/main/java/org/apache/iceberg/FileScanTask.java
```diff
@@ -26,15 +26,21 @@
  * A scan task over a range of a single file.
  */
 public interface FileScanTask extends ScanTask {
+
+  /**
+   * @return the file to scan
+   */
+  ContentFile<?> contentFile();
```
aokolnychyi, line 33: This PR raises a somehow similar question about the future of `FileScanTask` as in #4870. Maybe, we can discuss there a bit.

Instead of adding a new method like `contentFile`, what about wrapping a delete file as `DataFile` and using the task without any changes?

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
```diff
@@ -0,0 +1,44 @@
...
+    return MetadataTableType.POSITION_DELETES;
+  }
+
+  @Override
+  public TableScan newScan() {
+    return new PositionDeletesTableScan(operations(), table(), schema());
+  }
+
+  @Override
+  public Schema schema() {
```
aokolnychyi, line 41: It would be nice to include a column that would indicate either the snapshot ID when it was added or the sequence number of the row. We can then have filter pushdown on those columns.

aokolnychyi:
I feel it is reasonable to have a metadata table like this. I am not sure about the change in `FileScanTask`, though. I also have a few questions related to `FileScanTask` in #4870.

api/src/main/java/org/apache/iceberg/FileScanTask.java
szehon-ho: Yea I think that works although it felt not so clean,  DeleteDataFile?   It will definitely be less changes in code, I'm ok though if everyone agrees.  Will try to follow there as well

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
szehon-ho: Good idea

kbendick:
+1 for using a metadata table for this. It can then be used in numerous places and for various reasons.

aokolnychyi:
While I think a metadata table like this would be helpful, I am not sure it is something we can leverage during compaction. We will need to have access to a list of `DeleteFile` we compacted at the commit time so that we can replace them with the newly added files. That's why the data compaction action uses the staging approach.

stevenzwu:
Regarding the metadata table, why is it only limited to position deletes? do we anticipate a separate equality deletes metadata table too? This metadata table is at row level, like a source. For row-level deletes metadata, I am wondering if `delete_files` summary is good enough, which is similar to the `files` metadata table.

I assume this position_deletes metadata table is to not related to the CDC read like Ryan suggested in the CDC. Otherwise, the metadata table should cover inserts, position deletes, and equality deletes.

szehon-ho:
@stevenzwu yea I think equality deletes will be different table because the schema is not the same.

@aokolnychyi yea the idea wrt RewritePositionDeleteFiles was to use the building blocks in conjunction with file-scan-task-set-id, for the compaction case.  This pr should have most of the pieces though, and hopefully we just need to add some wiring.

stevenzwu:
> I thought this would be useful as we are lacking any way for user to see contents of delete files easily. 

Thanks for confirming.

api/src/main/java/org/apache/iceberg/FileScanTask.java
szehon-ho: @aokolnychyi reverted the refactor to move this forward.  As over there it seems there is no need to represent EqualityDelete as DataFile, it seems this becomes the only weird case (outside of course the metadata tables).  If we want to do it later, its still open.

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
szehon-ho: @aokolnychyi i added a test to verify all the metadata columns like _file_path work as expected.  Users can join position_delete._file_path column with delete_files table to get snapshot-id, sequence-number as they wish.  Though it is not for filtering currently.

I can take a look at adding a partition or other filter column.  It looks not very clean (need to add them as MetadataColumns as that's probably the only supported way now to get this to work with existing RowDataReader).

szehon-ho:
Addressed the review comments.  As mentioned in second comment , adding some filter columns like partition may require a bit of discussion on the implementation, may add this here or in a second pr.  But in any case, this should be ready as is.

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
szehon-ho: Update: metadata column works but unfortunately does not get pushed down for filtering by Iceberg in general:  SparkScanBuilder.pushFilters().  Seems like it will be useful but orthogonal change.  

Investigating making partition column as non-metadata column, though its a shame as there already exists the _partition metadata column.

core/src/main/java/org/apache/iceberg/DataFiles.java
```diff
@@ -108,6 +108,29 @@ public static DataFile fromManifest(ManifestFile manifest) {
         .build();
   }
 
+  public static DataFile fromPositionDelete(DeleteFile deleteFile, PartitionSpec spec) {
```
RussellSpitzer, line 111: Do we need an assert here to make sure someone doesn't pass through an Equality delete?

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,75 @@
...
+    Expression filter = ignoreResiduals ? Expressions.alwaysTrue() : context().rowFilter();
+    ResidualEvaluator residuals = ResidualEvaluator.unpartitioned(filter);
+    DeleteFileIndex deleteIndex = DeleteFileIndex.builderFor(tableOps().io(), snapshot().deleteManifests())
+        .specsById(table().specs())
+        .build();
+
+    CloseableIterable<DeleteFile> deleteFiles = CloseableIterable.withNoopClose(deleteIndex.referencedDeleteFiles());
+    CloseableIterable<DeleteFile> positionDeleteFiles = CloseableIterable.filter(deleteFiles,
+        df -> df.content().equals(FileContent.POSITION_DELETES));
+    return CloseableIterable.transform(positionDeleteFiles, f -> {
```
RussellSpitzer, line 57: nit: avoid one letter function args, could just be a "file"  or deleteFile unless we are out of o line length.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -167,6 +181,70 @@ public StructLikeSet rowSet(String name, Types.StructType projection, String...
...
+  private StructLikeSet actualPositionDeleteRowSet(String tableName,
+                                                   Table table,
+                                                   String filter,
+                                                   Types.StructType selectSchema) {
+    Dataset<Row> df = spark.read()
+        .format("iceberg")
+        .load("default." + tableName + ".position_deletes");
+    if (selectSchema != null) {
+      df = df.select(
+          selectSchema.fields().stream().map(
```
RussellSpitzer, line 207: looks like odd formatting here? maybe put the whole map on the following line?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -167,6 +181,70 @@ public StructLikeSet rowSet(String name, Types.StructType projection, String...
...
 
+  private StructLikeSet actualPositionDeleteRowSet(String tableName, Table table) {
+    return actualPositionDeleteRowSet(tableName, table, null, null);
+  }
+
+  private StructLikeSet actualPositionDeleteRowSet(String tableName, Table table, String filter) {
+    return actualPositionDeleteRowSet(tableName, table, filter, null);
+  }
+
+  private StructLikeSet actualPositionDeleteRowSet(String tableName,
```
RussellSpitzer, line 192: this and the following have formatting issues I think? We generally don't do 1 arg per line.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,75 @@
...
+          null, /* Deletes */
+          schemaString,
+          specString,
+          residuals);
+    });
+  }
+
+  @Override
+  public long targetSplitSize() {
+    long tableValue = tableOps().current().propertyAsLong(
```
RussellSpitzer, line 70: I think we may need something here to deal with splittable vs un-splittable files? Currently all of our delete files are parquet (I believe) which means we have to take into effect our offsets when planning and combining splits.

RussellSpitzer (commented):
Let me know when this PR is rebased on top of the ScanRefactor with the custom scan task we discussed

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
szehon-ho: Update 2:  Successfully made a non-metadata column partition column, and partition filter based on it.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: I think this should be fine, currently BaseFileScanTask checks if the file type is splittalbe before using this value.  I also added test in latest pr against all three file formats, against one test method that purposely hits the split limit.

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -60,8 +61,12 @@ protected BaseMetadataTable(TableOperations ops, Table table, String name) {
    * @return a spec used to rewrite the metadata table filters to partition filters using an inclusive projection
    */
   static PartitionSpec transformSpec(Schema metadataTableSchema, PartitionSpec spec) {
+    return transformSpec(metadataTableSchema, spec.partitionType());
+  }
+
+  static PartitionSpec transformSpec(Schema metadataTableSchema, Types.StructType partitionType) {
```
szehon-ho, line 67: There is a tricky bug here in partition filter push down that took me awhile to find.

In most metadata tables, the transformed partition spec used only in ManifestGroup to filter out matching manifests.  In that case, it seems the generated fieldId is not important.  However, the position delete metadata table needs to filter out matching manifest-entries (as it actually reads delete files).   A wrong fieldId causes the DeleteFile to be instantiated with wrong partition information (as the field id is used to project onto the partition struct to lookup the value), leading to various bugs in filtering out the wrong files.

szehon-ho:
Added partition column and got filter push down working, with all the partition spec evolution scenarios.  Added a lot of tests in dedicated file (TestPositionDeletesTable).

Will look to see how to incorporate #5077 is in.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,75 @@
...
+        .specsById(table().specs())
+        .build();
+
+    CloseableIterable<DeleteFile> deleteFiles = CloseableIterable.withNoopClose(deleteIndex.referencedDeleteFiles());
+    CloseableIterable<DeleteFile> positionDeleteFiles = CloseableIterable.filter(deleteFiles,
+        df -> df.content().equals(FileContent.POSITION_DELETES));
+    return CloseableIterable.transform(positionDeleteFiles, f -> {
+      PartitionSpec spec = table().specs().get(f.specId());
+      String specString = PartitionSpecParser.toJson(spec);
+      return new BaseFileScanTask(DataFiles.fromPositionDelete(f, spec),
```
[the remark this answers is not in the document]
szehon-ho: So I spent a bit of time on this, but it still is not compatible yet.  TableScan still returns FileScanTask so unless we change the interface I am not sure how this will get passed to Spark (Table.newScan is used everywhere to get TableScans).  

I think this is possible in https://github.com/apache/iceberg/pull/4812/files/051014682a2e876d2d5a7cb0b545bb5c1f8fa171 by making a new method on FileScanTask (alternatively, maybe we can add a new method on Table to return a generic Scan), but I guess it's not as clean.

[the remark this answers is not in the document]
szehon-ho: FYI @aokolnychyi

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
aokolnychyi, line 67: Is there a reason to add an overloaded method that accepts `StructType`?
Seems like it is not being used right now.

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -60,8 +61,12 @@ protected BaseMetadataTable(TableOperations ops, Table table, String name) {
...
     PartitionSpec.Builder identitySpecBuilder = PartitionSpec.builderFor(metadataTableSchema).checkConflicts(false);
-    spec.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), pf.name(), "identity"));
+    partitionType.fields().forEach(pf -> identitySpecBuilder.add(pf.fieldId(), pf.fieldId(), pf.name(), "identity"));
```
aokolnychyi, line 69: Was the switch from `spec.fields()` to `spec.partitionType().fields()` deliberate?

core/src/main/java/org/apache/iceberg/DataFiles.java
```diff
@@ -108,6 +108,31 @@ public static DataFile fromManifest(ManifestFile manifest) {
...
+    );
+
+    return DataFiles.builder(spec)
+        .withEncryptionKeyMetadata(deleteFile.keyMetadata())
+        .withFormat(deleteFile.format())
+        .withFileSizeInBytes(deleteFile.fileSizeInBytes())
+        .withPath(deleteFile.path().toString())
+        .withPartition(deleteFile.partition())
+        .withRecordCount(deleteFile.recordCount())
+        .withSortOrder(SortOrder.unsorted())
```
aokolnychyi, line 131: Shall we propagate the sort order? We can add `withSortOrderId` to the builder.

core/src/main/java/org/apache/iceberg/DataFiles.java
```diff
@@ -108,6 +108,31 @@ public static DataFile fromManifest(ManifestFile manifest) {
         .build();
   }
 
+  public static DataFile fromPositionDelete(DeleteFile deleteFile, PartitionSpec spec) {
```
aokolnychyi, line 111: Does this have to be specific to position deletes? Will constructing `DataFile` from an equality delete file be any different?

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
```diff
@@ -20,13 +20,38 @@
...
 import org.apache.iceberg.Schema;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types;
 
 public class DeleteSchemaUtil {
   private DeleteSchemaUtil() {
   }
 
+  public static Schema metadataTableSchema(Table table) {
```
aokolnychyi, line 33: Are we sure this logic belongs here? Up to you but `PositionDeletesTable` seemed more appropriate.

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
```diff
@@ -20,13 +20,38 @@
...
+    return metadataTableSchema(table, Partitioning.partitionType(table));
+  }
+
+  public static Schema metadataTableSchema(Table table, Types.StructType partitionType) {
+    Schema result = new Schema(
+        MetadataColumns.DELETE_FILE_PATH,
```
aokolnychyi, line 39: Are these real metadata columns or just columns that happen to have the same name? I don't think we should reuse column IDs as metadata columns are using reserved ones.

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
```diff
@@ -20,13 +20,38 @@
...
+        MetadataColumns.DELETE_FILE_POS,
+        Types.NestedField.optional(
+            MetadataColumns.DELETE_FILE_ROW_FIELD_ID, "row", table.schema().asStruct(),
+            MetadataColumns.DELETE_FILE_ROW_DOC));
+
+    if (partitionType.fields().size() > 0) {
```
aokolnychyi, line 45: I feel like using `TypeUtil.selectNot` will make it easier to read:

```
Types.StructType partitionType = Partitioning.partitionType(table);
Schema schema = new Schema(...);

if (partitionType.fields().size() < 1) {
  // avoid returning an empty struct, which is not always supported. instead, drop the partition field
  return TypeUtil.selectNot(schema, Sets.newHashSet(PARTITION_ID));
} else {
  return schema;
}
```

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
```diff
@@ -0,0 +1,59 @@
...
+    return DeleteSchemaUtil.metadataTableSchema(table());
+  }
+
+  @Override
+  public PartitionSpec spec() {
+    return table().spec();
```
aokolnychyi, line 52: I am not sure this is correct. All other metadata tables are unpartitioned and there is logic in the scan builder that assumes all metadata tables are unpartitioned. Do we need to make this one partitioned for a reason?

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  public Map<Integer, PartitionSpec> specs() {
```
aokolnychyi, line 56: Same here.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.stream.Collectors;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.expressions.ResidualEvaluator;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.util.Pair;
+import org.apache.iceberg.util.ParallelIterable;
+import org.apache.iceberg.util.PropertyUtil;
+
+public class PositionDeletesTableScan extends BaseMetadataTableScan {
```
aokolnychyi, line 32: Why not make it a nested class like in other metadata tables? Because it is too big?
Just out of curiosity.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public PositionDeletesTableScan(TableOperations ops, Table table, Schema schema) {
```
aokolnychyi, line 34: nit: can be package-private?

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,97 @@
...
+  protected TableScan newRefinedScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
+    return new PositionDeletesTableScan(ops, table, schema, context);
+  }
+
+  @Override
+  protected CloseableIterable<FileScanTask> doPlanFiles() {
+    Expression rowFilter = context().rowFilter();
+    String schemaString = SchemaParser.toJson(tableSchema());
+    boolean ignoreResiduals = context().ignoreResiduals();
+    Expression filter = ignoreResiduals ? Expressions.alwaysTrue() : rowFilter;
```
aokolnychyi, line 52: nit: use shouldIgnoreResiduals() directly?

```
Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : rowFilter;  
```

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,97 @@
...
+        .stream()
+        .map(e -> Pair.of(e.getKey(), BaseMetadataTable.transformSpec(tableSchema(), e.getValue())))
+        .collect(Collectors.toMap(Pair::first, Pair::second));
+
+    CloseableIterable<ManifestFile> deleteManifests = CloseableIterable.withNoopClose(
+        snapshot().deleteManifests(tableOps().io()));
+    CloseableIterable<CloseableIterable<FileScanTask>> results = CloseableIterable.transform(deleteManifests, m -> {
+
+      // Filter partitions
+      CloseableIterable<ManifestEntry<DeleteFile>> deleteFileEntries = ManifestFiles
```
aokolnychyi, line 65: Respect case sensitivity configured for this scan in `isCaseSensitive()`?

aokolnychyi, line 65: What about projection?

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
```diff
@@ -0,0 +1,97 @@
...
+          .filterRows(rowFilter)
+          .liveEntries();
+
+      // Filter delete file type
+      CloseableIterable<ManifestEntry<DeleteFile>> positionDeleteEntries = CloseableIterable.filter(deleteFileEntries,
+          entry -> entry.file().content().equals(FileContent.POSITION_DELETES));
+
+      return CloseableIterable.transform(positionDeleteEntries, entry -> {
+        PartitionSpec spec = transformedSpecs.get(entry.file().specId());
+        ResidualEvaluator residuals = ResidualEvaluator.of(spec, filter, context().caseSensitive());
```
aokolnychyi, line 76: nit: `context().caseSensitive()` -> `isCaseSensitive()`

aokolnychyi, line 76: Seems like we are doing repetitive work here for each task/spec.
Some sort of `LoadingCache<Integer, ResidualEvaluator> residualCache` would make sense here like in a few other places.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
aokolnychyi: Hm, I'd be surprised if a metadata table does not respect the metadata split size.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: Was using it in a test, but changed the test.

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
szehon-ho: Actually this crashes if it is not there.  See this code BaseDataReader::constantsMap, where partition field is populated, it will know to get it from the constants (FileScanTask) else it gets it from

core/src/main/java/org/apache/iceberg/DataFiles.java
szehon-ho: In my test, sort order id doesnt seem to be populated on delete files (returns null), but added a conditional set in case it is populated.

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
szehon-ho: One of the tricky things is to pick a column field id, because here it may conflict with the user field id.  (the row of position deletes is a exact struct of the user's schema).  So I will still put these constants on the class MetadataColumns, but as reserved columns instead of metadata columns.

szehon-ho: Actually looking into this, if I don't re-use the metadata column I have to change a lot of codes, like:

- BaseDataReader.constantsMap will need new method to populate partition based on new partition id, which I'll have to pass in based on what field is selected.  (Currently since both id's are the same it works)
- All the error checks for whether selected columns are in metadata columns need to take this into account.  Currently this uses MetadataColumns.metadataFieldIds(), which will need to be now something like MetadataColumns.metadataAndConstantFieldIds()  to take into account new column.

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
szehon-ho: This was emulated from other metadata table like BaseFilesTable, but good point.

core/src/main/java/org/apache/iceberg/DataFiles.java
szehon-ho: Good point

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: So I played with selecting some manifest fields, but actually we need all of them (even column stats).  Because we need to evaluate the filters on the actual delete file, we need to propagate this down.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: The default metadata split size is 32MB vs data file split size is 128 MB.  I thought initially the metadata split size is more suitable to read metadata than content files (which could be big?) but I can change it.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: Good point, done.

core/src/main/java/org/apache/iceberg/DataFiles.java
szehon-ho: In the end, decided to make the method work for both kind of delete files, as the DeleteFile object should not be different:  https://github.com/apache/iceberg/pull/4812#discussion_r918441075

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
szehon-ho: Changed it back.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: Yea initially when writing I thought it would be big, but in the end it is ok, so made it a nested class.

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
szehon-ho: Did the above, feel free to see if it is better or if its messier.

szehon-ho:
Rebase and support spark 3.3, resolved most comments.

core/src/main/java/org/apache/iceberg/PositionDeletesTableScan.java
szehon-ho: Changed.

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
szehon-ho: Also adding some background:  I had tried earlier to use the existing metadata column for this table (which is populated), but then pushdown does not work for metadata columns.    It seems the pruneSchema() function is called after filter() function on SparkScanBuilder, so that the binding in filter will fail.

core/src/main/java/org/apache/iceberg/MetadataColumns.java
```diff
@@ -53,6 +53,8 @@ private MetadataColumns() {
   public static final String DELETE_FILE_ROW_FIELD_NAME = "row";
   public static final int DELETE_FILE_ROW_FIELD_ID = Integer.MAX_VALUE - 103;
   public static final String DELETE_FILE_ROW_DOC = "Deleted row values";
+  public static final int POSITION_DELETE_TABLE_PARTITION_FIELD_ID = Integer.MAX_VALUE - 104;
```
aokolnychyi, line 56: It seems we have a static column in the metadata table that we plan to populate via the mechanism for metadata columns. It looks a little bit suspicious to me.

I feel we should pick one of these options:
- Have only `path`, `pos`, `row` columns in the table and use `_partition` and `_spec_id` metadata columns. That will mean we have to support filter pushdown on metadata columns. It is easy to handle this on the Spark side but we will also have to adapt ALL of our binding code to allow binding predicates with metadata columns. The last part will be a big change.
- Make `partition` and `spec_id` columns static and use `DataTask`.

I kind of like using `FileScanTask` for this effort to support vectorized reads so the first option seems preferable.

Thoughts, @szehon-ho @RussellSpitzer @rdblue?

rdblue: My concern with the marker `FileScanTask` is that if an engine is implementing metadata tables like normal reads, then we've introduced a correctness problem because it doesn't know to read the new task differently. I think the cleanest way is probably to use `DataTask`.

I think the concern about using `DataTask` is valid, since it exposes a row-based interface and isn't intended for large uses that benefit from vectorization. It was originally intended for small tables, like `snapshots`.

However, I've been thinking for a while that a significant improvement is to adapt Arrow record batches into rows, so we can take advantage of vectorized reads in all cases, not just when the engine supports a vectorized format. That is probably way faster, so we could explore doing that here and using a joined row.

szehon-ho: Thanks for taking a look.  It's getting a bit more messy than anticipated.

Was spending a lot of time yesterday looking at the marker FileScanTask approach and it does get messy.  Especially  trying to add another constant column (spec_id) to this table, not only is it another use of metadata column  to populate it, we have to figure out how to make a residual expression without the constant column as otherwise the default file read code tries to filter too aggressively if there is a spec_id filter (as the stats do not exist on the file and the pruning code skips it) .  In short , adding any constant columns to this table, gets a bit messy.

Makes sense to explore using some kind of StaticDataTask.  I think the problem for that, currently it would not do split() and implement any file residual filtering on position-delete row values if there are any filters there, and of course vectorization, and it'll be re-implementing these if we want it.  But maybe I am over-optimizing this and in the first cut we can live without those and add them later.  Can look at this approach, and see if I hit any major blockers.

szehon-ho: Hi guys, I hit the first issue , as StaticDataTask is in core module, there is no way currently for it to access Parquet, ORC, and the file readers.  So maybe engines, like Spark, can pass it in?  This would seem to require an API change in any case on DataTask.

So something like, adding a new method DataTask::rows(Function<DeleteFile, Schema, Expression> filter readerFunction)?  

This can also take care of encryption, which seems to also be a callback (for example DeleteFilter::inputFile(), which is implemented by Spark/Flink).

szehon-ho: I guess the only thing I can think to do is to provide a shared implementation of positionDeleteReader in data module, but each engine still needs to check the type of table (or FileScanTask) and then manually plug this in to the DataTask.

szehon-ho: cc @rdblue @aokolnychyi @rdblue appreciate any thoughts, I think its a good idea to try to make this table usable across engines using existing DataTask,  without extra if-else handling, but not sure we can do it as of today due to the dependency.  Will try to explore the above , but it  still will not be as clean

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
RussellSpitzer: What happens here if the table spec has evolved? Shouldn't this be the unified partition spec of the table?

core/src/main/java/org/apache/iceberg/MetadataColumns.java
szehon-ho: Hey guys I really tried hard on this, but couldn't find a way.  We looked with @RussellSpitzer  and looks like Trino actually wont break if we had a new table (they support each one manually, and actually i think a good portion don't even use our MetadataTable class altogether).   Discussed a bit with @aokolnychyi as well and coudn't find a good way.  Made a latest pr with a least bad way, piggy backing off of some of the latest RowReader refactoring

core/src/main/java/org/apache/iceberg/PositionDeletesTable.java
szehon-ho: Update, in latest implementation, removed partition spec from table, as original comment.

core/src/main/java/org/apache/iceberg/MetadataColumns.java
```diff
@@ -89,8 +92,15 @@ private MetadataColumns() {}
           ROW_POSITION.fieldId(),
           IS_DELETED.fieldId(),
           SPEC_ID.fieldId(),
-          PARTITION_COLUMN_ID);
+          PARTITION_COLUMN_ID,
```
szehon-ho, line 95: This is to avoid an issue when trying to read these constant columns via Avro reader, which supports but only if they are defined here..

```
Caused by: java.lang.IllegalArgumentException: Missing required field: partition
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
	at org.apache.iceberg.avro.BuildAvroProjection.record(BuildAvroProjection.java:104)
	at org.apache.iceberg.avro.BuildAvroProjection.record(BuildAvroProjection.java:42)
	at org.apache.iceberg.avro.AvroCustomOrderSchemaVisitor.visit(AvroCustomOrderSchemaVisitor.java:50)
	at org.apache.iceberg.avro.AvroSchemaUtil.buildAvroProjection(AvroSchemaUtil.java:126)
	at org.apache.iceberg.avro.ProjectionDatumReader.setSchema(ProjectionDatumReader.java:68)
	at org.apache.avro.file.DataFileStream.initialize(DataFileStream.java:133)
	at org.apache.avro.file.DataFileReader.<init>(DataFileReader.java:139)
	at org.apache.avro.file.DataFileReader.<init>(DataFileReader.java:131)
	at org.apache.avro.file.DataFileReader.openReader(DataFileReader.java:75)
	at org.apache.iceberg.avro.AvroIterable.newFileReader(AvroIterable.java:100)
	at org.apache.iceberg.avro.AvroIterable.iterator(AvroIterable.java:76)
	at org.apache.iceberg.io.CloseableIterable.lambda$filter$0(CloseableIterable.java:110)
	at org.apache.iceberg.io.CloseableIterable$2.iterator(CloseableIterable.java:72)
	at org.apache.iceberg.spark.source.RowDataReader.open(RowDataReader.java:61)
	at org.apache.iceberg.spark.source.RowDataReader.open(RowDataReader.java:38)
	at org.apache.iceberg.spark.source.BaseReader.next(BaseReader.java:128)
	at org.apache.spark.sql.execution.datasources.v2.PartitionIterator.hasNext(DataSourceRDD.scala:119)
	at org.apache.spark.sql.execution.datasources.v2.MetricsIterator.hasNext(DataSourceRDD.scala:156)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.$anonfun$hasNext$1(DataSourceRDD.scala:63)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.$anonfun$hasNext$1$adapted(DataSourceRDD.scala:63)
```

szehon-ho:
Some test was cancelled, re-triggering

core/src/main/java/org/apache/iceberg/io/DeleteSchemaUtil.java
szehon-ho: Update, this should be resolved, do not re-use metadata column ids anymore.

szehon-ho:
Closing as it's now broken into smaller prs

PR #4815, opened 2022-05-19
BooleanExpressionVisitor and visit method

samredai:
This adds the `BooleanExpressionVisitor` base class and a `visit` function to the expressions base. The visit function is missing a dispatched function for `UnboundPredicate` and `BoundPredicate`, which we can add once those classes are added.

cc: @CircArgs


python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+
+        Returns:
+            BooleanExpression: The resulting expression after any modifications performed by the visitor
+        """
+
+    @abstractmethod
+    def visit_always_false(self) -> BooleanExpression:
+        """Visit method for an AlwaysFalse boolean expression
+
+        Note: This visit method has no arguments since AlwaysTrue instances have no context.
```
samredai, line 366:
```suggestion
        Note: This visit method has no arguments since AlwaysFalse instances have no context.
```

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+            BooleanExpression: The resulting expression after any modifications performed by the visitor
+        """
+
+    @abstractmethod
+    def visit_not(self, result: BooleanExpression) -> BooleanExpression:
+        """Visit method for a Not boolean expression
+
+        Args:
+            left (BooleanExpression): The left side of the expression
+            right (BooleanExpression): The right side of the expression
```
rdblue, line 378: Looks like this was copy-pasted and not updated.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
             raise ValueError(f"Cannot find field '{self.name}' in schema: {schema}")
 
         return BoundReference(field=field, accessor=schema.accessor_for_field(field.field_id))
+
+
+class BooleanExpressionVisitor(ABC):
```
rdblue, line 351: I think this should be parameterized. The visitor will return some type, like the schema visitor.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+    @abstractmethod
+    def visit_always_true(self) -> BooleanExpression:
```
rdblue, line 353: `BooleanExpression` -> `T`

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+        """Visit method for an AlwaysFalse boolean expression
+
+        Note: This visit method has no arguments since AlwaysFalse instances have no context.
+
+        Returns:
+            BooleanExpression: The resulting expression after any modifications performed by the visitor
+        """
+
+    @abstractmethod
+    def visit_not(self, result: BooleanExpression) -> BooleanExpression:
```
rdblue, line 373: The result should have type `T` because this is passed the result of visiting the child.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+        Args:
+            left (BooleanExpression): The left side of the expression
+            right (BooleanExpression): The right side of the expression
+
+        Returns:
+            BooleanExpression: The resulting expression after any modifications performed by the visitor
+        """
+
+    @abstractmethod
+    def visit_predicate(self, predicate):
```
rdblue, line 409: I think it would be better to have `visit_unbound_predicate` and `visit_bound_predicate` so the implementation doesn't have to do an `isinstance` check.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+    The function traverses the expression in post-order fashion
+
+    Args:
+        obj(BooleanExpression): An instance of a BooleanExpression
+        visitor (BooleanExpressionVisitor): An instance of an implementation of the generic BooleanExpressionVisitor base class
+
+    Raises:
+        NotImplementedError: If attempting to visit an unrecognized object type
+    """
+    raise NotImplementedError(f"Cannot visit non-type: {obj}")
```
rdblue, line 431: I don't think "non-type" is correct. This should be "unsupported expression" right?

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+@visit.register(AlwaysTrue)
+def _(obj: AlwaysTrue, visitor: BooleanExpressionVisitor) -> BooleanExpression:
+    """Visit an AlwaysTrue boolean expression with a concrete BooleanExpressionVisitor"""
+    return visitor.visit_always_true()
+
+
+@visit.register(AlwaysFalse)
+def _(obj: AlwaysFalse, visitor: BooleanExpressionVisitor) -> BooleanExpression:
+    """Visit an AlwaysFalse boolean expression with a concrete BooleanExpressionVisitor"""
+    return visitor.visit_always_true()
```
rdblue, line 443: Should be `false` rather than `true`.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+        """Visit method for an AlwaysTrue boolean expression
+
+        Note: This visit method has no arguments since AlwaysTrue instances have no context.
+
+        Returns:
+            BooleanExpression: The resulting expression after any modifications performed by the visitor
+        """
+
+    @abstractmethod
+    def visit_always_false(self) -> BooleanExpression:
```
rdblue, line 363: While the expression itself is `AlwaysFalse`, it's really just a way to talk about the constant expression `False`. For the visitor, I'd just use `visit_true` and `visit_false`. The `Always` is there to avoid conflicting with built-in symbols like `True` and `False`.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+@visit.register(AlwaysFalse)
+def _(obj: AlwaysFalse, visitor: BooleanExpressionVisitor) -> BooleanExpression:
+    """Visit an AlwaysFalse boolean expression with a concrete BooleanExpressionVisitor"""
+    return visitor.visit_always_true()
+
+
+@visit.register(Not)
+def _(obj: Not, visitor: BooleanExpressionVisitor) -> BooleanExpression:
+    """Visit a Not boolean expression with a concrete BooleanExpressionVisitor"""
+    return visitor.visit_not(result=obj)
```
rdblue, line 449: This needs to visit the object and then call `visit_not` to pass in the result.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,116 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+
+
+@visit.register(And)
+def _(obj: And, visitor: BooleanExpressionVisitor) -> BooleanExpression:
+    """Visit a And boolean expression with a concrete BooleanExpressionVisitor"""
+    return visitor.visit_and(left=obj.left, right=obj.right)
```
rdblue, line 455: Same here, this needs to visit the left and right operands.

python/src/iceberg/expressions/base.py
samredai: Fixed!

python/src/iceberg/expressions/base.py
samredai: I see, updated to `BooleanExpressionVisitor(Generic[T], ABC)` 👍

python/src/iceberg/expressions/base.py
samredai: Fixed!

python/src/iceberg/expressions/base.py
samredai: Updated this to `visit_not(self, child_result: T) -> T` 👍

python/src/iceberg/expressions/base.py
samredai: Good idea, updated! I was initially thinking an implementation would just use singledispatchmethod but there's no reason to require that. Using your suggest still leaves open the option the implement a `visit_predicate` singledispatchmethod and just have it called by both the `visit_unbound_predicate` and `visit_bound_predicate` methods.

python/src/iceberg/expressions/base.py
samredai: Ah right, fixed!

python/src/iceberg/expressions/base.py
samredai: Thanks for catching this, updated!

python/src/iceberg/expressions/base.py
samredai: Makes sense, updated!

python/src/iceberg/expressions/base.py
samredai: Thanks, fixed!

python/src/iceberg/expressions/base.py
samredai: Done!

samredai:
@rdblue I incorporated your review so far and also updated the PR to add complete coverage. This is ready for another look, thanks!

samredai:
cc: @emkornfield @Fokko @kbendick @dramaticlly @dhruv-pratap

python/tests/expressions/test_expressions_base.py
```diff
@@ -84,6 +85,65 @@ def __str__(self):
...
+
+
+@base.visit.register(TestExpressionA)
+def _(obj: TestExpressionA, visitor: TestBooleanExpressionVisitor) -> List:
+    """Visit a TestExpressionA with a TestBooleanExpressionVisitor"""
+    return visitor.visit_test_expression_a()
+
+
+@base.visit.register(TestExpressionB)
+def _(obj: TestExpressionB, visitor: TestBooleanExpressionVisitor) -> List:
```
samredai, line 142: These are only for tests and dispatch a method for `TestExpressionA` and `TestExpressionB`. Otherwise a `NotImplementedError` would be raised.

python/tests/expressions/test_expressions_base.py
```diff
@@ -84,6 +85,65 @@ def __str__(self):
...
 
+class TestBooleanExpressionVisitor(base.BooleanExpressionVisitor[List]):
+    """A test implementation of a BooleanExpressionVisit
+
+    As this visitor visits each node, it appends an element to a `visit_histor` list. This enables testing that a given expression is
+    visited in an expected order by the `visit` method.
+    """
+
+    def __init__(self):
+        self.visit_history: List = []
```
samredai, line 96: Since we don't have predicates yet (coming in PR #4816), I added this test-only visitor. It's super simple and just appends a string for each visit method that gets called. I use this in the tests added to confirm that the visit methods are called in the expected post-order traversal of a given expression.

python/src/iceberg/expressions/base.py
```diff
@@ -346,3 +346,112 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
...
+    @abstractmethod
+    def visit_unbound_predicate(self, predicate) -> T:
+        """Visit method for an unbound predicate in an expression tree
+
+        Args:
+            predicate (UnboundPredicate): An instance of an UnboundPredicate
+        """
+
+    @abstractmethod
+    def visit_bound_predicate(self, predicate) -> T:
```
samredai, line 401: `predicate` has no typehint here because `BoundPredicate` and `UnboundPredicate` have not been added yet, but I'll circle back and add if after PR #4816 gets in (or rebase this PR if that one gets in first).

python/tests/expressions/test_expressions_base.py
```diff
@@ -257,3 +317,111 @@ def test_bound_reference(table_schema_simple, foo_struct):
...
+        "OR",
+        "TestExpressionA",
+        "NOT",
+        "AND",
+        "TestExpressionB",
+        "AND",
+    ]
+
+
+def test_or_with_always_true():
```
rdblue, line 349: This doesn't test the visitor, it is testing `Or.__new__` that simplifies the expression as it is created. I think you can remove these and/or/false/true tests and just go with the one above for now.

python/tests/expressions/test_expressions_base.py
```diff
@@ -257,3 +317,111 @@ def test_bound_reference(table_schema_simple, foo_struct):
...
+        base.AlwaysFalse(),
+        base.AlwaysFalse(),
+    )
+    visitor = TestBooleanExpressionVisitor()
+    result = base.visit(expr, visitor=visitor)
+    assert isinstance(expr, base.AlwaysFalse)
+    assert result == ["FALSE"]
+
+
+def test_boolean_expression_visit_raise_not_implemented_error():
```
rdblue, line 421: I'd keep this one, but you just need the one test above.

python/tests/expressions/test_expressions_base.py
```diff
@@ -257,3 +317,111 @@ def test_bound_reference(table_schema_simple, foo_struct):
     assert bound_ref1.eval(foo_struct) == "foovalue"
     assert bound_ref2.eval(foo_struct) == 123
     assert bound_ref3.eval(foo_struct) == True
+
+
+def test_boolean_expression_visitor():
```
rdblue, line 322: This test looks good to me.

python/tests/expressions/test_expressions_base.py
samredai: Updated!

python/tests/expressions/test_expressions_base.py
samredai: Removed 👍

rdblue (approved):

rdblue:
Thanks, @samredai! Looks great.

PR #4816, opened 2022-05-19
remove operations from base expressions; add in expression

CircArgs:
@samredai @rdblue 
Opening a draft PR introducing the `In` expression to get confirmation of the api. 

Is `def __init__(self, ref: 'UnboundReference', list: List[Literal]):` the correct signature?

I'll add the tests next and if this PR succeeds I will add the remaining boolean expressions

### DIFF
--- a/python/pyiceberg/expressions/base.py
+++ b/python/pyiceberg/expressions/base.py
@@ -101,7 +101,7 @@ class Term(ABC):
     """An expression that evaluates to a value."""
 
 
-class Reference(Generic[T], ABC):
+class BaseReference(Generic[T], Term, ABC):
     """Represents a variable reference in an expression."""
 
 
@@ -109,12 +109,12 @@ class BoundTerm(Bound[T], Term):
     """Represents a bound term."""
 
 
-class UnboundTerm(Unbound[T, BoundTerm[T]]):
+class UnboundTerm(Unbound[T, BoundTerm[T]], Term):
     """Represents an unbound term."""
 
 
 @dataclass(frozen=True)
-class BoundReference(BoundTerm[T], Reference[T]):
+class BoundReference(BoundTerm[T], BaseReference[T]):
     """A reference bound to a field in a schema
 
     Args:
@@ -136,7 +136,7 @@ class BoundReference(BoundTerm[T], Reference[T]):
 
 
 @dataclass(frozen=True)
-class UnboundReference(UnboundTerm[T], Reference[T]):
+class Reference(UnboundTerm[T], BaseReference[T]):
     """A reference not yet bound to a field in a schema
 
     Args:
@@ -148,7 +148,7 @@ class UnboundReference(UnboundTerm[T], Reference[T]):
 
     name: str
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
+    def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
         """Bind the reference to an Iceberg schema
 
         Args:
@@ -182,7 +182,7 @@ class BoundPredicate(Bound[T], BooleanExpression):
 
 @dataclass(frozen=True)  # type: ignore[misc]
 class UnboundPredicate(Unbound[T, BooleanExpression], BooleanExpression):
-    term: UnboundReference[T]
+    term: Reference[T]
     literals: Tuple[Literal[T], ...]
 
 
@@ -317,19 +317,19 @@ class AlwaysFalse(BooleanExpression, ABC, Singleton):
 
 
 @dataclass(frozen=True)
+class BoundIn(BoundPredicate[T]):
+    def __invert__(self):
+        raise TypeError("In expressions do not support negation.")
+
+
+@dataclass(frozen=True)
 class In(UnboundPredicate[T]):
     def __invert__(self):
         raise TypeError("In expressions do not support negation.")
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> "BoundIn":
+    def bind(self, schema: Schema, case_sensitive: bool) -> BoundIn:
         bound_ref = self.term.bind(schema, case_sensitive)
         return BoundIn(bound_ref, tuple(lit.to(bound_ref.field.field_type) for lit in self.literals))  # type: ignore
-
-
-@dataclass(frozen=True)
-class BoundIn(BoundPredicate[T]):
-    def __invert__(self):
-        raise TypeError("In expressions do not support negation.")
 
 
 class BooleanExpressionVisitor(Generic[T], ABC):
--- a/python/pyiceberg/expressions/base.py
+++ b/python/pyiceberg/expressions/base.py
@@ -242,13 +242,13 @@ class UnboundIn(BooleanExpression, UnboundPredicate):
     def __invert__(self):
         raise TypeError("In expressions do not support negation.")
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> "In":
+    def bind(self, schema: Schema, case_sensitive: bool) -> "BoundIn":
         bound_ref = self.term.bind(schema, case_sensitive)
-        return In(bound_ref, [lit.to(bound_ref.field.field_type) for lit in self.literal])
+        return BoundIn(bound_ref, [lit.to(bound_ref.field.field_type) for lit in self.literal])
 
 
 @dataclass
-class In(BooleanExpression, BoundPredicate):
+class BoundIn(BooleanExpression, BoundPredicate):
     term: "BoundReference"
     literal: List[Literal]
 
@@ -259,6 +259,7 @@ class In(BooleanExpression, BoundPredicate):
         return self.term.eval(struct)
 
 
+@dataclass
 class BoundReference:
     """A reference bound to a field in a schema
 
@@ -267,36 +268,11 @@ class BoundReference:
         accessor (Accessor): An Accessor object to access the value at the field's position
     """
 
-    def __init__(self, field: NestedField, accessor: Accessor):
-        self._field = field
-        self._accessor = accessor
-
-    def __str__(self):
-        return f"BoundReference(field={repr(self.field)}, accessor={repr(self._accessor)})"
-
-    def __repr__(self):
-        return f"BoundReference(field={repr(self.field)}, accessor={repr(self._accessor)})"
-
-    def __eq__(self, other) -> bool:
-        return id(self) == id(other) or (self.field == other.field and self.accessor == other.accessor)
-
-    @property
-    def field(self) -> NestedField:
-        """The referenced field"""
-        return self._field
-
-    def eval(self, struct: StructProtocol) -> Any:
-        """Returns the value at the referenced field's position in an object that abides by the StructProtocol
-
-        Args:
-            struct (StructProtocol): A row object that abides by the StructProtocol and returns values given a position
-
-        Returns:
-            Any: The value at the referenced field's position in `struct`
-        """
-        return self._accessor.get(struct)
+    field: NestedField
+    accessor: Accessor
 
 
+@dataclass
 class UnboundReference:
     """A reference not yet bound to a field in a schema
 
@@ -307,23 +283,7 @@ class UnboundReference:
         An unbound reference is sometimes referred to as a "named" reference
     """
 
-    def __init__(self, name: str):
-        if not name:
-            raise ValueError(f"Name cannot be null: {name}")
-        self._name = name
-
-    def __str__(self) -> str:
-        return f"UnboundReference(name={repr(self.name)})"
-
-    def __repr__(self) -> str:
-        return f"UnboundReference(name={repr(self.name)})"
-
-    def __eq__(self, other) -> bool:
-        return id(self) == id(other) or self.name == other.name
-
-    @property
-    def name(self) -> str:
-        return self._name
+    name: str
 
     def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference:
         """Bind the reference to an Iceberg schema
--- a/python/pyiceberg/expressions/base.py
+++ b/python/pyiceberg/expressions/base.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from enum import Enum, auto
 from functools import reduce, singledispatch
 from typing import (
@@ -233,66 +234,29 @@ class AlwaysFalse(BooleanExpression, ABC, Singleton):
         return "false"
 
 
+@dataclass
 class UnboundIn(BooleanExpression, UnboundPredicate):
-    """IN operation expression"""
+    term: "UnboundReference"
+    literal: List[Literal]
 
-    def __init__(self, left: "UnboundReference", right: List[Literal]):  # pylint: disable=super-init-not-called
-        self._left = left
-        self._right = right
-
-    @property
-    def left(self) -> "UnboundReference":
-        return self._left
-
-    @property
-    def right(self) -> List[Literal]:
-        return self._right
-
-    def __eq__(self, other) -> bool:
-        return id(self) == id(other) or (isinstance(other, UnboundIn) and self.left == other.left and self.right == other.right)
-
-    def __invert__(self) -> Not:
-        return Not(self)
-
-    def __repr__(self) -> str:
-        return f"UnboundIn({repr(self.left)}, {repr(self.right)})"
-
-    def __str__(self) -> str:
-        return f"({self.left} in {self.right})"
+    def __invert__(self):
+        raise TypeError("In expressions do not support negation.")
 
     def bind(self, schema: Schema, case_sensitive: bool) -> "In":
-        return In(self.left.bind(schema, case_sensitive), self.right)
+        return In(self.term.bind(schema, case_sensitive), self.literal)
 
 
 class In(BooleanExpression, BoundPredicate):
     """IN operation expression"""
 
-    def __init__(self, left: "BoundReference", right: List[Literal]):  # pylint: disable=super-init-not-called
-        self._left = left
-        self._right = right
+    term: "BoundReference"
+    literal: List[Literal]
 
-    @property
-    def left(self) -> "BoundReference":
-        return self._left
-
-    @property
-    def right(self) -> List[Literal]:
-        return self._right
-
-    def __eq__(self, other) -> bool:
-        return id(self) == id(other) or (isinstance(other, UnboundIn) and self.left == other.left and self.right == other.right)
-
-    def __invert__(self) -> Not:
-        return Not(self)
-
-    def __repr__(self) -> str:
-        return f"In({repr(self.left)}, {repr(self.right)})"
-
-    def __str__(self) -> str:
-        return f"({self.left} in {self.right})"
+    def __invert__(self):
+        raise TypeError("In expressions do not support negation.")
 
     def eval(self, struct: StructProtocol):
-        return self.left.eval(struct)
+        return self.term.eval(struct)
 
 
 class BoundReference:
--- a/python/pyiceberg/expressions/base.py
+++ b/python/pyiceberg/expressions/base.py
@@ -83,17 +83,17 @@ class BooleanExpression(ABC):
 
 
 class BoundPredicate(ABC):
-    def __init__(self, left, right):
-        """A concrete predicate must have a `left` and `right`"""
-        self._left = left
-        self._right = right
+    def __init__(self, term, literal):
+        """A concrete predicate must have a `term` and `literal`"""
+        self._term = term
+        self._literal = literal
 
 
 class UnboundPredicate:
-    def __init__(self, left, right):
-        """A concrete predicate must have a `left` and `right`"""
-        self._left = left
-        self._right = right
+    def __init__(self, term, literal):
+        """A concrete predicate must have a `term` and `literal`"""
+        self._term = term
+        self._literal = literal
 
     @abstractmethod
     def bind(self, schema: Schema, case_sensitive: bool) -> BoundPredicate:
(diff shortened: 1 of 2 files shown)


samredai:
> Is def __init__(self, ref: 'UnboundReference', list: List[Literal]): the correct signature?

That looks right to me, `java.util.List` is used in the java implementation of [in](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/expressions/Expressions.java#L206-L222).

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +204,34 @@ def __str__(self) -> str:
         return "false"
 
 
+class In(BooleanExpression):
```
samredai, line 207: Now that we're moving away from having the single `Predicate` class, I still think we can do some things to allow grouping downstream (for like methods that must take a "predicate" or more specifically an "unbound predicate"). What do you think of adding this?

```py
class Predicate:

    @abstractmethod
    def __init__(self, left, right):
        """A concrete predicate must have a `left` and `right`"""

class Bound:
    pass

class Unbound:
    pass

class UnboundIn(BooleanExpression, Predicate, Unbound):
    def __init__(self, left: 'UnboundReference', right: List[Literal]):
        ...
```

CircArgs: I've made some changes based on your suggestions and creating `UnboundIn`

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +228,34 @@ def __str__(self) -> str:
         return "false"
 
 
+class UnboundIn(BooleanExpression, UnboundPredicate):
```
samredai, line 231: To implement `UnboundPredicate`, this needs a `bind` method that returns an `In` instance.

python/src/iceberg/expressions/base.py
```diff
@@ -137,6 +74,30 @@ def __invert__(self) -> "BooleanExpression":
...
+class BoundPredicate(ABC):
+    def __init__(self, left, right):
+        """A concrete predicate must have a `left` and `right`"""
+        self._left = left
+        self._right = right
+
+    @abstractmethod
+    def eval(self, obj) -> bool:
+        """Evaluate the bound predicate"""
+        ...
```
samredai, line 86: The new linter stuff recently added is complaining about this line. The `...` isn't needed if you have a docstring. Same thing a few lines down in the `bind` method.

samredai:
It looks like some type of auto-formatting ran across the project (maybe from your IDE?). I would try to `tox -e format` it and if that doesn't work then just restore all the files that shouldn't change as part of this PR, which is everything besides `base.py` I think.

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
...
+
+class In(BooleanExpression, BoundPredicate):
+    """IN operation expression"""
+
+    def __init__(self, left: "BoundReference", right: List[Literal]):
+        self._left = left
+        self._right = right
+
+    @property
+    def ref(self) -> "BoundReference":
```
samredai, line 268: I think the `ref` property and `list` property should be named `left` and `right` respectively so that the visitor methods can stay consistent in terms of naming. As an example, here's what I did in PR #4815 for the method dispatched for `And`:
```py
@visit.register(And)
def _(obj: And, visitor: BooleanExpressionVisitor[T]) -> T:
    """Visit an And boolean expression with a concrete BooleanExpressionVisitor"""
    left_result: T = visit(obj.left, visitor=visitor)
    right_result: T = visit(obj.right, visitor=visitor)
    return visitor.visit_and(left_result=left_result, right_result=right_result)
```
Once this PR is in, we would add a `visit` dispatch method for `UnboundPredicate`/`BoundPredicate` that looks like:
```py
@visit.register(UnboundPredicate)
def _(obj: UnboundPredicate, visitor: BooleanExpressionVisitor[T]) -> T:
    """Visit an unbound predicate with a concrete BooleanExpressionVisitor"""
    left_result: T = visit(obj.left, visitor=visitor)
    right_result: T = visit(obj.right, visitor=visitor)
    return visitor.visit_unbound_predicate(left_result=left_result, right_result=right_result)

@visit.register(BoundPredicate)
def _(obj: BoundPredicate, visitor: BooleanExpressionVisitor[T]) -> T:
    """Visit a bound predicate with a concrete BooleanExpressionVisitor"""
    left_result: T = visit(obj.left, visitor=visitor)
    right_result: T = visit(obj.right, visitor=visitor)
    return visitor.visit_bound_predicate(left_result=left_result, right_result=right_result)
```

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
         return "false"
 
 
+class UnboundIn(BooleanExpression, UnboundPredicate):
+    """IN operation expression"""
+
+    def __init__(self, left: "UnboundReference", right: List[Literal]):
```
samredai, line 232:
```suggestion
    def __init__(self, left: "UnboundReference", right: List[Literal]):  # pylint: disable=super-init-not-called
```

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
...
+        return f"({self.ref} in {self.list})"
+
+    def bind(self, schema: Schema, case_sensitive: bool) -> "In":
+        return In(self.ref.bind(schema, case_sensitive), self.list)
+
+
+class In(BooleanExpression, BoundPredicate):
+    """IN operation expression"""
+
+    def __init__(self, left: "BoundReference", right: List[Literal]):
```
samredai, line 263:
```suggestion
    def __init__(self, left: "BoundReference", right: List[Literal]):  # pylint: disable=super-init-not-called
```

python/src/iceberg/expressions/base.py
```diff
@@ -137,6 +74,28 @@ def __invert__(self) -> "BooleanExpression":
         ...
 
 
+class BoundPredicate(ABC):
+    def __init__(self, left, right):
+        """A concrete predicate must have a `left` and `right`"""
+        self._left = left
+        self._right = right
```
kbendick, line 81: The usage of `left` and `right` here is somewhat confusing for me. Is this something that I missed that differs (or is renamed) from the Java library?

Or possibly `left` and `right` correspond tot the java `term` and `literal` arguments of `BoundLiteralPredicate`?

samredai: > Or possibly left and right correspond tot the java term and literal arguments of BoundLiteralPredicate?

Yes exactly. The plan is to not have a generic `BoundLiteralPredicate` that uses an `Operation` enum + a switch statement. Instead we'll have a different class for every operation, so `In(BooleanExpression, BoundPredicate)`, `LessThan(BooleanExpression, BoundPredicate)`, `GreaterThan(BooleanExpression, BoundPredicate)`, etc.

Since we're working with binary expression trees, I think it makes sense to use `left` and `right` here to keep the visitor logic intuitive, even though this deviates from the java `term` and `literal` argument names. IMO the fact that `left` is always a `term` and `right` is always a `literal` can simply be described through typehints and the docstring.

kbendick: Oh ok I see that left is the reference and right is the literal(s).

Could we possibhly name this something more specific than `left` / `right`?

samredai: Does `ref` and `literal` sound better? That's what it originally was in this PR. The visit function would then look like:
```py
@visit.register(UnboundPredicate)
def _(obj: UnboundPredicate, visitor: BooleanExpressionVisitor[T]) -> T:
    """Visit an unbound predicate with a concrete BooleanExpressionVisitor"""
    ref_result: T = visit(obj.ref, visitor=visitor)
    literal_result: T = visit(obj.literal, visitor=visitor)
    return visitor.visit_unbound_predicate(ref_result=ref_result, literal_result=literal_result)
```

rdblue: Let's not worry about the visitor just yet. I think we would probably pass the actual predicate rather than deconstructing it. Otherwise we'd need to reintroduce operation to pass it in.

I agree that this should be more specific about its children. I think it should use `term` (always left) and `literal`.

python/src/iceberg/expressions/base.py
rdblue: `UnboundPredicate` can be abstract and define `bind`.

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
         return "false"
 
 
+class UnboundIn(BooleanExpression, UnboundPredicate):
+    """IN operation expression"""
+
+    def __init__(self, left: "UnboundReference", right: List[Literal]):  # pylint: disable=super-init-not-called
```
rdblue, line 232: Why not call `__init__`?

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
...
+    @property
+    def left(self) -> "UnboundReference":
+        return self._left
+
+    @property
+    def right(self) -> List[Literal]:
+        return self._right
+
+    def __eq__(self, other) -> bool:
+        return id(self) == id(other) or (isinstance(other, UnboundIn) and self.left == other.left and self.right == other.right)
```
rdblue, line 245: I think using separate cases will help this be more readable.

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
...
+
+    def __invert__(self) -> Not:
+        return Not(self)
```
rdblue, line 248: I think this should be abstract. We will need to rewrite expressions to not include `Not` and we don't want to have this reintroducing it.

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
...
+
+    def __repr__(self) -> str:
```
rdblue, line 250: Does it make sense to use `@dataclass` so we get these for free?

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +226,68 @@ def __str__(self) -> str:
...
+        return f"UnboundIn({repr(self.left)}, {repr(self.right)})"
+
+    def __str__(self) -> str:
+        return f"({self.left} in {self.right})"
+
+    def bind(self, schema: Schema, case_sensitive: bool) -> "In":
+        return In(self.left.bind(schema, case_sensitive), self.right)
```
rdblue, line 257: Bind needs to happen for both the term and the literal. The value held by the literal must be rewritten to the type of the term. That's what `literal.to(type)` is responsible for.

python/src/iceberg/expressions/base.py
rdblue, line 248: For now, this should throw an exception. The logical inverse of IN is NOT_IN, rather than NOT(IN). When rewriting expressions, it's critical that we don't introduce new `Not` nodes because they can't be evaluated in many cases.

python/src/iceberg/expressions/base.py
CircArgs: is the `__str__` representation key and should look like it does here - that would be one reason but I assume no

I also don't think dataclasses provide automatic encapsulation in any way like our obfuscation and use of properties does currently.

I'll make the switch and revert if needed.

python/src/iceberg/expressions/base.py
```diff
@@ -137,6 +75,28 @@ def __invert__(self) -> "BooleanExpression":
...
 
 
+class BoundPredicate(ABC):
+    def __init__(self, left, right):
+        """A concrete predicate must have a `left` and `right`"""
+        self._left = left
+        self._right = right
+
+    @abstractmethod
+    def eval(self, struct: StructProtocol) -> bool:
```
rdblue, line 85: It may be misleading to have an eval method when there are so many different ways to evaluate a predicate.

python/src/iceberg/expressions/base.py
```diff
@@ -267,6 +227,31 @@ def __str__(self) -> str:
         return "false"
 
 
+@dataclass
+class UnboundIn(BooleanExpression, UnboundPredicate):
```
rdblue, line 231: You probably want to use `In` for `UnboundIn` and `BoundIn` for the bound version. Users interact the most with the unbound versions and don't need to know much about expression binding.

python/src/iceberg/expressions/base.py
```diff
@@ -285,6 +270,9 @@ def __str__(self):
     def __repr__(self):
         return f"BoundReference(field={repr(self.field)}, accessor={repr(self._accessor)})"
 
+    def __eq__(self, other) -> bool:
```
rdblue, line 273: If you convert this to a `@dataclass`, I think you'd get these for free.

CircArgs: changed to dataclasses

python/src/iceberg/expressions/base.py
CircArgs: updated this

python/src/iceberg/expressions/base.py
CircArgs: I got rid of all the eval methods

python/src/iceberg/expressions/base.py
CircArgs: now throws exception

python/src/iceberg/expressions/base.py
CircArgs: call `.to` in a list comprehension now

python/src/iceberg/expressions/base.py
CircArgs: uses dataclasses

python/src/iceberg/expressions/base.py
CircArgs: addressed by changing to dataclasses

rdblue:
@CircArgs, this is looking better, but I still think that we need to use `term` and `literal` in predicates, rather than `left` and `right`.

CircArgs:
Hey @rdblue sorry for the delay I was on vacation. I see I didn't change the abstract definition attribute names so I've changed that now. Do you also mean the `And` and `Or` as well? I've left those two alone for now because it's not clear to me why they would have the names `term` and `literal`

python/src/iceberg/expressions/base.py
```diff
@@ -268,41 +223,42 @@ def __str__(self) -> str:
         return "false"
 
 
-class BoundReference:
-    """A reference bound to a field in a schema
+@dataclass
```
rdblue, line 226: frozen?

python/src/iceberg/expressions/base.py
```diff
@@ -138,6 +75,24 @@ def __invert__(self) -> "BooleanExpression":
...
 
 
+class BoundPredicate(ABC):
+    def __init__(self, term, literal):
+        """A concrete predicate must have a `term` and `literal`"""
+        self._term = term
+        self._literal = literal
+
+
+class UnboundPredicate:
```
rdblue, line 85: I think this and `BoundPredicate` should extend `BooleanExpression` because they return boolean values.

Also, there should be no need for `BoundPredicate` to extend ABC right now, since there are no methods defined for it that aren't defined.

python/src/iceberg/expressions/base.py
```diff
@@ -268,41 +223,42 @@ def __str__(self) -> str:
         return "false"
 
 
-class BoundReference:
-    """A reference bound to a field in a schema
+@dataclass
+class In(BooleanExpression, UnboundPredicate):
+    term: "UnboundReference"
+    literal: List[Literal]
```
rdblue, line 229: How about `literals` since this is a `list`? And should it be a `tuple`?

python/src/iceberg/expressions/base.py
```diff
@@ -138,6 +75,24 @@ def __invert__(self) -> "BooleanExpression":
...
+        """A concrete predicate must have a `term` and `literal`"""
+        self._term = term
+        self._literal = literal
+
+
+class UnboundPredicate:
+    def __init__(self, term, literal):
+        """A concrete predicate must have a `term` and `literal`"""
+        self._term = term
+        self._literal = literal
```
rdblue, line 89: Is there value in having this `__init__` method? I don't think `In` delegates to it.

Also, every `UnboundPredicate` will have a `term`, but not all will have one or more literals. `IsNull`, `NotNull`, `IsNan`, and `NotNan` are all unary and have no term. `In` and `NotIn` have more than one literal. And the comparison predicates, like `LessThan` and `Equal` have exactly one literal. In the Java version, we model this as a list or iterable of literals and ensure requirements are handled when binding. Then the bound versions are more strict about the literals.

python/src/iceberg/expressions/base.py
```diff
@@ -138,6 +75,24 @@ def __invert__(self) -> "BooleanExpression":
         ...
 
 
+class BoundPredicate(ABC):
+    def __init__(self, term, literal):
```
rdblue, line 79: Having a literal in `BoundPredicate` doesn't quite work, since different bound predicates will have different literal counts. See my note below. You may just want to leave this as an empty class for now.

python/tests/expressions/test_expressions_base.py
```diff
@@ -175,6 +138,50 @@ def test_strs(op, string):
     assert str(op) == string
 
 
+@pytest.mark.parametrize(
+    "a,  schema, case_sensitive, success",
+    [
+        (
+            base.In(base.UnboundReference("foo"), [literal("hello"), literal("world")]),
+            "table_schema_simple",
+            True,
```
rdblue, line 147: It's hard to read test cases that are parameterized like this, with boolean literals. Can you separate these into different test methods? I think that it makes more sense to split out `success` since nearly the entire test code is different depending on its value. And it makes sense to split out case sensitive as well since it is easier to follow if you pass it directly to `bind`"

```
    expr.bind(schema, case_sensitive=True).term.field == ...
```

python/tests/expressions/test_expressions_base.py
```diff
@@ -191,29 +198,38 @@ def test_strs(op, string):
...
+        (base.Or(TestExpressionA(), TestExpressionB()), base.And(TestExpressionB(), TestExpressionA()), False),
+        (base.Not(TestExpressionA()), TestExpressionA(), False),
+        (base.In(base.UnboundReference("foo"), [literal("hello"), literal("world")]), None, True),
+        (TestExpressionA(), TestExpressionB(), False),
     ],
 )
-def test_negate(lhs, rhs):
-    assert ~lhs == rhs
+def test_negate(lhs, rhs, raises):
+    if not raises:
```
rdblue, line 228: Can you split out these cases as well?

rdblue: It may be a good idea to go ahead and add `NotIn` so you don't waste time writing tests for temporary behavior.

rdblue:
Thanks, @CircArgs! Looking much closer.

python/pyiceberg/expressions/base.py
```diff
@@ -131,11 +75,115 @@ def __ge__(self, other):
...
+
+    Args:
+        field (NestedField): A referenced field in an Iceberg schema
+        accessor (Accessor): An Accessor object to access the value at the field's position
+    """
+
+    field: NestedField
+    accessor: Accessor
+
+    def eval(self, struct: StructProtocol) -> Any:
```
samredai, line 128: @rdblue in [this](https://github.com/apache/iceberg/pull/4816#discussion_r884327792) comment did you mean that we shouldn't include this method at all or just that we shouldn't include it as an `@abstractmethod` on `BoundPredicate`? From my understanding we need this here to retrieve the value of a reference given a row/struct as tested [here](https://github.com/apache/iceberg/blob/master/python/tests/expressions/test_expressions_base.py#L318-L320).

python/pyiceberg/expressions/base.py
```diff
@@ -131,11 +75,115 @@ def __ge__(self, other):
...
+class Unbound(Generic[T, B], ABC):
+    """Represents an unbound expression node."""
+
+    @abstractmethod
+    def bind(self, schema: Schema, case_sensitive: bool) -> B:
+        ...  # pragma: no cover
+
+
+class Term(ABC):
+    """An expression that evaluates to a value."""
```
samredai, line 101: I know this is from the [javadoc](https://iceberg.apache.org/javadoc/latest/org/apache/iceberg/expressions/Term.html), but is the word "expression" in this docstring a little confusing? Currently a "term" is either a reference or a transform of a reference. Should we just be explicit here and says `"A transform or reference that evaluates to a value."`?

python/pyiceberg/expressions/base.py
```diff
@@ -131,11 +75,115 @@ def __ge__(self, other):
...
+
+    Args:
+        field (NestedField): A referenced field in an Iceberg schema
+        accessor (Accessor): An Accessor object to access the value at the field's position
+    """
+
+    field: NestedField
+    accessor: Accessor
+
+    def eval(self, struct: StructProtocol) -> Any:
```
rdblue, line 128: I think that this should return `T`. That's the type that this reference produces. That may be a little difficult since `Accessor` is not parameterized by the type that it returns. If it is hard, we can fix it later.

python/pyiceberg/expressions/base.py
```diff
@@ -131,11 +75,115 @@ def __ge__(self, other):
...
+        accessor = schema.accessor_for_field(field.field_id)
+
+        if not accessor:
+            raise ValueError(f"Cannot find accessor for field '{self.name}' in schema: {schema}")
+
+        return BoundReference(field=field, accessor=accessor)
+
+
+@dataclass(frozen=True)  # type: ignore[misc]
+class BoundPredicate(Bound[T], BooleanExpression):
```
rdblue, line 178: I think this would be a `BoundSetPredicate`. Other bound predicates have only one or zero literals.

rdblue (approved):

rdblue:
Thanks, @CircArgs!

CircArgs:
> Thanks, @CircArgs!

Thank you @rdblue!

PR #4823, opened 2022-05-20
AWS: support Path-Style Access

dungdm93:
AWS S3 path-style API requests are deprecated. However, there are many S3-compatible services like MinIO or Ceph are not able to use Virtual Hosted-Style because of legacy deployment or organization restriction.

Up until now, to config path-style access, we need to write custom `AwsClientFactory`. So it would be nice if Iceberg's `DefaultAwsClientFactory` support `s3.path-style-access` (`false` by default)


danielcweeks (approved):
Thanks @dungdm93 !

PR #4824, opened 2022-05-20
Test: Make Flink unit tests run properly on Windows

hililiwei:

Currently, iceberg flink unit tests cannot be started on Windows OS, this makes it very inconvenient for developers who use Windows as their workhorse (like me, huh). This PR aims to make unit tests compatible with Windows and Linux at the same time.

## Brief change log

The main reason for incompatibility with Windows is the error related to the file path, specifically due to: "URISyntaxException: Relative Path in Absolute URI". For example, `file:C:\\Users` should be changed to `file:/C:/Users`.

I've added a method to handle windows path.

## Verifying this change

On my own Windows PC, verifying this change with the following command:
```
./gradlew -DsparkVersions= -DhiveVersions= -DflinkVersions=1.15 :iceberg-flink:iceberg-flink-1.15:check :iceberg-flink:iceberg-flink-runtime-1.15:check -Pquick=true -x javadoc
```


rdblue:
I don't think there's much interest in supporting windows directly. I'm going to close this, but feel free to bring up support elsewhere if you think that more people are interested.

hililiwei:
Why can't we support development with Windows? This is unfriendly to developers using Windows. 😢

PR #4826, opened 2022-05-20
Nessie: Use unique path for different table with same name

ajantha-bhat:
In Nessie catalog, 
Each branch can have a new table with the same name as a table in another branch. When this happens, Nessie is using the same table path two different tables (when same warehouse location is configured).

So, use the table UUID in the table path to avoid two different tables sharing the same path. 

Steps:

```
docker run -p 19120:19120 projectnessie/nessie

                              
 /Users/ajantha/Documents/tools/spark-3.0.3-bin-hadoop2.7/bin/spark-sql --packages org.slf4j:slf4j-api:1.7.32 --jars /Users/ajantha/Downloads/iceberg-spark-runtime-3.0_2.12-0.14.0-SNAPSHOT.jar,/Users/ajantha/Downloads/nessie-spark-extensions-0.28.0.jar \
                                    --conf spark.sql.catalog.nessie=org.apache.iceberg.spark.SparkCatalog\
                                    --conf spark.sql.catalog.nessie.warehouse=$PWD/nessie2 \
                                --conf spark.sql.catalog.nessie.catalog-impl=org.apache.iceberg.nessie.NessieCatalog \
                                --conf spark.sql.catalog.nessie.uri=http://localhost:19120/api/v1 \
                                --conf spark.sql.catalog.nessie.ref=main \
                                --conf spark.sql.catalog.nessie.auth-type=NONE \
                                --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,org.projectnessie.spark.extensions.NessieSparkSessionExtensions
```

```
create branch dev in nessie from main;

use reference dev in nessie;

create table nessie.db1.t1(id int) using iceberg;
insert into nessie.db1.t1 select 42;

use reference main in nessie;

create table nessie.db1.t1(place string, name string) using iceberg;
insert into nessie.db1.t1 select 'x','y';
```
-- Now the path `$warehouse/db1/t1` has data and metadata files from both the tables.

**Why this needs to be fixed ?**
`remove_orphan` (Also Nessie GC tool) files does a list files (`BaseDeleteOrphanFilesSparkAction#buildActualFileDF()` -> `listDirsRecursively`) and try to remove the files that are not reachable from metadata. So, it will clean up other table in this case.

Not sure if there are any other Iceberg operations that do list files. If it is there. It will get impacted too.

### DIFF
--- a/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
+++ b/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.function.Function;
 import org.apache.hadoop.conf.Configurable;
 import org.apache.hadoop.conf.Configuration;
@@ -199,10 +200,16 @@ public class NessieCatalog extends BaseMetastoreCatalog
 
   @Override
   protected String defaultWarehouseLocation(TableIdentifier table) {
+    String location;
     if (table.hasNamespace()) {
-      return SLASH.join(warehouseLocation, table.namespace().toString(), table.name());
+      location = SLASH.join(warehouseLocation, table.namespace().toString(), table.name());
+    } else {
+      location = SLASH.join(warehouseLocation, table.name());
     }
-    return SLASH.join(warehouseLocation, table.name());
+    // Different tables with same table name can exist across references in Nessie.
+    // To avoid sharing same table path between two tables with same name, use uuid in the table
+    // path.
+    return location + "_" + UUID.randomUUID();
   }
 
   @Override
--- a/nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
@@ -20,9 +20,9 @@ package org.apache.iceberg.nessie;
 
 import static org.apache.iceberg.types.Types.NestedField.required;
 
-import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Path;
 import java.util.List;
 import org.apache.avro.generic.GenericData.Record;
 import org.apache.hadoop.conf.Configuration;
@@ -80,7 +80,7 @@ public abstract class BaseTestIceberg {
 
   private static final Logger LOG = LoggerFactory.getLogger(BaseTestIceberg.class);
 
-  @TempDir public File temp;
+  @TempDir public Path temp;
 
   protected NessieCatalog catalog;
   protected NessieApiV1 api;
@@ -140,7 +140,7 @@ public abstract class BaseTestIceberg {
             .put("ref", ref)
             .put(CatalogProperties.URI, uri)
             .put("auth-type", "NONE")
-            .put(CatalogProperties.WAREHOUSE_LOCATION, temp.toURI().toString());
+            .put(CatalogProperties.WAREHOUSE_LOCATION, temp.toUri().toString());
     if (null != hash) {
       options.put("ref.hash", hash);
     }
--- a/nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
@@ -20,9 +20,9 @@ package org.apache.iceberg.nessie;
 
 import static org.apache.iceberg.types.Types.NestedField.required;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URI;
-import java.nio.file.Path;
 import java.util.List;
 import org.apache.avro.generic.GenericData.Record;
 import org.apache.hadoop.conf.Configuration;
@@ -80,7 +80,7 @@ public abstract class BaseTestIceberg {
 
   private static final Logger LOG = LoggerFactory.getLogger(BaseTestIceberg.class);
 
-  @TempDir public Path temp;
+  @TempDir public File temp;
 
   protected NessieCatalog catalog;
   protected NessieApiV1 api;
@@ -140,7 +140,7 @@ public abstract class BaseTestIceberg {
             .put("ref", ref)
             .put(CatalogProperties.URI, uri)
             .put("auth-type", "NONE")
-            .put(CatalogProperties.WAREHOUSE_LOCATION, temp.toUri().toString());
+            .put(CatalogProperties.WAREHOUSE_LOCATION, temp.toURI().toString());
     if (null != hash) {
       options.put("ref.hash", hash);
     }
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
@@ -484,4 +484,32 @@ public class TestBranchVisibility extends BaseTestIceberg {
     nessieCatalog.createTable(identifier2, schema);
     Assertions.assertThat(nessieCatalog.listTables(namespace)).hasSize(2);
   }
+
+  @Test
+  public void testDifferentTableSameName() throws NessieConflictException, NessieNotFoundException {
+    String branch1 = "branch1";
+    String branch2 = "branch2";
+    createBranch(branch1, null);
+    createBranch(branch2, null);
+    Schema schema1 =
+        new Schema(Types.StructType.of(required(1, "id", Types.LongType.get())).fields());
+    Schema schema2 =
+        new Schema(
+            Types.StructType.of(
+                    required(1, "file_count", Types.IntegerType.get()),
+                    required(2, "record_count", Types.LongType.get()))
+                .fields());
+
+    TableIdentifier identifier = TableIdentifier.of("db.", "table1");
+
+    NessieCatalog nessieCatalog = initCatalog(branch1);
+    Table table1 = nessieCatalog.createTable(identifier, schema1);
+    Assertions.assertThat(table1.schema().asStruct()).isEqualTo(schema1.asStruct());
+
+    nessieCatalog = initCatalog(branch2);
+    Table table2 = nessieCatalog.createTable(identifier, schema2);
+    Assertions.assertThat(table2.schema().asStruct()).isEqualTo(schema2.asStruct());
+
+    Assertions.assertThat(table1.location()).isNotEqualTo(table2.location());
+  }
 }
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
@@ -500,7 +500,7 @@ public class TestBranchVisibility extends BaseTestIceberg {
                     required(2, "record_count", Types.LongType.get()))
                 .fields());
 
-    TableIdentifier identifier = TableIdentifier.of("db.", "table1");
+    TableIdentifier identifier = TableIdentifier.of("db", "table1");
 
     NessieCatalog nessieCatalog = initCatalog(branch1);
     Table table1 = nessieCatalog.createTable(identifier, schema1);
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestCustomNessieClient.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestCustomNessieClient.java
@@ -46,7 +46,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
         "nessie",
         ImmutableMap.of(
             CatalogProperties.WAREHOUSE_LOCATION,
-            temp.toUri().toString(),
+            temp.toURI().toString(),
             CatalogProperties.URI,
             uri));
   }
@@ -58,7 +58,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
         "nessie",
         ImmutableMap.of(
             CatalogProperties.WAREHOUSE_LOCATION,
-            temp.toUri().toString(),
+            temp.toURI().toString(),
             CatalogProperties.URI,
             uri,
             NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL,
@@ -75,7 +75,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
                   "nessie",
                   ImmutableMap.of(
                       CatalogProperties.WAREHOUSE_LOCATION,
-                      temp.toUri().toString(),
+                      temp.toURI().toString(),
                       CatalogProperties.URI,
                       uri,
                       NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL,
@@ -94,7 +94,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
                   "nessie",
                   ImmutableMap.of(
                       CatalogProperties.WAREHOUSE_LOCATION,
-                      temp.toUri().toString(),
+                      temp.toURI().toString(),
                       CatalogProperties.URI,
                       uri,
                       NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL,
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestCustomNessieClient.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestCustomNessieClient.java
@@ -46,7 +46,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
         "nessie",
         ImmutableMap.of(
             CatalogProperties.WAREHOUSE_LOCATION,
-            temp.toURI().toString(),
+            temp.toUri().toString(),
             CatalogProperties.URI,
             uri));
   }
@@ -58,7 +58,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
         "nessie",
         ImmutableMap.of(
             CatalogProperties.WAREHOUSE_LOCATION,
-            temp.toURI().toString(),
+            temp.toUri().toString(),
             CatalogProperties.URI,
             uri,
             NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL,
@@ -75,7 +75,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
                   "nessie",
                   ImmutableMap.of(
                       CatalogProperties.WAREHOUSE_LOCATION,
-                      temp.toURI().toString(),
+                      temp.toUri().toString(),
                       CatalogProperties.URI,
                       uri,
                       NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL,
@@ -94,7 +94,7 @@ public class TestCustomNessieClient extends BaseTestIceberg {
                   "nessie",
                   ImmutableMap.of(
                       CatalogProperties.WAREHOUSE_LOCATION,
-                      temp.toURI().toString(),
+                      temp.toUri().toString(),
                       CatalogProperties.URI,
                       uri,
                       NessieConfigConstants.CONF_NESSIE_CLIENT_BUILDER_IMPL,
--- a/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
+++ b/nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
@@ -31,6 +31,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecordBuilder;
+import org.apache.commons.io.FileUtils;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.HasTableOperations;
 import org.apache.iceberg.ManifestFile;
@@ -80,7 +81,7 @@ public class TestNessieTable extends BaseTestIceberg {
                   optional(2, "data", Types.LongType.get()))
               .fields());
 
-  private File tableLocation;
+  private String tableLocation;
 
   public TestNessieTable() {
     super(BRANCH);
@@ -90,7 +91,8 @@ public class TestNessieTable extends BaseTestIceberg {
   @BeforeEach
   public void beforeEach(@NessieUri URI uri) throws IOException {
     super.beforeEach(uri);
-    this.tableLocation = new File(catalog.createTable(TABLE_IDENTIFIER, schema).location());
+    this.tableLocation =
+        catalog.createTable(TABLE_IDENTIFIER, schema).location().replaceFirst("file:", "");
   }
 
   @Override
@@ -98,7 +100,7 @@ public class TestNessieTable extends BaseTestIceberg {
   public void afterEach() throws Exception {
     // drop the table data
     if (tableLocation != null) {
-      tableLocation.delete();
+      FileUtils.deleteDirectory(new File(tableLocation));
       catalog.dropTable(TABLE_IDENTIFIER, false);
     }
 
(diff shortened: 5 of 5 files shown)


core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
```diff
@@ -157,19 +161,27 @@ public TableBuilder withProperty(String key, String value) {
       return this;
     }
 
+    public TableMetadata newTableMetadata() {
```
ajantha-bhat, line 164: Extracting the logic that needs to be overridden for Nessie. 

Without changing this class, need to copy-paste a lot of code from here to NessieCatalog for overriding. So, code duplication will be more.

ajantha-bhat:
cc: @nastra, @snazy, @RussellSpitzer, @rdblue

ajantha-bhat:
ping...

nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
```diff
@@ -443,4 +443,26 @@ public void testWithRefAndHash() throws NessieConflictException, NessieNotFoundE
...
+    String branch1 = "branch1";
+    String branch2 = "branch2";
+    createBranch(branch1, null);
+    createBranch(branch2, null);
+    Schema schema1 = new Schema(Types.StructType.of(required(1, "id", Types.LongType.get())).fields());
+    Schema schema2 = new Schema(Types.StructType.of(
+        required(1, "file_count", Types.IntegerType.get()),
+        required(2, "record_count", Types.LongType.get())).fields());
+
+    TableIdentifier identifier = TableIdentifier.of("db1.", "table1");
```
nastra, line 458: nit: replace `db1.` with `db`

nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
```diff
@@ -443,4 +443,26 @@ public void testWithRefAndHash() throws NessieConflictException, NessieNotFoundE
...
+
+    NessieCatalog nessieCatalog = initCatalog(branch1);
+    String location1 = nessieCatalog.createTable(identifier, schema1).location();
```
nastra, line 461: can we also verify that each table has the correct schema?

nastra (commented):
in general this LGTM, just a few things

nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
ajantha-bhat: I didn't add it initially as it is a base functionality and it is already validated in some other test suite.

added now.

core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
snazy, line 164: Should be `protected`

core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
```diff
@@ -157,19 +161,27 @@ public TableBuilder withProperty(String key, String value) {
...
+      String baseLocation = location != null ? location : defaultWarehouseLocation(identifier);
+      tableProperties.putAll(tableOverrideProperties());
+      return TableMetadata.newTableMetadata(schema, spec, sortOrder, baseLocation, tableProperties);
+    }
+
+    public TableMetadata replacedTableMetadata(TableOperations ops) {
```
snazy, line 170: Should be `protected`

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -256,4 +263,31 @@ String currentRefName() {
...
+    }
+
+    private TableMetadata updateLocation(TableMetadata tableMetadata) {
+      if (location() != null) {
+        // create the table at the location specified.
+        return tableMetadata;
+      }
+      // Different tables with same table name can exist across references in Nessie.
+      // To avoid sharing same table path between two tables with same name, use table uuid in the table path.
+      return tableMetadata.updateLocation(tableMetadata.location() + "_" + tableMetadata.uuid());
```
snazy, line 290: I'd really prefer a new, random UUID instead of the table's UUID.

ajantha-bhat: ok. Used the random UUID instead of table UUID.

ajantha-bhat:
@snazy , @nastra : Addressed the comments. Please have a look at it again.

nastra (approved):

snazy (approved):

nastra:
@RussellSpitzer could you review/merge this one please?

rdblue:
@ajantha-bhat, do you think Nessie could move to the REST catalog API? That API fixes several problems, like the one here. The table location should be something that the catalog endpoint can override, rather than requiring the client to produce the correct one.

core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
```diff
@@ -157,19 +161,21 @@ public TableBuilder withProperty(String key, String value) {
       return this;
     }
 
+    protected TableMetadata newTableMetadata() {
+      String baseLocation = location != null ? location : defaultWarehouseLocation(identifier);
```
rdblue, line 165: @ajantha-bhat, why not override `defaultWarehouseLocation`? That is already intended to let the catalog customize the table locations.

ajantha-bhat: @rdblue:  Yes, overriding  `defaultWarehouseLocation` can solve the problem of two different tables with same identifier getting the same path (issue which this PR is trying to solve)

But `defaultWarehouseLocation()` is the public interface of the catalog. 
If user calls it `defaultWarehouseLocation()` N times for a table, then it gives N different location. Which is confusing.
 
So, Instead of overrding it, I fixed as described in the PR.

ajantha-bhat:
> @ajantha-bhat, do you think Nessie could move to the REST catalog API? That API fixes several problems, like the one here. The table location should be something that the catalog endpoint can override, rather than requiring the client to produce the correct one.

I am not sure at the moment. I believe REST catalog is a client for some SERVER that implements the endpoints for these REST catalog calls. Nessie already has its own end points which are not same as what REST catalog expects. I think it is too late to change Nessie endpoints as it will break the compatibility. I might be wrong here. Please correct me If I am mistaken.

ajantha-bhat:
I rebased and resolved the conflict.

core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
RussellSpitzer: Why would we be calling default warehouse location multiple times for the same table? Sounds like we would be safe here saying that "defaultWarehouseLocation" is not deterministic.

ajantha-bhat: Because it is a user-facing API,  (nothing stops the user from calling in two places while writing custom logic around it)
I felt it looks odd to return a different result for the same input.

ajantha-bhat: @RussellSpitzer , @rdblue : I have replied above that why I didn't modify `defaultWarehouseLocation() `, If you still feel it is better to override `defaultWarehouseLocation() ` please let me know. I just want this issue to be fixed. Thanks.

dimas-b:
I'm a bit out-of-date of the REST catalog impl., but could it have a common base class with the NessieCatalog perhaps?.. just an idea (for a future change)

ajantha-bhat:
@rdblue , @RussellSpitzer : As there was no response, I assumed people are more inclined toward modifying the default warehouse location.  I have modified it as the same. Please help in merging this issue fix.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -400,7 +405,7 @@ private void validateRegister(TableIdentifier identifier, String metadataVersion
 
   @Test
   public void testRegisterTableWithGivenBranch() {
-    List<String> metadataVersionFiles = metadataVersionFiles(TABLE_NAME);
+    List<String> metadataVersionFiles = metadataVersionFiles(tableNameWithUUID(catalog.loadTable(TABLE_IDENTIFIER)));
```
rdblue, line 408: It seems odd to use the location to get a directory name, but then add the rest of the location back to list files. Why not just update `metadataVersionFiles` and `manifestFiles` to accept a full path?

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -181,10 +182,15 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
...
     if (table.hasNamespace()) {
-      return SLASH.join(warehouseLocation, table.namespace().toString(), table.name());
+      location = SLASH.join(warehouseLocation, table.namespace().toString(), table.name());
+    } else {
+      location = SLASH.join(warehouseLocation, table.name());
     }
-    return SLASH.join(warehouseLocation, table.name());
+    // Different tables with same table name can exist across references in Nessie.
+    // To avoid sharing same table path between two tables with same name, use uuid in the table path.
+    return location + "_" + UUID.randomUUID();
```
rdblue, line 193: This looks reasonable to me.

rdblue:
> I'm a bit out-of-date of the REST catalog impl., but could it have a common base class with the NessieCatalog perhaps?.. just an idea (for a future change)

I think what makes the most sense is to eventually use the same protocol, especially since the REST protocol solves a number of these challenges (like delegating location defaults to the service). It also uses a diff-based approach that can avoid a lot more conflicts. I'm not sure how easy it would be for Nessie to update to be compatible with that client API, though. And I imagine it isn't a high priority just to avoid commit conflicts and enable metadata caching.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
ajantha-bhat: ok. Updated now.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -531,27 +531,26 @@ protected String getTableLocation(String tableName) {
...
 
-  protected List<String> metadataVersionFiles(String tableName) {
-    return filterByExtension(tableName, getFileExtension(TableMetadataParser.Codec.NONE));
+  protected List<String> metadataVersionFiles(Path tablePath) {
+    return filterByExtension(tablePath.toString(), getFileExtension(TableMetadataParser.Codec.NONE));
   }
 
-  protected List<String> manifestFiles(String tableName) {
-    return filterByExtension(tableName, ".avro");
+  protected List<String> manifestFiles(Path tablePath) {
```
rdblue, line 548: Why is this passing a `Path`, only to convert to string and strip the `file:` from the `toString` representation in `filterByExtension`? Can you just pass `Path`?

ajantha-bhat: ok. Avoided `toString` conversion and  stripping the `file:`

ajantha-bhat:
@rdblue: I have squashed the commits and rebased because of the spotless apply.

ajantha-bhat:
@rdblue , @RussellSpitzer : ping...

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -561,28 +561,26 @@ protected String getTableLocation(String tableName) {
...
-    return Paths.get(getTableBasePath(tableName), "metadata").toString();
-  }
-
   @SuppressWarnings(
       "RegexpSinglelineJava") // respecting this rule requires a lot more lines of code
-  private List<String> metadataFiles(String tableName) {
-    return Arrays.stream(Objects.requireNonNull(new File(metadataLocation(tableName)).listFiles()))
+  private List<String> metadataFiles(String tablePath) {
+    return Arrays.stream(
+            Objects.requireNonNull(new File((tablePath + "/" + "metadata")).listFiles()))
```
rdblue, line 568: I'd rather not mix file system APIs. This uses both Path and File. Can you get rid of `Path` and make that a string everywhere?

ajantha-bhat: I believe the original author of this file might referred this from `HiveTableBaseTest` which uses both `Path` and `File`

ajantha-bhat: @rdblue : Removed `Path` from the whole Nessie testsuite and using `File`

we can even do this for other modules like `HiveTableBaseTest` (but may be in a follow-up)

nessie/src/test/java/org/apache/iceberg/nessie/TestCustomNessieClient.java
```diff
@@ -46,7 +46,7 @@ public void testNoCustomClient() {
         "nessie",
         ImmutableMap.of(
             CatalogProperties.WAREHOUSE_LOCATION,
-            temp.toUri().toString(),
```
rdblue, line 49: This was correct before. Location should be a full URI including the scheme.

ajantha-bhat: ok. I have updated now.  I hope it is ok now.

ajantha-bhat:
ping @rdblue

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -567,40 +565,29 @@ public void testGCEnabled() {
...
-    return new Path("file", null, Paths.get(getTableBasePath(tableName)).toString());
-  }
-
-  protected String getTableLocation(String tableName) {
-    return getTableLocationPath(tableName).toString();
-  }
-
-  private String metadataLocation(String tableName) {
-    return Paths.get(getTableBasePath(tableName), "metadata").toString();
+    return temp.toURI() + DB_NAME + "/" + tableName;
```
rdblue, line 568: Does this automatically call `toString` on the URI from `temp`?

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -567,40 +565,29 @@ public void testGCEnabled() {
...
   }
 
   @SuppressWarnings(
       "RegexpSinglelineJava") // respecting this rule requires a lot more lines of code
-  private List<String> metadataFiles(String tableName) {
-    return Arrays.stream(Objects.requireNonNull(new File(metadataLocation(tableName)).listFiles()))
+  private List<String> metadataFiles(File tablePath) {
+    return Arrays.stream(
+            Objects.requireNonNull(
+                new File((tablePath + "/" + "metadata").replace("file:", "")).listFiles()))
```
rdblue, line 576: This looks like a safe way to work with strings, but not with files. It looks like you're still fighting the representation used everywhere. Here, you convert file to string, do custom manipulation, and go back to file. That seems needlessly complicated and has implications for readability and correctness (what does `new File(...)` return that requires `"file:"` to need removed??).

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -92,15 +90,15 @@ public TestNessieTable() {
   @BeforeEach
   public void beforeEach(@NessieUri URI uri) throws IOException {
     super.beforeEach(uri);
-    this.tableLocation = new Path(catalog.createTable(TABLE_IDENTIFIER, schema).location());
+    this.tableLocation = new File(catalog.createTable(TABLE_IDENTIFIER, schema).location());
```
rdblue, line 93: I don't really see why this isn't a `String` that gets converted to `File` when needed.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
dimas-b: Yes, apparently this is what it compiles to:
`javap -cp nessie/build/classes/java/test/ -p -c -s org.apache.iceberg.nessie.TestNessieTable`

```
  private java.lang.String getTableBasePath(java.lang.String);
    descriptor: (Ljava/lang/String;)Ljava/lang/String;
    Code:
       0: new           #88                 // class java/lang/StringBuilder
       3: dup
       4: invokespecial #89                 // Method java/lang/StringBuilder."<init>":()V
       7: aload_0
       8: getfield      #90                 // Field temp:Ljava/io/File;
      11: invokevirtual #91                 // Method java/io/File.toURI:()Ljava/net/URI;
      14: invokevirtual #92                 // Method java/lang/StringBuilder.append:(Ljava/lang/Object;)Ljava/lang/StringBuilder;
      17: ldc           #93                 // String db
      19: invokevirtual #94                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      22: ldc           #95                 // String /
      24: invokevirtual #94                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      27: aload_1
      28: invokevirtual #94                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      31: invokevirtual #96                 // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      34: areturn
```

but it's different with java 11, something like:
```
24: invokedynamic #9,  0              // InvokeDynamic #0:makeConcatWithConstants:(Ljava/net/URI;)Ljava/lang/String;
```

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
ajantha-bhat: Hmm, I just followed the original author's style instead of refactoring it. 

Fixed now.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
ajantha-bhat: When `Table.location()` is called, it returns a string with `file:` scheme. So, while creating a `new File()` for a `ListFiles()`, if the `pathname` has the file scheme, `ListFiles` return null. Hence, removing the file scheme. Similar to how it is in many existing test cases. 

But instead of doing for every time, now I moved this logic to the `tablePath` initialization

ajantha-bhat:
@gaborkaszab: I would like to get this one in 1.1 release.

nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
```diff
@@ -80,7 +80,7 @@ public abstract class BaseTestIceberg {
 
   private static final Logger LOG = LoggerFactory.getLogger(BaseTestIceberg.class);
 
-  @TempDir public Path temp;
+  @TempDir public File temp;
```
nastra, line 83: what's the reason from switching from `Path` to `File`? The `File` API is considered **old** and `Path` should generally be preferred imo. Also the tests should work without having to change this. 
The only place you'd have to adjust is `TestNessieTable#testCreate()`: `String expected = (temp.toUri() + DB_NAME + "/" + tableName).replace("///", "/")` to `String expected = (temp.toUri() + DB_NAME + "/" + tableName)`

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -567,40 +567,29 @@ public void testGCEnabled() {
...
-    return Paths.get(getTableBasePath(tableName), "metadata").toString();
+    return temp.toURI() + DB_NAME + "/" + tableName;
   }
 
   @SuppressWarnings(
       "RegexpSinglelineJava") // respecting this rule requires a lot more lines of code
-  private List<String> metadataFiles(String tableName) {
-    return Arrays.stream(Objects.requireNonNull(new File(metadataLocation(tableName)).listFiles()))
+  private List<String> metadataFiles(String tablePath) {
+
```
nastra, line 576: nit: unnecessary whitespace

nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
ajantha-bhat: > https://github.com/apache/iceberg/pull/4826#discussion_r937079863
I'd rather not mix file system APIs. This uses both Path and File. Can you get rid of Path and make that a string everywhere?

Removed `Path` after this comment.

ajantha-bhat: yeah, `Path` is more advanced but we don't need any specific function from `Path` that is not present in `File`.

ajantha-bhat: > The only place you'd have to adjust

previously `metadataFiles`, `metadataVersionFiles`, `manifestFiles`, `metadataLocation` used to work with `tableName`, now because of uuid addition, it has to be location based. Hence, we started modifying the test cases and ended up here.

RussellSpitzer: I agree that the change here probably isn't necessary. I would probably revert this change

nessie/src/test/java/org/apache/iceberg/nessie/TestBranchVisibility.java
```diff
@@ -484,4 +484,32 @@ public void testWithRefAndHash() throws NessieConflictException, NessieNotFoundE
...
+    Schema schema1 =
+        new Schema(Types.StructType.of(required(1, "id", Types.LongType.get())).fields());
+    Schema schema2 =
+        new Schema(
+            Types.StructType.of(
+                    required(1, "file_count", Types.IntegerType.get()),
+                    required(2, "record_count", Types.LongType.get()))
+                .fields());
+
+    TableIdentifier identifier = TableIdentifier.of("db.", "table1");
```
RussellSpitzer, line 503: nit: why "db." ?

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
RussellSpitzer: I would probably reserve larger refactors like this to another PR, not sure they are required to avoid usage of file here?

nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
ajantha-bhat: > what's the reason for switching from Path to File?

If I switch to `Path` completely,  that also needs many changes as `TestNessieTable` validations/asserts has many  
`Assertions.assertThat(new File(foo)).exists();`
Also, some functions like `metadataFiles` use `listFiles()` that needs to be changed to `Files.walk()`.

dimas-b (commented):
LGTM overall with minor comments.

The renaming suggestion for `defaultWarehouseLocation` is really optional. I'd appreciate some clean-up in test code though.

dimas-b (approved):

nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java
nastra: to me it seems that the switch from `Path` to `File` in this PR isn't required. I actually checked out the branch and reverted that change and everything was working after adjusting `TestNessieTable#testCreate()`

ajantha-bhat: Do you mean, just with the **first commit** of this PR?
If different, can you share the branch with me?

nastra: I mean in this PR. The diff that is being done in this file isn't required imo

ajantha-bhat: Oh, you mean just this file. I have reverted now.

nastra (approved):

ajantha-bhat:
@RussellSpitzer , @nastra : I have reverted the base test class changes and handled the nit.

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -199,10 +200,16 @@ protected TableOperations newTableOps(TableIdentifier tableIdentifier) {
...
-      return SLASH.join(warehouseLocation, table.namespace().toString(), table.name());
+      location = SLASH.join(warehouseLocation, table.namespace().toString(), table.name());
+    } else {
+      location = SLASH.join(warehouseLocation, table.name());
     }
-    return SLASH.join(warehouseLocation, table.name());
+    // Different tables with same table name can exist across references in Nessie.
+    // To avoid sharing same table path between two tables with same name, use uuid in the table
+    // path.
+    return location + "_" + UUID.randomUUID();
```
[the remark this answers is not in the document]
ajantha-bhat: `defaultWarehouseLocation` -> my initial argument was same with Ryan. Later I realised, it is called only once.

ajantha-bhat:
@RussellSpitzer, @rdblue: Can we please merge this PR if it is ok?

RussellSpitzer (approved):
I know our contract on defaultWarehouseLocation is a bit iffy here, but I agree this is what was discussed as the proper solution previously. If there are no further complaints I think we can merge this as is and reconsider the contract for defaultWarehouse or having an ancillary method in another PR.

ajantha-bhat:
@RussellSpitzer: Thanks for the review and approval.

I think we can merge this as there is already a consensus for the fix (from everyone). 
The confusion and discussions were mainly about base test case refactoring. This can be done in follow-up too if there is any concern.  

@gaborkaszab: Can we please add `1.1.0 milestone` label for this?

Issue #4827, opened 2022-05-20
[Python] Define a TableBuilder API to create a Table, without Transaction API


dhruv-pratap:
This is a child issue of the parent epic #3227  - Support table in iceberg python library. Please refer the conversations in the parent tracker for design and the Work Breakdown Structure. 

**Scope**: 
I want to define a TableBuilder API to create a Table, without Transaction to begin with.

Mandatory attributes for a table to be created: 
- `Table identifier`
- `Schema`

Optional attributes for a table to be created: 
- `PartitionSpec`
- `SortOrder`
- `Location`
- `Properties`

**Out-of-scope**:
- Implementation of `create()`API is out-of-scope for this and will follow in a subsequent tracker under the parent tracker #3227

rdblue:
What is the purpose of a TableBuilder API for Python? I think that the current `create` method works well because you can already use named arguments and omit unnecessary metadata.

Fokko:
I believe we can close this one: https://github.com/apache/iceberg/blob/master/python/pyiceberg/catalog/__init__.py#L136-L161

PR #4829, opened 2022-05-20
[Python] Define a TableBuilder API to create a Table, without Transaction API (#4827)

dhruv-pratap:
**WORK IN PROGRESS** This is a draft pull request, and is NOT yet ready for review.

The purpose of this PR is to define a TableBuilder API to create a Table, without Transaction API to begin with. This adresses #4827 which is a child issue for the larger epic #3227


Fokko:
@dhruv-pratap Looks like most of this PR is already in there. If you're okay with it, I'll close this PR 👍🏻

PR #4830, opened 2022-05-20
Core: Add session catalog implementation for REST

rdblue:
This adds `RESTSessionCatalog` that implements `SessionCatalog`. This is the last PR to replace #4575.

`RESTSessionCatalog` implements `SessionCatalog` and adds support for session-based auth handling using OAuth2.

For the catalog:
* If the catalog is configured with credentials, those credentials are used to fetch a bearer token using OAuth2 client credentials flow, and that token is used for the `Authorization` header
* If the catalog is configured with a bearer token, that token is used in the `Authorization` header.

Session context is updated to have a map of credentials. For each session context:
* If the session context is configured with a Bearer token (`"token"` credential), that token is used as an access token in the `Authorization` header
* If the session context is configured with a credential (`"credential"` credential), those credentials are used to fetch a bearer token using OAuth2 client credentials flow, and that token is used for the `Authorization` header
* If the session context has any OAuth2 token type as a key in its credential map, that token is exchanged for an access token using OAuth2 token exchange flow, and the access token is used for the `Authorization` header. This exchange uses the catalog's token, if present, as the `actor_token`.

When a table is loaded:
* If the table config contains a Bearer token, that token is used for the `Authorization` header
* If the table config contains credentials, the credentials are used to fetch an access token using OAuth2 client credentials flow, and that token is used for the `Authorization` header.
* If the table config contains any other OAuth2 token type, that token is exchanged for an access token using OAuth2 token exchange flow, and the access token is used for the `Authorization` header. This exchange uses the context token or catalog token as `actor_token`.

A scheduled thread pool is used to refresh tokens using the token exchange flow with no actor token. The token exchange uses the token that is being refreshed in the `Authorization` header.


rdblue:
Looks like the file names are misleading git. `RESTSessionCatalog` is based on `RESTCatalog`, so it is more useful to diff those. The `RESTCatalog` implementation in this PR is new and should not be diffed with the previous.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+        ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), fullConf, this.conf);
+    RESTClient tableClient = clientBuilder.apply(fullConf);
+
+    return Pair.of(tableClient, tableIO);
+  }
+
+  /**
+   * Class to handle authorization headers and token refresh.
+   */
+  private class AuthSession {
```
rdblue, line 525: This class handles refreshing tokens (when `refresh()` is called) and maintains headers. Instances are passed as `Supplier<Map<String, String>>` from `session::headers` to the `HTTPClient`, which gets the header map just before each request.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+      return token;
+    }
+
+    public String tokenType() {
+      return tokenType;
+    }
+
+    public AuthSession refresh() {
+      Map<String, String> request = OAuth2Util.tokenExchangeRequest(
+          token, tokenType, ImmutableList.of(RESTCatalogProperties.CATALOG_SCOPE));
```
rdblue, line 550: The token exchange request uses the current token in the `Authorization` header and as the `subject_token` to refresh.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+
+    String ioImpl = mergedProps.get(CatalogProperties.FILE_IO_IMPL);
+    this.io = CatalogUtil.loadFileIO(ioImpl != null ? ioImpl : ResolvingFileIO.class.getName(), mergedProps, conf);
+
+    super.initialize(name, mergedProps);
+  }
+
+  Supplier<Map<String, String>> headers(SessionContext context) {
+    AuthSession session = sessions.get(context.sessionId(), id -> {
+      Map<String, String> headers = catalogAuth.headers();
```
rdblue, line 145: Both flows (client credentials and token exchange) here will use the catalog's token in the `Authorization` header.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+
+      OAuthTokenResponse response;
+      if (context.credential() != null) {
+        response = fetchClientCredentials(
+            client, headers, context.credential(), RESTCatalogProperties.CATALOG_SCOPE);
+      } else if (context.properties().containsKey(RESTCatalogProperties.TOKEN)) {
+        response = exchangeToken(
```
rdblue, line 152: In addition to using the catalog's token in the `Authorization` header, this token exchange also uses the catalog token as the actor token.

core/src/main/java/org/apache/iceberg/rest/OAuth2Util.java
```diff
@@ -0,0 +1,142 @@
...
+  private static Pair<String, String> parseCredential(String credential) {
+    Preconditions.checkNotNull(credential, "Invalid credential: null");
+    List<String> parts = CREDENTIAL_SPLITTER.splitToList(credential);
+    switch (parts.size()) {
+      case 2:
+        // client ID and client secret
+        return Pair.of(parts.get(0), parts.get(1));
+      case 1:
+        // client ID
+        return Pair.of(null, parts.get(0));
```
kbendick, line 114: Is it intended that `client ID` is sometimes pair[0] and sometimes pair[1]?

That's how it looks presently based on the comments and return statements.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+  public void initialize(String name, Map<String, String> props) {
+    Preconditions.checkArgument(props != null, "Invalid configuration: null");
+
+    String initToken = props.get(RESTCatalogProperties.TOKEN);
+
+    // fetch auth and config to complete initialization
+    ConfigResponse config;
+    OAuthTokenResponse authResponse;
+    try (RESTClient initClient = clientBuilder.apply(props)) {
+      Map<String, String> initHeaders = merge(configHeaders(props), authHeaders(initToken));
```
rdblue, line 105: This starts off using a token if it is available. If both `token` and `credential` are set, the token is used for the `Authorization` header to exchange the credentials for a token that will be used. If the credential isn't set, then a session is created from the token later.

core/src/main/java/org/apache/iceberg/rest/OAuth2Util.java
rdblue: No. The secret is always present but the client ID may not be.

core/src/main/java/org/apache/iceberg/rest/HTTPClient.java
```diff
@@ -269,12 +269,6 @@ public Builder withHeaders(Map<String, String> headers) {
       return this;
     }
 
-    public Builder withBearerAuth(String token) {
-      Preconditions.checkNotNull(token, "Invalid auth token: null");
-      baseHeaders.put(HttpHeaders.AUTHORIZATION, asBearer(token));
```
danielcweeks, line 274: nit: seems like we can get by without a separate method to format the bearer token?

danielcweeks: We could also use the static value from RESTCatalogProperties as opposed to hard coded 'Bearer'

rdblue: Yeah, I'll remove the unnecessary method as well. That should fix both.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+        scope != null ? ImmutableList.of(scope) : ImmutableList.of());
+
+    OAuthTokenResponse response = client.postForm(
+        ResourcePaths.token(), request, OAuthTokenResponse.class, headers, ErrorHandlers.defaultErrorHandler());
+    response.validate();
+
+    return response;
+  }
+
+  public static OAuthTokenResponse fetchClientCredentials(RESTClient client, Map<String, String> headers,
```
danielcweeks, line 615: nit: naming is actually backwards here.  We're actually fetching a token (not credentials).

core/src/main/java/org/apache/iceberg/rest/OAuth2Util.java
kbendick: Oh. That's what I had thought. Then I believe the comment on line 113 should say `// client secret` and not `// client ID`, if that's what's present (and get's placed into the second part of the pair every time).

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,626 @@
...
+      OAuthTokenResponse response;
+      if (context.credential() != null) {
+        response = fetchClientCredentials(
+            client, headers, context.credential(), RESTCatalogProperties.CATALOG_SCOPE);
+      } else if (context.properties().containsKey(RESTCatalogProperties.TOKEN)) {
+        response = exchangeToken(
+            client,
+            headers,
+            context.properties().get(RESTCatalogProperties.TOKEN),
+            "urn:ietf:params:oauth:token-type:access_token",
```
danielcweeks, line 157: Might consider creating an `OAuth2Constants` or putting this in `OAuth2Util` as a constant

core/src/main/java/org/apache/iceberg/rest/OAuth2Util.java
```diff
@@ -0,0 +1,142 @@
...
+    formData.put(GRANT_TYPE, CLIENT_CREDENTIALS);
+    if (clientId != null) {
+      formData.put(CLIENT_ID, clientId);
+    }
+    formData.put(CLIENT_SECRET, clientSecret);
+
+    ImmutableList.Builder<String> scopeBuilder = ImmutableList.builder();
+    scopeBuilder.addAll(scopes);
+    scopeBuilder.add(CATALOG);
+    formData.put(SCOPE, SCOPE_JOINER.join(scopes));
```
kbendick, line 138: Nit: Could call `toScope(scopes)` here to reduce the usage of `SCOPE_JOINER` to _just_ that function.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
danielcweeks: That may have hit the wrong line.  I was referring to `urn:ietf:params:oauth:token-type:access_token`

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
danielcweeks: I understand that this is the client credential flow, but it feels awkward to read (maybe `fetchClientCredentialFlow`?)

danielcweeks:
Overall, the logic and sequencing looks sound.  The one thing I might suggest is actually creating a separate package `org...iceberg.rest.auth` and put much of the auth code there.  RESTSessionCatalog class not has a lot of internal logic that's really about OAuth2 which could be pulled out separately (e.g. `AuthSession` and the static methods supporting flows).  Some minor refactoring could help in terms of adding other auth in the future.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
kbendick: +1. There's a number of these long strings (which are defined in the OAuth2 spec) that would be easier to read with names.

core/src/main/java/org/apache/iceberg/rest/OAuth2Util.java
rdblue: Yeah, this looks suspicious to me. I'll rework it.

core/src/main/java/org/apache/iceberg/rest/responses/OAuthTokenResponse.java
```diff
@@ -0,0 +1,122 @@
...
+package org.apache.iceberg.rest.responses;
+
+import java.util.List;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.rest.OAuth2Util;
+import org.apache.iceberg.rest.RESTResponse;
+
+public class OAuthTokenResponse implements RESTResponse {
```
bryanck, line 29: It looked like this is deserialized using `ObjectMapper.readValue()` but I didn't see any setting for handling snake case field names, so just wanted to confirm that works...

rdblue: I added a serialization implementation and tests in #4833. Thanks for catching this, @bryanck!

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -0,0 +1,627 @@
...
+
+    // fetch auth and config to complete initialization
+    ConfigResponse config;
+    OAuthTokenResponse authResponse;
+    try (RESTClient initClient = clientBuilder.apply(props)) {
+      Map<String, String> initHeaders = merge(configHeaders(props), authHeaders(initToken));
+      String credential = props.get(RESTCatalogProperties.CREDENTIAL);
+      if (credential != null && !credential.isEmpty()) {
+        String scope = props.getOrDefault(RESTCatalogProperties.SCOPE, RESTCatalogProperties.CATALOG_SCOPE);
+        authResponse = fetchClientCredentials(initClient, initHeaders, credential, scope);
```
bryanck, line 109: It might be nice to have the option for a different URL base for auth requests, to support an authorization server separate from the catalog server

core/src/main/java/org/apache/iceberg/rest/OAuth2Util.java
rdblue: Yes, it should. Fixed in #4833.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
rdblue: Renamed to `fetchToken`.

baseline.gradle
```diff
@@ -82,6 +82,7 @@ subprojects {
           '-Xep:StrictUnusedVariable:OFF',
           '-Xep:TypeParameterShadowing:OFF',
           '-Xep:TypeParameterUnusedInFormals:OFF',
+          '-Xep:DangerousThreadPoolExecutorUsage:OFF',
```
rdblue, line 85: This was needed to add `ThreadPools.newScheduledPool`. For some reason, errorprone doesn't like it when you create a `ScheduledExecutorService`.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -222,11 +278,61 @@ public boolean updateNamespaceMetadata(SessionContext context, Namespace ns,
...
+
+    return refreshExecutor;
+  }
+
+  private void scheduleTokenRefresh(
+      AuthSession session, long startTimeMillis, long expiresIn, TimeUnit unit) {
+    // convert expiration interval to milliseconds
+    long expiresInMillis = unit.toMillis(expiresIn);
+    // how much ahead of time to start the request to allow it to complete
+    long refreshWindowMillis = Math.min(expiresInMillis / 10, MAX_REFRESH_WINDOW_MILLIS);
```
rdblue, line 298: The logic here is to start the refresh with 10% of the retry window left, but no more than some maximum (in case expires-in is long, like 10 hours).

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -87,18 +106,54 @@ public RESTSessionCatalog() {
...
+        config = fetchConfig(initClient, initHeaders);
+      }
+    } catch (IOException e) {
+      throw new UncheckedIOException("Failed to close HTTP client", e);
+    }
+
+    // build the final configuration and set up the catalog's auth
+    Map<String, String> mergedProps = config.merge(props);
+    Map<String, String> baseHeaders = configHeaders(mergedProps);
+    this.catalogAuth = new AuthSession(baseHeaders, null, null);
```
danielcweeks, line 134: Minor flow question here:  Why not just put this in an `else` block below as opposed to setting it and then resetting it in the following statements?

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -87,18 +106,54 @@ public RESTSessionCatalog() {
...
+    if (authResponse != null) {
+      this.catalogAuth = newSession(authResponse, startTimeMillis, catalogAuth);
+    } else if (initToken != null) {
+      this.catalogAuth = newSession(initToken, expiresInMs(mergedProps), catalogAuth);
```
danielcweeks, line 138: This feels a little awkward because if there's a token initially, the init doesn't go through the token exchange flow, so we don't actually know when the token is supposed to expire but still scheduling to refresh.  It feels like we should either exchange immediately and use the expires response to schedule the interval or assume that a set token has no expiration.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -222,11 +278,61 @@ public boolean updateNamespaceMetadata(SessionContext context, Namespace ns,
...
+    long elapsedMillis = System.currentTimeMillis() - startTimeMillis;
+    // how much time to actually wait
+    long timeToWait = Math.max(waitIntervalMillis - elapsedMillis, MIN_REFRESH_WAIT_MILLIS);
+
+    tokenRefreshExecutor().schedule(
+        () -> {
+          long refreshStartTime = System.currentTimeMillis();
+          Pair<Integer, TimeUnit> expiration = session.refresh(client);
+          if (expiration != null) {
+            scheduleTokenRefresh(session, refreshStartTime, expiration.first(), expiration.second());
```
danielcweeks, line 311: While I think it's probably ok for failing fast in initial OAuth exchanges, I think we should probably wrap this in a tasks for retry.  Failing to refresh the token due to intermittent issues would cause critical failure for long-running processes like Flink/Trino.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
rdblue: This is used as the parent for the sessions created in the `if`/`else` cases below. That ensures that those token exchanges use the correct base headers from merged properties.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
rdblue: Good point. I'll wrap the `session.refresh` call.

PR #4831, opened 2022-05-21
Add Parquet Row Group Bloom Filter Support

huaxingao:
Co-Authored-By: Xi Chen [jshmchenxi@163.com](mailto:jshmchenxi@163.com)
Co-Authored-By: Hao Lin [linhao@qiyi.com](mailto:linhao@qiyi.com)
Co-Authored-By: Huaxin Gao [huaxin_gao@apple.com](mailto:huaxin_gao@apple.com)

Currently, Iceberg has `ParquetMetricsRowGroupFilter` and `ParquetDictionaryRowGroupFilter`. This PR adds one more filter `ParquetBloomRowGroupFilter`, which takes advantage of the Parquet bloom filter to find out if a row group needs to be read or not.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/data/Record.java
+++ b/api/src/main/java/org/apache/iceberg/data/Record.java
@@ -58,4 +58,22 @@ public interface Record extends StructLike {
     return copy(overwriteValues);
   }
 
+  default Record copy(String field1, Object value1, String field2, Object value2, String field3, Object value3,
+                      String field4, Object value4, String field5, Object value5, String field6, Object value6,
+                      String field7, Object value7, String field8, Object value8, String field9, Object value9,
+                      String field10, Object value10, String field11, Object value11) {
+    Map<String, Object> overwriteValues = Maps.newHashMapWithExpectedSize(9);
+    overwriteValues.put(field1, value1);
+    overwriteValues.put(field2, value2);
+    overwriteValues.put(field3, value3);
+    overwriteValues.put(field4, value4);
+    overwriteValues.put(field5, value5);
+    overwriteValues.put(field6, value6);
+    overwriteValues.put(field7, value7);
+    overwriteValues.put(field8, value8);
+    overwriteValues.put(field9, value9);
+    overwriteValues.put(field10, value10);
+    overwriteValues.put(field11, value11);
+    return copy(overwriteValues);
+  }
 }
--- a/api/src/main/java/org/apache/iceberg/data/Record.java
+++ b/api/src/main/java/org/apache/iceberg/data/Record.java
@@ -57,23 +57,4 @@ public interface Record extends StructLike {
     overwriteValues.put(field3, value3);
     return copy(overwriteValues);
   }
-
-  default Record copy(String field1, Object value1, String field2, Object value2, String field3, Object value3,
-                      String field4, Object value4, String field5, Object value5, String field6, Object value6,
-                      String field7, Object value7, String field8, Object value8, String field9, Object value9,
-                      String field10, Object value10, String field11, Object value11) {
-    Map<String, Object> overwriteValues = Maps.newHashMapWithExpectedSize(11);
-    overwriteValues.put(field1, value1);
-    overwriteValues.put(field2, value2);
-    overwriteValues.put(field3, value3);
-    overwriteValues.put(field4, value4);
-    overwriteValues.put(field5, value5);
-    overwriteValues.put(field6, value6);
-    overwriteValues.put(field7, value7);
-    overwriteValues.put(field8, value8);
-    overwriteValues.put(field9, value9);
-    overwriteValues.put(field10, value10);
-    overwriteValues.put(field11, value11);
-    return copy(overwriteValues);
-  }
 }
--- a/api/src/main/java/org/apache/iceberg/data/Record.java
+++ b/api/src/main/java/org/apache/iceberg/data/Record.java
@@ -62,7 +62,7 @@ public interface Record extends StructLike {
                       String field4, Object value4, String field5, Object value5, String field6, Object value6,
                       String field7, Object value7, String field8, Object value8, String field9, Object value9,
                       String field10, Object value10, String field11, Object value11) {
-    Map<String, Object> overwriteValues = Maps.newHashMapWithExpectedSize(9);
+    Map<String, Object> overwriteValues = Maps.newHashMapWithExpectedSize(11);
     overwriteValues.put(field1, value1);
     overwriteValues.put(field2, value2);
     overwriteValues.put(field3, value3);
--- a/api/src/main/java/org/apache/iceberg/data/Record.java
+++ b/api/src/main/java/org/apache/iceberg/data/Record.java
@@ -57,4 +57,5 @@ public interface Record extends StructLike {
     overwriteValues.put(field3, value3);
     return copy(overwriteValues);
   }
+
 }
--- a/api/src/main/java/org/apache/iceberg/expressions/Binder.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Binder.java
@@ -96,7 +96,11 @@ public class Binder {
     }
     ReferenceVisitor visitor = new ReferenceVisitor();
     for (Expression expr : exprs) {
-      ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
+      if (expr instanceof Unbound) {
+        ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
+      } else {
+        ExpressionVisitors.visit(expr, visitor);
+      }
     }
     return visitor.references;
   }
--- a/api/src/main/java/org/apache/iceberg/expressions/Binder.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Binder.java
@@ -91,12 +91,17 @@ public class Binder {
   }
 
   public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
+    return boundReferences(struct, exprs, caseSensitive, false);
+  }
+
+  public static Set<Integer> boundReferences(
+      StructType struct, List<Expression> exprs, boolean caseSensitive, boolean alreadyBound) {
     if (exprs == null) {
       return ImmutableSet.of();
     }
     ReferenceVisitor visitor = new ReferenceVisitor();
     for (Expression expr : exprs) {
-      if (expr instanceof Unbound) {
+      if (!alreadyBound) {
         ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
       } else {
         ExpressionVisitors.visit(expr, visitor);
--- a/api/src/main/java/org/apache/iceberg/expressions/Binder.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Binder.java
@@ -91,10 +91,10 @@ public class Binder {
   }
 
   public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
-    return boundReferences(struct, exprs, caseSensitive, false);
+    return exprReferences(struct, exprs, caseSensitive, false);
   }
 
-  public static Set<Integer> boundReferences(
+  public static Set<Integer> exprReferences(
       StructType struct, List<Expression> exprs, boolean caseSensitive, boolean alreadyBound) {
     if (exprs == null) {
       return ImmutableSet.of();
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -167,12 +167,7 @@ public class TableProperties {
       "write.delete.parquet.row-group-check-max-record-count";
   public static final int PARQUET_ROW_GROUP_CHECK_MAX_RECORD_COUNT_DEFAULT = 10000;
 
-  public static final String DEFAULT_PARQUET_BLOOM_FILTER_ENABLED = "write.parquet.bloom-filter-enabled.default";
-  public static final boolean DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT = false;
-
   public static final String PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX = "write.parquet.bloom-filter-enabled.column.";
-  public static final String PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX =
-      "write.parquet.bloom-filter-expected-ndv.column.";
 
   public static final String PARQUET_BLOOM_FILTER_MAX_BYTES = "write.parquet.bloom-filter-max-bytes";
   public static final int PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT = 1024 * 1024;
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -167,6 +167,17 @@ public class TableProperties {
       "write.delete.parquet.row-group-check-max-record-count";
   public static final int PARQUET_ROW_GROUP_CHECK_MAX_RECORD_COUNT_DEFAULT = 10000;
 
+  public static final String DEFAULT_PARQUET_BLOOM_FILTER_ENABLED = "write.parquet.bloom-filter-enabled.default";
+  public static final boolean DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT = false;
+
+  public static final String PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX = "write.parquet.bloom-filter-enabled.column.";
+  public static final String PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX =
+      "write.parquet.bloom-filter-expected-ndv.column.";
+
+  public static final String PARQUET_BLOOM_FILTER_MAX_BYTES = "write.parquet.bloom-filter-max-bytes";
+  public static final int PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT = 1024 * 1024;
+
+
   public static final String AVRO_COMPRESSION = "write.avro.compression-codec";
   public static final String DELETE_AVRO_COMPRESSION = "write.delete.avro.compression-codec";
   public static final String AVRO_COMPRESSION_DEFAULT = "gzip";
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -177,7 +177,6 @@ public class TableProperties {
   public static final String PARQUET_BLOOM_FILTER_MAX_BYTES = "write.parquet.bloom-filter-max-bytes";
   public static final int PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT = 1024 * 1024;
 
-
   public static final String AVRO_COMPRESSION = "write.avro.compression-codec";
   public static final String DELETE_AVRO_COMPRESSION = "write.delete.avro.compression-codec";
   public static final String AVRO_COMPRESSION_DEFAULT = "gzip";
--- a/docs/tables/configuration.md
+++ b/docs/tables/configuration.md
@@ -50,6 +50,10 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.parquet.dict-size-bytes      | 2097152 (2 MB)     | Parquet dictionary page size                       |
 | write.parquet.compression-codec    | gzip               | Parquet compression codec: zstd, brotli, lz4, gzip, snappy, uncompressed |
 | write.parquet.compression-level    | null               | Parquet compression level                          |
+| write.parquet.bloom-filter-enabled.default | false | Whether to enable writing bloom filter for all columns |
+| write.parquet.bloom-filter-enabled.column.col1 | (not set) | Whether to enable writing bloom filter for column 'col1' to allow per-column configuration; This property overrides `bloom-filter-enabled.default` for the specified column; For example, setting both `write.parquet.bloom-filter-enabled.default=true` and `write.parquet.bloom-filter-enabled.column.some_col=false` will enable bloom filter for all columns except `some_col` |
+| write.parquet.bloom-filter-expected-ndv.column.col1 | (not set) | The expected number of distinct values in a column, it is used to compute the optimal size of the bloom filter; Note that the NDV is specific for a parquet file. If this property is not set, the bloom filter will use the maximum size set in `bloom-filter-max-bytes`; If this property is set for a column, then no need to enable the bloom filter with `write.parquet.bloom-filter-enabled` property |
+| write.parquet.bloom-filter-max-bytes | 1048576 (1 MB) | The maximum number of bytes for a bloom filter bitset |
 | write.avro.compression-codec       | gzip               | Avro compression codec: gzip(deflate with 9 level), zstd, snappy, uncompressed |
 | write.avro.compression-level       | null               | Avro compression level                             |
 | write.orc.stripe-size-bytes        | 67108864 (64 MB)   | Define the default ORC stripe size, in bytes       |
--- a/docs/tables/configuration.md
+++ b/docs/tables/configuration.md
@@ -49,10 +49,8 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.parquet.page-size-bytes      | 1048576 (1 MB)     | Parquet page size                                  |
 | write.parquet.dict-size-bytes      | 2097152 (2 MB)     | Parquet dictionary page size                       |
 | write.parquet.compression-codec    | gzip               | Parquet compression codec: zstd, brotli, lz4, gzip, snappy, uncompressed |
-| write.parquet.compression-level    | null               | Parquet compression level                          |
-| write.parquet.bloom-filter-enabled.default | false | Whether to enable writing bloom filter for all columns |
-| write.parquet.bloom-filter-enabled.column.col1 | (not set) | Whether to enable writing bloom filter for column 'col1' to allow per-column configuration; This property overrides `bloom-filter-enabled.default` for the specified column; For example, setting both `write.parquet.bloom-filter-enabled.default=true` and `write.parquet.bloom-filter-enabled.column.some_col=false` will enable bloom filter for all columns except `some_col` |
-| write.parquet.bloom-filter-expected-ndv.column.col1 | (not set) | The expected number of distinct values in a column, it is used to compute the optimal size of the bloom filter; Note that the NDV is specific for a parquet file. If this property is not set, the bloom filter will use the maximum size set in `bloom-filter-max-bytes`; If this property is set for a column, then no need to enable the bloom filter with `write.parquet.bloom-filter-enabled` property |
+| write.parquet.compression-level    | null               | Parquet compression level
+| write.parquet.bloom-filter-enabled.column.col1 | (not set) | Enables writing a bloom filter for the column |
 | write.parquet.bloom-filter-max-bytes | 1048576 (1 MB) | The maximum number of bytes for a bloom filter bitset |
 | write.avro.compression-codec       | gzip               | Avro compression codec: gzip(deflate with 9 level), zstd, snappy, uncompressed |
 | write.avro.compression-level       | null               | Avro compression level                             |
--- a/parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
+++ b/parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
@@ -245,13 +245,10 @@ public class Parquet {
       CompressionCodecName codec = context.codec();
       int rowGroupCheckMinRecordCount = context.rowGroupCheckMinRecordCount();
       int rowGroupCheckMaxRecordCount = context.rowGroupCheckMaxRecordCount();
-      boolean bloomFilterEnabled = PropertyUtil.propertyAsBoolean(config, DEFAULT_PARQUET_BLOOM_FILTER_ENABLED,
-          DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT);
-      int bloomFilterMaxBytes = PropertyUtil.propertyAsInt(config, PARQUET_BLOOM_FILTER_MAX_BYTES,
-          PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT);
-      Map<String, String> columnBloomFilterModes = getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX);
-      Map<String, String> columnBloomFilterNDVs =
-          getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX);
+      boolean bloomFilterEnabled = context.bloomFilterEnabled();
+      int bloomFilterMaxBytes = context.bloomFilterMaxBytes();
+      Map<String, String> columnBloomFilterEnabled = context.columnBloomFilterEnabled();
+      Map<String, String> columnBloomFilterNDVs = context.columnBloomFilterNDVs();
 
       if (compressionLevel != null) {
         switch (codec) {
@@ -291,16 +288,16 @@ public class Parquet {
             .withMaxBloomFilterBytes(bloomFilterMaxBytes)
             .withBloomFilterEnabled(bloomFilterEnabled);
 
-        for (Map.Entry<String, String> entry : columnBloomFilterModes.entrySet()) {
-          String col = entry.getKey();
-          String value = entry.getValue();
-          propsBuilder.withBloomFilterEnabled(col, Boolean.valueOf(value));
+        for (Map.Entry<String, String> entry : columnBloomFilterEnabled.entrySet()) {
+          String colPath = entry.getKey();
+          String bloomEnabled = entry.getValue();
+          propsBuilder.withBloomFilterEnabled(colPath, Boolean.valueOf(bloomEnabled));
         }
 
         for (Map.Entry<String, String> entry : columnBloomFilterNDVs.entrySet()) {
-          String col = entry.getKey();
-          String value = entry.getValue();
-          propsBuilder.withBloomFilterNDV(col, Long.valueOf(value));
+          String colPath = entry.getKey();
+          String numDistinctValue = entry.getValue();
+          propsBuilder.withBloomFilterNDV(colPath, Long.valueOf(numDistinctValue));
         }
 
         ParquetProperties parquetProperties = propsBuilder.build();
@@ -323,16 +320,16 @@ public class Parquet {
             // TODO: add .withMaxBloomFilterBytes(bloomFilterMaxBytes) once ParquetWriter.Builder supports it
             .withBloomFilterEnabled(bloomFilterEnabled);
 
-        for (Map.Entry<String, String> entry : columnBloomFilterModes.entrySet()) {
-          String col = entry.getKey();
-          String value = entry.getValue();
-          parquetWriteBuilder.withBloomFilterEnabled(col, Boolean.valueOf(value));
+        for (Map.Entry<String, String> entry : columnBloomFilterEnabled.entrySet()) {
+          String colPath = entry.getKey();
+          String bloomEnabled = entry.getValue();
+          parquetWriteBuilder.withBloomFilterEnabled(colPath, Boolean.valueOf(bloomEnabled));
         }
 
         for (Map.Entry<String, String> entry : columnBloomFilterNDVs.entrySet()) {
-          String col = entry.getKey();
-          String value = entry.getValue();
-          parquetWriteBuilder.withBloomFilterNDV(col, Long.valueOf(value));
+          String colPath = entry.getKey();
+          String numDistinctValue = entry.getValue();
+          parquetWriteBuilder.withBloomFilterNDV(colPath, Long.valueOf(numDistinctValue));
         }
 
         return new ParquetWriteAdapter<>(
@@ -341,16 +338,16 @@ public class Parquet {
       }
     }
 
-    private Map<String, String> getBloomColumnConfigMap(String prefix) {
-      Map<String, String> columnBloomFilterModes = Maps.newHashMap();
+    private static Map<String, String> getBloomColumnConfigMap(String prefix, Map<String, String> config) {
+      Map<String, String> columnBloomFilterConfig = Maps.newHashMap();
       config.keySet().stream()
           .filter(key -> key.startsWith(prefix))
           .forEach(key -> {
-            String columnAlias = key.replaceFirst(prefix, "");
+            String columnPath = key.replaceFirst(prefix, "");
             String bloomFilterMode = config.get(key);
-            columnBloomFilterModes.put(columnAlias, bloomFilterMode);
+            columnBloomFilterConfig.put(columnPath, bloomFilterMode);
           });
-      return columnBloomFilterModes;
+      return columnBloomFilterConfig;
     }
 
     private static class Context {
@@ -361,10 +358,16 @@ public class Parquet {
       private final String compressionLevel;
       private final int rowGroupCheckMinRecordCount;
       private final int rowGroupCheckMaxRecordCount;
+      private final boolean bloomFilterEnabled;
+      private final int bloomFilterMaxBytes;
+      private final Map<String, String> columnBloomFilterEnabled;
+      private final Map<String, String> columnBloomFilterNDVs;
 
       private Context(int rowGroupSize, int pageSize, int dictionaryPageSize,
                       CompressionCodecName codec, String compressionLevel,
-                      int rowGroupCheckMinRecordCount, int rowGroupCheckMaxRecordCount) {
+                      int rowGroupCheckMinRecordCount, int rowGroupCheckMaxRecordCount,
+                      boolean bloomFilterEnabled, int bloomFilterMaxBytes,
+                      Map<String, String> columnBloomFilterEnabled, Map<String, String> columnBloomFilterNDVs) {
         this.rowGroupSize = rowGroupSize;
         this.pageSize = pageSize;
         this.dictionaryPageSize = dictionaryPageSize;
@@ -372,6 +375,10 @@ public class Parquet {
         this.compressionLevel = compressionLevel;
         this.rowGroupCheckMinRecordCount = rowGroupCheckMinRecordCount;
         this.rowGroupCheckMaxRecordCount = rowGroupCheckMaxRecordCount;
+        this.bloomFilterEnabled = bloomFilterEnabled;
+        this.bloomFilterMaxBytes = bloomFilterMaxBytes;
+        this.columnBloomFilterEnabled = columnBloomFilterEnabled;
+        this.columnBloomFilterNDVs = columnBloomFilterNDVs;
       }
 
       static Context dataContext(Map<String, String> config) {
@@ -404,8 +411,22 @@ public class Parquet {
         Preconditions.checkArgument(rowGroupCheckMaxRecordCount >= rowGroupCheckMinRecordCount,
             "Row group check maximum record count must be >= minimal record count");
 
+        boolean bloomFilterEnabled = PropertyUtil.propertyAsBoolean(config, DEFAULT_PARQUET_BLOOM_FILTER_ENABLED,
+            DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT);
+
+        int bloomFilterMaxBytes = PropertyUtil.propertyAsInt(config, PARQUET_BLOOM_FILTER_MAX_BYTES,
+            PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT);
+        Preconditions.checkArgument(bloomFilterMaxBytes > 0, "bloom Filter Max Bytes must be > 0");
+
+        Map<String, String> columnBloomFilterEnabled =
+            getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX, config);
+
+        Map<String, String> columnBloomFilterNDVs =
+            getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX, config);
+
         return new Context(rowGroupSize, pageSize, dictionaryPageSize, codec, compressionLevel,
-            rowGroupCheckMinRecordCount, rowGroupCheckMaxRecordCount);
+            rowGroupCheckMinRecordCount, rowGroupCheckMaxRecordCount, bloomFilterEnabled, bloomFilterMaxBytes,
+            columnBloomFilterEnabled, columnBloomFilterNDVs);
       }
 
       static Context deleteContext(Map<String, String> config) {
@@ -441,8 +462,22 @@ public class Parquet {
         Preconditions.checkArgument(rowGroupCheckMaxRecordCount >= rowGroupCheckMinRecordCount,
             "Row group check maximum record count must be >= minimal record count");
 
+        boolean bloomFilterEnabled = PropertyUtil.propertyAsBoolean(config, DEFAULT_PARQUET_BLOOM_FILTER_ENABLED,
+            DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT);
+
+        int bloomFilterMaxBytes = PropertyUtil.propertyAsInt(config, PARQUET_BLOOM_FILTER_MAX_BYTES,
+            PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT);
+        Preconditions.checkArgument(bloomFilterMaxBytes > 0, "bloom Filter Max Bytes must be > 0");
+
+        Map<String, String> columnBloomFilterEnabled =
+            getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX, config);
+
+        Map<String, String> columnBloomFilterNDVs =
+            getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX, config);
+
         return new Context(rowGroupSize, pageSize, dictionaryPageSize, codec, compressionLevel,
-            rowGroupCheckMinRecordCount, rowGroupCheckMaxRecordCount);
+            rowGroupCheckMinRecordCount, rowGroupCheckMaxRecordCount, bloomFilterEnabled, bloomFilterMaxBytes,
+            columnBloomFilterEnabled, columnBloomFilterNDVs);
       }
 
       private static CompressionCodecName toCodec(String codecAsString) {
@@ -479,6 +514,22 @@ public class Parquet {
 
       int rowGroupCheckMaxRecordCount() {
         return rowGroupCheckMaxRecordCount;
+      }
+
+      boolean bloomFilterEnabled() {
+        return bloomFilterEnabled;
+      }
+
+      int bloomFilterMaxBytes() {
+        return bloomFilterMaxBytes;
+      }
+
+      Map<String, String> columnBloomFilterEnabled() {
+        return columnBloomFilterEnabled;
+      }
+
+      Map<String, String> columnBloomFilterNDVs() {
+        return columnBloomFilterNDVs;
       }
     }
   }
--- a/parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
+++ b/parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
@@ -78,8 +78,6 @@ import org.apache.parquet.hadoop.api.WriteSupport;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.apache.parquet.schema.MessageType;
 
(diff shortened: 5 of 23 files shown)


jshmchenxi:
@huaxingao Thanks for continuing this work!

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderWithBloomFilter.java
```diff
@@ -0,0 +1,248 @@
...
+
+    table.updateProperties()
+        .set(TableProperties.PARQUET_ROW_GROUP_SIZE_BYTES, "100")  // to have multiple row groups
+        .commit();
+    if (vectorized) {
+      table.updateProperties()
+          .set(TableProperties.PARQUET_VECTORIZATION_ENABLED, "true")
+          .set(TableProperties.PARQUET_BATCH_SIZE, "4")
+          .commit();
+    }
```
hililiwei, line 206: I don't quite understand this. Can we inject these properties directly when the table is created instead of changing it after it is created?

```
table = catalog.createTable(TableIdentifier.of("default", name), schema);
```

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
```diff
@@ -0,0 +1,950 @@
...
+    boolean shouldRead = new ParquetBloomRowGroupFilter(SCHEMA, equal("no_stats", "a"))
+        .shouldRead(parquetSchema, rowGroupMetadata, bloomStore);
+    Assert.assertFalse("Should skip: stats are missing but bloom filter is present", shouldRead);
+  }
+
+  @Test
+  public void testNot() {
+    // this test case must use a real predicate, not alwaysTrue(), otherwise binding will simplify it out
+    for (int i = INT_MIN_VALUE - 20; i < INT_MAX_VALUE + 20; i++) {
+      boolean shouldRead = new ParquetBloomRowGroupFilter(SCHEMA,   not(equal("id", i)))
```
hililiwei, line 410: format

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
```diff
@@ -0,0 +1,950 @@
...
+  @Test
+  public void testMissingStats() {
+    boolean shouldRead = new ParquetBloomRowGroupFilter(SCHEMA, equal("no_stats", "a"))
+        .shouldRead(parquetSchema, rowGroupMetadata, bloomStore);
+    Assert.assertFalse("Should skip: stats are missing but bloom filter is present", shouldRead);
+  }
+
+  @Test
+  public void testNot() {
+    // this test case must use a real predicate, not alwaysTrue(), otherwise binding will simplify it out
```
hililiwei, line 408: would you mind explaining it?

api/src/main/java/org/apache/iceberg/data/Record.java
```diff
@@ -58,4 +58,22 @@ default Record copy(String field1, Object value1, String field2, Object value2,
     return copy(overwriteValues);
   }
 
+  default Record copy(String field1, Object value1, String field2, Object value2, String field3, Object value3,
+                      String field4, Object value4, String field5, Object value5, String field6, Object value6,
+                      String field7, Object value7, String field8, Object value8, String field9, Object value9,
+                      String field10, Object value10, String field11, Object value11) {
+    Map<String, Object> overwriteValues = Maps.newHashMapWithExpectedSize(9);
```
hililiwei, line 65: It's more intuitive to just use 12. Although it's actually 12, even with 9.

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -269,19 +282,34 @@ public <D> FileAppender<D> build() throws IOException {
...
-        ParquetProperties parquetProperties = ParquetProperties.builder()
+        ParquetProperties.Builder propsBuilder = ParquetProperties.builder()
             .withWriterVersion(writerVersion)
             .withPageSize(pageSize)
             .withDictionaryPageSize(dictionaryPageSize)
             .withMinRowCountForPageSizeCheck(rowGroupCheckMinRecordCount)
             .withMaxRowCountForPageSizeCheck(rowGroupCheckMaxRecordCount)
-            .build();
+            .withMaxBloomFilterBytes(bloomFilterMaxBytes)
+            .withBloomFilterEnabled(bloomFilterEnabled);
```
hililiwei, line 292: When `bloomFilterEnabled` is true, all fields are turned on unless the fields are specified to be false in `columnBloomFilterModes`, right?

hililiwei (changes requested):
Great work. Left some comment.

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -167,6 +167,17 @@ private TableProperties() {
...
+  public static final boolean DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT = false;
+
+  public static final String PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX = "write.parquet.bloom-filter-enabled.column.";
+  public static final String PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX =
+      "write.parquet.bloom-filter-expected-ndv.column.";
+
+  public static final String PARQUET_BLOOM_FILTER_MAX_BYTES = "write.parquet.bloom-filter-max-bytes";
+  public static final int PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT = 1024 * 1024;
+
+
```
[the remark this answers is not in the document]
huaxingao: Removed. Thanks

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -239,6 +245,13 @@ public <D> FileAppender<D> build() throws IOException {
       CompressionCodecName codec = context.codec();
       int rowGroupCheckMinRecordCount = context.rowGroupCheckMinRecordCount();
       int rowGroupCheckMaxRecordCount = context.rowGroupCheckMaxRecordCount();
+      boolean bloomFilterEnabled = PropertyUtil.propertyAsBoolean(config, DEFAULT_PARQUET_BLOOM_FILTER_ENABLED,
+          DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT);
+      int bloomFilterMaxBytes = PropertyUtil.propertyAsInt(config, PARQUET_BLOOM_FILTER_MAX_BYTES,
+          PARQUET_BLOOM_FILTER_MAX_BYTES_DEFAULT);
+      Map<String, String> columnBloomFilterModes = getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX);
+      Map<String, String> columnBloomFilterNDVs =
+          getBloomColumnConfigMap(PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX);
```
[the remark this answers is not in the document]
huaxingao: It's cleaner to get it from the Context. Changed. Thanks

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -292,11 +320,39 @@ public <D> FileAppender<D> build() throws IOException {
             .withRowGroupSize(rowGroupSize)
             .withPageSize(pageSize)
             .withDictionaryPageSize(dictionaryPageSize)
-            .build(),
+            // TODO: add .withMaxBloomFilterBytes(bloomFilterMaxBytes) once ParquetWriter.Builder supports it
+            .withBloomFilterEnabled(bloomFilterEnabled);
+
+        for (Map.Entry<String, String> entry : columnBloomFilterModes.entrySet()) {
+          String col = entry.getKey();
+          String value = entry.getValue();
```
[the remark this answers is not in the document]
huaxingao: Changed

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -292,11 +320,39 @@ public <D> FileAppender<D> build() throws IOException {
...
+          parquetWriteBuilder.withBloomFilterEnabled(col, Boolean.valueOf(value));
+        }
+
+        for (Map.Entry<String, String> entry : columnBloomFilterNDVs.entrySet()) {
+          String col = entry.getKey();
+          String value = entry.getValue();
```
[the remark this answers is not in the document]
huaxingao: Changed

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -292,11 +320,39 @@ public <D> FileAppender<D> build() throws IOException {
...
       }
     }
 
+    private Map<String, String> getBloomColumnConfigMap(String prefix) {
+      Map<String, String> columnBloomFilterModes = Maps.newHashMap();
+      config.keySet().stream()
+          .filter(key -> key.startsWith(prefix))
+          .forEach(key -> {
+            String columnAlias = key.replaceFirst(prefix, "");
+            String bloomFilterMode = config.get(key);
```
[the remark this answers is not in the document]
huaxingao: Changed key to `columnPath`. I left `bloomFilterMode` as is because it could either be `columnBloomFilterEnabled` or `columnBloomFilterNDVs`.

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
```diff
@@ -0,0 +1,950 @@
...
+    TOO_LONG_FOR_STATS = sb.toString();
+  }
+
+  private static final int INT_MIN_VALUE = 30;
+  private static final int INT_MAX_VALUE = 79;
+  private static final int INT_VALUE_COUNT = INT_MAX_VALUE - INT_MIN_VALUE + 1;
+  private static final long LONG_BASE = 100L;
+  private static final double DOUBLE_BASE = 1000D;
+  private static final float FLOAT_BASE = 10000F;
+  private static final String BINARY_PREFIX = "BINARY测试_";
```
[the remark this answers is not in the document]
huaxingao: Actually I prefer to test both English and Chinese (for double bytes characters)

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderWithBloomFilter.java
huaxingao: These properties can be set either at table creation or updated later. I don't think it matters. The reason I picked the second choice is because when I wrote the test, I happened to use `TestSparkReaderDeletes` as my template and followed the style there.

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
huaxingao: Fixed. Thanks

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
huaxingao: `TestBloomRowGroupFilter` follows the test style of `TestMetricsRowGroupFilter` and `TestDictionaryRowGroupFilter` and this comment is actually copied. I think what it means is that a real predicate e.g. equal, lessThanOrEqual, etc. need to be used. If you use `alwaysTrue`, it will be simplified out and then the visitEvaluator path won't be tested.

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
huaxingao: That's right

huaxingao:
cc @aokolnychyi @RussellSpitzer @flyrain @szehon-ho 
This PR is ready for review. Thank you very much in advance!

huaxingao:
also cc @kbendick @chenjunjiedada @rdblue

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -167,6 +167,16 @@ private TableProperties() {
       "write.delete.parquet.row-group-check-max-record-count";
   public static final int PARQUET_ROW_GROUP_CHECK_MAX_RECORD_COUNT_DEFAULT = 10000;
 
+  public static final String DEFAULT_PARQUET_BLOOM_FILTER_ENABLED = "write.parquet.bloom-filter-enabled.default";
+  public static final boolean DEFAULT_PARQUET_BLOOM_FILTER_ENABLED_DEFAULT = false;
+
+  public static final String PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX = "write.parquet.bloom-filter-enabled.column.";
+  public static final String PARQUET_BLOOM_FILTER_COLUMN_EXPECTED_NDV_PREFIX =
```
chenjunjiedada, line 174: It would be better to document that the NDV is specific for a parquet file.

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -167,6 +167,16 @@ private TableProperties() {
       "write.delete.parquet.row-group-check-max-record-count";
   public static final int PARQUET_ROW_GROUP_CHECK_MAX_RECORD_COUNT_DEFAULT = 10000;
 
+  public static final String DEFAULT_PARQUET_BLOOM_FILTER_ENABLED = "write.parquet.bloom-filter-enabled.default";
```
stevenzwu, line 170: does it ever make sense to enable bloom filter for all columns? should we only allow bloom filter for explicitly specified columns?

huaxingao: It is not a common usage to enable bloom filter for all columns, but it's legal. This is consistent with the parquet-mr bloom filter implementations.

api/src/main/java/org/apache/iceberg/data/Record.java
```diff
@@ -58,4 +58,22 @@ default Record copy(String field1, Object value1, String field2, Object value2,
...
+    overwriteValues.put(field3, value3);
+    overwriteValues.put(field4, value4);
+    overwriteValues.put(field5, value5);
+    overwriteValues.put(field6, value6);
+    overwriteValues.put(field7, value7);
+    overwriteValues.put(field8, value8);
+    overwriteValues.put(field9, value9);
+    overwriteValues.put(field10, value10);
+    overwriteValues.put(field11, value11);
+    return copy(overwriteValues);
```
rdblue, line 77: Do we need to change the public API? And why not use a map for large changes like this?

core/src/main/java/org/apache/iceberg/TableProperties.java
rdblue: I agree with Steven. I think it only makes sense to enable bloom filters for some columns.

kbendick: +1. While it's consistent with the parquet-mr bloom filter implementaiton, we need to think of user experience first and foremost.

It doesn't make sense to enable bloom filters for _a lot_ of columns. And many users don't do any tuning of their metadata / statistics.

I think it's in-line with other things we do to make the users experience better, like turning off column level statistics after a certain number of columns. We can point it out in the docs under a big `!!!NOTE` (that's highlighted) that bloom filter is only used when turned on.

It's really an advanced thing to use at all imo.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderWithBloomFilter.java
kbendick: +1 to following existing style

docs/tables/configuration.md
```diff
@@ -50,6 +50,10 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.parquet.dict-size-bytes      | 2097152 (2 MB)     | Parquet dictionary page size                       |
 | write.parquet.compression-codec    | gzip               | Parquet compression codec: zstd, brotli, lz4, gzip, snappy, uncompressed |
 | write.parquet.compression-level    | null               | Parquet compression level                          |
+| write.parquet.bloom-filter-enabled.default | false | Whether to enable writing bloom filter for all columns |
+| write.parquet.bloom-filter-enabled.column.col1 | (not set) | Whether to enable writing bloom filter for column 'col1' to allow per-column configuration; This property overrides `bloom-filter-enabled.default` for the specified column; For example, setting both `write.parquet.bloom-filter-enabled.default=true` and `write.parquet.bloom-filter-enabled.column.some_col=false` will enable bloom filter for all columns except `some_col` |
```
rdblue, line 54: I think this can be a lot shorter. "Enables writing a bloom filter for the column."

docs/tables/configuration.md
```diff
@@ -50,6 +50,10 @@ Iceberg tables support table properties to configure table behavior, like the de
...
+| write.parquet.bloom-filter-expected-ndv.column.col1 | (not set) | The expected number of distinct values in a column, it is used to compute the optimal size of the bloom filter; Note that the NDV is specific for a parquet file. If this property is not set, the bloom filter will use the maximum size set in `bloom-filter-max-bytes`; If this property is set for a column, then no need to enable the bloom filter with `write.parquet.bloom-filter-enabled` property |
```
rdblue, line 55: Shouldn't this come from a table sketch instead of manual configuration? Since we're adding NDV estimation, I think we probably want to avoid setting this.

Also, if this eliminates the need to set `enabled` then we should remove the `enabled` property.

docs/tables/configuration.md
```diff
@@ -50,6 +50,10 @@ Iceberg tables support table properties to configure table behavior, like the de
...
+| write.parquet.bloom-filter-max-bytes | 1048576 (1 MB) | The maximum number of bytes for a bloom filter bitset |
```
rdblue, line 56: What is the behavior of this? If the NDV requires a size that is too large, does it skip writing the bloom filter?

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -292,11 +317,39 @@ public <D> FileAppender<D> build() throws IOException {
             .withRowGroupSize(rowGroupSize)
             .withPageSize(pageSize)
             .withDictionaryPageSize(dictionaryPageSize)
-            .build(),
+            // TODO: add .withMaxBloomFilterBytes(bloomFilterMaxBytes) once ParquetWriter.Builder supports it
```
rdblue, line 320: If this isn't supported, then we should remove the option to begin with, right?

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -292,11 +317,39 @@ public <D> FileAppender<D> build() throws IOException {
...
+          parquetWriteBuilder.withBloomFilterEnabled(colPath, Boolean.valueOf(bloomEnabled));
+        }
+
+        for (Map.Entry<String, String> entry : columnBloomFilterNDVs.entrySet()) {
+          String colPath = entry.getKey();
+          String numDistinctValue = entry.getValue();
+          parquetWriteBuilder.withBloomFilterNDV(colPath, Long.valueOf(numDistinctValue));
+        }
+
+        return new ParquetWriteAdapter<>(
```
rdblue, line 335: @aokolnychyi, what do you think about removing the old `ParquetWriteAdapter` code? I don't think that anyone uses it anymore.

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -292,11 +317,39 @@ public <D> FileAppender<D> build() throws IOException {
...
+            parquetWriteBuilder.build(),
             metricsConfig);
       }
     }
 
+    private static Map<String, String> getBloomColumnConfigMap(String prefix, Map<String, String> config) {
```
rdblue, line 341: Style: method name with `get`.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkRowLevelOperationsTestBase.java
```diff
@@ -112,7 +109,17 @@ public static Object[][] parameters() {
...
+            false
+        },
+        { "testhadoop", SparkCatalog.class.getName(),
+            ImmutableMap.of(
+                "type", "hadoop"
+            ),
+            "parquet",
+            RANDOM.nextBoolean(),
+            WRITE_DISTRIBUTION_MODE_HASH,
+            true
```
rdblue, line 122: Why not add bloom filter testing to one of the existing cases? I don't see a reason why this can't be tested along with other cases. And Spark tests already take a long time. We should avoid adding more cases.

parquet/src/main/java/org/apache/iceberg/parquet/ReadConf.java
```diff
@@ -106,7 +108,8 @@
       startRowPositions[i] = offsetToStartPos == null ? 0 : offsetToStartPos.get(rowGroup.getStartingPos());
       boolean shouldRead = filter == null || (
           statsFilter.shouldRead(typeWithIds, rowGroup) &&
-              dictFilter.shouldRead(typeWithIds, rowGroup, reader.getDictionaryReader(rowGroup)));
+              dictFilter.shouldRead(typeWithIds, rowGroup, reader.getDictionaryReader(rowGroup)) &&
+                  bloomFilter.shouldRead(typeWithIds, rowGroup, reader.getBloomFilterDataReader(rowGroup)));
```
rdblue, line 112: Nit: this indentation looks suspicious to me. The last line was also after `&&` but this one has a different indentation level.

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
```diff
@@ -0,0 +1,950 @@
...
+
+    @Override
+    public BloomFilter readBloomFilter(ColumnChunkMetaData meta) {
+      return null;
+    }
+  }
+
+  @Test
+  public void testIntegerIn() {
+    for (int i = 0; i < INT_VALUE_COUNT; i++) {
```
rdblue, line 809: I don't think this needs to be tested in loops like this. I think the case you want to test is that `in` behaves like `or`. So you should have a case where all values are present, no values are present, and where only one value is present.

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
```diff
@@ -0,0 +1,950 @@
...
+    Assert.assertFalse("Should not read: in on no nulls column (empty string is not within the set)", shouldRead);
+
+    shouldRead = new ParquetBloomRowGroupFilter(SCHEMA, in("no_nulls", "aaa", ""))
+        .shouldRead(parquetSchema, rowGroupMetadata, bloomStore);
+    Assert.assertTrue("Should read: in on no nulls column (empty string is within the set)", shouldRead);
+  }
+
+  @Test
+  public void testIntegerNotIn() {
+    for (int i = 0; i < INT_VALUE_COUNT; i++) {
```
rdblue, line 871: Here as well, I think you should test the same cases as `in`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+    @Override
+    public Boolean alwaysFalse() {
+      return ROWS_CANNOT_MATCH; // all rows fail
+    }
+
+    @Override
+    public Boolean not(Boolean result) {
+      // not() should be rewritten by RewriteNot
+      // bloom filter is based on hash and cannot eliminate based on not
+      return ROWS_MIGHT_MATCH;
```
rdblue, line 140: Since this calls `rewriteNot` on the filter, you may want to throw an `UnsupportedOperationException`. This is okay as well, but I don't think this will ever be hit.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+      this.fieldsWithNoBloom = Sets.newHashSet();
+      this.cols = Maps.newHashMap();
+      this.columnMetaMap = Maps.newHashMap();
+      this.types = Maps.newHashMap();
+
+      for (ColumnDescriptor desc : fileSchema.getColumns()) {
+        PrimitiveType colType = fileSchema.getType(desc.getPath()).asPrimitiveType();
+        if (colType.getId() != null) {
+          int id = colType.getId().intValue();
+          cols.put(id, desc);
```
rdblue, line 101: It looks like this is only used in `eq` and `in` to check whether the column is in the file or not. But isn't that decision the same as if the column has no bloom filter?

If `bloomCache` doesn't contain the ID, it doesn't matter if it was because the column doesn't exist or if the column exists but has no bloom filter. Either way, this should return `ROWS_MIGHT_MATCH`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+          return ROWS_MIGHT_MATCH;
+        }
+        BloomFilter bloom = getBloomById(id);
+        Type type = types.get(id);
+        T value = lit.value();
+        return shouldRead(col, value, bloom, type);
+      }
+    }
+
+    private BloomFilter getBloomById(int id) {
```
rdblue, line 219: Style: use of `get` in a method name.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+          if (!noBloom) {
+            hasBloom = true;
+          } else {
+            fieldsWithNoBloom.add(id);
+          }
+          columnMetaMap.put(id, meta);
+          types.put(id, icebergType);
+        }
+      }
+      if (!hasBloom) {
```
rdblue, line 120: What about when the filter's column set doesn't overlap with any bloom filter columns? I think that would be a stronger check to exit early.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+      }
+
+      for (ColumnChunkMetaData meta : rowGroup.getColumns()) {
+        PrimitiveType colType = fileSchema.getType(meta.getPath().toArray()).asPrimitiveType();
+        if (colType.getId() != null) {
+          int id = colType.getId().intValue();
+          Type icebergType = schema.findType(id);
+          boolean noBloom = ParquetUtil.hasNoBloomFilterPages(meta);
+          if (!noBloom) {
+            hasBloom = true;
```
rdblue, line 112: I think `hasBloom` can be removed, in favor of using a `fieldsWithBloomFilter`. Then you can update the check to do a set comparison:

```java
  if (Sets.intersection(fieldsWithBloomFilter, Binder.boundReferences(filter)).isEmpty()) {
    return ROWS_MIGHT_MATCH;
  }
```

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+      long hashValue = 0;
+      switch (type.typeId()) {
+        case INTEGER:
+        case LONG:
+        case FLOAT:
+        case DOUBLE:
+        case DATE:
+        case TIME:
+        case TIMESTAMP:
+          hashValue = bloom.hash(value);
```
rdblue, line 248: Why does this not cast `value`? As long as the type is checked in this method, I think it makes sense to call the right `hash` method, instead of calling a generic one for `Object`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+          return bloom.findHash(hashValue) ? ROWS_MIGHT_MATCH : ROWS_CANNOT_MATCH;
+        case STRING:
+          hashValue = bloom.hash(Binary.fromCharSequence((CharSequence) value));
+          return bloom.findHash(hashValue) ? ROWS_MIGHT_MATCH : ROWS_CANNOT_MATCH;
+        case BINARY:
+        case FIXED:
+          hashValue = bloom.hash(Binary.fromReusedByteBuffer((ByteBuffer) value));
```
rdblue, line 255: This is not a reused byte buffer. The contents won't change while `hash` is using it.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+          return bloom.findHash(hashValue) ? ROWS_MIGHT_MATCH : ROWS_CANNOT_MATCH;
+        case DECIMAL:
+          BigDecimal decimalValue = (BigDecimal) value;
+          switch (col.getPrimitiveType().getPrimitiveTypeName()) {
+            case INT32:
+              hashValue = bloom.hash(decimalValue.unscaledValue().intValue());
+              return bloom.findHash(hashValue) ? ROWS_MIGHT_MATCH : ROWS_CANNOT_MATCH;
```
rdblue, line 262: While I appreciate the use of `ROWS_MIGHT_MATCH` and `ROWS_CANNOT_MATCH`, I think this is no different than returning the result of `findHash` directly, right?

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+            case INT64:
+              hashValue = bloom.hash(decimalValue.unscaledValue().longValue());
+              return bloom.findHash(hashValue) ? ROWS_MIGHT_MATCH : ROWS_CANNOT_MATCH;
+            case FIXED_LEN_BYTE_ARRAY:
+            case BINARY:
+              DecimalMetadata metadata = col.getPrimitiveType().getDecimalMetadata();
+              int scale = metadata.getScale();
+              int precision = metadata.getPrecision();
+              byte[] requiredBytes = new byte[TypeUtil.decimalRequiredBytes(precision)];
+              byte[] binary = DecimalUtil.toReusedFixLengthBytes(precision, scale, (BigDecimal) value, requiredBytes);
```
rdblue, line 272: The underlying bytes are not reused.

Also, why does this need to copy? If `bloom` works with a Binary you can create it directly from the byte buffer. There should be no need to limit the number of bytes used to `TypeUtil.decimalRequiredBytes`, although if you wanted to be really careful you could check that the expected number of bytes in the buffer matches and return `ROWS_MIGHT_MATCH` if it doesn't.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+          throw new IllegalStateException("Failed to read required bloom filter for id: " + id);
+        } else {
+          bloomCache.put(id, bloomFilter);
+        }
+
+        return bloomFilter;
+      }
+    }
+
+    private <T> boolean shouldRead(ColumnDescriptor col, T value, BloomFilter bloom, Type type) {
```
rdblue, line 238: Can you move this utility method to the bottom of the class, rather than between override method implementations?

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,330 @@
...
+        BloomFilter bloom = getBloomById(id);
+        Type type = types.get(id);
+        T value = lit.value();
+        return shouldRead(col, value, bloom, type);
+      }
+    }
+
+    private BloomFilter getBloomById(int id) {
+      if (bloomCache == null) {
+        bloomCache = Maps.newHashMap();
```
rdblue, line 221: Rather than doing this all the time, can you create this cache in `eval`?

rdblue:
Thanks for working on this, @huaxingao! It looks close overall.

Does it make sense to split this into separate read and write PRs? It seems like adding the evaluator is quite a lot to review independently, and we don't need the write path for its tests.

core/src/main/java/org/apache/iceberg/TableProperties.java
huaxingao: Sounds good. I have removed `DEFAULT_PARQUET_BLOOM_FILTER_ENABLED`. Now user needs to enable bloom filter for individual column using `PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX`. If the column is a complex type, user needs to enable the column inside the complex type, for example, 
`set(PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX + "struct_col.int_field", "true")`

api/src/main/java/org/apache/iceberg/data/Record.java
huaxingao: This change is unnecessary. Removed.

docs/tables/configuration.md
huaxingao: Changed. Thanks!

docs/tables/configuration.md
huaxingao: Removed this property.

docs/tables/configuration.md
huaxingao: If the NDV requires a size that is too large, parquet still writes the bloom filter using the max bytes set by this property, not using the bitset calculated by NDV.

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
huaxingao: Right. Deleted this comment.

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
huaxingao: You mean to remove `get`, right? Changed to `bloomColumnConfigMap`.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkRowLevelOperationsTestBase.java
huaxingao: I want to have some bloom filter write/read tests that involve copy on write or merge on read. I think this is the only place that test copy on write or merge on read (I maybe wrong). That's why I added a parameter here to turn on bloom filter to test.

parquet/src/main/java/org/apache/iceberg/parquet/ReadConf.java
huaxingao: I changed to  
```
          statsFilter.shouldRead(typeWithIds, rowGroup) &&
              dictFilter.shouldRead(typeWithIds, rowGroup, reader.getDictionaryReader(rowGroup)) &&
              bloomFilter.shouldRead(typeWithIds, rowGroup, reader.getBloomFilterDataReader(rowGroup)));
```
Does this look right now?

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
huaxingao: Changed. Thanks!

parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
huaxingao: Changed. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Fixed. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: You are right. I deleted `cols`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: changed to `bloomById`

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Sounds good! Changed the code as you suggested. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Casted. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Changed to `fromConstantByteBuffer`

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Right. Fixed. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: If I use `decimalValue.unscaledValue().toByteArray()` without padding to the fixed length, I can't find the hash value. 
I took a look at parquet bloom filter's write path, seems it pads the bytes to the fixed length, and then hash.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Moved. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Fixed. Thanks!

huaxingao:
Thanks @rdblue @kbendick for reviewing!
I have addressed the comments and changed the code accordingly. I changed the code using this PR because I think it might be easier for you to check if I address the comments here. I can split this PR into two afterwards.

parquet/src/main/java/org/apache/iceberg/parquet/ReadConf.java
rdblue: Yes, looks good.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
```diff
@@ -91,12 +91,21 @@ static Expression bind(StructType struct,
   }
 
   public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
+    return boundReferences(struct, exprs, caseSensitive, false);
+  }
+
+  public static Set<Integer> boundReferences(
+      StructType struct, List<Expression> exprs, boolean caseSensitive, boolean alreadyBound) {
```
rdblue, line 98: I don't think it is a good idea to add a second boolean argument to this method. It is confusing enough with just one.

How about using a different method name for this and then renaming this to a be a private internal implementation?

docs/tables/configuration.md
rdblue: I guess there probably isn't much we can do about this, although that behavior makes no sense to me. Is it possible to set the expected false positive probability anywhere? Or is that hard-coded in the Parquet library?

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -291,12 +304,32 @@ public <D> FileAppender<D> build() throws IOException {
...
             metricsConfig);
       }
     }
 
+    private static Map<String, String> bloomColumnConfigMap(String prefix, Map<String, String> config) {
+      Map<String, String> columnBloomFilterConfig = Maps.newHashMap();
+      config.keySet().stream()
+          .filter(key -> key.startsWith(prefix))
+          .forEach(key -> {
+            String columnPath = key.replaceFirst(prefix, "");
```
rdblue, line 326: Since this uses column name in the config, is there any logic to update these configs when columns are renamed?

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,316 @@
...
+          types.put(id, icebergType);
+        }
+      }
+
+      Set<Integer> filterRefs = Binder.boundReferences(schema.asStruct(), ImmutableList.of(expr), caseSensitive, true);
+      // If the filter's column set doesn't overlap with any bloom filter columns, exit early with ROWS_MIGHT_MATCH
+      if (filterRefs.size() > 0 && Sets.intersection(fieldsWithBloomFilter, filterRefs).isEmpty()) {
+        return ROWS_MIGHT_MATCH;
+      }
+      return ExpressionVisitors.visitEvaluator(expr, this);
```
rdblue, line 118: Style: there should be a newline between a control flow block and the following statement.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,316 @@
...
+      return ROWS_MIGHT_MATCH;
+    }
+
+    @Override
+    public <T> Boolean in(BoundReference<T> ref, Set<T> literalSet) {
+      int id = ref.fieldId();
+      if (!fieldsWithBloomFilter.contains(id)) { // no bloom filter
+        return ROWS_MIGHT_MATCH;
+      }
+      BloomFilter bloom = bloomById(id);
```
rdblue, line 220: Style: control flow blocks should be separated from the following block or statement with an empty newline.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,316 @@
...
+      return ROWS_MIGHT_MATCH;
+    }
+
+    @Override
+    public <T> Boolean notStartsWith(BoundReference<T> ref, Literal<T> lit) {
+      // bloom filter is based on hash and cannot eliminate based on startsWith
+      return ROWS_MIGHT_MATCH;
+    }
+
+    private BloomFilter bloomById(int id) {
```
rdblue, line 248: I think `loadBloomFilter` is probably a better name.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,316 @@
...
+          bloomCache.put(id, bloomFilter);
+        }
+
+        return bloomFilter;
+      }
+    }
+
+    private <T> boolean shouldRead(PrimitiveType primitiveType, T value, BloomFilter bloom, Type type) {
+      long hashValue = 0;
+      switch (type.typeId()) {
```
rdblue, line 266: I think that this needs to use the Parquet type from the file, not the Iceberg type.

The Iceberg type comes from the read schema, which could be `long` when the file's type is `int` because of type promotion rules. We need to make sure that the hash function used matches the file's type because that was what produced the bloom filter. Hopefully, the same hash function is used for `long` and `int` in Parquet, but I'm not sure that is the case so the safest thing is to convert `value` to the Parquet type.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,316 @@
...
+          hashValue = bloom.hash(Binary.fromCharSequence((CharSequence) value));
+          return bloom.findHash(hashValue);
+        case BINARY:
+        case FIXED:
+          hashValue = bloom.hash(Binary.fromConstantByteBuffer((ByteBuffer) value));
+          return bloom.findHash(hashValue);
+        case DECIMAL:
+          BigDecimal decimalValue = (BigDecimal) value;
+          switch (primitiveType.getPrimitiveTypeName()) {
+            case INT32:
```
rdblue, line 292: I think this is correct.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
rdblue: Okay, I see. I think you're right.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,316 @@
...
+              DecimalMetadata metadata = primitiveType.getDecimalMetadata();
+              int scale = metadata.getScale();
+              int precision = metadata.getPrecision();
+              byte[] requiredBytes = new byte[TypeUtil.decimalRequiredBytes(precision)];
+              byte[] binary = DecimalUtil.toReusedFixLengthBytes(precision, scale, (BigDecimal) value, requiredBytes);
+              hashValue = bloom.hash(Binary.fromConstantByteArray(binary));
+              return bloom.findHash(hashValue);
+          }
+        case UUID:
+          hashValue = bloom.hash(Binary.fromReusedByteArray(UUIDUtil.convert((UUID) value)));
```
rdblue, line 309: I don't think this byte array is reused.

docs/tables/configuration.md
huaxingao: There isn't a property to set fpp in Parquet.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkRowLevelOperationsTestBase.java
huaxingao: I see. Thanks for the clarification. Instead of executing the entire suite, is it OK to add a new (but small) test suite to test bloom filter for copy on write or merge on read? I didn't find any existing copy on write/merge on read tests with filters.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
huaxingao: I changed this method name to `exprReferences` (please let me know if you have a better name), but I still need to keep this as `public` because I need to access this method from `ParquetBloomRowGroupFilter`, which is in a different package.

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
huaxingao: Good question. I don't have a logic to update the configs when columns are renamed. I think we are OK, though. At write path, I use these configs to write bloom filters at file creation time. I don't use these configs any more for read. At read path, the bloom filters are loaded using id instead of column name. If the columns are renamed after the bloom filters have been written, as long as the id are still the same, the bloom filters should be able to loaded OK.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Changed. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Changed. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Changed to `loadBloomFilter`. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: I changed the code to `switch (primitiveType.getPrimitiveTypeName())`

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Changed to `fromConstantByteArray`. Thanks!

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
rdblue: We're okay for adding read support, but we should consider how to configure write support then.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
rdblue: Maybe just `references` since it doesn't bind?

docs/tables/configuration.md
rdblue: What fpp is used by Parquet?

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,325 @@
...
+            case DECIMAL:
+              BigDecimal decimalValue = (BigDecimal) value;
+              hashValue = bloom.hash(decimalValue.unscaledValue().intValue());
+              return bloom.findHash(hashValue);
+            case INTEGER:
+            case DATE:
+              hashValue = bloom.hash(((Number) value).intValue());
+              return bloom.findHash(hashValue);
+          }
+        case INT64:
```
rdblue, line 280: Looks like this will fall through if the type ID is not handled? I think you need to add a `break;`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    private <T> boolean shouldRead(PrimitiveType primitiveType, T value, BloomFilter bloom, Type type) {
+      long hashValue = 0;
+      switch (primitiveType.getPrimitiveTypeName()) {
+        case INT32:
+          switch (type.typeId()) {
+            case DECIMAL:
+              BigDecimal decimalValue = (BigDecimal) value;
+              hashValue = bloom.hash(decimalValue.unscaledValue().intValue());
```
rdblue, line 273: This is what Parquet would do to hash a decimal stored as an int?

rdblue (commented):
@huaxingao, other than cleaning up `shouldRead` to add avoid falling through for unexpected Iceberg types, I think the read side of this is about ready to go. Do you want to open a separate PR for that?

docs/tables/configuration.md
huaxingao: Parquet uses 0.01 for fpp.
I chatted offline with @chenjunjiedada, we can probably add a config for fpp in parquet first, and then in iceberg.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
huaxingao: Sounds good. Will change.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Will fix.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Yes

huaxingao:
@rdblue I have a question on how to test the read path. If we separate the write and read path, I don't have a good way to test the read path. Shall we do the write path first? For write path, we can test if the bloom filter is written OK by calling the Parquet bloom filter APIs. Or I just submit the read path PR without a test, and add the test after the write path is in?

rdblue:
I'm going to close this because it is broken into separate read and write side PRs.

jia-zhengwei:
@huaxingao  
How to create a bloom filter index for iceberg with spark?
I know I need to set (write.parquet.bloom-filter-enabled.column.col1='true') when create table, do I need other operations when insert data by pyspark? 
How to make sure the bloom filter is  OK?  Will they have some new index file generated?

PR #4832, opened 2022-05-21
Core: Remove Authorization header handling from HTTPClient

rdblue:
This removes the `Authorization` header code from `HTTPClient` because the REST catalog will handle its own headers independent of the client.

This is part #4830.


danielcweeks (approved):

rdblue:
Thanks for the review, @danielcweeks!

PR #4841, opened 2022-05-23
Parquet: Fix Iceberg's parquet reader returning binary data instead of string data

lcspinter:
String columns from Parquet files written by writers that don't use logical type annotations are converted into binary data. This can be reproduced with older versions of Hive or Impala.

### DIFF
--- a/parquet/src/main/java/org/apache/iceberg/data/parquet/BaseParquetReaders.java
+++ b/parquet/src/main/java/org/apache/iceberg/data/parquet/BaseParquetReaders.java
@@ -291,7 +291,11 @@ public abstract class BaseParquetReaders<T> {
         case FIXED_LEN_BYTE_ARRAY:
           return new FixedReader(desc);
         case BINARY:
-          return new ParquetValueReaders.BytesReader(desc);
+          if (expected != null && expected.typeId() == org.apache.iceberg.types.Type.TypeID.STRING) {
+            return new ParquetValueReaders.StringReader(desc);
+          } else {
+            return new ParquetValueReaders.BytesReader(desc);
+          }
         case INT32:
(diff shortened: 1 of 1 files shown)


lcspinter:
cc: @pvary @szlta

pvary (approved):

szlta (approved):
Looks good to me, +1 non-binding

ConeyLiu:
LGTM too.

kbendick:
If you wanted to add a test for this, we might be able to generate data using Spark with some combination of `spark.sql.parquet.binaryAsString` and/or `spark.sql.parquet.writeLegacyFormat`.

lcspinter:
@kbendick I might be able to generate some binary data files with an older version of Impala and write some tests on reading those. Though, I'm not sure if it's a good practice to check in a data set generated by a 3rd party tool, rather than generating it in place. I would like to hear your opinion. Thanks

kbendick:
Hey @lcspinter. Sorry for the delayed response.

I agree that checking in files is generally something to avoid, though if need absolutely be (especially for interfacing with systems that we don't support directly within this repository) it might be ok.

If you can generate a file with Impala, we could potentially compare the output to generating code using Spark with the configuration values `spark.sql.parquet.binaryAsString` or something else. I can take a look this weekend.

kbendick:
All that said, I do wonder if we need to account for certain spark configs etc. Respecting the type as given in the schema is likely the best way to go though =)

rdblue (approved):

rdblue:
Thanks, @lcspinter!

lcspinter:
Thanks, @kbendick @pvary  @rdblue @szlta for the review!

PR #4842, opened 2022-05-23
[CORE] Support file filtering based on schema

ConeyLiu:
This patch adds the support of file filtering based on the schema. This aims to reduce the file scan time when we have done a schema evaluation. For the following example:
```
1. create a table with schema <id: long, name: string> and partition on id
2. add new files
3. update the schema and add a new column: address. Now the schema is <id: long, name: string, address: string>
4. add new files
// Before this patch, we have to read those files added at 2 and do the filter after reading those files.
6. scan with filter start_with('address', 'some_value') and id > 10;
```
This adds the `SchemaId` to both `DataFile` and `ManifestFile`, we will filter the `ManifestFile` first and then on `DataFile` when needed.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseFile.java
+++ b/core/src/main/java/org/apache/iceberg/BaseFile.java
@@ -54,6 +54,7 @@ abstract class BaseFile<F>
   private Types.StructType partitionType;
 
   private Long fileOrdinal = null;
+  private int schemaId = -1;
   private int partitionSpecId = -1;
   private FileContent content = FileContent.DATA;
   private String filePath = null;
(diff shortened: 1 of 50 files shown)


core/src/main/java/org/apache/iceberg/BaseFile.java
```diff
@@ -277,6 +280,9 @@ public void put(int i, Object value) {
         this.sortOrderId = (Integer) value;
         return;
       case 17:
+        this.schemaId = (int) value;
```
ConeyLiu, line 283: I updated the order because we add the `ROW_POSITION` after the schema fields of DataFile here: https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/BaseFile.java#L99. Some experts could give some advice for this update.

rdblue:
@ConeyLiu, can you break this down into smaller commits? I think it's reasonable to use the schema information to know whether a filter will succeed for fail, but there are a lot of changes that don't need to be combined into a single PR here. Thanks!

ConeyLiu:
@rdblue, thanks for the suggestion. The first part is here: #4898

PR #4844, opened 2022-05-23
Add Poetry build and release instructions

Fokko:
- Migrate from the setup.cfg to pyproject.toml
- Add poetry.lock for pinning all the dependencies, to make sure that we exactly know which libraries are installed
- Add detailed instruction on how to run a release


rdblue:
This looks good to me, but I'd love to hear what other python people think about moving to poetry.

python/README.md
```diff
@@ -30,26 +30,27 @@ cd iceberg/python
 pip install -e .
 ```
 
-## Testing
+Development is made easy using [Poetry](https://python-poetry.org/docs/#installation). 
 
-Testing is done using tox. The config can be found in `tox.ini` within the python directory of the iceberg project.
+## Development
```
rdblue, line 35: It would be good to have a setup section that instructs you to run `pip install poetry`.

rdblue: I also had to upgrade pip and virtualenv in Ubuntu 20.04 to get poetry working properly.

rdblue: `sudo pip install virtualenv --upgrade` and `sudo python3 -m pip install --upgrade pip`

rdblue (approved):
I think this is a good change overall. It's nice to have a lock file and poetry seems really friendly. What I like most is that there is an IntelliJ plugin that I was easily able to get working so my IDE environment and CLI environment are mostly the same (fixtures don't work yet, but I probably just need to configure IntelliJ).

python/README.md
Fokko: Cool, I've added the commands 👍🏻  btw, some interesting combinations you have there. `sudo` shouldn't be required, and `pip install` and `python3 -m pip install` are equivalent.

Fokko:
Hey @rdblue thanks for giving this a try. Unfortunately, the IDEA plugin seems to be outdated; https://plugins.jetbrains.com/plugin/14307-poetry/versions Only available for <= 2021 🤔

Fokko:
@rdblue It is actually part of Intellij now 🥳
image

PR #4846, opened 2022-05-23
Core: Refactor REST catalog to RESTSessionCatalog

rdblue:
This is a minimal set of changes to refactor the `RESTCatalog` to a `RESTSessionCatalog`. The current `RESTCatalog` is removed to show changes between the session version and the original catalog implementation. A wrapper will be added in a follow-up commit.


danielcweeks (approved):

rdblue:
Thanks for the review, @danielcweeks!

PR #4847, opened 2022-05-23
Core: Add reference_snapshot_id column and filter to AllManifest table

szehon-ho:
This is a pre-requisite for expire-snapshot performance improvement in : https://github.com/apache/iceberg/pull/4736

### DIFF
--- a/core/src/main/java/org/apache/iceberg/AllManifestsTable.java
+++ b/core/src/main/java/org/apache/iceberg/AllManifestsTable.java
@@ -20,18 +20,27 @@
 package org.apache.iceberg;
 
 import java.io.IOException;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
 import org.apache.iceberg.avro.Avro;
 import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.expressions.Binder;
+import org.apache.iceberg.expressions.BoundReference;
 import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.ExpressionVisitors;
+import org.apache.iceberg.expressions.ExpressionVisitors.BoundExpressionVisitor;
 import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.expressions.Literal;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.types.Comparators;
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.StructProjection;
 
@@ -43,6 +52,7 @@ import org.apache.iceberg.util.StructProjection;
  * This table may return duplicate rows.
  */
 public class AllManifestsTable extends BaseMetadataTable {
+  private static final int REF_SNAPSHOT_ID = 18;
   private static final Schema MANIFEST_FILE_SCHEMA = new Schema(
       Types.NestedField.required(14, "content", Types.IntegerType.get()),
       Types.NestedField.required(1, "path", Types.StringType.get()),
@@ -60,7 +70,8 @@ public class AllManifestsTable extends BaseMetadataTable {
           Types.NestedField.required(11, "contains_nan", Types.BooleanType.get()),
           Types.NestedField.optional(12, "lower_bound", Types.StringType.get()),
           Types.NestedField.optional(13, "upper_bound", Types.StringType.get())
-      )))
+      ))),
+      Types.NestedField.required(REF_SNAPSHOT_ID, "reference_snapshot_id", Types.LongType.get())
   );
 
   AllManifestsTable(TableOperations ops, Table table) {
@@ -112,21 +123,25 @@ public class AllManifestsTable extends BaseMetadataTable {
       Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : filter();
       ResidualEvaluator residuals = ResidualEvaluator.unpartitioned(filter);
 
-      return CloseableIterable.withNoopClose(Iterables.transform(table().snapshots(), snap -> {
+      SnapshotEvaluator snapshotEvaluator = new SnapshotEvaluator(filter, MANIFEST_FILE_SCHEMA.asStruct(),
+          isCaseSensitive());
+      Iterable<Snapshot> filteredSnapshots = Iterables.filter(table().snapshots(), snapshotEvaluator::eval);
+
+      return CloseableIterable.withNoopClose(Iterables.transform(filteredSnapshots, snap -> {
         if (snap.manifestListLocation() != null) {
           DataFile manifestListAsDataFile = DataFiles.builder(PartitionSpec.unpartitioned())
               .withInputFile(io.newInputFile(snap.manifestListLocation()))
               .withRecordCount(1)
               .withFormat(FileFormat.AVRO)
               .build();
-          return new ManifestListReadTask(io, schema(), specs, new BaseFileScanTask(
-              manifestListAsDataFile, null,
-              schemaString, specString, residuals));
+          return new ManifestListReadTask(io, schema(), specs,
+              new BaseFileScanTask(manifestListAsDataFile, null, schemaString, specString, residuals),
+              snap.snapshotId());
         } else {
           return StaticDataTask.of(
               io.newInputFile(tableOps().current().metadataFileLocation()),
               MANIFEST_FILE_SCHEMA, schema(), snap.allManifests(io),
-              manifest -> ManifestsTable.manifestFileToRow(specs.get(manifest.partitionSpecId()), manifest)
+              manifest -> manifestFileToRow(specs.get(manifest.partitionSpecId()), manifest, snap.snapshotId())
           );
         }
       }));
@@ -138,12 +153,15 @@ public class AllManifestsTable extends BaseMetadataTable {
     private final Schema schema;
     private final Map<Integer, PartitionSpec> specs;
     private final FileScanTask manifestListTask;
+    private final long referenceSnapshotId;
 
-    ManifestListReadTask(FileIO io, Schema schema, Map<Integer, PartitionSpec> specs, FileScanTask manifestListTask) {
+    ManifestListReadTask(FileIO io, Schema schema, Map<Integer, PartitionSpec> specs, FileScanTask manifestListTask,
+                         long referenceSnapshotId) {
       this.io = io;
       this.schema = schema;
       this.specs = specs;
       this.manifestListTask = manifestListTask;
+      this.referenceSnapshotId = referenceSnapshotId;
     }
 
     @Override
@@ -164,7 +182,7 @@ public class AllManifestsTable extends BaseMetadataTable {
           .build()) {
 
         CloseableIterable<StructLike> rowIterable =  CloseableIterable.transform(manifests,
-            manifest -> ManifestsTable.manifestFileToRow(specs.get(manifest.partitionSpecId()), manifest));
+            manifest -> manifestFileToRow(specs.get(manifest.partitionSpecId()), manifest, referenceSnapshotId));
 
         StructProjection projection = StructProjection.create(MANIFEST_FILE_SCHEMA, schema);
         return CloseableIterable.transform(rowIterable, projection::wrap);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/iceberg/util/PropertyUtil.java
```diff
@@ -74,6 +76,19 @@ public static String propertyAsString(Map<String, String> properties,
     return defaultValue;
   }
 
+  public static List<Long> propertyAsLongList(Map<String, String> properties,
+                                              String property) {
+    String value = properties.get(property);
+    if (value == null) {
+      return null;
+    }
+    return Splitter.on(",")
```
kbendick, line 85: Nit: Might be good to declare the splitter one time as a private static field, as opposed to constructing it every time (unless there's some serialization-related reason it has to be reinstantiated).

aokolnychyi:
Let me take a look.

core/src/main/java/org/apache/iceberg/util/PropertyUtil.java
aokolnychyi: +1 to Kyle's suggestion

core/src/main/java/org/apache/iceberg/ScanProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+public class ScanProperties {
```
aokolnychyi, line 22: I am thinking about the best way to support this. I agree about having a Spark read option but I wonder whether we should have scan options or some sort of `useSnapshots(ids)` method in `TableScan`. We can only support that method in `ALL_*` metadata tables and throw an exception otherwise.

Thoughts, @rdblue @szehon-ho @RussellSpitzer?

rdblue: This isn't terrible, but I would much rather expose this in some other way. I don't think adding it to `TableScan` makes sense because it is very specific to this feature and would cause confusion in other cases. A data scan, for example, can't scan more than one snapshot.

What about exposing a `reference_snapshot_id` column (or with an _ for a metadata column)? If we did that, we could accept a filter from Spark, like `reference_snapshot_id IN (1, 2, 3, 4)` and special case the pushdown. We'd need to ensure that we give a consistent value for that column, which could be the latest snapshot that has the manifest.

szehon-ho: @rdblue yea , I considered that.  My thought at the time is that same manifest may have different reference_snapshot_id .  Currently I guess we don't do de-duplication in all-manifest, but this would enforce that we return one manifest per reference_snapshot_id (ie, for an append-only table, the first manifest would appear as many times as snapshots.).  Also seems like this would require a bit complex code like adding a ProjectionEvaluator for snapshot filters.  Not totally against it, and maybe this is a useful feature even outside, but seems a bit harder to do in my view.

aokolnychyi: > A data scan, for example, can't scan more than one snapshot.

It is true but `ALL_*` tables, on the other hand, don't support `useSnapshot`, `asOfTime` and incremental features.

> What about exposing a reference_snapshot_id column

I'd be open for that too. It will be the snapshot ID of the manifest list we read these manifests from, right? Adding a new column is fairly easy. How can we do the filtering? We need to filter `Snapshot`s and we have a generic filter `Expression` that can reference any columns. Is there an easy way to perform the filtering?

rdblue: Yeah, I think to perform the filtering we just need to go through all the snapshots and evaluate the expression on each one.

aokolnychyi: Essentially a custom expression visitor applied on `Snapshot` object?

rdblue:
> adding this column this would enforce that we return duplicates

Not necessarily, we could return any one or maybe the latest one. There aren't that many snapshots, so we can figure out an ordering and return the latest.

core/src/main/java/org/apache/iceberg/ScanProperties.java
rdblue: @aokolnychyi, that, or we could just pattern match and reject anything but IN predicates. As long as we're the ones controlling what filters we use, we have some flexibility.

szehon-ho:
@aokolnychyi change is ready if you guys have time for another pass, thanks

szehon-ho:
FYI @rdblue if you have time to take a look, as well, thanks

aokolnychyi:
I'll have time to take a look today. Sorry for the delay!

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+        manifest.content() == ManifestContent.DATA ? manifest.deletedFilesCount() : 0,
+        manifest.content() == ManifestContent.DELETES ? manifest.addedFilesCount() : 0,
+        manifest.content() == ManifestContent.DELETES ? manifest.existingFilesCount() : 0,
+        manifest.content() == ManifestContent.DELETES ? manifest.deletedFilesCount() : 0,
+        ManifestsTable.partitionSummariesToRows(spec, manifest.partitions()),
+        referencedSnapshotId
+    );
+  }
+
+  private static class SnapshotEvaluator extends BoundExpressionVisitor<Boolean> {
```
aokolnychyi, line 250: What about following what we did in `Evaluator`? The main idea is to have an evaluator that you instantiate once. That evaluator would have `eval(snapshot)` method and an inner visitor class. That way, we can do the binding only once instead of per snapshot and reuse the same evaluator for all snapshots.

Let me know what you think about the `Evaluator` class.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+        Literal<Long> longLit = lit.to(Types.LongType.get());
+        int cmp = longLit.comparator().compare(snapshotId, longLit.value());
+        if (!compareResult.apply(cmp)) {
+          return ROWS_CANNOT_MATCH;
+        }
+      }
+      return ROWS_MIGHT_MATCH;
+    }
+
+    private <T> boolean snapshotIdRef(BoundReference<T> ref) {
```
aokolnychyi, line 401: nit: will it be a bit more readable if we added `is` to the method name? I know we don't always do that but I feel like it would make sense. What do you think?

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+    }
+
+    @Override
+    public <T> Boolean notStartsWith(BoundReference<T> ref, Literal<T> lit) {
+      return ROWS_MIGHT_MATCH;
+    }
+
+    /**
+     * Comparison of snapshot reference and literal
+     * @param ref bound reference, comparison attempted only if reference is for reference_snapshot_id
```
aokolnychyi, line 382: nit: I'd add an empty line before params

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+     * @param lit literal value to compare with snapshot id.
+     * @param compareResult function to apply to comparison result, which returns true if rows might match
+     * @return boolean true if rows might match, false if cannot match
+     */
+    private <T> Boolean compare(BoundReference<T> ref,
+                                Literal<T> lit,
+                                Function<Integer, Boolean> compareResult) {
```
aokolnychyi, line 389: nit: is there a better name than compareResult for this function?

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+    }
+
+    @Override
+    public <T> Boolean notEq(BoundReference<T> ref, Literal<T> lit) {
+      return compare(ref, lit, compareResult -> compareResult != 0);
+    }
+
+    @Override
+    public <T> Boolean in(BoundReference<T> ref, Set<T> literalSet) {
+      Comparator<Object> longComparator = Comparators.forType(Types.LongType.get());
```
aokolnychyi, line 348: nit: Can we move this line into the if block not to instantiate it all the time?

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+      if (snapshotIdRef(ref)) {
+        boolean in = literalSet.stream().anyMatch(lit -> longComparator.compare(snapshotId, lit) == 0);
```
aokolnychyi, line 350: nit: If you want to, you can use `noneMatch` to avoid the negation in `if`:

```
boolean noneMatch = literalSet.stream().noneMatch(lit -> comparator.compare(snapshotId, lit) == 0);
if (noneMatch) {
  return ROWS_CANNOT_MATCH;
}
```

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+        }
+      }
+      return ROWS_MIGHT_MATCH;
+    }
+
+    @Override
+    public <T> Boolean notIn(BoundReference<T> ref, Set<T> literalSet) {
+      Comparator<Object> longComparator = Comparators.forType(Types.LongType.get());
+      if (snapshotIdRef(ref)) {
+        boolean notIn = literalSet.stream().noneMatch(lit -> longComparator.compare(snapshotId, lit) == 0);
```
aokolnychyi, line 362: nit: Same here, you can use `anyMatch` instead.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +228,178 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+        if (!in) {
+          return ROWS_CANNOT_MATCH;
+        }
+      }
+      return ROWS_MIGHT_MATCH;
+    }
+
+    @Override
+    public <T> Boolean notIn(BoundReference<T> ref, Set<T> literalSet) {
+      Comparator<Object> longComparator = Comparators.forType(Types.LongType.get());
```
aokolnychyi, line 360: nit: Same here

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
aokolnychyi, line 348: `Comparator<Object>` looked suspicious at first but I think it should work.

aokolnychyi (commented):
The implementation looks correct to me. I left some minor comments.
Will take a look at the tests tomorrow.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Makes sense, done

szehon-ho:
Thanks for review, addressed the round of comments

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -43,6 +52,7 @@
  * This table may return duplicate rows.
  */
 public class AllManifestsTable extends BaseMetadataTable {
+  private static final int REF_SNAPSHOT_ID = 18;
```
RussellSpitzer, line 55: Are we keeping this a constant because we refer to it again in the isSnapshotRef function? I just wonder because we don't do this for any of our other fields in the schema.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -112,21 +123,29 @@ protected CloseableIterable<FileScanTask> doPlanFiles() {
...
         if (snap.manifestListLocation() != null) {
           DataFile manifestListAsDataFile = DataFiles.builder(PartitionSpec.unpartitioned())
               .withInputFile(io.newInputFile(snap.manifestListLocation()))
               .withRecordCount(1)
               .withFormat(FileFormat.AVRO)
               .build();
-          return new ManifestListReadTask(io, schema(), specs, new BaseFileScanTask(
-              manifestListAsDataFile, null,
-              schemaString, specString, residuals));
+          return new ManifestListReadTask(
```
RussellSpitzer, line 138: We generally don't do 1 arg a line, minor formatting nit

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Yea that's it actually, its used in two places.  Thoughts?

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +229,186 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+      }
+
+      @Override
+      public <T> Boolean lt(BoundReference<T> ref, Literal<T> lit) {
+        return compareSnapshotRef(ref, lit, compareResult -> compareResult < 0);
+      }
+
+      @Override
+      public <T> Boolean ltEq(BoundReference<T> ref, Literal<T> lit) {
+        return compareSnapshotRef(ref, lit, compareResult -> compareResult <= 0);
```
RussellSpitzer, line 330: We talked about these a bit offline, I think the gt/lt's here probably won't every be useful but I don't see a problem in including them

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +229,186 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+      @Override
+      public <T> Boolean notEq(BoundReference<T> ref, Literal<T> lit) {
+        return compareSnapshotRef(ref, lit, compareResult -> compareResult != 0);
+      }
+
+      @Override
+      public <T> Boolean in(BoundReference<T> ref, Set<T> literalSet) {
+        if (isSnapshotRef(ref)) {
+          Comparator<Object> longComparator = Comparators.forType(Types.LongType.get());
+          boolean noneMatch = literalSet.stream().noneMatch(lit -> longComparator.compare(snapshotId, lit) == 0);
```
RussellSpitzer, line 358: Can we do this with 

```literalSet.contains(snapshotId)?```

File: core/src/main/java/org/apache/iceberg/AllManifestsTable.java [hunk 15]
```diff
@@ -204,4 +229,186 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+          }
+        }
+        return ROWS_MIGHT_MATCH;
+      }
+
+      @Override
+      public <T> Boolean notIn(BoundReference<T> ref, Set<T> literalSet) {
+        if (isSnapshotRef(ref)) {
+          Comparator<Object> longComparator = Comparators.forType(Types.LongType.get());
+          boolean anyMatch = literalSet.stream().anyMatch(lit -> longComparator.compare(snapshotId, lit) == 0);
```
RussellSpitzer, line 370: Another call of "contains?"

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +229,186 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+
+      /**
+       * Comparison of snapshot reference and literal, using long comparator.
+       *
+       * @param ref           bound reference, comparison attempted only if reference is for reference_snapshot_id
+       * @param lit           literal value to compare with snapshot id.
+       * @param desiredResult function to apply to long comparator result, returns true if result is as expected.
+       * @return false if comparator does not achieve desired result, true otherwise
+       */
+      private <T> Boolean compareSnapshotRef(BoundReference<T> ref,
```
RussellSpitzer, line 396: This (an others have little formatting issues)

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -138,12 +154,18 @@ static class ManifestListReadTask implements DataTask {
     private final Schema schema;
     private final Map<Integer, PartitionSpec> specs;
     private final FileScanTask manifestListTask;
+    private final long referenceSnapshotId;
 
-    ManifestListReadTask(FileIO io, Schema schema, Map<Integer, PartitionSpec> specs, FileScanTask manifestListTask) {
+    ManifestListReadTask(FileIO io,
```
RussellSpitzer, line 159: formatting

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +226,190 @@ public Iterable<FileScanTask> split(long splitSize) {
...
   }
+
+  static StaticDataTask.Row manifestFileToRow(PartitionSpec spec, ManifestFile manifest, long referenceSnapshotId) {
+    return StaticDataTask.Row.of(
+        manifest.content().id(),
+        manifest.path(),
+        manifest.length(),
+        manifest.partitionSpecId(),
+        manifest.snapshotId(),
+        manifest.content() == ManifestContent.DATA ? manifest.addedFilesCount() : 0,
```
RussellSpitzer, line 237: I think 0 may be the wrong thing here, shouldn't it be "null" if we don't have the info?

RussellSpitzer (approved):
I had a few comments on using "Contains" in the SnapshotEvaluator but this looks close to me.

There are also number of formatting nits as well we should clean up before merging.

aokolnychyi:
Let me take another look as well.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
aokolnychyi: I think the problem is in the original schema in `ManifestsTable`, where the counts are required. I also did not notice that data file counts are optional in `AllManifestsTable` when I recently added counts for delete files. I think we should continue to use 0 for compatibility but I wonder why we have such a mismatch in nullability between these tables.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -112,21 +123,26 @@ protected CloseableIterable<FileScanTask> doPlanFiles() {
       Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : filter();
       ResidualEvaluator residuals = ResidualEvaluator.unpartitioned(filter);
 
-      return CloseableIterable.withNoopClose(Iterables.transform(table().snapshots(), snap -> {
+      SnapshotEvaluator snapshotEvaluator = new SnapshotEvaluator(filter, MANIFEST_FILE_SCHEMA.asStruct(),
+          isCaseSensitive());
+      Iterable<Snapshot> filteredSnapshots = Iterables.filter(table().snapshots(),
+          snapshot -> snapshotEvaluator.eval(snapshot));
```
aokolnychyi, line 129: nit: this can fit one one line

```
Iterable<Snapshot> filteredSnapshots = Iterables.filter(table().snapshots(), snapshotEvaluator::eval);
```

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -112,21 +123,26 @@ protected CloseableIterable<FileScanTask> doPlanFiles() {
       Expression filter = shouldIgnoreResiduals() ? Expressions.alwaysTrue() : filter();
       ResidualEvaluator residuals = ResidualEvaluator.unpartitioned(filter);
 
-      return CloseableIterable.withNoopClose(Iterables.transform(table().snapshots(), snap -> {
+      SnapshotEvaluator snapshotEvaluator = new SnapshotEvaluator(filter, MANIFEST_FILE_SCHEMA.asStruct(),
```
aokolnychyi, line 126: nit: if you want to keep it on one line, you could just use `MANIFEST_FILE_SCHEMA` in the constructor of the evaluator class (you have access to it there)

aokolnychyi (approved):

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
aokolnychyi: I'd still use 0 for data files count when reading a delete manifest as null means unknown but we know there is no data file in a delete manifest.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Yea I think its the same in this case, but thought its safer to use the Comparator lookup for Long type to get the official Iceberg way to compare (if it ever differs from Java's comparator)

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Ah great.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: I might keep it as argument, the class is more re-usable if it takes the schema as an argument, was thinking we could add this for other metadata tables at some point.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
kbendick, line 55: Side comment / non-blocking: I'm a _big_ fan of using named constants for schema element IDs.

Should we update to use named position constants throughout, particularly for schemas like the metadata tables that are evolving (note that `content` is listed first but has ID 14). But is the cherry-picking that would be required by a large diff like this considered an unnecessary change?

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
```diff
@@ -204,4 +222,188 @@ public Iterable<FileScanTask> split(long splitSize) {
...
+       *
+       * @param ref           bound reference, comparison attempted only if reference is for reference_snapshot_id
+       * @param lit           literal value to compare with snapshot id.
+       * @param desiredResult function to apply to long comparator result, returns true if result is as expected.
+       * @return false if comparator does not achieve desired result, true otherwise
+       */
+      private <T> Boolean compareSnapshotRef(BoundReference<T> ref, Literal<T> lit,
+                                             Function<Integer, Boolean> desiredResult) {
+        if (isSnapshotRef(ref)) {
+          Literal<Long> longLit = lit.to(Types.LongType.get());
```
kbendick, line 395: I imagine it’s unlikely to happen as the query should throw before this, but any concern with potential `null` being passed in for`lit`?

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -991,18 +1005,34 @@ public void testAllManifestsTable() {
...
+
+    table.newRowDelta()
+        .addDeletes(deleteFile)
+        .commit();
 
     table.newDelete().deleteFromRowFilter(Expressions.alwaysTrue()).commit();
 
-    manifests.addAll(table.currentSnapshot().allManifests());
+    Stream<Pair<Long, ManifestFile>> snapshotIdToManifests =
+        StreamSupport.stream(table.snapshots().spliterator(), false)
```
kbendick, line 1033: Nit: Can you add an inline comment for what the `false` stands for? If its something like parallel or something then don’t worry about it, but usually it helps me reading the code to have the parameter name in-lined like `, /* caseSensitive */ false);`.

But again, thinking on it (and the usage of spliterator), if it’s parallel or something then feel free to ignore it.

szehon-ho: Yea I usually do, but this one is just parallel and pretty common, so maybe skip this time?  Would be great if StreamSupport has a default value.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Yea I am not sure if there is consensus, Russell made a discussion in https://github.com/apache/iceberg/pull/4847#discussion_r906391975 but didn't finished (not sure if there is a big value for the other constants if they are never referenced outside?).  But If so, we should do it in another pr as its unrelated.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Yea it seems it's checked when trying to make the predicate:
```
java.lang.NullPointerException: Cannot create expression literal from null
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkNotNull(Preconditions.java:907)
	at org.apache.iceberg.expressions.Literals.from(Literals.java:62)
	at org.apache.iceberg.expressions.UnboundPredicate.<init>(UnboundPredicate.java:40)
	at org.apache.iceberg.expressions.Expressions.equal(Expressions.java:175)
```

So i think it's not necessary, as it's pretty internal.  Also, none of the other Evaluators check this.

szehon-ho:
Restarting as Flink test is frozen.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: Yea this is copy + add from ManifestsTable class (should have mentioned), so if we fix it can be in another pr.

szehon-ho:
Fixed and added to new Spark 3.3 test module, merged.  Thanks all for review @kbendick @aokolnychyi @RussellSpitzer

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
rdblue: Sorry I'm late to this review, but the `literalSet` will use the correct comparator for a given type.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
rdblue: Yeah, we should eventually update this. The set's `contains` method will be correct. It's probably not a big deal right now since there are so few snapshots.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
rdblue: @szehon-ho, once the expression is bound, you are guaranteed that the `lit` corresponding to the ref will be a `LongType` because binding coerces the literal types to match the corresponding term. This shouldn't be needed, unless you're doing it to get a `Long` comparator because you have a `long snapshotId`.

core/src/main/java/org/apache/iceberg/AllManifestsTable.java
szehon-ho: OK works for me , I will make a follow up pr then!

PR #4848, opened 2022-05-23
Spark: Fix Alignment of Merge Commands with Mixed Case

RussellSpitzer:
Prior to this a mixed-case insert statement would fail to be marked
as aligned after our alignement rule was applied. This would occur
because Spark is allowed to opperate without case sensitivity. Although
we would correctly align the fields, our check for alignment required
an exact match even with the system was set to be case-insensitive. Failing this
check would mean our RewriteToMerge analysis rule would never get applied.

### DIFF
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
@@ -84,16 +84,7 @@ object AlignRowLevelCommandAssignments
           throw new AnalysisException("Not matched actions can only contain INSERT")
       }
 
-      val alignedMerge = m.copy(
-        matchedActions = alignedMatchedActions,
-        notMatchedActions = alignedNotMatchedActions)
-
-      if (!alignedMerge.aligned) {
-        // This should never happen, but it is better than failing silently and allowing analysis to continue
-        throw new AnalysisException(s"Alignment of row level commands failed")
-      }
-
-      alignedMerge
+      m.copy(matchedActions = alignedMatchedActions, notMatchedActions = alignedNotMatchedActions)
   }
 
   private def alignInsertActionAssignments(
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
@@ -84,7 +84,16 @@ object AlignRowLevelCommandAssignments
           throw new AnalysisException("Not matched actions can only contain INSERT")
       }
 
-      m.copy(matchedActions = alignedMatchedActions, notMatchedActions = alignedNotMatchedActions)
+      val alignedMerge = m.copy(
+        matchedActions = alignedMatchedActions,
+        notMatchedActions = alignedNotMatchedActions)
+
+      if (!alignedMerge.aligned) {
+        // This should never happen, but it is better than failing silently and allowing analysis to continue
+        throw new AnalysisException(s"Alignment of row level commands failed")
+      }
+
+      alignedMerge
   }
 
   private def alignInsertActionAssignments(
--- a/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignedRowLevelIcebergCommandCheck.scala
+++ b/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignedRowLevelIcebergCommandCheck.scala
@@ -29,9 +29,9 @@ object AlignedRowLevelIcebergCommandCheck extends (LogicalPlan => Unit) {
   override def apply(plan: LogicalPlan): Unit = {
     plan foreach {
       case m: MergeIntoIcebergTable if !m.aligned =>
-        throw new AnalysisException(s"Could not align Iceberg MERGE INT: $m")
+        throw new AnalysisException(s"Could not align Iceberg MERGE INTO: $m")
       case u: UpdateIcebergTable if !u.aligned =>
-        throw new AnalysisException(s"Could not align Iceberg UPDATE was never aligned: $u")
+        throw new AnalysisException(s"Could not align Iceberg UPDATE: $u")
       case _ => // OK
     }
   }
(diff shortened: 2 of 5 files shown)


spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
```diff
@@ -84,7 +84,16 @@ object AlignRowLevelCommandAssignments
           throw new AnalysisException("Not matched actions can only contain INSERT")
       }
 
-      m.copy(matchedActions = alignedMatchedActions, notMatchedActions = alignedNotMatchedActions)
+      val alignedMerge = m.copy(
+        matchedActions = alignedMatchedActions,
+        notMatchedActions = alignedNotMatchedActions)
+
+      if (!alignedMerge.aligned) {
```
aokolnychyi, line 91: An alternative way of solving this is to provide a check like `MergeIntoIcebergTableResolutionCheck`. It will be run after all resolution rules and we will know for sure that we failed to align the assignments. If we fail here, we won't give Spark any chances to fix the alignments using other rules.

aokolnychyi: I wonder whether we should also cover UPDATEs too?

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/expressions/AssignmentUtils.scala
```diff
@@ -40,7 +40,15 @@ object AssignmentUtils extends SQLConfHelper {
     sameSize && table.output.zip(assignments).forall { case (attr, assignment) =>
       val key = assignment.key
       val value = assignment.value
-      toAssignmentRef(attr) == toAssignmentRef(key) &&
+      val refsEqual = if (conf.caseSensitiveAnalysis) {
```
aokolnychyi, line 43: Seems like we can use `conf.resolver` that would abstract this away? Then we will have only one case to handle.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMerge.java
```diff
@@ -1186,6 +1186,32 @@ public void testMergeAlignsUpdateAndInsertActions() {
         sql("SELECT * FROM %s ORDER BY id", tableName));
   }
 
+  @Test
+  public void testMergeMixedCaseAlignsUpdateAndInsertActions() {
+    createAndInitTable("id INT, a INT, b STRING", "{ \"id\": 1, \"a\": 2, \"b\": \"str\" }");
+    createOrReplaceView(
+        "source",
+        "{ \"id\": 1, \"c1\": -2, \"c2\": \"new_str_1\" }\n" +
+            "{ \"id\": 2, \"c1\": -20, \"c2\": \"new_str_2\" }");
```
aokolnychyi, line 1195: nit: the rest of this file aligns the json records slightly differently (no extra indentation for 2nd record)

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMerge.java
```diff
@@ -1186,6 +1186,32 @@ public void testMergeAlignsUpdateAndInsertActions() {
...
+
+    sql("MERGE INTO %s t USING source " +
+            "ON t.iD == source.Id " +
```
aokolnychyi, line 1198: nit: same for MERGE statements

aokolnychyi (commented):
Great catch!

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
RussellSpitzer: I did add in a test case for Update. I think that's a good point, I can also just add a case to the Merge Rewrite Rule itself to match on any unaligned merge to an Iceberg table

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/expressions/AssignmentUtils.scala
RussellSpitzer: That is a great idea

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMerge.java
RussellSpitzer: ugh it was the intellij helper annotations, they made it look aligned :)

kbendick: Hahaha that happens to me all the time. So annoying.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala
RussellSpitzer: Add check rule for both Updates and Merges

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignedRowLevelIcebergCommandCheck.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.plans.logical.MergeIntoIcebergTable
+import org.apache.spark.sql.catalyst.plans.logical.UpdateIcebergTable
+
+object AlignedRowLevelIcebergCommandCheck extends (LogicalPlan => Unit) {
+
+  override def apply(plan: LogicalPlan): Unit = {
+    plan foreach {
+      case m: MergeIntoIcebergTable if !m.aligned =>
+        throw new AnalysisException(s"Could not align Iceberg MERGE INT: $m")
```
aokolnychyi, line 32: nit: looks like a typo at the end?

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignedRowLevelIcebergCommandCheck.scala
```diff
@@ -0,0 +1,38 @@
...
+      case u: UpdateIcebergTable if !u.aligned =>
+        throw new AnalysisException(s"Could not align Iceberg UPDATE was never aligned: $u")
```
aokolnychyi, line 34: nit: do we need "was never aligned" part?

aokolnychyi (approved):
LGTM

RussellSpitzer:
I blame all typos on the fact that I don't have my external monitor

PR #4851, opened 2022-05-24
Build: upgrade github actions

XN137:
```
- use ubuntu-20.04 explicitly in anticipation of the next ubuntu release
- upgrade to actions/cache@v3
- upgrade to actions/checkout@v3
- upgrade to actions/labeler@v4
- upgrade to actions/setup-java@v3
- upgrade to actions/setup-python@v3
- upgrade to actions/upload-artifact@v3
```

this brings various improvements for each action, see changelogs:
https://github.com/actions/cache#v3
https://github.com/actions/checkout/#checkout-v3
https://github.com/actions/setup-java#v2-vs-v1
https://github.com/actions/setup-python#whats-new
https://github.com/actions/labeler/releases
https://github.com/actions/upload-artifact/releases

a common change across all actions is that they are now based on node16 LTS since node12 LTS is going EOL later this year.

we explicitly use the `ubuntu-20.04` runner, since `ubuntu-latest` will get updated to `22.04` in the future, and it should be preferable to make this upgrade intentionally, instead of automatically.

all updates were done with `sed` apart from `setup-java` which needs another property (`zulu` distribution was the default in earlier versions):
https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md#java-distribution 

before:
```
~/code/iceberg$ rg --no-filename -o 'uses:.*$' .github | sort | uniq -c
      9 uses: actions/cache@v2
     16 uses: actions/checkout@v2
      1 uses: actions/labeler@v2
     12 uses: actions/setup-java@v1
      2 uses: actions/setup-python@v2
      9 uses: actions/upload-artifact@v2
```

after:
```
/code/iceberg$ rg --no-filename -o 'uses:.*$' .github | sort | uniq -c
      9 uses: actions/cache@v3
     16 uses: actions/checkout@v3
      1 uses: actions/labeler@v4
     12 uses: actions/setup-java@v3
      2 uses: actions/setup-python@v3
      9 uses: actions/upload-artifact@v3
```

sanity check for the manual modification:
```
~/code/iceberg$ rg "distribution: zulu" .github | wc -l
12
~/code/iceberg$ rg "setup-java" .github | wc -l
12
```

### DIFF
--- a/.github/workflows/flink-ci.yml
+++ b/.github/workflows/flink-ci.yml
@@ -55,7 +55,7 @@ jobs:
 
   # Only test the latest flink version with scala 2.11 for saving testing time.
   flink-scala-2-11-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         jvm: [ 8, 11 ]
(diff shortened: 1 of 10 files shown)


nastra (approved):

Fokko (approved):

kbendick (approved):
LGTM. I would keep in mind that these changes are _not_ the tests that got run as part of CI in this PR (github uses the current definition in master).

But these specific changes could be run in a fork by merging it into the master branch of your fork etc (if you've not done so already).

Question about explicitly setting the ubuntu version: Why not just move to 22 now? Just to keep the changes somewhat separated?

XN137:
thanks for the reviews.

> I would keep in mind that these changes are not the tests that got run as part of CI in this PR (github uses the current definition in master).

is this really true?
when i check the workflow runs on this PR i see all the latest action versions being used, for example:
Screenshot from 2022-05-25 08-15-38
https://github.com/apache/iceberg/runs/6568370241?check_suite_focus=true

Afaik only the labeler workflow isnt being exercised as part of this PR (so there is a very low risk of breaking it on merge, but I would be checking that right after).

> Question about explicitly setting the ubuntu version: Why not just move to 22 now? Just to keep the changes somewhat separated?

the image is still in beta, see https://github.com/actions/virtual-environments/issues/5490
while we could try upgrading, we should do so in a separate PR imo.
as stated earlier the idea to avoid using "ubuntu-latest" is to have full control on when that upgrade happens (to make sure CI remains stable).

rdblue (approved):

rdblue:
Thanks, @XN137!

PR #4853, opened 2022-05-24
Flink: Modify Table Location Strings in Tests to Fix Local Test Failures

wuwenchi:
In the original code, the path is directly written `'location'='/tmp/location'` , and hdfs will access the local port 9000.
However, in general, hadoop is not installed during local testing, which causes the connection port to fail and the test case to report an error. 
So, I modified the path to store it on disk `'location'='file:///tmp/location'` , which is more convenient for local testing.


wuwenchi:
@openinx can you review it? thx!

stevenzwu (approved):

kbendick (approved):
Looks good to me.

kbendick:
You might want to rename this PR to `Flink - Modify Table Locations in Test to Fix Local Test Failures`.

You'll probably get more eyes that way 👍  cc @wuwenchi

wuwenchi:
@kbendick Great! More understandable!

rdblue (approved):
Looks reasonable to me. Since we intend to use the local FS, we should just specify that in the table location.

rdblue:
Thanks, @wuwenchi!

Issue #4828, opened 2022-05-20
[REST] Add MetadataUpdate.AssignUUID action types to OpenAPI spec


kbendick:
The MetadataUpdate actions AssignUUID is not listed in the OpenAPI spec.

It's also not currently fully implemented, but we do have the code in place for JSON SerDe, as well as the class (only the `applyTo` logic has not been fleshed out yet).

We will need to add it to the REST OpenAPI spec in the following enum, as well as add a class for it: https://github.com/apache/iceberg/blob/b8626eecc139f638b162f06a5f8f3c8bb210346a/open-api/rest-catalog-open-api.yaml#L1175-L1195

The class itself is defined here: https://github.com/apache/iceberg/blob/b8626eecc139f638b162f06a5f8f3c8bb210346a/core/src/main/java/org/apache/iceberg/MetadataUpdate.java#L34-L49

PR #4860, opened 2022-05-24
[REST Catalog Spec] Add MetadataUpdate.AssignUUID to the OpenAPI Specification

kbendick:
This closes issue https://github.com/apache/iceberg/issues/4828

For some very old tables (e.g. from the 0.9.x-era or incubating era), the table does not have UUID in it.

This class allows server side implementors to add a UUID to those tables (in case they havent been touched in a very long time).

This is a _very_ rare use case, so we've encoded it in the `MetadataUpdate` if people need it, but it's a very rare corner case.

This is encoded in the class MetadataUpdate.AssignUUID, but it's not present in the REST Catalog's spec.
https://github.com/apache/iceberg/blob/b8626eecc139f638b162f06a5f8f3c8bb210346a/core/src/main/java/org/apache/iceberg/MetadataUpdate.java#L34-L49

Normally, a table's UUID will be udated as part of the normal update table / create table workload.

### DIFF
--- a/open-api/rest-catalog-open-api.yaml
+++ b/open-api/rest-catalog-open-api.yaml
@@ -1222,6 +1222,7 @@ components:
         action:
           type: string
           enum:
+            - assign-uuid
             - upgrade-format-version
             - add-schema
             - set-current-schema
@@ -1235,6 +1236,17 @@ components:
             - set-location
             - set-properties
             - remove-properties
+
+    AssignUUIDUpdate:
+      allOf:
+        - $ref: '#/components/schemas/BaseUpdate'
+        - type: object
+          required:
+            - uuid
+          properties:
+            uuid:
+              type: string
+              description: Table UUID to set as current
 
     UpgradeFormatVersionUpdate:
       allOf:
@@ -1379,6 +1391,7 @@ components:
 
     TableUpdate:
       anyOf:
+        - $ref: '#/components/schemas/AssignUUIDUpdate'
         - $ref: '#/components/schemas/UpgradeFormatVersionUpdate'
         - $ref: '#/components/schemas/AddSchemaUpdate'
         - $ref: '#/components/schemas/SetCurrentSchemaUpdate'


singhpk234 (approved):
LGTM !!

Found [[RemoveSnapshotRef](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/MetadataUpdate.java#L218-#L233) also missing, added a [PR](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/MetadataUpdate.java#L218-#L233) for the same.

kbendick:
> LGTM !!
> 
> Found [RemoveSnapshotRef](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/MetadataUpdate.java#L218-#L233) also missing, added a [PR](https://github.com/apache/iceberg/pull/4866) for the same.

Good catch. That was actually the main one I was thinking of.

This one is actually _not_ as important. The use case for this is actually just if people have very old table's that _don't_ already have UUID in them. I wrote the issue comment in a rush and completely misrepresented it.

I am going to close this PR in favor of adding `RemoveSnapshotRef` (which is the one I was looking for). This class is mostly available in case some 3rd party needs to add UUIDs to existing tables, or make table UUIDs meet their internal standards (not something I myself use, hence the confusion).

In an effort to keep the open api spec consumable, we've decided to leave this class out as it's really just in the off chance somebody has very very old tables that haven't been touched in a while but need to add UUIDs for them. Or some third party finds a use case for it to match their internal systems (of which I don't know any).

So thanks for the find @singhpk234. As mentioned, RemoveSnapshotRef was actually the one I was thinking of when I made this issue.

I'lll be writing a lot of documentation on the REST catalog coming up. (I also submitted for a talk on it to ApacheCon so if that gets accepted there will likely be more documentation than anybody wants on the subject). Cheers.

PR #4863, opened 2022-05-25
Fix invalid Stream to Iterable conversion in Tasks

findepi:
`Stream.iterator` is a one-shot operation, cannot be called multiple
times, so `stream::iterator` shouldn't be used as an implementation of
`Iterable`.

Similar to https://github.com/apache/iceberg/pull/4806


core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -205,11 +215,12 @@ private <E extends Exception> boolean runSingleThreaded(
       List<I> succeeded = Lists.newArrayList();
       List<Throwable> exceptions = Lists.newArrayList();
 
-      Iterator<I> iterator = items.iterator();
+      Preconditions.checkState(!hasStarted, "Cannot run twice");
+      hasStarted = true;
```
kbendick, line 219: Nit: normally we only use `this.` for field when mutating them. In most cases, this winds up just being the constructor or when set in the builder.

I'm not sure if we would use it here or not.

I wouldn't block on this by any means, but figured I'd mention it 🙂

core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -88,8 +89,17 @@ public UnrecoverableException(Throwable cause) {
     private long maxSleepTimeMs = 600000; // 10 minutes
     private long maxDurationMs = 600000;  // 10 minutes
     private double scaleFactor = 2.0;     // exponential
+    private boolean hasStarted = false;
```
kbendick, line 92: Since this can be updated in `runParallel`, should we consider making this an `AtomicBoolean` like the booleans defined inside of `runParallel` are?

kbendick (commented):
Thanks for tackling this work @findepi! This has bugged for me a while.

I left a comment (as well as a few nits that are _absolutely_ non-blocking). Given the importance / widespread use of this code, I'll hold off on approving for the moment but will come back to this later today if nobody approves between then.

core/src/main/java/org/apache/iceberg/util/Tasks.java
kbendick: Looking closer, it seems like this wouldn't actually be set in some other thread, so I suppose it's ok. I would defer to your expertise on the issue 🙂

core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -88,8 +89,17 @@ public UnrecoverableException(Throwable cause) {
...
 
+    /**
+     * @deprecated Use {@link #Builder(Iterator)} instead.
```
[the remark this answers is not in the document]
findepi: > I think you can use `@deprecated since 0.14.0`.

I don't think it adds value. If a user is < 0.14, they won't see deprecation notice.
If they are on >= 0.14, they see it. 

It would be valuable if it comes with the "will be removed", but it's not my call to make promises like this.

Please advise what should i do.

core/src/main/java/org/apache/iceberg/util/Tasks.java
findepi: Should i also reference `hasStarted` with `this.` in the checkState above?

should i also reference `items` with `this.` when calling `.next()`? (this is a mutation after all)

Please advise what should i do.

core/src/main/java/org/apache/iceberg/util/Tasks.java
findepi: `runParallel` starts threads, but i agree with you, the field isn't mutated in such a thread.

core/src/main/java/org/apache/iceberg/util/Tasks.java
rdblue: Only use `this.` when modifying instance state. It helps the reader know that it is instance state being modified rather than local state. No need to use it when accessing state, only when modifying.

core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -88,8 +89,17 @@ public UnrecoverableException(Throwable cause) {
...
+     */
+    @Deprecated
```
rdblue, line 97: There's no need to deprecate this. It isn't invalid to pass an `Iterable`.

core/src/main/java/org/apache/iceberg/util/Tasks.java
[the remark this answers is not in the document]
kbendick: It doesn’t have to be added if it truly doesn’t add values.

We have some places where it’s not present, but I also tend to think of it as a notice for people who fork the project in case they’re using it.

I also think of it as something that reminds us (or specifically release managers) when something can be removed entirely.

Burr I agree here it wouldn’t present much end user facing value.

core/src/main/java/org/apache/iceberg/util/Tasks.java
```diff
@@ -570,7 +584,7 @@ public static <I> Builder<I> foreach(I... items) {
   }
 
   public static <I> Builder<I> foreach(Stream<I> items) {
-    return new Builder<>(items::iterator);
+    return new Builder<>(items.iterator());
```
rdblue, line 587: I think a better fix is to introduce a one-time use `Iterable` that enforces `hasStarted` internally. That way we don't need to modify the code in `Tasks`, which is a much larger risk.

core/src/main/java/org/apache/iceberg/util/Tasks.java
kbendick: Isn’t this line of code modifying instances state though (assigning a class variable from false to true)?

core/src/main/java/org/apache/iceberg/util/Tasks.java
findepi: > I think a better fix is to introduce a one-time use `Iterable` that enforces `hasStarted` internally.

i would assume `stream::iterator` does that. Still, it's not a real `Iterable` IMO.

rdblue: No, `stream::iterator` will allow calling `iterator` more than once. Adding a single-use iterable would fix this without lots of code changes.

findepi: > No, `stream::iterator` will allow calling `iterator` more than once.

I didn't know that, thanks!

> Adding a single-use iterable would fix this 

That's true. But `Iterable` interface contract isn't clear about whether "a single-use iterable" is a valid implementation, so I wouldn't want to create such a class. I wouldn't stop you if you wanted to, though. 

> without lots of code changes. 

from long-term design perspective this may or may not be a dominant factor

PR #4866, opened 2022-05-25
Core / OpenAPI: Add RemoveSnapshotRef to MetadataUpdateParser and OpenAPI spec

singhpk234:
### About the change
This PR added the `RemoveSnapshotRef` to both MetadataUpdateParser and OpenAPI spec. 

Closes https://github.com/apache/iceberg/issues/4876

P.S Thanks @kbendick for beautifully documenting the change in the above issue :) !!!

### Testing Done 
Added UT for the same.

----- 

cc @rdblue @kbendick

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdate.java
@@ -216,19 +216,19 @@ public interface MetadataUpdate extends Serializable {
   }
 
   class RemoveSnapshotRef implements MetadataUpdate {
-    private final String name;
+    private final String refName;
 
-    public RemoveSnapshotRef(String name) {
-      this.name = name;
+    public RemoveSnapshotRef(String refName) {
+      this.refName = refName;
     }
 
     public String name() {
-      return name;
+      return refName;
     }
 
     @Override
     public void applyTo(TableMetadata.Builder metadataBuilder) {
-      metadataBuilder.removeRef(name);
+      metadataBuilder.removeRef(refName);
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
+++ b/core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
@@ -51,8 +51,8 @@ public class MetadataUpdateParser {
   static final String SET_DEFAULT_SORT_ORDER = "set-default-sort-order";
   static final String ADD_SNAPSHOT = "add-snapshot";
   static final String REMOVE_SNAPSHOTS = "remove-snapshots";
-  static final String SET_SNAPSHOT_REF = "set-snapshot-ref";
   static final String REMOVE_SNAPSHOT_REF = "remove-snapshot-ref";
+  static final String SET_SNAPSHOT_REF = "set-snapshot-ref";
   static final String SET_PROPERTIES = "set-properties";
   static final String REMOVE_PROPERTIES = "remove-properties";
   static final String SET_LOCATION = "set-location";
@@ -89,7 +89,7 @@ public class MetadataUpdateParser {
   private static final String SNAPSHOT_IDS = "snapshot-ids";
 
   // SetSnapshotRef
-  private static final String REF_NAME = "ref-name";
+  private static final String REF_NAME = "ref-name"; // Also used in RemoveSnapshotRef
   private static final String SNAPSHOT_ID = "snapshot-id";
   private static final String TYPE = "type";
   private static final String MIN_SNAPSHOTS_TO_KEEP = "min-snapshots-to-keep";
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -52,6 +52,7 @@ private MetadataUpdateParser() {
   static final String ADD_SNAPSHOT = "add-snapshot";
   static final String REMOVE_SNAPSHOTS = "remove-snapshots";
   static final String SET_SNAPSHOT_REF = "set-snapshot-ref";
+  static final String REMOVE_SNAPSHOT_REF = "remove-snapshot-ref";
```
kbendick, line 55: Nit: Can you move this up to be above `set-snapshot-ref`? This follows the order that the classes are declared in. So move this to line 54 (and let `set-snapshot-ref` go down one line).

core/src/main/java/org/apache/iceberg/MetadataUpdate.java
```diff
@@ -216,19 +216,19 @@ public void applyTo(TableMetadata.Builder metadataBuilder) {
   }
 
   class RemoveSnapshotRef implements MetadataUpdate {
-    private final String name;
+    private final String refName;
```
kbendick, line 219: 👍  This change is good. If somebody were to implement their own small enums in another language or with something that could go from kebab-case to camel-case automatically, then this change would be sufficient to allow them to update the parser themselves.

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -117,6 +118,7 @@ private MetadataUpdateParser() {
       .put(MetadataUpdate.AddSnapshot.class, ADD_SNAPSHOT)
       .put(MetadataUpdate.RemoveSnapshot.class, REMOVE_SNAPSHOTS)
       .put(MetadataUpdate.SetSnapshotRef.class, SET_SNAPSHOT_REF)
+      .put(MetadataUpdate.RemoveSnapshotRef.class, REMOVE_SNAPSHOT_REF)
```
kbendick, line 121: Nit: Same thing about moving this to be below `REMOVE_SNAPSHOTS`.

open-api/rest-catalog-open-api.yaml
```diff
@@ -1343,6 +1344,17 @@ components:
...
+    RemoveSnapshotRefUpdate:
+      allOf:
+        - $ref: '#/components/schemas/BaseUpdate'
+        - type: object
+          required:
+            - ref-name
+          properties:
+            name:
+              type: string
+              description: SnapshotRef name to remove
```
kbendick, line 1356: Nit: Maybe `Name of the snapshot ref to remove`? We tend to prefer plain English sentences in descriptions.

kbendick (approved):
Thanks @singhpk234. I appreciate you taking this off my plate for me.

This is pretty much complete, just some minor style changes if you don't mind. Having the classes and functons ordered int he same way as they are listed in `MetadataUpdate` helps me personally to match things together when my brain isn't functioning optimally.

Buit on a technical level, this is perfect.

One last nit, but could you add a comment on line 91 of `MetadataUpdateParser` that `REF_NAME` is also used in `RemoveSnapshotRef`?

Maybe just the following would be sufficient.
```java
private static final String REF_NAME = "ref-name";  // Also used in RemoveSnapshotRef
```

Thanks again!

singhpk234:
> Thanks @singhpk234. I appreciate you taking this off my plate for me.

It was pleasure :) !!!

> Having the classes and functons ordered int he same way as they are listed in MetadataUpdate

This makes perfect sense, have addressed it along with the remaining feedbacks :).

Thanks for thorough review @kbendick !!!

core/src/main/java/org/apache/iceberg/MetadataUpdateParser.java
```diff
@@ -324,6 +331,12 @@ private static void writeSetSnapshotRef(MetadataUpdate.SetSnapshotRef update, Js
     JsonUtil.writeLongFieldIf(update.maxRefAgeMs() != null, MAX_REF_AGE_MS, update.maxRefAgeMs(), gen);
   }
 
+  private static void writeRemoveSnapshotRef(
+      MetadataUpdate.RemoveSnapshotRef update,
+      JsonGenerator gen) throws IOException {
```
kbendick, line 336: Nit: if this could be on one line, possibly with the `throws IOException` on its own line if it doesn't all fit on one line, that would be great / fit what is done elsewhere.

But I wouldn't block on this PR for that.

```java
  private static void writeRemoveSnapshotRef(MetadataUpdate.RemoveSnapshotRef update, JsonGenerater gen)
      throws IOException {
    gen.writeStringField(REF_NAME, update.name());
  }
```

kbendick: `writeSetDefaultSortOrder` is written in this style iirc.

kbendick (approved):
Reconfirming my approval. cc @rdblue

rdblue (approved):

rdblue:
Thanks, @singhpk234! Good to have this fixed.

By the way, I made a slight change to the spec since the `ref-name` property was in the required list, but was still `name` in the list of properties. I fixed that and I'm going to merge before CI passes because it doesn't affect tests.

singhpk234:
Apologies from my oversight here, Thanks @rdblue :) !!!

PR #4867, opened 2022-05-25
Github: Add issue form

Fokko:
This helps the reporter to submit a nicely structured bug report which helps the community to reproduce the issue and makes sure that it can be resolved quickly.

This uses the Github issue forms: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

Inspired by the awesome bug report of Airflow: https://github.com/apache/airflow/issues/new?assignees=&labels=kind%3Abug%2Carea%3Acore&template=airflow_bug_report.yml


.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+        - "Apache Spark"
+        - "Apache Flink"
+        - "Apache Hive"
+        - "Trino"
+        - "Presto"
+        - "Dremio"
+        - "Amazon Athena"
+        - "Amazon EMR"
+        - "0.9.1"
+        - "0.9.0"
```
kbendick, line 70: Question: Do we really think we should go back that far?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+        We really appreciate the community's efforts to improve Iceberg."
+  - type: dropdown
+    attributes:
+      label: Apache Iceberg version
+      description: >
+        What Apache Iceberg version are you using?
+      multiple: false
+      options:
+        - "0.13.1 (latest released)"
+        - "0.13.0"
```
kbendick, line 21: One thing that comes up _all the time_ with Spark (and similarly with Flink, but less so), is that we updated the runtime JAR names to be by version, e.g. `iceberg-spark-runtime-3.2_2.12:0.13.1`.

However, we left Spark 3.0 as the legacy name `iceberg-spark3-runtime_2.12:0.13.1`. Can we possibly ask which jar they are using?

I have wanted to _stop_ publishing `iceberg-spark3-runtime` entirely, and change that also to `iceberg-spark-runtime-3.0_2.12:0.14.0`, but a few people have mentioned that as "breaking". I personally don't see that as a problem, as the entire change is technically breaking (moving to the per version JAR release).

**TLDR** - In any case, because we use version specific jars, I feel it's appropriate to ask users which jar they're using (even if we do make the full change - though with the update it will likely be _somewhat less_ of an issue).

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: I wouldn't expect people to be still on 0.9.0, but where would be the cutoff point? Maybe only the active releases?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: I've added another field to define the version of the Query Engine. Would that be sufficient?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
kbendick: We don't even really do backports tbh at this point.

If there's a bug, we release a patch version and expect users to upgrade.

Generally speaking, our dependencies are small enough (or more specifically, well shaded) such that this has been ok.

But having people putthe version they're on is still beneficial.

I'd personally stop at 0.11.0 because anybody on anything earlier than that is.... not something I'm looking to debug. But probably not my call to make.

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
kbendick: Yeah that's definitely sufficient 👍

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,97 @@
+---
+name: Iceberg Bug report 🐞
+description: Problems, bugs and issues with Apache Iceberg
+labels: ["kind:bug"]
```
kbendick, line 4: Should we be defaulting all new issues as `kind:bug`? I would think we should be adding that when something is determined to be a bug (or possible bug). We get a lot of issues that are seeking assistance, or hoping to implement new features etc.

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,97 @@
...
+      multiple: false
+      options:
+        - "Apache Spark"
+        - "Apache Flink"
+        - "Apache Hive"
+        - "Trino"
+        - "Presto"
+        - "Dremio"
+        - "Amazon Athena"
+        - "Amazon EMR"
```
kbendick, line 69: Can we have an `Other` option and then ask them to explain in the `Version of Query Engine` text box? More and more systems integrate with Iceberg (for example Apache Drill does) and so we might want to leave that option open?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: I'll remove everything <0.11.0

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
kbendick: Looking at the Airflow issues, I see now how this would be a good label to use by default (given the user selected bug): https://github.com/apache/airflow/issues/new/choose

.github/ISSUE_TEMPLATE/iceberg_question.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: Iceberg Bug report 🐞
```
danielcweeks, line 2: Shouldn't this be Iceberg Question?

kbendick, line 2: Typo: I think this one should be question?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,35 @@
...
+      label: Query engine
+      description: Which query engine (Flink, Spark, EMR, etc) are you using? What version?
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Please describe the bug 🐞
+      description: >
+        Please describe the problem, what to expect, and how to reproduce.
+        Feel free to include stacktraces. You can include files by dragging and dropping them here.
```
nastra, line 33: maybe slightly update this to: `Feel free to include stacktraces and the Iceberg catalog configuration`

nastra (commented):
I think having an issue template is a good starting point towards the right direction. LGTM (with some nits) and we can always improve once we see what's missing from the template

.github/ISSUE_TEMPLATE/iceberg_question.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: Iceberg Question
+description: Problems, bugs and issues with Apache Iceberg
```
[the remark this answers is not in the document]
Fokko: Thanks, done!

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: Done! 👍🏻

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,36 @@
...
+labels: ["kind:bug"]
+body:
+  - type: dropdown
+    attributes:
+      label: Apache Iceberg version
+      description: >
+        What Apache Iceberg version are you using?
+      multiple: false
+      options:
+        - "0.13.1 (latest released)"
```
rdblue, line 13: 0.14.0?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,36 @@
...
+        - "0.13.0"
+        - "0.12.1"
+        - "0.12.0"
+        - "0.11.1"
+        - "0.11.0"
+        - "main (development)"
+    validations:
+      required: true
```
rdblue, line 21: Do we need to make this required? I'd probably start with optional and then see whether that's a problem. If people fill it out most of the time then that's better than nothing.

.github/ISSUE_TEMPLATE/iceberg_improvement.yml
```diff
@@ -0,0 +1,11 @@
+---
+name: Iceberg Improvement / Feature Request
+description: New features with Apache Iceberg
+labels: ["kind:new feature"]
```
rdblue, line 4: feature request?

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,36 @@
...
+  - type: textarea
```
rdblue, line 22: What about `input` instead of `textarea`? I doubt this needs to be multi-line. We could also have a dropdown if we think that might be easier, although it may have a lot of options, since there is support for quite a few that I can think of:
- Spark
- Flink
- Trino
- Starburst
- Snowflake
- Dremio
- Starrocks
- Doris
- EMR
- Athena
- PrestoDB
- Flink
- Impala
- Hive

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
```diff
@@ -0,0 +1,36 @@
...
+    attributes:
+      label: Query engine
+      description: Which query engine (Flink, Spark, EMR, etc) are you using? What version?
+    validations:
+      required: true
```
rdblue, line 27: What if this weren't required either?

.github/ISSUE_TEMPLATE/iceberg_improvement.yml
```diff
@@ -0,0 +1,11 @@
...
+body:
+  - type: textarea
+    attributes:
+      label: Feature Request / Improvement
+      description: Please describe the feature and elaborate on the use case and motivation behind it
+    validations:
+      required: true
```
rdblue, line 11: We may want an optional query engine input here as well, since it could be for a specific engine.

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: Agree, made it optional.

.github/ISSUE_TEMPLATE/iceberg_improvement.yml
Fokko: I like it, updated!

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: You got Flink twice ;) I've updated it to a list and also added Other (also made it optional to keep the bar low).

.github/ISSUE_TEMPLATE/iceberg_bug_report.yml
Fokko: Yes, less red * 👍🏻

.github/ISSUE_TEMPLATE/iceberg_improvement.yml
Fokko: I like it, thanks!

rdblue (approved):

rdblue:
Merged! Thanks for getting this done, @Fokko!

I didn't wait for CI because this doesn't change any code.

PR #4868, opened 2022-05-25
Uniform partition field names generation

szlta:
We have two different ways of generating the partition field column name:
https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/BaseUpdatePartitionSpec.java#L460 generates data_bucket_16, while
https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L484 generates data_bucket.

This PR tries to unify this, so that the former method would be applicable everywhere.
Related PR: [4662](https://github.com/apache/iceberg/pull/4662)

### DIFF
--- a/api/src/main/java/org/apache/iceberg/transforms/PartitionSpecVisitor.java
+++ b/api/src/main/java/org/apache/iceberg/transforms/PartitionSpecVisitor.java
@@ -86,6 +86,10 @@ public interface PartitionSpecVisitor<T> {
     throw new UnsupportedOperationException("Void transform is not supported");
   }
 
+  default T alwaysNull(String sourceName, int sourceId) {
+    throw new UnsupportedOperationException("Void transform is not supported");
+  }
+
   default T unknown(int fieldId, String sourceName, int sourceId, String transform) {
     throw new UnsupportedOperationException(String.format("Unknown transform %s is not supported", transform));
   }
--- a/api/src/main/java/org/apache/iceberg/PartitionSpec.java
+++ b/api/src/main/java/org/apache/iceberg/PartitionSpec.java
@@ -36,6 +36,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Multimaps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.transforms.PartitionNameGenerator;
 import org.apache.iceberg.transforms.Transform;
 import org.apache.iceberg.transforms.Transforms;
 import org.apache.iceberg.transforms.UnknownTransform;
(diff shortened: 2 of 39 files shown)


api/src/main/java/org/apache/iceberg/transforms/PartitionSpecVisitor.java
```diff
@@ -86,6 +86,10 @@ default T alwaysNull(int fieldId, String sourceName, int sourceId) {
     throw new UnsupportedOperationException("Void transform is not supported");
   }
 
+  default T alwaysNull(String sourceName, int sourceId) {
+    throw new UnsupportedOperationException("Void transform is not supported");
```
kbendick, line 90: Nit: Copy paste error? This is not the Void transform. Should be `Always null transform is not supoprted` (if needed at all).

kbendick (commented):
Hi @szlta.

I'm still ramping up on previous discussion etc, but have we identified actual cases (e.g. when using an engine) that one might get `xxxx_bucket_n` AND `xxx_bucket`? It's possible that there is a function that exists but isn't really used in a user-facing way (by which I mean used by an engine, preferably amongst the best supported / tested engines).

My immediate concern is that if an engine is currently generating columns in one style, and begins to generate them in another, this could be a _very_ breaking change for people's tables.

Additionally, if we append the number of buckets / truncation width at the end of the generated column name, are we opening ourselves up to the possibility that users will be able to partition twice on the same column using the same transform (e.g. bucket_data_6 and bucket_data_8 within the same partition spec)?

I also have some minor style notes, but more importantly I'd like to discuss:
- what we see in practice (as opposed to just reading through the code or even in unit tests, some of which might be somewhat old).
- any places we might be depending on the naming at present that would cause a breaking change for people who use Iceberg in the most common ways (i.e. via Spark, Flink, Trino, Dremio, etc).

Is there a discussion somewhere of that or could you possibly provide an example so we're all on the same page?

api/src/main/java/org/apache/iceberg/transforms/PartitionSpecVisitor.java
szlta: Nope this was meant to be intentional. Each method here has a pair that doesn't take a `int fieldId`, and this was missing for the `alwaysNull` method.
I believe this is actually the void transform, which other method should it be if not this one?

rdblue:
I don't see much benefit to making this change. What is the underlying problem you're trying to solve? Changing 40 files just to make a slight change to field names in metadata tables doesn't seem worth it.

szlta:
Hi @rdblue, @kbendick - could you share your opinion on the above please?

rdblue:
@szlta, I appreciate your thoroughness and that you're aiming to make this less confusing overall. I think this is probably better off the way it is. I don't think that it is very confusing to have slightly different names generated in different situations and I don't think this large of a change is worth it.

One thing that I do think warrants attention is the reuse of old partition fields. If there's a partition spec with an equivalent field, then we should bring it back rather than replacing it. That helps quite a bit more than more uniform names, I think.

szlta:
@rdblue I can accept this reasoning. I just hope we won't have to do this change in the future either, but thanks for sharing your insights.
I have amended my original change at https://github.com/apache/iceberg/pull/4662/ so that this name generation discrepancy would not cause a test failure there while reusing old partition fields.. Kindly take a look on that PR, while I'll go ahead and close this. Thanks!

PR #4873, opened 2022-05-25
API: Pass FileIO into Snapshot methods that read metadata

kbendick:
# Motivation
As part of the REST Catalog, and for interoperability with other systems, we need to be able to accept a `TableMetadata` instance as JSON from an external source.

Currently, `TableMetadata` takes in `FileIO` as an argument, which it pushes down to `BaseSnapshot` (the main implementation of `Snapshot`).

File IO is used lazily in `BaseSnapshot`, only when we need to fetch added / deleted files or any kind of manifest from the snapshot. The call for these changes is done once, and then the results are cached.

# Changes
Because we cannot be sure of which FileIO is the appropriate one to use when parsing `TableMetadata`, and even if we were sure, it would be very difficult to pass it to the `TableMetadataDeserializer`, this PR deprecates the stored `FileIO` state from `BaseSnapshot`.

With these changes, users will need to call `addedFiles` / `deletedFiles` / `allManifests` / `dataManifests` / `deleteManifests` on `Snapshot` with the desired FileIO implementation.

This is relatively easy to do, as the correct `FileIO` tends to be readily available where these functions are called. This also makes it much easier for session based use cases, which might have different FileIO with different levels of access, within the same program, such as multi-tenant Trino-style programs.

# Remaining
This PR only updates latest Spark and Flink and Hive versions. We'll need to backport for earlier versions, which can either be done in this PR or in another PR. But for the sake of reviewers, I've left it to just the latest versions of all.

cc @rdblue


core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -84,6 +89,25 @@ class BaseSnapshot implements Snapshot {
...
+    this.io = null;
+    this.sequenceNumber = sequenceNumber;
+    this.snapshotId = snapshotId;
+    this.parentId = parentId;
+    this.timestampMillis = timestampMillis;
+    this.operation = operation;
+    this.summary = summary;
+    this.schemaId = schemaId;
+    this.manifestListLocation = manifestList;
+  }
```
kbendick, line 109: This constructor is not presently used. In most situations, it's ok to simply pass `null` as the FileIO.

I didn't update the specific constructor used in an attempt to make this easier to cherry-pick.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -150,14 +192,40 @@ private void cacheManifests() {
...
+  /**
+   * @deprecated since 1.0.0 - Use {@link Snapshot#allManifests(FileIO)} instead.
+   */
+  @Override
+  @Deprecated
   public List<ManifestFile> allManifests() {
+    // TODO - Consider throwing here if allManifests are not already cached or if called at all.
     if (allManifests == null) {
       cacheManifests();
     }
```
kbendick, line 211: @rdblue and I discussed leaving this not deprecated, but throwing if the function _needs_ to call `cacheManifests()`.

I'd personally like to see the older signatures removed prior to 1.0, but given we're not doing a 0.14.0 release that might not be possible.

However, we should be doing as much as possible to discourage people from using the function signatures that don't require `FileIO`. It might be ok to keep them for backwards compatibility, provided that `cacheManifests` and `cacheChanges` have already been called.

kbendick (commented):
cc @nastra as `TestNessieTable` has been touched. Additionally, I believe you were previously working around this problem (not having `FileIO` on hand when parsing `TableMetadata` from external JSON).

Hopefully this can help simplify some changes for you in the medium to long term 🙂

kbendick:
Also cc @danielcweeks

api/src/main/java/org/apache/iceberg/Snapshot.java
```diff
@@ -68,23 +69,58 @@ public interface Snapshot extends Serializable {
    * Return all {@link ManifestFile} instances for either data or delete manifests in this snapshot.
    *
    * @return a list of ManifestFile
+   * @deprecated since 1.0.0 - Use {@link Snapshot#allManifests(FileIO)} instead.
```
rdblue, line 72: Instead of saying when it was deprecated, I think we should say when it will be removed. I'm thinking that will probably be 1.0.0, assuming that we do an 0.14.0 and then follow up with a 1.0.0 to remove deprecations, but open to other opinions.

api/src/main/java/org/apache/iceberg/Snapshot.java
```diff
@@ -68,23 +69,58 @@ public interface Snapshot extends Serializable {
...
    */
+  @Deprecated
   List<ManifestFile> allManifests();
 
+  /**
+   * Return all {@link ManifestFile} instances for either data or delete manifests in this snapshot.
+   * <p>
+   * The FileIO instance is passed in for cases where the correct FileIO to use is not known when the
+   * TableMetadata (and thus BaseSnapshot implementations) is not known, such as parsing JSON for the
+   * RESTCatalog or for highly sessionzed applications like Trino or shared Spark servers.
```
rdblue, line 82: I don't think this paragraph is needed. The FileIO is passed in because it doesn't make sense for `Snapshot` to have one. The reason why we didn't require it before is because we originally kept manifests in the Snapshot metadata itself.

api/src/main/java/org/apache/iceberg/Snapshot.java
```diff
@@ -68,23 +69,58 @@ public interface Snapshot extends Serializable {
...
+   *
+   * @param io the {@link FileIO} instance to be used for reading Snapshot manifests, if not already cached
```
rdblue, line 84: How about `a FileIO instance for reading files in object storage`?

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -131,6 +155,24 @@ public Integer schemaId() {
     return schemaId;
   }
 
+  private void cacheManifests(FileIO fileIO) {
+    if (fileIO == null) {
+      throw new IllegalStateException("Cannot cache changes: FileIO is null");
```
rdblue, line 160: I think this is now an `IllegalArgumentException`.

api/src/main/java/org/apache/iceberg/Snapshot.java
kbendick: Yeah I think that's a good idea. I believe Yufei did something similar (noting when it would be deprecated) when deprecating the old metadata write directory configurations.

I also think that this gives a little bit more of an argument for a 0.14.0. We can go through, determine what needs to be deprecated, cut a release, and then turn around and cut a 1.0.0 release that from then on has stronger API / ABI guarantees.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -150,14 +192,40 @@ private void cacheManifests() {
...
+    return allManifests;
+  }
+
+  /**
+   * @deprecated since 1.0.0 - Use {@link Snapshot#allManifests(FileIO)} instead.
+   */
+  @Override
+  @Deprecated
   public List<ManifestFile> allManifests() {
+    // TODO - Consider throwing here if allManifests are not already cached or if called at all.
```
rdblue, line 208: It's okay for callers to use this for now. It should behave as it always has.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
rdblue: I think I'm okay with deprecating this.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -194,16 +298,51 @@ public String manifestListLocation() {
...
+    }
+
+    this.cachedAdds = adds.build();
+    this.cachedDeletes = deletes.build();
+  }
+
   private void cacheChanges() {
     if (io == null) {
-      throw new IllegalStateException("Cannot cache changes: FileIO is null");
+      throw new IllegalStateException("Cannot cache changes: file io is null");
```
rdblue, line 338: Do we need to change this? It doesn't really seem worth it.

api/src/main/java/org/apache/iceberg/Snapshot.java
kbendick: I'll go with `a FileIO instance for reading files from storage` so as not to confuse any potential HDFS users. Trivial difference but let me know if you disagree

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -194,16 +298,51 @@ public String manifestListLocation() {
     return manifestListLocation;
   }
 
+  private void cacheChanges(FileIO fileIO) {
```
rdblue, line 301: As with `cacheManifests`, can we combine the implementations so that it is clear there are no changes other than passing in `fileIO`?

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
kbendick: The existing `cacheManifest` function uses `IllegalStateException`, but I agree that `IllegalArgumentException` makes more sense here as we're not relying on state within the class but arguments.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -131,6 +155,24 @@ public Integer schemaId() {
     return schemaId;
   }
 
+  private void cacheManifests(FileIO fileIO) {
```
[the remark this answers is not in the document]
kbendick: Updated to use one `cacheManifests` function - passing in the class level `io` for the existing (deprecated) functions that call it.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
kbendick: Nah it doesn't. Agreed.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
kbendick: Done.

kbendick: I think changing the signature is better in case people are using this method in a fork somehow. They'll be more likely to notice the change and update their (additional) code as needed.

The likelihood somebody is using this particular `private` method in some unique way though is likely pretty low in my opinion.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
kbendick: Yes same. I left a note that it's deprecated and will be removed in 1.0.0.

api/src/main/java/org/apache/iceberg/Snapshot.java
kbendick: Copied existing language and mentioned it's deprecated since 0.14.0 and will be deprecated in 1.0.0.

I found a deprecation notice saying something would be removed in 0.15.0, so I think some ambiguity about 0.14.0 vs 1.0.0 is fine for now.

api/src/main/java/org/apache/iceberg/Snapshot.java
kbendick: Removed.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
kbendick: Done.

rdblue (approved):

rdblue:
Looks great! Thanks @kbendick. It's great to have this cleaned up. I've been wanting to do it for a long time now.

kbendick:
> Looks great! Thanks @kbendick. It's great to have this cleaned up. I've been wanting to do it for a long time now.

Thanks @rdblue. This makes a lot of sense to me as well, as the TableMetadata should be transmissible in a way that doesn’t immediately couple itself to the FileSystem (or choice of FileIO etc).

I know it seems like a large change but it’s really mostly just tests and the FileIO is already properly serialized for Kryo etc where need be. So the change isn’t nearly as large as it seems.

**Question**: Should I update tests and code for earlier Spark versions and Flink versions in this PR or in another? It would make code maintenance simpler to keep other versions in sync (either here or in another PR). But happy to slice the PRs however is deemed best.

api/src/main/java/org/apache/iceberg/Snapshot.java
kbendick: Also I’ll make a note of the 0.15.0 thing and we’ll get that cleared up. Found it basically git grep-Ing and it’s very unlikely we’ll have a 0.15.0.

rdblue:
> Should I update tests and code for earlier Spark versions and Flink versions in this PR or in another?

I think in another PR would be best.

rdblue:
Thanks, @kbendick! I merged this so we can move on to the follow-ups.

kbendick:
> @kbendick thanks, I'll review on monday but if anything comes up we could just do a follow-up, so no need to block on merging this.

Cool. Let me know if there's anything in particular you'd like to go over or have any questions on as I believe this will benefit you as well.

nastra:
retro +1

PR #4877, opened 2022-05-26
API: Pass FileIO into Snapshot methods that read metadata - Backport

kbendick:
This backports https://github.com/apache/iceberg/pull/4873 to Spark 2.4, 3.0, and 3.1 as well as Flink 1.13 and 1.14.

The only usage of the deprecated functions that remains is the call to [`snapshot.allManifests()` in SnapshotParser at line 86](https://github.com/apache/iceberg/blob/128d5a161fda076118a7cab1d95ab5064400e08a/core/src/main/java/org/apache/iceberg/SnapshotParser.java#L80-L90), which we explicitly chose to leave in for now as there should be very few tables that do not already have the manifest list in them (and are storing manifests instead).

cc @rdblue @nastra @danielcweeks

### DIFF
--- a/spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteManifestsSparkAction.java
+++ b/spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteManifestsSparkAction.java
@@ -261,7 +261,7 @@ public class BaseRewriteManifestsSparkAction
       return ImmutableList.of();
     }
 
-    return currentSnapshot.dataManifests().stream()
+    return currentSnapshot.dataManifests(table.io()).stream()
         .filter(manifest -> manifest.partitionSpecId() == spec.specId() && predicate.test(manifest))
         .collect(Collectors.toList());
   }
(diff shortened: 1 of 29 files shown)


spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteManifestsSparkAction.java
```diff
@@ -261,7 +261,7 @@ private List<ManifestFile> findMatchingManifests() {
       return ImmutableList.of();
     }
 
-    return currentSnapshot.dataManifests().stream()
+    return currentSnapshot.dataManifests(fileIO).stream()
```
kbendick, line 264: For this usage, there's a class level `fileIO` that is wrapped to avoid Kryo serialization issues. So we use that instead of `table..io()`.

kbendick:
It's worth noting that this changes only 4 lines of code that are in source files. The rest is in test files.

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #4878, opened 2022-05-27
API: Cleanup SnapshotRef to avoid unnecessary boxing in equality check by comparing snapshotID via ==

amogh-jahagirdar:
Cleanup SnapshotRef to avoid unnecessary boxing in equality check by comparing snapshotID via ==

### DIFF
--- a/api/src/main/java/org/apache/iceberg/SnapshotRef.java
+++ b/api/src/main/java/org/apache/iceberg/SnapshotRef.java
@@ -86,7 +86,7 @@ public class SnapshotRef implements Serializable {
     }
 
     SnapshotRef ref = (SnapshotRef) other;
-    return Objects.equals(ref.snapshotId(), snapshotId) &&
+    return ref.snapshotId == snapshotId &&
         Objects.equals(ref.type(), type) &&
         Objects.equals(ref.maxRefAgeMs(), maxRefAgeMs) &&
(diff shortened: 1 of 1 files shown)


kbendick:
I believe there are other places that do the same equality check. Want to handle them or leave that as a follow up issue?

amogh-jahagirdar:
> I believe there are other places that do the same equality check. Want to handle them or leave that as a follow up issue?

Good point! This case was just something I noticed in the warning logs when building, let me double check if there are any others. If there are, I think I may as well tackle them here!

amogh-jahagirdar:
As far as the error prone https://errorprone.info/bugpattern/ObjectEqualsForPrimitives check goes I'm not seeing any other cases when I compile, so this seems to be the only place where this applies. Let me know if I'm missing some cases @kbendick!

jackye1995 (approved):
yeah I remember I did a full pass of this and did not find any other places with this issue when creating this task in our internal backlog, so I assume it is still the case now. Let's see if new things pop up.

kbendick:
Thanks @amogh-jahagirdar and @jackye1995! If we find other places then we can update them later. Not a high priority issue.

I think there's some code in `MetadataUpdate` that uses `SnapshotRef`, but also with a `name` field and that's where it's copied (it came to me in the night). So I'll check and update it if so.

More then the performance gains, I think the use of primitives and treating them as such signifies something to the reader (especially given we don't tend to use `Optional`).

PR #4880, opened 2022-05-27
Core - Use unbound variants of SortOrder / PartitionSpec for REST Catalog's CreateTableRequest

kbendick:
As I've been working on some serialization related tests for the REST catalog, it came to my attention that we should be storing the create table requests sort order and partition spec as `UnboundSortOrder` and `UnboundPartitionSpec`, respectively.

Generally speaking, we are moving towards using `Unbound` variants of all things whenever possible, and letting the binding to actual fields take place later in the process.

I am working on some PRs that add additional serialization tests in, but I wanted to get these changes in sooner so that anybody who might be implementing or testing agains the RESTCatalog would have these changes available to them.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/requests/CreateTableRequest.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/CreateTableRequest.java
@@ -24,6 +24,8 @@ import java.util.Objects;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SortOrder;
+import org.apache.iceberg.UnboundPartitionSpec;
+import org.apache.iceberg.UnboundSortOrder;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/iceberg/rest/requests/CreateTableRequest.java
```diff
@@ -38,8 +40,8 @@ public class CreateTableRequest implements RESTRequest {
   private String name;
   private String location;
   private Schema schema;
-  private PartitionSpec spec;
-  private SortOrder order;
+  private UnboundPartitionSpec spec;
+  private UnboundSortOrder order;
```
rdblue, line 44: We need to make sure we follow up with a serialization test for this class.

rdblue (approved):
+1 to unblock people testing with this, but I really would like to get a serialization test for it soon!

rdblue:
Thanks, @kbendick!

PR #4885, opened 2022-05-27
Docs: Fix a typo. Add using for writeTo when creating a table.

wypoon:
Fix two issues in Spark documentation.
1. Typo in Spark DDL doc.
2. Examples for creating/replacing a table using `writeTo` need `using("iceberg")`.

### DIFF
--- a/docs/spark/spark-writes.md
+++ b/docs/spark/spark-writes.md
@@ -287,14 +287,20 @@ To run a CTAS or RTAS, use `create`, `replace`, or `createOrReplace` operations:
 
 ```scala
 val data: DataFrame = ...
-data.writeTo("prod.db.table").using("iceberg").create()
+data.writeTo("prod.db.table").create()
+```
+
+If you have replaced the default Spark catalog (`spark_catalog`) with Iceberg's `SparkSessionCatalog`, do:
+
+```scala
+val data: DataFrame = ...
+data.writeTo("db.table").using("iceberg").create()
 ```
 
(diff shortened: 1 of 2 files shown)


kbendick (approved):
Good catch.

singhpk234 (approved):

docs/spark/spark-writes.md
```diff
@@ -287,13 +287,14 @@ To run a CTAS or RTAS, use `create`, `replace`, or `createOrReplace` operations:
 
 ```scala
 val data: DataFrame = ...
-data.writeTo("prod.db.table").create()
+data.writeTo("prod.db.table").using("iceberg").create()
```
rdblue, line 290: This depends on the catalog. If you're using an Iceberg catalog, it shouldn't be necessary to do this.

wypoon: Thanks for pointing that out. I did not actually know that. I usually use the `SparkSessionCatalog`.

rdblue: For the `SparkSessionCatalog`, you do need this. It might be better just to recommend it here, although it isn't needed unless we are trying to determine whether to pass the create through to the underlying catalog.

rdblue (approved):

rdblue:
Looks great. Thanks for updating this, @wypoon!

docs/spark/spark-writes.md
wypoon: I reverted the change and added a clarification for when you use `SparkSessionCatalog`. Please let me know if you think it is better now.

PR #4886, opened 2022-05-27
Core: Address Anton's comments from the PR 4744 (incremental append scan impl)

stevenzwu:


stevenzwu:
@aokolnychyi 

For the [boolean flag comment](https://github.com/apache/iceberg/pull/4744#discussion_r883936190), do you think a builder class for `IncrementalScanEvent` would be better?

For the [helper methods comment](https://github.com/apache/iceberg/pull/4744#discussion_r883943057), I feel we are adding a bunch of methods without making the code much simpler. 

For the [schema evolution comment](https://github.com/apache/iceberg/pull/4744#discussion_r883912168), I think it is outside the scope of streaming source.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -131,6 +126,17 @@ public CloseableIterable<CombinedScanTask> planTasks() {
     return TableScanUtil.planTasks(splitFiles, targetSplitSize(), splitLookback(), splitOpenFileCost());
   }
 
+  private long toSnapshotIdExclusive() {
```
rdblue, line 129: Isn't `toSnapshotId` always inclusive, not exclusive?

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -71,46 +71,41 @@ public IncrementalAppendScan toSnapshot(long toSnapshotId) {
...
+      // simply return an empty iterable. In this case, listener notification is also skipped.
       return CloseableIterable.empty();
     }
 
+    long toSnapshotIdInclusive = toSnapshotIdExclusive();
     // fromSnapshotIdExclusive can be null. appendsBetween handles null fromSnapshotIdExclusive properly
     // by finding the oldest ancestor of end snapshot.
-    Long fromSnapshotIdExclusive = context().fromSnapshotId();
-    if (context().fromSnapshotId() != null) {
+    Long fromSnapshotIdExclusive = fromSnapshotId;
```
rdblue, line 85: If `fromSnapshotId` may be inclusive, then it should never be assigned to `fromSnapshotIdExclusive` without first checking. Otherwise, there is temporarily a value for `fromSnapshotIdExclusive` that is incorrect. If you have a guarantee in the variable name, you should never violate that guarantee because someone might not know about it later.

core/src/main/java/org/apache/iceberg/BaseScan.java
```diff
@@ -59,8 +59,8 @@ protected TableScanContext context() {
     return context;
   }
 
-  protected abstract T newRefinedScan(
-      TableOperations newOps, Table newTable, Schema newSchema, TableScanContext newContext);
+  protected abstract T newRefinedScan(TableOperations newOps, Table newTable, Schema newSchema,
+                                      TableScanContext newContext);
```
rdblue, line 63: I don't think these reformatting changes are worth it, since the alternative is already in master.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -132,7 +132,7 @@ public static Snapshot oldestAncestor(Table table) {
    * @param lookup lookup function from snapshot ID to snapshot
    * @return null if there is no current snapshot in the table, else the oldest Snapshot.
    */
-  public static Snapshot oldestAncestor(long snapshotId, Function<Long, Snapshot> lookup) {
+  public static Snapshot oldestAncestorOf(long snapshotId, Function<Long, Snapshot> lookup) {
```
rdblue, line 135: Didn't this exist before the other commit? If this has been released, I doubt it is a good idea to rename it just to add "Of".

stevenzwu: this was added by the other PR #4744 (incremental append scan impl) that just merged. It is not released and should be safe to rename

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: good catch. I meant inclusive. somehow typed exclusive

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
stevenzwu: good point. will update

core/src/main/java/org/apache/iceberg/BaseScan.java
stevenzwu: ok. will revert

core/src/main/java/org/apache/iceberg/TableScanContext.java
```diff
@@ -169,7 +169,7 @@ Long fromSnapshotId() {
     return fromSnapshotId;
   }
 
-  TableScanContext fromSnapshotId(long id) {
+  TableScanContext fromSnapshotIdExclusive(long id) {
     return new TableScanContext(snapshotId, rowFilter, ignoreResiduals,
         caseSensitive, colStats, projectedSchema, selectedColumns, options, id, toSnapshotId,
         planExecutor, fromSnapshotInclusive);
```
aokolnychyi, line 175: nit: I think we should probably pass `false` explicitly instead of using `fromSnapshotInclusive`.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
aokolnychyi: It was just added and hasn't been released. All other methods in this class follow this name pattern.

aokolnychyi (approved):
Thanks, @stevenzwu! I had one question in `TableScanContext` but looks good to me otherwise.

core/src/main/java/org/apache/iceberg/TableScanContext.java
stevenzwu: agree. explicit `false` is more clear.

PR #4888, opened 2022-05-28
Core: Support _deleted metadata column in vectorized read

flyrain:
The vectorized version of #4683.

cc @aokolnychyi @szehon-ho @RussellSpitzer @chenjunjiedada @stevenzwu @Reo-LEI @hameizi @singhpk234 @rajarshisarkar @kbendick @rdblue

Benchmarks for pos delete and eq delete
Screen Shot 2022-06-17 at 2 28 40 PM
Screen Shot 2022-06-17 at 2 30 03 PM

### DIFF
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
@@ -104,6 +104,10 @@ public class VectorHolder {
     return new ConstantVectorHolder(numRows, constantValue);
   }
 
+  public static <T> VectorHolder isDeletedHolder(int numRows) {
+    return new IsDeletedVectorHolder(numRows);
+  }
+
   public static VectorHolder dummyHolder(int numRows) {
     return new ConstantVectorHolder(numRows);
   }
@@ -143,6 +147,19 @@ public class VectorHolder {
   public static class PositionVectorHolder extends VectorHolder {
     public PositionVectorHolder(FieldVector vector, Type type, NullabilityHolder nulls) {
       super(vector, type, nulls);
+    }
+  }
+
+  public static class IsDeletedVectorHolder extends VectorHolder {
+    private final int numRows;
+
+    public IsDeletedVectorHolder(int numRows) {
+      this.numRows = numRows;
+    }
+
+    @Override
+    public int numValues() {
+      return numRows;
     }
   }
 
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
@@ -104,8 +104,8 @@ public class VectorHolder {
     return new ConstantVectorHolder(numRows, constantValue);
   }
 
-  public static <T> VectorHolder isDeletedHolder(int numRows) {
-    return new IsDeletedVectorHolder(numRows);
+  public static VectorHolder deletedVectorHolder(int numRows) {
+    return new DeletedVectorHolder(numRows);
   }
 
   public static VectorHolder dummyHolder(int numRows) {
@@ -150,10 +150,10 @@ public class VectorHolder {
     }
   }
 
-  public static class IsDeletedVectorHolder extends VectorHolder {
+  public static class DeletedVectorHolder extends VectorHolder {
     private final int numRows;
 
-    public IsDeletedVectorHolder(int numRows) {
+    public DeletedVectorHolder(int numRows) {
       this.numRows = numRows;
     }
 
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
@@ -517,5 +517,32 @@ public class VectorizedArrowReader implements VectorizedReader<VectorHolder> {
     }
   }
 
+  /**
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns a dummy
+   * VectorHolder which indicates whether the row is deleted.
+   */
+  public static class DeletedVectorReader extends VectorizedArrowReader {
+    public DeletedVectorReader() {
+    }
+
+    @Override
+    public VectorHolder read(VectorHolder reuse, int numValsToRead) {
+      return VectorHolder.isDeletedHolder(numValsToRead);
+    }
+
+    @Override
+    public void setRowGroupInfo(PageReadStore source, Map<ColumnPath, ColumnChunkMetaData> metadata, long rowPosition) {
+    }
+
+    @Override
+    public String toString() {
+      return "DeletedVectorReader";
+    }
+
+    @Override
+    public void setBatchSize(int batchSize) {
+    }
+  }
+
 }
 
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
@@ -518,8 +518,8 @@ public class VectorizedArrowReader implements VectorizedReader<VectorHolder> {
   }
 
   /**
-   * A Dummy Vector Reader which doesn't actually read files, instead it returns a dummy
-   * VectorHolder which indicates whether the row is deleted.
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns an
+   * IsDeleted Vector Holder which indicates whether a given row is deleted.
    */
   public static class DeletedVectorReader extends VectorizedArrowReader {
     public DeletedVectorReader() {
--- a/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
+++ b/arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
@@ -518,8 +518,8 @@ public class VectorizedArrowReader implements VectorizedReader<VectorHolder> {
   }
 
   /**
-   * A Dummy Vector Reader which doesn't actually read files, instead it returns an
-   * IsDeleted Vector Holder which indicates whether a given row is deleted.
+   * A Dummy Vector Reader which doesn't actually read files. Instead, it returns a
+   * Deleted Vector Holder which indicates whether a given row is deleted.
    */
   public static class DeletedVectorReader extends VectorizedArrowReader {
     public DeletedVectorReader() {
@@ -527,7 +527,7 @@ public class VectorizedArrowReader implements VectorizedReader<VectorHolder> {
 
     @Override
     public VectorHolder read(VectorHolder reuse, int numValsToRead) {
-      return VectorHolder.isDeletedHolder(numValsToRead);
+      return VectorHolder.deletedVectorHolder(numValsToRead);
     }
 
     @Override
--- a/spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
+++ b/spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
@@ -70,8 +70,7 @@ public class TestSparkParquetReadMetadataColumns {
   private static final Schema PROJECTION_SCHEMA = new Schema(
       required(100, "id", Types.LongType.get()),
       required(101, "data", Types.StringType.get()),
-      MetadataColumns.ROW_POSITION,
-      MetadataColumns.IS_DELETED
+      MetadataColumns.ROW_POSITION
   );
 
   private static final int NUM_ROWS = 1000;
@@ -104,7 +103,6 @@ public class TestSparkParquetReadMetadataColumns {
       }
       row.update(1, UTF8String.fromString("str" + i));
       row.update(2, i);
-      row.update(3, false);
       EXPECTED_ROWS.add(row);
     }
   }
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.data.vectorized;
+
+import org.apache.iceberg.arrow.vectorized.VectorHolder;
+import org.apache.iceberg.arrow.vectorized.VectorHolder.ConstantVectorHolder;
+import org.apache.iceberg.arrow.vectorized.VectorHolder.IsDeletedVectorHolder;
+import org.apache.iceberg.types.Types;
+import org.apache.spark.sql.vectorized.ColumnVector;
+
+public class ColumnVectorBuilder {
+  private final VectorHolder holder;
+  private final int numRows;
+
+  private boolean[] isDeleted;
+  private int[] rowIdMapping;
+
+  public ColumnVectorBuilder(VectorHolder holder, int numRows) {
+    this.holder = holder;
+    this.numRows = numRows;
+  }
+
+  public ColumnVectorBuilder withDeletedRows(int[] rowIdMappingArray, boolean[] isDeletedArray) {
+    this.rowIdMapping = rowIdMappingArray;
+    this.isDeleted = isDeletedArray;
+    return this;
+  }
+
+  public ColumnVector build() {
+    if (holder.isDummy()) {
+      if (holder instanceof IsDeletedVectorHolder) {
+        return new DeletedMetaColumnVector(Types.BooleanType.get(), isDeleted);
+      } else if (holder instanceof ConstantVectorHolder) {
+        return new ConstantColumnVector(Types.IntegerType.get(), numRows,
+            ((ConstantVectorHolder) holder).getConstant());
+      } else {
+        throw new IllegalStateException("Unknown dummy vector holder: " + holder);
+      }
+    } else if (rowIdMapping != null) {
+      return new ColumnVectorWithFilter(holder, rowIdMapping);
+    } else {
+      return new IcebergArrowColumnVector(holder);
+    }
+  }
+}
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
@@ -21,21 +21,12 @@ package org.apache.iceberg.spark.data.vectorized;
 
 import org.apache.iceberg.arrow.vectorized.VectorHolder;
 import org.apache.iceberg.arrow.vectorized.VectorHolder.ConstantVectorHolder;
-import org.apache.iceberg.arrow.vectorized.VectorHolder.IsDeletedVectorHolder;
 import org.apache.iceberg.types.Types;
 import org.apache.spark.sql.vectorized.ColumnVector;
 
-public class ColumnVectorBuilder {
-  private final VectorHolder holder;
-  private final int numRows;
-
+class ColumnVectorBuilder {
   private boolean[] isDeleted;
   private int[] rowIdMapping;
-
-  public ColumnVectorBuilder(VectorHolder holder, int numRows) {
-    this.holder = holder;
-    this.numRows = numRows;
-  }
 
   public ColumnVectorBuilder withDeletedRows(int[] rowIdMappingArray, boolean[] isDeletedArray) {
     this.rowIdMapping = rowIdMappingArray;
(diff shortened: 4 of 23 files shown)


spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
```diff
@@ -70,8 +70,7 @@ public class TestSparkParquetReadMetadataColumns {
   private static final Schema PROJECTION_SCHEMA = new Schema(
       required(100, "id", Types.LongType.get()),
       required(101, "data", Types.StringType.get()),
-      MetadataColumns.ROW_POSITION,
-      MetadataColumns.IS_DELETED
+      MetadataColumns.ROW_POSITION
```
flyrain, line 73: We need this change since VectorizedReaderBuilder is shared by all spark versions. The change in line 94 of VectorizedReaderBuilder changes the type of the reader as the following code shows. Then, the read throws exception in the method `IcebergArrowColumnVector.forHolder()` of the old Spark version. This change should be fine, since the old Spark doesn't really support _deleted metadata column.
```
        reorderedFields.add(new VectorizedArrowReader.DeletedVectorReader());
```

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
```diff
@@ -104,6 +104,10 @@ public static <T> VectorHolder constantHolder(int numRows, T constantValue) {
     return new ConstantVectorHolder(numRows, constantValue);
   }
 
+  public static <T> VectorHolder isDeletedHolder(int numRows) {
```
RussellSpitzer, line 107: This is another kinda on the fence for me, while the return type isn't boolean, this does seem like a boolean method. Maybe it should just be deletedHolder ? Just skip the "is" since it's a bit confusing in this context?

RussellSpitzer: I think the class name is fine, just this method seems a little confusing to me, but maybe it's just me :)

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
```diff
@@ -517,5 +517,32 @@ public void setBatchSize(int batchSize) {
     }
   }
 
+  /**
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns a dummy
```
RussellSpitzer, line 521: Just a little note here, I think you don't have to say "dummy" twice. 

Small suggestion?
"A Vector Reader that returns a IsDeleted Vector Holder which returns true when a given row is deleted when read"?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -209,7 +231,32 @@ void applyEqDelete() {
...
+      return currentRowId;
+    }
+
+    /**
+     * Convert the row id mapping array to the isDeleted array.
+     *
+     * @param numRowsInRowIdMapping the num of rows in the row id mapping array
+     */
+    void rowIdMappingToIsDeleted(int numRowsInRowIdMapping) {
+      if (isDeleted == null || rowIdMapping == null) {
```
RussellSpitzer, line 244: Is this ok to return on? Seems like we should fail if isDeleted is set but rowIdMapping is null? Or is is rowIdMapping null if there are no delete files?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -209,7 +231,32 @@ void applyEqDelete() {
...
+        return;
+      }
+
+      for (int i = 0; i < numRowsToRead; i++) {
```
RussellSpitzer, line 248: Arrays.fill(isDeleted, true) ?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -209,7 +231,32 @@ void applyEqDelete() {
...
+        isDeleted[i] = true;
+      }
+
+      for (int i = 0; i < numRowsInRowIdMapping; i++) {
+        isDeleted[rowIdMapping[i]] = false;
+      }
+
+      // reset the row id mapping array, so that it doesn't filter out the deleted rows
+      for (int i = 0; i < numRowsToRead; i++) {
```
RussellSpitzer, line 257: Could we just set rowIdMapping to null? Isn't that an allowed state here?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
RussellSpitzer: Oh wait I think I see, so if rowIdMapping is null it means there are no deletes. Does that mean we treat isDeleted as always false if null? I think that may need to be a javadoc comment?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/DeletedMetaColumnVector.java
```diff
@@ -0,0 +1,123 @@
...
+    return false;
+  }
+
+  @Override
+  public boolean getBoolean(int rowId) {
+    return isDeleted[rowId];
+  }
+
+  @Override
+  public byte getByte(int rowId) {
```
RussellSpitzer, line 65: Not sure if we did this in the others, but IMHO all the accessors should through `UnsupportedOperationException` except for `getBoolean`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -72,31 +84,44 @@ public final ColumnarBatch read(ColumnarBatch reuse, int numRowsToRead) {
   }
 
   private class ColumnBatchLoader {
-    private int[] rowIdMapping; // the rowId mapping to skip deleted rows for all column vectors inside a batch
-    private int numRows;
     private ColumnarBatch columnarBatch;
+    private final int numRowsToRead;
+    private int[] rowIdMapping; // the rowId mapping to skip deleted rows for all column vectors inside a batch
+    private boolean[] isDeleted; // the array to indicate if a row is deleted or not
```
RussellSpitzer, line 90: For my confusion below can we indicate here for these two arrays to describe when these two can be null?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
```diff
@@ -173,27 +174,28 @@ public void testReadRowNumbers() throws IOException {
 
   @Test
   public void testReadRowNumbersWithDelete() throws IOException {
-    if (vectorized) {
-      List<InternalRow> expectedRowsAfterDelete = Lists.newArrayList(EXPECTED_ROWS);
-      // remove row at position 98, 99, 100, 101, 102, this crosses two row groups [0, 100) and [100, 200)
-      for (int i = 1; i <= 5; i++) {
-        expectedRowsAfterDelete.remove(98);
-      }
+    Assume.assumeTrue(vectorized);
```
RussellSpitzer, line 177: This change is unrelated to the current pr correct? Looks like we are just using the correct Junit ignore functionality but for a test which doesn't use the "isdeleted" column

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -277,8 +276,6 @@ public void testPosDeletesAllRowsInBatch() throws IOException {
 
   @Test
   public void testPosDeletesWithDeletedColumn() throws IOException {
-    Assume.assumeFalse(vectorized);
```
RussellSpitzer, line 280: +1

RussellSpitzer (commented):
I think this is pretty close, I just have a few questions about how we propagate around null rowIdMapping and isDeletedColumns and a few naming nits.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
flyrain: Exactly.

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorHolder.java
flyrain: We may keep it as is to keep the naming consistent since we don't change the class name.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/DeletedMetaColumnVector.java
flyrain: Make sense. We did in class `RowPositionColumnVector`.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
flyrain: We've set the ColumnVectorWithFilter with rowIdMapping at line 133. The instance of ColumnVectorWithFilter will throw NPE if we set rowIdMapping to null.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
flyrain: yeah, let me add doc.

flyrain:
Thanks @RussellSpitzer for the review. Refactor class `ColumnBatchReader` a bit to remove two loop on isDeleted array. My benchmark shows a bit perf gain.

flyrain:
Hi @aokolnychyi and @RussellSpitzer, vectorized read is enabled by default several months ago. But the benchmark still assumes it false by default. I have set it false explicitly, and run the benchmark again. Now we can see the big performance gain between vectorized and non-vectorized read, as the following diagram shows.
Screen Shot 2022-06-22 at 4 22 09 PM
I also profile the benchmarks. Here is the flame graph for 25% vectorized read. It looks normal to me. The program spent majority time to read pos delete file and the data file. The read of position delete file is still non-vectorized, which takes a big portion. Would suggest to enable the vectorized read on delete files to improve the overall perf. That's probably next step we can do.
Screen Shot 2022-06-22 at 4 26 57 PM

RussellSpitzer (approved):
I'm good to go on this, @aokolnychyi are you ready as well?

aokolnychyi:
Sorry for the delay. Let me see.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/DeletedMetaColumnVector.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.spark.SparkSchemaUtil;
+import org.apache.iceberg.types.Type;
+import org.apache.spark.sql.types.Decimal;
+import org.apache.spark.sql.vectorized.ColumnVector;
+import org.apache.spark.sql.vectorized.ColumnarArray;
+import org.apache.spark.sql.vectorized.ColumnarMap;
+import org.apache.spark.unsafe.types.UTF8String;
+
+public class DeletedMetaColumnVector extends ColumnVector {
```
aokolnychyi, line 31: The naming in new classes is a bit inconsistent. Can we align that?

```
IsDeletedVectorHolder
DeletedMetaColumnVector
DeletedVectorReader
```

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReadMetadataColumns.java
```diff
@@ -70,8 +70,7 @@ public class TestSparkParquetReadMetadataColumns {
   private static final Schema PROJECTION_SCHEMA = new Schema(
       required(100, "id", Types.LongType.get()),
       required(101, "data", Types.StringType.get()),
-      MetadataColumns.ROW_POSITION,
```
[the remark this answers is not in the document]
flyrain: I put the reason here https://github.com/apache/iceberg/pull/4888#discussion_r884039841.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+package org.apache.iceberg.spark.data.vectorized;
+
+import org.apache.iceberg.arrow.vectorized.VectorHolder;
+import org.apache.iceberg.arrow.vectorized.VectorHolder.ConstantVectorHolder;
+import org.apache.iceberg.arrow.vectorized.VectorHolder.IsDeletedVectorHolder;
+import org.apache.iceberg.types.Types;
+import org.apache.spark.sql.vectorized.ColumnVector;
+
+public class ColumnVectorBuilder {
```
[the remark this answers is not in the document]
flyrain: It's not necessary. Let me change it to package-level.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+    return this;
+  }
+
+  public ColumnVector build() {
+    if (holder.isDummy()) {
+      if (holder instanceof IsDeletedVectorHolder) {
+        return new DeletedMetaColumnVector(Types.BooleanType.get(), isDeleted);
+      } else if (holder instanceof ConstantVectorHolder) {
+        return new ConstantColumnVector(Types.IntegerType.get(), numRows,
+            ((ConstantVectorHolder) holder).getConstant());
```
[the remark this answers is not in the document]
flyrain: it cannot with the `<?>`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -66,35 +78,50 @@ public final ColumnarBatch read(ColumnarBatch reuse, int numRowsToRead) {
...
       if (hasEqDeletes()) {
-        applyEqDelete();
+        applyEqDelete(newColumnarBatch);
       }
-      return columnarBatch;
+
+      if (hasIsDeletedColumn && rowIdMapping != null) {
+        // reset the row id mapping array, so that it doesn't filter out the deleted rows
+        for (int i = 0; i < numRowsToRead; i++) {
+          rowIdMapping[i] = i;
```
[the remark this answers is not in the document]
flyrain: That's a good question. In short, I'm using row ID mapping to improve eq deletes perf when we have both pos deletes and eq deletes. I think it is worth to do that since applying eq deletes is expensive, it has to go row by row.  Here is an example, after the pos deletes, we will only need to iterate 6 rows instead of 8 rows for applying eq delete.
```
     * Filter out the equality deleted rows. Here is an example,
     * [0,1,2,3,4,5,6,7] -- Original status of the row id mapping array
     * [F,F,F,F,F,F,F,F] -- Original status of the isDeleted array
     * Position delete 2, 6
     * [0,1,3,4,5,7,-,-] -- After applying position deletes [Set Num records to 6]
     * [F,F,T,F,F,F,T,F] -- After applying position deletes
     * Equality delete 1 <= x <= 3
     * [0,4,5,7,-,-,-,-] -- After applying equality deletes [Set Num records to 4]
     * [F,T,T,T,F,F,T,F] -- After applying equality deletes
```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -105,35 +132,31 @@ ColumnVector[] readDataToColumnVectors(int numRowsToRead) {
             "Number of rows in the vector %s didn't match expected %s ", numRowsInVector,
             numRowsToRead);
 
-        arrowColumnVectors[i] = hasDeletes() ?
-            ColumnVectorWithFilter.forHolder(vectorHolders[i], rowIdMapping, numRows) :
-            IcebergArrowColumnVector.forHolder(vectorHolders[i], numRowsInVector);
+        arrowColumnVectors[i] = new ColumnVectorBuilder(vectorHolders[i], numRowsInVector)
```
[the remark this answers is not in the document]
flyrain: Nice suggestion. Made the change.

arrow/src/main/java/org/apache/iceberg/arrow/vectorized/VectorizedArrowReader.java
```diff
@@ -517,5 +517,32 @@ public void setBatchSize(int batchSize) {
     }
   }
 
+  /**
+   * A Dummy Vector Reader which doesn't actually read files, instead it returns an
+   * IsDeleted Vector Holder which indicates whether a given row is deleted.
+   */
+  public static class DeletedVectorReader extends VectorizedArrowReader {
```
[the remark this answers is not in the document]
flyrain: I've change `IsDeletedVectorHolder` to `DeletedVectorHolder`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/DeletedMetaColumnVector.java
flyrain: Made the following changes
```
IsDeletedVectorHolder -> DeletedVectorHolder 
DeletedMetaColumnVector -> DeletedColumnVector
DeletedVectorReader
```

flyrain:
Hi @aokolnychyi, this is ready for review. I have to apply the same changes to Spark 3.3, otherwise unit test won't pass.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.iceberg.arrow.vectorized.VectorHolder;
+import org.apache.iceberg.arrow.vectorized.VectorHolder.ConstantVectorHolder;
+import org.apache.iceberg.types.Types;
+import org.apache.spark.sql.vectorized.ColumnVector;
+
+class ColumnVectorBuilder {
+  private boolean[] isDeleted;
+  private int[] rowIdMapping;
+
+  public ColumnVectorBuilder withDeletedRows(int[] rowIdMappingArray, boolean[] isDeletedArray) {
```
aokolnychyi, line 31: nit: I feel we better make this a constructor and pass these arrays only once during the construction.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
```diff
@@ -66,37 +69,53 @@ public final ColumnarBatch read(ColumnarBatch reuse, int numRowsToRead) {
...
-      return columnarBatch;
+
+      return newColumnarBatch;
     }
 
-    ColumnVector[] readDataToColumnVectors(int numRowsToRead) {
+    ColumnVector[] readDataToColumnVectors() {
       ColumnVector[] arrowColumnVectors = new ColumnVector[readers.length];
 
+      ColumnVectorBuilder columnVectorBuilder = new ColumnVectorBuilder();
```
aokolnychyi, line 118: nit: This is probably where you can pass `rowIdMapping` and `isDeleted` as those two don't change.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnarBatchReader.java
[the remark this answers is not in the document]
aokolnychyi: Sounds good to me.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.iceberg.arrow.vectorized.VectorHolder;
+import org.apache.iceberg.arrow.vectorized.VectorHolder.ConstantVectorHolder;
+import org.apache.iceberg.types.Types;
+import org.apache.spark.sql.vectorized.ColumnVector;
+
+class ColumnVectorBuilder {
+  private boolean[] isDeleted;
+  private int[] rowIdMapping;
+
+  public ColumnVectorBuilder withDeletedRows(int[] rowIdMappingArray, boolean[] isDeletedArray) {
```
aokolnychyi, line 31: nit: Same for 3.3

aokolnychyi (approved):
LGTM. I had only one nit (same in 3.2 and 3.3).

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ColumnVectorBuilder.java
flyrain: I am trying to make the builder more generic so that it can also be used for creation of vectors without deletes.

aokolnychyi: Okay, I see now. Then it is fine.

aokolnychyi:
Thanks, @flyrain! Great to have this done. Thanks for reviewing, @RussellSpitzer!

flyrain:
Thanks for the review, @aokolnychyi @RussellSpitzer. Per discussion with @aokolnychyi, I will file a followup to throw an exception when `_deleted` column is used in spark 2.4/3.0/3.1. It always return false no matter whether a row is deleted.

flyrain:
@aokolnychyi, checked the module Spark2.4/3.0/3.1. Metadata column `_deleted` isn't supported. For example, I've added the following tests into class `TestSelect`. 
```
@Test
public void testSelectDeletedMetaColumn() {
  List<Object[]> expected = ImmutableList.of(
      row(1L, "a", 1.0F), row(2L, "b", 2.0F), row(3L, "c", Float.NaN));

  assertEquals("Should return all expected rows", expected, sql("SELECT * FROM %s where _deleted=false", tableName));

  expected = ImmutableList.of();
  assertEquals("Should return all expected rows", expected, sql("SELECT * FROM %s where _deleted=true", tableName));
}
```
It reports the following errors. We don't need to change anything in that sense. 
```
cannot resolve '`_deleted`' given input columns: [table.id, table.data, table.doubleVal]; line 1 pos 26;
'Project [*]
+- 'Filter ('_deleted = false)
   +- SubqueryAlias `table`
      +- RelationV2 iceberg[id#12, data#13, doubleVal#14] (Options: [path=file:/var/folders/69/j9m_r8gj69753xfnsjnlsl_m0000gn/T/junit2794346723320750539/junit3760903...)

org.apache.spark.sql.AnalysisException: cannot resolve '`_deleted`' given input columns: [table.id, table.data, table.doubleVal]; line 1 pos 26;
'Project [*]
+- 'Filter ('_deleted = false)
   +- SubqueryAlias `table`
      +- RelationV2 iceberg[id#12, data#13, doubleVal#14] (Options: [path=file:/var/folders/69/j9m_r8gj69753xfnsjnlsl_m0000gn/T/junit2794346723320750539/junit3760903...)
```
And the change I did in class `TestSparkParquetReadMetadataColumns.java`, they are unit tests for internal classes. User cannot use them directly.
In conclusion, we are OK to leave it as is.

PR #4890, opened 2022-05-28
[0.13] Core: Fix query failure when using projection on top of partitions metadata table (#4720)

szehon-ho:
Backport #4720 for 0.13.2 release.

Regression where user cannot have projections on Partitions table.


szehon-ho:
@nastra @RussellSpitzer @rdblue , let me know, if its ok for 0.13.2?

rdblue (approved):

rdblue:
Looks good to me. The vote has already started, so we can discussion whether to do another RC for this on that thread.

PR #4891, opened 2022-05-29
OpenAPI: Have the type before type_required

Fokko:
For Python, when we generate classes it is required to have the non-default variables first:

This is not allowed:
```python
class ListType(IcebergType, rest_catalog.ListType):

    def __init__(
        cls,
        element_id: int,
        element_required: bool = True,
        element: IcebergType
    ):
```
It should be:
```python
class ListType(IcebergType, rest_catalog.ListType):

    def __init__(
        cls,
        element_id: int,
        element: IcebergType,
        element_required: bool = True
    ):
```

It would be nice to have the type first (which is also more important), and then required/optional. Of course, we can just adjust this in the Python init method but would be nice to have this order the same in the open-api spec as well :)

### DIFF
--- a/open-api/rest-catalog-open-api.yaml
+++ b/open-api/rest-catalog-open-api.yaml
@@ -907,17 +907,17 @@ components:
       required:
         - id
         - name
-        - required
         - type
+        - required
       properties:
         id:
           type: integer
         name:
           type: string
-        required:
-          type: boolean
         type:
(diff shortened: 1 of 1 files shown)


rdblue (approved):

rdblue:
Sounds reasonable to me. Thanks, @Fokko!

PR #4894, opened 2022-05-29
Core: Add REST catalog session timeout

rdblue:
This adds a timeout for REST catalog sessions. After the timeout is reached after a session is used, it will be removed and its token will no longer be refreshed using token exchange.

This also cleans up catalog options for token refresh:
* Renamed token expiration property to `token-expires-in-ms`
* `auth.default-refresh-enabled` should be set to enable token refresh when `token-expires-in-ms` is not present

The default refresh option is needed for long-running environments that need to refresh tokens because references are kept for a long time, like Flink jobs. The default is to not refresh for environments like Trino that should not continuously refresh tokens that are no longer referenced and used.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Properties.java
+++ b/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Properties.java
@@ -37,8 +37,8 @@ public class OAuth2Properties {
    * Interval in milliseconds to wait before attempting to exchange the configured catalog Bearer token.
    * By default, token exchange will be attempted after 1 hour.
    */
-  public static final String EXCHANGE_TOKEN_MS = "exchange-token-in-ms";
-  public static final long EXCHANGE_TOKEN_MS_DEFAULT = 3_600_000; // 1 hour
+  public static final String TOKEN_EXPIRES_IN_MS = "token-expires-in-ms";
+  public static final long TOKEN_EXPIRES_IN_MS_DEFAULT = 3_600_000; // 1 hour
 
   /**
    * Additional scope for OAuth2.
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Properties.java
```diff
@@ -37,8 +37,8 @@ private OAuth2Properties() {
    * Interval in milliseconds to wait before attempting to exchange the configured catalog Bearer token.
    * By default, token exchange will be attempted after 1 hour.
    */
-  public static final String EXCHANGE_TOKEN_MS = "exchange-token-in-ms";
-  public static final long EXCHANGE_TOKEN_MS_DEFAULT = 3_600_000; // 1 hour
+  public static final String TOKEN_EXPIRES_IN_MS = "token-expires-in-ms";
```
rdblue, line 40: This rename brings the property more in line with the OAuth2 `expires_in` field.

danielcweeks: OAuth spec has this as seconds.  Should we just use the unit from the spec?  I'm not sure milliseconds add much value.

danielcweeks (approved):

core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Properties.java
rdblue: All other durations in Iceberg config use milliseconds, so I did this for consistency. I agree it doesn't help much, but I think consistency is important so people don't always need to look up the expected unit.

rdblue:
Thanks for reviewing, @danielcweeks!

PR #4896, opened 2022-05-29
Core: Add Basic auth for client credentials flow

rdblue:
This adds a Basic auth header to client credentials request as recommended in [RFC 6749 Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1).

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Util.java
+++ b/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Util.java
@@ -23,6 +23,8 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JsonNode;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


rdblue:
@bryanck, can you take a look? This adds the Basic authentication header for client credentials exchanges.

core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Util.java
```diff
@@ -144,8 +148,19 @@ public static OAuthTokenResponse fetchToken(RESTClient client, Map<String, Strin
...
         credential, scope != null ? ImmutableList.of(scope) : ImmutableList.of());
 
+    Map<String, String> requestHeaders;
+    if (headers.containsKey(AUTHORIZATION_HEADER)) {
+      requestHeaders = headers;
+    } else {
+      String credentialAsBase64 = Base64.getEncoder().encodeToString(credential.getBytes(StandardCharsets.UTF_8));
+      requestHeaders = ImmutableMap.<String, String>builder()
+          .putAll(headers)
+          .put(AUTHORIZATION_HEADER, BASIC_PREFIX + credentialAsBase64)
```
bryanck, line 158: It seems like this will put the credential as a basic auth header, but `clientCredentialsRequest()` already puts it in the payload so it is really needed?

rdblue: I thought that you said we preferred to use the Basic header, so I added it. I didn't remove the ID and secret from the payload because it is used in some cases, like when there is a Bearer token that needs to be passed in Authorization for the catalog. We can do whatever you think is right here.

bryanck: I see, thanks for the context. If you need to keep it in the payload anyway, then I'd vote to not add the header also. The spec wording seemed more about avoiding putting it in the payload.

Issue #2542, opened 2021-04-28
Improve backward compatibility tests for spec changes introduced in all table versions


yyanyy:
Stem from this comment: https://github.com/apache/iceberg/pull/2521#issuecomment-828793923

We had a NPE bug (#2405) caused by reading a newly introduced field as  `boolean`, which assigns null to this `boolean` field when reading old tables. While Iceberg has backward compatibility tests to test reading V1 tables, this field is introduced for all specs, and thus when creating V1 table for testing, this field will still be populated. NPE only occurs in a wider setting - during table commits (fixed in #2495) or reading manifest table (fixed in #2521). 

I think to prevent this from happening in future, we may need to introduce a mechanism that creates table not through Iceberg code itself, so that the new fields introduced to the spec can continue to be null. The comment link above describes a proposal to achieve it but there might be better approaches.

szehon-ho:
A bit late to the discussion, I wonder if we can make Testwriters for the various metadata that writes null for all optional fields, and then try to deserialize it with the regular readers, to test all newly-fields at once.

szehon-ho:
Was chatting with @dramaticlly about it, cc if you want to look at it.

PR #4898, opened 2022-05-30
API, Core: Add schema_id to ContentFile

ConeyLiu:
This is the first part of #4842. Add the schema id to DataFile/DeteFile/ManifestFile and which could be used to evaluate the filter expression based on the schema.


ConeyLiu:
Hi @rdblue @kbendick, could you help to review this when you are free? Thanks a lot.

api/src/main/java/org/apache/iceberg/ContentFile.java
```diff
@@ -39,6 +39,13 @@
    */
   int specId();
 
+  /**
+   * Return id of the schema when write this file.
+   */
+  default int schemaId() {
+    return -1;
```
rdblue, line 46: Instead of returning an incorrect value, I think this should throw an exception.

api/src/main/java/org/apache/iceberg/DataFile.java
```diff
@@ -88,7 +89,8 @@ static StructType getType(StructType partitionType) {
         KEY_METADATA,
         SPLIT_OFFSETS,
         EQUALITY_IDS,
-        SORT_ORDER_ID
+        SORT_ORDER_ID,
+        SCHEMA_ID
```
rdblue, line 93: I think this should be added just above `SPEC_ID`.

api/src/main/java/org/apache/iceberg/ManifestFile.java
```diff
@@ -64,14 +64,15 @@ public interface ManifestFile {
       "Summary for each partition");
   Types.NestedField KEY_METADATA = optional(519, "key_metadata", Types.BinaryType.get(),
       "Encryption key metadata blob");
-  // next ID to assign: 520
+  Types.NestedField SCHEMA_ID = optional(520, "schema_id", Types.IntegerType.get(), "Schema ID");
```
rdblue, line 67: What is the value of adding the schema ID to a manifest file? Manifests can contain files with different schemas and I don't think there is a way to filter manifests based on the schema of the data files it may contain. At best, you could fill this in if all of the files in the manifest had the same schema ID, but that's really unlikely. I'm not sure it would be worth the code and the metadata.

core/src/main/java/org/apache/iceberg/avro/Avro.java
```diff
@@ -338,13 +340,19 @@ public DataWriteBuilder withSortOrder(SortOrder newSortOrder) {
       return this;
     }
 
+    public DataWriteBuilder withSchemaId(int newSchemaId) {
+      this.schemaId = newSchemaId;
```
rdblue, line 344: Why isn't this taken from the schema that is passed in?

core/src/main/java/org/apache/iceberg/V2Metadata.java
```diff
@@ -47,7 +47,9 @@ private V2Metadata() {
       ManifestFile.ADDED_ROWS_COUNT.asRequired(),
       ManifestFile.EXISTING_ROWS_COUNT.asRequired(),
       ManifestFile.DELETED_ROWS_COUNT.asRequired(),
-      ManifestFile.PARTITION_SUMMARIES
+      ManifestFile.PARTITION_SUMMARIES,
+      ManifestFile.KEY_METADATA,
```
rdblue, line 51: Why does this add key metadata?

rdblue:
@ConeyLiu, thanks for breaking this up, but this is still far too large for a single PR. There's no need to update all of the writers in a single PR. Instead, this should focus on core and API classes and just read null values for the new field. You can add writes later.

api/src/main/java/org/apache/iceberg/ManifestFile.java
ConeyLiu: For newly added files, those files should have the same schema ID. I think it should be a common situation that manifest entries have the same schema ID. Such as batch write/overwrite case.

core/src/main/java/org/apache/iceberg/avro/Avro.java
ConeyLiu: I see the schema passed into the `Avro`/`Parquet`/`ORC` is the one converted from Spark/Flink schema. And that schema has the default ID `-1`.

core/src/main/java/org/apache/iceberg/V2Metadata.java
ConeyLiu: I am not sure this is right, please correct me. Because I need to keep the schema matching the `IndexedManifestFile.get` method. 

```java
        case 13:
          return wrapped.partitions();
        case 14:
          return wrapped.keyMetadata();
        case 15:
          return wrapped.schemaId();
        default:
          throw new UnsupportedOperationException("Unknown field ordinal: " + pos);
```

api/src/main/java/org/apache/iceberg/ContentFile.java
ConeyLiu: I tried with throw an exception and then we need to add more `try catch` the wrap the function call. I keep this with -1 and add comments for the -1 means.

ConeyLiu:
Hi @rdblue, thanks for the review. I have removed the changes for the writers. Please take another look, thanks a lot.

core/src/main/java/org/apache/iceberg/avro/Avro.java
rdblue: Where is this happening?

rdblue:
> Is this something we can do or need to wait until V3?

@szehon-ho, this is something we can do because it is backward compatible. Older readers will ignore new fields, so we can add it safely. And if we don't have a schema ID for a column, then we just return null and skip the optimization.

api/src/main/java/org/apache/iceberg/ContentFile.java
rdblue: None of the implementations should actually use this implementation. The `default` is needed for binary compatibility, but if anyone calls this it should throw an exception.

ConeyLiu: Updated

api/src/main/java/org/apache/iceberg/ManifestFile.java
ConeyLiu: Removed the schema ID for manifest file.

core/src/main/java/org/apache/iceberg/avro/Avro.java
ConeyLiu: You could find it in the following code. The `new Schema(struct.fields())` will create the schema with the default schema ID 0.

```java
  Schema writeSchema = validateOrMergeWriteSchema(table, dsSchema, writeConf);

  // validateOrMergeWriteSchema
  private static Schema validateOrMergeWriteSchema(Table table, StructType dsSchema, SparkWriteConf writeConf) {
    Schema writeSchema;
    if (writeConf.mergeSchema()) {
      ...
    } else {
      writeSchema = SparkSchemaUtil.convert(table.schema(), dsSchema);
      TypeUtil.validateWriteSchema(
          table.schema(), writeSchema, writeConf.checkNullability(), writeConf.checkOrdering());
    }

    return writeSchema;
  }

  // SparkSchemaUtil.convert
  public static Schema convert(Schema baseSchema, StructType sparkType) {
    // convert to a type with fresh ids
    Types.StructType struct = SparkTypeVisitor.visit(sparkType, new SparkTypeToType(sparkType)).asStructType();
    // reassign ids to match the base schema
    Schema schema = TypeUtil.reassignIds(new Schema(struct.fields()), baseSchema);
    // fix types that can't be represented in Spark (UUID and Fixed)
    return SparkFixupTypes.fixup(schema, baseSchema);
  }
```

format/spec.md
```diff
@@ -441,6 +441,8 @@ The schema of a manifest file is a struct called `manifest_entry` with the follo
 | _optional_ | _optional_ | **`132  split_offsets`**          | `list<133: long>`            | Split offsets for the data file. For example, all row group offsets in a Parquet file. Must be sorted ascending |
 |            | _optional_ | **`135  equality_ids`**           | `list<136: int>`             | Field ids used to determine row equality in equality delete files. Required when `content=2` and should be null otherwise. Fields with ids listed in this column must be present in the delete file |
 | _optional_ | _optional_ | **`140  sort_order_id`**          | `int`                        | ID representing sort order for this file [3]. |
+| _optional_ | _optional_ | **`141  spec_id`**                | `int`                        | ID representing Partition Spec for this file |
```
szehon-ho, line 444: Spec id is actually a generated field, its not persisted in spec.

spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/SparkDataFile.java
```diff
@@ -81,6 +82,7 @@ public SparkDataFile(Types.StructType type, StructType sparkType) {
     keyMetadataPosition = positions.get("key_metadata");
     splitOffsetsPosition = positions.get("split_offsets");
     sortOrderIdPosition = positions.get("sort_order_id");
+    schemaIdPosition = positions.get("schema_id");
```
szehon-ho, line 85: Nit: should we put it in the consistent position, after file format?

szehon-ho (commented):
I wonder, is it possible to add a test to try to deserialize an older manifest entry without schema_id?

format/spec.md
ConeyLiu: Thanks for the explanation. Updated

spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/SparkDataFile.java
ConeyLiu: Updated

ConeyLiu:
> I wonder, is it possible to add a test to try to deserialize an older manifest entry without schema_id?

It seems to need to implement customized V1Writer/V2Writer and those `Indexed` classes. Is there any suggestion for this?

szehon-ho:
Yea there is some limited discussion in this related issue, but I guess no good conclusions:  https://github.com/apache/iceberg/issues/2542.  

Maybe a test writer that creates metadata files with all optional columns as null?  That way can test all the new columns at once.

By the way,  change mostly looks good, and it's good you put the new field is optional to avoid the issue like the one mentioned.  I just dont know enough about the serialization/deserialization code to be sure if there are any other problems with previous serialized metadata, so was hoping to have a test to verify it.  Though can leave to @rdblue to approve if he is confident , and we can tackle the backward compat tests on the side.

ConeyLiu:
Thanks @szehon-ho for the review and suggestion.

> Maybe a test writer that creates metadata files with all optional columns as null? That way can test all the new columns at once.

I will add the test later.

core/src/test/java/org/apache/iceberg/TestManifestReader.java
```diff
@@ -142,9 +147,61 @@ public void testDeleteFilePositions() throws IOException {
...
+        1587L,
+        ImmutableMap.of(1, 15L, 2, 122L, 3, 4021L, 4, 9411L, 5, 15L), // sizes
+        ImmutableMap.of(1, 100L, 2, 100L, 3, 100L, 4, 100L, 5, 100L),  // value counts
+        ImmutableMap.of(1, 0L, 2, 0L, 3, 0L, 4, 0L, 5, 0L), // null value counts
+        ImmutableMap.of(5, 10L), // nan value counts
+        ImmutableMap.of(1, Conversions.toByteBuffer(Types.IntegerType.get(), 1)),  // lower bounds
+        ImmutableMap.of(1, Conversions.toByteBuffer(Types.IntegerType.get(), 1))); // upper bounds
+    Integer sortOrderId = 2;
+
+    String fileName = String.format("OldManifestFileV%s.avro", formatVersion);
```
ConeyLiu, line 186: The `OldManifestFileV1.avro/OldManifestFileV2.avro/` is the previous DataFile(the file instance is https://github.com/apache/iceberg/blob/master/core/src/test/java/org/apache/iceberg/TestManifestWriterVersions.java#L74). Hi, @szehon-ho I think this test covered your concern.

ConeyLiu:
Hi @rdblue @szehon-ho, I am sorry for the late update. The compatible test has been added. Hopeful, you could take another look when you are free.

core/src/test/java/org/apache/iceberg/TestManifestReader.java
szehon-ho: This is really great to have this test.  I was initially thinking that we want to have a TestV1Writer / TestV2Writer that writes all the optional fields as null, instead of checking in the old version avro file.  Is that possible?   I can also take a look myself if that is possible.

core/src/main/java/org/apache/iceberg/deletes/EqualityDeleteWriter.java
```diff
@@ -42,18 +42,26 @@
   private final ByteBuffer keyMetadata;
   private final int[] equalityFieldIds;
   private final SortOrder sortOrder;
+  private final int schemaId;
   private DeleteFile deleteFile = null;
 
   public EqualityDeleteWriter(FileAppender<T> appender, FileFormat format, String location,
                               PartitionSpec spec, StructLike partition, EncryptionKeyMetadata keyMetadata,
                               SortOrder sortOrder, int... equalityFieldIds) {
+    this(appender, format, location, spec, partition, keyMetadata, sortOrder, -1, equalityFieldIds);
```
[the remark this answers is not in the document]
ConeyLiu: This is a public class and constructor, I think we should keep the compatibility.

core/src/main/java/org/apache/iceberg/BaseFile.java
```diff
@@ -54,6 +54,7 @@ public PartitionData copy() {
   private Types.StructType partitionType;
 
   private Long fileOrdinal = null;
+  private int schemaId = -1;
```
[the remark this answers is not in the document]
ConeyLiu: I am OK with either one. I followed the `partitionSpecId` behavior.

core/src/test/java/org/apache/iceberg/TestManifestReader.java
```diff
@@ -142,9 +147,61 @@ public void testDeleteFilePositions() throws IOException {
...
+        required(5, "double", Types.DoubleType.get()));
+
+    PartitionSpec spec = PartitionSpec.builderFor(schema)
+        .identity("category")
+        .hour("timestamp")
+        .bucket("id", 16)
+        .build();
+    PartitionData partition = DataFiles.data(spec, "category=cheesy/timestamp_hour=10/id_bucket=3");
+
+    String path = "s3://bucket/table/category=cheesy/timestamp_hour=10/id_bucket=3/file.avro";
```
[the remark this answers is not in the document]
ConeyLiu: There is a check: `Assert.assertEquals("File path should match", path, entry.file().path());`

core/src/test/java/org/apache/iceberg/TestManifestReader.java
ConeyLiu: I am not sure I have fully understood your ways. In this pr, we not only add a new field to the `DataFile` and also changed the return StructType of `DataFile.getType`. Shouldn't we need to customize DataFile/V1Metadata/V2Metadata to write the DataFile with old data spec?

core/src/main/java/org/apache/iceberg/BaseFile.java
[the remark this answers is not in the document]
szehon-ho: I think partitionSpecId is different because that's a derived field. Slightly prefer Integer more reflects what's persisted on disk.

core/src/main/java/org/apache/iceberg/deletes/EqualityDeleteWriter.java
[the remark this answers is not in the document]
szehon-ho: I'm not a huge fan of it, I think we can have a builder like SparkAppenderFactory (where we did some refactor to not have one constructor per new argument)

szehon-ho (commented):
Hey Im sorry for the late responses, as Im definitely not the most familiar to review this part of the code.  I was chatting with @dramaticlly and he could possibly help take a look at the idea for making a cleaner backward compatiblity test in general, over at https://github.com/apache/iceberg/issues/2542.  Dont want to block the change until that's there, but for me its more re-assuring to have those tests overall :)

Also I noticed, schemaId is already written in the header of each manifest.  As far as I can tell, it seems even during rewriteManifests.  Is it adequate for the optimization you are planning?

core/src/main/java/org/apache/iceberg/deletes/EqualityDeleteWriter.java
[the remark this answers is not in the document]
ConeyLiu: Should we do it in this patch or a separate patch?

core/src/main/java/org/apache/iceberg/BaseFile.java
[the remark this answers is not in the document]
ConeyLiu: Updated to use Integer.

ConeyLiu:
Hi @szehon-ho thanks for reviewing this again. 

> Also I noticed, spec-id and schema is already written in the header of each manifest. As far as I can tell, it seems to be the right one that the manifest was written in, even after rewriteManifests. Wondering at a high level, is it adequate for the optimization you are planning?

What happened after the rewritten data file? It seems like the schema of the manifest file is the current table schema. And those manifest entries in the same manifest file could have a different schema after rewrite.

szehon-ho:
@ConeyLiu that's a good question, I think (may be wrong) rewriteDataFiles groups files by partition/partition spec, and may not preserve the old schemas.  Ie, all the data files are rewritten with latest schema of that partition spec.  

I think the situation would be the same even in your proposal to add new schemaid field to data_file, right?  After rewriteDataFiles we have to carry over the latest schema-id of each spec , in order for your initial proposed optimization to be accurate?  Because there may be data in the new file that was written by a later schema.

ConeyLiu:
> I think the situation would be the same even in your proposal to add new schemaid field to data_file, right? After rewriteDataFiles we have to carry over the latest schema-id of each spec , in order for your initial proposed optimization to be accurate? Because there may be data in the new file that was written by a later schema.

You are correct. The data file with the new spec after rewrite. We can not benefit from the schema evaluation because we lost the original schema information.

> As far as I can tell, it seems to be the right one that the manifest was written in, even after rewriteManifests. 

In RewriteManifest, we use the current table partition [spec](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteManifestsSparkAction.java#L103) or the specified spec with spec ID. I think the schema used in the current space is not the same as the original schema for the old manifest file. That's because we will [rewrite the partition spec](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/TableMetadata.java#L952) when updating the table schema. Please correct me if I am wrong.

szehon-ho:
Yea you are right, it seems it will set the latest schema of each spec on the rewritten manifests, so the information is lost if you evolve schemas within a spec.

chenjunjiedada:
@szehon-ho，@rdblue  Do we have any agreement here?

core/src/test/java/org/apache/iceberg/TestManifestReader.java
szehon-ho: Yea I was thinking to make a Test version of those writers..  not sure if its possible.  Anyway, I guess this works too, only issue is wont be too debuggable if something goes wrong.

szehon-ho (commented):
Yea general direction makes sense to me.  But as its changing the spec, would love to get another opinion as well .  Pinged @aokolnychyi  on this if he has time

aokolnychyi:
Will this mean all evaluator logic will have to change to be schema specific? Is there a simple example how this will be consumed?

ConeyLiu:
Thanks @szehon-ho @aokolnychyi

> Will this mean all evaluator logic will have to change to be schema specific? Is there a simple example how this will be consumed?

We don't need to change those existing evaluators. Just need to create a `SchemaEvaluator` by the filter expression and schema, and use it to evaluate the file by its schema.. You can refer [here](https://github.com/apache/iceberg/pull/4842/files#diff-79baa98562948784b4c255ee2dd14ae3a26f875d246f1607f76d58668285ef31R206):
```java
return CloseableIterable.filter(
      open(projection(fileSchema, fileProjection, projectColumns, caseSensitive)),
      entry -> {
        boolean keep = entry != null;
        if (keep && schemaEvaluator != null && entry.file().schemaId() > -1) {
          // evaluate based on the schemaId
          keep = schemaEvaluator.eval(schemasById.get(entry.file().schemaId()));
        }

       return keep &&
            evaluator.eval(entry.file().partition()) &&
            metricsEvaluator.eval(entry.file()) &&
            inPartitionSet(entry.file());
       });
```

ConeyLiu:
Hi @rdblue @szehon-ho @aokolnychyi do you have any time to look at this again?

api/src/main/java/org/apache/iceberg/DataFile.java
```diff
@@ -98,18 +98,20 @@ public interface DataFile extends ContentFile<DataFile> {
...
-  // NEXT ID TO ASSIGN: 142
+  // NEXT ID TO ASSIGN: 143
 
   static StructType getType(StructType partitionType) {
     // IDs start at 100 to leave room for changes to ManifestEntry
     return StructType.of(
         CONTENT,
         FILE_PATH,
         FILE_FORMAT,
+        SCHEMA_ID,
```
ConeyLiu, line 114: @szehon-ho if you are concerned about this order. We may put the `SCHEMA_ID` to last to align other new fields. Then we don't need to update too many methods of `get/put` in `BaseFile`.  And some of the spark UTs are not needed to update as well.

api/src/main/java/org/apache/iceberg/DataFile.java
```diff
@@ -123,7 +124,8 @@ static StructType getType(StructType partitionType) {
         KEY_METADATA,
         SPLIT_OFFSETS,
         EQUALITY_IDS,
-        SORT_ORDER_ID);
+        SORT_ORDER_ID,
+        SCHEMA_ID);
```
ConeyLiu, line 128: Put it in the last one to reduce the code changes.

ConeyLiu:
Hi @rdblue @szehon-ho @aokolnychyi @RussellSpitzer @nastra, could you help to review this? This should be useful for tables with frequent column additions or deletions.

api/src/main/java/org/apache/iceberg/DataFile.java
ConeyLiu: Moved the last

PR #4902, opened 2022-05-30
Spark-3.2: Support Zorder option for rewrite_data_files stored procedure

ajantha-bhat:
Zorder during compaction is not supported in stored procedure. Only supported in spark actions. Hence this PR.

Also, strengthen the validation in the base Zorder implementation. 

Co-authored-by: Ryan Blue <blue@apache.org>


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
```diff
@@ -133,6 +133,14 @@ public SparkZOrderStrategy(Table table, SparkSession spark, List<String> zOrderC
           "Cannot perform ZOrdering, all columns provided were identity partition columns and cannot be used.");
     }
 
+    zOrderColNames.forEach(col -> {
+      NestedField nestedField = table.schema().caseInsensitiveFindField(col);
+      if (nestedField == null) {
+        throw new IllegalArgumentException(
+            String.format("Cannot find field '%s' in struct: %s", col, table.schema().asStruct()));
+      }
+    });
```
ajantha-bhat, line 142: the above validation was implicitly handled from this

https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java#L185-L187

But it is too late and it doesn't enter here for an empty table. So, added the above to catch the error early.

ajantha-bhat:
cc: @RussellSpitzer , @szehon-ho : please review.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
```diff
@@ -133,6 +133,14 @@ public SparkZOrderStrategy(Table table, SparkSession spark, List<String> zOrderC
           "Cannot perform ZOrdering, all columns provided were identity partition columns and cannot be used.");
     }
 
+    zOrderColNames.forEach(col -> {
+      NestedField nestedField = table.schema().caseInsensitiveFindField(col);
```
rdblue, line 137: This should use the case sensitivity setting from Spark, rather than always using case insensitive.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -52,6 +52,7 @@ class RewriteDataFilesProcedure extends BaseProcedure {
           ProcedureParameter.required("table", DataTypes.StringType),
           ProcedureParameter.optional("strategy", DataTypes.StringType),
           ProcedureParameter.optional("sort_order", DataTypes.StringType),
+          ProcedureParameter.optional("z_order", DataTypes.StringType),
```
rdblue, line 55: Is it violating the stored procedure's contract to add a new argument in the middle?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -98,15 +99,19 @@ public InternalRow[] call(InternalRow args) {
       if (sortOrderString != null) {
         sortOrder = collectSortOrders(table, sortOrderString);
       }
-      if (strategy != null || sortOrder != null) {
-        action = checkAndApplyStrategy(action, strategy, sortOrder);
+      String zOrderString = args.isNullAt(3) ? null : args.getString(3);
+      if (sortOrderString != null && zOrderString != null) {
+        throw new IllegalArgumentException("Both sort order and z order is configured. Please configure one of them");
+      }
+      if (strategy != null || sortOrder != null || zOrderString != null) {
```
rdblue, line 106: Style: control flow blocks should be separated by an empty newline.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -129,24 +134,34 @@ private RewriteDataFiles checkAndApplyFilter(RewriteDataFiles action, String whe
...
       return action.sort(sortOrder);
     }
     if (strategy.equalsIgnoreCase("binpack")) {
       RewriteDataFiles rewriteDataFiles = action.binPack();
       if (sortOrder != null) {
         // calling below method to throw the error as user has set both binpack strategy and sort order
         return rewriteDataFiles.sort(sortOrder);
+      } else if (zOrderString != null) {
+        // calling below method to throw the error as user has set both binpack strategy and z order
+        return rewriteDataFiles.zOrder(zOrderString.split(","));
```
rdblue, line 164: This should allow spaces, not just `,`. Can you add `trim` to the results?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
ajantha-bhat: I initially kept it in the end. But then I thought it is good to be together and I "assumed" customers will use named arguments. 

I will move it to the end. Thanks.

docs/spark/spark-procedures.md
```diff
@@ -268,6 +268,7 @@ Iceberg can compact data files in parallel using Spark with the `rewriteDataFile
 | `sort_order`  |    | string | Comma separated sort_order_column. Where sort_order_column is a space separated sort order info per column (ColumnName SortDirection NullOrder). <br/> SortDirection can be ASC or DESC. NullOrder can be NULLS FIRST or NULLS LAST |
 | `options`     | ️   | map<string, string> | Options to be used for actions|
 | `where`       | ️   | string | predicate as a string used for filtering the files. Note that all files that may contain data matching the filter will be selected for rewriting|
+| `z_order`     |    | string | Comma separated column names that are to be considered for z ordering |
```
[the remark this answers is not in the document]
ajantha-bhat: I am not sure how to make it. Can you elaborate? (how to specify z_order without providing an extra argument?) 

I know that z_order extends the sort strategy itself and at a given point of time only sort order or z_order can exist. 
But the arguments for sort order will have sort direction, nulls first and last info which is not required to have.

ajantha-bhat:
@rdblue : Thanks for the review. I have addressed the comments.

docs/spark/spark-procedures.md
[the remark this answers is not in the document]
rdblue: I think the intent is to make zorder look like a function: `zorder(col1, col2)`. +1 for unifying sort order and zorder.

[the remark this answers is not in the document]
ajantha-bhat: ok, I will make the `strategy` to accept binpack, sort or **zorder**.

So, that same sort order can be used. Is that fine? @rdblue , @RussellSpitzer

ajantha-bhat:
@RussellSpitzer , @rdblue : I have unified sort order and z order config. Please have a look at it again. Thanks.

ajantha-bhat:
@RussellSpitzer , @rdblue : Ping...

docs/spark/spark-procedures.md
```diff
@@ -264,7 +264,7 @@ Iceberg can compact data files in parallel using Spark with the `rewriteDataFile
 | Argument Name | Required? | Type | Description |
 |---------------|-----------|------|-------------|
 | `table`       | ✔️  | string | Name of the table to update |
-| `strategy`    |    | string | Name of the strategy - binpack or sort. Defaults to binpack strategy |
+| `strategy`    |    | string | Name of the strategy - binpack, sort or zorder. Defaults to binpack strategy |
```
rdblue, line 267: I thought the plan was to add zorder support to sort?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
```diff
@@ -258,7 +295,7 @@ public void testRewriteDataFilesWithInvalidInputs() {
 
     // Test for invalid strategy
     AssertHelpers.assertThrows("Should reject calls with unsupported strategy error message",
-        IllegalArgumentException.class, "unsupported strategy: temp. Only binpack,sort is supported",
+        IllegalArgumentException.class, "unsupported strategy: temp. Only binpack,sort or zorder is supported",
```
rdblue, line 298: This should be `Only binpack, sort, or zorder are supported`.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
```diff
@@ -133,6 +133,43 @@ public void testRewriteDataFilesWithSortStrategy() {
...
+  public void testRewriteDataFilesWithZOrder() {
+    createTable();
+    // create 10 files under non-partitioned table
+    insertData(10);
+    List<Object[]> expectedRecords = currentData();
+
+    // set z_order = c1,c2
+    List<Object[]> output = sql(
+        "CALL %s.system.rewrite_data_files(table => '%s', " +
+        "strategy => 'zorder', sort_order => 'c1,c2')",
```
rdblue, line 146: @RussellSpitzer, what do you think about this work-around?

i would like to see something like `c3, zorder(c1, c2)` supported, but I don't know if that's going to be too complicated.

RussellSpitzer: I thought we were going to have the strategy just be sort. After that change, I think it's ok to special case zorder support at the moment and eventually support expressions like you are describing.

ajantha-bhat: Sorry, I am confused. 
Is the current implementation of `strategy => 'zorder', sort_order => 'c1,c2'` is ok or do we need to change this?
If we need to change this, can you mention an example of what the arguments look like?

RussellSpitzer: Strategy = sort, sort_order = zorder(c1, c2)

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
```diff
@@ -133,9 +133,23 @@ public SparkZOrderStrategy(Table table, SparkSession spark, List<String> zOrderC
           "Cannot perform ZOrdering, all columns provided were identity partition columns and cannot be used.");
     }
 
+    validateColumnsExistence(table, spark, zOrderColNames);
+
     this.zOrderColNames = zOrderColNames;
   }
 
+  private void validateColumnsExistence(Table table, SparkSession spark, List<String> colNames) {
+    boolean caseSensitive = Boolean.parseBoolean(spark.conf().get("spark.sql.caseSensitive"));
```
RussellSpitzer, line 142: Rather than this we can use the session resolver class which automatically cases or not based on the conf

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
```diff
@@ -133,9 +133,23 @@ public SparkZOrderStrategy(Table table, SparkSession spark, List<String> zOrderC
...
+    Schema schema = table.schema();
+    colNames.forEach(col -> {
+      NestedField nestedField = caseSensitive ? schema.findField(col) : schema.caseInsensitiveFindField(col);
+      if (nestedField == null) {
+        throw new IllegalArgumentException(
+            String.format("Cannot find field '%s' in struct: %s", col, schema.asStruct()));
```
RussellSpitzer, line 148: "Cannot find field x which is specified in the ZOrder in the table schema x,y,z"

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -139,7 +140,10 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
 
   private RewriteDataFiles checkAndApplyStrategy(RewriteDataFiles action, String strategy, SortOrder sortOrder) {
     // caller of this function ensures that between strategy and sortOrder, at least one of them is not null.
-    if (strategy == null || strategy.equalsIgnoreCase("sort")) {
+    if (strategy == null || strategy.equalsIgnoreCase("sort") || strategy.equalsIgnoreCase("zorder")) {
```
RussellSpitzer, line 143: What I was thinking was we just keep Binpack and sort order as options then we add in a special casing like

If sort_order.startsWith(zorder) {
   zorder
} else {
   Normal sort path
}

And this is what we would do till we have better transform handling for multi-column transforms.

RussellSpitzer (commented):
I think I'm still in favor of Zorder being a special case of sort order, the Action api at the moment is really just a work around of that not being easy to specify at the moment. If everyone else thinks it should permanently be a separate strategy I'm ok with that too. It may make sense to keep the APIs parallel for now and then in the future allow SORT to just take all sorts of expressions.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
ajantha-bhat: there can be some column called `zorder` it will conflict with the assumption?

RussellSpitzer: Ok "zorder\s*\(.*\)"

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
ajantha-bhat: Can you point me to an example code? I am not aware of it and couldn't find it either. 

Also, I did like this because some part of the iceberg code is doing similarly. (by lookingup `spark.sql.caseSensitive` in the code) . Maybe I could replace them too.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
```diff
@@ -292,6 +329,13 @@ public void testRewriteDataFilesWithInvalidInputs() {
         IllegalArgumentException.class, "Cannot parse predicates in where option: col1 = 3",
         () -> sql("CALL %s.system.rewrite_data_files(table => '%s', " +
             "where => 'col1 = 3')", catalogName, tableIdent));
+
+    // Test for z_order with invalid column name
+    AssertHelpers.assertThrows("Should reject calls with error message",
+        IllegalArgumentException.class, "Cannot find field 'col1' in which is specified in the ZOrder " +
```
rdblue, line 335: No need for "in the ZOrder" here. That just makes the error confusing and doesn't add much. It's clear from the sort order where the missing field is.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
rdblue: Other places in Iceberg do this because it is Iceberg binding expressions. Where we can use the Spark resolver, we should.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,20 +139,34 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
...
+      String strategy,
+      String sortOrderString,
+      Table table) {
+    Pattern zOrderPattern = Pattern.compile("zorder\\s*\\(.*\\)", Pattern.CASE_INSENSITIVE);
+    boolean isZOrder = sortOrderString != null && zOrderPattern.matcher(sortOrderString).matches();
     // caller of this function ensures that between strategy and sortOrder, at least one of them is not null.
     if (strategy == null || strategy.equalsIgnoreCase("sort")) {
-      return action.sort(sortOrder);
+      if (isZOrder) {
+        String columns = sortOrderString.substring(
```
rdblue, line 152: Why not add rules to the SQL parser so that we can parse these strings? This follows the same pattern as transforms in the SQL parser, so we can easily produce a real parser rule:

```antlr
fieldAndTransformList
    : '(' fields+=transform (',' fields+=transform)* ')'
    ;
```

Then you could create a call for that in the parser and convert. I think that would be better.

The problem with customer parsing is that this may produce strange results that don't correspond to what a real parser would do. For example, `zorder(col1, col2), col3)` matches and makes it to the column names. Then it produces an error about not being able to find `col2)` instead of complaining that there's an unexpected char after col3.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
ajantha-bhat: originally it was like that. Russel suggested to add the Zorder in error message. 
I am fine both the ways. I will revert it back.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
RussellSpitzer: https://github.com/apache/iceberg/blob/master/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AlignRowLevelCommandAssignments.scala

Has an example of how to use this, see `conf.resolver`

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
ajantha-bhat: It is a very good suggestion.
we can even support Zorder with "ALTER TABLE ..WRITE ORDER BY" syntax too. If we have our own parser.

But I am just not expert enough to implement it. I will explore this in my free time.

ajantha-bhat: I mean, If we support this we should even support sort order = c1, Zorder(c2,c3).  Not just the zorder but a mixture of zorder and sort order I guess.

ajantha-bhat:
@RussellSpitzer, @rdblue : I have updated this PR to use the extended parser from (https://github.com/apache/iceberg/pull/5185)

api/src/main/java/org/apache/iceberg/expressions/Zorder.java
```diff
@@ -17,24 +17,19 @@
...
-import org.apache.iceberg.NullOrder
-import org.apache.iceberg.Schema
-import org.apache.iceberg.SortDirection
-import org.apache.iceberg.SortOrder
-import org.apache.iceberg.expressions.Term
+import java.util.Arrays;
+import java.util.List;
 
-class SortOrderParserUtil {
+public class Zorder implements Term {
```
rdblue, line 25: Thinking about this a bit more, I think we should move this out of the API module. We don't want to make guarantees about this class since it isn't a "real" part of the expression library yet. It is mostly a placeholder. Can you move this to core?

api/src/main/java/org/apache/iceberg/expressions/Expressions.java
```diff
@@ -301,4 +303,12 @@ public static <T> NamedReference<T> ref(String name) {
   public static <T> UnboundTerm<T> transform(String name, Transform<?, T> transform) {
     return new UnboundTransform<>(ref(name), transform);
   }
+
+  public static Term zorder(String... names) {
```
rdblue, line 307: If we move the `Zorder` class to core, these factory methods won't be possible anymore. I think that's okay, we can have Spark instantiate Zorder directly instead of calling these.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
```diff
@@ -292,6 +341,19 @@ public void testRewriteDataFilesWithInvalidInputs() {
...
+    // Test for z_order with invalid column name
+    AssertHelpers.assertThrows("Should reject calls with error message",
+        IllegalArgumentException.class, "Cannot find field 'col1' in struct:" +
+            " struct<1: c1: optional int, 2: c2: optional string, 3: c3: optional string>",
+        () -> sql("CALL %s.system.rewrite_data_files(table => '%s', strategy => 'sort', " +
+            "sort_order => 'zorder(col1)')", catalogName, tableIdent));
+
+    // Test for z_order with sort_order
+    AssertHelpers.assertThrows("Should reject calls with error message",
+        IllegalArgumentException.class, "Both SortOrder and Zorder is configured: c1,zorder(c2,c3)",
```
rdblue, line 354: Is this not something we can support? Seems like we could build the Spark sort expression with multiple fields like this.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseRewriteDataFilesSparkAction.java
```diff
@@ -141,6 +141,8 @@ public RewriteDataFiles sort() {
 
   @Override
   public RewriteDataFiles zOrder(String... columnNames) {
+    Preconditions.checkArgument(this.strategy == null,
+        "Cannot set strategy to zOrder, it has already been set to %s", this.strategy);
```
rdblue, line 145: `zorder` instead of `zOrder`?

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
ajantha-bhat: Underlaying sparkAction uses a separate strategy for Zorder instead of handling in the sort strategy. 
So, it is not possible to set more than one strategy now. 

cc: @RussellSpitzer

api/src/main/java/org/apache/iceberg/expressions/Expressions.java
ajantha-bhat: done

api/src/main/java/org/apache/iceberg/expressions/Zorder.java
ajantha-bhat: moved.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
ajantha-bhat: I got `Function2<String, String, Object> resolver = SQLConf.get().resolver();` in java but still couldn't figure out how to  use this in my case. 

Scala I can do find and match. But not sure in java.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,36 +141,55 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
...
+      if (sortOrderString != null) {
         // calling below method to throw the error as user has set both binpack strategy and sort order
-        return rewriteDataFiles.sort(sortOrder);
+        return rewriteDataFiles.sort(buildSortOrder(sortOrderFields, schema));
       }
       return rewriteDataFiles;
     } else {
-      throw new IllegalArgumentException("unsupported strategy: " + strategy + ". Only binpack,sort is supported");
+      throw new IllegalArgumentException(
+          "unsupported strategy: " + strategy + ". Only binpack or sort is supported");
```
rdblue, line 185: Is this change needed?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,36 +141,55 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
...
+
     // caller of this function ensures that between strategy and sortOrder, at least one of them is not null.
     if (strategy == null || strategy.equalsIgnoreCase("sort")) {
-      return action.sort(sortOrder);
+      if (!zOrderFields.isEmpty()) {
+        String[] columnNames = zOrderFields.stream().flatMap(
+            field -> ((Zorder) field.term()).refs().stream().map(NamedReference::name)).toArray(String[]::new);
+        return action.zOrder(columnNames);
+      }
+      return action.sort(buildSortOrder(sortOrderFields, schema));
```
rdblue, line 174: Minor: since this is purely an alternative, I would prefer putting it in an `else` block.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
ajantha-bhat: just a better wording.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,36 +141,55 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
...
+      RewriteDataFiles action,
+      String strategy,
+      String sortOrderString,
+      Schema schema) {
+    List<ExtendedParser.RawOrderField> zOrderFields;
+    List<ExtendedParser.RawOrderField> sortOrderFields;
+    if (sortOrderString != null) {
+      List<ExtendedParser.RawOrderField> rawOrderFields = ExtendedParser.parseSortOrder(spark(), sortOrderString);
+      Map<Boolean, List<ExtendedParser.RawOrderField>> partitions = rawOrderFields.stream().collect(
+              Collectors.partitioningBy(field -> field.term() instanceof Zorder));
```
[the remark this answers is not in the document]
ajantha-bhat: agree. Refactored now.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
ajantha-bhat: done

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
RussellSpitzer: To compare two strings A and B
```java
(Boolean) SQLConf.get().resolver().apply(stringA, stringB)

  @Test
  public void useResolver() {
    Assert.assertTrue((Boolean) SQLConf.get().resolver().apply("Hello","hello"));
    Assert.assertTrue((Boolean) SQLConf.get().resolver().apply("Hello","HeLLO"));
    Assert.assertFalse((Boolean) SQLConf.get().resolver().apply("yellow","HeLLO"));
  }

```

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
RussellSpitzer: I still think "struct" is ambiguous even if we don't want to add the other parts I suggested. I would probably just say "in table schema"

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,36 +140,55 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
     return action.options(options);
   }
 
-  private RewriteDataFiles checkAndApplyStrategy(RewriteDataFiles action, String strategy, SortOrder sortOrder) {
+  private RewriteDataFiles checkAndApplyStrategy(
+      RewriteDataFiles action,
```
RussellSpitzer, line 144: formatting issue on this, we tend to not put every arg on it's own line

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
RussellSpitzer: Also maybe instead of field we use "column"

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,36 +140,55 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
...
+        if (field.term() instanceof Zorder) {
+          zOrderFields.add(field);
+        } else {
+          sortOrderFields.add(field);
+        }
+      });
+
+      if (!zOrderFields.isEmpty() && !sortOrderFields.isEmpty()) {
+        // TODO: we need to allow this in future when SparkAction has handling for this.
+        throw new IllegalArgumentException("Both SortOrder and Zorder is configured: " + sortOrderString);
```
RussellSpitzer, line 161: Perhaps, "Cannot mix identity sort columns and a Zorder sort expression"? Since I'm not sure an end user would know what we are referring to here

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/procedures/RewriteDataFilesProcedure.java
```diff
@@ -140,36 +140,55 @@ private RewriteDataFiles checkAndApplyOptions(InternalRow args, RewriteDataFiles
     return action.options(options);
   }
 
-  private RewriteDataFiles checkAndApplyStrategy(RewriteDataFiles action, String strategy, SortOrder sortOrder) {
+  private RewriteDataFiles checkAndApplyStrategy(
+      RewriteDataFiles action,
+      String strategy,
+      String sortOrderString,
+      Schema schema) {
+    List<ExtendedParser.RawOrderField> zOrderFields = Lists.newArrayList();
```
RussellSpitzer, line 148: Could just be a list of Zorder terms at the moment since you check the type before adding to this list and the other fields are meaningless of RawOrderField in this case

RussellSpitzer (commented):
Just a few remaining questions about formatting, the Resolver issue, and some error messages I think need a little work.

The test class has the one ordering issue, I think we can just remove the order checking test unless you want to make sure some non hierarchal sort is being applied. I worry about checking the exact order as we may change the algorithm in the future.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
ajantha-bhat: But here the problem statement is to find the column using the `currentCase` or `lowerCase` of a column name from a map having column name as keys. So, I think resolver (equals()) will not help here.

spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestRewriteDataFilesProcedure.java
```diff
@@ -133,6 +145,43 @@ public void testRewriteDataFilesWithSortStrategy() {
...
+    assertEquals("Action should rewrite 10 data files and add 1 data files",
+        ImmutableList.of(row(10, 1)),
+        output);
+
+    List<Object[]> actualRecords = currentData();
+    assertEquals("Data after compaction should not change", expectedRecords, actualRecords);
+
+    // Due to Z_order, the data written will be in the below order.
+    // As there is only one small output file, we can validate the query ordering (as it will not change).
+    ImmutableList<Object[]> expectedRows = ImmutableList.of(
```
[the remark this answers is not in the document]
ajantha-bhat: line 166 has `orderBy` in the `CurrentData()`. So, it will be sorted always. 

And `assertEquals` will not ignore the order. 

Agree that No need to check two times, (one time for records presence and another time for order)
I will remove 'CurrentData()` logic and keep hardcoded order check logic. 

order checking is needed as that is the only way to prove zorder sort happened.

ajantha-bhat:
> I worry about checking the exact order as we may change the algorithm in the future.

@RussellSpitzer : I believe we can update the test case when we change the underlying order in future. Right now order checking is needed as it is the only way to confirm zorder is actually used.

RussellSpitzer (approved):

RussellSpitzer:
Thanks for the work @ajantha-bhat ! And thank you for reviewing @rdblue !

I assume we'll need to put this in 3.3 as well?

ajantha-bhat:
@RussellSpitzer : Thanks for merging. I will do a port for spark 3.3 very soon.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/SparkZOrderStrategy.java
rdblue: After seeing what this is doing with the `caseSensitive` value, I think this is reasonable. This is using the value to decide whether to use `findField` or `caseInsensitiveFindField`. I don't see a good way to pass a resolver for those cases, so this seems fine to me.

PR #4905, opened 2022-05-30
Return an empty list for child namespaces in the REST catalog client

bryanck:
Currently, an exception is thrown as part of a precondition check when the REST catalog client attempts to list namespaces under a parent. Some versions of Spark attempt to list child namespaces before performing a drop of the namespace, so this throws an exception and a namespace drop is not possible. This PR changes the behavior to return an empty list instead, which allows the Spark namespace drop to proceed. Given the REST catalog does not currently support nested namespaces, this seems semantically correct.


bryanck:
Closing this, as this will be resolved when `listNamespaces()` is fully implemented in the REST catalog client.

PR #4907, opened 2022-05-30
Core: Implement namespace listing with a parent in REST catalog

rdblue:
This implements support for passing a parent namespace to list.


bryanck (approved):

kbendick (approved):
This looks good to me.

We should add a test that calls listNamespaces on a parent namespace (assuming we don’t have one already - which I didn’t see). We can handle that in a follow up though.

danielcweeks (approved):

rdblue:
Thanks for the reviews, everyone!

PR #4909, opened 2022-05-30
AWS: use pre-created IAM role for AssumeRole related integration tests

jackye1995:
as a part of AWS CI effort, remove IAM role that is created during tests and use a pre-created IAM role. The role creation and configuration step will be defined in a CDK stack for easy developer setup.

@amogh-jahagirdar @xiaoxuandev @rajarshisarkar @singhpk234

### DIFF
--- a/aws/src/integration/java/org/apache/iceberg/aws/TestAssumeRoleAwsClientFactory.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/TestAssumeRoleAwsClientFactory.java
@@ -26,87 +26,35 @@ import org.apache.iceberg.aws.s3.S3FileIO;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
-import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
-import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.glue.model.AccessDeniedException;
 import software.amazon.awssdk.services.glue.model.GlueException;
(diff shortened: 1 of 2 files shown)


aws/src/integration/java/org/apache/iceberg/aws/TestAssumeRoleAwsClientFactory.java
```diff
@@ -117,7 +65,7 @@ public void testAssumeRoleGlueCatalog() throws Exception {
       Assert.assertEquals(AccessDeniedException.class, e.getClass());
     }
 
-    Namespace namespace = Namespace.of("allowed_" + UUID.randomUUID().toString().replace("-", ""));
+    Namespace namespace = Namespace.of("iceberg_aws_ci_" + UUID.randomUUID().toString().replace("-", ""));
```
amogh-jahagirdar, line 68: Nit: iceberg_aws_ci_ maybe could be a static prefix? Not blocking though considering it was already like that

aws/src/integration/java/org/apache/iceberg/aws/TestAssumeRoleAwsClientFactory.java
```diff
@@ -129,41 +77,20 @@ public void testAssumeRoleGlueCatalog() throws Exception {
...
-    waitForIamConsistency();
-
-    S3FileIO s3FileIO = new S3FileIO();
-    s3FileIO.initialize(assumeRoleProperties);
-    InputFile inputFile = s3FileIO.newInputFile("s3://" + AwsIntegTestUtil.testBucketName() + "/denied/file");
+  public void testAssumeRoleS3FileIO() {
+    S3Client s3Client = AwsClientFactories.from(assumeRoleProperties).s3();
     try {
-      inputFile.exists();
+      s3Client.getBucketAccelerateConfiguration(GetBucketAccelerateConfigurationRequest.builder()
```
amogh-jahagirdar, line 83: I guess here we can use any method, so it does not matter but would getBucket just be less verbose to read?

amogh-jahagirdar (approved):
Looks good to me, just some minor non-blocking stuff.

PR #4911, opened 2022-05-31
Flink: reduce the scope and duration of holding checkpoint lock.

stevenzwu:
Keep the expensive split planning outside the lock and only lock for emitting splits and update lastSnapshotId position.

### DIFF
--- a/flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
+++ b/flink/v1.13/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
@@ -116,11 +116,7 @@ public class StreamingMonitorFunction extends RichSourceFunction<FlinkInputSplit
   public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
     this.sourceContext = ctx;
     while (isRunning) {
-      synchronized (sourceContext.getCheckpointLock()) {
-        if (isRunning) {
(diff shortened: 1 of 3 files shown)


stevenzwu:
Please see more context from the Slack thread started by @chenjunjiedada : https://apache-iceberg.slack.com/archives/C025PH0G1D4/p1653955614339389

cc @openinx @rdblue @kbendick @Fokko

chenjunjiedada (approved):

stevenzwu:
there are two potentially expensive steps (1) split discovery (2) emitting the discovered splits to downstream readers. Right now, both steps are executed inside the checkpoint lock. This PR moved the first step outside the checkpoint lock section.

kbendick (commented):
So I like this change as the lock should be held less.

And this gives Flink _far_ more chance to deal with backpressure records (though if this is a source maybe it won’t have them).

I think we should combine this with something like https://github.com/apache/iceberg/pull/4943 (if not 4943 directly once it’s finished).

stevenzwu:
@kbendick this PR is orthogonal to PR #4943 . Both are useful.

hililiwei (approved):

stevenzwu:
@openinx can you help take a look?

kbendick (approved):
This LGTM

rdblue (approved):

rdblue:
Thanks, @stevenzwu!

PR #4912, opened 2022-05-31
AWS: add retry logic to S3InputStream

jackye1995:
@danielcweeks @rajarshisarkar @amogh-jahagirdar @xiaoxuandev @singhpk234 

Add retry for the `S3InputStream` so that when we encounter network failures (mostly `SSLException` for server side connection reset and `SocketTimoutException` for client side connection reset), we can quickly retry the read without failing the entire operation.

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
@@ -354,6 +354,30 @@ public class AwsProperties implements Serializable {
   public static final boolean CLIENT_ENABLE_ETAG_CHECK_DEFAULT = false;
 
   /**
+   * Number of times to retry S3 read operation.
+   */
+  public static final String S3_READ_RETRY_NUM_RETRIES = "s3.read.retry.num-retries";
+  public static final int S3_READ_RETRY_NUM_RETRIES_DEFAULT = 7;
+
+  /**
+   * Minimum wait time to retry a S3 read operation
+   */
+  public static final String S3_READ_RETRY_MIN_WAIT_MS = "s3.read.retry.min-wait-ms";
+  public static final long S3_READ_RETRY_MIN_WAIT_MS_DEFAULT = 500; // 0.5 seconds
+
+  /**
+   * Maximum wait time to retry a S3 read operation
+   */
+  public static final String S3_READ_RETRY_MAX_WAIT_MS = "s3.read.retry.max-wait-ms";
+  public static final long S3_READ_RETRY_MAX_WAIT_MS_DEFAULT = 2 * 60 * 1000; // 2 minute
+
+  /**
+   * Total retry time for a S3 read operation
+   */
+  public static final String S3_READ_RETRY_TOTAL_TIMEOUT_MS = "s3.read.retry.total-timeout-ms";
+  public static final long S3_READ_RETRY_TOTAL_TIMEOUT_MS_DEFAULT = 10 * 60 * 1000; // 10 minutes
+
+  /**
    * Used by {@link LakeFormationAwsClientFactory}.
    * The table name used as part of lake formation credentials request.
    */
@@ -380,6 +404,10 @@ public class AwsProperties implements Serializable {
   private int s3FileIoDeleteThreads;
   private boolean isS3DeleteEnabled;
   private final Map<String, String> s3BucketToAccessPointMapping;
+  private int s3ReadRetryNumRetries;
+  private long s3ReadRetryMinWaitMs;
+  private long s3ReadRetryMaxWaitMs;
+  private long s3ReadRetryTotalTimeoutMs;
 
   private String glueCatalogId;
   private boolean glueCatalogSkipArchive;
@@ -404,6 +432,10 @@ public class AwsProperties implements Serializable {
     this.s3FileIoDeleteThreads = Runtime.getRuntime().availableProcessors();
     this.isS3DeleteEnabled = S3_DELETE_ENABLED_DEFAULT;
     this.s3BucketToAccessPointMapping = ImmutableMap.of();
+    this.s3ReadRetryNumRetries = S3_READ_RETRY_NUM_RETRIES_DEFAULT;
+    this.s3ReadRetryMinWaitMs = S3_READ_RETRY_MIN_WAIT_MS_DEFAULT;
+    this.s3ReadRetryMaxWaitMs = S3_READ_RETRY_MAX_WAIT_MS_DEFAULT;
+    this.s3ReadRetryTotalTimeoutMs = S3_READ_RETRY_TOTAL_TIMEOUT_MS_DEFAULT;
 
     this.glueCatalogId = null;
     this.glueCatalogSkipArchive = GLUE_CATALOG_SKIP_ARCHIVE_DEFAULT;
@@ -472,6 +504,14 @@ public class AwsProperties implements Serializable {
             Runtime.getRuntime().availableProcessors());
     this.isS3DeleteEnabled = PropertyUtil.propertyAsBoolean(properties, S3_DELETE_ENABLED, S3_DELETE_ENABLED_DEFAULT);
     this.s3BucketToAccessPointMapping = PropertyUtil.propertiesWithPrefix(properties, S3_ACCESS_POINTS_PREFIX);
+    this.s3ReadRetryNumRetries = PropertyUtil.propertyAsInt(properties, S3_READ_RETRY_NUM_RETRIES,
+        S3_READ_RETRY_NUM_RETRIES_DEFAULT);
+    this.s3ReadRetryMinWaitMs = PropertyUtil.propertyAsLong(properties, S3_READ_RETRY_MIN_WAIT_MS,
+        S3_READ_RETRY_MIN_WAIT_MS_DEFAULT);
+    this.s3ReadRetryMaxWaitMs = PropertyUtil.propertyAsLong(properties, S3_READ_RETRY_MAX_WAIT_MS,
+        S3_READ_RETRY_MAX_WAIT_MS_DEFAULT);
+    this.s3ReadRetryTotalTimeoutMs = PropertyUtil.propertyAsLong(properties, S3_READ_RETRY_TOTAL_TIMEOUT_MS,
+        S3_READ_RETRY_TOTAL_TIMEOUT_MS_DEFAULT);
 
     this.dynamoDbTableName = PropertyUtil.propertyAsString(properties, DYNAMODB_TABLE_NAME,
         DYNAMODB_TABLE_NAME_DEFAULT);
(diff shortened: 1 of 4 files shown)


aws/src/test/java/org/apache/iceberg/aws/s3/TestS3InputStream.java
```diff
@@ -20,45 +20,216 @@
...
 
   @Before
   public void before() {
-    s3.createBucket(CreateBucketRequest.builder().bucket("bucket").build());
+    s3Client.createBucket(CreateBucketRequest.builder().bucket("bucket").build());
   }
 
   @Test
-  public void testRead() throws Exception {
+  public void testReadWithNormalClient() throws Exception {
```
jackye1995, line 78: cannot really figure out a way to do parameterized test here due to the use of `S3MockRule`, parameterization always gives me initialization failure... I will continue to see if I can remove the redundant tests, meanwhile I will also update integration test to run against both normal client and a fuzzy client.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -178,25 +281,56 @@ private void positionStream() throws IOException {
...
+
+    if (exception instanceof S3Exception) {
+      switch (((S3Exception) exception).statusCode()) {
+        case HttpURLConnection.HTTP_NOT_FOUND:
+        case 416: // range not satisfied
+          return false;
+      }
+    }
+
+    return true;
```
amogh-jahagirdar, line 333: Curious is this the same retry policy that Hadoop S3A has?

singhpk234:
[doubt] SDK V2 also has a possibility to define [retry policy / condition](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/retry/PredefinedRetryPolicies.html#DEFAULT_RETRY_CONDITION) by override [ClientConfiguration](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/ClientConfiguration.html) while building the s3 client itself, why are we opting do it here via `Tasks` here ?

danielcweeks (changes requested):
Ok, I understand what you're trying to achieve here, but I think this is best handled by the S3Client via configuration for [Retry Policy](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/retry/RetryPolicy.html) and not custom handling in Iceberg.

Introducing our own retries can ultimately conflict with other retries and can have some bad side-effects (like failures resulting in retries on top of retries).

Unless there's some reason we can't handle this via the SDK, then don't think we should introduce retries here.  @jackye1995 did you look into the retry policy?  It seems like we could use many of these configs and just apply them in the AwsClientFactory.

jackye1995:
> Introducing our own retries can ultimately conflict with other retries and can have some bad side-effects (like failures resulting in retries on top of retries).

Yes you are right, I actually plan to add one more PR for configuring retry policy. The issue here is a bit different. I think there are 2 issues that cannot be handled at SDK level:
1. AWS SDK only handles exceptions up to the AWS client calls, in this case `s3.getObject`. Once you get the input stream, further exceptions are not handled during operations like `read` and `seek`, and that's typically when network issues like connection timeout or socket timeout might happen. I haven't found logic in the SDK that could close a stream, re-seek to the right position and retry a read operation, but maybe I missed some code places.
2. When closing the input stream, the closing should call `abort` instead of `close` to terminate the HTTP request. This should be done both at retry and stream close time. See [this javadoc](https://github.com/aws/aws-sdk-java-v2/blob/56e21b1d8e18f692ba9193c625661aa786e9946b/core/sdk-core/src/main/java/software/amazon/awssdk/core/ResponseInputStream.java#L29-L31) for more details.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -353,6 +353,30 @@ public class AwsProperties implements Serializable {
   @Deprecated
   public static final boolean CLIENT_ENABLE_ETAG_CHECK_DEFAULT = false;
 
+  /**
+   * Number of times to retry S3 read operation.
+   */
+  public static final String S3_READ_RETRY_NUM_RETRIES = "s3.read.retry.num-retries";
+  public static final int S3_READ_RETRY_NUM_RETRIES_DEFAULT = 6;
```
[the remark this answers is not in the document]
jackye1995: thanks for the pointer, updated

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -88,23 +97,68 @@ public void seek(long newPos) {
 
   @Override
   public int read() throws IOException {
-    Preconditions.checkState(!closed, "Cannot read: already closed");
-    positionStream();
+    AtomicInteger byteRef = new AtomicInteger(0);
+    try {
+      Tasks.foreach(0)
```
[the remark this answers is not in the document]
jackye1995: I think we cannot, because the input stream needs to be closed and re-opened, but if we put it here then the retry will always retry against the same stream that is already closed after the first failure.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
jackye1995: It's not. It's closer to the ones in Presto and Trino. Basically it retires almost all IO exceptions except for EOF, because they are most likely network issues. For AWS side exceptions, this logic seems sufficient to me if they are proven sufficient in Presto and Trino. I am not sure if we need to list every single possible exception class like S3A did.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -88,23 +97,68 @@ public void seek(long newPos) {
 
   @Override
   public int read() throws IOException {
-    Preconditions.checkState(!closed, "Cannot read: already closed");
-    positionStream();
+    AtomicInteger byteRef = new AtomicInteger(0);
```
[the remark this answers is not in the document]
jackye1995: Yes good point, it's only used to hold the value, not really trying to prevent any multi-thread access.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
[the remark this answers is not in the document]
jackye1995: Let's discuss this in the thread above https://github.com/apache/iceberg/pull/4912#discussion_r889726156

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -172,31 +274,62 @@ private void positionStream() throws IOException {
...
+        }
+      } catch (IOException | AbortedException e) {
+        // ignore failure to abort or close stream
+      }
     }
   }
 
-  public void setSkipSize(int skipSize) {
-    this.skipSize = skipSize;
+  private static boolean shouldRetry(Exception exception) {
```
danielcweeks, line 309: I think we need to be more explicit about the cases where we do want to retry rather than just defaulting to retry.  I think there were specific cases that were mentined as known issues (e.g. socket timeout).  However, the problem we've had in other retry scenarios is that the retry cases are overly broad and retry when they really shouldn't.  

I think the default here should return false and only return true for the known Exceptions.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -88,23 +96,69 @@ public void seek(long newPos) {
 
   @Override
   public int read() throws IOException {
-    Preconditions.checkState(!closed, "Cannot read: already closed");
-    positionStream();
+    int[] byteRef = new int[1];
+    try {
+      Tasks.foreach(0)
```
danielcweeks, line 101: This is a lot of duplicate code for the retry logic.  Could we consolidate to a method that wraps just the logic we want to execute?

For example:

```java
retry(() -> {
    Preconditions.checkState(!closed, "Cannot read: already closed");
    positionStream();

    byteRef[0] =  stream.read();
})
```

(you many have to use consumer/function for the range versions, but seems like there might be an more concise way to write this)

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -172,31 +274,62 @@ private void positionStream() throws IOException {
...
+
+  private void closeStream() {
+    closeServerSideStream(stream);
+    stream = null;
   }
 
-  private void closeStream() throws IOException {
-    if (stream != null) {
-      stream.close();
+  private static void closeServerSideStream(InputStream streamToClose) {
```
danielcweeks, line 292: I'd just put this in the close stream (not have a separate method for it).

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -88,23 +96,69 @@ public void seek(long newPos) {
 
   @Override
   public int read() throws IOException {
-    Preconditions.checkState(!closed, "Cannot read: already closed");
-    positionStream();
+    int[] byteRef = new int[1];
```
danielcweeks, line 99: This is less than ideal but Tasks is a little limited in this area because it's really based on threaded execution and we're reusing it here for retry.  It might be worth exploring whether we can tweak Tasks to support this usecase:

Ideally we'd have something like:

```java
int read = Tasks.single()
    .exponentialBackoff(...)
    ...
    run(<funciton>);
```

danielcweeks: We could also consider bringing in [Failsafe](https://failsafe.dev/), which is more closely aligned with what we want for these cases.  Though it is a new dependency (but zero deps and apache licensed).

javrasya:
Why was this PR closed? We are experiencing exactly the same problem and S3FileIO is literally unusable for us at the moment. no matter what we did to tune the apache http client such as TcpKeepAlive or socket timeout, it still hits this often and it breaks the whole pipeline. 

We created our own S3FileIO and all the classes till S3InputStream where a SocketException arises, got the inspiration from this PR and our problem is solved. But it is annoying to maintain our own version of those classes. 

Is it not possible to revive this PR?

puchengy:
@jackye1995 Hi, I also encountered the same issue as described in https://github.com/apache/iceberg/issues/10340 can we continue the PR to address the issue here? ALSo I wonder have you seen similar AWS clients having same issue? How do they end up solving the issue? Thanks.

danielcweeks:
@puchengy I spoke with @amogh-jahagirdar the other day and he's taking a look at this.  There have been a few different attempts to address this issue and we're trying to find a solution that fits cleanly into the implementation.

puchengy:
thank you @danielcweeks and @amogh-jahagirdar !

puchengy:
@amogh-jahagirdar Hello, i wonder is there any update on this? Our clients from time to time hit the issue and cause some interruptions to their workload. Thanks!

amogh-jahagirdar:
Hey @puchengy I'm wokring on it! I am aiming to have a PR out sometime this week to address this.

puchengy:
@amogh-jahagirdar Hi, I am checking in to see if there is any update? thank you!

SandeepSinghGahir:
Hi,

Is there any update on this PR? We have been facing the same issue while reading data from iceberg tables in glue job blocking the entire pipeline. Would really appreciate if we can get update and a probable fix for this.

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3InputStream.java
```diff
@@ -178,12 +192,16 @@ private byte[] randomData(int size) {
...
 
   private void writeS3Data(S3URI uri, byte[] data) throws IOException {
-    s3.putObject(
+    s3Client.putObject(
         PutObjectRequest.builder()
             .bucket(uri.bucket())
             .key(uri.key())
             .contentLength((long) data.length)
             .build(),
         RequestBody.fromBytes(data));
```
steveloughran, line 201: FWIW, this triggers a full array copy of the data.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -88,23 +96,69 @@ public void seek(long newPos) {
...
+          .throwFailureWhenFinished()
+          .run(ignored -> {
+            try {
+              Preconditions.checkState(!closed, "Cannot read: already closed");
+              positionStream();
+
+              byteRef[0] =  stream.read();
+            } catch (IOException e) {
+              closeStream();
+              throw new UncheckedIOException(e);
```
steveloughran, line 118: on an unrecoverable network error, `abortStream()` is required to avoid the failed TLS connection being recycled. This adds a new challenge: identify an unrecoverable network error from the deep nested stacks which can come from the SDK, and distinguish from other failures

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
```diff
@@ -172,31 +274,62 @@ private void positionStream() throws IOException {
...
-    if (stream != null) {
-      stream.close();
+  private static void closeServerSideStream(InputStream streamToClose) {
+    if (streamToClose != null) {
+      try {
+        if (streamToClose instanceof Abortable) {
+          // Stated in the ResponseInputStream javadoc:
+          // If it is not desired to read remaining data from the stream,
+          // you can explicitly abort the connection via abort().
+          ((Abortable) streamToClose).abort();
```
steveloughran, line 299: abort() removes it from the HTTP pool.

* this is good if you are closing with an unrecoverable network error
* this is absolutely what you do not want to do on a normal read.

aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java
steveloughran: reasonably confident its a lot more complicated than this, especially as SDK-level failures often create deep chains with the underlying cause at the bottom

steveloughran (commented):
commented. It's a start

PR #4054, opened 2022-02-07
Core: Add InMemory FileIO/Catalog and extract common catalog tests from HadoopCatalog into a common test

mayursrivastava:
The purpose of this pr is to create pure in-memory implementations of FileIO and Catalog+SupportsNamespace that can be utilized to write unit tests that don't use external resources such local disk, databases, etc.

The FileIO implementation is backed by in-memory data-structures, specifically, InMemoryFileStore which uses ConcurrentHashMap.
The Catalog implementation uses in-memory data-structures to store the namespaces and tables, specifically InMemoryCatalogDb which uses ConcurrentHashMaps. The InMemoryCatalog uses InMemoryFileIO.

There was an old corrupted (closed) pr ([commit](https://github.com/apache/iceberg/pull/3294/commits/31cab1d7f06cf3439c3ed1fc8a40ae9e1ef36f96)) with this same change.

@rdblue apologies for messing up the old pr. I've created this new pr with the same changes in the commit mentioned above. Could you please take a look again?

@nastra this is the new inmemory catalog pr.

Thanks,
Mayur

PR #4918, opened 2022-05-31
Core: Add InMemoryFileIO implementation

nastra:
This is a revival of https://github.com/apache/iceberg/pull/4054 to get this over the line. I also addressed review feedback and improved/changed a few other things. Once this PR is in, I'll open a new one with an InMemoryCatalog implementation


core/src/main/java/org/apache/iceberg/io/inmemory/InMemoryFileStore.java
```diff
@@ -0,0 +1,69 @@
...
+package org.apache.iceberg.io.inmemory;
+
+import java.nio.ByteBuffer;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentMap;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.util.ByteBuffers;
+
+/** An in-memory collection based storage for file contents keyed by a string location. */
+class InMemoryFileStore {
```
snazy, line 29: Maybe move the code from this class into `InMemoryFileIO` and make `InMemoryFileIO`'s inner classes top-level classes?

core/src/main/java/org/apache/iceberg/io/inmemory/InMemoryFileStore.java
```diff
@@ -0,0 +1,69 @@
...
+  public void put(String location, ByteBuffer data) {
+    // Copy the contents and store it.
+    store.put(location, ByteBuffers.toByteArray(data));
+  }
+
+  /** Get the file contents for the given location. */
+  public Optional<ByteBuffer> get(String location) {
+    return Optional.ofNullable(store.get(location))
+        .map(
+            bytes -> {
```
snazy, line 47: You can store the `ByteBuffer` in `put()` using `ByteBuffer.asReadOnlyBuffer()` and then return just return `ByteBuffer.duplicate()` here instead of the byte[] copy + wrapping.

nastra: good point, done

core/src/test/java/org/apache/iceberg/TestScansAndSchemaEvolution.java
```diff
@@ -70,7 +70,8 @@ public TestScansAndSchemaEvolution(int formatVersion) {
   private DataFile createDataFile(String partValue) throws IOException {
     List<GenericData.Record> expected = RandomAvroData.generate(SCHEMA, 100, 0L);
 
-    OutputFile dataFile = new InMemoryOutputFile(FileFormat.AVRO.addExtension(UUID.randomUUID().toString()));
+    OutputFile dataFile =
+        new InMemoryFileIO().newOutputFile(FileFormat.AVRO.addExtension(UUID.randomUUID().toString()));
```
kbendick, line 74: Should we store the `InMemoryFileIO` in its own variable so that the underlying `store` is only instantiated once?

core/src/test/java/org/apache/iceberg/avro/TestAvroDeleteWriters.java
```diff
@@ -74,7 +75,7 @@ public void createDeleteRecords() {
 
   @Test
   public void testEqualityDeleteWriter() throws IOException {
-    OutputFile out = new InMemoryOutputFile();
+    OutputFile out = new InMemoryFileIO().newOutputFile(UUID.randomUUID().toString());
```
kbendick, line 78: Same comment here about saving and reusing the `InMemoryFileIO`, especially as this class uses it multiple times.

core/src/test/java/org/apache/iceberg/io/inmemory/InMemoryFileIOTest.java
```diff
@@ -0,0 +1,256 @@
...
+    buffer.putInt(8);
+    // Add magic number
+    buffer.put("par1".getBytes(UTF_8));
+
+    // Put the data in the store
+    buffer.flip();
+    store.put(fileName, buffer);
+
+    // magic number, chunk1, chunk2, footer, footer size, magic number
+    int expectedFileSize = 4 + numRowsInChunk * 4 + numRowsInChunk * 8 + 8 + 4 + 4;
```
kbendick, line 128: Nit: Some of these might be better off as constants. But since it's all in the same method and we'll defined above I'm ok either way.

kbendick (commented):
Thanks for working on this @nastra!

Left a few small comments, mostly about saving and reusing the `InMemoryFileIO` within a test.

Also, for the test that emulates a parquet file, thats fine for the test. But how would we do that within the `core` codebase? Would we even do so? I assume we’d just use the normal `ParquetWriter` etc, but it would be good to eventually have an example of that (possibly as a follow up). It seems to work well with the `PuffinWriter`.

I’m not sure how far we’re trying to take this work for use within the code base though.

nastra:
Closing this as we already have InMemoryFileIO in the codebase

PR #4919, opened 2022-05-31
Dev: Release Script Fix

RussellSpitzer:
Default is wrong on remote
Key values are not used

### DIFF
--- a/dev/source-release.sh
+++ b/dev/source-release.sh
@@ -36,7 +36,7 @@ usage () {
 }
 
 # Default repository remote name
-remote="origin"
+remote="apache"
 
 while getopts "v:r:k:g:d" opt; do
   case "${opt}" in
--- a/dev/source-release.sh
+++ b/dev/source-release.sh
@@ -112,7 +112,7 @@ tarball=$tag.tar.gz
 git archive $release_hash --worktree-attributes --prefix $tag/ -o $projectdir/$tarball
 
 echo "Signing the tarball..."
-[[ -z "$keyid" ]] && keyopt="-u $keyid"
+[[ -n "$keyid" ]] && keyopt="-u $keyid"
 gpg $keyopt --armor --output ${projectdir}/${tarball}.asc --detach-sig ${projectdir}/$tarball
 shasum -a 512 $tarball > ${projectdir}/${tarball}.sha512
 


dev/source-release.sh
```diff
@@ -36,7 +36,7 @@ usage () {
 }
 
 # Default repository remote name
-remote="origin"
+remote="apache"
```
RussellSpitzer, line 39: https://github.com/apache/iceberg/pull/4919/files#diff-ee187088d9f56516cc2abd9e4b4294a92d23dc09746db5b0976799131c293d2bR33

kbendick: I think this depends on where the script gets run (in the case of origin) and then how the user / box has their git remote set up.

Like mine is `upstream` vs apache, but I know most people / PMC members specifically probably use apache.

RussellSpitzer: Sorry, the link is meant to show this line above
```    echo "  -g      Specify Git remote name. Defaults to \"apache\""```

File: dev/source-release.sh [hunk 2]
```diff
@@ -112,7 +112,7 @@ tarball=$tag.tar.gz
 git archive $release_hash --worktree-attributes --prefix $tag/ -o $projectdir/$tarball
 
 echo "Signing the tarball..."
-[[ -z "$keyid" ]] && keyopt="-u $keyid"
+[[ -n "$keyid" ]] && keyopt="-u $keyid"
```
kbendick, line 115: 👍

dev/source-release.sh
kbendick: Ahhh! Ok then absolutely yes let’s make it apache. This typically gets run manually in somebody’s machine, so `origin` is very likely incorrect (typically one’s fork).

But if it says the default is `apache`, it should be that.

kbendick:
Also as an aside, during the _actual_ release process, we need to tag the SHA we build off of in master as `release-base-X.Y.0` for a new (non-patch) release.

I don’t think that step is automated anywhere but it’s how the next SNAPSHOT version gets its name (via a function in build.gradle).

dev/source-release.sh
kbendick: What I've done in the past is add a function that logs either way, so logs X on true and Y on false, so we'd get `No keyid is found in variable \\$keyid, skipping adding signature to tarball` in the untrue path (just some indicator it was taken).

RussellSpitzer: Yeah trick here is missing that arg still signs, just with the default key

kbendick: Oh haha you’re right. This is why I advocate for logging somewhat verbosely in shell scripts.

PR #4920, opened 2022-05-31
Python: Add Avro read path

Fokko:
Reads Avro file by first reading the headers, and then extracting the schema Then we convert the Avro schema into Iceberg, and read the actual binary using the schema visitor.

The binary decoder and codecs have been copied from `apache/avro` because I didn't want to depend on the library for just that. Also, `apache/avro` uses the general IO interface, while we have our own FileStream interface for reading files.

To not make the PR too big, I'm working on the follow-up PRs:

- Map the accessor onto an actual Python class
- Support Manifest version 1 & 2
- Support write/read schema


python/tests/avro/test_reader.py
```diff
@@ -0,0 +1,162 @@
...
+# under the License.
+# pylint: disable=W0212
+import pathlib
+
+from iceberg.avro.reader import AvroStructProtocol, DataFileReader
+from tests.io.test_io_base import LocalInputFile
+
+
+def test_read_header():
+    test_file_path = str(pathlib.Path(__file__).parent.resolve()) + "/manifest.avro"
```
samredai, line 25: What do you think of adding a `tests/avro/data` directory, and then using a `pytest.parametrize` decorator for this test. We could then include a collection of manifest files and test different combinations of `filename`, `expected_header_magic`, `expected_header_meta`, and `expected_header_sync`.

We could do this for `test_reader_data(...)` as well.

python/tests/avro/test_reader.py
```diff
@@ -0,0 +1,162 @@
...
+        "partition-spec": '[{"name":"VendorID","transform":"identity","source-id":1,"field-id":1000}]',
+    }
+    assert header.sync == b"B\x9e\xb0\xc3\x04,vG\xd1O\x8e\xf5\x9c\xdd\x17\x8d"
```
samredai, line 40: Should we also test `header.get_schema()`?

```suggestion
    assert header.sync == b"B\x9e\xb0\xc3\x04,vG\xd1O\x8e\xf5\x9c\xdd\x17\x8d"
    assert header.get_schema() == Schema(
        NestedField(field_id=0, name="status", field_type=IntegerType(), is_optional=False),
        NestedField(
            field_id=1, name="snapshot_id", field_type=LongType(), is_optional=True
        ),
        NestedField(
            field_id=2,
            name="data_file",
            field_type=StructType(
                fields=(
                    NestedField(
                        field_id=100,
                        name="file_path",
                        field_type=StringType(),
                        is_optional=False,
                    ),
                    NestedField(
                        field_id=101,
                        name="file_format",
                        field_type=StringType(),
                        is_optional=False,
                    ),
                    NestedField(
                        field_id=102,
                        name="partition",
                        field_type=StructType(
                            fields=(
                                NestedField(
                                    field_id=1000,
                                    name="VendorID",
                                    field_type=IntegerType(),
                                    is_optional=True,
                                ),
                            )
                        ),
                        is_optional=False,
                    ),
                    NestedField(
                        field_id=103,
                        name="record_count",
                        field_type=LongType(),
                        is_optional=False,
                    ),
                    NestedField(
                        field_id=104,
                        name="file_size_in_bytes",
                        field_type=LongType(),
                        is_optional=False,
                    ),
                    NestedField(
                        field_id=105,
                        name="block_size_in_bytes",
                        field_type=LongType(),
                        is_optional=False,
                    ),
                    NestedField(
                        field_id=108,
                        name="column_sizes",
                        field_type=MapType(
                            key_id=117,
                            key_type=IntegerType(),
                            value_id=118,
                            value_type=LongType(),
                            value_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=109,
                        name="value_counts",
                        field_type=MapType(
                            key_id=119,
                            key_type=IntegerType(),
                            value_id=120,
                            value_type=LongType(),
                            value_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=110,
                        name="null_value_counts",
                        field_type=MapType(
                            key_id=121,
                            key_type=IntegerType(),
                            value_id=122,
                            value_type=LongType(),
                            value_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=137,
                        name="nan_value_counts",
                        field_type=MapType(
                            key_id=138,
                            key_type=IntegerType(),
                            value_id=139,
                            value_type=LongType(),
                            value_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=125,
                        name="lower_bounds",
                        field_type=MapType(
                            key_id=126,
                            key_type=IntegerType(),
                            value_id=127,
                            value_type=BinaryType(),
                            value_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=128,
                        name="upper_bounds",
                        field_type=MapType(
                            key_id=129,
                            key_type=IntegerType(),
                            value_id=130,
                            value_type=BinaryType(),
                            value_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=131,
                        name="key_metadata",
                        field_type=BinaryType(),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=132,
                        name="split_offsets",
                        field_type=ListType(
                            element_id=133,
                            element_type=LongType(),
                            element_is_optional=False,
                        ),
                        is_optional=True,
                    ),
                    NestedField(
                        field_id=140,
                        name="sort_order_id",
                        field_type=IntegerType(),
                        is_optional=True,
                    ),
                )
            ),
            is_optional=False,
        ),
        schema_id=1,
        identifier_field_ids=[],
    )
```

python/tests/avro/test_reader.py
```diff
@@ -0,0 +1,162 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint: disable=W0212
+import pathlib
+
+from iceberg.avro.reader import AvroStructProtocol, DataFileReader
```
samredai, line 20: Style nit/opinion: I like to fully import the module that's being directly tested in a test file. I feel like it helps easily identify what's being tested with much less cognitive load (just look for `reader.*`. So what do you think of making this `from iceberg.avro import reader` and then using `reader.AvroStructProtocol` and `reader.DataFileReader` down below?

python/tests/utils/test_schema_conversion.py
```diff
@@ -229,3 +234,127 @@ def test_avro_list_required_record():
...
+
+def test_unknown_logical_type():
+    avro_logical_type = {"type": "bytes", "logicalType": "date"}
+    with pytest.raises(ValueError) as exc_info:
+        AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+
+    assert "Unknown logical/physical type combination:" in str(exc_info.value)
+
+
+def test_logical_map_with_invalid_fields():
```
samredai, line 342: I see a pattern here so we may be able to consolidate these into a handful of parametrized tests, making them easily extendable too. How about making a single function per method that's being tested?

- `AvroSchemaConversion()._convert_logical_type`
- `AvroSchemaConversion()._convert_logical_map_type`
- `AvroSchemaConversion()._convert_schema`
- `AvroSchemaConversion()._convert_field`
- `AvroSchemaConversion()._convert_record_type`
- `AvroSchemaConversion()._convert_array_type`

As an example, for `AvroSchemaConversion()._convert_logical_type`:
```py
@pytest.mark.parametrize(
    "avro_logical_type,expected",
    [
        ({"type": "int", "logicalType": "date"}, DateType()),
        (
            {"type": "bytes", "logicalType": "decimal", "precision": 19, "scale": 25},
            DecimalType(precision=19, scale=25),
        ),
        ...,
    ],
)
def test_schema_conversion_convert_logical_type(avro_logical_type, expected):
    assert AvroSchemaConversion()._convert_logical_type(avro_logical_type) == expected
```

python/tests/utils/test_schema_conversion.py
```diff
@@ -229,3 +234,127 @@ def test_avro_list_required_record():
...
+    assert actual == expected
+
+
+def test_convert_date_type():
+    avro_logical_type = {"type": "int", "logicalType": "date"}
+    actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+    assert actual == DateType()
+
+
+def test_unknown_logical_type():
```
samredai, line 334: nit: One-liner docstrings may be helpful, especially since they'll come up in IDE searches. Here for example I'd add something like:
```py
def test_unknown_logical_type():
    """Test raising a ValueError when converting an unknown logical type as part of an Avro schema conversion"""
    ...
```

samredai (commented):
This is awesome @Fokko! I left some comments. For some of the comments in `src/iceberg/avro/`, I know we're vendoring some of that so please feel free to ignore any nit/style comments there. I'm super excited that we'll have a read path that utilizes all standard lib stuff. 😄

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        return float(STRUCT_FLOAT.unpack(self.read(4))[0])
+
+    def read_double(self) -> float:
+        """
+        A double is written as 8 bytes.
+        The double is converted into a 64-bit integer using a method equivalent to
+        Java's doubleToLongBits and then encoded in little-endian format.
+        """
+        return float(STRUCT_DOUBLE.unpack(self.read(8))[0])
```
rdblue, line 93: Do any of these methods come from another library, or are they all mostly new? Is there code you were basing this on? Knowing that would help review.

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,284 @@
...
+    StructType,
+    TimestampType,
+    TimestamptzType,
+    TimeType,
+)
+from iceberg.utils.schema_conversion import AvroSchemaConversion
+
+
+@dataclass(frozen=True)
+class AvroStructProtocol(StructProtocol):
```
rdblue, line 63: It looks like this implements the protocol, but is not a protocol itself. Should it be `AvroStruct` instead?

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,284 @@
...
+    _data: List[Union[Any, StructProtocol]] = field(default_factory=list)
+
+    def set(self, pos: int, value: Any) -> None:
+        self._data[pos] = value
+
+    def get(self, pos: int) -> Any:
+        return self._data[pos]
+
+
+class _AvroReader(SchemaVisitor[Union[AvroStructProtocol, Any]]):
```
rdblue, line 73: It looks like this calls a decoder when it visits a schema, but I was expecting an implementation that creates a reader tree that accepts a decoder, like the other one. Why did you decide to go with this approach over the other one?

python/tests/utils/test_schema_conversion.py
```diff
@@ -229,3 +234,127 @@ def test_avro_list_required_record():
     iceberg_schema = AvroSchemaConversion().avro_to_iceberg(avro_schema)
 
     assert expected_iceberg_schema == iceberg_schema
+
+
+def test_resolve_union():
```
rdblue, line 239: Should these conversion test cases go in a separate PR? It doesn't seem related to the reader.

python/src/iceberg/avro/codec.py
```diff
@@ -0,0 +1,179 @@
...
+
+# snappy
+try:
+    import snappy
+
+    has_snappy = True
+except ImportError:
+    has_snappy = False
+
+# z-standard
```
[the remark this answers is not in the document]
Fokko: Not really :)

python/src/iceberg/avro/codec.py
```diff
@@ -0,0 +1,179 @@
...
+import zlib
+from typing import Dict, Tuple, Type
+
+from iceberg.avro.decoder import BinaryDecoder
+from iceberg.io.memory import MemoryInputStream
+
+STRUCT_CRC32 = struct.Struct(">I")  # big-endian unsigned int
+
+
+def _check_crc32(bytes_: bytes, checksum: bytes) -> None:
```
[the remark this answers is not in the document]
Fokko: Nice, thanks!

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def read_int(self) -> int:
+        """
+        int and long values are written using variable-length, zigzag coding.
+        """
+        return self.read_long()
+
+    def read_long(self) -> int:
+        """
+        int and long values are written using variable-length, zigzag coding.
```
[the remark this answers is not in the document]
Fokko: There is no minimal nit threshold from my side. I like the suggestion, thanks!

python/tests/utils/test_schema_conversion.py
Fokko: Yes, let me split those out

python/tests/avro/test_reader.py
Fokko: It works both for me. I have a slight preference for just having everything imported at the top because it makes the code below more concise (having to repeat the file all the time).

python/tests/avro/test_reader.py
Fokko: I've added it to the top of the file to not bloat the conftest.py

python/tests/utils/test_schema_conversion.py
Fokko: Personally, I'm not a fan of parameterized tests:

- I find them hard to read with all the (curly)braces
- If you have many tests, and the last one is failing, you have to rerun the earlier ones all the time, which is kind of annoying if you have breakpoints everywhere.
- The parameterized arguments are evaluated every time, even if you run an unrelated test.

I don't mind a few additional tests.

python/tests/utils/test_schema_conversion.py
Fokko: Added it, thanks!

python/src/iceberg/avro/reader.py
Fokko: Hey @rdblue. This made the most sense to me at the time. Which one is the other one? I don't have a strong opinion on this way or the other. Probably you have done this more often than me :)

python/tests/utils/test_schema_conversion.py
Fokko: https://github.com/apache/iceberg/pull/4972

python/src/iceberg/avro/decoder.py
rdblue: Great, that helps. We should also document where the code came from with a comment at the top, and also start collecting the LICENSE and NOTICE files.

python/src/iceberg/avro/reader.py
rdblue: I suggested the other approach for a few reasons. First, I think when it is reasonable to match the approach taken by the Java codebase, that's a good idea. That way we don't have completely different implementations to validate and maintain.

Second, this approach traverses the schema for each record read. That is inefficient compared to building a tree of readers that handle this. The reader tree approach allows us to create basically a state machine that is ready to read records.

Last, this is harder to update. The next step is to reconcile the differences between the read and write schemas. Doing that for every record is hard to write with this approach. Same thing with reusing structs, dicts, and lists.

python/tests/avro/data/README.md
```diff
@@ -0,0 +1,571 @@
...
+ - Unless required by applicable law or agreed to in writing, software
+ - distributed under the License is distributed on an "AS IS" BASIS,
+ - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ - See the License for the specific language governing permissions and
+ - limitations under the License.
+ -->
+
+# Avro tests
+
+This directory contains a manifest file to running tests against.
```
samredai, line 20: s/to running/for running

python/NOTICE
```diff
@@ -6,3 +6,22 @@ This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
 --------------------------------------------------------------------------------
+
+This project includes code from Apache Avro from the Apache Software Foundation with
+the following copyright notice:
```
rdblue, line 11: We don't need to add copyright notices into NOTICE. This only needs to include relevant NOTICE contents from other projects. For Avro, there's nothing that applies to the code that we're including because all of the NOTICE contents are for the C# implementation. So you can remove this.

We do need to update the LICENSE file for Python to cover the parts of Avro that we copied. Here's how we did that for the Java project: https://github.com/apache/iceberg/blob/master/LICENSE#L214-L223

python/src/iceberg/avro/codec.py
```diff
@@ -0,0 +1,196 @@
...
+except Exception as e:
+    logger.exception(f"snappy compression not available; {e}")
+
+has_zstandard = False
+try:
+    import zstandard as zstd
+
+    has_zstandard = True
+except ImportError:
+    logger.info("zstandard compression not available, please install with zstandard")
```
rdblue, line 69: Shouldn't we only log this if the user tries to read a file with Zstd compression and it isn't installed? I always find it annoying when code tells me that optional things aren't available because I have chosen not to install them.

python/src/iceberg/avro/codec.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class NullCodec(Codec):
+    @staticmethod
+    def compress(data: bytes) -> Tuple[bytes, int]:
+        return data, len(data)
+
+    @staticmethod
+    def decompress(readers_decoder: BinaryDecoder) -> BinaryDecoder:
+        _ = readers_decoder.read_long()
```
rdblue, line 109: Compression stores the length even if it is uncompressed?

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+)
+
+
+@dataclass(frozen=True)
+class AvroFileHeader:
+    magic: bytes
+    meta: dict[str, str]
+    sync: bytes
+
+    def get_compression_codec(self) -> type[Codec]:
```
rdblue, line 64: Is it a convention in python to use `get_` in method names? In Java, we tend not to use it because it can usually be replaced with a better verb (like find or fetch) or isn't helpful and can be omitted.

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+        """Get the file's compression codec algorithm from the file's metadata."""
+        codec_key = "avro.codec"
```
rdblue, line 66: Could this be a constant somewhere rather than redefining it in this method?

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+
+            if codec_name not in KNOWN_CODECS:
+                raise ValueError(f"Unsupported codec: {codec_name}. (Is it installed?)")
+
+            return KNOWN_CODECS[codec_name]
+        else:
+            return NullCodec
+
+    def get_schema(self) -> Schema:
+        schema_key = "avro.schema"
```
rdblue, line 78: Same here. Use a constant?

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+        if new_block > 0:
+            return self.__next__()
+        else:
+            raise StopIteration
+
+    def _read_header(self) -> AvroFileHeader:
+        self.input_stream.seek(0, SEEK_SET)
+        reader = visit(META_SCHEMA, ConstructReader())
+        _header = reader.read(self.decoder)
+        meta = {k: v.decode("utf-8") for k, v in _header.get(1).items()}
```
rdblue, line 156: I know that it is more correct to use the schema from the Avro spec, but as long as you're converting every binary to a UTF-8 string anyway, why not just update the header struct's type so the reader reads the bytes as a string directly?

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+        self.header = self._read_header()
+        self.schema = self.header.get_schema()
+        self.file_length = len(self.input_file)
+        self.reader = visit(self.schema, ConstructReader())
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.input_stream.close()
+
+    def __iter__(self) -> AvroFile:
```
rdblue, line 123: Is this equivalent to `iterator` in Java? If so, then structuring this class to return `self` rather than a reader means that you can't share the instance. That's not a problem, but the best practice in Java is to use a separate `Iterator` class that has its own open stream and state. I'm wondering if that's the best practice in Python as well?

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+        return self
+
+    def _read_block(self) -> int:
+        if self.block > 0:
+            sync_marker_len = len(self.header.sync)
```
rdblue, line 128: `len(self.header.sync)` is already `SYNC_SIZE`

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+        self._input_stream = input_stream
+
+    def read(self, n: int) -> bytes:
+        """
+        Read n bytes.
+        """
+        if n < 0:
+            raise ValueError(f"Requested {n} bytes to read, expected positive integer.")
+        read_bytes = self._input_stream.read(n)
+        if len(read_bytes) != n:
```
rdblue, line 48: Is this a guarantee from `InputStream`? I thought that it was possible to return fewer bytes.

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+        self.block_decoder = self.header.get_compression_codec().decompress(self.decoder)
+        self.block_position = 0
+        return self.block_records
+
+    def __next__(self) -> AvroStruct:
+        if self.block_position < self.block_records:
+            self.block_position = self.block_position + 1
+            return self.reader.read(self.block_decoder)
+
+        new_block = self._read_block()
```
rdblue, line 145: `new_block` is actually the number of records in the block. This name doesn't seem to fit, although the logic based on it looks correct.

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,160 @@
...
+        if self.block > 0:
+            sync_marker_len = len(self.header.sync)
+            sync_marker = self.decoder.read(sync_marker_len)
+            if sync_marker != self.header.sync:
+                raise ValueError(f"Expected sync bytes {self.header.sync!r}, but got {sync_marker!r}")
+            if self.is_EOF():
+                raise StopIteration
+        self.block = self.block + 1
+        self.block_records = self.decoder.read_long()
+        self.block_decoder = self.header.get_compression_codec().decompress(self.decoder)
```
rdblue, line 136: It's a bit odd to me that the decompression codec accepts and produces a decoder, rather than decompressing bytes and returning bytes. It's okay for now, but I think it mixes Avro-specific classes into the compression classes, which don't need to be specific to Avro.

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,318 @@
...
+)
+from iceberg.utils.singleton import Singleton
+
+
+@dataclass(frozen=True)
+class AvroStruct(StructProtocol):
+    _data: list[Any | StructProtocol] = field()
+
+    def set(self, pos: int, value: Any) -> None:
+        self._data[pos] = value
```
rdblue, line 67: Where is `_data` initialized? If I create a list and try to assign to a position, it fails with `IndexError: list assignment index out of range`:

```python
In [1]: _data = []

In [2]: _data[3] = "foo"
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 _data[3] = "foo"

IndexError: list assignment index out of range
```

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+class TimestampReader(Reader):
+    def read(self, decoder: BinaryDecoder) -> datetime:
+        return decoder.read_timestamp_micros_from_long()
+
+
+class TimestamptzReader(Reader):
+    def read(self, decoder: BinaryDecoder) -> datetime:
+        return decoder.read_timestamp_micros_from_long()
```
rdblue, line 126: I think this needs to call a variant that sets `tzInfo` to UTC.

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,318 @@
...
+    def read(self, decoder: BinaryDecoder) -> Decimal:
+        return decoder.read_decimal_from_bytes(self.precision, self.scale)
+
+
+@dataclass(frozen=True)
+class OptionReader(Reader):
+    option: Reader = field()
+
+    def read(self, decoder: BinaryDecoder) -> Any | None:
+        # For the Iceberg spec it is required to set the default value to null
```
rdblue, line 166: This comment is true, but for a long time there was nothing in Avro that actually enforced this so I've definitely seen out of order schemas. I would opt to figure out which option is non-null and test for it directly:

```python
class OptionReader(Reader):
    option_reader: Reader = field()
    non_null_option: int = field()

    def read(self, decoder: BinaryDecoder) -> Any | None:
        if decoder.read_int() == non_null_option:
            return self.option_reader.read(decoder)
        return None
```

python/src/iceberg/avro/reader.py
rdblue: Okay, I see that the data list is passed in. Seems okay for now.

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,318 @@
...
+class ListReader(Reader):
+    element: Reader
+
+    def read(self, decoder: BinaryDecoder) -> list:
+        read_items = []
+        block_count = decoder.read_long()
+        while block_count != 0:
+            if block_count < 0:
+                block_count = -block_count
+                # We ignore the block size for now
```
rdblue, line 198: I think the intent is for `read` to always skip the block size. It is only used for data skipping.

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,318 @@
...
+
+@dataclass(frozen=True)
+class MapReader(Reader):
+    key: Reader
+    value: Reader
+
+    def read(self, decoder: BinaryDecoder) -> dict:
+        read_items = {}
+        block_count = decoder.read_long()
+        if block_count < 0:
```
rdblue, line 214: Should this be moved into the `while` loop like it is in `ListReader`?

python/src/iceberg/utils/singleton.py
```diff
@@ -0,0 +1,30 @@
...
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+from abc import ABCMeta
+from typing import Dict
+
+
+class Singleton(type, metaclass=ABCMeta):
```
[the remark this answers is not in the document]
rdblue: Can you help me understand what this is doing? I understand the other `Singleton` class. Why is this needed?

python/tests/avro/test_reader.py
```diff
@@ -0,0 +1,347 @@
...
+    ListType,
+    LongType,
+    MapType,
+    NestedField,
+    StringType,
+    StructType,
+)
+from tests.io.test_io_base import LocalInputFile
+
+MANIFEST_SCHEMA = Schema(
```
rdblue, line 35: Will this move when we add a manifest reader class?

python/tests/avro/test_reader.py
```diff
@@ -0,0 +1,347 @@
...
+                    None,
+                    [4],
+                    0,
+                ]
+            ),
+        ]
+    )
+
+
+def test_singleton():
```
rdblue, line 343: Should this be moved to a different suite? Seems strange to include it here in the reader tests.

python/tests/avro/test_reader.py
```diff
@@ -0,0 +1,347 @@
...
+        "partition-spec": '[{"name":"VendorID","transform":"identity","source-id":1,"field-id":1000}]',
+    }
+    assert header.get_schema() == MANIFEST_SCHEMA
+    assert header.sync == b"B\x9e\xb0\xc3\x04,vG\xd1O\x8e\xf5\x9c\xdd\x17\x8d"
+
+
+def test_reader_data():
```
rdblue, line 221: I like that this test is practical, but I'm generally prefer not to add binary files to a git repository. They can't be removed in the future.

Is there another good way to test this? Maybe we could pull in fastavro (or the ASF one) as a test dependency and validate that when we write files with it we can read it with this library? That seems like a better overall test for the behavior we're adding here, and it doesn't require either a new runtime dependency (only test) or adding files to the git repo.

python/src/iceberg/utils/schema_conversion.py
```diff
@@ -147,7 +145,14 @@ def _resolve_union(self, type_union: dict | list | str) -> tuple[str | dict[str,
...
 
-        is_optional = "null" in avro_types
+        # For the Iceberg spec it is required to set the default value to null
+        # From https://iceberg.apache.org/spec/#avro
+        # Optional fields must always set the Avro field default value to null.
+        #
+        # This means that null has to come first:
+        # https://avro.apache.org/docs/current/spec.html
+        # type of the default value must match the first element of the union.
+        is_optional = "null" == avro_types[0]
```
rdblue, line 155: I'd prefer to be a bit more careful and work with types where `null` isn't first. I know it's required, but we should be strict with what this library produces and liberal with what it accepts.

python/src/iceberg/avro/reader.py
```diff
@@ -0,0 +1,318 @@
...
+        while block_count != 0:
+            for _ in range(block_count):
+                key = self.key.read(decoder)
+                read_items[key] = self.value.read(decoder)
+            block_count = decoder.read_long()
+
+        return read_items
+
+
+class ConstructReader(SchemaVisitor[Reader]):
```
rdblue, line 229: This looks great!

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+        return unix_epoch_datetime + timedelta
+
+    def read_timestamp_micros_from_long(self) -> datetime.datetime:
+        """
+        long is decoded as python datetime object which represents
+        the number of microseconds from the unix epoch, 1 January 1970.
+        """
+        timestamp_micros = self.read_long()
+        timedelta = datetime.timedelta(microseconds=timestamp_micros)
+        unix_epoch_datetime = datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone.utc)
```
rdblue, line 192: This value is incorrect for Iceberg because Iceberg's `timestamp` type should not have `tzinfo` set. Can you split this into `read_timestamp_micros_from_long` and `read_timestamptz_micros_from_long`?

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def read_timestamp_millis_from_long(self) -> datetime.datetime:
+        """
+        long is decoded as python datetime object which represents
+        the number of milliseconds from the unix epoch, 1 January 1970.
+        """
+        timestamp_millis = self.read_long()
+        timedelta = datetime.timedelta(microseconds=timestamp_millis * 1000)
+        unix_epoch_datetime = datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=datetime.timezone.utc)
+        return unix_epoch_datetime + timedelta
```
rdblue, line 183: Can you refactor the date/time/timestamp handling here to share the methods in `utils/datetime.py`? I'd like to have consistent implementations for converting between micros and datetime objects.

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def read_decimal_from_fixed(self, precision: int, scale: int, size: int) -> decimal.Decimal:
+        """
+        Decimal is encoded as fixed. Fixed instances are encoded using the
+        number of bytes declared in the schema.
+        """
+        datum = self.read(size)
+        unscaled_datum = 0
+        msb = struct.unpack("!b", datum[0:1])[0]
+        leftmost_bit = (msb >> 7) & 1
```
rdblue, line 107: @Fokko, do you know why this uses `ord` and `unpack`? I don't think it needs to be this complicated. If you access a byte, then it is returned as the unsigned value.

```python
datum = b'\x01\x02\x03\xFF'
datum[0] # => 1
datum[-1] # => 255
```

Maybe I'm missing something, but I don't see how the line to produce `msb` isn't equivalent to `datum[0]`. I think this could be:

```java
    unscaled_datum = 0
    if datum[0] & 0x80 != 0:
        # negative two's complement
        ...
    else:
        # positive
        ...
```

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+        if leftmost_bit == 1:
+            modified_first_byte = ord(datum[0:1]) ^ (1 << 7)
+            datum = bytearray([modified_first_byte]) + datum[1:]
+            for offset in range(size):
+                unscaled_datum <<= 8
+                unscaled_datum += ord(datum[offset : 1 + offset])
```
rdblue, line 113: Similarly, I think this could be `datum[offset]` rather than producing another binary and converting it with `ord`.

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+        Decimal is encoded as fixed. Fixed instances are encoded using the
+        number of bytes declared in the schema.
+        """
+        datum = self.read(size)
+        unscaled_datum = 0
+        msb = struct.unpack("!b", datum[0:1])[0]
+        leftmost_bit = (msb >> 7) & 1
+        if leftmost_bit == 1:
+            modified_first_byte = ord(datum[0:1]) ^ (1 << 7)
+            datum = bytearray([modified_first_byte]) + datum[1:]
```
rdblue, line 110: Rather than creating a new byte array, this could just handle the first byte separately. Then it also wouldn't need to rely on `unscaled_datum <<= 8` being a no-op the first time through the loop.

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,193 @@
...
+            for offset in range(size):
+                unscaled_datum <<= 8
+                unscaled_datum += ord(datum[offset : 1 + offset])
+            unscaled_datum += pow(-2, (size * 8) - 1)
+        else:
+            for offset in range(size):
+                unscaled_datum <<= 8
+                unscaled_datum += ord(datum[offset : 1 + offset])
+
+        original_prec = decimal.getcontext().prec
```
rdblue, line 120: There's no need to modify the context like this. We have a more direct way to produce the decimal value in `utils.decimal.py`. Can you update this to use `unscaled_to_decimal`?

python/src/iceberg/avro/reader.py
rdblue: I commented about this in the decoder.

rdblue (changes requested):
@Fokko, this is great! I made a few comments but it's really close.

python/src/iceberg/avro/decoder.py
rdblue: There may also be a much better way to do this, according to the answers on two's complement here: https://stackoverflow.com/questions/1604464/twos-complement-in-python

python/tests/avro/data/README.md
Fokko: Thanks!

python/src/iceberg/utils/singleton.py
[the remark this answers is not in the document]
Fokko: Hey Ryan, since you've merged https://github.com/apache/iceberg/pull/5008 we can re-use that singleton. I'll update the code.

I've also added a docstring for the singleton module to elaborate why we need it:
```
This is a singleton metaclass that can be used to cache and re-use existing objects

In the Icebeg codebase we have a lot of objects that are stateless (for example Types such as StringType,
BooleanType etc). FixedTypes have arguments (eg. Fixed[22]) that we also make part of the key when caching
the newly created object.

The Singleton uses a metaclass which essentially defines a new type. When the Type gets created, it will first
evaluate the `__call__` method with all the arguments. If we already initialized a class earlier, we'll just
return it.

More information on metaclasses: https://docs.python.org/3/reference/datamodel.html#metaclasses
```

python/src/iceberg/avro/codec.py
Fokko: Yes, I've refactored the whole codecs into a separate module. I couldn't handle all the `has_{snappy,bzip2,...}` to keep all the imports at the top. I've refactored them into separate files, and changed the dynamic dict containing all the codecs to a static one. Instead of logging, I'm initializing an empty class that will throw an error when being used.

I also think we might want to do re-using of buffers, later on, this will also cause the classes to grow in the number of lines. So maybe better to split them now.

python/src/iceberg/avro/codec.py
Fokko: I had to check, and this is the case. I've added a test with a snappy and a null-codec.

python/src/iceberg/avro/file.py
Fokko: No, that isn't a convention. I've removed it since it can be ommitted.

python/src/iceberg/avro/file.py
Fokko: I've moved it to the file itself.

python/src/iceberg/avro/file.py
Fokko: I've combined Java's closable and the iterator on this one. I could refactor this one a bit, for example by for clarity by splitting out the block

python/src/iceberg/avro/file.py
Fokko: Thanks!

python/src/iceberg/avro/decoder.py
Fokko: I don't think that's possible. We know how many bytes we need to read, if we got fewer that indicates that something is off.

python/src/iceberg/avro/file.py
Fokko: I've wrapped the block in a block object, which splits out some of the logic

python/src/iceberg/avro/reader.py
Fokko: We can optimize this later by reusing objects. But I think it makes more sense to get everything working, and then do some profiling

python/src/iceberg/avro/reader.py
Fokko: oof

We lose this information when we go to the Iceberg schema, since we don't keep the position there. This would either require storing this information in the Iceberg type, which I don't really like. Or build the reader from the Avro schema directly.

python/src/iceberg/avro/reader.py
Fokko: Yes, great catch!

python/tests/avro/test_reader.py
Fokko: I've moved it to `test_singleton` 👍🏻

python/src/iceberg/utils/singleton.py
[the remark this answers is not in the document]
rdblue: Thanks, I think I get why this is using `__call__` now instead of `__new__`.

python/src/iceberg/avro/codec.py
rdblue: This is actually an artifact of how the decoder is used. Since the block length is right before the block, the decoder does the right thing when you call `read_bytes`. But in this case there's no need to create a new decoder. It just needs to consume the length bytes and return the existing one. It makes sense, although I do think it is a bit strange to couple the compression API with decoders.

python/src/iceberg/avro/decoder.py
rdblue: Actually, I think that we can have this happen. `InputStream` is a protocol to describe our requirements for IO classes, which is intended to allow passing something that extends `IOBase`. The [docs for `RawIOBase`](https://docs.python.org/3/library/io.html#io.RawIOBase) show that there isn't a guarantee that all the bytes will be returned:

> Fewer than size bytes may be returned if the operating system call returns fewer than size bytes.

I think we should handle small reads. Maybe add a `read_fully` method since we always want to get the exact number of bytes that are expected?

python/src/iceberg/avro/reader.py
rdblue: Ah, I see. Let me think about this more, but it should be fine for now.

python/src/iceberg/utils/schema_conversion.py
rdblue: Nevermind, based on the comment above. I understand the limitation now.

python/src/iceberg/avro/decoder.py
Fokko: Added!

python/src/iceberg/avro/decoder.py
Fokko: Yes, thanks for pointing out!

python/src/iceberg/avro/reader.py
Fokko: I'm not able to judge how often this happens in practice. Once we run into it, we can change it

python/src/iceberg/avro/decoder.py
Fokko: LOL, you're right. I just copied this code. We can simplify it to:
```python
    def read_decimal_from_fixed(self, precision: int, scale: int, size: int) -> decimal.Decimal:
        """
        Decimal is encoded as fixed. Fixed instances are encoded using the
        number of bytes declared in the schema.
        """
        data = self.read(size)
        unscaled_datum = int.from_bytes(data, byteorder="big", signed=True)
        decimal_context.prec = precision
        return decimal_context.create_decimal(unscaled_datum).scaleb(
            -scale, decimal_context
        )
```
And that's what I did. Also added a couple of tests (should add more).

Fokko: I read the comment below, it is now:
```python
    def read_decimal_from_fixed(self, precision: int, scale: int, size: int) -> decimal.Decimal:
        """
        Decimal is encoded as fixed. Fixed instances are encoded using the
        number of bytes declared in the schema.
        """
        data = self.read(size)
        unscaled_datum = int.from_bytes(data, byteorder="big", signed=True)
        return unscaled_to_decimal(unscaled_datum, scale)
```

python/src/iceberg/avro/decoder.py
Fokko: Yes, I was unaware of the helper functions. Much better now

python/src/iceberg/avro/decoder.py
rdblue: Wow, quite a bit smaller!

python/tests/avro/test_reader.py
Fokko: I'm also not a fan of binary files in a repo, but it is a very solid approach when it comes to testing, especially when we're evolving the spec over time.

Let me pull in fastavro and generate some files in a temp folder 👍🏻

python/NOTICE
```diff
@@ -6,3 +6,22 @@ This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
 
 --------------------------------------------------------------------------------
+
+This project includes code from Apache Avro from the Apache Software Foundation with
```
rdblue, line 10: The update to LICENSE looks good. And now we can remove this because there isn't anything in the Avro NOTICE that needs to be included. ASF projects don't need to add the boilerplate, too.

python/README.md
```diff
@@ -1,3 +1,4 @@
+
```
rdblue, line 1: Looks like an accidental change made it in?

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,167 @@
...
+        return self.read_bytes().decode("utf-8")
+
+    def read_date_from_int(self) -> date:
+        """
+        int is decoded as python date object.
+        int stores the number of days from
+        the unix epoch, 1 January 1970 (ISO calendar).
+        """
+        days_since_epoch = self.read_int()
+        return date(1970, 1, 1) + timedelta(days_since_epoch)
```
rdblue, line 136: Minor: I'd prefer to use `days_to_date` just like this uses `micros_to_time` and `micros_to_timestamp`.

python/src/iceberg/utils/datetime.py
```diff
@@ -36,6 +39,15 @@ def micros_to_days(timestamp: int) -> int:
     return (datetime.fromtimestamp(timestamp / 1_000_000) - EPOCH_TIMESTAMP).days
 
 
+def micros_to_time(micros: int) -> time:
```
rdblue, line 42: Jun just added `time_from_micros` below that is basically the same thing (though I like this version better because it uses `divmod`. Can you replace the use of `time_from_micros` with this?

python/src/iceberg/utils/datetime.py
```diff
@@ -77,6 +89,12 @@ def timestamptz_to_micros(timestamptz_str: str) -> int:
     raise ValueError(f"Invalid timestamp with zone: {timestamptz_str} (must be ISO-8601)")
 
 
+def micros_to_timestamp(micros: int, tzinfo: timezone | None = None):
+    dt = timedelta(microseconds=micros)
+    unix_epoch_datetime = datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=tzinfo)
```
rdblue, line 94: Can you use `EPOCH_TIMESTAMP` and `EPOCH_TIMESTAMPTZ` instead of creating a new `datetime` here?

```python
def micros_to_timestamp(micros: int):
    return EPOCH_TIMESTAMP + timedelta(microseconds=micros)

def micros_to_timestamptz(micros: int):
    return EPOCH_TIMESTAMPTZ + timedelta(microseconds=micros)
```

python/src/iceberg/avro/codec.py
Fokko: Ah, you're right and this doesn't look good. This is actually a bug. I'll refactor this, including passing the bytes to the compression API instead of the decoder, which indeed doesn't make much sense.

python/src/iceberg/avro/decoder.py
Fokko: Alright, I tried to implement this which isn't that hard, but we also have other interfaces that might not have implemented the `read_fully`, breaking the protocol of InputStream. This happens with PyArrows NativeFile as well:
```python
    def test_pyarrow_input_file():
        """Test reading a file using PyArrowFile"""
    
        with tempfile.TemporaryDirectory() as tmpdirname:
            file_location = os.path.join(tmpdirname, "foo.txt")
            with open(file_location, "wb") as f:
                f.write(b"foo")
    
            # Confirm that the file initially exists
            assert os.path.exists(file_location)
    
            # Instantiate the input file
            absolute_file_location = os.path.abspath(file_location)
            input_file = PyArrowFile(location=f"{absolute_file_location}")
    
            # Test opening and reading the file
            f = input_file.open()
>           assert isinstance(f, InputStream)  # Test that the file object abides by the InputStream protocol
E           assert False
E            +  where False = isinstance(<pyarrow.lib.NativeFile object at 0x111bbf6f0>, InputStream)
```

Looks like pyarrow does return the exact number of bytes, and also has a `read1` method fwiw:

image

python/README.md
Fokko: Oops, thanks!

python/src/iceberg/avro/decoder.py
Fokko: I like it 👍🏻  Updated

python/src/iceberg/utils/datetime.py
Fokko: Sure thing, updated 👍🏻

python/src/iceberg/avro/codecs/codec.py
```diff
@@ -0,0 +1,33 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+
+class Codec(ABC):
```
rdblue, line 22: Minor: would it make sense to put this in the `__init__.py` file?

python/src/iceberg/avro/decoder.py
rdblue: I might not understand you correctly, but it sounds like you're adding `read_fully` to the Protocol and that's why PyArrow's `NativeFile` doesn't conform to `InputStream`? I was thinking that we should add an IO util class that has a `read_fully` method, like we have in Java: https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/io/IOUtil.java

python/src/iceberg/avro/decoder.py
```diff
@@ -0,0 +1,166 @@
...
+        return self.read_bytes().decode("utf-8")
+
+    def read_date_from_int(self) -> date:
+        """
+        int is decoded as python date object.
+        int stores the number of days from
+        the unix epoch, 1 January 1970 (ISO calendar).
+        """
+        days_to_date = self.read_int()
+        return date(1970, 1, 1) + timedelta(days_to_date)
```
rdblue, line 135: What about adding `days_to_date` to `datetime.py`? Then you could use `EPOCH_DATE` and this would be reusable.

python/src/iceberg/avro/file.py
```diff
@@ -0,0 +1,181 @@
...
+    file_length: int
+    reader: StructReader
+
+    decoder: BinaryDecoder
+    block: Block | None = None
+
+    def __init__(self, input_file: InputFile) -> None:
+        self.input_file = input_file
+
+    def __enter__(self):
```
rdblue, line 122: As a follow up, I think we should start a base class for our readers that handles `__iter__`, `__enter__`, and `__exit__`. This should probably use `threading.local()` to ensure that `with` is thread-safe if the file is shared, and I think we can make `__iter__` return an iterator class. But these all work great for now.

python/src/iceberg/avro/codecs/codec.py
Fokko: Some projects do this, and some don't :) For me, it makes sense to add base classes in the init. Mostly because they need to be loaded anyway, and by adding them to the `__init__.py` they are read when you access the module. WDYT?

python/src/iceberg/avro/file.py
Fokko: I'm happy to do this, but I would suggest doing this in a separate PR. We could also extend the Input- and OutputFile. That `__enter__` calls `open()` or `create()`.

With regard to the threading, I think that's another can of worms. (At least as a start) reading a file should not be considered thread-safe, and we should not share the file across threads. I would rather suggest reading the files in parallel using something like [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.starmap). We could also split out the blocks in the Avro file if we like. Another option would be to make the read [async](https://docs.python.org/3/library/asyncio.html). We could also go fancy and go for a [async iterator](https://peps.python.org/pep-0525/). But looking at the size of this PR, I think we should split that out. Let me know if you feel otherwise.

python/src/iceberg/avro/decoder.py
Fokko: Auch, that one slipped through the cracks. I just updated the code

python/src/iceberg/avro/codecs/codec.py
rdblue: +1 for adding it to `__init__.py`. Fewer files cuts down on load time and we will need to load it anyway.

python/src/iceberg/avro/file.py
rdblue: Yeah, definitely as a separate PR. I didn't mean to suggest doing it now.

Fokko:
@rdblue I've fixed the linting violation introduced by https://github.com/apache/iceberg/pull/5055

python/src/iceberg/avro/decoder.py
Fokko: Then we should add another class in between. Adding it to the protocol means that it should be implemented to NativeFile as well. Happy to do this, but we can also do this in a separate PR.

rdblue: Yeah, I agree. We should have some additional methods that help fill the gap between the IO API and what we need for usage.

rdblue (approved):

rdblue:
Thanks, @Fokko! Everything looks great in here. Thanks for updating the tests. Looking forward to the next steps!

PR #4922, opened 2022-05-31
Core: Assign main branch to table's current snapshot if there is no main but there is a current table state

amogh-jahagirdar:
As part of https://github.com/apache/iceberg/pull/4428/files#r884916051 and other PRs, a common theme has been around checking if main even exists in different API implementations. In this PR, main will always be set in a table metadata's refs. If refs don't exist or the main ref does not exist for some reason when parsing, it will be set to the table's current snapshot. 

This change also allows -1 to be set on different table metadata builder APIs to allow main to be created when there is no current table snapshot.

I'm still mulling over the implications of this change, the benefit is that this change allows different operations to be implemented with the assumption that main exists. However, I think the real long term solution is that we produce a snapshot upon table creation and assign main at that point. In this change, we only allow -1 for main, but this prevents the ability to create a branch before a snapshot is produced on main, which is a bit awkward but I think is needed until we produce a snapshot upon table creation. 

@rdblue @jackye1995 let me know your thoughts! If we determine that it makes more sense to just explicitly handle the non-existing main branch cases, then I will just close this PR.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
@@ -174,7 +174,7 @@ public abstract class BaseMetastoreTableOperations implements TableOperations {
   protected void refreshFromMetadataLocation(String newLocation, Predicate<Exception> shouldRetry,
                                              int numRetries) {
     refreshFromMetadataLocation(newLocation, shouldRetry, numRetries,
-        metadataLocation -> TableMetadataParser.read(io(), metadataLocation));
+        metadataLocation -> TableMetadata.buildFromLocation(io(), metadataLocation).build());
   }
 
   protected void refreshFromMetadataLocation(String newLocation, Predicate<Exception> shouldRetry,
--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
@@ -31,6 +31,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.iceberg.exceptions.ValidationException;
+import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Objects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
@@ -755,6 +756,10 @@ public class TableMetadata implements Serializable {
     return new Builder(base);
   }
 
+  static Builder buildFromLocation(FileIO io, String location) {
+    return buildFrom(TableMetadataParser.read(io, location));
+  }
+
   public static class Builder {
     private static final int LAST_ADDED = -1;
 
@@ -1165,6 +1170,11 @@ public class TableMetadata implements Serializable {
     }
 
     public TableMetadata build() {
+      if (refs.isEmpty() && currentSnapshotId != -1) {
+        SnapshotRef main = SnapshotRef.branchBuilder(currentSnapshotId).build();
+        setRef(SnapshotRef.MAIN_BRANCH, main);
+      }
+
       if (!hasChanges()) {
         return base;
       }


amogh-jahagirdar:
Tests are failing, definitely related to this change. Looking into it.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -529,6 +531,7 @@ public TableMetadata replaceProperties(Map<String, String> rawProperties) {
 
     return new Builder(this)
         .setProperties(updated)
+        .setRef(SnapshotRef.MAIN_BRANCH, SnapshotRef.branchBuilder(currentSnapshotId).build())
```
rdblue, line 534: I don't think it is a good idea to overwrite the main branch anywhere. If there are no refs, but there is a current snapshot, then the builder should produce a main branch. I think that's the only place where this logic should be handled.

amogh-jahagirdar: Good point, I should be setting this in the builder as that's the whole point of it. I'll update this

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -752,7 +753,19 @@ private static Map<String, SnapshotRef> validateRefs(Long currentSnapshotId,
   }
 
   public static Builder buildFrom(TableMetadata base) {
```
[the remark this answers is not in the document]
amogh-jahagirdar: Actually, I think I can just do this in build()

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -120,13 +120,15 @@ static TableMetadata newTableMetadata(Schema schema,
     // break existing tables.
     MetricsConfig.fromProperties(properties).validateReferencedColumns(schema);
 
+    SnapshotRef main = SnapshotRef.branchBuilder(-1).build();
```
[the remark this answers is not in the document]
amogh-jahagirdar: Good point, I was originally thinking that at least the ref can exist even if there is no table state. But it's easier to just maintain the invariant that so long as there's no table state, main won't exist.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -1187,6 +1192,20 @@ public TableMetadata build() {
           previousFiles, previousFileLocation, base.lastUpdatedMillis(), properties);
       List<HistoryEntry> newSnapshotLog = updateSnapshotLog(snapshotLog, snapshotsById, currentSnapshotId, changes);
 
+      if (refs.isEmpty() && currentSnapshotId != -1) {
```
amogh-jahagirdar, line 1195: Will add tests for this if we deem setting main here is the right way to go

hililiwei: +1

hililiwei:
Is there any update here?

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
```diff
@@ -174,7 +174,7 @@ protected void refreshFromMetadataLocation(String newLocation, int numRetries) {
   protected void refreshFromMetadataLocation(String newLocation, Predicate<Exception> shouldRetry,
                                              int numRetries) {
     refreshFromMetadataLocation(newLocation, shouldRetry, numRetries,
-        metadataLocation -> TableMetadataParser.read(io(), metadataLocation));
+        metadataLocation -> TableMetadata.buildFromLocation(io(), metadataLocation).build());
```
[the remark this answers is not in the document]
rdblue: I don't think we should change the metadata after reading it, so it is probably best to fix this up in the `TableMetadata` constructor rather than in the builder, which is used to produce new `TableMetadata` objects.

rdblue:
Reopening this. I accidentally closed it through a different PR.

amogh-jahagirdar:
This PR is no longer needed due to https://github.com/apache/iceberg/pull/5669 which already includes the change here

PR #4923, opened 2022-05-31
Python: Use taskipy for development task execution

dramaticlly:
Inspired by #4811 , instead of using introduce a new `Makefile` for defining arbitrary development task execution, we can actually leverage the `taskipy` and put task execution logic inside the `pyproject.toml`.

- It adds a new development dependency on `taskipy`, it defines custom logic in single source of truth `pyproject.toml` and it can also integrate with poetry nicely in upcoming #4844 
- It can potentially replace the Makefile (also version pinned in setup.cfg), instead of calling `make lint`, we can use `task lint`. With poetry it will become `poetry run task lint`.
- Task discovery is easier with 
```bash
$ task -l
install pip install -e ".[dev,arrow]"
lint    pre-commit run --all-files
test        coverage run --source=src/ -m pytest tests/ &&     coverage report -m --fail-under=90 &&     coverage html &&     coverage xml
```

---
Example Usage is similar to what we have today, once installed, we can just call
`task install`

`task lint` 

`task test`

Disclaim: 

It might be a bit dumb today, as to do `task install` we need to install the taskipy first. But I think it will get better once we adopt poetry 

CC @Fokko @samredai @rdblue

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -47,3 +47,13 @@ ignore_missing_imports = true
 
 [tool.coverage.run]
 source = ['src/']
+
+[tool.taskipy.tasks]
+install = "pip install -e \".[dev,arrow]\""
+lint = "pre-commit run --all-files"
+test = """
+    coverage run --source=src/ -m pytest tests/ &&
+    coverage report -m --fail-under=90 &&
+    coverage html &&
+    coverage xml
+"""
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -52,8 +52,8 @@ source = ['src/']
 install = "pip install -e \".[dev,arrow]\""
 lint = "pre-commit run --all-files"
 test = """
-    coverage run --source=src/ -m pytest tests/ &&
-    coverage report -m --fail-under=90 &&
-    coverage html &&
-    coverage xml
+coverage run --source=src/ -m pytest tests/ &&
+coverage report -m --fail-under=90 &&
+coverage html &&
+coverage xml
 """
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -55,5 +55,6 @@ dev=
     pytest-checkdocs==2.7.1
     pyenchant==3.2.2
     pre-commit==2.19.0
+    taskipy==1.10.2
 [options.packages.find]
 where = src


python/pyproject.toml
```diff
@@ -47,3 +47,13 @@ ignore_missing_imports = true
 
 [tool.coverage.run]
 source = ['src/']
+
+[tool.taskipy.tasks]
+install = "pip install -e \".[dev,arrow]\""
+lint = "pre-commit run --all-files"
+test = """
+    coverage run --source=src/ -m pytest tests/ &&
```
Fokko, line 55: I would suggest getting rid of the indentation here, probably this will give us a nicer listing:

Before:
```
$ task -l
install pip install -e ".[dev,arrow]"
lint    pre-commit run --all-files
test        coverage run --source=src/ -m pytest tests/ &&     coverage report -m --fail-under=90 &&     coverage html &&     coverage xml
```

after:
```
➜  python git:(taskipy) ✗ task -l             
install pip install -e ".[dev,arrow]"
lint    pre-commit run --all-files
test    coverage run --source=src/ -m pytest tests/ && coverage report -m --fail-under=90 && coverage html && coverage xml
```

Fokko (commented):
I like it! If we want to go for this, we need to add:

- Docs in the python/CONTRIBUTING.md
- Remove the Makefile to avoid confusion

My only concern is that it seems to be quite a young project, with not that many contributors.

Fokko:
Just bumped into a similar project: https://github.com/nat-n/poethepoet

samredai:
I like that this consolidates the tasks into the pyproject.toml, however make is much more common. I think we should consider that devs in other parts of the Iceberg codebase may want to run these tests and it's more likely they'd already have make.

dramaticlly:
> Just bumped into a similar project: https://github.com/nat-n/poethepoet

Thanks @Fokko , yeah I guess there's true. People needs a task execution to be implemented and there's probably no dominant package yet. I took a look at the poethepoet one, looks like it have better auto complete and more stars on github. But the pypi download stats are pretty similar
- https://pypistats.org/packages/poethepoet
- https://pypistats.org/packages/taskipy
Personally, i dont have strong preference, so I am open to everyone's input, thank you for the quick response

rdblue:
Personally, I don't think I'd add this. An extra development dependency just to run tasks doesn't seem worth it. The Makefile is not really required, it's just a convenience. Adding a dependency that makes the configuration less obvious doesn't seem like a win to me.

That said, if people strongly prefer this I'm happy with whatever the community decides!

dramaticlly:
Thanks @rdblue , @samredai  and @Fokko  for the review, I am just proposing some ideas and I agree its benefits is tangential and probably nor worth the effort. I can close this PR

PR #4657, opened 2022-04-28
[WIP] API/Core: View support

jzhuge:
Support view.

The view spec has been added in #3188.

TODO:

- Fix schema handling
- Add unit tests

Follow-ups:

- Add transactions for replace or createOrReplace
- Update field comments
- Support Spark view catalog

Issue #4567, opened 2022-04-15
How to determine iceberg version after table created


sontt710:
Hello,

I used **iceberg-spark-runtime-3.2_2.12-0.13.1.jar** to create Iceberg table, and I want to verify version of this table is 0.13.1. 

How can I get this version from Spark or catalog, etc..? 

Thanks.

RussellSpitzer:
Iceberg tables only have two possible versions (1 or 2) and this is stored in the format-version table property

sontt710:
Thanks @RussellSpitzer , I agree with you.

But in Our environment have multiple iceberg-spark-runtime-* for testing, so I want to know table was created from which runtime version

RussellSpitzer:
This is not recorded in any way as the version shouldn't make a difference in the spec the table is written with. There should be no material difference other than possibly changing defaults.

sontt710:
thanks @RussellSpitzer

PR #3188, opened 2021-09-27
Add common view format spec

anjalinorwood:

PR #4925, opened 2022-06-01
API: Add view interfaces

jzhuge:
Co-authored-by: anjalinorwood@gmail.com

### DIFF
--- a/api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
+++ b/api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.catalog;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.iceberg.exceptions.AlreadyExistsException;
+import org.apache.iceberg.exceptions.NoSuchViewException;
+import org.apache.iceberg.exceptions.NotFoundException;
+import org.apache.iceberg.view.View;
+import org.apache.iceberg.view.ViewBuilder;
+
+/** A Catalog API for view create, drop, and load operations. */
+public interface ViewCatalog {
+
+  /**
+   * Return the name for this catalog.
+   *
+   * @return this catalog's name
+   */
+  String name();
+
+  /**
+   * Return all the identifiers under this namespace.
+   *
+   * @param namespace a namespace
+   * @return a list of identifiers for views
+   * @throws NotFoundException if the namespace is not found
+   */
+  List<TableIdentifier> listViews(Namespace namespace);
+
+  /**
+   * Load a view.
+   *
+   * @param identifier a view identifier
+   * @return instance of {@link View} implementation referred by the identifier
+   * @throws NoSuchViewException if the view does not exist
+   */
+  View loadView(TableIdentifier identifier);
+
+  /**
+   * Check whether view exists.
+   *
+   * @param identifier a view identifier
+   * @return true if the view exists, false otherwise
+   */
+  default boolean viewExists(TableIdentifier identifier) {
+    try {
+      loadView(identifier);
+      return true;
+    } catch (NoSuchViewException e) {
+      return false;
+    }
+  }
+
+  /**
+   * Instantiate a builder to create or replace a SQL view.
+   *
+   * @param identifier a view identifier
+   * @return a view builder
+   */
+  ViewBuilder buildView(TableIdentifier identifier);
+
+  /**
+   * Drop a view.
+   *
+   * @param identifier a view identifier
+   * @return true if the view was dropped, false if the view did not exist
+   */
+  boolean dropView(TableIdentifier identifier);
+
+  /**
+   * Rename a view.
+   *
+   * @param from identifier of the view to rename
+   * @param to new view identifier
+   * @throws NoSuchViewException if the "from" view does not exist
+   * @throws AlreadyExistsException if the "to" view already exists
+   */
+  void renameView(TableIdentifier from, TableIdentifier to);
+
+  /**
+   * Invalidate cached view metadata from current catalog.
+   *
+   * <p>If the view is already loaded or cached, drop cached data. If the view does not exist or is
+   * not cached, do nothing.
+   *
+   * @param identifier a view identifier
+   */
+  default void invalidateView(TableIdentifier identifier) {}
+
+  /**
+   * Initialize a view catalog given a custom name and a map of catalog properties.
+   *
+   * <p>A custom view catalog implementation must have a no-arg constructor. A compute engine like
+   * Spark or Flink will first initialize the catalog without any arguments, and then call this
+   * method to complete catalog initialization with properties passed into the engine.
+   *
+   * @param name a custom name for the catalog
+   * @param properties catalog properties
+   */
+  default void initialize(String name, Map<String, String> properties) {}
+}
--- a/api/src/main/java/org/apache/iceberg/exceptions/NoSuchViewException.java
+++ b/api/src/main/java/org/apache/iceberg/exceptions/NoSuchViewException.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.exceptions;
+
+import com.google.errorprone.annotations.FormatMethod;
+
+/** Exception raised when attempting to load a view that does not exist. */
+public class NoSuchViewException extends RuntimeException {
+  @FormatMethod
+  public NoSuchViewException(String message, Object... args) {
+    super(String.format(message, args));
+  }
+
+  @FormatMethod
+  public NoSuchViewException(Throwable cause, String message, Object... args) {
+    super(String.format(message, args), cause);
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
+++ b/api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.List;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.catalog.Namespace;
+
+public interface SQLViewRepresentation extends ViewRepresentation {
+
+  @Override
+  default Type type() {
+    return Type.SQL;
+  }
+
+  /** The view query SQL text. */
+  String query();
+
+  /** The view query SQL dialect. */
+  String dialect();
+
+  /** The default catalog when the view is created. */
+  String defaultCatalog();
+
+  /** The default namespace when the view is created. */
+  Namespace defaultNamespace();
+
+  /** The query output schema at version create time, without aliases. */
+  Schema schema();
+
+  /** The view field aliases. */
+  List<String> fieldComments();
+
+  /** The view field comments. */
+  List<String> fieldAliases();
+}
--- a/api/src/main/java/org/apache/iceberg/view/View.java
+++ b/api/src/main/java/org/apache/iceberg/view/View.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.iceberg.Schema;
+
+/** Interface for view definition. */
+public interface View {
+
+  String name();
+
+  /**
+   * Return the {@link Schema schema} for this view.
+   *
+   * @return this table's schema
+   */
+  Schema schema();
+
+  /**
+   * Return a map of {@link Schema schema} for this view.
+   *
+   * @return this table's schema map
+   */
+  Map<Integer, Schema> schemas();
+
+  /**
+   * Get the current version for this view, or null if there are no versions.
+   *
+   * @return the current view version.
+   */
+  ViewVersion currentVersion();
+
+  /**
+   * Get the versions of this view.
+   *
+   * @return an Iterable of versions of this view.
+   */
+  Iterable<ViewVersion> versions();
+
+  /**
+   * Get a version in this view by ID.
+   *
+   * @param versionId version ID
+   * @return a version, or null if the ID cannot be found
+   */
+  ViewVersion version(int versionId);
+
+  /**
+   * Get the version history of this table.
+   *
+   * @return a list of {@link ViewHistoryEntry}
+   */
+  List<ViewHistoryEntry> history();
+
+  /**
+   * Return a map of string properties for this view.
+   *
+   * @return this view's properties map
+   */
+  Map<String, String> properties();
+
+  /**
+   * Create a new {@link UpdateViewProperties} to update view properties.
+   *
+   * @return a new {@link UpdateViewProperties}
+   */
+  UpdateViewProperties updateProperties();
+}
--- a/api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
+++ b/api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
@@ -1,0 +1,134 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.ViewCatalog;
+
+/**
+ * A builder used to create or replace a SQL {@link View}.
+ *
+ * <p>Call {@link ViewCatalog#buildView} to create a new builder.
+ */
+public interface ViewBuilder {
+  /**
+   * Set the view schema.
+   *
+   * @param schema view schema
+   * @return this for method chaining
+   */
+  ViewBuilder withSchema(Schema schema);
+
+  /**
+   * Set the view query.
+   *
+   * @param query view query
+   * @return this for method chaining
+   */
+  ViewBuilder withQuery(String query);
+
+  /**
+   * Set the view SQL dialect.
+   *
+   * @param dialect view SQL dialect
+   * @return this for method chaining
+   */
+  ViewBuilder withDialect(String dialect);
+
+  /**
+   * Set the view default catalog.
+   *
+   * @param defaultCatalog view default catalog
+   * @return this for method chaining
+   */
+  ViewBuilder withDefaultCatalog(String defaultCatalog);
+
+  /**
+   * Set the view default namespace.
+   *
+   * @param defaultNamespace view default namespace
+   * @return this for method chaining
+   */
+  ViewBuilder withDefaultNamespace(Namespace defaultNamespace);
+
+  /**
+   * Set the view query column names.
+   *
+   * @param queryColumnNames view query column names
+   * @return this for method chaining
+   */
+  ViewBuilder withQueryColumnNames(List<String> queryColumnNames);
+
+  /**
+   * Set the view field aliases.
+   *
+   * @param fieldAliases view field aliases
+   * @return this for method chaining
+   */
+  ViewBuilder withFieldAliases(List<String> fieldAliases);
+
+  /**
+   * Set the view field comments.
+   *
+   * @param fieldComments view field comments
+   * @return this for method chaining
+   */
+  ViewBuilder withFieldComments(List<String> fieldComments);
+
+  /**
+   * Add key/value properties to the view.
+   *
+   * @param properties key/value properties
+   * @return this for method chaining
+   */
+  ViewBuilder withProperties(Map<String, String> properties);
+
+  /**
+   * Add a key/value property to the view.
+   *
+   * @param key a key
+   * @param value a value
+   * @return this for method chaining
+   */
+  ViewBuilder withProperty(String key, String value);
+
+  /**
+   * Create the view.
+   *
+   * @return the view created
+   */
+  View create();
+
+  /**
+   * Replace the view.
+   *
+   * @return the {@link View} replaced
+   */
+  View replace();
+
+  /**
+   * Create or replace the view.
+   *
+   * @return the {@link View} created or replaced
+   */
+  View createOrReplace();
+}
--- a/api/src/main/java/org/apache/iceberg/view/ViewHistoryEntry.java
+++ b/api/src/main/java/org/apache/iceberg/view/ViewHistoryEntry.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+/**
+ * View history entry.
+ *
+ * <p>An entry contains a change to the view state. At the given timestamp, the current version was
+ * set to the given version ID.
+ */
+public interface ViewHistoryEntry {
+  /** Return the timestamp in milliseconds of the change */
+  long timestampMillis();
+
+  /** Return ID of the new current version */
+  int versionId();
+}
--- a/api/src/main/java/org/apache/iceberg/view/ViewRepresentation.java
+++ b/api/src/main/java/org/apache/iceberg/view/ViewRepresentation.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.Locale;
+
+public interface ViewRepresentation {
+
+  enum Type {
+    SQL;
+
+    public static Type fromString(String typeName) {
+      return valueOf(typeName.toUpperCase(Locale.ENGLISH));
+    }
+
+    public String typeName() {
+      return name().toLowerCase(Locale.ENGLISH);
+    }
+  }
+
+  Type type();
+}
--- a/api/src/main/java/org/apache/iceberg/view/ViewVersion.java
+++ b/api/src/main/java/org/apache/iceberg/view/ViewVersion.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A version of the view at a point in time.
+ *
+ * <p>A version consists of a view metadata file.
+ *
+ * <p>Versions are created by view operations, like Create and Replace.
+ */
+public interface ViewVersion {
+
+  /** Return this version's id. Version ids are monotonically increasing */
+  int versionId();
+
+  /**
+   * Return this version's timestamp.
+   *
+   * <p>This timestamp is the same as those produced by {@link System#currentTimeMillis()}.
+   *
+   * @return a long timestamp in milliseconds
+   */
+  long timestampMillis();
+
+  /**
+   * Return the version summary such as the name of the operation that created that version of the
+   * view
+   *
+   * @return a version summary
+   */
+  Map<String, String> summary();
+
+  /**
+   * Return the list of other view representations.
+   *
+   * <p>May contain SQL view representations for other dialects.
+   *
+   * @return the list of view representations
+   */
+  List<ViewRepresentation> representations();
+}
--- a/api/src/main/java/org/apache/iceberg/view/UpdateViewProperties.java
+++ b/api/src/main/java/org/apache/iceberg/view/UpdateViewProperties.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.Map;
+import org.apache.iceberg.PendingUpdate;
+
+/**
+ * API for updating view properties.
+ *
+ * <p>Apply returns the updated view properties as a map for validation.
+ *
+ * <p>When committing, these changes will be applied to the current view metadata. Commit conflicts
+ * will be resolved by applying the pending changes to the new view metadata.
+ */
+public interface UpdateViewProperties extends PendingUpdate<Map<String, String>> {
+
+  /**
+   * Add a key/value property to the view.
+   *
+   * @param key a String key
+   * @param value a String value
+   * @return this for method chaining
+   * @throws NullPointerException If either the key or value is null
+   */
+  UpdateViewProperties set(String key, String value);
+
+  /**
+   * Remove the given property key from the view.
+   *
+   * @param key a String key
+   * @return this for method chaining
+   * @throws NullPointerException If the key is null
+   */
+  UpdateViewProperties remove(String key);
+}


jzhuge:
Split from #4567

nastra:
@jzhuge I think you wanted to link to #4657 here :)

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,185 @@
...
+  /**
+   * Create a view.
+   *
+   * @param identifier a view identifier
+   * @param representations a list of view representations
+   * @param properties a string map of view properties
+   */
+  View createView(
+      TableIdentifier identifier,
+      List<ViewRepresentation> representations,
```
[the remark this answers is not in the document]
jackye1995: I think it's for cross engine compatibility, in case there is a better way to represent the view that more engines can interpret.

api/src/main/java/org/apache/iceberg/view/ViewDefinition.java
```diff
@@ -0,0 +1,79 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.view;
+
+import java.util.List;
+import org.apache.iceberg.Schema;
+
+public interface ViewDefinition extends ViewRepresentation {
```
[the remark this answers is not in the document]
jackye1995: +1, why is `ViewDefinition` not just an implementation of the presentation?

api/src/main/java/org/apache/iceberg/view/ViewUpdateProperties.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * API for updating view properties.
+ * <p>
+ * Apply returns the updated view properties as a map for validation.
+ * <p>
+ * When committing, these changes will be applied to the current view metadata.
+ * Commit conflicts will be resolved by applying the pending changes to the new view metadata.
+ */
+public interface ViewUpdateProperties extends PendingUpdate<Map<String, String>> {
```
jackye1995, line 33: UpdateViewProperties?

amogh-jahagirdar: +1, this is the naming convention we follow for UpdateProperties for table properties. I was originally thinking we extend UpdateProperties for this case as well, but it may be better to keep them independent otherwise it ties us into an interface contract we may not want.

api/src/main/java/org/apache/iceberg/view/ViewDefinition.java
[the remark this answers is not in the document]
amogh-jahagirdar: Yeah the only difference is in the type, so we can just have 1 be an interface (say ViewRepresentation) and then just have SqlViewRepresentation etc which implement that contract

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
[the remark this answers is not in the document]
jzhuge: https://github.com/apache/iceberg/pull/3188#discussion_r718985188

api/src/main/java/org/apache/iceberg/view/ViewDefinition.java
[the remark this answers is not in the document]
jzhuge: Per discussion, I will turn ViewDefinition into an implementation of ViewRepresentation.
And move it to core.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * Invalidate cached view metadata from current catalog.
+   * <p>
+   * If the view is already loaded or cached, drop cached data. If the view does not exist or is not cached, do
+   * nothing.
+   *
+   * @param identifier a view identifier
+   */
+  default void invalidateView(TableIdentifier identifier) {
```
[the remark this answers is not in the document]
jzhuge: Yeah, I keep this for parity with table catalog.

api/src/main/java/org/apache/iceberg/exceptions/NoSuchViewException.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.apache.iceberg.exceptions;
+
+import com.google.errorprone.annotations.FormatMethod;
+
+/**
+ * Exception raised when attempting to load a view that does not exist.
+ */
+public class NoSuchViewException extends RuntimeException {
+  @FormatMethod
```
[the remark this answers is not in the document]
jzhuge: Neither NoSuchTableException nor NoSuchNamespaceException has a newline.
What do you see this specific style being followed in Iceberg codebase?

api/src/main/java/org/apache/iceberg/view/ViewUpdateProperties.java
jzhuge: `UpdateProperties` has a method `defaultFormat` that does not apply to views.

jzhuge: Name `ViewUpdateProperties` has the same prefix as `ViewHistoryEntry`, `ViewRepresentation`, and `ViewVersion`.

api/src/main/java/org/apache/iceberg/exceptions/NoSuchViewException.java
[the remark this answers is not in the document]
amogh-jahagirdar: Good point, It just looked odd to me but that could just be my style. Looking at the other exceptions this seems to be the standard , so makes sense to leave it as is!

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,153 @@
...
+      Map<String, String> properties);
+
+  /**
+   * Drop a view and delete all data and metadata files.
+   *
+   * @param identifier a view identifier
+   * @return true if the view was dropped, false if the view did not exist
+   */
+  default boolean dropView(TableIdentifier identifier) {
+    return dropView(identifier, true /* drop data and metadata files */);
```
amogh-jahagirdar, line 106: One thing I wanted to clarify, do you have an example of the behavior what it means to drop data files when we purge the view? Sorry it may be a silly question but just wanted to validate my understanding

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Drop a view; optionally delete data and metadata files.
+   * <p>
+   * If purge is set to true the implementation should delete all data and metadata files.
+   *
+   * @param identifier a view identifier
+   * @param purge      if true, delete all data and metadata files in the view
+   * @return true if the view was dropped, false if the view did not exist
+   */
+  boolean dropView(TableIdentifier identifier, boolean purge);
```
amogh-jahagirdar, line 118: One thing I wanted to clarify, do you have an example of the behavior what it means to drop data files (I understand removing metadata files) when we purge the view? Sorry it may be a silly question but just wanted to validate my understanding

amogh-jahagirdar, line 118: One thing I wanted to clarify, do you have an example of the behavior what it means to drop data files (I understand removing metadata files) when we purge the view? Sorry it may be a silly question but just wanted to validate my understanding

api/src/main/java/org/apache/iceberg/view/ViewUpdateProperties.java
jzhuge: Sounds good

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: No data file at this point. I will update the javadoc.

amogh-jahagirdar (approved):
Looks good to me, thanks @jzhuge for the contribution! @jackye1995 @nastra let us know if you have any further comments. 

cc: @rdblue @danielcweeks

jzhuge:
Thanks @amogh-jahagirdar.

BTW, I am also ok if we decide to remove `UpdateViewProperties` and just use the existing `UpdateProperties`.

nastra (approved):

jackye1995 (approved):
Looks good to me!

jackye1995:
@rdblue do you have any additional comment for adding these view APIs?

api/src/main/java/org/apache/iceberg/view/View.java
```diff
@@ -0,0 +1,74 @@
...
+
+package org.apache.iceberg.view;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Interface for view definition.
+ */
+public interface View {
```
amogh-jahagirdar, line 28: I think we would also want to add schema() to this as well. But I think we can tackle that in a follow on PR

jzhuge:
If no more comment, is it possible to get it merged so that I can move on to the Core PR?

jzhuge:
Hi @danielcweeks, have you got a chance to take a look?

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Create a view.
+   *
+   * @param identifier a view identifier
+   * @param representations a list of view representations
+   * @param properties a string map of view properties
+   */
+  View createView(
+      TableIdentifier identifier,
+      List<ViewRepresentation> representations,
```
danielcweeks, line 84: I find it a little odd that we're exposing the list of representations at this level without a wrapper class.  I just feel like there are a number of cases where we would want encapsulating data/information.  I would suggest adding a `ViewMetadata` as a container for this list.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,71 @@
...
+   * Return this version's timestamp.
+   * <p>
+   * This timestamp is the same as those produced by {@link System#currentTimeMillis()}.
+   *
+   * @return a long timestamp in milliseconds
+   */
+  long timestampMillis();
+
+  /**
+   * Returns the version summary such as the name and genie-id of the operation that created that version of the view
```
danielcweeks, line 57: You might want to remove references to genie.

danielcweeks:
@jzhuge or @anjalinorwood I feel like I'm missing something in terms of where we actually access the view content / text?  I would think that would live in `ViewRepresentation` or `ViewVersion`?

amogh-jahagirdar:
@danielcweeks I'll let @jzhuge @anjalinorwood confirm but the model at least in my mind is that there is an implementation of ViewRepresentation; for the SQL view representation, this would be SqlViewRepresentation which would expose a sql() method for surfacing the literal text. Down the line there would be other view representation types such as for substrait which would expose the view representation in their own way (serialized plan node for example).

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jackye1995: Having that `ViewMetadata` encapsulation seems to be against the structure defined in spec overall. Also I think we cannot use the term `ViewMetadata` since that is a term which means the entire view metadata including all versions in the spec: https://github.com/apache/iceberg/blob/master/format/view-spec.md?plain=1#L57

In my opinion, in the long term what we need is a builder here like `Catalog.buildTable(...)`, so something like 

```
ViewCatalog.buildView(TableIdentifier.of("db","table"))
  .addRepresentation(SqlViewRepresentation.of("select * from table"))
  .addProperty("key", "val")
  .build()
```

But for now this `createView` method should be sufficient

amogh-jahagirdar: Yeah ViewMetadata is reserved for the overall metadata for the view. The list of view reprsentations is part of the "versions" in the view.

api/src/main/java/org/apache/iceberg/view/View.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Return a map of string properties for this view.
+   *
+   * @return this view's properties map
+   */
+  Map<String, String> properties();
+
+  /**
+   * Update view properties and commit the changes.
```
stevenzwu, line 69: This Javadoc is not very accurate. maybe sth like this.

```
Create a new {@link UpdateViewProperties} to update view properties and commit the changes.
```

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Map;
+
+/**
+ * A version of the view at a point in time.
+ * <p>
+ * A version consists of a view metadata file.
+ * <p>
+ * Versions are created by view operations, like Create and Replace.
+ */
+public interface ViewVersion {
```
stevenzwu, line 32: To me, `version` only carries the meaning of some version id or number. Just brainstorming some ideas: `ViewPoint` or `PointOfView`?

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Return this version's ID.
+   *
+   * @return a long ID
```
stevenzwu, line 36: type mismatch: int vs long. same for the parentId below.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
amogh-jahagirdar: View Version does contain a version id which identifies the actual version of the view, but I think I get what you mean; the name "version" should be tied to just the numeric id. Some other ideas in my mind is `ViewInstance` or `ViewSnapshot` but maybe snapshot will get confusing, although that is what this really is at least in my mind.

stevenzwu: Yeah, version typically means just the numeric id to me. I also thought about `ViewSnapshot`. I think it can get confused with table snapshot. Or if we use git analogy, maybe `ViewCommit`?

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
stevenzwu: also is version a monotonically increased numbers? if yes, maybe document it in the Javadoc?

jzhuge:
Merged Amogh's PR, rebased, and applied spotless.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
jzhuge: It is not snapshot. We deliberately chose `version` to differentiate from `snapshot`.
As for `ViewCommit`, unfortunately that is another over-used word :(

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
jzhuge: Done

amogh-jahagirdar:
@danielcweeks @stevenzwu @rdblue Any more comments on the API changes? Would be nice so that we could get to the core implementation.

stevenzwu (approved):
I still have a little reservation on `ViewVersion`. However, since I can't come up with a better name, I am gonna to approve this.

api/src/main/java/org/apache/iceberg/view/ViewHistoryEntry.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * <p>An entry contains a change to the view state. At the given timestamp, the current version was
+ * set to the given version ID.
+ */
+public interface ViewHistoryEntry {
+  /** Returns the timestamp in milliseconds of the change */
+  long timestampMillis();
+
+  /** Returns ID of the new current version */
+  int versionId();
```
amogh-jahagirdar, line 32: I just realized this should also be a long.

amogh-jahagirdar:
Follow up from our discussion in yesterday's sync.

## Create View Semantics

I think we'll probably need to better define what createView exactly entails. There are 2 cases I can think of that should be handled in createView. 

CreateView can either:

1.) Create a completely new view with an initial set of representations for the view. i.e. the view is not even defined in the catalog 

OR

2.) Add new representation(s) to an existing view if and only if the new representations or sql dialects don't already exist in the current view (if they exist this means that a user must replace the view). This case should still increment the view version. 

This could be a separate API AddViewRepresentation but this puts more work on a caller to identify if a given view representation or dialect already exists in a view which does not seem like the right behavior. 

## Schema Validation of different representations

There's another issue here for if and how does CreateView validate that the representations all have the expected schema for case 2? Considering a view represents a result of a computation there must be a single schema. Following this, ideally CreateView should validate that the schema is the same for the representations. 

This is challenging considering different representations and SQL dialects. In the long run, I think projects like Coral or Substrait would be used to help perform the schema validation across the different SQL dialects. As an extension, even there could be a Substrait View Representation down the line. 

@wmoustafa @jacques-n Any recommendations on a good way for handling this schema validation within the view format itself? I'm taking a look in Coral and Substrait (apologies if this is a naive question)

For addressing schema validation in the short term, I think in createView API engines should pass in the schema for the dialects. Then createView performs a validation that the passed in schema aligns with the view schema. This could be bypassed by an engine just passing in the view.schema() even though it doesn't actually align with the view definition. So we could 

Happy to get thoughts and suggestions on this! @jzhuge @danielcweek  @jackye1995

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
amogh-jahagirdar: Also I think we'll need to pass in a Schema @jzhuge ?

wmoustafa:
@amogh-jahagirdar very good points. Following this line of thought, we should probably also address the question of whether the SQL text should be validated in Iceberg/the API implementation, or validation is left to the enclosing engine. If it is validated in Iceberg/the API implementation, then Iceberg/implementation needs to understand the dialects as well. If it is left to the enclosing engine, there is the risk of passing an invalid view text. 

Similarly, whether it is responsibility of Iceberg (or the API implementation) to ensure the semantic equivalence of all the representations (and not only the schemas).

Finally, if we assume there is a translation tool, we may consider if it makes sense to store one representation and let the remaining representations be handled by translation, as opposed to letting the caller generate all the representations, which as you indicated, could fall out of sync.

By the way, I believe adding new representations (case #2) should not increment the view version.

To answer your question about Coral schema validation, Coral infers the schema of a view given the view text and base table schemas (see some [test cases](https://github.com/linkedin/coral/blob/ceb8fdfc8c91eb58a0864ca92c22361361f73c3c/coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java#L24)). It uses Avro schemas for representing the schemas (due to its rich semantics), but extension to use Iceberg schemas is straightforward.

wmoustafa:
Should the spec define rules for schema evolution of views across different versions? The implementation should probably enforce them.

jackye1995:
>  Following this line of thought, we should probably also address the question of whether the SQL text should be validated in Iceberg/the API implementation, or validation is left to the enclosing engine. If it is validated in Iceberg/the API implementation, then Iceberg/implementation needs to understand the dialects as well. If it is left to the enclosing engine, there is the risk of passing an invalid view text.

I would vote for leaving to the enclosing engine or translation tool to validate the correctness and semantic equivalence in the initial API implementation.

Imagine in the most manual way, there is a data engineer manually adding 2 representations for Spark and Trino, and the engineer does the manual translation, and the translation might not be perfect but works for their use case. Then the engineer invokes this API to update the Iceberg view spec. We should make sure at least this use case can proceed without issue without Iceberg throwing unnecessary validation exceptions.

We can always introduce features like `CoralValidatedSqlRepresentation` for people who would like to use certain tool, and plugin points in the Iceberg library, such that the integrations could live as plugin packages that don't even need to live in the Iceberg repository.

amogh-jahagirdar:
Thanks @wmoustafa for your thoughts and sorry for the late reply. So here are my thoughts:

> Following this line of thought, we should probably also address the question of whether the SQL text should be validated in Iceberg/the API implementation, or validation is left to the enclosing engine. If it is validated in Iceberg/the API implementation, then Iceberg/implementation needs to understand the dialects as well. If it is left to the enclosing engine, there is the risk of passing an invalid view text.

Yes, after some thought, my thinking is in the long run the Iceberg View format should enforce correctness and semantic equality between the different representations. However, I think for an initial implementation though it suffices to just put it on the engine or translation layer to validate. Later on, validation can be implemented within the View API. 

>By the way, I believe adding new representations (e.g., dialects) of semantically equivalent views (case 2 above) should not increment the view version. Do you agree? It seems this might result in having to distinguish between view representations that only add a dialect and preserve the semantics and view representations that change the semantics.

So view versions in my mind should represent every single change to the view, similar to there's a new table metadata version regardless of if there is a new snapshot or not. Another argument is that view version should only increment for semantic or schema changes in the view (which if i'm not mistaken is what you're suggesting), but I think the mental model would be hard to grasp here when people try and compare with Iceberg table specification.

 Also for an initial implementation if we don't have any validation controls, then I think it becomes necessary to increment on any new version representations which are updated, so that people can rollback the view if it's in a state they do not like.
 
 I'm on board with having a separated updateViewRepresentations which has a "merge" like semantic for updating view representations (either replacing existing ones of the same type and dialect OR if non-existing, create the new representation)
 
 Thoughts @jzhuge @danielcweeks ?

amogh-jahagirdar:
@jzhuge I saw schema was optional in the spec, I see the rationale here https://github.com/apache/iceberg/pull/3188#discussion_r748303084

I think we'll want a separate createView which also accepts schema. 
I've raised a PR to your branch https://github.com/jzhuge/iceberg/pull/2/files for addressing this and a definition for updateViewRepresentations for adding new representations of new types and replacing existing representations of the same type. The PR also has some other fixes for long versionId in the history entry and doc adjustments.

Let me know what you think!

api/src/main/java/org/apache/iceberg/view/View.java
rdblue: I agree. This needs to expose the view schema.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * Drop a view; optionally delete metadata files.
+   *
+   * <p>If purge is set to true the implementation should delete all metadata files.
+   *
+   * @param identifier a view identifier
+   * @param purge if true, delete all metadata files in the view
+   * @return true if the view was dropped, false if the view did not exist
+   */
+  boolean dropView(TableIdentifier identifier, boolean purge);
```
rdblue, line 115: Why is there a `purge` option here? Is there value to requiring this?

The `purge` option for tables is not very valuable and is in a lot of cases misunderstood. The problem is that it confuses two needs: "clean up resources" and "do this immediately". Cleaning up resources (i.e., metadata and data files) should be done on a schedule determined by the catalog and administrators. That way undrop can be supported. Doing something immediately should be used when it is urgent that the underlying resources are removed for some other reason, like needing to completely remove PII that has leaked.

In the view case, I see no need for `purge` because I doubt there is a case where deleting a view's description or representation must be done immediately.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,148 @@
...
+      Map<String, String> properties);
+
+  /**
+   * Drop a view and delete all metadata files.
+   *
+   * @param identifier a view identifier
+   * @return true if the view was dropped, false if the view did not exist
+   */
+  default boolean dropView(TableIdentifier identifier) {
+    return dropView(identifier, true /* drop metadata files */);
```
rdblue, line 103: I don't think there should be a purge option, and I definitely don't think that it should be set to `true` by default. Otherwise, you're instructing the catalog to discard the information that may be needed to un-drop the view later.

api/src/main/java/org/apache/iceberg/view/ViewHistoryEntry.java
rdblue: The spec is ambiguous here. I think it's unlikely that a view will have 4 billion versions, so I'd leave it as an int.

api/src/main/java/org/apache/iceberg/view/View.java
```diff
@@ -0,0 +1,71 @@
...
+   */
+  Iterable<ViewVersion> versions();
+
+  /**
+   * Get a version in this view by ID.
+   *
+   * @param versionId version ID
+   * @return a version, or null if the ID cannot be found
+   */
+  ViewVersion version(long versionId);
```
rdblue, line 49: This should be an `int` to correspond to `ViewHistoryEntry`.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>Versions are created by view operations, like Create and Replace.
+ */
+public interface ViewVersion {
+
+  /** Returns this version's id. Version ids are monotonically increasing */
+  long versionId();
+
+  /** Returns this version's parent id or null if there is no parent */
+  Long parentId();
```
rdblue, line 37: What does it mean to have a parent version? If it is just the order in which view versions are created, can't we get that by looking at view with `versionId - 1`?

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
rdblue: I think `ViewVersion` is fine. We could also go with just `Version`, since this is in the `view` package.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,64 @@
...
+ * A version of the view at a point in time.
+ *
+ * <p>A version consists of a view metadata file.
+ *
+ * <p>Versions are created by view operations, like Create and Replace.
+ */
+public interface ViewVersion {
+
+  /** Returns this version's id. Version ids are monotonically increasing */
+  long versionId();
```
rdblue, line 34: The type of `version` should be consistent. Probably int?

api/src/main/java/org/apache/iceberg/view/ViewRepresentation.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.iceberg.view;
+
+import java.util.Locale;
+
+public interface ViewRepresentation {
+
+  enum Type {
+    SQL;
+
+    public static Type parse(String typeName) {
```
rdblue, line 28: I think we typically call this `fromString`.

api/src/main/java/org/apache/iceberg/view/ViewRepresentation.java
```diff
@@ -0,0 +1,38 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.view;
+
+import java.util.Locale;
+
+public interface ViewRepresentation {
```
rdblue, line 23: I'd like to see the SQL view representation in this PR. Otherwise it's hard to understand where different pieces of information are coming from.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
rdblue: > But for now this createView method should be sufficient

I think we should create a builder now so that we can see the whole proposed API, not just part of it. We took the approach of updating `createTable` for different cases for a long time and needed to introduce the builder API because it was getting so annoying to maintain all the methods.

This is missing critical parts of the view spec right now, like schema, and I don't think we can move forward until those are represented.

```java
catalog.buildView(identifier, schema)
    .withQuery(selectStatement, dialect)
    .withProperty("key", "value")
    .withProperties(props)
    .withLocation(s3Location)
    .create();
```

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
rdblue: What about `Definition`? We can use a version ID for the definition, but call it a "definition".

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,148 @@
...
+      Map<String, String> properties);
+
+  /**
+   * Replace a view.
+   *
+   * @param identifier a view identifier
+   * @param representations a list of view representations
+   * @param properties a string map of view properties
+   */
+  View replaceView(
```
rdblue, line 91: This method assumes that a view will be replaced using a catalog method, but I'm not sure that's a good idea.

The table builder API exposes `create`, `createTransaction`, `replaceTransaction`, and `createOrReplaceTransaction`. The `Transaction` methods return a Transaction that will create or replace the table when the transaction is committed. I don't think that there are transactions for views, though. It would make little sense to create a view, then update its properties in a transaction rather than create a view with the properties already set.

I could see adding a way to make it easy to create or replace a view, but that's about it.

For me, this question comes down to how we are going to evolve views in the API. `View` exposes `updateProperties` and it makes a lot of sense to me to use the same pattern that we use for tables:

```java
view.updateProperties().set("prop", "val").commit();

view.updateDefinition()
    .withSchema(schemaFromSelect)
    .withQuery(selectStatement)
    .withAliases(aliases, docs)
    .withCatalog("prod")
    .withNamespace("db")
    .withDialect("spark-3.3")
    .commit();
```

With that, I'd probably just have the view create builder expose a `createOrReplace` method to finish the building.

rdblue:
@jzhuge, I just had a look at this and it's mostly good. I think a few places should be updated though. Can you make the changes? Thanks!

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
stevenzwu: +1 Ryan's latest suggestion of `ViewDefinition`. It is the most intuitive name so far.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Will remove

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Same action as above.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Adding the buildView API.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
jzhuge: TODO: sounds good, checking the code to be sure.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
jzhuge: You are right. Will we support rollback? If yes, the parent-child relation makes a little bit sense, for maintaining a lineage tree.

However, since nothing is shared between 2 views, "rollback" can be treated as a "replace" with all representations re-computed.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Will do

jzhuge:
3 ways to update view metadata:

1. `buildView() + create/replace`: create or replace the entire view
2. `loadView().updateRepresentations + commit`: update representations
3. `loadView().updateProperties + commit`: update properties

jzhuge:
To create a view with SQL representation:

```java
ViewRepresentation sqlViewRepresentation = SQLViewRepresentation.builder()
    .sql("SELECT foo FROM base")
    .dialect("spark")
    .defaultCatalog("prodhive")
    .defaultNamespace(ICollections.singleList("default"))
    .build()

viewCatalog.buildView(ident)
    .withSchema(schema)
    .withRepresentation(sqlViewRepresentation)
    .createOrReplace()
```

jzhuge:
Added `ViewBuilder.withLocation` and rebased to master branch.

jzhuge:
@rdblue Added `buildSqlView` for convenience, so the above example can be shortened as:
```java
viewCatalog.buildSqlView(ident)
    .withSchema(schema)
    .withSql("SELECT foo FROM base")
    .withDialect("spark")
    .withDefaultCatalog("prodhive")
    .withDefaultNamespace(Collections.singleList("default"))
    .withProperty("k1", "v1")
    .withLocation(s3Loc)
    .createOrReplace()
```

Should we add this convenience api for common use cases?
For builder method names, should we use `with` prefix or not? I have examples of both in the existing code.

api/src/main/java/org/apache/iceberg/view/SqlViewBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Map;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.catalog.ViewCatalog;
+
+/**
+ * A builder used to create or replace a {@link View} with only SQL view representation.
+ *
+ * <p>Call {@link ViewCatalog#buildSqlView} to create a new builder.
+ */
+public interface SqlViewBuilder {
```
stevenzwu, line 31: there seems some overlap btw `ViewBuilder` and `SqlViewBuilder`. maybe we can extract a base interface.

You can see `Scan`, `IncrementalScan`, `IncrementalAppendScan`, `IncrementalChangelogScan` as an example.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,180 @@
...
+package org.apache.iceberg.view;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/** SQL definition for a view */
+public class SQLViewRepresentation implements ViewRepresentation {
```
stevenzwu, line 28: should this live in the iceberg-core module? it doesn't seem to be referenced anywhere in the api module.

rdblue: Yes, I think that implementation classes should live in core. We only want to put the interfaces in API.

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Create a view.
+   *
+   * @param identifier a view identifier
+   * @param schema view schema
+   * @param representation a view representation
+   * @param properties a string map of view properties
+   */
+  default View createView(
```
rdblue, line 82: Do we need this method? What about using just `buildView` instead?

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,149 @@
...
+   */
+  ViewBuilder buildView(TableIdentifier identifier);
+
+  /**
+   * Instantiate a builder to create or replace a view with only SQL view representation.
+   *
+   * @param identifier a view identifier
+   * @return a view builder
+   */
+  ViewBuilder buildSqlView(TableIdentifier identifier);
```
rdblue, line 108: I would rather have a single `buildView` method that allows you to create a SQL view. There are no other representations right now so we should focus on SQL. Rather than exposing `ViewRepresentation` right now, we can use methods that update the SQL representation, like `withQuery(String)`.

api/src/main/java/org/apache/iceberg/view/SqlViewBuilder.java
rdblue, line 31: In Iceberg, we typically keep commonly upper case words as upper case, like `RESTCatalog`. In this case, I think `SQLViewBuilder` would be a better name, although I think we will probably want a single view builder, not one specific to SQL.

api/src/main/java/org/apache/iceberg/view/SqlViewBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+   */
+  SqlViewBuilder withSchema(Schema schema);
+
+  /**
+   * Set the view sql.
+   *
+   * @param sql view sql
+   * @return this for method chaining
+   */
+  SqlViewBuilder withSql(String sql);
```
rdblue, line 46: I'd prefer `withQuery` because I think that is more clear that the SQL should be a `SELECT` statement.

api/src/main/java/org/apache/iceberg/view/SqlViewBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Set the view sql dialect.
+   *
+   * @param dialect view sql dialect
+   * @return this for method chaining
+   */
+  SqlViewBuilder withDialect(String dialect);
```
rdblue, line 54: I think `withSql` and `withDialect` are the only options that are specific to a SQL-defined view. I think that means we should just merge this with `ViewBuilder` and remove the places that pass `ViewRepresentation`. I generally don't think it is very friendly to expose representation like that and require users to figure out how to create one, when we could just build one internally.

We shouldn't expose internal complexity through the API unless it is really needed.

api/src/main/java/org/apache/iceberg/view/UpdateViewRepresentations.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * API for updating view representations.
+ *
+ * <p>Apply returns the updated view representations as a list for validation.
+ *
+ * <p>When committing, these changes will be applied to the current view metadata. Commit conflicts
+ * will be resolved by applying the pending changes to the new view metadata.
+ */
+public interface UpdateViewRepresentations extends PendingUpdate<List<ViewRepresentation>> {
```
rdblue, line 33: I'm not confident that this is how we want to evolve a view. Like I said above, I don't like exposing a big complex object (`ViewRepresentation`) to users.

I'd prefer to start off with just `updateDefinition` or similar that allows replacing the current view version. We can adapt that to add representations implicitly. That update can contain a call to both `withQuery` or a future `withSubstrait` or something that creates parallel representations.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
jzhuge: Sounds good

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Fine with me

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Totally fine with this approach.

- buildView() + create/replace(): create or replace sql view representation
- loadView().updateRepresentations() + commit(): update other representations

api/src/main/java/org/apache/iceberg/view/UpdateViewRepresentations.java
jzhuge: Sounds good. We can add this or something similar when the use cases come up.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
jzhuge: Unfortunately, "view definition" has different meaning in the spec:

```
##### Original View Definition in SQL

This type of representation stores the original view definition in SQL and its SQL dialect.
```

api/src/main/java/org/apache/iceberg/view/View.java
jzhuge: Added schema() and schemas(), similar to `Table` interface.

api/src/main/java/org/apache/iceberg/view/ViewRepresentation.java
jzhuge: Added interface `SQLViewRepresentation` that extends this interface.

api/src/main/java/org/apache/iceberg/view/SqlViewBuilder.java
jzhuge: Done

api/src/main/java/org/apache/iceberg/view/SqlViewBuilder.java
jzhuge: Done

jzhuge:
@rdblue @danielcweeks @stevenzwu @amogh-jahagirdar @jackye1995 @nastra Please take another look.

Summary of discussions:

- ViewBuilder API focuses on SQL view first. Do not expose vague APIs to manage a list of representations.
- Interface name `ViewVersion` not ideal, but no good alternative.
- Version ID type is integer, not long.

Usage example:

```
viewCatalog.buildView(ident)
    .withSchema(schema)
    .withQuery("SELECT foo FROM base")
    .withDialect("spark")
    .withDefaultCatalog("prodhive")
    .withDefaultNamespace(Collections.singleList("default"))
    .withProperty("k1", "v1")
    .withLocation(s3Loc)
    .createOrReplace();
```

jzhuge:
> How about `ViewMetadataVersion`, `ViewMetadataSnapshot`

Like `ViewMetadataVersion`.

> I am still unclear on whether we could store different dialects for the same definition as one version or multiple dialects (even for the same definition) imply different versions.

Multiple dialects are available at the same time, so in one version.

> 
> Related, how can a `ViewVersion` return a list of representations?

Each `Version` contains a list of representations. See view spec, section `Versions`.

> 
> Finally, what in the example you gave instructs building a `SQLViewRepresentation`?

In the created `View`, current version contains a single SQL view representation.

wmoustafa:
> 
> Multiple dialects are available at the same time, so in one version.
>
What is an API example to achieve that?
 
> Each `Version` contains a list of representations. See view spec, section `Versions`. 
> In the created `View`, current version contains a single SQL view representation.

Could you clarify if the above two statements are consistent? It seems the former implies a version contains a list of representations and the latter indicates a version contains a single representation.

>> Finally, what in the example you gave instructs building a SQLViewRepresentation?

This question was about how to implement the builder. Will this builder result in creating a `SQLViewRepresentation`? If yes, how would it know it is a `SQLViewRepresentation` vs something else (e.g., not SQL representation)?

jzhuge:
> > Multiple dialects are available at the same time, so in one version.
> 
> What is an API example to achieve that?

Not exposed currently. Do you have a use case? What do u need in the API to achieve it?

> 
> > Each `Version` contains a list of representations. See view spec, section `Versions`.
> > In the created `View`, current version contains a single SQL view representation.
> 
> Could you clarify if the above two statements are consistent? It seems the former implies a version contains a list of representations and the latter indicates a version contains a single representation.

What I meant was the current builder API only supports creating a view view with a single representation, but the view spec and on-disk format does support multiple representations.

> 
> > > Finally, what in the example you gave instructs building a SQLViewRepresentation?
> 
> This question was about how to implement the builder. Will this builder result in creating a `SQLViewRepresentation`? If yes, how would it know it should create a `SQLViewRepresentation` vs something else (e.g., not SQL representation)?

Yes. The current view builder only supports SQLViewRepresentation. For other representations, one idea is to add more `withXXX` methods. Any suggestion?

wmoustafa:
> Not exposed currently. Do you have a use case? What do u need in the API to achieve it?

I do not necessarily have a use case, but was trying to understand the API contracts. It sounds like something that should be either explicitly supported or not supported by the API (i.e., without having to introduce a backward incompatibility in the future to support it), and probably we should show a pathway for this API to be compatible with the spec.

As a follow up, in case both multiple representations could share the same version, and additional representations could end up with a new version, semantically what does a new version mean? Is it an addition to the previous version, or a replacement? For example, if I add the Spark representation in one version and the equivalent Trino representation in another (with different syntax), is it allowed? Should Spark queries start to fail?

> Yes. The current view builder only supports SQLViewRepresentation. For other representations, one idea is to add more withXXX methods. Any suggestion?

Just making sure that `withQuery` is the one that signifies the `SQLViewRepresentation`. It sounded there was another source that it is a SQL view (e.g., the fact that the dialect is Spark). Would be good to find out how to normalize those things (so there are not two independent hints about the same representation).

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+   */
+  ViewBuilder withSchema(Schema schema);
+
+  /**
+   * Set the view query.
+   *
+   * @param query view query
+   * @return this for method chaining
+   */
+  ViewBuilder withQuery(String query);
```
wmoustafa, line 46: If this is supposed to return `SQLViewReperentation`, then it should enforce it somehow?

jzhuge: > If this is supposed to return `SQLViewReperentation`, then it should enforce it somehow?

`ViewVersion` returns a list of `ViewRepresentation`. How about adding a field explicitly for `SQLViewReperentation`? In this way, we enforce views should have at least SQL view representation. On disk, we can still lump together and store as a JSON array.

jzhuge: Good idea since we always have  at least 1 SQL view representation.

jzhuge: If there are multiple SQL view representations for different dialects, I assume the first/default one (specified in buildView) should be set in this new field.

jzhuge: Added `ViewVersion.sqlViewRepresentation()`.

jzhuge:
@wmoustafa Thanks for the valuable questions and feedbacks. Echo your point that it is confusing to store representations in different view versions.

I'd suggest this API contract:

- Add the main SQL view representation and other dialects in `buildView` API so all are stored in one view version
- For a new representation down the road, add a new interface and a new `withXXX` in `buildView`
- No need for something like `View.updateRepresentions()`, as `buildView() + replace()` should be enough

jzhuge:
@rdblue @wmoustafa Could you take look at the PR again? Especially `ViewBuilder`.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Add a SQL {@link View} for a different dialect.
+   *
+   * <p>If its dialect is the same as the SQL view being built, {@link IllegalArgumentException}
+   * will be thrown when creating or replacing the view.
+   *
+   * @param sqlViewRepresentation a SQL view representation
+   * @return this for method chaining
+   */
+  ViewBuilder withOtherSQLView(SQLViewRepresentation sqlViewRepresentation);
```
amogh-jahagirdar, line 122: Nit: I think this should be called  `withOtherSQLRepresentation` to differentiate that it's another representation and not a "View" in our terms.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
amogh-jahagirdar: I think this should be fine for now because it for now we will practically always have sql representation. Down the line, if engines use something use a library like Substrait to serialize common plans then this assumption may not hold.

amogh-jahagirdar (approved):
Overall looks great to me, just a comment on naming on a method in the builder.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
jzhuge: Thanks @amogh-jahagirdar !

One of these 3 choices :)

- withOtherSQLView
- withOtherSQLViewRepresentation
- withOtherSQLRepresentation

jzhuge: Applied the change

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,119 @@
...
+   * @return instance of {@link View} implementation referred by the identifier
+   * @throws NoSuchViewException if the view does not exist
+   */
+  View loadView(TableIdentifier identifier);
+
+  /**
+   * Check whether view exists.
+   *
+   * @param identifier a view identifier
+   * @return true if the table exists, false otherwise
```
stevenzwu, line 61: nit: table -> view?

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
```diff
@@ -0,0 +1,119 @@
...
+    try {
+      loadView(identifier);
+      return true;
+    } catch (NoSuchViewException e) {
+      return false;
+    }
+  }
+
+  /**
+   * Instantiate a builder to create or replace a SQL view.
```
stevenzwu, line 73: should `SQL view` be just `view`?

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,41 @@
...
+    return Type.SQL;
+  }
+
+  String query();
+
+  String dialect();
+
+  String defaultCatalog();
+
+  List<String> defaultNamespace();
```
stevenzwu, line 36: if it is singular, return type should be `String`. if return is a `List`, the method name probably should be plural.

stevenzwu: oh. I guess the list is for the namespace hierachy.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+   */
+  ViewBuilder withDefaultCatalog(String defaultCatalog);
+
+  /**
+   * Set the view default namespaces.
+   *
+   * @param defaultNamespaces view default namespaces
+   * @return this for method chaining
+   */
+  ViewBuilder withDefaultNamespace(List<String> defaultNamespaces);
```
stevenzwu, line 70: similar, should we use `Namespace` class here?

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
stevenzwu: if this is intended for SQL view/query, should this method be called `withSQLQuery`?

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+   *
+   * @param location a location
+   * @return this for method chaining
+   */
+  ViewBuilder withLocation(String location);
+
+  /**
+   * Add a SQL {@link View} for a different dialect.
+   *
+   * <p>If its dialect is the same as the SQL view being built, {@link IllegalArgumentException}
```
stevenzwu, line 116: should this be written as `@throws`?

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: Fixed

api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java
jzhuge: It is SQL View. `ViewBuilder` create a SQL view primarily.

It can add an additional SQL view for a different dialect using `withOtherSQLRepresentation`.

In the future, we will extend it with another `withNNNN` for a new representation.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
jzhuge: Changed to use Namespace class and use singular

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
jzhuge: Changed to use Namespace class and use singular

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
jzhuge: Fixed

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /** The view query SQL text. */
+  String query();
+
+  /** The view query SQL dialect. */
+  String dialect();
+
+  /** The default catalog when the view is created. */
+  String defaultCatalog();
```
stevenzwu, line 38: this is a getter. wondering why `default...`? should it just be `catalog()`?

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /** The default namespace when the view is created. */
+  Namespace defaultNamespace();
```
stevenzwu, line 41: similar question, should this just be `namespace()`?

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
jzhuge: In a Spark/Presto SQL session, you can run `USE` to set default catalog and namespace, thus the "default".

jzhuge:
Created #6134 to add the missing field `query-column-names` to SQL view representation in the view spec.

wmoustafa:
> * No need for something like `View.updateRepresentions()`, as `buildView() + replace()` should be enough

Can one replace the current version or an old version by adding a new dialect? Does it result in a new version?

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
wmoustafa: My original question was that this builder returns a `View`, and in this discussion, you mentioned that `withQuery()` is the method responsible for the fact that it should return `SQLViewRepresentation`. However, I think this method still does not enforce it in this iteration.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,53 @@
...
+
+public interface SQLViewRepresentation extends ViewRepresentation {
+
+  @Override
+  default Type type() {
+    return Type.SQL;
+  }
+
+  /** The view query SQL text. */
+  String query();
```
wmoustafa, line 32: Why would not this method be also common with other representations?

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /** The view query SQL dialect. */
+  String dialect();
```
wmoustafa, line 35: I feel this needs to be an `Enum` and somehow be merged with `Type`. @rdblue, @danielcweeks, @amogh-jahagirdar, thoughts? Further, it is better for the spec to take a position on supported dialects, or not take a position at all, but in this case SQL should not be a special type. Right now the position sounds a bit fuzzy.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
rdblue: This is the SQL text of the view. Other view representations don't necessarily have a SQL query. I think this is reasonable.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
rdblue: I disagree that we want to define supported dialects. We don't want engine writers to need to come to the Iceberg community and get a new symbol in an enum before storing a view. I don't see much benefit to doing that.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
```diff
@@ -0,0 +1,53 @@
...
+  /** The default catalog when the view is created. */
+  String defaultCatalog();
+
+  /** The default namespace when the view is created. */
+  Namespace defaultNamespace();
+
+  /**
+   * The output column names of the query that creates this view. The field aliases are not applied.
+   */
+  List<String> queryColumnNames();
```
rdblue, line 46: Why is this not a schema?

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+  ViewBuilder withLocation(String location);
+
+  /**
+   * Add a SQL {@link View} for a different dialect.
+   *
+   * @param sqlViewRepresentation a SQL view representation
+   * @return this for method chaining
+   * @throws IllegalArgumentException if the dialect is the same as the SQL view being built
+   */
+  ViewBuilder withOtherSQLRepresentation(SQLViewRepresentation sqlViewRepresentation);
```
rdblue, line 129: Let's remove this. We should not be exposing view representation to callers using this builder yet.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+   */
+  ViewBuilder withProperty(String key, String value);
+
+  /**
+   * Set a location for the view.
+   *
+   * @param location a location
+   * @return this for method chaining
+   */
+  ViewBuilder withLocation(String location);
```
rdblue, line 120: What does the location control? Do we need to expose this right now?

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
rdblue: I don't think parent makes sense because each view definition is independent. Maybe history matters, but we have history for that. Let's remove parent ID.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * <p>Versions are created by view operations, like Create and Replace.
+ */
+public interface ViewVersion {
+
+  /** Return this version's id. Version ids are monotonically increasing */
+  int versionId();
+
+  /** Returns this version's parent id or null if there is no parent */
+  int parentId();
```
rdblue, line 37: Let's remove parent ID. I don't think it makes sense for views.

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
```diff
@@ -0,0 +1,71 @@
...
+   * @return a version summary
+   */
+  Map<String, String> summary();
+
+  /**
+   * Return the main SQL view representation.
+   *
+   * @return a SQL view representation
+   */
+  SQLViewRepresentation sqlViewRepresentation();
```
rdblue, line 61: Let's remove this. There isn't necessarily a SQL representation.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
rdblue: Thinking about this more, I think it would make sense if this were a schema ID rather than query column names. Since this is derived from the query by the engine that creates the view, we should keep the full version.

We may also want to simply remove this and add a schema ID to `ViewVersion` since the schema should be consistent across representations.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
wmoustafa: I think it is fine to do either: not take a position at all or take a well defined position. There could be a few scenarios where the intended meaning is not achieved. For example:
* One defines the view SQL using a non-SQL string (e.g., some other DSL), and still leverages `SQLViewReperesentation` for that. Even if this is permitted, it is not clear from the spec if it is okay.
* The dialect could be the same but not standardized, e.g., dialects such as `spark`, `Spark`, `SparkSQL`, `Spark SQL`.

wmoustafa: Maybe what we need is an API to add supported dialects to the metadata? It is up to each catalog implementation to add their dialects, and view definitions can be associated with a dialect ID.

jzhuge: > Maybe what we need is an API to add supported dialects to the metadata? It is up to each catalog implementation to add their dialects, and view definitions can be associated with a dialect ID.

Already added
```java
  /**
   * Add a SQL {@link View} for a different dialect.
   *
   * @param sqlViewRepresentation a SQL view representation
   * @throws IllegalArgumentException if the dialect is the same as the SQL view being built
   */
  ViewBuilder withOtherSQLRepresentation(SQLViewRepresentation sqlViewRepresentation);
```

wmoustafa: I was referring to making dialects (not just view representations) a construct in the metadata where people can define and add them, e.g., some catalogs define `SparkSQL, id = 1` and `TrinoSQL, id = 2`, and each view representation references a standardized dialect ID. They are not part of the Iceberg spec, but defining and adding them dynamically is.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
jzhuge: Not needed. Can add later if necessary.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
jzhuge: Needed to add a different dialet. See https://github.com/apache/iceberg/pull/4925#discussion_r1015685367.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
rdblue: > One defines the view SQL using a non-SQL string

I don't think this is allowed. If you create a SQL view then it should contain SQL. In any case, being strict about dialects doesn't help prevent this.

> The dialect could be the same but not standardized

This is a risk, but I think it isn't very likely. We should document the known dialect strings, but we shouldn't switch it to be an enum.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
rdblue: @jzhuge, I still think we should remove this. We don't need to expose multiple representations in the builder yet. And when we do, I'm very, very skeptical that we will do so by passing a `SQLViewRepresentation` object.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
wmoustafa: Also relevant discussion here: https://github.com/apache/iceberg/pull/6134#discussion_r1015129057

I think we should try to avoid thinking in terms what Spark stores in its metadata and try to understand the higher level objective.

wmoustafa: I think "original" query column names are specifically useful for `SELECT *` queries only, and for full utility they should be accompanied by table names as well to avoid ambiguity. I think we should also decouple the data communicated from an engine to the catalog as part of a `CREATE VIEW` statement, and the what we capture in the View Representation. Having `ViewRepresentation` heavily influenced by `CREATE VIEW` syntax could be the reason for this complexity?

Adding the right tag for Erik @xkrogen.

jzhuge: Yeah, lots of complexity comes from

- column alias feature of CREATE VIEW in Spark
- store column names for SELECT star query

jzhuge:
> > * No need for something like `View.updateRepresentions()`, as `buildView() + replace()` should be enough
> 
> Can one replace the current version or an old version by adding a new dialect? Does it result in a new version?

Not supported.

To add a new dialect, `buildView() + withOtherSQLRepresentation() + replace()`

wmoustafa:
`buildView() + withOtherSQLRepresentation() + replace()` automatically copies existing dialects from previous versions or if we want to preserve them, we need to add them again?

api/src/main/java/org/apache/iceberg/view/ViewVersion.java
jzhuge: I will create a PR to update the view spec.

api/src/main/java/org/apache/iceberg/view/ViewBuilder.java
wmoustafa: Relevant discussion here: https://github.com/apache/iceberg/pull/4925#issuecomment-1253381372.

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
rdblue: @jzhuge, I think that we need the schema stored and accessible, and I think that should be the pre-alias schema that is derived from the SQL query.

Storing the pre-alias column names isn't sufficient because engines may choose to validate more than just the top-level columns that are selected. For example, an engine may want to validate that the type of an `id` column is still a bigint and not, for example, a string. The same logic also applies to nested types for evolution.

The most strict validation of a view is to ensure that the SQL currently produces the same schema, or one that can be cast to the same schema.

rdblue (approved):

api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java
rdblue: I updated this. Will merge when tests are passing.

rdblue:
@jzhuge, this looks good to go. Can you rebase and fix tests?

jzhuge:
@rdblue Please merge.

Issue #3896, opened 2022-01-13
Support Iceberg branching for all the snapshot producer operations


jackye1995:
With the introduction of branching, we can produce snapshots to a specific branch like the following:

```java
table.newAppend()
    .appendFile(someFile)
    .branch("some_branch")
    .commit();
```

flyrain:
I guess people can use branches for testing some things without impact the trunk. The trunk should still be there for any read in that case. Would you mind share any solid use cases? Like in which scenario this would be pretty useful?

jackye1995:
> I guess people can use branches for testing new changes without impact the trunk
yes that is one primary use case. Another one is for a separated branch that maintains a historical snapshot. A compacted version could be updated to the branch for optimized read performance.

flyrain:
Do you mean "a compact version could be updated to the **trunk** for optimized read"? Make sense to me.

RussellSpitzer:
Are we going to allow merging a branch back into the main line? I was thinking about this for transaction support.

flyrain:
IIRC, we got a discussion on this, the verdict is not to support merging a branch into the trunk.

amogh-jahagirdar:
There is also some more clarity with what it could look like based on the discussion in https://github.com/apache/iceberg/pull/3883 (in particular https://github.com/apache/iceberg/pull/3883#discussion_r795274832). @jackye1995 I am happy to take this issue if possible!

namratha2403:
@jackye1995 @amogh-jahagirdar I have been working working on the above issue and have a raised a PR for same. Any comments or suggestions appreciated. PR is:  https://github.com/apache/iceberg/pull/4926

syun64:
Hi folks, I see this PR was merged to enable support for writing to branches with table identifiers in Spark 3.3. https://github.com/apache/iceberg/pull/6965

Do we need to go through a separate effort to backport it to Spark 3.2, and is it possible to backport to 3.2? (Are there any Spark features that are only available in 3.3 that's required for introducing custom table identifiers?)

PR #4926, opened 2022-06-01
Core, API: Performing operations on a snapshot branch ref

namrathamyske:
This PR is related to the open issue https://github.com/apache/iceberg/issues/3896 to perform table operations on a branch ref. Right now throws UnsupportedOperationException. In laters PR's will include implementation for each operation.


amogh-jahagirdar (changes requested):
Thanks so much for contributing this!

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,27 @@ public void testIncludedPartitionSummaryLimit() {
...
+            .commit();
+
+    table.newFastAppend()
+            .appendFile(FILE_B)
+            .commit();
+
+    table.manageSnapshots().createBranch("branch",table.currentSnapshot().snapshotId());
+
+    table.newFastAppend()
+            .appendFile(FILE_A)
```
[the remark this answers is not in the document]
namratha2403: yes , thats right.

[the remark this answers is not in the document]
namratha2403: Should i remove this test for now? I would need a way to test the refs from table.

namratha2403:
@jackye1995 @rdblue any review comments to this ?

core/src/test/java/org/apache/iceberg/TestFastAppend.java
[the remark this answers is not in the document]
amogh-jahagirdar: For getting the refs one way is to check the underlying metadata. table.ops().current().ref(ref) and then check if the snapshot id is expected. But imo I think it's worth waiting for the proper table APIs.

amogh-jahagirdar:
Also a more fundamental thing:

In apply https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L170

the current snapshot should be the tip of the branch if a branch is set, or the latest snapshot if it's main. We can do that by getting the ref and then the snapshot.

amogh-jahagirdar (changes requested):
There is a more fundamental issue here for keeping the snapshot lineage correct. The parent of the new snapshot being produced needs to be set to the tip of the branch.

In apply https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L170

the current snapshot should be the tip of the branch if a branch is set, or the latest snapshot if it's main. We can do that by getting the branch and then the snapshot id for that. Then the new base snapshot produced will have the parent id we expect (the parent id of the tip of the branch). Then the commit logic can be kept as is.

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
```diff
@@ -60,4 +60,9 @@
    * @return this for method chaining
    */
   ThisT scanManifestsWith(ExecutorService executorService);
+
+  /**
+   * Perform operations a particular branch
```
rdblue, line 65: This needs a `@param` to document branch. I think we also need to specify behavior:
* What happens if the branch does not exist? I think we would create one from the `main` branch.
* How does this change validation? What if the validation's from snapshot is not an ancestor of the branch?

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
```diff
@@ -60,4 +60,9 @@
...
+   */
+  ThisT branch(String branch);
```
rdblue, line 67: I think we need to add another word to make it clear what is happening here, like `toBranch` or `forBranch`. Just branch isn't obvious when you're reading it:

```java
table.newOverwrite()
    .overwriteByRowFilter(filter)
    .addFile(one)
    .addFile(two)
    .toBranch(b)
    .commit()
```

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -298,11 +306,11 @@ public void commit() {
             TableMetadata.Builder update = TableMetadata.buildFrom(base);
             if (base.snapshot(newSnapshot.snapshotId()) != null) {
               // this is a rollback operation
-              update.setBranchSnapshot(newSnapshot.snapshotId(), SnapshotRef.MAIN_BRANCH);
+              update.setBranchSnapshot(newSnapshot.snapshotId(), branch);
             } else if (stageOnly) {
               update.addSnapshot(newSnapshot);
             } else {
-              update.setBranchSnapshot(newSnapshot, SnapshotRef.MAIN_BRANCH);
+              update.setBranchSnapshot(newSnapshot.snapshotId(), branch);
```
rdblue, line 313: I think this correctly writes to the branch, but I think we also need to check validations. That probably means copying _all_ the tests for every table operation and running them to make sure they work when writing to a branch.

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,26 @@ public void testIncludedPartitionSummaryLimit() {
     String changedPartitions = table.currentSnapshot().summary().get(SnapshotSummary.CHANGED_PARTITION_COUNT_PROP);
     Assert.assertEquals("Should set changed partition count", "2", changedPartitions);
   }
+
+  @Test
+  public void testAppendToBranch() {
```
rdblue, line 483: I think we need more tests for fast append:
* Where no snapshots or branches exist
* Where the named branch does not exist, but main does
* Possibly others?

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
namratha2403: @rdblue I agree with this. Do mean having a method like following in current PR?

`@Override
  public ThisT toBranch(String branch){
    throw new UnsupportedOperationException("Performing operations on a branch is currently not supported");
  }`
  
 And having the following method in next PR?
 
 `  @Override
  public ThisT toBranch(String branch){
    this.branchToUpdate = branch;
    return self();
  }`
  
  That would mean for this PR 

`update.setBranchSnapshot(newSnapshot.snapshotId(), branch);`

is not required ?

namratha2403:
> There is a more fundamental issue here for maintaining the correct snapshot lineage on a branch. The parent of the new snapshot needs to be set to the tip of the branch so that when the operation commits the lineage is correct.
> 
> Also there should be validation that the ref is actually a branch and not a tag since we cannot produce snapshots to a tag.
> 
> In apply https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L170
> 
> If the branch is main, the parent is the current snapshot
> 
> If the branch is not main, the parent is the tip of the branch which would be retrieved via table.ref(branch).snapshotId().

@amogh-jahagirdar  Not only the parent ID, even the manifests need to merged by taking the parent tip into account. 

Something like:

`@Override
  public List<ManifestFile> apply(TableMetadata base) {
    Snapshot current;
    if(branchToUpdate != null){
      current = base.snapshot(base.ref(branchToUpdate).snapshotId());
    }
    else{
      current = base.currentSnapshot();
      ...........
      ...........
    }`

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
namratha2403: @rdblue made changes to to parameter to take the type SnapshotRef so that end customer will be passing a valid snapshot ref.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
namratha2403: Made changes to this, i will be setting branch in next PR.

core/src/test/java/org/apache/iceberg/TestFastAppend.java
namratha2403: Since the signature of the method now changes, the end user is required to pass SnapshotRef, are these checks required?

namratha2403:
@rdblue Any other review comments on this PR or is this good to go ?

core/src/test/java/org/apache/iceberg/TestFastAppend.java
namratha2403: @rdblue Added checks for above

namratha2403:
@rdblue @amogh-jahagirdar I have introduced a method called ToBranch which checks for null branch and valid refs. In the next PR, will add the methods for using the set Branch. let me know if this looks right.

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
namratha2403: @rdblue  i added a check to check if branch exists. If it does not exist should we create new branch? I was thinking to throw an illegal arg exception.

_How does this change validation? What if the validation's from snapshot is not an ancestor of the branch? - > I did not understand this scenario._

namratha2403:
@jackye1995 Is it possible for you to review this?

cc: @SreeramGarlapati

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
amogh-jahagirdar: Each child of SnapshotProducer can implement their own validation logic https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L156

This gets called when trying to apply the operation as part of transaction conflict detection

amogh-jahagirdar:
Replied here https://github.com/apache/iceberg/pull/4926#discussion_r902380766 , but yes as @rdblue pointed out I think we'll need to think of defining conflict detection validation behavior in the presence of branching and tagging (such as the case where validation starts at a snapshot which is not an ancestor of the branch). I'll need to think more on this.

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
namrathamyske: @amogh-jahagirdar All validations must start from tip of branch, aka the current snapshot in the branch. So all the operations which override the ` protected void validate(TableMetadata currentMetadata) { }` aka - BaseOverwriteFiles, BaseReplacePartitions, BaseRewriteFiles, BaseRowDelta need to reference the current snapshot from the branch tip. 

Is any other validation needed here ?

namrathamyske:
@rdblue @amogh-jahagirdar can someone sign off on this PR so that Implementation for child operations can forward ?

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +117,38 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
     return self();
   }
 
+  @Override
+  public ThisT toBranch(String branch) {
+    throw new UnsupportedOperationException("Performing operations on a branch is currently not supported");
```
rdblue, line 122: Minor: I think this is a good time to use `this.getClass().getName()` so that it's clear which operation doesn't support committing to a branch.

```java
throw new UnsupportedOperationException(Sting.format(
    "Cannot commit to branch %s: %s does not support branch commits", branch,
    this.getClass().getName());
```

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +117,38 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
...
+   */
+  protected void setTargetBranch(String branch) {
+    this.targetBranch = branch;
+  }
+
+  /***
+   * A getter for the target branch on which snapshot producer operation should be performed
+   * @return target branch
+   */
+  protected String getTargetBranch() {
```
rdblue, line 148: Iceberg typically does not use `get` in method names because it is useless. Either it should be replaced by a more specific verb or it should be omitted.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +117,38 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
...
+
+  /***
+   * Will be used by snapshot producer operations to create a new ref if an invalid branch is passed
+   * @param branch ref name on which operation is to performed
+   */
+  protected void createNewRef(String branch) {
+    SnapshotRef branchRef = SnapshotRef.branchBuilder(this.current().currentSnapshot().snapshotId()).build();
+    TableMetadata.Builder updatedBuilder = TableMetadata.buildFrom(this.current());
+    updatedBuilder.setRef(branch, branchRef);
+    ops.commit(ops.current(), updatedBuilder.build());
```
rdblue, line 133: Committing to a branch while creating it should not commit twice. Instead, the commit should add a snapshot and then create a branch pointing to that snapshot.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -167,8 +200,7 @@ protected void validate(TableMetadata currentMetadata) {
   @Override
   public Snapshot apply() {
     refresh();
-    Long parentSnapshotId = base.currentSnapshot() != null ?
-        base.currentSnapshot().snapshotId() : null;
+    Long parentSnapshotId = base.ref(targetBranch) != null ? base.ref(targetBranch).snapshotId() : null;
```
rdblue, line 203: This should be the existing branch state, the current snapshot (if creating the branch), or null.

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,13 @@ public void testIncludedPartitionSummaryLimit() {
     String changedPartitions = table.currentSnapshot().summary().get(SnapshotSummary.CHANGED_PARTITION_COUNT_PROP);
     Assert.assertEquals("Should set changed partition count", "2", changedPartitions);
   }
+
+  @Test(expected = UnsupportedOperationException.class)
```
rdblue, line 482: To test exceptions, we use either `TestHelpers.assertThrows` or `Assertions`. That's important so that you can add assertions about the error message, and so that you can run assertions to validate that state did not change. In this case, the the branch name should not exist and the snapshot list should be the same size.

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,13 @@ public void testIncludedPartitionSummaryLimit() {
...
+  public void testAppendToBranch() throws UnsupportedOperationException {
+    table.newFastAppend()
+            .appendFile(FILE_A)
+            .commit();
+    table.manageSnapshots().createBranch("ref", table.currentSnapshot().snapshotId()).commit();
+    table.newFastAppend().appendFile(FILE_B).toBranch("ref").commit();
```
rdblue, line 488: Could this implement branch commits for `FastAppend`? That way we can see one operation that works from end-to-end when reviewing the changes to `SnapshotUpdate`.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -167,8 +200,7 @@ protected void validate(TableMetadata currentMetadata) {
...
     long sequenceNumber = base.nextSequenceNumber();
 
     // run validations from the child operation
```
rdblue, line 206: This needs to pass the parent snapshot ID into `validate` along with table metadata (`base`) so that validations can check changes between the starting snapshot ID and the parent snapshot, rather than between the starting snapshot ID and the current snapshot.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
rdblue: I see from #5010 that this is used in each child implementation to get the parent snapshot.

I think it would be better for maintainability and future implementations if the `apply` method were passed the parent snapshot.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
rdblue: Same thing with `apply`.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
amogh-jahagirdar: Yeah if we're creating a new branch, the transaction needs to encapsulate the branch creation and the producing of the snapshot.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +117,38 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
     return self();
   }
 
+  @Override
+  public ThisT toBranch(String branch) {
+    throw new UnsupportedOperationException("Performing operations on a branch is currently not supported");
+  }
+
+  /***
+   * Will be used by snapshot producer operations to create a new ref if an invalid branch is passed
```
amogh-jahagirdar, line 126: Can we also update the comment, instead of "invalid branch" I think it should be "create a new branch if it does not exist"

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +117,38 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
...
+   * @param branch ref name on which operation is to performed
+   */
+  protected void createNewRef(String branch) {
```
amogh-jahagirdar, line 129: Also this will always be a branch right, so can we call this method, createNewBranch?

core/src/test/java/org/apache/iceberg/TestFastAppend.java
namrathamyske: @rdblue Have written the complete test-case in https://github.com/apache/iceberg/pull/5010.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
namrathamyske: @rdblue My thought process is to check the whether starting snapshot ID is an ancestor of parent snapshot. If not then throw invalid starting snapshot ID exception. Else we should go ahead with checking between starting snapshot ID  & parent ID.

amogh-jahagirdar:
@namrathamyske Thanks for all the work on this! I've raised a PR https://github.com/namrathamyske/iceberg/pull/1/files to your fork, just addressing some of the comments @rdblue brought up, would love to get your thoughts on it.

amogh-jahagirdar:
Thanks for the review on the PR @namrathamyske.  @rdblue could we get a review on the updated PR when you get a chance?  Thanks!

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -487,4 +487,54 @@ public void testIncludedPartitionSummaryLimit() {
         table.currentSnapshot().summary().get(SnapshotSummary.CHANGED_PARTITION_COUNT_PROP);
     Assert.assertEquals("Should set changed partition count", "2", changedPartitions);
   }
+
+  @Test
+  public void testAppendToExistingBranch() {
+    table.newFastAppend().appendFile(FILE_A).commit();
+    table.manageSnapshots().createBranch("branch", table.currentSnapshot().snapshotId()).commit();
+    table.newFastAppend().appendFile(FILE_B).toBranch("branch").commit();
```
amogh-jahagirdar, line 495: Just realized I just hardcoded branch throughout the tests, probably cleaner to just define a variable in the tests and use that.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -150,28 +173,37 @@ public ThisT deleteWith(Consumer<String> deleteCallback) {
...
+  protected abstract List<ManifestFile> apply(TableMetadata metadataToUpdate, Snapshot snapshot);
 
   @Override
   public Snapshot apply() {
     refresh();
-    Long parentSnapshotId =
-        base.currentSnapshot() != null ? base.currentSnapshot().snapshotId() : null;
-    long sequenceNumber = base.nextSequenceNumber();
+    Snapshot parentSnapshot = base.currentSnapshot();
+    if (targetBranch != null) {
```
amogh-jahagirdar, line 193: Also considering targetBranch will always be set (we have it default to Main, we don't need this check).

core/src/test/java/org/apache/iceberg/TestOverwrite.java
```diff
@@ -300,4 +300,17 @@ public void testValidatedOverwriteWithAppendSuccess() {
...
+  public void testOverwriteToBranchUnsupported() {
+    AssertHelpers.assertThrows(
+        "Cannot commit to branch someBranch: org.apache.iceberg.BaseOverwriteFiles does not support branch commits",
+        UnsupportedOperationException.class,
+        () ->
+            table
+                .newOverwrite()
+                .overwriteByRowFilter(and(equal("date", "2018-06-09"), lessThan("id", 20)))
+                .addFile(FILE_10_TO_14)
+                .toBranch("someBranch"));
```
amogh-jahagirdar, line 314: I should fix this assertion so it actually checks the expected error message;  I think the assertThrows method with 3 params just surfaces this message in case the test fails, it doesn't actually check the expected error message. Like what's being done in TestReplacePartitions.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
amogh-jahagirdar: https://github.com/namrathamyske/iceberg/pull/1/files should take care of what @rdblue is talking about here which if I'm not mistaken is more about a cleaner abstraction for applying a change to a specific snapshot, and running a validation to a specific snapshot which will be needed when we implement branch support for the other producers.

core/src/test/java/org/apache/iceberg/TestOverwrite.java
rdblue: This is minor because it is going to be replaced very soon.

rdblue (approved):

core/src/test/java/org/apache/iceberg/TestFastAppend.java
rdblue: I think this is okay.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -181,7 +181,7 @@ protected void validate(TableMetadata currentMetadata, Snapshot snapshot) {}
    * Apply the update's changes to the base table metadata and return the new manifest list.
    *
    * @param metadataToUpdate the base table metadata to apply changes to
-   * @param snapshot
+   * @param snapshot parent snapshot passed to child operations.
```
amogh-jahagirdar, line 184: Ah thanks for adding the commenting.  Just a nit, I think this should say something like "snapshot to apply the changes to". Child operations doesn't make too much sense imo, here we're just saying "Apply the changes to this this metadata and specific snapshot".

api/src/main/java/org/apache/iceberg/SnapshotUpdate.java
```diff
@@ -59,4 +59,11 @@
    * @return this for method chaining
    */
   ThisT scanManifestsWith(ExecutorService executorService);
+
+  /**
+   * Perform operations on a particular branch
+   *
+   * @param branch which is name of SnapshotRef of type branch.
+   */
+  ThisT toBranch(String branch);
```
rdblue, line 68: This needs a default that throws `UnsupportedOperationException`.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -113,6 +114,28 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
     return self();
   }
 
+  @Override
+  public ThisT toBranch(String branch) {
+    throw new UnsupportedOperationException(
+        String.format(
+            "Cannot commit to branch %s: %s does not support branch commits",
+            branch, this.getClass().getName()));
+  }
```
rdblue, line 123: This should be in the interface, not here.

rdblue:
Looks like this is close, but we need to fix the `toBranch` definition and revapi checks.

amogh-jahagirdar:
@namrathamyske raised a PR to your fork to address the revapi and moving the default implementation up to the interface level. https://github.com/namrathamyske/iceberg/pull/3/files

namrathamyske:
@rdblue Merged PR with @amogh-jahagirdar's changes which had changes wrt to review comments.

amogh-jahagirdar:
Thanks @jackye1995 for triggering the approval workflow. @rdblue when you get a chance, let us know how the change looks!

jackye1995 (approved):
also looks good to me!

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
amogh-jahagirdar: good point, fixed

amogh-jahagirdar (approved):
Thanks for the contribution @namrathamyske !

jackye1995:
Looks like we are all in agreement, I will merge this to unblock next step implementations. thanks for the contribution @namrathamyske @amogh-jahagirdar ! and thanks for the review @rdblue !

PR #4929, opened 2022-06-01
Improve binary pack/unpack performance using cached struct

chulucninh09:
Instead of construct `Struct` each time in conversion functions using `struct(c_type).unpack`, we can create struct one in module and use the cached structs to pack and unpack, getting performance gain.

### DIFF
--- a/python/src/iceberg/conversions.py
+++ b/python/src/iceberg/conversions.py
@@ -27,7 +27,7 @@ Note:
     implementation, a concrete function is registered for each generic conversion function. For PrimitiveType
     implementations that share the same conversion logic, registrations can be stacked.
 """
-import struct
+from struct import Struct
 import uuid
 from decimal import Decimal
 from functools import singledispatch
@@ -52,6 +52,13 @@ from iceberg.types import (
 )
 from iceberg.utils.decimal import decimal_to_bytes, unscaled_to_decimal
 
+
+_bool_struct = Struct("<?")
+_int_struct = Struct("<i")
+_long_struct = Struct("<q")
+_float_struct = Struct("<f")
+_double_struct = Struct("<d")
+_uuid_struct = Struct(">QQ")
 
 def handle_none(func):
     """A decorator function to handle cases where partition values are `None` or "__HIVE_DEFAULT_PARTITION__"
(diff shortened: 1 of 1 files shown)


python/src/iceberg/conversions.py
```diff
@@ -53,6 +53,13 @@
 from iceberg.utils.decimal import decimal_to_bytes, unscaled_to_decimal
 
 
+_bool_struct = Struct("<?")
```
samredai, line 56: Do these count as constants? If so the python style guide recommends all caps for the name.

*PEP8 - Constants*
> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.

samredai (approved):
LGTM, thank you!

python/src/iceberg/conversions.py
Fokko: Since they are private (starting with an underscore), they shouldn't be used outside of the file. We could also consolidate this later on with my open PR: https://github.com/apache/iceberg/pull/4920/files#diff-9747a5712c71edff07e52c56ae999d30fa9dde758ba94e376acd16819e055731R23-R27

chulucninh09: @samredai @Fokko correct me if I'm wrong. I consider this cached instance instead of constant because most of the time const is scalar value to me. These are class instance, kind of too complex object to consider it const by my understanding

kbendick: I would personally still think of it as a constant regardless of complexity, but I agree so long as they're not used outside of this file that's the more important thing to me.

kbendick (approved):

Fokko:
@chulucninh09 Can you run the linter? You can easily do this by running:
```python
make install
make lint
```
This will make the CI pass 👍🏻

chulucninh09:
> @chulucninh09 Can you run the linter? You can easily do this by running:
> 
> ```shell
> make install
> make lint
> ```
> 
> This will make the CI pass 👍🏻

I linted the code. And also I change the naming to all cap with `_` to follow the convention

python/src/iceberg/conversions.py
chulucninh09: I changed it to all caps with leading underscore to follow the convention @Fokko @kbendick @samredai

rdblue (approved):

rdblue:
Thanks, @chulucninh09!

PR #4935, opened 2022-06-01
Python: Add pyupgrade as a pre-commit hook

Fokko:
The pyupgrade plugin will check the syntax, and make sure that you're using the latest version. For example, instead of Union[str, int] we have the str | int notation in Python 3.8.

This hook will not change any behavior of the code, but will just make sure that we're all using the same syntax.

More information on:
https://github.com/asottile/pyupgrade

### DIFF
--- a/python/.pre-commit-config.yaml
+++ b/python/.pre-commit-config.yaml
@@ -45,3 +45,8 @@ repos:
       hooks:
         - id: pycln
           args: [--config=python/pyproject.toml]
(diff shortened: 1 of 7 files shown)


samredai (approved):
LGTM, thanks @Fokko!

rdblue (approved):

rdblue:
Thanks, @Fokko!

PR #4936, opened 2022-06-01
Core, API: Add support for renaming tags

amogh-jahagirdar:
From discussion on https://docs.google.com/document/d/1tbATFPrKF3vNlzkgZQdaW8CAJmbjvryfrlg6C2Ci_aA/edit#, there's interest in renaming tags which is currently unsupported.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/UpdateSnapshotReferencesOperation.java
+++ b/core/src/main/java/org/apache/iceberg/UpdateSnapshotReferencesOperation.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg;
 
+import java.util.Locale;
 import java.util.Map;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/UpdateSnapshotReferencesOperation.java
```diff
@@ -85,16 +86,29 @@ public UpdateSnapshotReferencesOperation removeTag(String name) {
...
+    renameRef(name, newName, SnapshotRefType.BRANCH);
+    return this;
+  }
+
+  public UpdateSnapshotReferencesOperation renameTag(String name, String newName) {
+    renameRef(name, newName, SnapshotRefType.TAG);
+    return this;
+  }
+
+  private void renameRef(String name, String newName, SnapshotRefType type) {
```
amogh-jahagirdar, line 98: Still don't really know about this helper method. I added it to reduce code duplication for renameBranch/renameTag as they are basically the same (check if the ref exists and it's the expected type. Then verify the target ref doesn't already exist. Remove the old ref and add the new ref), it's just the messaging that's different and we expose the branch/tag specific methods so it's more explicit to the user of the API .

Code duplication is reduced but there's branching logic just for the different names in the message which seems overkill. May be worth the code duplication if the community feels it's more readable that way

core/src/main/java/org/apache/iceberg/UpdateSnapshotReferencesOperation.java
```diff
@@ -85,16 +86,30 @@ public UpdateSnapshotReferencesOperation removeTag(String name) {
...
     Preconditions.checkArgument(!name.equals(SnapshotRef.MAIN_BRANCH), "Cannot rename main branch");
+    String refTypeInMessage = type == SnapshotRefType.BRANCH ? "Branch" : "Tag";
```
ccforbes, line 100: nit: could just reference the enum value as a string in the checks, could make the argument about having more than two types of snapshots in the future, but i wouldn't know what those could be. this does help with readability though

core/src/main/java/org/apache/iceberg/UpdateSnapshotReferencesOperation.java
```diff
@@ -85,16 +86,30 @@ public UpdateSnapshotReferencesOperation removeTag(String name) {
...
+    Preconditions.checkNotNull(name, "%s to rename cannot be null", refTypeInMessage);
+    Preconditions.checkNotNull(newName, "New %s name cannot be null", refTypeInMessage.toLowerCase(Locale.ROOT));
     SnapshotRef ref = updatedRefs.get(name);
-    Preconditions.checkArgument(ref != null, "Branch does not exist: %s", name);
-    Preconditions.checkArgument(ref.isBranch(), "Ref %s is a tag not a branch", name);
+    Preconditions.checkArgument(ref != null, "%s does not exist: %s", refTypeInMessage, name);
```
ccforbes, line 104: could be a checkNotNull?

PR #4938, opened 2022-06-02
Support reading Parquet row group bloom filter

huaxingao:
Co-Authored-By: Xi Chen [jshmchenxi@163.com](mailto:jshmchenxi@163.com)
Co-Authored-By: Hao Lin [linhao@qiyi.com](mailto:linhao@qiyi.com)
Co-Authored-By: Huaxin Gao [huaxin_gao@apple.com](mailto:huaxin_gao@apple.com)

This is the read path of parquet row group bloom filter. The original PR is [here](https://github.com/apache/iceberg/pull/4831/)

### DIFF
--- a/api/src/main/java/org/apache/iceberg/expressions/Binder.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Binder.java
@@ -91,20 +91,19 @@ public class Binder {
   }
 
   public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
-    return references(struct, exprs, caseSensitive, false);
-  }
-
-  public static Set<Integer> references(
-      StructType struct, List<Expression> exprs, boolean caseSensitive, boolean alreadyBound) {
     if (exprs == null) {
       return ImmutableSet.of();
     }
     ReferenceVisitor visitor = new ReferenceVisitor();
     for (Expression expr : exprs) {
-      if (!alreadyBound) {
+      try {
         ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
-      } else {
-        ExpressionVisitors.visit(expr, visitor);
+      } catch (IllegalStateException e) {
+        if (e.getMessage().contains("Found already bound predicate")) {
+          ExpressionVisitors.visit(expr, visitor);
+        } else {
+          throw e;
+        }
       }
     }
     return visitor.references;
--- a/api/src/main/java/org/apache/iceberg/expressions/Binder.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Binder.java
@@ -91,12 +91,21 @@ public class Binder {
   }
 
   public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
+    return references(struct, exprs, caseSensitive, false);
+  }
+
+  public static Set<Integer> references(
+      StructType struct, List<Expression> exprs, boolean caseSensitive, boolean alreadyBound) {
     if (exprs == null) {
       return ImmutableSet.of();
     }
     ReferenceVisitor visitor = new ReferenceVisitor();
     for (Expression expr : exprs) {
-      ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
+      if (!alreadyBound) {
+        ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
+      } else {
+        ExpressionVisitors.visit(expr, visitor);
+      }
     }
     return visitor.references;
   }
--- a/api/src/main/java/org/apache/iceberg/expressions/Binder.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/Binder.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Set;
 import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.expressions.ExpressionVisitors.ExpressionVisitor;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.types.Type;
@@ -96,17 +97,27 @@ public class Binder {
     }
     ReferenceVisitor visitor = new ReferenceVisitor();
     for (Expression expr : exprs) {
-      try {
+      if (isBound(expr)) {
+        ExpressionVisitors.visit(expr, visitor);
+      } else {
         ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
-      } catch (IllegalStateException e) {
-        if (e.getMessage().contains("Found already bound predicate")) {
-          ExpressionVisitors.visit(expr, visitor);
-        } else {
-          throw e;
-        }
       }
     }
     return visitor.references;
+  }
+
+  /**
+   * Returns whether an expression is bound.
+   * <p>
+   * An expression is bound if all of its predicates are bound.
+   *
+   * @param expr an {@link Expression}
+   * @return true if the expression is bound
+   * @throws IllegalArgumentException if the expression has both bound and unbound predicates.
+   */
+  public static boolean isBound(Expression expr) {
+    Boolean isBound = ExpressionVisitors.visit(expr, new IsBoundVisitor());
+    return isBound != null ? isBound : false; // assume unbound if undetermined
   }
 
   private static class BindVisitor extends ExpressionVisitor<Expression> {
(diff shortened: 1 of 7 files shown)


huaxingao:
I don't have a good way to test the read path, but I have tested these changes with the write path on my local.

huaxingao:
cc @rdblue @RussellSpitzer @kbendick @chenjunjiedada @hililiwei

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,331 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.parquet;
```
rdblue, line 20: Is it possible to include the unit test in this PR? I think all you'd need to do is configure the bloom filter settings using the Parquet settings in a Hadoop Configuration rather than through the Iceberg write settings.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
```diff
@@ -91,12 +91,21 @@ static Expression bind(StructType struct,
   }
 
   public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
+    return references(struct, exprs, caseSensitive, false);
+  }
+
+  public static Set<Integer> references(
+      StructType struct, List<Expression> exprs, boolean caseSensitive, boolean alreadyBound) {
```
rdblue, line 98: Can we detect that an expression is unbound? Maybe identify named refs and return? Then we could just use one method for everything.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: I tried to config using  `ParquetOutputFormat.BLOOM_FILTER_ENABLED` by replacing this [line](
https://github.com/apache/iceberg/blob/1fabb2db59d6db33609f549024c4371873ad3b10/parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java#L201) with `.set(ParquetOutputFormat.BLOOM_FILTER_ENABLED + "#_id", "true") `, but it doesn't work.

Seems Iceberg only honors the Iceberg's properties (which are set by `Context` at [here](https://github.com/apache/iceberg/blob/master/parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java#L233)), but it doesn't really take the properties set by Parquet.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
huaxingao: Can i just use `instanceof Unbound` to detect if an expression is unbound? Something like this:

```
  private static boolean isUnbound(Expression expr) {
    switch (expr.op()) {
      case TRUE:
        return false;
      case FALSE:
        return false;
      case NOT:
        Not not = (Not) expr;
        return isUnbound(not.child());
      case AND:
        And and = (And) expr;
        return isUnbound(and.left()) || isUnbound(and.right());
      case OR:
        Or or = (Or) expr;
        return isUnbound(or.left()) || isUnbound(or.right());
      default:
        return expr instanceof Unbound;
    }
  }
```
Then the method `boundReferences` can be
```
  public static Set<Integer> boundReferences(StructType struct, List<Expression> exprs, boolean caseSensitive) {
    if (exprs == null) {
      return ImmutableSet.of();
    }
    ReferenceVisitor visitor = new ReferenceVisitor();
    for (Expression expr : exprs) {
      if (isUnbound(expr)) {
        ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
      } else {
        ExpressionVisitors.visit(expr, visitor);
      }
    }
    return visitor.references;
  }
```

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
rdblue: I think that properties will be passed through to the Hadoop `Configuration` automatically. Is that no longer true?

api/src/main/java/org/apache/iceberg/expressions/Binder.java
huaxingao: Sounds good to me!

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
huaxingao: Should I set the Parquet properties to `encodingPropsBuilder`? If the same properties is also set by iceberg properties, I will reset to overwrite.

rdblue: I see. Is it possible to get some tests working without the write-side changes? Maybe write a Parquet file directly and use name mapping? If not then let's try to make the minimal write-side changes to get the test in.

huaxingao: I made some minimal write-side changes to get the test in. Hope this is OK.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetBloomRowGroupFilter.java
```diff
@@ -0,0 +1,331 @@
...
+
+  /**
+   * Tests whether the bloom for a row group may contain records that match the expression.
+   *
+   * @param fileSchema  schema for the Parquet file
+   * @param bloomReader a bloom filter reader
+   * @return false if the file cannot contain rows that match the expression, true otherwise.
+   */
+  public boolean shouldRead(MessageType fileSchema, BlockMetaData rowGroup,
+      BloomFilterReader bloomReader) {
```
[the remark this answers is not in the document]
huaxingao: Thanks for the reminder. I will pay attention to the style next time.

api/src/main/java/org/apache/iceberg/expressions/Binder.java
```diff
@@ -96,7 +96,15 @@ public static Set<Integer> boundReferences(StructType struct, List<Expression> e
     }
     ReferenceVisitor visitor = new ReferenceVisitor();
     for (Expression expr : exprs) {
-      ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
+      try {
+        ExpressionVisitors.visit(bind(struct, expr, caseSensitive), visitor);
+      } catch (IllegalStateException e) {
+        if (e.getMessage().contains("Found already bound predicate")) {
```
rdblue, line 102: We shouldn't rely on exception messages like this.

I think instead we should just add a utility to detect whether an expression is bound. Here's an implementation:

```java
  /**
   * Returns whether an expression is bound.
   * <p>
   * An expression is bound if all of its predicates are bound.
   *
   * @param expr an {@link Expression}
   * @return true if the expression is bound
   * @throws IllegalArgumentException if the expression has both bound and unbound predicates.
   */
  public static boolean isBound(Expression expr) {
    Boolean isBound = ExpressionVisitors.visit(expr, new IsBoundVisitor());
    return isBound != null ? isBound : false; // assume unbound if undetermined
  }

  private static class IsBoundVisitor extends ExpressionVisitors.ExpressionVisitor<Boolean> {
    @Override
    public Boolean not(Boolean result) {
      return result;
    }

    @Override
    public Boolean and(Boolean leftResult, Boolean rightResult) {
      return combineResults(leftResult, rightResult);
    }

    @Override
    public Boolean or(Boolean leftResult, Boolean rightResult) {
      return combineResults(leftResult, rightResult);
    }

    @Override
    public <T> Boolean predicate(BoundPredicate<T> pred) {
      return true;
    }

    @Override
    public <T> Boolean predicate(UnboundPredicate<T> pred) {
      return false;
    }

    private Boolean combineResults(Boolean isLeftBound, Boolean isRightBound) {
      if (isLeftBound != null) {
        Preconditions.checkArgument(isRightBound == null || isLeftBound.equals(isRightBound),
            "Found partially bound expression");
        return isLeftBound;
      } else {
        return isRightBound;
      }
    }
  }
```

parquet/src/main/java/org/apache/iceberg/parquet/Parquet.java
```diff
@@ -291,9 +292,17 @@ public <D> FileAppender<D> build() throws IOException {
             .withWriteMode(writeMode)
             .withRowGroupSize(rowGroupSize)
             .withPageSize(pageSize)
-            .withDictionaryPageSize(dictionaryPageSize)
-            .build(),
-            metricsConfig);
+            .withDictionaryPageSize(dictionaryPageSize);
+        // Todo: The following code needs to be improved in the bloom filter write path PR.
```
rdblue, line 296: I think this is okay for now to get this in with tests. Thanks, @huaxingao!

api/src/main/java/org/apache/iceberg/expressions/Binder.java
huaxingao: Sounds great! Thank you very much for the implementation!

rdblue: No problem! Thanks for all your work on this.

rdblue (approved):

rdblue:
Thanks, @huaxingao!

huaxingao:
Thank you very much @rdblue!  Also thank you @RussellSpitzer @kbendick @chenjunjiedada @hililiwei @stevenzwu

kbendick:
Thank you for making support for Parquet bloom filters happen @huaxingao! Your work is highly appreciated.

Issue #4931, opened 2022-06-01
Rewrite Data Files with Manual Sort Order should also use Table Partitioning in Sort Order


RussellSpitzer:
Currently when a user provides a SortOrder to RewriteDataFiles to manually override the table sort order it ignores the underlying table partitioning sort. We should still include this even if we are ignoring the table's set SortOrder for other columns.

rajarshisarkar:
I would like to contribute this enhancement. One possible approach I can think of is to combine both the Sort Order (table sort order + manual sort order). Please let me know your thoughts.

RussellSpitzer:
@rajarshisarkar If you have a chance see if you can get a repo test as well. I think in the particular case that this was reported to me it may have also been a rewrite between partition specs with a manually specified order...

PR #4941, opened 2022-06-02
Core: Use Table Partitioning sort with manual Sort Order in Rewrite Data Files

rajarshisarkar:
Currently when a user provides a SortOrder to RewriteDataFiles to manually override the table sort order it ignores the underlying table partitioning sort. We should still include this even if we are ignoring the table's set SortOrder for other columns.

Closes https://github.com/apache/iceberg/issues/4931

----
cc: @RussellSpitzer

### DIFF
--- a/spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
+++ b/spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
@@ -18,6 +18,8 @@
  */
 package org.apache.iceberg.spark.actions;
 
+import static org.apache.iceberg.NullOrder.NULLS_LAST;
+import static org.apache.iceberg.expressions.Expressions.bucket;
 import static org.apache.iceberg.types.Types.NestedField.optional;
 import static org.apache.iceberg.types.Types.NestedField.required;
 import static org.apache.spark.sql.functions.current_date;
@@ -1318,6 +1320,19 @@ public class TestRewriteDataFilesAction extends SparkTestBase {
     Assert.assertEquals("Number of files order should be descending", actual, expected);
     Collections.reverse(expected);
     Assert.assertNotEquals("Number of files order should not be ascending", actual, expected);
+  }
+
+  @Test
+  public void testPartitioningSortWithManualSortOnSameColumn() {
+    Table table = createTablePartitioned(4, 2);
+    writeRecords(4, SCALE, 4);
+    shouldHaveFiles(table, 24);
+    table.updateSpec().addField(bucket("c3", 16)).commit();
+
+    basicRewrite(table)
+        .sort(SortOrder.builderFor(SCHEMA).asc(Expressions.bucket("c3", 16), NULLS_LAST).build())
+        .execute();
+    basicRewrite(table).sort(SortOrder.builderFor(SCHEMA).asc("c3").build()).execute();
   }
 
   private Stream<RewriteFileGroup> toGroupStream(Table table, RewriteDataFilesSparkAction rewrite) {


core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -48,7 +49,7 @@ public abstract class SortStrategy extends BinPackStrategy {
    * @return this for method chaining
    */
   public SortStrategy sortOrder(SortOrder order) {
-    this.sortOrder = order;
+    this.sortOrder = SortOrderUtil.buildSortOrder(table(), order);
```
rajarshisarkar, line 52: This builds a sort order using both the table partition spec and the user supplied sort order.

core/src/test/java/org/apache/iceberg/actions/TestSortStrategy.java
```diff
@@ -86,9 +86,6 @@ private List<FileScanTask> tasksForSortOrder(int sortOrderId, int... fileSizesMB
 
   @Test
   public void testInvalidSortOrder() {
-    AssertHelpers.assertThrows("Should not allow an unsorted Sort order", IllegalArgumentException.class,
-        () -> defaultSort().sortOrder(SortOrder.unsorted()).options(Collections.emptyMap()));
```
rajarshisarkar, line 90: This test fails now as the table has a partition spec:
```
[
  1000: data_bucket: bucket[16](2)
]
```

Resulting in the overall Sort Order to be:

```
[
  bucket[16](2) ASC NULLS FIRST
]
```

Hence, no scope of `IllegalArgumentException`

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
rdblue: This looks correct to me.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
```diff
@@ -48,7 +49,7 @@ public abstract class SortStrategy extends BinPackStrategy {
    * @return this for method chaining
    */
   public SortStrategy sortOrder(SortOrder order) {
-    this.sortOrder = order;
```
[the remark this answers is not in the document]
rajarshisarkar: @rdblue, I have added the check.

RussellSpitzer (commented):
I have a small request that we add a quick test to make sure this is doing what we expect. The error we saw in practice was a "already closed files for partition x ... " which I think we could repo here just by having a bucketed table and then requesting a manual sort on the bucketed column.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
rdblue: Yes, this does happen. We can technically suppress the second occurrence of `X`, but it seems minor to me since it is just going to make the sort use a bit more data. I think this is something that we can clean up in `buildSortOrder` in the future, after this PR.

rajarshisarkar:
Sure, @rdblue!

@RussellSpitzer I have added the test. Although, I couldn't reproduce the error `already closed files for partition x ... `.

core/src/main/java/org/apache/iceberg/actions/SortStrategy.java
rajarshisarkar: Sounds good!

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1264,6 +1266,23 @@ public void testRewriteJobOrderFilesDesc() {
     Assert.assertNotEquals("Number of files order should not be ascending", actual, expected);
   }
 
+  @Test
+  public void testPartitioningSortWithManualSortOnSameColumn() {
+    Table table = createTablePartitioned(4, 2);
+    table.refresh();
```
rdblue, line 1272: Does this need to be called? The table instance returned by `createTablePartitioned` should be up to date.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
```diff
@@ -1264,6 +1266,23 @@ public void testRewriteJobOrderFilesDesc() {
...
+    table.updateSpec().addField(bucket("c3", 16)).commit();
+    Assert.assertFalse(table.specs().isEmpty());
+    Assert.assertFalse(table.spec().fields().isEmpty());
```
rdblue, line 1275: I don't think that these assertions are needed, since this isn't trying to validate the behavior of `updateSpec`. This should just assume that it the update worked.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
rajarshisarkar: It shouldn't be required. I have removed it.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestRewriteDataFilesAction.java
rajarshisarkar: I have removed it.

PR #4942, opened 2022-06-02
[Spark][Test]: Check before creating default namespace to avoid noisy AlreadyExistsExceptions in test logs

sumeetgajjar:
This PR aims at removing the unwanted `AlreadyExistsException` stacktrace from the spark-iceberg test logs.

The test logs for spark tests are filled with `AlreadyExistsException(message:Database default already exists)` stacktrace even when all the tests pass.

These `AlreadyExistsException` exceptions are not from Spark per se but from HMS.

When "CREATE NAMESPACE IF NOT EXISTS default" SQL command is executed in Spark, Spark invokes `hive.createDatabase` command. 
https://github.com/apache/spark/blob/89fdb8a6fb6a669c458891b3abeba236e64b1e89/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala#L574

Hive client invokes internally invokes HMS API to create the database. If the DB already exists HMS throws `AlreadyExistsException`. When the `ifNotExist` flag is set to true, the Hive client simply ignores the exception.
https://github.com/apache/hive/blob/63326ff775206e59547b6b1332e25279e90ef5ee/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L608-L619

The HMS logs this exception to STDERR and for iceberg tests since a standalone HMS is running in the same JVM as that of the test, these logs are part of the info output of the tests.

This generates a lot of noise in the logs and might overshadow an actual exception.

### DIFF
--- a/spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
+++ b/spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
@@ -87,7 +87,10 @@ public class TestSparkReaderDeletes extends DeleteReadTests {
                 HiveCatalog.class.getName(), "hive", ImmutableMap.of(), hiveConf);
 
     try {
-      catalog.createNamespace(Namespace.of("default"));
+      Namespace defaultNamespace = Namespace.of("default");
+      if (!catalog.namespaceExists(defaultNamespace)) {
+        catalog.createNamespace(defaultNamespace);
+      }
     } catch (AlreadyExistsException ignored) {
       // the default namespace already exists. ignore the create error
     }
--- a/spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
+++ b/spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
@@ -41,7 +41,6 @@ import org.apache.iceberg.data.DeleteReadTests;
 import org.apache.iceberg.data.FileHelpers;
 import org.apache.iceberg.data.GenericRecord;
 import org.apache.iceberg.data.Record;
-import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.hive.HiveCatalog;
 import org.apache.iceberg.hive.TestHiveMetastore;
 import org.apache.iceberg.io.CloseableIterable;
@@ -86,13 +85,9 @@ public class TestSparkReaderDeletes extends DeleteReadTests {
             CatalogUtil.loadCatalog(
                 HiveCatalog.class.getName(), "hive", ImmutableMap.of(), hiveConf);
 
-    try {
-      Namespace defaultNamespace = Namespace.of("default");
-      if (!catalog.namespaceExists(defaultNamespace)) {
-        catalog.createNamespace(defaultNamespace);
-      }
-    } catch (AlreadyExistsException ignored) {
-      // the default namespace already exists. ignore the create error
+    Namespace defaultNamespace = Namespace.of("default");
+    if (!catalog.namespaceExists(defaultNamespace)) {
+      catalog.createNamespace(defaultNamespace);
     }
   }
 
--- a/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
+++ b/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
@@ -76,7 +76,10 @@ public abstract class SparkTestBase {
                 HiveCatalog.class.getName(), "hive", ImmutableMap.of(), hiveConf);
 
     try {
-      catalog.createNamespace(Namespace.of("default"));
+      Namespace defaultNamespace = Namespace.of("default");
+      if (!catalog.namespaceExists(defaultNamespace)) {
+        catalog.createNamespace(defaultNamespace);
+      }
     } catch (AlreadyExistsException ignored) {
       // the default namespace already exists. ignore the create error
     }
--- a/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
+++ b/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
@@ -249,9 +249,9 @@ public abstract class SparkTestBase {
     Assert.assertEquals("Expected metric value not match", expectedMetrics, currentMetrics);
   }
 
-  protected static void createNamespace(SupportsNamespaces catalog, Namespace namespace) {
-    if (!catalog.namespaceExists(namespace)) {
-      catalog.createNamespace(namespace);
+  protected static void createNamespace(SupportsNamespaces supportsNamespaces, Namespace namespace) {
+    if (!supportsNamespaces.namespaceExists(namespace)) {
+      supportsNamespaces.createNamespace(namespace);
     }
   }
 
--- a/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
+++ b/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
@@ -30,6 +30,7 @@ import java.util.stream.IntStream;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.iceberg.CatalogUtil;
 import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.SupportsNamespaces;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.hive.HiveCatalog;
 import org.apache.iceberg.hive.TestHiveMetastore;
@@ -75,14 +76,7 @@ public abstract class SparkTestBase {
             CatalogUtil.loadCatalog(
                 HiveCatalog.class.getName(), "hive", ImmutableMap.of(), hiveConf);
 
-    try {
-      Namespace defaultNamespace = Namespace.of("default");
-      if (!catalog.namespaceExists(defaultNamespace)) {
-        catalog.createNamespace(defaultNamespace);
-      }
-    } catch (AlreadyExistsException ignored) {
-      // the default namespace already exists. ignore the create error
-    }
+    createNamespace(catalog, Namespace.of("default"));
   }
 
   @AfterClass
@@ -254,6 +248,16 @@ public abstract class SparkTestBase {
             .filter(x -> metricsIds.containsKey(x.getKey()))
             .collect(Collectors.toMap(x -> metricsIds.get(x.getKey()), x -> x.getValue()));
     Assert.assertEquals("Expected metric value not match", expectedMetrics, currentMetrics);
+  }
+
+  protected static void createNamespace(SupportsNamespaces catalog, Namespace namespace) {
+    try {
+      if (!catalog.namespaceExists(namespace)) {
+        catalog.createNamespace(namespace);
+      }
+    } catch (AlreadyExistsException ignored) {
+      // the namespace already exists. ignore the create error
+    }
   }
 
   @FunctionalInterface
--- a/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
+++ b/spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
@@ -249,7 +249,8 @@ public abstract class SparkTestBase {
     Assert.assertEquals("Expected metric value not match", expectedMetrics, currentMetrics);
   }
 
-  protected static void createNamespace(SupportsNamespaces supportsNamespaces, Namespace namespace) {
+  protected static void createNamespace(
+      SupportsNamespaces supportsNamespaces, Namespace namespace) {
     if (!supportsNamespaces.namespaceExists(namespace)) {
       supportsNamespaces.createNamespace(namespace);
     }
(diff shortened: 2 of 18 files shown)


sumeetgajjar:
Hi @kbendick, @RussellSpitzer, @rdblue, @aokolnychyi 
Can you please review this PR?

sumeetgajjar:
cc: @marton-bod @pvary

pvary:
Is this exception in the logs is soo confusing, so this worth the complexity in the test code? Coming from Hive background seeing this exception seems perfectly okay, but I have 6 years of indoctrination so I am not a representative example.

sumeetgajjar:
Hi @pvary, thank you for your comments.

> Is this exception in the logs is soo confusing, so this worth the complexity in the test code?

In this case, the refactored test code does add additional checks but I believe those are justifiable given they reduce noise from the test logs. IMO, the less noise we generate during tests the better it is while triaging test failures. Especially spark tests where we also have tons of logs from the spark as well.

> Coming from Hive background seeing this exception seems perfectly okay, but I have 6 years of indoctrination so I am not a representative example.

Maybe, I believe it is subjective. 

I was recently triaging a couple of spark-iceberg test failures in our downstream iceberg branches and the test logs were full of the above exception stacktrace. The actual failures were shadowed by them and it took me a while to narrow down on the right stacktrace.

pvary:
@sumeetgajjar: Could you please check the failures?

I am comfortable with this:
```
      Namespace defaultNamespace = Namespace.of("default");
      if (!catalog.namespaceExists(defaultNamespace)) {
        catalog.createNamespace(defaultNamespace);
      }
``` 

I am not entirely sure about these:
```
    boolean createNamespace = isHadoopCatalog || spark.sql("SHOW NAMESPACES IN " + catalogName)
        .filter("namespace = 'default'")
        .isEmpty();
    if (createNamespace) {
      sql("CREATE NAMESPACE IF NOT EXISTS " + catalogName + ".default");
    }
```

sumeetgajjar:
> Could you please check the failures?

Fixed the failures. 

> I am not entirely sure about these:

@pvary simplified the logic a bit, can you please check again? 


Also, can you please approve the workflows?

kbendick:
> @sumeetgajjar: Could you please check the failures?
> 
> I am comfortable with this:
> 
> ```
>       Namespace defaultNamespace = Namespace.of("default");
>       if (!catalog.namespaceExists(defaultNamespace)) {
>         catalog.createNamespace(defaultNamespace);
>       }
> ```
> 
> I am not entirely sure about these:
> 
> ```
>     boolean createNamespace = isHadoopCatalog || spark.sql("SHOW NAMESPACES IN " + catalogName)
>         .filter("namespace = 'default'")
>         .isEmpty();
>     if (createNamespace) {
>       sql("CREATE NAMESPACE IF NOT EXISTS " + catalogName + ".default");
>     }
> ```

Ditto. I also have concerns about the additional usage of `spark.sql` (and the general complexity of that part). I also consider this to be a normal-ish log, along the lines of "illegal reflective access" and similar things.

sumeetgajjar:
Hi @pvary @kbendick,
Thank you for the comments.

> I also consider this to be a normal-ish log, along the lines of "illegal reflective access" and similar things.

I agree that some exceptions (in this case `AlreadyExistsException` and as Kyle mentioned "illegal reflective access") can be considered normal after prolonged exposure to them and we grow oblivious to them. 

However, I still feel if we can avoid them in the first place, then we should. 

> I also have concerns about the additional usage of spark.sql (and the general complexity of that part).

I've removed the use of `spark.sql` and simplified the namespace exists check logic to simply use `catalog.namespaceExists`. Please let me know how you feel about those. 

If it still feels complex, I can revert the changes in `SparkTestBaseWithCatalog` and simply keep the part in `SparkTestBase` that was deemed comfortable.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/SparkTestBaseWithCatalog.java
```diff
@@ -80,13 +80,24 @@ public SparkTestBaseWithCatalog(String catalogName, String implementation, Map<S
...
 
-    sql("CREATE NAMESPACE IF NOT EXISTS default");
+    // When Hive API is invoked to create Namespace/Database using CREATE IF NOT EXISTS, the standalone HMS logs an
+    // AlreadyExistsException exception in the test logs thereby generating unwanted error logs. Thus checking if a
+    // namespace exists before creating one to keep the test logs noise-free.
+    // HMS is not involved in case HadoopCatalog, thus skipping the check for HadoopCatalog.
+    if (isHadoopCatalog) {
+      sql("CREATE NAMESPACE IF NOT EXISTS default");
+    } else {
+      if (!validationNamespaceCatalog.namespaceExists(Namespace.of("default"))) {
```
pvary, line 97: Would this work for HadoopCatalog as well? So we can just keep the code more simple?

sumeetgajjar: Hi @pvary - simplified the namespace creation logic further.
 
However, that failed `TestNamespaceSQL#testListNamespace` test so had to fix it. The fix actually simplifies the assertion logic in `TestNamespaceSQL#testListNamespace` and removes a conditional as well :)

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkReaderDeletes.java
```diff
@@ -85,7 +85,10 @@ public static void startMetastoreAndSpark() {
         CatalogUtil.loadCatalog(HiveCatalog.class.getName(), "hive", ImmutableMap.of(), hiveConf);
 
     try {
-      catalog.createNamespace(Namespace.of("default"));
+      Namespace defaultNamespace = Namespace.of("default");
```
pvary, line 88: Do we need the try-catch after this change?

sumeetgajjar: In a perfect world no. 
However, this is just due diligence to avoid any unwanted failures.

pvary: This is test code. I would remove anything which is not strictly needed.

Also getting an AlreadyExistsException after checking for existence should be a bug if there is no concurrency.

sumeetgajjar: Noted, removed try-catch while creating namespace in the latest commit :)

spark/v3.0/spark/src/test/java/org/apache/iceberg/spark/SparkTestBase.java
```diff
@@ -73,11 +73,7 @@ public static void startMetastoreAndSpark() {
     SparkTestBase.catalog = (HiveCatalog)
         CatalogUtil.loadCatalog(HiveCatalog.class.getName(), "hive", ImmutableMap.of(), hiveConf);
 
-    try {
-      catalog.createNamespace(Namespace.of("default"));
-    } catch (AlreadyExistsException ignored) {
-      // the default namespace already exists. ignore the create error
-    }
+    createNamespace(catalog, Namespace.of("default"));
```
pvary, line 76: This is really just a question: does it worth to create a twoliner method which is used only 2 places?

sumeetgajjar: In this case, it would be a good idea to have that method to ensure the namespace is created in the same manner i.e using Catalog methods.
I'm fine with removing the method if required.

pvary: Your answer helped me to understand, why I felt ambiguous about this solution. We are in the same situation here than in the other places, like TestSparkReaderDeletes.java, and there we chose to inline the check.

sumeetgajjar: > We are in the same situation here than in the other places, like TestSparkReaderDeletes.java, and there we chose to inline the check.

`SparkTestBaseWithCatalog` extends `SparkTestBase`, thus the method can be reused in the child class. 
However, `TestSparkReaderDeletes` does not fall in the above inheritance hierarchy, thus we had to inline it. 

There also does not exist a TestUtil class for Spark tests or else we could have added the method to the util class and reused it everywhere.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/SparkTestBaseWithCatalog.java
```diff
@@ -86,7 +86,7 @@ public SparkTestBaseWithCatalog(String catalogName, String implementation, Map<S
 
     this.tableName = (catalogName.equals("spark_catalog") ? "" : catalogName + ".") + "default.table";
 
-    sql("CREATE NAMESPACE IF NOT EXISTS default");
+    createNamespace(validationNamespaceCatalog, Namespace.of("default"));
```
rdblue, line 89: Why is this needed? `IF NOT EXISTS` should check existence first, right?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/sql/TestNamespaceSQL.java
```diff
@@ -134,15 +134,9 @@ public void testListNamespace() {
...
-      Set<String> namespaceNames = namespaces.stream().map(arr -> arr[0].toString()).collect(Collectors.toSet());
-      Assert.assertEquals("Should have only db namespace", ImmutableSet.of("db"), namespaceNames);
-    } else {
-      Assert.assertEquals("Should have 2 namespaces", 2, namespaces.size());
-      Set<String> namespaceNames = namespaces.stream().map(arr -> arr[0].toString()).collect(Collectors.toSet());
-      Assert.assertEquals("Should have default and db namespaces", ImmutableSet.of("default", "db"), namespaceNames);
-    }
+    Assert.assertEquals("Should have 2 namespaces", 2, namespaces.size());
+    Set<String> namespaceNames = namespaces.stream().map(arr -> arr[0].toString()).collect(Collectors.toSet());
+    Assert.assertEquals("Should have default and db namespaces", ImmutableSet.of("default", "db"), namespaceNames);
```
rdblue, line 139: What changed this behavior?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/SparkTestBaseWithCatalog.java
sumeetgajjar: Hi Ryan - thank you for your comment.

> Why is this needed? IF NOT EXISTS should check existence first, right?

Yes - in a way it does perform the check.

When "CREATE NAMESPACE IF NOT EXISTS default" SQL command is executed in Spark, Spark invokes hive.createDatabase command.
https://github.com/apache/spark/blob/89fdb8a6fb6a669c458891b3abeba236e64b1e89/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala#L574

Hive client invokes internally invokes HMS API to create the database. If the DB already exists HMS throws AlreadyExistsException. When the ifNotExist flag is set to true, the Hive client simply ignores the exception.
https://github.com/apache/hive/blob/63326ff775206e59547b6b1332e25279e90ef5ee/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L608-L619

The HMS logs this exception to STDERR and for iceberg tests since a standalone HMS is running in the same JVM as that of the test, these logs are part of the info output of the tests.

This generates a lot of noise in the logs and might overshadow an actual exception.

Thus in order to avoid the `AlreadyExistsException` in the info logs, we are performing an explicit check and using the Catalog APIs instead of the Spark SQL to create a namespace.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/sql/TestNamespaceSQL.java
sumeetgajjar: The assertion checks if the namespace that `sql("CREATE NAMESPACE %s", fullNamespace);` creats is in fact created or not.

Now, instead of having only one namespace i.e. "db" which is created using `CREATE NAMESPACE` stmt we have two; one is "db" and the other is "default" that is created in the constructor of the `SparkTestBaseWithCatalog` class during setup.

Assuming the objective of the assertion was to check if the "db" namespace was created or not and even after the changes that this PR introduces the aim is still met, thus there's no change in behavior.

sumeetgajjar:
Hi @pvary @kbendick @rdblue,
Gentle ping to review the new changes.

sumeetgajjar:
Hi @rdblue and @pvary,
A gentle ping to review the latest changes.
If you feel this change is something that we don't need, please let me know as well, I can close this PR :-)

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/sql/TestNamespaceSQL.java
pvary: The question is whether the test aims to test the creation using the `CREATE NAMESPACE IF NOT EXISTS default` sql too, or the only goal was to test if the NS is created by using the catalog API.

Someone with more knowledge about the Spark test could be good here.

rdblue:
@sumeetgajjar, I'm fine with these changes, although I think we still need to find out why changing how a database gets created caused it to exist for Hadoop. Feel free to rebase and we can work on this.

sumeetgajjar:
Thanks for the reviews @rdblue @pvary, really appreciate it.

There was an internal redistribution of tasks and a change of ownership of some GitHub PRs due to internal changes.
I am closing this PR and moving forward @wypoon or @sririshindra would be going leading the changes suggested in this PR.
They would pull the public branch associated with this PR, add their commits on top of it and create a new PR to ease the logistics of the transition.

Thanks.

PR #4943, opened 2022-06-02
Flink: add an option to set monitoring snapshot number

chenjunjiedada:
This adds an option to control how many snapshots to monitor at once when using iceberg table as a Flink source. 

Currently, the monitor operator generates file splits from the last consumed snapshot to the latest snapshot, which may lead to backpressure when the consumer lag behind as follow image shows. We can reduce the checkpoint lock scope (https://github.com/apache/iceberg/pull/4911) or increase the network buffer to mitigate the situation while the problem still cannot be completely avoided since the number of the splits is unknown, especially when starting a consumer for the first time.
image

With the option, the user can tune the monitoring flow according to backpressure and busy metrics. 
image

### DIFF
--- a/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
+++ b/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
@@ -72,6 +72,9 @@ class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Integer> MONITOR_SNAPSHOT_NUMBER =
+      ConfigOptions.key("monitor-snapshot-number").intType().defaultValue(Integer.MAX_VALUE);
+
   private final boolean caseSensitive;
   private final boolean exposeLocality;
   private final Long snapshotId;
@@ -90,12 +93,13 @@ class ScanContext implements Serializable {
   private final long limit;
   private final boolean includeColumnStats;
   private final Integer planParallelism;
+  private final int monitorSnapshotNumber;
 
   private ScanContext(boolean caseSensitive, Long snapshotId, Long startSnapshotId, Long endSnapshotId,
                       Long asOfTimestamp, Long splitSize, Integer splitLookback, Long splitOpenFileCost,
                       boolean isStreaming, Duration monitorInterval, String nameMapping, Schema schema,
                       List<Expression> filters, long limit, boolean includeColumnStats, boolean exposeLocality,
-                      Integer planParallelism) {
+                      Integer planParallelism, int monitorSnapshotNumber) {
     this.caseSensitive = caseSensitive;
     this.snapshotId = snapshotId;
     this.startSnapshotId = startSnapshotId;
@@ -114,6 +118,7 @@ class ScanContext implements Serializable {
     this.includeColumnStats = includeColumnStats;
     this.exposeLocality = exposeLocality;
     this.planParallelism = planParallelism;
+    this.monitorSnapshotNumber = monitorSnapshotNumber;
   }
 
   boolean caseSensitive() {
@@ -184,6 +189,10 @@ class ScanContext implements Serializable {
     return planParallelism;
   }
 
+  int monitorSnapshotNumber() {
+    return monitorSnapshotNumber;
+  }
+
   ScanContext copyWithAppendsBetween(long newStartSnapshotId, long newEndSnapshotId) {
     return ScanContext.builder()
         .caseSensitive(caseSensitive)
@@ -203,6 +212,7 @@ class ScanContext implements Serializable {
         .includeColumnStats(includeColumnStats)
         .exposeLocality(exposeLocality)
         .planParallelism(planParallelism)
+        .monitorSnapshotNumber(monitorSnapshotNumber)
         .build();
   }
 
@@ -225,6 +235,7 @@ class ScanContext implements Serializable {
         .includeColumnStats(includeColumnStats)
         .exposeLocality(exposeLocality)
         .planParallelism(planParallelism)
+        .monitorSnapshotNumber(monitorSnapshotNumber)
         .build();
   }
 
@@ -250,6 +261,7 @@ class ScanContext implements Serializable {
     private boolean includeColumnStats = INCLUDE_COLUMN_STATS.defaultValue();
     private boolean exposeLocality;
     private Integer planParallelism = FlinkConfigOptions.TABLE_EXEC_ICEBERG_WORKER_POOL_SIZE.defaultValue();
+    private int monitorSnapshotNumber = MONITOR_SNAPSHOT_NUMBER.defaultValue();
 
     private Builder() {
     }
@@ -339,6 +351,11 @@ class ScanContext implements Serializable {
       return this;
     }
 
+    Builder monitorSnapshotNumber(int newMonitorSnapshotNumber) {
+      this.monitorSnapshotNumber = newMonitorSnapshotNumber;
+      return this;
+    }
+
     Builder fromProperties(Map<String, String> properties) {
       Configuration config = new Configuration();
       properties.forEach(config::setString);
@@ -354,14 +371,15 @@ class ScanContext implements Serializable {
           .streaming(config.get(STREAMING))
           .monitorInterval(config.get(MONITOR_INTERVAL))
           .nameMapping(properties.get(DEFAULT_NAME_MAPPING))
-          .includeColumnStats(config.get(INCLUDE_COLUMN_STATS));
+          .includeColumnStats(config.get(INCLUDE_COLUMN_STATS))
+          .monitorSnapshotNumber(config.get(MONITOR_SNAPSHOT_NUMBER));
     }
 
     public ScanContext build() {
       return new ScanContext(caseSensitive, snapshotId, startSnapshotId,
           endSnapshotId, asOfTimestamp, splitSize, splitLookback,
           splitOpenFileCost, isStreaming, monitorInterval, nameMapping, projectedSchema,
-          filters, limit, includeColumnStats, exposeLocality, planParallelism);
+          filters, limit, includeColumnStats, exposeLocality, planParallelism, monitorSnapshotNumber);
     }
   }
 }
--- a/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
+++ b/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.flink.source;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.List;
 import java.util.concurrent.ExecutorService;
 import org.apache.flink.api.common.functions.RuntimeContext;
 import org.apache.flink.api.common.state.ListState;
@@ -134,11 +135,17 @@ public class StreamingMonitorFunction extends RichSourceFunction<FlinkInputSplit
   public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
     this.sourceContext = ctx;
     while (isRunning) {
+      LOG.info("Start polling snapshots from snapshot id: {}, monitor snapshot number {}", lastSnapshotId,
+          scanContext.monitorSnapshotNumber());
+      long start = System.currentTimeMillis();
+      long startSnapshotId = lastSnapshotId;
       synchronized (sourceContext.getCheckpointLock()) {
         if (isRunning) {
           monitorAndForwardSplits();
         }
       }
+      LOG.info("Forwarded splits from {}(exclusive) to {}(inclusive), time elapsed {}ms",
+          startSnapshotId, lastSnapshotId, System.currentTimeMillis() - start);
       Thread.sleep(scanContext.monitorInterval().toMillis());
     }
   }
@@ -155,6 +162,12 @@ public class StreamingMonitorFunction extends RichSourceFunction<FlinkInputSplit
       if (lastSnapshotId == INIT_LAST_SNAPSHOT_ID) {
         newScanContext = scanContext.copyWithSnapshotId(snapshotId);
       } else {
+        List<Long> snapshotIds = SnapshotUtil.snapshotIdsBetween(table, lastSnapshotId, snapshot.snapshotId());
+        if (snapshotIds.size() < scanContext.monitorSnapshotNumber()) {
+          snapshotId = snapshot.snapshotId();
+        } else {
+          snapshotId = snapshotIds.get(snapshotIds.size() - scanContext.monitorSnapshotNumber());
+        }
         newScanContext = scanContext.copyWithAppendsBetween(lastSnapshotId, snapshotId);
       }
 
--- a/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
+++ b/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
@@ -208,6 +208,37 @@ public class TestStreamingMonitorFunction extends TableTestBase {
     }
   }
 
+  @Test
+  public void testConsumeWithDifferentMonitorSnapshotNumbers() throws Exception {
+    List<List<Record>> recordsList = generateRecordsAndCommitTxn(10);
+
+    for (int monitorNumber = 1; monitorNumber < 11; monitorNumber = monitorNumber + 1) {
+      ScanContext scanContext = ScanContext.builder()
+          .monitorInterval(Duration.ofMillis(100))
+          .monitorSnapshotNumber(monitorNumber)
+          .build();
+
+      StreamingMonitorFunction function = createFunction(scanContext);
+      try (AbstractStreamOperatorTestHarness<FlinkInputSplit> harness = createHarness(function)) {
+        harness.setup();
+        harness.open();
+
+        CountDownLatch latch = new CountDownLatch(1);
+        TestSourceContext sourceContext = new TestSourceContext(latch);
+        runSourceFunctionInTask(sourceContext, function);
+
+        Assert.assertTrue("Should have expected elements.", latch.await(WAIT_TIME_MILLIS, TimeUnit.MILLISECONDS));
+        Thread.sleep(1000L);
+
+        // Stop the stream task.
+        function.close();
+
+        Assert.assertEquals("Should produce the expected splits", 1, sourceContext.splits.size());
+        TestHelpers.assertRecords(sourceContext.toRows(), Lists.newArrayList(Iterables.concat(recordsList)), SCHEMA);
+      }
+    }
+  }
+
   private List<List<Record>> generateRecordsAndCommitTxn(int commitTimes) throws IOException {
     List<List<Record>> expectedRecords = Lists.newArrayList();
     for (int i = 0; i < commitTimes; i++) {
--- a/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSource.java
+++ b/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSource.java
@@ -184,6 +184,11 @@ public class FlinkSource {
       return this;
     }
 
+    public Builder monitorSnapshotNumber(int newMonitorSnapshotNumber) {
+      contextBuilder.monitorSnapshotNumber(newMonitorSnapshotNumber);
+      return this;
+    }
+
     public Builder flinkConf(ReadableConfig config) {
       this.readableConfig = config;
       return this;


flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -155,6 +162,12 @@ private void monitorAndForwardSplits() {
       if (lastSnapshotId == INIT_LAST_SNAPSHOT_ID) {
         newScanContext = scanContext.copyWithSnapshotId(snapshotId);
       } else {
+        List<Long> snapshotIds = SnapshotUtil.snapshotIdsBetween(table, lastSnapshotId, snapshot.snapshotId());
+        if (snapshotIds.size() < scanContext.monitorSnapshotNumber()) {
+          snapshotId = snapshot.snapshotId();
+        } else {
+          snapshotId = snapshotIds.get(snapshotIds.size() - scanContext.monitorSnapshotNumber());
+        }
```
kbendick, line 170: Can you elaborate on this logic here / walk me through an example case where `snapshotId` needs to be determined because it's equal to (or possibly greater than?) the `monitorSnapshotNumber`?

kbendick (commented):
Thanks for this contribution @chenjunjiedada! I know that especially the case of streaming a table from the start is difficult and can cause high memory pressure.

I would also love to get @stevenzwu's input here.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -72,6 +72,9 @@ class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Integer> MONITOR_SNAPSHOT_NUMBER =
+      ConfigOptions.key("monitor-snapshot-number").intType().defaultValue(Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
chenjunjiedada: +1 to use "max" word. How about `max-snapshots-per-monitor-interval`?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -339,6 +351,11 @@ Builder planParallelism(Integer parallelism) {
       return this;
     }
 
+    Builder monitorSnapshotNumber(int newMonitorSnapshotNumber) {
+      this.monitorSnapshotNumber = newMonitorSnapshotNumber;
```
hililiwei, line 355: Should we add a check here? It should be greater than 0, right?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
hililiwei: Instead, I suggest using `lastSnapshotId` and `monitorSnapshotNumber` directly to get to the destination.

Specifically, there should be two ways:

1.  use `lastSnapshotId` and `monitorSnapshotNumber` to get `snapshotId`, instead of getting `currentSnapshotId` and comparing,
2.  Use `lastSnapshotId` and `monitorSnapshotNumber` directly to get the snapshot list

The above may require modifications to the core module, but I think it will make the logic more intuitive.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -134,11 +135,17 @@ public void snapshotState(FunctionSnapshotContext context) throws Exception {
   public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
     this.sourceContext = ctx;
     while (isRunning) {
+      LOG.info("Start polling snapshots from snapshot id: {}, monitor snapshot number {}", lastSnapshotId,
```
hililiwei, line 138: Use `debug` to prevent too many logs?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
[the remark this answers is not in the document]
kbendick: Yeah I’m good with that. That’s much more clear to me what that is, especially as we have a monitor-interval in the configs already.

So users will go to look up what that is  - which we should ensure is documented as a follow up.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +208,37 @@ public void testCheckpointRestore() throws Exception {
...
 
+  @Test
+  public void testConsumeWithDifferentMonitorSnapshotNumbers() throws Exception {
+    List<List<Record>> recordsList = generateRecordsAndCommitTxn(10);
+
+    for (int monitorNumber = 1; monitorNumber < 11; monitorNumber = monitorNumber + 1) {
+      ScanContext scanContext = ScanContext.builder()
+          .monitorInterval(Duration.ofMillis(100))
+          .monitorSnapshotNumber(monitorNumber)
+          .build();
```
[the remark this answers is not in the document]
kbendick: Yes definitely. A Precondition on an invalid configuration is _much_ preferred. 👍

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
kbendick: I’m thinking through these situations now.

But users would _never enter this block_ if they don’t opt into the new behavior, is that correct?

If so, can we add a conditional case that this block won’t be entered unless the user have a non `INT_MAX` value (eg we make a Precondition and just make a sentence that starts with `[bug] this shouldn’t happen <because>`). We have one or two other places that use the same `[bug]` syntax and this new logic change would ideally not apply to users that keep the default behavior.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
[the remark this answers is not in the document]
kbendick: It’s always better in my opinion to throw on an invalid configuration vs try to adjust for the user’s behavior.

We should check that it’s non-negative in. Precondition check.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
kbendick: I do like the idea of using the snapshot ID though generally speaking (in this case I need to review).

However, It’s possible to turn off snapshot ID inheritance. So we’d need to consider that.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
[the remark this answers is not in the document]
kbendick: We should also be testing that we only have at most `max-snapshots-per-monitor-interval` amount of snapshots processed at the end of the large loop.

That should be true, right?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
kbendick: Also, can we make this its own method?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
kbendick: +1. A precondition check. And INT_MAX can be the value that disables this behavior.

Or a negative value (eg -1). A negative value is probably more in-line with what we typically do and more cross-language friendly (as Iceberg table format is a specification first and foremost… it should be able to be rewritten in a language that doesn’t have JVM INT_MAX).

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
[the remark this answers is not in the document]
kbendick: If you could please rename the code instances from `monitorNumber` that would be great. That name is much more descriptive.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
kbendick: Yeah this could get very excessive very quickly.

This should be a debug log as users can opt in via their logging configuration if need be.

kbendick:
It might be a good idea (in a follow up) to make a metric that monitors how many snapshots are processed per monitor interval.

Also, since this grabs the checkpoint lock, will it possibly not be an even multiple when the checkpoint happens? If so, we should add a metric that tracks that as well (again, in a follow up).

Does that interest you @hililiwei? Or @chenjunjiedada? If so, feel free to make the ticket and we’ll deal with that as soon as we can 🙂

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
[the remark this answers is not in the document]
chenjunjiedada: Instead of testing at the outside loop, I compare the exact split number to the planning result inside the loop. Does that make sense to you?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
chenjunjiedada: Since we already consider the inheritance in state initialization, do we need another check?

@hililiwei I agree computing using `lastComsumedSnapshotId` and `maxSnapshotsPerMonitorInterval` is more accurate. But I haven't found direct methods or utils to compute that, maybe need more utils in `SnapshotUtil` as you said. Can we have such cases in which the last consumed snapshot is not an ancestor of the latest snapshot?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -134,15 +140,33 @@ public void snapshotState(FunctionSnapshotContext context) throws Exception {
...
 
+  private long maxReachableSnapshotId(long lastConsumedSnapshotId, long latestSnapshotId,
+                                      int maxSnapshotsPerMonitorInterval) {
+    List<Long> snapshotIds = SnapshotUtil.snapshotIdsBetween(table, lastConsumedSnapshotId, latestSnapshotId);
+
+    if (snapshotIds.size() <= maxSnapshotsPerMonitorInterval) {
+      return latestSnapshotId;
+    } else {
+      // This doesn't consider snapshot inheritance since it is already checked in state initialization.
+      return snapshotIds.get(snapshotIds.size() - maxSnapshotsPerMonitorInterval);
```
stevenzwu, line 166: This logic seems incorrect to me. `SnapshotUtil.snapshotIdsBetween` returns the snapshot ids in the reverse order (most recent snapshot first). I think we need to use the reversed list.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -134,15 +140,33 @@ public void snapshotState(FunctionSnapshotContext context) throws Exception {
...
+      LOG.debug("Start polling snapshots from snapshot id: {}, max-snapshots-per-monitor-interval: {}", lastSnapshotId,
+          scanContext.maxSnapshotsPerMonitorInterval());
+      long start = System.currentTimeMillis();
+      long startSnapshotId = lastSnapshotId;
       synchronized (sourceContext.getCheckpointLock()) {
         if (isRunning) {
           monitorAndForwardSplits();
         }
       }
+      LOG.debug("Forwarded splits from {}(exclusive) to {}(inclusive), time elapsed {}ms",
```
stevenzwu, line 152: I think it is better to log the durations for the two steps separately if we want to have better understanding of the bottleneck

stevenzwu: The logging seems wrong to me. both `startSnapshotId` and `lastSnapshotId` are pointing to the same snapshotId. Moving the logging inside `maxReachableSnapshotId`, we can correctly apply the endSnapshotId calculated from `maxReachableSnapshotId`.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkConfigOptions.java
```diff
@@ -48,6 +48,8 @@
  */
 public class FlinkConfigOptions {
 
+  public static final int MAX_SNAPSHOTS_PER_MONITOR_INTERVAL_DEFAULT = -1;
```
stevenzwu, line 51: we can default it to Integer.MAX_VALUE. then there is no need to check default value.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
stevenzwu: I got it now. it is actually correct. We should improve the comment. Because it is a revert list (list size - maxSnapshotsPerMonitorInterval) will actually point to the snapshot that results in (fromSnapshotId, toSnapshotId) with size of maxSnapshotsPerMonitorInterval

stevenzwu:
In general, I think this is a right direction. It is not mutually exclusive with the PR #4911 (reduce the checkpoint lock scope). We should have both. This is focused on making the plan smaller and faster. PR #4911 can avoid holding the lock beyond what is actually necessary.

For the new FLIP-27 source, I have been thinking about something very similar. There is no point of eagerly discover all splits/snapshots if the Flink job is falling behind too much. We need to throttle the split discovery. In addition to limit the number of snapshots per discovery cycle, I am also thinking about that we should pause/skip the split discovery, if the number of pending splits is over a certain threshold. It is like a backpressure mechanism. This can help control the memory footprint. This won't be in the MVP version of FLIP-27 source. We can follow up on the optimization after MVP version is merged.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +210,67 @@ public void testCheckpointRestore() throws Exception {
...
+    );
+
+    for (int maxSnapshotsNum = 1; maxSnapshotsNum < 15; maxSnapshotsNum = maxSnapshotsNum + 1) {
+      ScanContext scanContext = ScanContext.builder()
+          .monitorInterval(Duration.ofMillis(100))
+          .splitSize(1000L)
+          .maxSnapshotsPerMonitorInterval(maxSnapshotsNum)
+          .build();
+
+      FlinkInputSplit[] expectedSplits = FlinkSplitPlanner
```
stevenzwu, line 250: I think we should directly define the number of expected splits and avoid using `FlinkSplitPlanner`. if `FlinkSplitPlanner` is not honoring  `maxSnapshotsPerMonitorInterval`, the assertion later will still be correct

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -72,6 +72,10 @@ class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Integer> MAX_SNAPSHOTS_PER_MONITOR_INTERVAL =
+      ConfigOptions.key("max-snapshot-per-monitor-interval").intType()
```
stevenzwu, line 76: it will be more clear if the config is defined as `max-snapshot-count-per-monitor-interval` or `max-snapshot-count-per-incremental-scan`

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkConfigOptions.java
chenjunjiedada: @stevenzwu , Kyle suggested using -1 here: https://github.com/apache/iceberg/pull/4943#discussion_r889652532. I think that makes sense to me as well.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
chenjunjiedada: `lastSnapshotId` is updated in `monitorAndForwardSplits`.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
chenjunjiedada: `FlinkSplitPlanner` here is using the same `scanContext` as `StreamingMonitorFunction`, will it still produce different splits?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkConfigOptions.java
stevenzwu: I see the `ScanSummary` also use `Integer.MAX_VALUE`. -1 default seems to be used mostly for invalid ids (like snapshotId, fieldId etc.)
```
    private int limit = Integer.MAX_VALUE;
```

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
stevenzwu: yes. I missed `lastSnapshotId`. It also means that the code is difficult to read. we are relying on the side effect of the `monitorAndForwardSplits` method. It is more clear to avoid such side effect. Plus it is better to measure the latency separately: planning vs emitting.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
stevenzwu: If `FlinkSplitPlanner` didn't honor the `maxSnapshotsPerMonitorInterval` option correctly, this unit test won't be able to detect it. Both the expected and actual splits use the same planner and will generate the same planning result

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
hililiwei: I wonder if that would be the case after supporting `Branch and Tag` feature?

hililiwei:
> It might be a good idea (in a follow up) to make a metric that monitors how many snapshots are processed per monitor interval.
> 
> Also, since this grabs the checkpoint lock, will it possibly not be an even multiple when the checkpoint happens? If so, we should add a metric that tracks that as well (again, in a follow up).
> 
> Does that interest you @hililiwei? Or @chenjunjiedada? If so, feel free to make the ticket and we’ll deal with that as soon as we can 🙂

Sounds interesting. if you don't mind, I'll try to follow up.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkConfigOptions.java
chenjunjiedada: @kbendick What do you think for this?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
chenjunjiedada: I'm not sure whether it can route to different branches without changing table states or properties. Like git, if we don't explicitly `checkout` a branch we should be able to traverse the history, right?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
stevenzwu: @chenjunjiedada I prefer we address the above comment regarding logging

chenjunjiedada: Will do.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +211,94 @@ public void testCheckpointRestore() throws Exception {
...
+        // Take lock to prevent monitoring loop
+        synchronized (sourceContext.getCheckpointLock()) {
+          if (maxSnapshotsNum < 10) {
+            // it produces one split for one snapshot.
+            Assert.assertEquals("Should produce same splits as max-snapshot-count-per-monitor-interval",
+                maxSnapshotsNum, sourceContext.splits.size());
+          }
+        }
+
+        Assert.assertTrue("Should have expected elements.",
```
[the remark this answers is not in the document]
chenjunjiedada: @stevenzwu I removed lock/wait/sleep logic and kept the loop from 1 to 15 since the total execution time of the unit test is just a few seconds.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +211,94 @@ public void testCheckpointRestore() throws Exception {
     }
   }
 
+  @Test
+  public void testConsumeWithMaxSnapshotCountPerMonitorInterval() throws Exception {
+    List<List<Record>> recordsList = generateRecordsAndCommitTxn(10);
+
+    final ScanContext scanContext1 = ScanContext.builder()
```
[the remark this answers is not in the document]
chenjunjiedada: Done.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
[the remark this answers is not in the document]
stevenzwu: a few seconds is long for unit test. We just need to test the 3 scenarios of the `maxSnapshotsNum`: less, equal, greater.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -154,14 +179,28 @@ private void monitorAndForwardSplits() {
       ScanContext newScanContext;
       if (lastSnapshotId == INIT_LAST_SNAPSHOT_ID) {
         newScanContext = scanContext.copyWithSnapshotId(snapshotId);
+        LOG.debug("Start generating splits for {}", snapshotId);
       } else {
+        if (scanContext.maxSnapshotCountPerMonitorInterval() ==
```
stevenzwu, line 184: Related to an earlier comment. if the default value is `Integer.MAX_VALUE`, we won't need the if-else

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -154,14 +179,28 @@ private void monitorAndForwardSplits() {
...
+            FlinkConfigOptions.MAX_SNAPSHOT_COUNT_PER_MONITOR_INTERVAL_DEFAULT) {
+          snapshotId = snapshot.snapshotId();
+        } else {
+          snapshotId = maxReachableSnapshotId(lastSnapshotId, snapshotId,
+              scanContext.maxSnapshotCountPerMonitorInterval());
+        }
         newScanContext = scanContext.copyWithAppendsBetween(lastSnapshotId, snapshotId);
+        LOG.debug("Start generating splits from {}(exclusive) to {}(inclusive),", lastSnapshotId, snapshotId);
```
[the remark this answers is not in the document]
chenjunjiedada: Do you mean it is ok to show `from -1 to $snapshotId`?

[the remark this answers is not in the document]
chenjunjiedada: Done.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
chenjunjiedada: Updated.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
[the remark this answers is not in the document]
chenjunjiedada: Updated.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +211,82 @@ public void testCheckpointRestore() throws Exception {
...
+        .build();
+
+    FlinkInputSplit[] expectedSplits = FlinkSplitPlanner
+        .planInputSplits(table, scanContext3, ThreadPools.getWorkerPool());
+
+    Assert.assertEquals("should produce 9 splits", 9, expectedSplits.length);
+
+    // This covers three cases that maxSnapshotCount is less than, equal or greater than the total splits number.
+    List<Integer> maxSnapshotCounts = ImmutableList.of(1, 9, 15);
+    for (int maxSnapshotCount : maxSnapshotCounts) {
```
stevenzwu, line 264: nit: directly put `ImmutableList.of(1, 9, 15)` or `Arrays.asList(....)` here

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -143,7 +147,25 @@ public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
...
+      return snapshotIds.get(snapshotIds.size() - maxSnapshotCountPerMonitorInterval);
+    }
+  }
+
+  @VisibleForTesting
+  void sourceContext(SourceContext<FlinkInputSplit> ctx) {
+    this.sourceContext = ctx;
+  }
+
+  void monitorAndForwardSplits() {
```
stevenzwu, line 168: nit: add `@VisibleForTesting` annotation here too

stevenzwu (approved):
LGTM except for a couple of nit comments

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +211,81 @@ public void testCheckpointRestore() throws Exception {
     }
   }
 
+  @Test
+  public void testInvalidMaxSnapshotCountPerMonitorInterval() {
+    final ScanContext scanContext1 = ScanContext.builder()
```
kbendick, line 216: Nit: We normally don't use `final` for variables inside of methods. Is this usage of `final` necessary?

kbendick (approved):
A few non-blocking nits but overall this looks good to me. Thanks @chenjunjiedada!

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
```diff
@@ -208,6 +211,81 @@ public void testCheckpointRestore() throws Exception {
...
+  }
+
+  @Test
+  public void testConsumeWithMaxSnapshotCountPerMonitorInterval() throws Exception {
+    generateRecordsAndCommitTxn(10);
+
+    // Use the oldest snapshot as starting to avoid the initial case.
+    long oldestSnapshotId = SnapshotUtil.oldestAncestor(table).snapshotId();
+
+    ScanContext scanContext3 = ScanContext.builder()
```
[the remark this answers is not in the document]
chenjunjiedada: Not necessary actually, updated.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamingMonitorFunction.java
chenjunjiedada: Not necessary, removed.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -143,7 +147,26 @@ public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
     }
   }
 
-  private void monitorAndForwardSplits() {
+  private long maxReachableSnapshotId(long lastConsumedSnapshotId, long latestSnapshotId,
+                                      int maxSnapshotCountPerMonitorInterval) {
+    // This doesn't consider snapshot inheritance since it is already checked in state initialization.
```
rdblue, line 152: What do you mean by "snapshot inheritance" here?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -143,7 +147,26 @@ public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
     }
   }
 
-  private void monitorAndForwardSplits() {
+  private long maxReachableSnapshotId(long lastConsumedSnapshotId, long latestSnapshotId,
+                                      int maxSnapshotCountPerMonitorInterval) {
```
rdblue, line 151: I think this should be `maxSnapshots`

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -155,13 +178,21 @@ private void monitorAndForwardSplits() {
       if (lastSnapshotId == INIT_LAST_SNAPSHOT_ID) {
         newScanContext = scanContext.copyWithSnapshotId(snapshotId);
       } else {
+        snapshotId = maxReachableSnapshotId(lastSnapshotId, snapshotId,
+            scanContext.maxSnapshotCountPerMonitorInterval());
         newScanContext = scanContext.copyWithAppendsBetween(lastSnapshotId, snapshotId);
       }
 
+      LOG.debug("Start discovering splits from {}(exclusive) to {}(inclusive),", lastSnapshotId, snapshotId);
```
rdblue, line 186: This is missing spaces between the snapshot IDs and the inclusive or exclusive clarification.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -82,6 +82,9 @@ public class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Integer> MAX_SNAPSHOT_COUNT_PER_MONITOR_INTERVAL =
+      ConfigOptions.key("max-snapshot-count-per-monitor-interval").intType().defaultValue(Integer.MAX_VALUE);
```
rdblue, line 86: I don't think it is very clear what this is setting from the name. Not many people know Flink internals well enough to understand what the "monitor interval" is. Is there a simpler name?

What about "max-planning-group-size" or "snapshot-group-limit"?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
chenjunjiedada: The latest table snapshot id might not be the ancestor of the last consumed one.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
chenjunjiedada: `max-planning-group-size` lack information about what things are in the group, `group` can be used to replace monitor interval. How about `max-snapshot-count-per-group` or `max-snapshots-per-group`?

chenjunjiedada: `max-planning-group-size` lacks information about the group items, `snapshot-group-limit` looks better to me. "group" is more concise than "per-monitor-interval".

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
chenjunjiedada: Changed to snapshotLimit.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
stevenzwu: I agree with @chenjunjiedada that `group` is vague. what about `max-planning-snapshot-count`?

chenjunjiedada: I meant `group` is better than `per-monitor-interval`. I'm OK with both since I think we definitely need a doc to describe what it is and how it impacts the planning and the backpressure.

stevenzwu: I found `snapshot-group-limit` less intuitive. what does the `group` mean in this context. `monitor-interval` is an existing config that user is already familiar.

I do agree that we should make the config name clear. Just brainstorming more names. what about `max-snapshot-count-per-incremental-scan` or `max-snapshot-count-per-planning`?

chenjunjiedada: Both `max-planning-snapshot-count` and `max-snapshot-count-per-incremental-scan` are OK with me. @kbendick @rdblue what do you think?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -82,6 +82,9 @@ public class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Integer> SNAPSHOT_GROUP_LIMIT =
+      ConfigOptions.key("snapshot-group-limit").intType().defaultValue(Integer.MAX_VALUE);
```
rdblue, line 86: I talked with @stevenzwu about this and the best name we could come up with was `max-planning-snapshot-count`. What do you think, @chenjunjiedada? Could you rename this?

kbendick: +1 to this name. Config keys should ideally be concise as well as being as short as possible and `max-planning-snapshot-count` achieves that.

The documentation can potentially use the language `Maximum number of snapshots to consume and plan per group in each iteration of an incremental scan` or something similar (might need to work on that language too but the language from the other ideas can be used in the docs possibly).

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkConfigOptions.java
kbendick: I’m ok with INT_MAX if that’s what’s used elsewhere in `ScanSummary`. I suggested -1 as we use -1 to disable cache expiration in the caching catalog and because it’s easier to use in things like Python where INT_MAX is less commonly used. -1 is also easier to use as in-line SQL option imo.

But if INT_MAX is already used in the `ScanSummary`, I suggest we keep that consistent.

We can then later on consider using -1 in both places within Flink (or more places). But consistency is better imo.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
chenjunjiedada: Done.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -139,7 +143,24 @@ public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
     }
   }
 
-  private void monitorAndForwardSplits() {
+  private long maxReachableSnapshotId(long lastConsumedSnapshotId, long latestSnapshotId, int maxSnapshotCount) {
```
stevenzwu, line 146: nit: maybe call the args `fromSnapshotIdExclusive` and `toSnapshotIdInclusive` to be consistent with recent incremental API naming

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -139,7 +143,24 @@ public void run(SourceContext<FlinkInputSplit> ctx) throws Exception {
...
+    List<Long> snapshotIds = SnapshotUtil.snapshotIdsBetween(table, lastConsumedSnapshotId, latestSnapshotId);
+
```
stevenzwu, line 148: nit: we add empty line after control block (not before)

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
stevenzwu: actually maybe call this method as `toSnapshotIdInclusive(long lastConsumedSnapshotId, long currentSnapshotId, int maxPlanningSnapshotCount)`

chenjunjiedada:
@rdblue @kbendick, Could you please take another look?

rdblue (approved):

PR #4944, opened 2022-06-02
Add Puffin, indices and stats file format specification

findepi:
Add a specification for Puffin format, a container file format to store
indices and stats for Iceberg tables.

### DIFF
--- a/format/puffin-spec.md
+++ b/format/puffin-spec.md
@@ -1,0 +1,144 @@
+---
+url: puffin-spec
+toc: false
+---
+<!--
+ - Licensed to the Apache Software Foundation (ASF) under one or more
+ - contributor license agreements.  See the NOTICE file distributed with
+ - this work for additional information regarding copyright ownership.
+ - The ASF licenses this file to You under the Apache License, Version 2.0
+ - (the "License"); you may not use this file except in compliance with
+ - the License.  You may obtain a copy of the License at
+ -
+ -   http://www.apache.org/licenses/LICENSE-2.0
+ -
+ - Unless required by applicable law or agreed to in writing, software
+ - distributed under the License is distributed on an "AS IS" BASIS,
+ - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ - See the License for the specific language governing permissions and
+ - limitations under the License.
+ -->
+
+# Puffin file format
+
+This is a specification for Puffin, a file format designed to store
+information such as indexes and statistics about data managed in an
+Iceberg table that cannot be stored directly within the Iceberg manifest. A
+Puffin file contains arbitrary pieces of information (here called "blobs"),
+along with metadata necessary to interpret them. The blobs supported by Iceberg
+are documented at [Blob types](#blob-types).
+
+## Format specification
+
+A file conforming to the Puffin file format specification should have the structure
+as described below.
+
+### Versions
+
+Currently, there is a single version of the Puffin file format, described below.
+
+### File structure
+
+The Puffin file has the following structure
+
+```
+Magic Blob₁ Blob₂ ... Blobₙ Footer
+```
+
+where
+
+- `Magic` is four bytes 0x50, 0x46, 0x41, 0x31 (short for: Puffin _Fratercula
+  arctica_, version 1),
+- `Blobᵢ` is i-th blob contained in the file, to be interpreted by application
+  according to the footer,
+- `Footer` is defined below.
+
+### Footer structure
+
+Footer has the following structure
+
+```
+Magic FooterPayload FooterPayloadSize Flags Magic
+```
+
+where
+
+- `Magic`: four bytes, same as at the beginning of the file
+- `FooterPayload`: optionally compressed, UTF-8 encoded JSON payload describing the
+  blobs in the file, with the structure described below
+- `FooterPayloadSize`: a length in bytes of the `FooterPayload` (after compression,
+  if compressed), stored as 4 byte integer
+- `Flags`: 4 bytes for boolean flags
+  - byte 0 (first)
+    - bit 0 (lowest bit): whether `FooterPayload` is compressed
+    - all other bits are reserved for future use and should be set to 0 on write
+  - all other bytes are reserved for future use and should be set to 0 on write
+
+A 4 byte integer is always signed, in a two's complement representation, stored
+little-endian.
+
+### Footer Payload
+
+Footer payload bytes is either uncompressed or LZ4-compressed (as a single
+[LZ4 compression frame](https://github.com/lz4/lz4/blob/77d1b93f72628af7bbde0243b4bba9205c3138d9/doc/lz4_Frame_format.md)
+with content size present), UTF-8 encoded JSON payload representing a single
+`FileMetadata` object.
+
+#### FileMetadata
+
+`FileMetadata` has the following fields
+
+
+| Field Name | Field Type                              | Required | Description |
+| ---------- | --------------------------------------- | -------- | ----------- |
+| blobs      | list of BlobMetadata objects            | yes      |
+| properties | JSON object with string property values | no       | storage for arbitrary meta-information, like writer identification/version. See [Common properties](#common-properties) for properties that are recommended to be set by a writer.
+
+#### BlobMetadata
+
+`BlobMetadata` has the following fields
+
+| Field Name        | Field Type                              | Required | Description |
+|-------------------|-----------------------------------------|----------| ----------- |
+| type              | JSON string                             | yes      | See [Blob types](#blob-types)
+| fields            | list of JSON long                       | yes      | List of field IDs the blob was computed for; the order of items is used to compute sketches stored in the blob.
+| offset            | JSON long                               | yes      | The offset in the file where the blob contents start
+| length            | JSON long                               | yes      | The length of the blob stored in the file (after compression, if compressed)
+| compression-codec | JSON string                             | no       | See [Compression codecs](#compression-codecs). If omitted, the data is assumed to be uncompressed.
+| properties        | JSON object with string property values | no       | storage for arbitrary meta-information about the blob
+
+### Blob types
+
+The blobs can be of a type listed below
+
+#### `apache-datasketches-theta-v1` blob type
+
+A serialized form of a "compact" Theta sketch produced by the [Apache
+DataSketches](https://datasketches.apache.org/) library. The sketch is obtained by
+constructing Alpha family sketch with default seed, and feeding it with individual
+distinct values converted to bytes using Iceberg's single-value serialization.
+
+The blob metadata for this blob may include following properties:
+
+- `ndv`: estimate of number of distinct values, derived from the sketch.
+
+### Compression codecs
+
+The data can also be uncompressed. If it is compressed the codec should be one of
+codecs listed below. For maximal interoperability, other codecs are not supported.
+
+| Codec name | Description                                                                                                                                                                                     |
+|------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| lz4        | Single [LZ4 compression frame](https://github.com/lz4/lz4/blob/77d1b93f72628af7bbde0243b4bba9205c3138d9/doc/lz4_Frame_format.md), with content size present                                     |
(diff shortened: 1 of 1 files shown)


findepi:
cc @alexjo2144 @RussellSpitzer @jackye1995

findepi:
This was already reviewed at https://github.com/apache/iceberg-docs/pull/69.
Should be good to go, per my understanding.

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+
+`BlobMetadata` has the following fields
+
+| Field Name        | Field Type        | Required | Description |
+|-------------------|-------------------| -------- | ----------- |
+| type              | JSON string       | yes      | See [Blob types](#blob-types)
+| fields            | list of JSON long | yes      | List of field IDs the blob was computed for; the order of items is used to compute sketches stored in the blob.
+| offset            | JSON long         | yes      | The offset in the file where the blob contents start
+| length            | JSON long         | yes      | The length of the blob stored in the file
+| compression-codec | JSON string       | no       | See [Compression codecs](#compression-codecs). If omitted, the data is assumed to be uncompressed.
```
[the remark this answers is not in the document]
karuppayya: nevermind, Saw the comment from @rdblue regarding consistency in usage, in other PR

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+Magic FooterPayload FooterPayloadSize Flags Magic
+```
+
+where
+
+- `Magic`: four bytes, same as at the beginning of the file.
+- `FooterPayload`: optionally compressed, UTF-8 encoded JSON payload describing the
+  blobs in the file, with the structure described below,
```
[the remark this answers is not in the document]
findepi: thanks, i think it would be best if it was a comma. however, i removed all the EOL punctucation, because it looked weird on the item with sub-items.

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+- `FooterPayloadSize`: a length in bytes of the `FooterPayload` (compressed),
```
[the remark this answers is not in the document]
findepi: indeed this isn't clear, changing to `(after compression, if compressed)`

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+| properties | JSON object with string property values | no       | storage for arbitrary meta-information, like writer identification/version. See [Common properties](#common-properties) for properties that are recommended to be set by a writer.
+
+#### BlobMetadata
+
+`BlobMetadata` has the following fields
+
+| Field Name        | Field Type        | Required | Description |
+|-------------------|-------------------| -------- | ----------- |
+| type              | JSON string       | yes      | See [Blob types](#blob-types)
+| fields            | list of JSON long | yes      | List of field IDs the blob was computed for; the order of items is used to compute sketches stored in the blob.
```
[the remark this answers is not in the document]
findepi: the wording comes from https://github.com/apache/iceberg-docs/pull/69#discussion_r871536918

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+distinct values converted to bytes using Iceberg's single-value serialization.
+
+### Compression codecs
+
+The data can also be uncompressed. If it is compressed the code should be one of
+codecs listed below. For maximal interoperability, other codecs are not supported.
+
+| Codec name | Description                                                                                                                                                                                     |
+|------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| lz4        | Single [LZ4 compression frame](https://github.com/lz4/lz4/blob/77d1b93f72628af7bbde0243b4bba9205c3138d9/doc/lz4_Frame_format.md), with content size present                                     |
```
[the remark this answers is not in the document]
findepi: I think the precise version (commit) is the only way to guarantee the link remains valid.
for example, the .md file could be renamed on the release branch.

findepi:
@karuppayya thanks for your review!

updated

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+# Puffin file format
+
+This is a specification for the Puffin, a file format designed to store
+information such as indexes and statistics about data managed in an
+Iceberg table that cannot be stored directly within the Iceberg manifest. A
+Puffin file contains arbitrary pieces of information (here called "blobs"),
+along with metadata necessary to interpret them. The blobs supported by Iceberg
+are documented at [Blob types](#blob-types).
+
+## Format specification
```
danielcweeks, line 31: @rdblue should we have a qualifier in this spec relating to restrictions for how Puffiin data can be used in context of Iceberg?  Specifically, I feel like we should be explicit that Puffin data must not be required in order to read/interpret data in iceberg format (i.e. it can be used for optimizations and annotating data, but must not be required for correct interpretation of table data).

We don't want Puffin to be a vector to extend the table format for representation.

danielcweeks (approved):
Overall, I'm +1 with a small comment about being more explicit about the context where Puffin can be used.

format/puffin-spec.md
findepi: @danielcweeks   This is a valid concern.
This PR doesn't introduce any relation between Puffin and Iceberg table format, so there is nothing to worry about yet.
This relation between file format and Iceberg spec is being added in #4945. Can you please take a look there and ensure it doesn't leave any unwanted ambiguities?

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+ - Unless required by applicable law or agreed to in writing, software
+ - distributed under the License is distributed on an "AS IS" BASIS,
+ - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ - See the License for the specific language governing permissions and
+ - limitations under the License.
+ -->
+
+# Puffin file format
+
+This is a specification for the Puffin, a file format designed to store
```
rdblue, line 24: Nit: Should be "for Puffin" instead of "for the Puffin" (unless it was "for the Puffin file format")

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+
+`BlobMetadata` has the following fields
+
+| Field Name        | Field Type        | Required | Description |
+|-------------------|-------------------| -------- | ----------- |
+| type              | JSON string       | yes      | See [Blob types](#blob-types)
+| fields            | list of JSON long | yes      | List of field IDs the blob was computed for; the order of items is used to compute sketches stored in the blob.
+| offset            | JSON long         | yes      | The offset in the file where the blob contents start
+| length            | JSON long         | yes      | The length of the blob stored in the file
+| compression-codec | JSON string       | no       | See [Compression codecs](#compression-codecs). If omitted, the data is assumed to be uncompressed.
```
rdblue, line 107: What about summary metadata, like the NDV result from the theta sketch?

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+### Blob types
+
+The blobs can be of a type listed below
+
+#### `apache-datasketches-theta-v1` blob type
+
+A serialized form of a "compact" Theta sketch produced by the [Apache
+DataSketches](https://datasketches.apache.org/) library. The sketch is obtained by
+constructing Alpha family sketch with default seed, and feeding it with individual
+distinct values converted to bytes using Iceberg's single-value serialization.
```
rdblue, line 118: What about documenting the extra blob metadata that this will produce? It would be nice to have that documented for the blob type.

format/puffin-spec.md
```diff
@@ -0,0 +1,139 @@
...
+#### BlobMetadata
+
+`BlobMetadata` has the following fields
+
+| Field Name        | Field Type        | Required | Description |
+|-------------------|-------------------| -------- | ----------- |
+| type              | JSON string       | yes      | See [Blob types](#blob-types)
+| fields            | list of JSON long | yes      | List of field IDs the blob was computed for; the order of items is used to compute sketches stored in the blob.
+| offset            | JSON long         | yes      | The offset in the file where the blob contents start
+| length            | JSON long         | yes      | The length of the blob stored in the file
```
rdblue, line 106: It would be nice to state that this is the compressed length.

format/puffin-spec.md
findepi: added blob-level `properties` field for this

format/puffin-spec.md
findepi: added

findepi:
thanks @rdblue for your detailed review.

changes applied.

format/puffin-spec.md
danielcweeks: I'll refer to this from the other PR as it doesn't address this.

format/puffin-spec.md
```diff
@@ -0,0 +1,144 @@
...
+Footer has the following structure
+
+```
+Magic FooterPayload FooterPayloadSize Flags Magic
+```
+
+where
+
+- `Magic`: four bytes, same as at the beginning of the file
+- `FooterPayload`: optionally compressed, UTF-8 encoded JSON payload describing the
```
kbendick, line 67: The mentioning of lz4 as the compression option should ideally be placed near the first mention of compression imo. Or perhaps a mention of the compression section would suffice.

I had to go looking for an indication of the compression type after reading this line which was somewhat confusing for me.

findepi: i would  remove mention of compression from here rather than elaborate this enumeration item with compression specification. However, it was added for the clarity here.

rdblue (approved):

rdblue:
I think this is ready to commit. Next step is to bring it up on the dev list and vote on it.

findepi:
thanks for the merge!

> I think this is ready to commit. Next step is to bring it up on the dev list and vote on it.

@rdblue  who can do this?

findepi:
per offline conversation, sent an email to the iceberg dev list (https://lists.apache.org/thread/950rz31y3kr3kz0zzncwokvgzbrmmz4q)
hope this is what i was supposed to do.

format/puffin-spec.md
```diff
@@ -0,0 +1,144 @@
...
+| offset            | JSON long                               | yes      | The offset in the file where the blob contents start
+| length            | JSON long                               | yes      | The length of the blob stored in the file (after compression, if compressed)
+| compression-codec | JSON string                             | no       | See [Compression codecs](#compression-codecs). If omitted, the data is assumed to be uncompressed.
+| properties        | JSON object with string property values | no       | storage for arbitrary meta-information about the blob
+
+### Blob types
+
+The blobs can be of a type listed below
+
+#### `apache-datasketches-theta-v1` blob type
```
szehon-ho, line 114: So to clarify, alpha-sketch is just one example, and the idea here is it support various statistics/secondary index that is to be defined, right?  Example, other sketches, or even bloom filter?

szehon-ho, line 114: So to clarify, alpha-sketch is just one example, and the idea here is it support various statistics/secondary index that is to be defined, right?  Example, other sketches, or even bloom filter?

rdblue: Yeah, the idea is to support bloom filters, dictionaries, and other index types in addition to sketches like this.

We'd also like to get histograms in there.

szehon-ho: Great, thanks

szehon-ho: Yea, led me to wonder on https://github.com/apache/iceberg/pull/4945, whether these stats/indices would apply to an individual data file, as well as across whole snapshot (I guess both will be very useful).

rdblue: It doesn't matter for this file format. I think the next step is to start thinking about adding these at the partition level.

szehon-ho: Yea makes sense, its separate.

lirui-apache:
Hey @findepi , do we have any docs regarding how the puffin files should be tracked by snapshots/manifests? We had implemented data file level indices, e.g. bloom filter, bitmap, and wonder whether we could adapt to this new format.

lirui-apache:
@findepi Thanks for the pointers! Our indices are used for data skipping so they have to be file-level. Currently we just store serialized indices in the files. I guess switching to Puffin format can help us better track metadata like snapshot and sequence numbers.

findepi:
@lirui-apache sounds very interesting.
BTW there is `#secondary-index` slack channel (https://apache-iceberg.slack.com/archives/C02TZ3NMBRC/p1654850392577929) which contains more people interested in use-case like yours!

PR #4945, opened 2022-06-02
Add table spec changes for statistics information in table snapshot

findepi:


findepi:
cc @alexjo2144 @RussellSpitzer

findepi:
Supersedes https://github.com/apache/iceberg-docs/pull/77 (just for reference, no useful content/comments there)

format/puffin-spec.md
```diff
@@ -0,0 +1,144 @@
...
+# Puffin file format
+
+This is a specification for Puffin, a file format designed to store
+information such as indexes and statistics about data managed in an
+Iceberg table that cannot be stored directly within the Iceberg manifest. A
+Puffin file contains arbitrary pieces of information (here called "blobs"),
+along with metadata necessary to interpret them. The blobs supported by Iceberg
+are documented at [Blob types](#blob-types).
+
+## Format specification
```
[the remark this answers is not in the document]
rdblue: I think that's something we would note in the Iceberg spec when we add how to reference Puffin files.

format/spec.md
```diff
@@ -513,6 +514,17 @@ Manifests for a snapshot are tracked by a manifest list.
 
 Valid snapshots are stored as a list in table metadata. For serialization, see Appendix C.
 
+Statistics files' metadata within `statistics` field is a struct with the following fields:
+
+| Field name                      | Type                               | Description                                                                                                                                                                                                              |
+|---------------------------------|------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **`location`**                  | `string`                           | Location of the statistics file. See [Puffin file format](../puffin).                                                                                                                                                    |
```
rdblue, line 521: Other places in the spec use "path" rather than "location" for files. How about using `stats-path` or `statistics-path`?

Also, this is out of date. It should link to `puffin-spec`.

format/spec.md
```diff
@@ -513,6 +514,17 @@ Manifests for a snapshot are tracked by a manifest list.
 
 Valid snapshots are stored as a list in table metadata. For serialization, see Appendix C.
 
+Statistics files' metadata within `statistics` field is a struct with the following fields:
```
rdblue, line 517: I think this should be a sub-section, like `Manifest Lists` below. It should also be located after the `Snapshot Retention Policy` sub-section. There's no need to place this before manifest lists.

In the new section for `Statistics file`, I think this should add any disclaimers about the use of stats, like @danielcweeks suggested.

format/spec.md
```diff
@@ -513,6 +514,17 @@ Manifests for a snapshot are tracked by a manifest list.
...
+
+| Field name                      | Type                               | Description                                                                                                                                                                                                              |
```
rdblue, line 519: Can you add the version columns to this table so it is clear that this is optional?

format/spec.md
```diff
@@ -513,6 +514,17 @@ Manifests for a snapshot are tracked by a manifest list.
...
+| **`file-size-in-bytes`**        | `long`                             | Size of the statistics file.                                                                                                                                                                                             |
+| **`file-footer-size-in-bytes`** | `long`                             | Size of the statistics file's footer. See [Puffin file format](../puffin) for footer definition.                                                                                                                         |
+| **`source-sequence-number`**    | `long`                             | Table sequence number at which the stats were calculated                                                                                                                                                                 |
+| **`statistics-fields-sets`**    | `map<string, list<list<integer>>>` | A map indicating which statistics are contained in the statistics file and on which columns they were calculated. The map keys are statistics sketch names and map values represent sets of columns, given by column ID. |
```
rdblue, line 525: I think this should include more information and be a bit easier to understand. Because this uses a map, you have to also use a list of lists of integers in case there is more than one set of columns covered by a blob. Instead, I think it is better to make the metadata a list of objects. That has a couple advantages:
* There can be one object per field set rather than combining by blob name
* The objects are extensible and can carry more information, like the NDV estimate (to avoid reading the file entirely)

format/puffin-spec.md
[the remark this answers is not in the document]
findepi: Note that this file is not being changed here. Just that the current PR is based on https://github.com/apache/iceberg/pull/4944, so it shows up in the diff.

format/spec.md
findepi: changed, thanks

format/spec.md
findepi: It's versioned and marked as optional the "A snapshot consists of the following fields" table.

Within "Statistics files' metadata" all fields are required. How should i indicate that?

format/spec.md
findepi: moved & added

format/spec.md
rdblue: Are these field required or optional?

format/puffin-spec.md
```diff
@@ -0,0 +1,144 @@
+---
+url: puffin-spec
```
rdblue, line 2: I merged https://github.com/apache/iceberg/pull/4944 so I think we can remove this.

format/spec.md
```diff
@@ -513,7 +514,6 @@ Manifests for a snapshot are tracked by a manifest list.
 
 Valid snapshots are stored as a list in table metadata. For serialization, see Appendix C.
 
-
```
rdblue, line 516: Can you revert this unnecessary whitespace change?

format/spec.md
```diff
@@ -495,6 +495,7 @@ A snapshot consists of the following fields:
 | _optional_ |            | **`manifests`**          | A list of manifest file locations. Must be omitted if `manifest-list` is present |
 | _optional_ | _required_ | **`summary`**            | A string map that summarizes the snapshot changes, including `operation` (see below) |
 | _optional_ | _optional_ | **`schema-id`**          | ID of the table's current schema when the snapshot was created |
+|            | _optional_ | **`statistics`**         | A list of [statistics files' metadata](#statistics-file)                                              |
```
rdblue, line 498: A list?

format/spec.md
```diff
@@ -630,6 +630,21 @@ When expiring snapshots, retention policies in table and snapshot references are
     2. The snapshot is not one of the first `min-snapshots-to-keep` in the branch (including the branch's referenced snapshot)
 5. Expire any snapshot not in the set of snapshots to retain.
 
+#### Statistics file
```
rdblue, line 633: I think this should start with some context about stats files. I think that we require them to be valid Puffin files and explain how they're used. They are also informational. The paragraph below the table is good and can be moved up here to clarify that.

format/puffin-spec.md
findepi: thanks for the merge! rebased

format/spec.md
findepi: thanks for pointing this out, fixed

format/spec.md
findepi: There can be two or more stat files

format/spec.md
findepi: good idea, added a word of intro, please take another look

format/spec.md
findepi: > * There can be one object per field set rather than combining by blob name

changed



> * The objects are extensible and can carry more information,

good point about extensibility.

i pushed a new version which allows that, and also added properties field

findepi:
comments applied. @rdblue mind taking another look?

format/spec.md
findepi: Here, all fields are required.

findepi: actually no, the newly added properties field is optional. i added a column in the table to indicate that

format/spec.md
```diff
@@ -496,6 +496,7 @@ A snapshot consists of the following fields:
 | _optional_ |            | **`manifests`**          | A list of manifest file locations. Must be omitted if `manifest-list` is present |
 | _optional_ | _required_ | **`summary`**            | A string map that summarizes the snapshot changes, including `operation` (see below) |
 | _optional_ | _optional_ | **`schema-id`**          | ID of the table's current schema when the snapshot was created |
+|            | _optional_ | **`statistics`**         | A list of [statistics files' metadata](#statistics-file). The field should be retained by writers, unless writer updates the statistics, or knows they became obsolete. |
```
rdblue, line 499: Why would there be more than one stats file? Wouldn't multiple files be merged together or one would make another obsolete?

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+
+| v2         | Field name                      | Type                                   | Description                                                                                            |
+|------------|---------------------------------|----------------------------------------|--------------------------------------------------------------------------------------------------------|
+| _required_ | **`statistics-path`**           | `string`                               | Path of the statistics file. See [Puffin file format](../puffin-spec).                                 |
+| _required_ | **`file-size-in-bytes`**        | `long`                                 | Size of the statistics file.                                                                           |
+| _required_ | **`file-footer-size-in-bytes`** | `long`                                 | Size of the statistics file's footer. See [Puffin file format](../puffin-spec) for footer definition.  |
+| _required_ | **`source-sequence-number`**    | `long`                                 | Table sequence number at which the stats were calculated                                               |
```
rdblue, line 647: Shouldn't this also include a snapshot ID at which the stats were calculated?

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+ignore statistics information. Statistics support is not required to read the table correctly.
+
+Statistics files' metadata within `statistics` table snapshot field is a struct with the following fields:
+
+| v2         | Field name                      | Type                                   | Description                                                                                            |
+|------------|---------------------------------|----------------------------------------|--------------------------------------------------------------------------------------------------------|
+| _required_ | **`statistics-path`**           | `string`                               | Path of the statistics file. See [Puffin file format](../puffin-spec).                                 |
+| _required_ | **`file-size-in-bytes`**        | `long`                                 | Size of the statistics file.                                                                           |
+| _required_ | **`file-footer-size-in-bytes`** | `long`                                 | Size of the statistics file's footer. See [Puffin file format](../puffin-spec) for footer definition.  |
```
rdblue, line 646: I think this needs to be more clear, since the Puffin footer contains a size. This is not the size from the Puffin footer, it is the size of the entire footer, which is the Puffin footer payload size + 16 (start magic, footer payload size, flags, end magic). Saying this is the "total footer size (not the footer payload size)" should be sufficient.

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+| _required_ | **`source-sequence-number`**    | `long`                                 | Table sequence number at which the stats were calculated                                               |
+| _required_ | **`statistics-metadata`**       | `list<statistic metadata>` (see below) | A list of the statistics metadata for statistics contained in the file with structure described below. |
+
+Statistic metadata is a struct with the following fields:
+
+| v2         | Field name           | Type                  | Description                                                                                      |
+|------------|----------------------|-----------------------|--------------------------------------------------------------------------------------------------|
+| _required_ | **`statistic-type`** | `string`              | Type of the statistic. Matches Blob type in the Puffin file.                                     |
```
rdblue, line 654: Why not just `type` or `blob-type`? I think this structure is probably going to be reused for indexes, so making it generic seems like a good idea to me. This is basically `type`, `fields`, and `properties` from the Puffin `BlobMetadata`.

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+|------------|---------------------------------|----------------------------------------|--------------------------------------------------------------------------------------------------------|
+| _required_ | **`statistics-path`**           | `string`                               | Path of the statistics file. See [Puffin file format](../puffin-spec).                                 |
+| _required_ | **`file-size-in-bytes`**        | `long`                                 | Size of the statistics file.                                                                           |
+| _required_ | **`file-footer-size-in-bytes`** | `long`                                 | Size of the statistics file's footer. See [Puffin file format](../puffin-spec) for footer definition.  |
+| _required_ | **`source-sequence-number`**    | `long`                                 | Table sequence number at which the stats were calculated                                               |
+| _required_ | **`statistics-metadata`**       | `list<statistic metadata>` (see below) | A list of the statistics metadata for statistics contained in the file with structure described below. |
```
rdblue, line 648: Can we call this something other than `statistics-metadata`? Maybe `blob-metadata` instead since this is likely to be shared with indexes?

format/spec.md
findepi: We don't have implementation for merging stat files yet.
One application can calculate some statistic on some columns and then someone else may can calculate a different statistic (or on different list of columns). This would lead to multiple stat files.

format/spec.md
findepi: > Also, if this is to be used in v1,

my intention was for this to be v2 feature.

> Shouldn't this also include a snapshot ID at which the stats were calculated?

also, or instead of?

please advise

format/spec.md
findepi: cool, will go with `type`

format/spec.md
findepi: cool, will change to that

findepi:
Applied comments. @rdblue can you please give it another look?

format/spec.md
szehon-ho: Trying to understand this (sorry if missed previous discussion), does a puffin file or list of puffin files have statistics for all data referenced by a snapshot?  We will not support statistics per data file?

szehon-ho: Trying to understand this (sorry if missed previous discussion), does a puffin file or list of puffin files have statistics for all data referenced by a snapshot?  We will not support statistics per data file?

rdblue: > We don't have implementation for merging stat files yet.

I think we want to make spec decisions based on what makes the most sense long term, not on what we've already built. It should be easy enough to merge these. I think we should have just one statistics file.

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
     2. The snapshot is not one of the first `min-snapshots-to-keep` in the branch (including the branch's referenced snapshot)
 5. Expire any snapshot not in the set of snapshots to retain.
 
+#### Statistics file
+
+Statistics files are valid [Puffin files](../puffin-spec). Statistics are informational. A reader can choose to
+ignore statistics information. Statistics support is not required to read the table correctly.
+
+Statistics files' metadata within `statistics` table snapshot field is a struct with the following fields:
```
rdblue, line 640: I think these sentences should refer to a singular statistics file if we agree there should be just one.

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+| _required_ | **`statistics-path`**           | `string`                               | Path of the statistics file. See [Puffin file format](../puffin-spec).                                                                    |
+| _required_ | **`file-size-in-bytes`**        | `long`                                 | Size of the statistics file.                                                                                                              |
+| _required_ | **`file-footer-size-in-bytes`** | `long`                                 | Total size of the statistics file's footer (not the footer payload size). See [Puffin file format](../puffin-spec) for footer definition. |
+| _required_ | **`source-sequence-number`**    | `long`                                 | Table sequence number at which the stats were calculated                                                                                  |
+| _required_ | **`blob-metadata`**             | `list<statistic metadata>` (see below) | A list of the statistics metadata for statistics contained in the file with structure described below.                                    |
```
rdblue, line 648: Looks like the type still has `list<statistic metadata>`.

format/spec.md
findepi: That means the transaction adding new statistics will need to rewrite the existing ones in the commit phase.
That would also imply different API -- it will no longer make sense for the stats writing application to write the stats file (and then commit its registration). Instead the API should take statistics in-memory, because their file destination is to be determined.
I am not yet convinced these are fair tradeoffs.

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+|------------|---------------------------------|-----------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
+| _required_ | **`statistics-path`**           | `string`                          | Path of the statistics file. See [Puffin file format](../puffin-spec).                                                                    |
+| _required_ | **`file-size-in-bytes`**        | `long`                            | Size of the statistics file.                                                                                                              |
+| _required_ | **`file-footer-size-in-bytes`** | `long`                            | Total size of the statistics file's footer (not the footer payload size). See [Puffin file format](../puffin-spec) for footer definition. |
+| _required_ | **`source-sequence-number`**    | `long`                            | Table sequence number at which the stats were calculated                                                                                  |
```
krisdas, line 647: curious to know, what's the recommended frequency to run stats writing application ? will not the stats file become outdated after a new commit on the same partition?

findepi: Stats are supposed to be approximate information about the data, so they don't need to be exact to still be useful.
When to use stats and when not to -- it's supposed to be up to the query engine to decide.

For writing stats with Trino, there is a PR: https://github.com/trinodb/trino/pull/12317.

format/spec.md
rdblue: I looked at the other PR, but I don't follow the argument here that it would make more sense to pass stats in memory. Whether you base stats on the previous file or not doesn't matter to the commit phase. The process updating stats produces a file and then sets it as the current one on a snapshot.

format/spec.md
rdblue: Yes, I think this should have the snapshot for at which stats were calculated.

There's also no reason to not support this in v1. As long as the snapshot ID is here for tracking, it should work just fine.

findepi: Changed to `source-snapshot-id` and made the whole thing supported in v1.

findepi: A side question -- now that we have v2 and we will have v3 of the spec, why do we want to add a new feature to v1?

format/spec.md
findepi: I asked for further explanation in the other PR (https://github.com/apache/iceberg/pull/4741#discussion_r895458863)

format/spec.md
```diff
@@ -631,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+|------------|------------|---------------------------------|-----------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`statistics-path`**           | `string`                          | Path of the statistics file. See [Puffin file format](../puffin-spec).                                                                    |
+| _required_ | _required_ | **`file-size-in-bytes`**        | `long`                            | Size of the statistics file.                                                                                                              |
+| _required_ | _required_ | **`file-footer-size-in-bytes`** | `long`                            | Total size of the statistics file's footer (not the footer payload size). See [Puffin file format](../puffin-spec) for footer definition. |
+| _required_ | _required_ | **`source-snapshot-id`**        | `long`                            | Table sequence number at which the stats were calculated                                                                                  |
```
flyrain, line 647: Kind of confused by the field name and description. Is it a snapshot id or sequence number? I assume it is a snapshot id.

findepi: good catch. Anyway this field needs to be removed now (https://github.com/apache/iceberg/pull/5021#issuecomment-1163533152)

format/spec.md
findepi: updated

format/spec.md
```diff
@@ -486,16 +486,17 @@ When reading v1 manifests with no sequence number column, sequence numbers for a
...
+| _optional_ | _required_ | **`manifest-list`**      | The location of a manifest list for this snapshot that tracks manifest files with additional metadata                                                           |
+| _optional_ |            | **`manifests`**          | A list of manifest file locations. Must be omitted if `manifest-list` is present                                                                                |
+| _optional_ | _required_ | **`summary`**            | A string map that summarizes the snapshot changes, including `operation` (see below)                                                                            |
+| _optional_ | _optional_ | **`schema-id`**          | ID of the table's current schema when the snapshot was created                                                                                                  |
+| _optional_ | _optional_ | **`statistics`**         | A [statistics file's metadata](#statistics-file). The field should be retained by writers, unless writer updates the statistics, or knows they became obsolete. |
```
flyrain, line 499: Do we need statistics in the snapshot level if we are going to "have one stats file for a table"?

flyrain: If there is only one statistics file per table, we can have it in [Table metadata](https://iceberg.apache.org/spec/#table-metadata) instead of [snapsthots](https://iceberg.apache.org/spec/#snapshots).
Is there a link why we want only one statistics file per table? I assume per-snapshot one makes more sense. For example, we can remove the statistics files of a snapshot when we expire the snapshot.

findepi: @rdblue do you have any thoughts about this?

format/spec.md
```diff
@@ -486,16 +486,17 @@ When reading v1 manifests with no sequence number column, sequence numbers for a
...
-|            | _required_ | **`sequence-number`**    | A monotonically increasing long that tracks the order of changes to a table |
-| _required_ | _required_ | **`timestamp-ms`**       | A timestamp when the snapshot was created, used for garbage collection and table inspection |
-| _optional_ | _required_ | **`manifest-list`**      | The location of a manifest list for this snapshot that tracks manifest files with additional metadata |
-| _optional_ |            | **`manifests`**          | A list of manifest file locations. Must be omitted if `manifest-list` is present |
-| _optional_ | _required_ | **`summary`**            | A string map that summarizes the snapshot changes, including `operation` (see below) |
-| _optional_ | _optional_ | **`schema-id`**          | ID of the table's current schema when the snapshot was created |
+| v1         | v2         | Field                    | Description                                                                                                                                                     |
```
rdblue, line 489: Can you revert the changes to the other fields? We don't need to have the whitespace changes affecting diffs.

format/spec.md
rdblue: This is a good point, but there are tradeoffs in both directions.

If we attach a stats file at the table level, then we don't need to add logic for carrying it through from one snapshot to the next. That's nice because older writers could be unaware of stats and not propagate the file. That means that clients would need to look through commit history to find the most recent stats file attached to an ancestor snapshot. In short, there's some compatibility work to do here if we use snapshot. On the other hand, if we attach stats to snapshots, then we can have stats at the branch level.

We could design a new table-level structure that tracks stats files per branch, something like this:

```json
"stats": [
    { "ref": "main", "last-updated-snapshot-id": ..., "stats-path": "...", "file-size-in-bytes": ..., "footer-size-in-bytes": ..., blob-metadata: [ ... ] },
    { "ref": "test", "last-updated-snapshot-id": ..., "stats-path": "...", "file-size-in-bytes": ..., "footer-size-in-bytes": ..., blob-metadata: [ ... ] }
  ]
```

I would probably prefer that new structure, or something similar. What do you think, @findepi and @flyrain?

flyrain: Trying to understand the table level stats files, assume the table only has one branch, so the stats file will be updated every time a new snapshot created. For example, instead of having (snapshot1, stats-file1), (snapshot2, stats-file2)... There is only one stats file. How does time travel work with one stats file?
With snapshot-level stats files, expireSnapshot and time travel would be more natural to handle. 
> That means that clients would need to look through commit history to find the most recent stats file attached to an ancestor snapshot. 

I understand the compatibility concern, and the snapshots created by the old client won't have the stats files. Maybe I'm missing something, but it shouldn't be so hard for the new client find the most recent stats files and carry them over, right?

format/spec.md
findepi: Sorry, I am not doing this on purpose. it's my IntelliJ reformatting the tables in .md files automatically.

format/spec.md
findepi: > it shouldn't be so hard for the new client find the most recent stats files and carry them over, right?

@flyrain 
with current implementation --
iceberg library will do this automatically.

however, clients using old Iceberg library will "forget" stats when creating new snapshot (eg inserting new data)

> I would probably prefer that new structure, or something similar. What do you think, @findepi and @flyrain?


@rdblue i am not fluent in Iceberg Table / Snapshot intricacies, lifecycles, and branching, so I don't have opinion.

I know what we need to be able to
- attach a stats file to a table 
- update/replace/remove existing stats file
- retain existing stats file when doing eg DML on a table 
- reading old snapshot (time travel) should be able to use the stat file as was present in the table at that time

i believe attaching the stats file to snapshot addresses these needs.
if there is some other and better way to address these needs, please be explicit. I still need guidance in Iceberg world.

format/spec.md
findepi: Fixed

format/spec.md
```diff
@@ -631,6 +632,30 @@ When expiring snapshots, retention policies in table and snapshot references are
     2. The snapshot is not one of the first `min-snapshots-to-keep` in the branch (including the branch's referenced snapshot)
 5. Expire any snapshot not in the set of snapshots to retain.
 
+#### Statistics file
+
+Statistics files are valid [Puffin files](../puffin-spec). Statistics are informational. A reader can choose to
```
ajantha-bhat, line 637: Do we need to add one more abstraction layer for `Statistics`.

Thinking of partition stats, For each partition value, if I am storing row count and file count stats. There can be million rows.
So, puffin may not be efficient for that. 

So, can we have an abstraction for statistics metadata?
Where stats like sketches are stored in Puffin and simple stats are stored in parquet or Hfile or some other format ?
I can also see more than one stats file per snapshot (one in puffin format and some in other format)

format/spec.md
ajantha-bhat: +1 for storing at snapshot level.

rdblue: After thinking about this more and talking with @danielcweeks, I think that stats files should be tracked by a new top-level `stats` field in `TableMetadata`.

The issues with tracking stats in a snapshot are:
1. There is now a rule about carrying forward stats files that all writers must implement. This makes it difficult to manage an environment with multiple writer versions. The stats files will be dropped by some writers, and there isn't a good way to detect whether omitting stats was intentional (they are no longer accurate) or accidental (the writer didn't know about stats).
2. Adding or updating stats creates a new snapshot. The new snapshot has no data changes, but will cause retries and validations for concurrent writers.

A top-level `stats` field addresses the writer challenge because modifying stats files is separate from adding a snapshot and updating a ref. The catalog can detect the difference between removing a stats file (`stats` list exists) and accidentally dropping stats (`stats` list is omitted).

This also allows us to add to the Table API, rather than adding to the Snapshot API. The table API is better because matches are fuzzy -- there isn't necessarily a stats file that was calculated for a given snapshot so we don't want to associated a snapshot 1-to-1 with a stats file. (Side benefit: we don't copy all the stats metadata into each snapshot!)

The main drawback is that this would need to contain more than one stats file per branch in the list in order to use the stats file that was current as of a snapshot. But there is a benefit to this as well because you can always calculate old stats files and add them. For example, you can calculate stats for all of the tags in a table after they are tagged, even if the tagged snapshot itself didn't have stats originally.

Over all, I'm more and more confident that a top-level `stats` field is the right way to go.

Zhangg7723: +1 for attaching stats file to table level, stats data needs to be re-calculated for each snapshot that insert or update data, the history stats files is only used for time travel query and not useful next calculation, it seems to be lower value.

findepi: If this go into table metadata, how should i find out stat file appropriate to read from some previous table snapshot?
Should we keep multiple entries with `"ref": "main"` and basically pick the latest one with `value of "last-updated-snapshot-id" < selected_snapshot_id`?

Also, when updating the stats, how do I select which stat file to update from (which is current, and which historical)?
To avoid any confusion, we could make "ref", the branch name, as a key in the map, with value being an ordered list of stat file entries.

RussellSpitzer: I'm not really set on either location but I do have concerns like @findepi about issuing queries against older snapshots. Does this means we keep a history of stat files in the table metadata as well?

format/spec.md
```diff
@@ -631,6 +632,30 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+ignore statistics information. Statistics support is not required to read the table correctly.
+
+Statistics file's metadata within `statistics` table snapshot field is a struct with the following fields:
+
+| v1         | v2         | Field name                      | Type                              | Description                                                                                                                               |
+|------------|------------|---------------------------------|-----------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`statistics-path`**           | `string`                          | Path of the statistics file. See [Puffin file format](../puffin-spec).                                                                    |
+| _required_ | _required_ | **`file-size-in-bytes`**        | `long`                            | Size of the statistics file.                                                                                                              |
```
RussellSpitzer, line 645: Do we need an encryption key_metadata here as in the DataFile Spec?

format/spec.md
RussellSpitzer: I'm trying to think through the table level structure, I assume I need to be able to store data like
Stats files : Valid for Snapshots [x -> z]. It seems like we then have some difficulty in extending this range for commits which do not invalidate the snapshot unless the writer is clear on the stats file. A stats file tracking table wide aggregates could survive a `replace` but a bloom filter which pointed to individual data files would not.

I guess it is then still up to the writer of the new metadata.json to know whether or not the Stats file is valid for its new commit.

rdblue: @findepi, you would find the stats file written for the closest snapshot. That may be a snapshot that was written after: if a compaction runs and then stats are calculated for the compacted snapshot it would be closest for the snapshot that was compacted.

The file that should be updated is the newest ancestor with stats in a given snapshot's history. There are utils for traversing ancestors, so you'd do that until you find an ancestor with a stats file.

For the storage, I think we should use a list. Table metadata can be indexed however is convenient in memory. Adding requirements for ordering within the format is difficult (what happens when one is out of order?) and leads to issues with time skew.

@RussellSpitzer, I think it is simple to traverse ancestors to find the last stats file. And we can add stats files for older snapshots later if needed (the tagging case). I think it is actually much cleaner by not keeping these in snapshots.

RussellSpitzer: I don't think Snapshots is the right place to hold them @rdblue , I just am not sure how we are traversing ancestors in this case. Are you suggesting we have a list of stats files, each of which is associated with a single snapshot file. Then we we traverse our ancestors in the snapshot tree to find stats files? 

When we don't have a stats file for the current snapshot (or snapshot in question) do we return no stats files? I think that would make sense and instead we rely on a metadata only operation for associating old stats files with newer snapshots.

For example, older or non-puffin capable writer adds a snapshot B to existing table with Snapshot A

```
Table : { snaps {A, B} , stats { A -> A' } }
```

When reading B I think we have two options

1. Report there are no stats files associated with B
2. Use ancestor, find A -> A' and then attempt to determine if the operation to make B invalidated A'

I think 1. is probably safest and then we would add an operation for a client that understands A', `propagate_stats` or something which would add a new metadata.json
```
Table {snaps {A, B}, stats { A -> A', B -> A'}
```

findepi: > We could design a new table-level structure that tracks stats files per branch, something like this:
> 
> ```json
> "stats": [
>     { "ref": "main", "last-updated-snapshot-id": 1, "stats-path": "...", "file-size-in-bytes": 1024, "footer-size-in-bytes": 100, "blob-metadata": [ {}, {} ] },
>     { "ref": "test", "last-updated-snapshot-id": 2, "stats-path": "...", "file-size-in-bytes": 1024, "footer-size-in-bytes": 100, "blob-metadata": [ {}, {} ] }
>   ]
> ```

@rdblue what do we need "ref" for?

format/spec.md
RussellSpitzer: I'm not sure our security review team will agree. Something like first and last name combos which have a very finite set of values, the ability to detect whether the dataset contains a value +- a false positive rate may be considered a vulnerability. I'm gonna see if they'll accept it.

ggershinsky: +1 to adding encryption key_metadata here, similar to the ManifestFile structures (we can also use the same approach to their encryption). 
Regarding sensitivity of this metadata - as an example, Parquet encrypts the bloom filters. Here goes the doc section with details (and reasons) 
https://github.com/apache/parquet-format/blob/master/BloomFilter.md#encryption

rdblue: Sounds reasonable.

rdblue:
@findepi, can you update this to add the stats structure that we're discussing in the thread?

format/spec.md
flyrain: +1 for the idea to detach stats files from a snapshot. This is due to immutability of snapshot, which means any internal state change of a snapshot will have to create a new snapshot. We can update/generate stats files without creating a new snapshot by detaching them.

We can still consider stats file as an external property of snapshot, so that stats files are indexed by snapshot. So that we search the stats files by searching the snapshot history.
```
"stats": [
    { "snapshot-id": 1, stats-files[{"stats-path": "path/file1", "file-size-in-bytes": 1024, "footer-size-in-bytes": 100, "blob-metadata": [ {}, {} ]},{"stats-path": "path/file2", "file-size-in-bytes": 1024, "footer-size-in-bytes": 100, "blob-metadata": [ {}, {} ]}]},
    { "snapshot-id": 2, stats-files[{"stats-path": "path/file3", "file-size-in-bytes": 1024, "footer-size-in-bytes": 100, "blob-metadata": [ {}, {} ]},{"stats-path": "path/file4", "file-size-in-bytes": 1024, "footer-size-in-bytes": 100, "blob-metadata": [ {}, {} ]}]},
  ]
```

rdblue: @flyrain, I think there should still be just one stats file per snapshot. It is easy to merge them and we don't want to accumulate them.

rdblue: @findepi, do you think you can update this PR to add the `stats` structure as discussed here?

format/spec.md
```diff
@@ -632,6 +633,30 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+
+| v1         | v2         | Field name                           | Type                   | Description                                                                                      |
+|------------|------------|--------------------------------------|------------------------|--------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`type`**                           | `string`               | Type of the statistic. Matches Blob type in the Puffin file.                                     |
+| _required_ | _required_ | **`fields`**                         | `list<integer>`        | Ordered list of fields, given by field ID, on which the statistic was calculated.                |
+| _required_ | _required_ | **`source-snapshot-id`**             | `long`                 | ID of the Iceberg table's snapshot the blob was computed from.                                   |
+| _required_ | _required_ | **`source-snapshot-sequence-numer`** | `long`                 | Sequence number of the Iceberg table's snapshot the blob was computed from.                      |
```
findepi, line 657: if we attach stats file to a given snapshot (per https://github.com/apache/iceberg/pull/4945#discussion_r905295067) then these fields become redundant

format/spec.md
findepi: Sure, i can add `key_metadata | Base64-encoded implementation-specific key metadata for encryption` here.

do we want whole-file encryption, or do we need to encrypt individual blobs?
Also, if a file is encrypted, is it more useful to store file-size-in-bytes before encryption, or we also/instead need to have a number after the encryption?

format/spec.md
findepi: Updated accordingly.

findepi:
> @findepi, can you update this to add the stats structure that we're discussing in the thread?

Done (https://github.com/apache/iceberg/pull/4945#discussion_r927705073)

format/puffin-spec.md
```diff
@@ -144,3 +142,5 @@ When writing a Puffin file it is recommended to set the following fields in the
 
 - `created-by` - human-readable identification of the application writing the file,
   along with its version. Example "Trino version 381".
+- `source-snapshot-id` - a table snapshot which was used to calculate blob contents
+- `source-sequence-number` - sequence number of the table snapshot used to calculate blob contents
```
rdblue, line 146: Why did these move to properties rather than required blob metadata?

core/src/main/java/org/apache/iceberg/puffin/Blob.java
```diff
@@ -30,29 +30,23 @@
...
+      String type, List<Integer> inputFields, ByteBuffer blobData,
+      @Nullable PuffinCompressionCodec requestedCompression, Map<String, String> properties) {
     Preconditions.checkNotNull(type, "type is null");
     Preconditions.checkNotNull(inputFields, "inputFields is null");
     Preconditions.checkNotNull(blobData, "blobData is null");
     Preconditions.checkNotNull(properties, "properties is null");
     this.type = type;
     this.inputFields = ImmutableList.copyOf(inputFields);
-    this.snapshotId = snapshotId;
-    this.sequenceNumber = sequenceNumber;
```
rdblue, line 55: I think the snapshot and sequence number for the blob should be retained. That allows preserving blobs that were not updated.

format/spec.md
```diff
@@ -632,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+#### Snapshot statistics
+
+Snapshot statistics files are valid [Puffin files](../puffin-spec). Statistics are informational. A reader can choose to
+ignore statistics information. Statistics support is not required to read the table correctly.
+
+Statistics files metadata within `snapshot-statistics` table metadata field is a struct with the following fields:
+
+| v1         | v2         | Field name                      | Type                              | Description                                                                                                                               |
+|------------|------------|---------------------------------|-----------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`snapshot-id`**               | `string`                          | ID of the Iceberg table's snapshot the blob was computed from.                                                                            |
```
rdblue, line 644: This refers to "the blob", but I think it should be for the stats file?

format/spec.md
```diff
@@ -632,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+| _required_ | _required_ | **`statistics-path`**           | `string`                          | Path of the statistics file. See [Puffin file format](../puffin-spec).                                                                    |
+| _required_ | _required_ | **`file-size-in-bytes`**        | `long`                            | Size of the statistics file.                                                                                                              |
+| _required_ | _required_ | **`file-footer-size-in-bytes`** | `long`                            | Total size of the statistics file's footer (not the footer payload size). See [Puffin file format](../puffin-spec) for footer definition. |
+| _required_ | _required_ | **`blob-metadata`**             | `list<blob metadata>` (see below) | A list of the blob metadata for statistics contained in the file with structure described below.                                          |
+
+Statistic metadata is a struct with the following fields:
```
rdblue, line 650: Do you mean blob metadata?

format/spec.md
```diff
@@ -632,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+
+| v1         | v2         | Field name       | Type                  | Description                                                                                      |
+|------------|------------|------------------|-----------------------|--------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`type`**       | `string`              | Type of the statistic. Matches Blob type in the Puffin file.                                     |
```
rdblue, line 654: This isn't a type for the statistic, it is a type for the blob.

format/spec.md
```diff
@@ -632,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
...
+| _required_ | _required_ | **`fields`**     | `list<integer>`       | Ordered list of fields, given by field ID, on which the statistic was calculated.                |
+| _required_ | _optional_ | **`properties`** | `map<string, string>` | Additional properties associated with the statistic. Matches Blob properties in the Puffin file. |
```
rdblue, line 656: I don't think that we need to require that the properties match what is in the file. The file could contain more information that isn't relevant here. The important thing is that this is where actual values can be stored to avoid reading Puffin files during job planning.

format/spec.md
```diff
@@ -632,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
     2. The snapshot is not one of the first `min-snapshots-to-keep` in the branch (including the branch's referenced snapshot)
 5. Expire any snapshot not in the set of snapshots to retain.
 
+#### Snapshot statistics
+
+Snapshot statistics files are valid [Puffin files](../puffin-spec). Statistics are informational. A reader can choose to
+ignore statistics information. Statistics support is not required to read the table correctly.
+
+Statistics files metadata within `snapshot-statistics` table metadata field is a struct with the following fields:
```
rdblue, line 640: I don't think that these should be "snapshot statistics" since these aren't part of `snapshot`. These are actually table statistics that are associated with a particular snapshot of the table. I think that `statistics` or `table-statistics` are better names.

format/spec.md
```diff
@@ -632,6 +632,29 @@ When expiring snapshots, retention policies in table and snapshot references are
     2. The snapshot is not one of the first `min-snapshots-to-keep` in the branch (including the branch's referenced snapshot)
 5. Expire any snapshot not in the set of snapshots to retain.
 
+#### Snapshot statistics
```
rdblue, line 635: Can you move this to be `Table statistics` just below `Table Metadata Fields`? I think that's a better place since these aren't part of a snapshot.

format/spec.md
ggershinsky: > add key_metadata | Base64-encoded implementation-specific key metadata for encryption here.

sgtm

> do we want whole-file encryption, or do we need to encrypt individual blobs?

Both key_metadata and the encryption itself are implementation-specific, so probably no need to add these details here. (wrt implementations, we can start with a simple file encrypting stream, similar to manifest files; and expand to other techniques later)

> if a file is encrypted, is it more useful to store file-size-in-bytes before encryption, or we also/instead need to have a number after the encryption

this depends on the intended use of this field. For manifest files, we don't use the file-size-in-bytes field for decryption - we get the full file length directly from InputFile.getLength() which is sourced from the file system. In other words, decryption is ok with any definition of this field (because it doesn't use it).

findepi:
applied/answered the comments

findepi:
Thank you @rdblue for your detailed review.

format/spec.md
findepi: @rdblue are you OK with me adding the following

`add key_metadata | Base64-encoded implementation-specific key metadata for encryption here.`

within this PR?

rdblue: Yes, I think it is a reasonable addition.

core/src/main/java/org/apache/iceberg/puffin/Blob.java
rdblue: It is related to the metadata added here, but unnecessary for the purpose of this PR. I think it needs more discussion and I'd like to focus on making progress on the main content of the PR.

Right now, I need to think about this more. The rationale for these fields is unchanged, although we could reconsider it by making the search for relevant metrics more complicated.

format/spec.md
```diff
@@ -665,9 +665,34 @@ Table metadata consists of the following fields:
...
+| _required_ | _required_ | **`blob-metadata`**             | `list<blob metadata>` (see below) | A list of the blob metadata for statistics contained in the file with structure described below.                                          |
+
+Blob metadata is a struct with the following fields:
+
+| v1         | v2         | Field name       | Type                  | Description                                                                                        |
+|------------|------------|------------------|-----------------------|----------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`type`**       | `string`              | Type of the blob. Matches Blob type in the Puffin file.                                            |
+| _required_ | _required_ | **`fields`**     | `list<integer>`       | Ordered list of fields, given by field ID, on which the statistic was calculated.                  |
+| _required_ | _optional_ | **`properties`** | `map<string, string>` | Additional properties associated with the statistic. Subset of Blob properties in the Puffin file. |
```
rdblue, line 694: I think it is probably a typo that this is required in v1 but optional in v2.

format/spec.md
```diff
@@ -665,9 +665,34 @@ Table metadata consists of the following fields:
...
+| _required_ | _required_ | **`blob-metadata`**             | `list<blob metadata>` (see below) | A list of the blob metadata for statistics contained in the file with structure described below.                                          |
+
+Blob metadata is a struct with the following fields:
+
+| v1         | v2         | Field name       | Type                  | Description                                                                                        |
+|------------|------------|------------------|-----------------------|----------------------------------------------------------------------------------------------------|
+| _required_ | _required_ | **`type`**       | `string`              | Type of the blob. Matches Blob type in the Puffin file.                                            |
+| _required_ | _required_ | **`fields`**     | `list<integer>`       | Ordered list of fields, given by field ID, on which the statistic was calculated.                  |
```
rdblue, line 693: Right now, blob metadata includes the snapshot ID and sequence number. I think that should be included here until we decide to remove them from blob metadata in Puffin files. These should mostly match.

format/spec.md
```diff
@@ -665,9 +665,34 @@ Table metadata consists of the following fields:
 | _optional_ | _required_ | **`sort-orders`**| A list of sort orders, stored as full sort order objects. |
 | _optional_ | _required_ | **`default-sort-order-id`**| Default sort order id of the table. Note that this could be used by writers, but is not used when reading because reads use the specs stored in manifest files. |
 |            | _optional_ | **`refs`** | A map of snapshot references. The map keys are the unique snapshot reference names in the table, and the map values are snapshot reference objects. There is always a `main` branch reference pointing to the `current-snapshot-id` even if the `refs` map is null. |
+| _optional_ | _optional_ | **`snapshot-statistics`** | A list (optional) of [table statistics](#table-statistics). |
```
rdblue, line 668: I'd still prefer `statistics` instead of `snapshot-statistics`. I can see how `table-statistics` is a bit too broad since they are based on a snapshot, but `snapshot-statistics` sounds more specific than needed.

findepi:
Thank you @rdblue for your detailed review. Applied comments!

rdblue (approved):

PR #4946, opened 2022-06-02
API: Support registering filesystem tables to metastore based catalogs

### LINKED ISSUE #4794: Error Registering Hadoop Based Table Into Hive Metastore based table.
This current registerTable function is  not working for converting hadoop catalog based tables to glue based tables. because tables written using hadoop catalog writes metadata.json file in the form v<V+1>.metadata.json

And tables written using glue and hive catalog have format <V+1>-.metadata.json

Hence when we call registerTable with HadoopCatalog based table [parseVersion ](https://github.com/apache/iceberg/blob/387e414689c59d5a3849adec35823d7aecb209d7/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java#L334) function would fail while parsing.

after renaming the latest metadata.json file I was able to register and query the data.

we can handle this in registerTable function as follows.

1.) We shall call parseVersion() in registerTable. If parsing gets failed in [BaseMetastoreTableOperations](https://iceberg.apache.org/javadoc/0.13.1/index.html?org/apache/iceberg/BaseMetastoreTableOperations.html) then the file can be in File-System-Tables spec and we can reparse the file using File-System-Tables spec.

2.) We can then rename the metadata file in required spec and use the renamed file path as metadataFileLocation.

ajantha-bhat:
Filesystem tables and metastore tables have different naming conventions for TableMetadata file as per spec.
So, when a filesystem table is registered to hive catalog (or any metastore based catalog), the register was successful but loading the table used to fail in `parseVersion` as it was expecting the naming convention of metastore table.

Fixed it by using metastore table naming convention while registering. 

Fixes #4794

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
@@ -151,7 +151,7 @@ public abstract class BaseMetastoreTableOperations implements TableOperations {
     this.shouldRefresh = false;
   }
 
-  protected String writeNewMetadata(TableMetadata metadata, int newVersion) {
+  public String writeNewMetadata(TableMetadata metadata, int newVersion) {
     String newTableMetadataFilePath = newTableMetadataFilePath(metadata, newVersion);
     OutputFile newMetadataLocation = io().newOutputFile(newTableMetadataFilePath);
 
--- a/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
@@ -151,7 +151,7 @@ public abstract class BaseMetastoreTableOperations implements TableOperations {
     this.shouldRefresh = false;
   }
 
-  public String writeNewMetadata(TableMetadata metadata, int newVersion) {
+  protected String writeNewMetadata(TableMetadata metadata, int newVersion) {
     String newTableMetadataFilePath = newTableMetadataFilePath(metadata, newVersion);
     OutputFile newMetadataLocation = io().newOutputFile(newTableMetadataFilePath);
 
@@ -336,7 +336,10 @@ public abstract class BaseMetastoreTableOperations implements TableOperations {
     int versionEnd = metadataLocation.indexOf('-', versionStart);
     try {
       return Integer.valueOf(metadataLocation.substring(versionStart, versionEnd));
-    } catch (NumberFormatException e) {
+    } catch (NumberFormatException | StringIndexOutOfBoundsException e) {
+      // As per the spec, metastore tables and filesystem tables
+      // will have different representation for metadata file names.
+      // So, StringIndexOutOfBoundsException can happen when parsing a filesystem based metadata file name.
       LOG.warn("Unable to parse version from metadata location: {}", metadataLocation, e);
       return -1;
     }
--- a/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
@@ -331,15 +331,24 @@ public abstract class BaseMetastoreTableOperations implements TableOperations {
     return metadataFileLocation(meta, String.format("%05d-%s%s", newVersion, UUID.randomUUID(), fileExtension));
   }
 
+  /**
+   * Parse the version from table metadata file name.
+   *
+   * @param metadataLocation table metadata file location
+   * @return version of the table metadata file in success case and
+   * -1 if the version is not parsable (as a sign that the metadata is not part of this catalog)
+   */
   private static int parseVersion(String metadataLocation) {
     int versionStart = metadataLocation.lastIndexOf('/') + 1; // if '/' isn't found, this will be 0
     int versionEnd = metadataLocation.indexOf('-', versionStart);
+    if (versionEnd < 0) {
+      // found filesystem table's metadata
+      return -1;
+    }
+
     try {
       return Integer.valueOf(metadataLocation.substring(versionStart, versionEnd));
-    } catch (NumberFormatException | StringIndexOutOfBoundsException e) {
-      // As per the spec, metastore tables and filesystem tables
-      // will have different representation for metadata file names.
-      // So, StringIndexOutOfBoundsException can happen when parsing a filesystem based metadata file name.
+    } catch (NumberFormatException e) {
       LOG.warn("Unable to parse version from metadata location: {}", metadataLocation, e);
       return -1;
     }
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
@@ -246,9 +246,24 @@ public class HiveCatalog extends BaseMetastoreCatalog implements SupportsNamespa
     TableOperations ops = newTableOps(identifier);
     InputFile metadataFile = fileIO.newInputFile(metadataFileLocation);
     TableMetadata metadata = TableMetadataParser.read(ops.io(), metadataFile);
+
+    if (!isMetastoreTable(metadataFileLocation)) {
+      // Metastore table and filesystem tables will have different naming conventions as per the spec.
+      // So, to support importing filesystem tables to metastore catalog,
+      // create a new metadata file with contents of filesystem table but with a metastore table naming convention.
+      BaseMetastoreTableOperations baseOps = (BaseMetastoreTableOperations) ops;
+      String newFileName = baseOps.writeNewMetadata(metadata, baseOps.currentVersion() + 1);
+      metadataFile = fileIO.newInputFile(newFileName);
+      metadata = TableMetadataParser.read(ops.io(), metadataFile);
+    }
+
     ops.commit(null, metadata);
 
     return new BaseTable(ops, identifier.toString());
+  }
+
+  private boolean isMetastoreTable(String metadataFileLocation) {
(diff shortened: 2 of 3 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -246,11 +246,26 @@ public org.apache.iceberg.Table registerTable(TableIdentifier identifier, String
...
+      metadata = TableMetadataParser.read(ops.io(), metadataFile);
+    }
+
     ops.commit(null, metadata);
 
     return new BaseTable(ops, identifier.toString());
   }
 
+  private boolean isMetastoreTable(String metadataFileLocation) {
+    return metadataFileLocation.substring(metadataFileLocation.lastIndexOf("/") + 1).contains("-");
```
rdblue, line 266: I think this should be more specific. Why not confirm that there's a number using a regex?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -246,11 +246,26 @@ public org.apache.iceberg.Table registerTable(TableIdentifier identifier, String
     TableOperations ops = newTableOps(identifier);
     InputFile metadataFile = fileIO.newInputFile(metadataFileLocation);
     TableMetadata metadata = TableMetadataParser.read(ops.io(), metadataFile);
+
+    if (!isMetastoreTable(metadataFileLocation)) {
+      // Metastore table and filesystem tables will have different naming conventions as per the spec.
```
rdblue, line 251: This seems unnecessary. Why not just update `parseVersion` to return `-1` if the version isn't valid? I think that's a better solution than rewriting the metadata.

rdblue:
@pvary, I think that this is a case that makes sense for the register method. We highly recommend using a real metastore rather than using the Hadoop strategy. To convert from a Hadoop catalog to a metastore catalog, you'd need this to work. To me, the register table method is mainly for migration between catalogs.

pvary:
@rdblue: Putting that in this way, and I agree with you. If this is a migration, then this is the way to go.

Thanks for the explanation!

ajantha-bhat:
@pvary : Yes, this was for migration purposes. As we are aware that keeping the same table in two catalog is dangerous and leads to data mismatch. The user who migrates needs to take care of deleting the table in source catalog using drop table with purge=false  option using the catalog interface.

@rdblue : Thanks for clarification and review. I will address the comments.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
ajantha-bhat: I like this suggestion. For filesystem tables, along with fixing register problem for hive, it can fix register problem for other catalogs also once they support it.

ajantha-bhat:
@rdblue : I have addressed the comments. Thanks.

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
```diff
@@ -336,7 +336,10 @@ private static int parseVersion(String metadataLocation) {
     int versionEnd = metadataLocation.indexOf('-', versionStart);
     try {
       return Integer.valueOf(metadataLocation.substring(versionStart, versionEnd));
-    } catch (NumberFormatException e) {
+    } catch (NumberFormatException | StringIndexOutOfBoundsException e) {
```
rdblue, line 339: Why not check `versionEnd`? If `versionEnd` is < 0, then return -1. No need to catch an exception.

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveTableTest.java
```diff
@@ -404,6 +389,80 @@ public void testRegisterTable() throws TException {
...
+            "Hive metastore should not have this table", NoSuchObjectException.class,
+            "table not found",
+            () -> metastoreClient.getTable(DB_NAME, "table1"));
+    AssertHelpers.assertThrows(
+            "Hive catalog should fail to load the table", NoSuchTableException.class,
+            "Table does not exist:",
+            () -> catalog.loadTable(identifier));
+
+    // register the table to hive catalog using the latest metadata file
+    metadataFiles.sort(Collections.reverseOrder());
```
rdblue, line 423: Rather than guessing which file is the metadata file, I think you should just get the location from the table:

```java
  String metadataLocation = ((BaseTable) table).operations().current().metadataFileLocation();
```

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveTableTest.java
```diff
@@ -404,6 +389,80 @@ public void testRegisterTable() throws TException {
...
+    Assert.assertNotNull(metastoreClient.getTable(DB_NAME, "table1"));
+
+    // load the table in hive catalog
+    table = catalog.loadTable(identifier);
+    Assert.assertNotNull(table);
+
+    // insert some data
+    appendData(table, "file2");
+    tasks = Lists.newArrayList(table.newScan().planFiles());
+    Assert.assertEquals("Should scan 2 files", 2, tasks.size());
```
rdblue, line 434: This should now verify that the table has `file1` and `file2`.

ajantha-bhat:
@rdblue : Thanks for the review again. I have addressed all the comments.

ajantha-bhat:
ping @rdblue

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
```diff
@@ -334,6 +334,12 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
   private static int parseVersion(String metadataLocation) {
     int versionStart = metadataLocation.lastIndexOf('/') + 1; // if '/' isn't found, this will be 0
     int versionEnd = metadataLocation.indexOf('-', versionStart);
+    if (versionEnd < 0) {
+      // As per the spec, metastore tables and filesystem tables
+      // will have different representation for metadata file names.
+      LOG.warn("Found filesystem table's metadata : {}", metadataLocation);
+      return -1;
+    }
```
pvary, line 342: nit: we usually leave a new line after blocks

ajantha-bhat:
@pvary : I have addressed the comments. Please check. Thanks.

pvary (approved):
+1 from my side. Let's see if @rdblue has other comments, or not. If not I will merge this around at the end of the week.

Thanks @ajantha-bhat!

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
```diff
@@ -331,9 +331,21 @@ private String newTableMetadataFilePath(TableMetadata meta, int newVersion) {
...
+   *
+   * @param metadataLocation table metadata file location
+   * @return version of the table metadata file in success case and
+   * -1 if the version is not parsable (as a sign that the metadata is not part of this catalog)
+   */
   private static int parseVersion(String metadataLocation) {
     int versionStart = metadataLocation.lastIndexOf('/') + 1; // if '/' isn't found, this will be 0
     int versionEnd = metadataLocation.indexOf('-', versionStart);
+    if (versionEnd < 0) {
+      LOG.warn("Found filesystem table's metadata : {}", metadataLocation);
```
rdblue, line 345: I don't think this is a good thing to warn about. Warnings are for potential problems, while this is a normal case if you convert a table or control a name some other way.

This is also inaccurate. We can't really know why the file didn't include a version and shouldn't guess.

ajantha-bhat: removed the log warning and added it as a code comment.

pvary: The more we talk about this, the less convinced I am that we should differentiate between the parsing errors. Nothing really prevents a developer to create their own `TableOperations` and using their own metadata filename format.

I think it would be good to have a log for the event in any case, at least in info level.

Maybe rewrite the whole parsing part to this:
```
  private static int parseVersion(String metadataLocation) {
    try {
      int versionStart = metadataLocation.lastIndexOf('/') + 1; // if '/' isn't found, this will be 0
      int versionEnd = metadataLocation.indexOf('-', versionStart);
      return Integer.valueOf(metadataLocation.substring(versionStart, versionEnd));
    } catch (Throwable e) {
      LOG.info("Unable to parse version from metadata location: {}", metadataLocation, e);
      return -1;
    }
  }
```

What do you think?

ajantha-bhat: SGTM. waiting for @rdblue 's opinion to avoid me working back and forth on this.

pvary:
> @pvary : I have addressed the comments. Please check. Thanks.

Let's give him a few more days, then I will merge.
Thanks, Peter

core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java
rdblue: I think this should check for `versionEnd`. That indicates that we're using a file with no version and should return -1.

If we also want to return -1 for `NumberFormatException` and warn, I'm all for that. But we should definitely not catch `Throwable`. That's dangerous. Let's handle the cases we know about.

ajantha-bhat: Thanks for the inputs @rdblue. Agree that catching `Throwable` is not a good practice. I have reverted the latest change. So, final code is same as what you described above.

rdblue (approved):

rdblue:
Thanks, @ajantha-bhat!

PR #4947, opened 2022-06-02
API: Add expression equivalence testing

rdblue:
This adds `ExpressionUtil.equivalent` and `ExpressionUtil.selectsPartitions` to determine whether an expression will select whole files in a partition spec. This is needed to determine whether a filter is guaranteed to delete only whole files, or whether a row-level plan should be used. It is also needed to implement aggregate pushdown with non-trivial filters because filters must match whole files in order to use file metadata to satisfy a query.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/expressions/BoundLiteralPredicate.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/BoundLiteralPredicate.java
@@ -122,7 +122,7 @@ public class BoundLiteralPredicate<T> extends BoundPredicate<T> {
             break;
           case GT_EQ:
             if (other.op() == Operation.GT) {
-              // >= 5 is equivalent to > 5
+              // >= 5 is equivalent to > 4
               return toLong(literal()) == toLong(other.literal()) + 1L;
             }
             break;
--- a/api/src/main/java/org/apache/iceberg/expressions/BoundLiteralPredicate.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/BoundLiteralPredicate.java
@@ -20,9 +20,19 @@
 package org.apache.iceberg.expressions;
 
 import java.util.Comparator;
+import java.util.Set;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.types.Type;
 
 public class BoundLiteralPredicate<T> extends BoundPredicate<T> {
+  private static final Set<Type.TypeID> INTEGRAL_TYPES = Sets.newHashSet(
+      Type.TypeID.INTEGER, Type.TypeID.LONG, Type.TypeID.DATE, Type.TypeID.TIME, Type.TypeID.TIMESTAMP);
+
+  private static long toLong(Literal<?> lit) {
+    return ((Number) lit.value()).longValue();
+  }
+
   private final Literal<T> literal;
 
   BoundLiteralPredicate(Operation op, BoundTerm<T> term, Literal<T> lit) {
@@ -74,6 +84,53 @@ public class BoundLiteralPredicate<T> extends BoundPredicate<T> {
       default:
         throw new IllegalStateException("Invalid operation for BoundLiteralPredicate: " + op());
     }
+  }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public boolean isEquivalentTo(Expression expr) {
+    if (op() == expr.op()) {
+      BoundLiteralPredicate<?> other = (BoundLiteralPredicate<?>) expr;
+      if (term().isEquivalentTo(other.term())) {
+        // because the term is equivalent, the literal must have the same type, T
+        Literal<T> otherLiteral = (Literal<T>) other.literal();
+        Comparator<T> cmp = literal.comparator();
+        return cmp.compare(literal.value(), otherLiteral.value()) == 0;
+      }
+
+    } else if (expr instanceof BoundLiteralPredicate) {
+      BoundLiteralPredicate<?> other = (BoundLiteralPredicate<?>) expr;
+      if (INTEGRAL_TYPES.contains(term().type().typeId()) && term().isEquivalentTo(other.term())) {
+        switch (op()) {
+          case LT:
+            if (other.op() == Operation.LT_EQ) {
+              // < 6 is equivalent to <= 5
+              return toLong(literal()) == toLong(other.literal()) + 1L;
+            }
+            break;
+          case LT_EQ:
+            if (other.op() == Operation.LT) {
+              // <= 5 is equivalent to < 6
+              return toLong(literal()) == toLong(other.literal()) - 1L;
+            }
+            break;
+          case GT:
+            if (other.op() == Operation.GT_EQ) {
+              // > 5 is equivalent to >= 6
+              return toLong(literal()) == toLong(other.literal()) - 1L;
+            }
+            break;
+          case GT_EQ:
+            if (other.op() == Operation.GT) {
+              // >= 5 is equivalent to > 5
+              return toLong(literal()) == toLong(other.literal()) + 1L;
+            }
+            break;
+        }
+      }
+    }
+
+    return false;
   }
 
   @Override
--- a/api/src/main/java/org/apache/iceberg/expressions/BoundTransform.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/BoundTransform.java
@@ -58,6 +58,18 @@ public class BoundTransform<S, T> implements BoundTerm<T> {
   }
 
   @Override
+  public boolean isEquivalentTo(BoundTerm<?> other) {
+    if (other instanceof BoundTransform) {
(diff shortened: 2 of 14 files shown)


api/src/main/java/org/apache/iceberg/expressions/Expression.java
```diff
@@ -124,4 +124,21 @@ public Operation flipLR() {
...
+   * If this returns true, the expressions are guaranteed to return the same evaluation for the same input. However, if
+   * this returns false the expressions may return the same evaluation for the same input. That is, expressions may
+   * be equivalent even if this returns false.
+   * <p>
+   * For best results, rewrite not and bind expressions before calling this method.
+   *
+   * @param other another expression
+   * @return true if the expressions are equivalent
+   */
+  default boolean isEquivalentTo(Expression other) {
```
aokolnychyi, line 140: In which cases we use `is` prefix for boolean methods and in which no?
We are not very consistent in the code right now.

aokolnychyi, line 140: In which cases we use `is` prefix for boolean methods and in which no?
We are not very consistent in the code right now.

rdblue: I typically try to make the method name form a natural sentence in English, like `if (someFilter.isEquivalentTo(other))` reads like "if some filter is equivalent to other". In some cases it makes more sense to use other words, like `has`.

I wasn't sure what to do with `ExpressionUtil.equivalent` since it compares two things, but `areEquivalent(f1, f2, schema)` doesn't feel right. I just left that as `ExpressionUtil.equivalent`. Suggestions are welcome!

api/src/main/java/org/apache/iceberg/expressions/BoundLiteralPredicate.java
```diff
@@ -76,6 +86,53 @@ public boolean test(T value) {
...
+            break;
+          case GT:
+            if (other.op() == Operation.GT_EQ) {
+              // > 5 is equivalent to >= 6
+              return toLong(literal()) == toLong(other.literal()) - 1L;
+            }
+            break;
+          case GT_EQ:
+            if (other.op() == Operation.GT) {
+              // >= 5 is equivalent to > 5
```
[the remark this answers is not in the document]
rdblue: Fixed.

api/src/main/java/org/apache/iceberg/expressions/BoundTransform.java
```diff
@@ -57,6 +57,18 @@ public Type type() {
     return transform.getResultType(ref.type());
   }
 
+  @Override
+  public boolean isEquivalentTo(BoundTerm<?> other) {
+    if (other instanceof BoundTransform) {
+      BoundTransform<?, ?> bound = (BoundTransform<?, ?>) other;
+      return ref.isEquivalentTo(bound.ref()) && transform.equals(((BoundTransform<?, ?>) other).transform());
```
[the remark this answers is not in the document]
rdblue: Fixed.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
```diff
@@ -66,6 +67,40 @@ public static String toSanitizedString(Expression expr) {
...
+   * If this returns true, the expressions are guaranteed to return the same evaluation for the same input. However, if
+   * this returns false the expressions may return the same evaluation for the same input. That is, expressions may
+   * be equivalent even if this returns false.
+   *
+   * @param left an unbound expression
+   * @param right an unbound expression
+   * @param struct a struct type for binding
+   * @return true if the expressions are equivalent
+   */
+  public static boolean equivalent(Expression left, Expression right, Types.StructType struct) {
```
aokolnychyi, line 82: Do we care about case sensitivity in any of the methods?

aokolnychyi, line 82: Do we care about case sensitivity in any of the methods?

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
```diff
@@ -66,6 +67,40 @@ public static String toSanitizedString(Expression expr) {
...
+   * Returns whether an expression selects whole partitions for a partition spec.
+   * <p>
+   * For example, ts &lt; '2021-03-09T10:00:00.000' selects whole partitions in an hourly spec, [hours(ts)], but does
+   * not select whole partitions in a daily spec, [days(ts)].
+   *
+   * @param expr an unbound expression
+   * @param spec a partition spec
+   * @return true if the expression will select whole partitions in the given spec
+   */
+  public static boolean selectsPartitions(Expression expr, PartitionSpec spec) {
```
aokolnychyi, line 97: Am I correct this should allow us to optimize metadata-only deletes by handling obvious cases without the need to plan files with potential matches and invoking evaluators?

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: Maybe? My thought was that since this is binding both expressions, the important thing is that we use the setting consistently. But you're right that we should probably just pass it in because the expressions may not match the struct.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: Yes.

aokolnychyi: Cool, I'll follow up with that once this is merged.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: Added.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
```diff
@@ -66,6 +67,41 @@ public static String toSanitizedString(Expression expr) {
...
+   * For example, ts &lt; '2021-03-09T10:00:00.000' selects whole partitions in an hourly spec, [hours(ts)], but does
+   * not select whole partitions in a daily spec, [days(ts)].
+   *
+   * @param expr an unbound expression
+   * @param spec a partition spec
+   * @return true if the expression will select whole partitions in the given spec
+   */
+  public static boolean selectsPartitions(Expression expr, PartitionSpec spec, boolean caseSensitive) {
+    return equivalent(
+        Projections.inclusive(spec).project(expr),
```
aokolnychyi, line 100: Shall we pass `caseSensitive` to `Projections` calls too?

aokolnychyi (approved):
Looks great to me. One question about case sensitivity when building projections.

api/src/main/java/org/apache/iceberg/expressions/ExpressionUtil.java
rdblue: Fixed.

rdblue:
Thanks for reviewing, @aokolnychyi!

wypoon:
@rdblue after I picked up this commit in master, my local build fails with
```
Execution failed for task ':iceberg-api:revapi'.
> There were Java public API/ABI breaks reported by revapi:
  
  java.method.addedToInterface: Method was added to an interface.
  
  old: <none>
  new: method boolean org.apache.iceberg.expressions.BoundTerm<T>::isEquivalentTo(org.apache.iceberg.expressions.BoundTerm<?>)
  
  SOURCE: BREAKING, BINARY: NON_BREAKING, SEMANTIC: POTENTIALLY_BREAKING
  
  From old archive: <none>
  From new archive: iceberg-api-0.14.0-SNAPSHOT.jar
  
  If this is an acceptable break that will not harm your users, you can ignore it in future runs like so for:
  
    * Just this break:
        ./gradlew :iceberg-api:revapiAcceptBreak --justification "{why this break is ok}" \
          --code "java.method.addedToInterface" \
          --new "method boolean org.apache.iceberg.expressions.BoundTerm<T>::isEquivalentTo(org.apache.iceberg.expressions.BoundTerm<?>)"
    * All breaks in this project:
        ./gradlew :iceberg-api:revapiAcceptAllBreaks --justification "{why this break is ok}"
    * All breaks in all projects:
        ./gradlew revapiAcceptAllBreaks --justification "{why this break is ok}"
  ----------------------------------------------------------------------------------------------------
```
Do you need to add something to .palantir/revapi.yml?

rdblue:
@wypoon, I merged a PR that added `BoundTerm.isEquivalentTo` earlier today: https://github.com/apache/iceberg/commit/c9efc4b8ccdaf047f32dc134096f68cb7133c545

@kbendick, can you make sure we're running the checks on PRs correctly?

wypoon:
> @wypoon, I merged a PR that added `BoundTerm.isEquivalentTo` earlier today: [c9efc4b](https://github.com/apache/iceberg/commit/c9efc4b8ccdaf047f32dc134096f68cb7133c545)
> 
Thanks. I haven't picked up #4965. I'll do so then.

Issue #4948, opened 2022-06-02
Create a Github Action to automatically mark issues as stale and later close if inactive


kbendick:
There is occasionally a backlog of issues, many of which are essentially abandoned or have been resolved but never closed.

To deal with this and ensure that eyes are kept on "fresh" / pending issues, we can use a Github Action to mark issues as `stale` (via the label) and then close them after a specified period of time. https://github.com/actions/stale

We don't want to be too aggressive, and we can allow users to mark issues as `not-stale` (or some equivalent label).

kbendick:
We need to add a label `not-stale` (per the PR, issues marked as `not-stale` will be ignored).

PR #4949, opened 2022-06-02
Infra - Add a GH Action to Mark and Close Stale Issues

kbendick:
This closes issue https://github.com/apache/iceberg/issues/4948 

Adds a Github Action to mark issues as `stale` after the configured period of time (180 days), leaving the designated comment on those issues, and then closing them after 14 days if no activity occurs after that.

Issues can be marked as `not-stale` so that they will not be considered by the stale bot action. We'll need to create this label (whatever we wind up choosing).

Github Stale Action Docs / Configuration Options: https://github.com/actions/stale
Inspired By Airflow's usage of stale action: https://github.com/apache/airflow/blob/main/.github/workflows/stale.yml

cc @danielcweeks @Fokko

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -41,7 +41,8 @@ jobs:
           stale-issue-message: >
             This issue has been automatically marked as stale because it has been open for 180 days
             with no activity. It will be closed in next 14 days if no further activity occurs. To
-            prevent this issue from being closed as stale, add the label 'not-stale'.
+            permanently prevent this issue from being considered stale, add the label 'not-stale',
+            but commenting on the issue is preferred when possible.
           close-issue-message: >
             This issue has been closed because it has not received any activity in the last 14 days
             since being marked as 'stale'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,47 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: "Close Stale Issues"
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  # All other permissions are set to none
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/stale@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/stale@v4
+        with:
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'not-stale'
+          days-before-issue-stale: 180
+          days-before-issue-close: 14
+          # Only close stale issues, leave PRs alone
+          days-before-pr-stale: -1
```
kbendick, line 40: This action is generically designed for stale PRs and issues, but I've scoped it down to just issues for now.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,47 @@
...
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/stale@v4
+        with:
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'not-stale'
```
kbendick, line 36: In addition, if an issue is part of a milestone we can also set it up to ignore those. But adding the `not-stale` label is simplest as I don't think we use milestones too much.

kbendick:
To not immediately mark all older issues as stale, we can set the configuration `start-date` to only consider issues created after a certain date. However, by default this runs in batches of 30 so I don't think we'll see massive marking of stale either way.

See https://github.com/actions/stale#start-date

Fokko (approved):
Looks good. I think we should be okay with issues older than 180 days. Thanks @kbendick

.github/workflows/stale.yml
```diff
@@ -0,0 +1,48 @@
...
```
rdblue, line 36: Where I've seen this used before, there was a label for "awaiting reply" so you could make it so issues only go stale if the person that opened the issue stops responding. Is that something we should consider?

danielcweeks: I think that behavior is largely covered, just in a slightly different way.  The issue will be marked with the `stale` label after the period of inactivity and if anyone (author or otherwise) response, the plugin will remove the `stale` label and the clock resets.

We could change the label to be `awaiting reply` but I think that is less clear in this context than `stale`.  We aren't specifically waiting for a reply and if you were to look at an issue with that label, it's wouldn't be as clear that it is considered stale.

danielcweeks (approved):

.github/workflows/stale.yml
kbendick: Yeah. I agree with Ryan’s assessment. Simply commenting is fine.

And I think personally that “awaiting-response”, unless automated to be added / removed appropriately, could cause a bit more confusion about which issues should be looked at or not.

whereas “stale” is clear that it’s been untouched for basically half a year (presently 180 days). That’s sufficient time to be able to not need to assess it further.

kbendick: And sometimes people submit issues, and GitHub only allows one person as the author / submitter, but it’s clear from the discussion that 2+ people are working on it

eg from a company maybe submitting an issue. In which case a reply from either should be sufficient to make it not stale, but that would be hard to automate.

kbendick: EDITED - Sorry said I agreed with Ryan’s assessment but I meant Dan’s assessment. Updated to reflect.

nastra (approved):

rdblue:
I think the next steps here are to bring this up on the dev list and make sure there is consensus around adding it.

rdblue (approved):

PR #4950, opened 2022-06-02
Core: Support metadata table loading in the REST catalog

rdblue:


bryanck (approved):

kbendick (approved):
This looks good to me. +1.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -209,12 +211,40 @@ private LoadTableResponse loadInternal(SessionContext context, TableIdentifier i
...
     RESTTableOperations ops = new RESTTableOperations(
         clients.first(), paths.table(identifier), session::headers, clients.second(), response.tableMetadata());
-    return new BaseTable(ops, fullTableName(identifier));
+
+    BaseTable table = new BaseTable(ops, fullTableName(identifier));
+    if (metadataType != null) {
+      return MetadataTableUtils.createMetadataTableInstance(table, metadataType);
+    }
+
+    return table;
```
karuppayya, line 247: nit: `return new BaseTable(ops, fullTableName(identifier));`
or create the object below `if.. else`

karuppayya (approved):

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
kbendick: The `table` is being used inside of `createMetadataTableInstance(table, metadataType)` regardless.

But I do somewhat agree with your style point, that it might be more readable to duplicate the line of code.

rdblue: Yeah, the intent was to either return or wrap the base table like Kyle said.

PR #4951, opened 2022-06-03
Core - Upgrade Parquet to 1.12.3

kbendick:
This patch upgrades Parquet from 1.12.2 to 1.12.3.

The change-log between the two can be found here: https://github.com/apache/parquet-mr/compare/apache-parquet-1.12.2...apache-parquet-1.12.3

A few notes of particular interest:
- [Enabled zstd buffer pool by default](https://github.com/apache/parquet-mr/pull/903)
   - Sets `parquet.compression.codec.zstd.bufferPool.enabled` to true by default
   - Some users of current Iceberg have reported that this makes working with tables with very very high cardinality (e.g. hundreds or even thousands of columns) much more memory friendly
- [PARQUET-2030 - Expose page size writer configurations](https://github.com/apache/parquet-mr/commit/8e40e69ce3e02cf4f8f0c624ff9e6173509961ee)
- [Bump Avro from 1.10.1 to 1.10.2](https://github.com/apache/parquet-mr/commit/d96b19bb97caf6f358579c9e22626553e8dc986d)
  - **We might need to upgrade this as well as we currently use 1.10.1**


rdblue:
@kbendick, what's the status of this? Should Iceberg set `parquet.compression.codec.zstd.bufferPool.enabled=true` for all Parquet files? That sounds like a good thing to me.

kbendick:
For reference, [1.12.3 is the default parquet version in Spark master](https://github.com/apache/spark/blob/c0a12cfbb82a710d786158bcd70a7f7bd531751b/pom.xml#L134), although [Spark 3.3.0 still uses Parquet 1.12.2](https://github.com/apache/spark/blob/f74867bddfbcdd4d08076db36851e88b15e66556/pom.xml#L134).

spark/v3.2/spark/src/jmh/java/org/apache/iceberg/spark/source/IcebergSourceFlatDataBenchmark.java
```diff
@@ -54,6 +54,8 @@ protected final Table initTable() {
     HadoopTables tables = new HadoopTables(hadoopConf());
     Map<String, String> properties = Maps.newHashMap();
     properties.put(TableProperties.METADATA_COMPRESSION, "gzip");
+    // Force usage of zstd compression for all Iceberg Source Flat Data Benchmarks that use Parquet
+    properties.put(TableProperties.PARQUET_COMPRESSION, "zstd");
```
rdblue, line 58: Can you remove the unnecessary changes like this? If you want to do this, then let's move it to a separate PR.

versions.props
```diff
@@ -6,7 +6,7 @@ org.apache.hadoop:* = 2.7.3
 org.apache.hive:* = 2.3.8
 org.apache.httpcomponents.client5:* = 5.1
 org.apache.orc:* = 1.7.5
-org.apache.parquet:* = 1.12.2
+org.apache.parquet:* = 1.12.3
```
rdblue, line 9: Looks like this is the only necessary change. Can you update this PR to just this change, or can you open a separate PR for this?

kbendick: Yeah absolutely.

This PR was originally for running the JMH tests.

I’ll remove all changes but the version bump and consider opening a separate PR for them if we want them.

kbendick:
We might also want to bump the Avro version to match what’s used in parquet 1.12.3.

kbendick:
> > We might also want to bump the Avro version to match what’s used in parquet 1.12.3.
> 
> Is that safe? What is the version change?

The version change is from 1.10.1 to 1.10.2. https://github.com/apache/parquet-mr/commit/d96b19bb97caf6f358579c9e22626553e8dc986d

We are on 1.10.1, so we would be following the same upgrade path.

kbendick:
> @kbendick, what's the status of this? Should Iceberg set `parquet.compression.codec.zstd.bufferPool.enabled=true` for all Parquet files? That sounds like a good thing to me.

I have spoken to some users with very high cardinality tables (potentially over 1000 columns) who have told me that enabling the buffer pool via this configuration has resolved OOMs for them, so I believe it will be beneficial to all users.

kbendick:
The Hive tests are failing due to the lack of a method.

rdblue:
Looks like Hive might be bringing in a different copy of Parquet and that is conflicting in the test. We should be able to exclude Hive's Parquet version to work around this.

kbendick:
> Looks like Hive might be bringing in a different copy of Parquet and that is conflicting in the test. We should be able to exclude Hive's Parquet version to work around this.

I'll give that a try.

kbendick:
Closed in favor of https://github.com/apache/iceberg/pull/5188, which gets this working.

PR #4955, opened 2022-06-03
Remove version.txt pointing to 0.13.2 from master

singhpk234:
Version.txt pointing to 0.13.2, recently got added to master.

As per my understanding master should not be tied to a version, and hence seems like it was added accidentally. This PR aims to remove that.

---- 

cc @RussellSpitzer


RussellSpitzer:
Yep, sorry that shouldn't have gone in, missed that

singhpk234:
Thanks @RussellSpitzer, for the quick review :) !!!

PR #4956, opened 2022-06-03
support adding extra commit metadata with SQL in Spark

CodingCat:
this PR implements the functionality for users to add extra commit metadata when operating tables with SQL. It also allows users to use multi threading to commit data to tables while keeping metadata thread local 

new usage:

```scala
(0 until 10).foreach { _ =>


    new Thread() {
      override def run() {
        CommitMetadata.withCommitProperties(Map("metadata-key" -> "thread-local-metadata-value").asJava,
          () => {
            SparkSession.getActiveSession.get.sql("INSERT INTO target VALUES (3, 'c'), (4, 'd')");
          })
      }
    }.start()

  }
}
```


spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
     Assert.assertTrue(table.currentSnapshot().summary().get("extra-key").equals("someValue"));
     Assert.assertTrue(table.currentSnapshot().summary().get("another-key").equals("anotherValue"));
   }
+
+  @Test
+  public void testExtraSnapshotMetadataWithSQL() throws IOException {
+    String tableLocation = temp.newFolder("iceberg-table").toString();
+    HadoopTables tables = new HadoopTables(CONF);
+    int threadsCount = 3;
+    ExecutorService executorService = Executors.newFixedThreadPool(threadsCount, new ThreadFactory() {
```
rdblue, line 423: Can you use the helper in `ThreadPools` and also wrap it in a try/finally to close the threadpool?

I'm also not entirely sure this requires a threadpool to test. I think it would be fine to test a single write in the current thread.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
...
+            }
+        );
+    Set<String> threadNames = new HashSet<>();
+    for (Snapshot snapshot : table.snapshots()) {
+      threadNames.add(snapshot.summary().get("writer-thread"));
+    }
+    assertTrue(threadNames.contains("thread-0"));
+    assertTrue(threadNames.contains("thread-1"));
+    assertTrue(threadNames.contains("thread-2"));
+
```
rdblue, line 472: Unnecessary newline.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
...
+              } catch (Exception e) {
+                e.printStackTrace();
+              }
+            }
+        );
+    Set<String> threadNames = new HashSet<>();
+    for (Snapshot snapshot : table.snapshots()) {
+      threadNames.add(snapshot.summary().get("writer-thread"));
+    }
+    assertTrue(threadNames.contains("thread-0"));
```
rdblue, line 469: Add empty newlines between a control flow blocks and the following statement.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
...
+    );
+    Dataset<Row> originalDf = spark.createDataFrame(expectedRecords, SimpleRecord.class);
+    originalDf.select("id", "data").write()
+        .format("iceberg")
+        .mode("append")
+        .option(SparkWriteOptions.SNAPSHOT_PROPERTY_PREFIX + ".extra-key", "someValue")
+        .option(SparkWriteOptions.SNAPSHOT_PROPERTY_PREFIX + ".another-key", "anotherValue")
+        .save(tableLocation);
+    spark.read().format("iceberg").load(tableLocation).createOrReplaceTempView("target");
+    Tasks
```
rdblue, line 447: I'd add empty lines before and after this to make the method more readable.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
...
+        .run(index -> {
+              Map<String, String> properties = Maps.newHashMap();
+              properties.put("writer-thread", String.valueOf(Thread.currentThread().getName()));
+              try {
+                CommitMetadata.withCommitProperties(properties, () -> {
+                  spark.sql("INSERT INTO target VALUES (3, 'c'), (4, 'd')");
+                  return 0;
+                });
+              } catch (Exception e) {
+                e.printStackTrace();
```
rdblue, line 461: This shouldn't print. Instead, wrap it in a `RuntimeException`.

Another option is to parameterize `withCommitProperties` so that you can specify what type of exception may be thrown. Then you can use `RuntimeException` here and not add the try/catch. An example of this is the `runSafely` method: https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java#L56

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
...
+    List<SimpleRecord> expectedRecords = Lists.newArrayList(
+        new SimpleRecord(1, "a"),
+        new SimpleRecord(2, "b")
+    );
+    Dataset<Row> originalDf = spark.createDataFrame(expectedRecords, SimpleRecord.class);
+    originalDf.select("id", "data").write()
+        .format("iceberg")
+        .mode("append")
+        .option(SparkWriteOptions.SNAPSHOT_PROPERTY_PREFIX + ".extra-key", "someValue")
+        .option(SparkWriteOptions.SNAPSHOT_PROPERTY_PREFIX + ".another-key", "anotherValue")
```
rdblue, line 444: Are these checked? If we don't need them, I think it would be better to omit them.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/CommitMetadata.java
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.iceberg.spark;
+
+import java.util.Map;
+import java.util.concurrent.Callable;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+
+/**
+ * utility class to accept thread local commit properties
+ */
+public class CommitMetadata {
```
kbendick, line 52: Does this class need to be `public`? Could it be made package-private?

I have concerns around the usage of `ThreadLocal` for things that most cases don't need to be thread local. I don't want to give users too much room to hurt themselves because they don't consider that `CommitMetadata` is _only_ threadlocal and then their writes not working properly in the common case of writes without user-side multithreading (e.g. it gets set in one thread somewhere, but another thread is used for commit).

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
kbendick: I would also prefer a test that uses a single write in the current thread without any additional threading business. I worry that `CommitMetadata` doesn't _seem_ thread local to users.

And then if a multi-threaded test is needed, using the helpers from `ThreadPools` as suggested.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/CommitMetadata.java
```diff
@@ -0,0 +1,67 @@
...
+  private static final ThreadLocal<Map<String, String>> COMMIT_PROPERTIES = ThreadLocal.withInitial(ImmutableMap::of);
+
+  public static <R> R withCommitProperties(Map<String, String> properties, Callable<R> callable) throws Exception {
+    COMMIT_PROPERTIES.set(properties);
+    try {
+      return callable.call();
+    } finally {
+      COMMIT_PROPERTIES.set(ImmutableMap.of());
+    }
+  }
```
kbendick, line 62: I guess since this requires a `Callable`, then it's not a huge deal that the class only supports ThreadLocal CommitMetadata.

But to not conflict with Structured Streaming's `CommitMetadata` (especially in Java programs that can't rename imports), it might be a good idea to update the name as mentioned above.

kbendick (commented):
Thinking this over more, there's really no way to achieve this use-case without this. So outside of the style nits Ryan mentioned, this looks pretty much good to me.

Thanks @CodingCat!

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
CodingCat: updated to ThreadPool, I think multi-threading testing is still necessary? as we need to have something guarding that the commit metadata change is thread safe no matter we use ThreadLocal as now or later we change to something else for any reason

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
```diff
@@ -404,4 +414,61 @@ public void testExtraSnapshotMetadata() throws IOException {
...
+                CommitMetadata.withCommitProperties(properties, () -> {
+                  spark.sql("INSERT INTO target VALUES (3, 'c'), (4, 'd')");
+                  return 0;
+                });
+              } catch (Exception e) {
+                e.printStackTrace();
+              }
+            }
+        );
+    Set<String> threadNames = new HashSet<>();
```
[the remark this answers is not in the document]
CodingCat: updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
CodingCat: updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
CodingCat: updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
CodingCat: updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
CodingCat: updated

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
CodingCat: updated

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/CommitMetadata.java
CodingCat: sure, changed to CallerWithCommitMetadata....but...eh...not sure if it is better or worse....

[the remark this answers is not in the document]
CodingCat: updated

CodingCat:
thanks! @rdblue and @kbendick , just updated the PR

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestDataSourceOptions.java
rdblue: I don't think multi-threaded testing is needed. It's enough to know that we're using a thread-local. This also is not guaranteed to run the way this test assumes that it will. There is not a guarantee that the thread pool will scale all the way up, and there's no guarantee that the tasks will each run in a separate thread. I think it's likely that those will happen, but this could still be a source of flakiness later on. Also, this doesn't necessarily test that the thread-local is working properly because there's no guarantee of concurrency across tasks.

While it's probably working the way you expect, there's no guarantee that it must. So I'd prefer to keep the test simple.

CodingCat: updated

rdblue (approved):

CodingCat:
just to confirm, seems we don't support SQL based table insert/merge in Spark 2.4 at all right? (so this functionality is not relevant there)

rdblue (changes requested):
What was wrong with the `CommitMetadata` name? I think we should have a simpler name than `CallerWithCommitMetadata`. That's a bit too confusing. I think the original name was good.

CodingCat:
> What was wrong with the `CommitMetadata` name? I think we should have a simpler name than `CallerWithCommitMetadata`. That's a bit too confusing. I think the original name was good.

https://github.com/apache/iceberg/pull/4956#discussion_r889199722 @kbendick mentioned some potential conflict with SS's CommitMetadata here, I don't have strong opinion on this

spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/CallerWithCommitMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+   * @param exClass the expected type of exception which would be thrown from callable
+   */
+  public static <R, E extends Exception> R withCommitProperties(
+      Map<String, String> properties, Callable<R> callable, Class<E> exClass) throws E {
+    COMMIT_PROPERTIES.set(properties);
+    try {
+      return callable.call();
+    } catch (Throwable e) {
+      ExceptionUtil.castAndThrow(e, exClass);
+      return null;
```
singhpk234, line 52: [minor] is this required as we throw in the line above ?

spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -174,6 +175,10 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
       extraSnapshotMetadata.forEach(operation::set);
     }
 
+    if (!CallerWithCommitMetadata.commitProperties().isEmpty()) {
+      CallerWithCommitMetadata.commitProperties().forEach(operation::set);
+    }
```
singhpk234, line 180: [doubt](Probably not in scope of PR) I see there is no validation happening in the keys passed from here doesn't override the keys already present in the summary, should we also add one ?

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -192,6 +193,10 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
       extraSnapshotMetadata.forEach(operation::set);
     }
 
+    if (!CallerWithCommitMetadata.commitProperties().isEmpty()) {
+      CallerWithCommitMetadata.commitProperties().forEach(operation::set);
+    }
```
singhpk234, line 198: [question] Should we also add this to SparkPositionWrite
https://github.com/apache/iceberg/blob/b06a89cebd5099b40b188c4c40ea7b1a23d3427a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java#L250


since starting 3.2 iceberg support MOR with pos deletes

spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
rdblue: I don't think that we need to worry about this. It is unlikely to conflict and if it does conflict, it's up to the caller to decide what to do.

rdblue:
@CodingCat, I think renaming the class back is about the only thing left to fix.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
rdblue: Yes, that's a good idea.

spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/CallerWithCommitMetadata.java
rdblue: I don't think the compiler sees that `castAndThrow` will always throw, so it needs this to know what to do.

kbendick:
> > What was wrong with the `CommitMetadata` name? I think we should have a simpler name than `CallerWithCommitMetadata`. That's a bit too confusing. I think the original name was good.
> 
> [#4956 (comment)](https://github.com/apache/iceberg/pull/4956#discussion_r889199722) @kbendick mentioned some potential conflict with SS's CommitMetadata here, I don't have strong opinion on this

I’m good with the original name too (and prefer it to the new one).

CodingCat:
@kbendick @rdblue @singhpk234 updated the PR accordingly, thanks!

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/CommitMetadata.java
rdblue: > Does this class need to be public? Could it be made package-private?

Yes, this does need to be public because it is a way for Iceberg users to pass metadata.

rdblue (approved):

rdblue:
Thanks, @CodingCat!

PR #4957, opened 2022-06-03
Use CatalogUtil classloader instead of context classloader for FileIO

danielcweeks:
Using the context classloader to dynamically load FileIO will cause problems in systems like Trino where the context classloader comes from a different plugin (e.g. information schema or describe table).

The classloader that loaded the `CatalogUtil` should have be the same classloader as the FileIO.


bryanck (approved):

rdblue (approved):
Thanks for catching this. We've hit problems like this in the past (with Avro reads, for example) where we want to use the classloader that loaded Iceberg itself. At least by default.

This looks good to me when tests are passing.

rdblue (approved):

PR #4958, opened 2022-06-04
Shutdown refresh token executor service during REST catalog client close

bryanck:
This PR adds a shutdown to the refresh token executor service when the REST catalog client is closed. Without shutting down the executor, refresh tasks continue to be scheduled (and fail because the REST client is closed).

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
@@ -369,6 +369,7 @@ public class RESTSessionCatalog extends BaseSessionCatalog implements Configurab
       ScheduledExecutorService service = refreshExecutor;
       this.refreshExecutor = null;
 
+      service.shutdown();
       try {
         if (service.awaitTermination(1, TimeUnit.MINUTES)) {
           LOG.warn("Timed out waiting for refresh executor to terminate");


core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -369,6 +369,7 @@ public void close() throws IOException {
       ScheduledExecutorService service = refreshExecutor;
       this.refreshExecutor = null;
 
+      service.shutdown();
```
kbendick, line 372: Good catch.

One question: do we potentially want to do this inside of a try block? [Looking at the javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdown--), it seems like the only exception it can throw is `SecurityException` which likely shouldn't happen if the executor service was started.

kbendick (approved):
Good catch @bryanck! This looks good to me overall.

kbendick (approved):
Do we potentially want to add additional calls to `shutdownNow` if the initial 1 minute timeout isn't sufficient? I found this in the javadocs usage examples of `ExecutorService`

```
// The following method shuts down an ExecutorService in two phases, first by calling
// shutdown to reject incoming tasks, and then calling shutdownNow, if necessary, to cancel
// any lingering tasks:
 
 void shutdownAndAwaitTermination(ExecutorService pool) {
   pool.shutdown(); // Disable new tasks from being submitted
   try {
     // Wait a while for existing tasks to terminate
     if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
       pool.shutdownNow(); // Cancel currently executing tasks
       // Wait a while for tasks to respond to being cancelled
       if (!pool.awaitTermination(60, TimeUnit.SECONDS))
           System.err.println("Pool did not terminate");
     }
   } catch (InterruptedException ie) {
     // (Re-)Cancel if current thread also interrupted
     pool.shutdownNow();
     // Preserve interrupt status
     Thread.currentThread().interrupt();
   }
 }
```

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
bryanck: My thought is that a try/catch block might be overkill, but I can add that if desired

kbendick: Agreed. This is sufficient for now in my opinion. I also think the example from the javadoc might be overkill (if anything the results will simply be processed and not returned to anything by the server, which happens all the time during shutdown etc).

bryanck: I'm thinking it is better to throw, so if the `shudown()` fails, it won't proceed to the `awaitTermination()`

bryanck:
The way it is now is similar to the shutdown in `RewriteDataFilesCommitManager.java`.

One minor thing is that the client is being closed before the executor is shut down, so any refresh tasks in progress will most likely fail. Awaiting termination might not be that useful. The refreshed token isn't going to be used if the catalog is being closed AFAIK. Alternatively, closing the client could come after the awaitTermination if that's a concern.

rdblue (approved):

rdblue:
> One minor thing is that the client is being closed before the executor is shut down, so any refresh tasks in progress will most likely fail.

Why not move the thread pool shutdown before closing the client?

rdblue:
Thanks, @bryanck!

PR #4960, opened 2022-06-05
delegate sql text to spark parser with passThrough statement instead of regular expression

PengleiShi:
This PR uses a better way instead of regular expression to delegate sql text to spark parser, which is used in DeltaLake and Hudi. And if it's ok,  i will do it in other version spark extensions.


rdblue (approved):

rdblue:
This looks good to me, but I'd like to hear what @aokolnychyi and @RussellSpitzer think.

RussellSpitzer:
The reason we aren't doing this currently is present in the tests that we failing on this branch :)

When we do the pass through it gives us an error that looks like the commands are completely invalid.

RussellSpitzer (changes requested):
As mentioned in the longer comment, I wouldn't want to make this change unless we can continue having useful error messages. Currently this breaks our check that parse errors are correctly reported.

rdblue:
Thanks, @RussellSpitzer! I missed that we wouldn't have nicer error messages. I agree, we should use the current approach.

PR #4963, opened 2022-06-06
Flink 1.15: Set format-version to 2 when Upsert is enabled and format-version is lt 2

hililiwei:
Set format-version to 2 when Upsert is enabled and format-version is lt 2


rdblue:
This should not be merged. Upgrading the format version used for a table can't be reversed and should not happen automatically. The user should do this explicitly.

hililiwei:
It only works when creating a table, and version 1 does not work properly when creating an UPSERT table. If this is not feasible, can we send a warning message when creating a UPSERT table but not using version 2. 

Our developers sometimes come to us because of this oversight, 😄

PR #4964, opened 2022-06-06
Python: Disable autouse of the fixtures

Fokko:
When debugging code, the fixtures get initialized when you run tests in PyCharm. This is a bit annoying since it will also hit all of the breakpoints that you have in the code

I don't think we really need to set those on autouse, we can just initialize on the first time we use them.`


rdblue:
Thanks, @Fokko!

PR #4965, opened 2022-06-06
Fix :iceberg-api:revapi failure - allow the new method

findepi:
Before the change, a build run locally with

    ./gradlew -x test -x integrationTest clean build

would fail at the :iceberg-api:revapi task.


findepi:
Per @ajantha-bhat 's comment https://apache-iceberg.slack.com/archives/C025PH0G1D4/p1654519428401729?thread_ts=1654517622.709609&cid=C025PH0G1D4, the `:iceberg-api:revapi` gets SKIPPED when run on `master branch from the `api-binary-compatibility.yml` workflow.

Example in currently last run on the master branch: https://github.com/apache/iceberg/runs/6747965003?check_suite_focus=true

perhaps same or similar problem applies to the PR builds.

ajantha-bhat:
@findepi : Thanks for the Fix. But better fix the CI task (github actions of rev api) to catch it before merging this? Else some other new PR might produce new issue and we keep on fixing it.

findepi:
> But better fix the CI task (github actions of rev api) to catch it before merging this?

Good idea, someone should look into that.

rdblue (approved):

PR #4969, opened 2022-06-06
Flink 1.15: include overwrite operation when read incremental data

hililiwei:
When data is inserted into a table where `Upsert` is enabled, stream reading cannot get the latest data.  Include Overwrite in an incremental scan.

Note, however, that only the new data (even though the same key already exists) is fetched here, and the deletion file is not included. It is not strictly a change log,  is only to be able to read the latest data.


rdblue:
I don't think we can merge this without a plan for how to handle incremental data in Flink. There's a similar effort happening for Spark and I think it makes sense to have a similar plan for how Flink is going to work. Would you like to work on a design? Maybe you and @stevenzwu could come up with the Flink design.

hililiwei:
@rdblue Yes, I'm glad to do. Does Spark have a design for reference?
@stevenzwu, what do you think of this? Or has there been some progress?

PR #4972, opened 2022-06-06
Python: Add more tests for schema_conversion

Fokko:


python/tests/utils/test_schema_conversion.py
```diff
@@ -229,3 +234,128 @@ def test_avro_list_required_record():
...
+def test_unknown_logical_type():
+    """Test raising a ValueError when converting an unknown logical type as part of an Avro schema conversion"""
+    avro_logical_type = {"type": "bytes", "logicalType": "date"}
+    with pytest.raises(ValueError) as exc_info:
+        AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+
+    assert "Unknown logical/physical type combination:" in str(exc_info.value)
+
+
+def test_logical_map_with_invalid_fields():
```
rdblue, line 343: Can you add a test for a valid logical map type?

rdblue: I guess that's covered by the test for the Iceberg metadata schema.

PR #4973, opened 2022-06-06
Python: (de)serialize using Pydantic

Fokko:
https://github.com/samuelcolvin/pydantic provides data validation and settings management using Python type-hints.

Fast and extensible, pydantic plays nicely with your linters/IDE/brain. Define how data should be in pure, canonical Python 3.7+; validate it with pydantic.

I took the basis from @samredai's PR in https://github.com/apache/iceberg/pull/3677

It allows us to easily (de)serialize the `TableMetadata` including all the nested fields (schema, fields and types). While working on this, I've also noticed some discrepancies with the spec. For example:

- Python uses `is_optional`, while the spec talks about `required`
- The `map.key` is a field in Python, while in the spec, this is a type.
- For the list, the element is a field in Python, while in the spec it is a type.

In this PR I also correct some of these discrapencies to avoid confusion among the different implementations. This PR comes with a lot of test to check proper (de)serialization of the tablemetadata, schema and types.

https://github.com/apache/iceberg/pull/3677


Fokko:
1.

PR #4976, opened 2022-06-06
Format: Cleanup language in spec ORC type notes

szehon-ho:
Clean up some rough language in the user-facing spec documentation.


rdblue:
Thanks, @szehon-ho! Good catch.

PR #4978, opened 2022-06-06
Annotate selected FileScanTask attributes as not null

findepi:
Within Iceberg API some attributes are nullable and some are not. We
should make clear to the user of the API which attributes do not require
a null check (are guaranteed to be non-null). This PR uses
`@NonNull` annotation for that purpose, but I am open to other suggestions.

cc @alexjo2144

### DIFF
--- a/api/src/main/java/org/apache/iceberg/FileScanTask.java
+++ b/api/src/main/java/org/apache/iceberg/FileScanTask.java
@@ -20,6 +20,7 @@
 package org.apache.iceberg;
 
 import java.util.List;
+import javax.annotation.Nonnull;
 import org.apache.iceberg.expressions.Expression;
 
 /**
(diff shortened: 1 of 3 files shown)


danielcweeks:
@findepi Iceberg doesn't currently annotate methods or parameters using `@NonNull` or `@NotNull` though, there are places where `@Nullable` is currently used.

I would be favor of making this more explicit, but I would prefer to use either `javax.validation.constraints.NotNull` or `javax.annotation.Nonnull` as opposed to introducing dependencies on projects like `findbugs` or related IDE or framework based annotations.

I think based on currently accepted usage, `javax.validation.constraints.NotNull` might be the best option.

findepi:
> as opposed to introducing dependencies on projects like findbugs or related IDE or framework based annotations.

i think i picked `edu.umd.cs.findbugs.annotations.NonNull` because that was readily available on the classpath, but i agree it wasn't the best choice.

AFAICT `javax.validation.constraints.NotNull` is preferred for bean validation, and not preferred for defining contracts, so  I  switched to `javax.annotation.Nonnull`. 

@danielcweeks can you please take another look?

PR #4979, opened 2022-06-06
Flink: port PR 4329 (FLIP-27 enumerator) to Flink 1.15

stevenzwu:


stevenzwu:
I confirmed that the git diff output doesn't show anything from PT #4329 : https://gist.github.com/stevenzwu/d56179d147d22ff9d54fec13366a2af7
```
git diff --no-index flink/v1.14/flink/src flink/v1.15/flink/src > flink.diff
```

PR #4987, opened 2022-06-07
Fix HTML javadoc warning

ericlgoodman:
Simple fix for the warning received when currently building:

```
IdToOrcName.java:42: warning: [UnescapedEntity] This HTML entity is invalid. Enclosing the code in this <pre>/<code> tag with a {@code } block will force Javadoc to interpret HTML literally.
 * struct -> field                      struct.field      struct.field
           ^
(see https://errorprone.info/bugpattern/UnescapedEntity)
Did you mean '* <pre><code>{@code'?
```


jackye1995:
Thanks for fixing this! Have you checked if there are any other warnings related to `UnescapedEntity`? If so I think we can fix in the same PR. Also make sure the PR title is in the format of `Area: Commit message` and be more concrete, for eaxmple `Doc: Fix UnescapedEntity HTML javadoc warning`

ericlgoodman:
Just rebuilt and found no `UnescapedEntity` warnings.

jackye1995 (approved):
thanks for the fix!

PR #4989, opened 2022-06-07
Infra - Fix API compatibility Github Action

kbendick:
The API Binary compatibility task `revapi`, was previously skipping or occasionally passing when it shouldn't be (and when it was _not_ passing for local gradle runs).

The issues were two:
1. The cache action was causing the action to be skipped.
2. The git checkout action checks out by default at a depth of 1, which is a detached head state. Therefore, the output of `git describe` or the first line of the output of `git tags` was not showing up. The action uses this tag to find the proper existing overrides and "old" version to pull from maven (currently iceberg-api:0.13.0) 

This removes the cache action, adds a step to output the `git describe` (so we can easily verify which tag is considered current, which corresponds to what's in `.palantir/revapi.yml`), and then runs the revapi task pretty aggressively (with --rerun-tasks, --no-build-cache, and --no-daemon).

We can try adding back in the `actions/cache` action as well as removing some of the aggressive flags for iceberg-api:revapi, but for now this is what I tested with in my fork and this will work. 🙂

### DIFF
--- a/.github/workflows/api-binary-compatibility.yml
+++ b/.github/workflows/api-binary-compatibility.yml
@@ -37,19 +37,13 @@ jobs:
   revapi:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-        if: github.event_name == 'pull_request'
+      - uses: actions/checkout@v3
         with:
           # fetch-depth of zero ensures that the tags are pulled in and we're not in a detached HEAD state
           # revapi depends on the tags, specifically the tag from git describe, to find the relevant override
           # in the .palantir/revapi.yml file
           #
           # See https://github.com/actions/checkout/issues/124
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.ref }}
-      - uses: actions/checkout@v2
-        if: github.event_name == 'push'
-        with:
           fetch-depth: 0
       - uses: actions/setup-java@v1
         with:
--- a/.github/workflows/api-binary-compatibility.yml
+++ b/.github/workflows/api-binary-compatibility.yml
@@ -43,6 +43,8 @@ jobs:
           # fetch-depth of zero ensures that the tags are pulled in and we're not in a detached HEAD state
           # revapi depends on the tags, specifically the tag from git describe, to find the relevant override
           # in the .palantir/revapi.yml file
+          #
+          # See https://github.com/actions/checkout/issues/124
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.ref }}
       - uses: actions/checkout@v2
--- a/.github/workflows/api-binary-compatibility.yml
+++ b/.github/workflows/api-binary-compatibility.yml
@@ -40,6 +40,9 @@ jobs:
       - uses: actions/checkout@v2
         if: github.event_name == 'pull_request'
         with:
+          # fetch-depth of zero ensures that the tags are pulled in and we're not in a detached HEAD state
+          # revapi depends on the tags, specifically the tag from git describe, to find the relevant override
+          # in the .palantir/revapi.yml file
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.ref }}
       - uses: actions/checkout@v2
@@ -51,7 +54,7 @@ jobs:
           java-version: 11
       - run: |
           echo "Using the old version tag, as per git describe, of $(git describe)";
-      - run: ./gradlew :iceberg-api:revapi --rerun-tasks --no-build-cache --no-daemon
+      - run: ./gradlew :iceberg-api:revapi --rerun-tasks
       - uses: actions/upload-artifact@v2
         if: failure()
         with:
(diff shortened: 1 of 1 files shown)


kbendick:
cc @singhpk234 @ajantha-bhat @findepi 

I tested this _exact_ configuration in my fork and it worked as seen here: https://github.com/kbendick/iceberg/runs/6780458052?check_suite_focus=true

Please see the PR summary for my explanation of the issue.

We can likely slowly add back in some of the cache layers, as I think the _primary_ issue was the fact that it was checked out in a detached HEAD state (as shown by the log output of checkout action and as discussed [in this issue](https://github.com/actions/checkout/issues/124).

However for now I would like to merge in something that I have seen work for sure and then we can make this more efficient.

kbendick:
I just checked in my fork and we can remove these flags ` --rerun-tasks, --no-build-cache, and --no-daemon`. Though `--no-daemon` might be desirable to some degree.

.github/workflows/api-binary-compatibility.yml
```diff
@@ -38,15 +38,25 @@ jobs:
...
       - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request'
+        with:
+          # fetch-depth of zero ensures that the tags are pulled in and we're not in a detached HEAD state
+          # revapi depends on the tags, specifically the tag from git describe, to find the relevant override
+          # in the .palantir/revapi.yml file
+          #
+          # See https://github.com/actions/checkout/issues/124
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
```
findepi, line 49: will this run on a merge commit between target branch and the PR?

kbendick: Yes, I believe it should. Somewhat hard to test admittedly. But the merge commit counts as a `push` against the target branch. For us, that’s typically master so keeping the default is fine.

I got this from the docs of the action and the associated issue. That’s why there’s a checkout if push step just below.

We can double check once it is merged in though. For our current uses (merging into master), it’ll be fine. I’ll double check whether the GitHub context has anything on push that we might grab as well though for the ref field.

Good question @findepi

singhpk234:
Thanks @kbendick for the change and comprehensive deep-dive, the above approach LGTM !!

[question] : based on this [article](https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-does-pull_request-affect-actionscheckout)

> that a pull_request workflow `ref` would look like refs/remotes/pull/##/merge, SHA of a pull_request workflow doesn't  matches the commit that triggered the workflow; instead the SHA of the pull_request is the resulting commit that was created from merging the base to the head.

Are we ok failing in the CI incase there is a merge conflict when merging base to the head ? Thoughts.
 (I think we should be ok, considering we need to resolve them any how before final merge.)

.github/workflows/api-binary-compatibility.yml
findepi: the point is -- i don't know why you want to specify `ref` at all, instead of relying on checkout action's default behavior.
if you did go with defaults, you wouldn't need these `if`s, so it would be simpler, so there must be a reason not to.
IMO the reason should be documented as a code-level comment (unless it's obvious, which doesn't seem to be the case)

kbendick:
> Thanks @kbendick for the change and comprehensive deep-dive, the above approach LGTM !!
> 
> [question] : Have a small question, based on this [article](https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-does-pull_request-affect-actionscheckout)
> 
> > that a pull_request workflow `ref` would look like refs/remotes/pull/##/merge, SHA of a pull_request workflow doesn't  matches the commit that triggered the workflow; instead the SHA of the pull_request is the resulting commit that was created from merging the base to the head.
> 
> Are we ok failing in the CI incase there is a merge conflict when merging base to the head ? Thoughts. (I think we should be ok, considering we need to resolve them any how before final merge.)

This has been removed @singhpk234. So we're using the default checkout branch, but just depth of 0 instead of 1. =)

kbendick:
This should be good to go now once tests pass @findepi @singhpk234

kbendick:
> Thanks @kbendick for the change and comprehensive deep-dive, the above approach LGTM !!
> 
> [question] : Have a small question, based on this [article](https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-does-pull_request-affect-actionscheckout)
> 
> > that a pull_request workflow `ref` would look like refs/remotes/pull/##/merge, SHA of a pull_request workflow doesn't  matches the commit that triggered the workflow; instead the SHA of the pull_request is the resulting commit that was created from merging the base to the head.
> 
> Are we ok failing in the CI incase there is a merge conflict when merging base to the head ? Thoughts. (I think we should be ok, considering we need to resolve them any how before final merge.)

So if I understand the article correctly, we _might _ still need to use the specific branch for `pull_request`.

I tested in my fork without it, and it was fine.

Additionally, the article states that in the worse case we'll get false positives. I'd rather that than false negatives as mentioned.

Let's go with this simpler workflow for now as I've tested it extensively in my fork and if we get false positives - then we'll catch them and deal with it then. But thanks for the article link, it's very helpful. Same applies for false negatives which we _should_ catch eventually as people run via the CLI which should always generate the proper result.

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #4991, opened 2022-06-07
Improved shutdown of token refresh executor during REST catalog close

bryanck:
This PR improves the shutdown of the refresh executor in the REST catalog client. During catalog close, all pending scheduled tasks will also be canceled in addition to shutting down the executor. It also moves the HTTP client close until after the executor shutdown to help prevent spurious errors in the scheduled tasks during shutdown.

Finally, the max refresh window was bumped from 1 min to 5 min to give a little bit more leeway for connection timeouts and such, to prevent the token from expiring before calling the refresh endpoint.


core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -83,7 +84,7 @@
 
 public class RESTSessionCatalog extends BaseSessionCatalog implements Configurable<Configuration>, Closeable {
   private static final Logger LOG = LoggerFactory.getLogger(RESTSessionCatalog.class);
-  private static final long MAX_REFRESH_WINDOW_MILLIS = 60_000; // 1 minute
+  private static final long MAX_REFRESH_WINDOW_MILLIS = 300_000; // 5 minutes
```
kbendick, line 87: 👍

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -361,15 +362,25 @@ private void scheduleTokenRefresh(
...
       ScheduledExecutorService service = refreshExecutor;
       this.refreshExecutor = null;
 
-      service.shutdown();
+      List<Runnable> tasks = service.shutdownNow();
+      tasks.forEach(task -> {
+        if (task instanceof Future) {
+          ((Future<?>) task).cancel(true);
+        }
+      });
```
kbendick, line 382: What I saw in the j[avadocs for `ExecutorService`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html) used the following code, which first calls `shutdown`, then `awaitTermination`, then if that is unsuccessful, calls `shutdownNow`, to still allow tasks some time to finish.

Do we want to consider that instead?

```java
 void shutdownAndAwaitTermination(ExecutorService pool) {
   pool.shutdown(); // Disable new tasks from being submitted
   try {
     // Wait a while for existing tasks to terminate
     if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
       pool.shutdownNow(); // Cancel currently executing tasks
       // Wait a while for tasks to respond to being cancelled
       if (!pool.awaitTermination(60, TimeUnit.SECONDS))
           System.err.println("Pool did not terminate");
     }
   } catch (InterruptedException ie) {
     // (Re-)Cancel if current thread also interrupted
     pool.shutdownNow();
     // Preserve interrupt status
     Thread.currentThread().interrupt();
   }
 }
```

I admittedly don't have a super strong opinion either way, but it is in the official javadocs and doesn't involve explicit cancelling of futures.

Otherwise, I'm not sure if the `awaitTermination` afterwards is necesssary.

kbendick (commented):
Thanks @bryanck! I think the 5 minute refresh interval is better.  Eventually we might consider configuring it.

I left some notes about possible otherways to shutdown (vs `shutdownNow` immediately with explicit future cancelling). Would like to hear your thoughts on it / what others think.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
bryanck: If we did that, we'd almost always await termination, as the next scheduled task will be pending, so we don't want to wait for it to run (it could be scheduled several minutes or more in the future). If a refresh task is actually running, letting it finish doesn't really add benefit as the response won't be used for anything once the catalog is closed.

rdblue (approved):

rdblue:
Thanks, @bryanck!

PR #4810, opened 2022-05-18
Spark: Provide Procedure for Catalog Register Procedure

RussellSpitzer:
Adds the ability to invoke a Catalog's register method via a Spark
procedure. This allows a user to create a catalog entry for a metadata.json
file which already exists but does not have a corrosponding catalog identifier.

PR #5006, opened 2022-06-09
Spark: Show metadata.json path in `DESC TABLE EXTENDED`

singhpk234:
### About Change
The change attempts to show metadata location in `DESC TABLE EXTENDED`, this adds `metadata-location` to the table properties.

### Why this change is Required

We can use this to figure out  the metadata pointer this table points to and use it to register table via [StoredProcedure](https://github.com/apache/iceberg/pull/4810) recently introduced. This could be convenient for users who *only* want to use SPARK-SQL.

### Testing

Added a UT.
Add one more UT for path validation

### DIFF
--- a/spark/v3.2/spark-runtime/src/integration/java/org/apache/iceberg/spark/SmokeTest.java
+++ b/spark/v3.2/spark-runtime/src/integration/java/org/apache/iceberg/spark/SmokeTest.java
@@ -20,7 +20,9 @@
 package org.apache.iceberg.spark;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.spark.extensions.SparkExtensionsTestBase;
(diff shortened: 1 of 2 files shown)


rdblue:
@singhpk234, is this something we want to expose?

singhpk234:
@rdblue, Considering now we have register table stored-procedure in, was thinking if there is a pure SQL way to find this property. I had a migration use case (i.e migrate hive tables to glue), had to go to glue UI to find this path and then register it via SQL and also vice-versa went to hive table params to find it and register in glue. 
```
MariaDB [hive]> SELECT * from TBLS where TBL_NAME='store_sales'
MariaDB [hive]> SELECT * FROM TABLE_PARAMS where TBL_ID = xx
```
Apologies if there exists some other SQL way to find it, I thought that having this exposed in describe table could have helped me. I agree this is not a property of the table but a state of the table. This was my rationale behind putting a pr out for this change. Would love to know your thoughts about the same.

Agree with you, there exists an alternative way for us to find this as well, so it's not like it can't be worked around :
```java
Table table = Spark3Util.loadIcebergTable(spark, tableName);
String metadataJson = ((HiveTableOperations) (((HasTableOperations) table).operations())).currentMetadataLocation();
```

rdblue:
Right now, you can use `input_file_name()` on some metadata tables to get this, but that's mostly a hack. I don't think that we want to expose this detail to users, but I could be convinced otherwise. I'm skeptical about the register table use case. Wouldn't you want to export from the current metastore so you don't have a duplicate table? That use case seems like a dangerous way to migrate.

singhpk234:
>  I don't think that we want to expose this detail to users, but I could be convinced otherwise

Agree with you, hence this https://github.com/apache/iceberg/pull/5063 could be a better way for achieving the same, @jackye1995  also suggested the same, hence opened https://github.com/apache/iceberg/pull/5063 for the same. Apologies I forgot to close the PR.

> Wouldn't you want to export from the current metastore so you don't have a duplicate table? That use case seems like a dangerous way to migrate

In the case above I wanted to have duplicate tables (1 per each catalog), was bench-marking tpc-ds perf using various catalog (hive / glue). This is a very niche use though and very dangerous as well for prod use cases.

singhpk234:
Superceded by https://github.com/apache/iceberg/pull/5063

PR #5008, opened 2022-06-09
Python: Extend the use of the singleton

Fokko:
Instead of having a specialized caching mechanism per class, we can generalize it as a singleton metaclass that we can reuse across different classes.

This moves the `Singleton` out of the `types.py` and implements it as a singleton.

Changing `Class(ABC)` to `Class(metaclass=ABCMeta)` is equivalent, but nicer since it more explicitly shows that we're setting a metaclass. From abc:
```python
class ABC(metaclass=ABCMeta):
    """Helper class that provides a standard way to create an ABC using
    inheritance.
    """
    __slots__ = ()
```

### DIFF
--- a/python/src/iceberg/expressions/base.py
+++ b/python/src/iceberg/expressions/base.py
@@ -14,14 +14,15 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from abc import ABC, abstractmethod
+from abc import ABCMeta, abstractmethod
 from enum import Enum, auto
 from functools import reduce, singledispatch
 from typing import Any, Generic, TypeVar
 
 from iceberg.files import StructProtocol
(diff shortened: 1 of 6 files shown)


python/src/iceberg/expressions/base.py
```diff
@@ -88,7 +89,7 @@ def negate(self) -> "Operation":
 }
 
 
-class Literal(Generic[T], ABC):
+class Literal(Generic[T], metaclass=ABCMeta):
```
rdblue, line 92: Why do we want the reader to know that this is changing a metaclass and not just inheriting from ABC? It seems like bringing a metaclass into the code is more clear about what's happening, but introduces a distinction that people (like me) may not understand or need.

rdblue (approved):

rdblue:
Looks fine to me. It's nice that it removes more code.

Issue #4689, opened 2022-05-03
Partition Summaries Found in Snapshot Summaries for Unpartitioned Tables


kbendick:
When writing to a table that is _not_ partitioned, if the `write.summary.partition-limit` property of the table is set to some non-zero value, partition summaries will be written still.

Here's a sample snapshot summary for an unpartitioned table that received one write. Notice that the prefix `partitions.` is included before the partition summary, which is the [changed partitions prefix](https://github.com/apache/iceberg/blob/3584c79022ec70f79b326550736b4600d249e4a2/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L54) which is usually followed by a field name and the value for that partition.

```
"summary" : {
      "operation" : "append",
      "spark.app.id" : "local-1651499999454",
      "added-data-files" : "1",
      "added-records" : "1",
      "added-files-size" : "608",
      "changed-partition-count" : "1",
      "partition-summaries-included" : "true",
      "partitions." : "added-data-files=1,added-records=1,added-files-size=608",
      "total-records" : "1",
      "total-files-size" : "608",
      "total-data-files" : "1",
      "total-delete-files" : "0",
      "total-position-deletes" : "0",
      "total-equality-deletes" : "0"
    }
```

Also of note is that if a table was previously partitioned, and then the partition field is dropped, and data is then inserted afterwards, the partition summary can have a reference to the partition field for the value `null`.

```
summary" : {
      "operation" : "append",
      "spark.app.id" : "local-1651499999454",
      "added-data-files" : "2",
      "added-records" : "2",
      "added-files-size" : "1215",
      "changed-partition-count" : "1",
      "partition-summaries-included" : "true",
      "partitions.cat=null" : "added-data-files=2,added-records=2,added-files-size=1215",
...
```

Is there some use case where we intentionally want to include partition summaries for the "global" partition of an unpartitioned table? The same data seems to be available in the rest of the summary, but maybe this was intentional.

I'm almost certain that number 2 is not intentional (though maybe that has something to do with the void transform).

cc @rdblue @RussellSpitzer @flyrain @aokolnychyi @szehon-ho @danielcweeks for visibility.

flyrain:
I'm not aware of any use case that needs partition summary when the table is no partitioned. This looks confusing to me. We may remove it.

danielcweeks:
I think both cases are erroneous.  Partition summaries should only be emitted if the current schema has partitions and the summaries should reflect that schema.

[...]

PR #5009, opened 2022-06-09
[CORE] - Fix snapshot summary generating partition summary metrics for unpartitioned tables

kbendick:
This partially fixes issue https://github.com/apache/iceberg/issues/4689

Currently, if a table has the `WRITE_PARTITION_SUMMARY_LIMIT`, `write.summary.partition-limit`, set to a non-zero value, it will generate partition summary data for a partition field with an empty string name.

This field essentially comes from the unpartitioned spec.

Notice that `partitions.` field, which is the [changed partitions prefix](https://github.com/apache/iceberg/blob/3584c79022ec70f79b326550736b4600d249e4a2/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L54), should normally have the partition field name after it, as it's a summary for that partition field.

However, for unpartitioned tables, if the partition summary limit is greater than zero, a `partitions.` field will be added for an empty string column.

Here's an example:
```json
"summary" : {
      "operation" : "append",
      "spark.app.id" : "local-1651499999454",
      "added-data-files" : "1",
      "added-records" : "1",
      "added-files-size" : "608",
      "changed-partition-count" : "1",
      "partition-summaries-included" : "true",
      "partitions." : "added-data-files=1,added-records=1,added-files-size=608",
      "total-records" : "1",
      "total-files-size" : "608",
      "total-data-files" : "1",
      "total-delete-files" : "0",
      "total-position-deletes" : "0",
      "total-equality-deletes" : "0"
    }
```

This PR removes partition summaries for commits on unpartitioned tables.

It does not handle void transforms (e.g. tables that were once partitioned but now are not).


kbendick:
After this change, here's an actual snapshot summary from Spark 3.2 for an unpartitioned table:

```json
{"spark.app.id":"local-1654805490169","added-data-files":"2","added-records":"3","added-files-size":"1286","changed-partition-count":"0","total-records":"3","total-files-size":"1286","total-data-files":"2","total-delete-files":"0","total-position-deletes":"0","total-equality-deletes":"0"}
```

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -823,16 +823,14 @@ public void testAddDeleteFilesMultipleSpecs() {
...
+    Assert.assertEquals("Should change 3 partitions", "3", summary.get(CHANGED_PARTITION_COUNT_PROP));
     Assert.assertEquals("Should add 1 data file", "1", summary.get(ADDED_FILES_PROP));
     Assert.assertEquals("Should have 4 data files", "4", summary.get(TOTAL_DATA_FILES_PROP));
     Assert.assertEquals("Should add 3 delete files", "3", summary.get(ADDED_DELETE_FILES_PROP));
     Assert.assertEquals("Should have 3 delete files", "3", summary.get(TOTAL_DELETE_FILES_PROP));
     Assert.assertEquals("Should add 3 position deletes", "3", summary.get(ADDED_POS_DELETES_PROP));
     Assert.assertEquals("Should have 3 position deletes", "3", summary.get(TOTAL_POS_DELETES_PROP));
 
-    Assert.assertTrue("Partition metrics must be correct",
-        summary.get(CHANGED_PARTITION_PREFIX).contains(ADDED_DELETE_FILES_PROP + "=1"));
```
kbendick, line 835: This partition summary field is for `"partition.": "..."` (the unpartitioned portion of the table).

As we're not writing summaries for unpartitioned fields, this needs to be removed.

We could in theory still add a "global partition" summary if partition summaries are enabled, but update it from `"partition."` to `"partition.unpartitioned"` or something.

kbendick: We could also still increment `changed-partition-count` by 1, given that the `partitions` metadata table _does_ have a row for the "global partition".

kbendick:
This changes some of the existing behavior:
- `changed-partition-count` does not include "unpartitioned" - so the number of changed partitions could potentially go down by 1 in some scenarios (or not be present at all).
  -  We can decouple the function that writes the `partition.` stats and the count of changed partitions to keep the old behavior.
- Stops writing partition summaries for unpartitioned tables (the goal of the PR). So there won't be a `partitions.` summary in the file for the unpartitioned section of the table.
  - Void partition transforms (dropped partitions) in v1 tables will still be included in partition summary stats as they transform to `alwaysNull` and where the change was made, it's not possible to differentiate between an actual partition field with a null value vs the void transform.
  - This does not affect v2 tables which don't use the void transform

The big one is the `changed-partition-count`. We can decouple the logic so that `changed-partition-count` is still 1 for the unpartitioned parts of the table when they are touched (technically at present, if a V2 table has multiple unpartitioned specs, there will only be 1 value in total added to `changed-partition-count`, instead of the value being incremented for each unpartitioned section of the table).

If anybody has a real-world use case where the original behavior is needed, I'm very open to hearing it. My goal is just to remove `partition.` partition summary fields caused by the PartitionSpec for `PartitionSpec.unpartitioned()` being defined as an empty list of fields.

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
```diff
@@ -186,7 +187,13 @@ public Map<String, String> build() {
...
       metrics.addTo(builder);
       setIf(deletedDuplicateFiles > 0, builder, DELETED_DUPLICATE_FILES, deletedDuplicateFiles);
-      Set<String> changedPartitions = partitionMetrics.keySet();
+      // Exclude the key that corresponds to PartitionSpec.unpartitioned().
+      // We might want to keep it if and only if the table has had its partition spec updated over time and the
+      // write hit two specs (unpartitioned / VOID and an actual bound partition spec).
+      Set<String> changedPartitions = partitionMetrics.keySet()
+          .stream()
+          .filter(partKey -> partKey != null && !partKey.isEmpty())
+          .collect(Collectors.toSet());
```
rdblue, line 196: Why not just add a check on the line that adds partition summaries?

```java
  if (trustPartitionMetrics && isPartitioned && changedPartitions.size() <= maxChangedPartitionsForSummaries) {
    ...
  }
```

core/src/test/java/org/apache/iceberg/TestRowDelta.java
rdblue: I think this should still be present. It is okay to have an entry for the table root in cases like this.

spark/v2.4/spark/src/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -783,7 +783,7 @@ public void testSnapshotsTable() {
             .set("summary", ImmutableMap.of(
                 "added-records", "1",
                 "added-data-files", "1",
-                "changed-partition-count", "1",
+                "changed-partition-count", "0",
```
rdblue, line 786: This depends on whether we consider the root to be a partition.

I think we should leave this as-is because there isn't a compelling reason to set this to 0 partitions, and that is a behavior change. We should suppress the root partition summary since it is identical to the table summary, but we shouldn't change this.

kbendick: I agree that the root should be considered a partition of its own. Suppressing the root partition summary is sufficient.

Will update 👍

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
kbendick: I considered that, but we don't have access to `isPartitioned` here. What we know about the table's partitioning comes from the metrics and the partition keys.

I agree that's cleaner but requires more of a refactor. I'm going to update to simply not write a summary for the root partition (i.e. unpartitioned tables).

kbendick: And as mentioned elsewhere, we _do_ want to include some of the metrics about the root partition regardless (just not the summary).

PR #5010, opened 2022-06-09
Core, API: Append to branch Impl

namrathamyske:
Issue addressed from: https://github.com/apache/iceberg/issues/3896

This PR contains the Implementation part from https://github.com/apache/iceberg/pull/4926 for only Append to Table operation.

Changes addressed:

1. Code to update to a branch.
2. Fetching the parent snapshot in the branch so as to make changes only to the specific branch.
3. Similarly fetching manifests from parent snapshot in a specific branch.


namratha2403:
@amogh-jahagirdar @rdblue Here are the implementation changes as addressed in https://github.com/apache/iceberg/pull/4926#discussion_r889296626

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +118,14 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
     return self();
   }
 
+  @Override
+  public ThisT toBranch(String branch){
+      Preconditions.checkArgument(branch != null,"branch cannot be null");
+      Preconditions.checkArgument(ops.current().ref(branch) != null, "%s is not a valid ref", branch);
```
amogh-jahagirdar, line 124: I think there should be another validation that this ref is actually a branch.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -167,8 +177,12 @@ protected void validate(TableMetadata currentMetadata) {
...
   public Snapshot apply() {
     refresh();
-    Long parentSnapshotId = base.currentSnapshot() != null ?
-        base.currentSnapshot().snapshotId() : null;
+    Long parentSnapshotId = null;
+    if(toBranch != null){
+      parentSnapshotId = base.ref(toBranch).snapshotId();
+    } else {
+      parentSnapshotId = base.currentSnapshot() != null ? base.currentSnapshot().snapshotId() : null;
+    }
```
amogh-jahagirdar, line 185: We can default assign the parentSnapshotId. I also think the SnapshotProducer can maintain the current branch itself as a field, (defaulting to main and then getting overrode in the toBranch).

namrathamyske:
@amogh-jahagirdar incorporated your feedback. Thanks for your review.

namrathamyske:
@rdblue Can you review this ?

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -89,6 +89,8 @@ public void accept(String file) {
 
   private ExecutorService workerPool = ThreadPools.getWorkerPool();
 
+  protected String toBranch = null;
```
amogh-jahagirdar, line 92: Imo, I think the name targetBranch is more clear.  it's used in a few other places as well. Also if we're assigning it, should we just assign it to main instead of null?

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +118,19 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
...
   }
 
+  @Override
+  public ThisT toBranch(String branch){
+      Preconditions.checkArgument(branch != null,"branch cannot be null");
+      if (ops.current().ref(branch) == null) {
+        SnapshotRef branchRef = SnapshotRef.branchBuilder(ops.current().currentSnapshot().snapshotId()).build();
+        TableMetadata.Builder updatedBuilder = TableMetadata.buildFrom(base);
+        updatedBuilder.setRef(branch, branchRef).build();
+      }
```
amogh-jahagirdar, line 128: Newline after the if block

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -167,8 +182,10 @@ protected void validate(TableMetadata currentMetadata) {
   @Override
   public Snapshot apply() {
     refresh();
-    Long parentSnapshotId = base.currentSnapshot() != null ?
-        base.currentSnapshot().snapshotId() : null;
+    Long parentSnapshotId = base.currentSnapshot() != null ? base.currentSnapshot().snapshotId() : null;
+    if(toBranch != null){
+      parentSnapshotId = base.ref(toBranch).snapshotId();
+    }
```
amogh-jahagirdar, line 188: Same, newline after the if block

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -282,6 +299,7 @@ protected TableMetadata refresh() {
   @Override
   public void commit() {
     // this is always set to the latest commit attempt's snapshot id.
+    String branch = toBranch == null ? SnapshotRef.MAIN_BRANCH : toBranch;
```
amogh-jahagirdar, line 302: If we set the default value of targetBranch as main, then the branch to produce snapshots to just becomes targetBranch. Then we don't need to declare a separate variable, and so we can remove this line

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,37 @@ public void testIncludedPartitionSummaryLimit() {
     String changedPartitions = table.currentSnapshot().summary().get(SnapshotSummary.CHANGED_PARTITION_COUNT_PROP);
     Assert.assertEquals("Should set changed partition count", "2", changedPartitions);
   }
+
+  @Test
+  public void testAppendToBranch() throws UnsupportedOperationException {
+    table.newFastAppend()
+            .appendFile(FILE_A)
+            .commit();
```
amogh-jahagirdar, line 486: Indentation, the continuing indentation should be 4

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,37 @@ public void testIncludedPartitionSummaryLimit() {
...
+    table.newFastAppend().toBranch("ref").appendFile(FILE_B).commit();
+    Snapshot branch = table.snapshot(table.ops().current().ref("ref").snapshotId());
+    Assert.assertEquals(currSnapshot, branch.parentId());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testAppendToNullBranch() {
+    table.newFastAppend()
+            .appendFile(FILE_A)
+            .commit();
```
amogh-jahagirdar, line 499: Ditto on indentation

core/src/test/java/org/apache/iceberg/TestFastAppend.java
```diff
@@ -478,4 +478,37 @@ public void testIncludedPartitionSummaryLimit() {
...
+
+    table.manageSnapshots().createBranch("ref", table.currentSnapshot().snapshotId()).commit();
+    table.newDelete().toBranch(null).deleteFile(FILE_A);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testAppendToInValidBranch() {
+    table.newFastAppend()
+            .appendFile(FILE_A)
+            .commit();
```
amogh-jahagirdar, line 509: Ditto on indentation

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
namrathamyske: Agree with your suggestion. I have defaulted targetBranch to main and just changed the way we access snapshot by ref.

namrathamyske:
@amogh-jahagirdar Thanks for your review ! Incorporated your changes.

namrathamyske:
@rdblue @amogh-jahagirdar I have added toBranch method only for Append operations as per suggestion in https://github.com/apache/iceberg/pull/4926#discussion_r889296626. For other operations it will throw an **UnsupportedOperationException**. Please let me know if this looks good

namrathamyske:
@rdblue @amogh-jahagirdar After appending to a branch, The sequence Number for main ref is continued from last number in branch ref was left off. Should I take the sequence number from last last commit in main ref or branch ref ? Any thoughts appreciated.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -116,6 +117,38 @@ public ThisT scanManifestsWith(ExecutorService executorService) {
     return self();
   }
 
+  @Override
+  public ThisT toBranch(String branch) {
+    throw new UnsupportedOperationException("Performing operations on a branch is currently not supported");
```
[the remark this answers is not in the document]
namrathamyske: @dimas-b , In the PR i have implemented the `toBranch` only in one of the subclasses of Snapshot producer aka FastAppend class. This is done in order to avoid creating a huge PR and be thorough in testing each of the child operations of snapshot producer. In future my aim is to incrementally raise PR for each of them.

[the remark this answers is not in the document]
dimas-b: Thanks for the clarification, @namrathamyske ! Would you mind updating the message to something like `Performing [SnapshotProducer operations on a branch is currently not supported` to avoid making a (false) impression that it is not supported in "general"?.. or use different wording as you see fit. My only concern is that getting the old message in one context might be interpreted that it is not supported anywhere.

core/src/main/java/org/apache/iceberg/FastAppend.java
```diff
@@ -114,6 +114,21 @@ public FastAppend appendManifest(ManifestFile manifest) {
     return this;
   }
 
+  @Override
+  public FastAppend toBranch(String branch) {
+    Preconditions.checkArgument(branch != null, "branch cannot be null");
```
rdblue, line 119: Message should be `Invalid branch name: null`

core/src/main/java/org/apache/iceberg/FastAppend.java
```diff
@@ -114,6 +114,21 @@ public FastAppend appendManifest(ManifestFile manifest) {
...
+    if (ops.current().ref(branch) == null) {
+      super.createNewRef(branch);
```
rdblue, line 121: This should be handled by `SnapshotProducer`, as I noted on #4926.

core/src/main/java/org/apache/iceberg/FastAppend.java
```diff
@@ -114,6 +114,21 @@ public FastAppend appendManifest(ManifestFile manifest) {
...
+    }
+
+    Preconditions.checkArgument(ops.current()
+            .ref(branch).type()
+            .equals(SnapshotRefType.BRANCH),
```
rdblue, line 126: Does this check need to be split across 3 lines? Or is that autoformatting doing it?

core/src/main/java/org/apache/iceberg/FastAppend.java
```diff
@@ -114,6 +114,21 @@ public FastAppend appendManifest(ManifestFile manifest) {
...
+        "%s is not a ref to type branch", branch);
```
amogh-jahagirdar, line 127: Also now that we're creating a branch on commit, we should basically do a null check. It's okay if the branch is null or if it's not null, it must be a ref of type branch.

PR #5011, opened 2022-06-09
Python: Use Pydantic for (de)serialization

Fokko:
https://github.com/samuelcolvin/pydantic provides data validation and settings management using Python type-hints.

Fast and extensible, pydantic plays nicely with your linters/IDE/brain. Define how data should be in pure, canonical Python 3.7+; validate it with pydantic.

I took the basis from @samredai's PR in https://github.com/apache/iceberg/pull/3677

It allows us to easily (de)serialize the TableMetadata including all the nested fields (schema, fields, and types).


python/src/iceberg/utils/iceberg_base_model.py
```diff
@@ -0,0 +1,29 @@
...
+
+class IcebergBaseModel(BaseModel):
+    class Config:
+        allow_population_by_field_name = True
+
+    def dict(self, exclude_none=True, **kwargs):
+        return super().dict(exclude_none=exclude_none, **kwargs)
+
+    def json(self, exclude_none=True, by_alias=True, **kwargs):
+        return super().json(exclude_none=exclude_none, by_alias=True, **kwargs)
```
rdblue, line 29: What is this implementation changing? Wouldn't the superclass's `json` method be called with these args? Is this excluding allowed arguments?

python/src/iceberg/utils/iceberg_base_model.py
```diff
@@ -0,0 +1,29 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from pydantic import BaseModel
+
+
+class IcebergBaseModel(BaseModel):
```
rdblue, line 21: Is there a better name for this? How about `IcebergObject`? I don't think `BaseModel` is a clear name that we want to carry forward, unless this is a convention that I'm missing. To me, `IcebergObject` does signal that it is layering in some common functionality.

python/src/iceberg/schema.py
```diff
@@ -134,7 +138,7 @@ def _lazy_id_to_accessor(self) -> dict[int, Accessor]:
 
     def as_struct(self) -> StructType:
         """Returns the underlying struct"""
-        return self._struct
+        return StructType(*self.fields)
```
rdblue, line 141: I think I would opt to keep an inner struct like this did before, and then define `fields` to return `this.struct.fields`

python/src/iceberg/types.py
```diff
@@ -59,134 +62,176 @@ class IcebergType:
...
+    @classmethod
+    def validate(cls, v):
+        # When Pydantic is unable to determine the subtype
+        # In this case we'll help pydantic a bit by parsing the
+        # primitive type ourselves, or pointing it at the correct
+        # complex type by looking at the type field
+
+        if isinstance(v, str):
+            if v.startswith("decimal"):
+                m = re.search(r"decimal\((\d+),\s*(\d+)\)", v)
```
rdblue, line 81: Can you create a constant for the regex instead of compiling it every time?

python/src/iceberg/types.py
```diff
@@ -59,134 +62,176 @@ class IcebergType:
...
-        return self.string_type
+    @classmethod
+    def __get_validators__(cls):
+        # one or more validators may be yielded which will be called in the
+        # order to validate the input, each validator will receive as an input
+        # the value returned from the previous validator
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
```
rdblue, line 73: Why is the parsing done in the validate method?

python/src/iceberg/types.py
```diff
@@ -59,134 +62,176 @@ class IcebergType:
...
 
 
-@dataclass(frozen=True, eq=True)
 class PrimitiveType(IcebergType):
-    """Base class for all Iceberg Primitive Types
+    """Base class for all Iceberg Primitive Types"""
 
-    Example:
-        >>> str(PrimitiveType())
-        'PrimitiveType()'
```
rdblue, line 80: Why remove this example? Is it no longer accurate?

python/src/iceberg/types.py
```diff
@@ -59,134 +62,176 @@ class IcebergType:
...
         >>> FixedType(8)
         FixedType(length=8)
         >>> FixedType(8) == FixedType(8)
         True
+        >>> FixedType(19) == FixedType(25)
+        False
     """
 
-    length: int = field()
+    __root__: str = Field()
```
rdblue, line 132: What is `__root__`?

python/src/iceberg/types.py
```diff
@@ -59,134 +62,176 @@ class IcebergType:
...
+        return self._scale
+
+    def __repr__(self) -> str:
+        return f"DecimalType(precision={self._precision}, scale={self._scale})"
 
 
-@dataclass(frozen=True)
 class NestedField(IcebergType):
     """Represents a field of a struct, a map key, a map value, or a list element.
-
```
rdblue, line 137: Can you add the whitespace back? Or is this required by formatting?

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+
+from iceberg.schema import Schema
+from iceberg.utils.iceberg_base_model import IcebergBaseModel
+
+
+class TableMetadataCommonFields(IcebergBaseModel):
+    """Metadata for an Iceberg table as specified in the Apache Iceberg
+    spec (https://iceberg.apache.org/spec/#iceberg-table-spec)"""
+
+    table_uuid: str = Field(alias="table-uuid")
```
rdblue, line 30: Should this be a UUID class rather than a string? The Avro code uses a UUID: https://github.com/apache/iceberg/pull/4920/files#diff-cf9ad84745a2a233b007a20fc7c8fa1862de0d38a93aa7f89166114624e045b1R136

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    """A UUID that identifies the table, generated when the table is created.
```
rdblue, line 31: Minor: Why are these doc strings after the fields? Should this be passed to `Field`?

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+     in manifest files."""
+
+
+class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
+    format_version: Literal[1] = Field(alias="format-version")
+    """An integer version number for the format. Currently, this can be 1 or 2
+    based on the spec. Implementations must throw an exception if a table’s
+    version is higher than the supported version."""
+
+    schema_: Schema = Field(alias="schema")
```
rdblue, line 112: Does this correctly translate?

When reading, we want to accept a `schema`, but convert it into a `schemas` list and `current-schema-id`. In v1, we want to write both `schema` and `schemas` / `current-schema-id` because adding that metadata is compatible.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    """An integer version number for the format. Currently, this can be 1 or 2
+    based on the spec. Implementations must throw an exception if a table’s
+    version is higher than the supported version."""
+
+    last_sequence_number: int = Field(alias="last-sequence-number")
+    """The table’s highest assigned sequence number, a monotonically
+    increasing long that tracks the order of snapshots in a table."""
+
+
+class TableMetadata:
```
rdblue, line 135: Why is this distinct from `TableMetadataCommenFields`? It seems like all of the table metadata classes should be `TableMetadata`.

python/tests/catalog/test_base.py
```diff
@@ -157,7 +158,7 @@ def update_namespace_properties(
 TEST_TABLE_IDENTIFIER = ("com", "organization", "department", "my_table")
 TEST_TABLE_NAMESPACE = ("com", "organization", "department")
 TEST_TABLE_NAME = "my_table"
-TEST_TABLE_SCHEMA = Schema(schema_id=1)
+TEST_TABLE_SCHEMA = Schema(NestedField(1, "foo", StringType(), True), schema_id=1)
```
rdblue, line 161: Is it still okay to have an empty table schema?

python/tests/table/test_metadata.py
```diff
@@ -0,0 +1,328 @@
...
+@pytest.mark.parametrize(
+    "metadata",
+    [
+        EXAMPLE_TABLE_METADATA_V1,
+        EXAMPLE_TABLE_METADATA_V2,
+    ],
+)
+def test_from_dict(metadata: dict):
+    """Test initialization of a TableMetadata instance from a dictionary"""
+    TableMetadata.parse_obj(metadata)
```
rdblue, line 167: Is this just testing that an object is produced? Shouldn't it validate the contents?

python/tests/test_schema.py
```diff
@@ -399,4 +399,20 @@ def get(self, pos: int) -> Any:
 
     accessors = build_position_accessors(table_schema_nested)
     container = TestStruct({6: TestStruct({0: "name"})})
-    assert accessors.get(16).get(container) == "name"
+    inner_accessor = accessors.get(16)
+    assert inner_accessor
+    assert inner_accessor.get(container) == "name"
+
+
+def test_serialize_schema(table_schema_simple: Schema):
```
rdblue, line 407: Can you add some tests for nested types (structs, maps, and arrays) as well?

python/tests/test_types.py
```diff
@@ -204,3 +206,388 @@ def test_non_parameterized_type_equality(input_index, input_type, check_index, c
...
+def test_str_binary():
+    assert str(BinaryType()) == "binary"
+
+
+def test_repr_binary():
+    assert repr(BinaryType()) == "BinaryType()"
+
+
+def test_serialization_decimal():
+    assert DecimalType(19, 25).json() == '"decimal(19, 25)"'
```
rdblue, line 432: Isn't this type invalid because the scale is greater than the precision? (I think we check that in Java)

python/tests/test_types.py
```diff
@@ -204,3 +206,388 @@ def test_non_parameterized_type_equality(input_index, input_type, check_index, c
...
+                "type": "string",
+                "required": true,
+                "doc": "this is a doc"
+            },
+            {
+                "id": 2,
+                "name": "optional_field",
+                "type": "int",
+                "required": true,
+                "doc": null
```
rdblue, line 527: Minor: Is it possible to omit the field rather than making it explicitly null?

python/tests/test_types.py
```diff
@@ -204,3 +206,388 @@ def test_non_parameterized_type_equality(input_index, input_type, check_index, c
...
+        """{"type": "map", "key-id": 19, "key": "string", "value-id": 25, "value": "double", "value-required": false}"""
+    )
+    assert actual == simple_map
+
+
+def test_str_map(simple_map: MapType):
+    assert str(simple_map) == "map<string, double>"
+
+
+def test_repr_map(simple_map: MapType):
```
rdblue, line 589: It would be nice to have a more complex type that mixes maps, lists, and structs within one another, but I think this is good overall.

rdblue (changes requested):
This looks great overall! I really like how quickly this is able to implement the JSON serialization.

My main concern about this approach is the `TableMetadata` handling. We need to read into a common representation that works for both v1 and v2, and serialize to the requirements of each version. I don't think that's quite working now, but I may just understand.

python/src/iceberg/utils/iceberg_base_model.py
Fokko: In the superclass, the `by_alias` is set to False by default. But for Iceberg it is better to set this to `True` all the time because we rely on aliases when we convert JSON we always want to use the aliases (`schema-id` instead of `schema_id`, since `-` is an operator in Python). The same goes for excluding non-values in the case of the doc element of the NestedField that we don't want to include if it is `None`.

By overriding this function, we set these values by default to True.

python/src/iceberg/utils/iceberg_base_model.py
Fokko: I'm also okay with `IcebergObject`, but that might also be confusing since it is much more than the plain Python object. I don't think this is a Python convention, but I do like calling it a model because it connects it to Pydantic. It has much more than an object, for example, it has a schema: https://pydantic-docs.helpmanual.io/usage/schema/

python/src/iceberg/types.py
Fokko: Done!

python/src/iceberg/types.py
Fokko: Good one, I've moved this into the Types itself.

python/src/iceberg/types.py
Fokko: I've converted into an abstract class that doesn't have a `__root__` defined, and therefore no representation. I'm not sure when we would want to instantiate the PrimitiveType, therefore I removed the example.

python/src/iceberg/types.py
Fokko: This is something special to Pydantic: https://pydantic-docs.helpmanual.io/usage/models/#custom-root-types

In most cases, a class converts into a JSON dict. Using the `__root__` we can set it to a string (or a int, etc)

python/src/iceberg/table/metadata.py
Fokko: We can pass it to the field, but that doesn't add any value right now, until we start exporting the models (which I don't see happening any time soon :)

python/tests/catalog/test_base.py
Fokko: This was during testing. This wasn't possible for some time because of all the changes I made to the code, but since this is initialized before the test, it would crash right away before hitting any breakpoints. I'll revert this.

python/tests/table/test_metadata.py
Fokko: This is just checking the logic. This will be obsolete once https://github.com/samuelcolvin/pydantic/pull/3847 has been merged :)

python/tests/test_types.py
Fokko: Good one, I've created an issue https://github.com/apache/iceberg/issues/5038

python/tests/test_types.py
Fokko: This is a deserialization input, so that should be possible. I've updated the code.

python/src/iceberg/table/metadata.py
Fokko: This is actually quite nice of Pydantic. The default json dumps doesn't allow us to encode a uuid into json:
```python
➜  python git:(fd-read-avro) python3     
Python 3.9.13 (main, May 19 2022, 13:48:47) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from uuid import UUID
>>> example = UUID("f79c3e09-677c-4bbd-a479-3f349cb785e7")
>>> import json
>>> json.dumps({"uuid": example})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Cellar/python@3.9/3.9.13/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type UUID is not JSON serializable
```

python/src/iceberg/schema.py
Fokko: I tried this, but that creates a lot of awkward code since it serializes as fields. I can cache it on the first call, wdyt?

python/src/iceberg/table/metadata.py
Fokko: schema is a reserved keyword in Pydantic:

print(Model.schema())
```python
>>> table_metadata.schema()
{
	'title': 'TableMetadataV2',
	'description': 'Metadata for an Iceberg table as specified in the Apache Iceberg\nspec (https://iceberg.apache.org/spec/#iceberg-table-spec)',
	'type': 'object',
	'properties': {
		'table-uuid': {
			'title': 'Table-Uuid',
			'type': 'string',
			'format': 'uuid'
		},
		'location': {
			'title': 'Location',
			'type': 'string'
		},
		'last-updated-ms': {
			'title': 'Last-Updated-Ms',
			'type': 'integer'
		},
		'last-column-id': {
			'title': 'Last-Column-Id',
			'type': 'integer'
		},
		'schemas': {
			'title': 'Schemas',
			'type': 'array',
			'items': {
				'$ref': '#/definitions/Schema'
			}
		},
		'current-schema-id': {
			'title': 'Current-Schema-Id',
			'type': 'integer'
		},
		'partition-specs': {
			'title': 'Partition-Specs',
			'type': 'array',
			'items': {}
		},
		'default-spec-id': {
			'title': 'Default-Spec-Id',
			'type': 'integer'
		},
		'last-partition-id': {
			'title': 'Last-Partition-Id',
			'type': 'integer'
		},
		'properties': {
			'title': 'Properties',
			'type': 'object'
		},
		'current-snapshot-id': {
			'title': 'Current-Snapshot-Id',
			'type': 'integer'
		},
		'snapshots': {
			'title': 'Snapshots',
			'type': 'array',
			'items': {}
		},
		'snapshot-log': {
			'title': 'Snapshot-Log',
			'type': 'array',
			'items': {}
		},
		'metadata-log': {
			'title': 'Metadata-Log',
			'type': 'array',
			'items': {}
		},
		'sort-orders': {
			'title': 'Sort-Orders',
			'type': 'array',
			'items': {}
		},
		'default-sort-order-id': {
			'title': 'Default-Sort-Order-Id',
			'type': 'integer'
		},
		'format-version': {
			'title': 'Format-Version',
			'enum': [2],
			'type': 'integer'
		},
		'last-sequence-number': {
			'title': 'Last-Sequence-Number',
			'type': 'integer'
		}
	},
	'required': ['table-uuid', 'location', 'last-updated-ms', 'last-column-id', 'schemas', 'current-schema-id', 'partition-specs', 'default-spec-id', 'last-partition-id', 'properties', 'current-snapshot-id', 'snapshots', 'snapshot-log', 'metadata-log', 'sort-orders', 'default-sort-order-id', 'format-version', 'last-sequence-number'],
	'definitions': {
		'IcebergType': {
			'title': 'IcebergType',
			'description': "Base type for all Iceberg Types\n\nExample:\n    >>> str(IcebergType())\n    'IcebergType()'\n    >>> repr(IcebergType())\n    'IcebergType()'",
			'type': 'object',
			'properties': {}
		},
		'NestedField': {
			'title': 'NestedField',
			'description': 'Represents a field of a struct, a map key, a map value, or a list element.\n\nThis is where field IDs, names, docs, and nullability are tracked.\nExample:\n    >>> str(NestedField(\n    ...     field_id=1,\n    ...     name=\'foo\',\n    ...     field_type=FixedType(22),\n    ...     required=False,\n    ... ))\n    \'1: foo: optional fixed[22]\'\n    >>> str(NestedField(\n    ...     field_id=2,\n    ...     name=\'bar\',\n    ...     field_type=LongType(),\n    ...     is_optional=False,\n    ...     doc="Just a long"\n    ... ))\n    \'2: bar: required long (Just a long)\'',
			'type': 'object',
			'properties': {
				'id': {
					'title': 'Id',
					'type': 'integer'
				},
				'name': {
					'title': 'Name',
					'type': 'string'
				},
				'type': {
					'$ref': '#/definitions/IcebergType'
				},
				'required': {
					'title': 'Required',
					'default': True,
					'type': 'boolean'
				},
				'doc': {
					'title': 'Doc',
					'type': 'string'
				}
			},
			'required': ['id', 'name', 'type']
		},
		'Schema': {
			'title': 'Schema',
			'description': 'A table Schema\n\nExample:\n    >>> from iceberg import schema\n    >>> from iceberg import types',
			'type': 'object',
			'properties': {
				'fields': {
					'title': 'Fields',
					'type': 'array',
					'items': {
						'$ref': '#/definitions/NestedField'
					}
				},
				'schema-id': {
					'title': 'Schema-Id',
					'type': 'integer'
				},
				'identifier-field-ids': {
					'title': 'Identifier-Field-Ids',
					'type': 'array',
					'items': {
						'type': 'integer'
					}
				}
			},
			'required': ['schema-id']
		}
	}
}
"""

Therefore we need to set the alias to properly map it.

I've added code to properly map it. In v1 the schema's are optional, if it is set, I'll check if the schema is in there, otherwise, I'll add it.

File: python/src/iceberg/table/metadata.py (hunk 12 above)
Comment (Fokko) [reply python/src/iceberg/table/metadata.py:135]: Once https://github.com/samuelcolvin/pydantic/issues/3846 has been resolved, we can replace the TableMetadata with:

```python
TableMetadata = Annotated[Union[TableMetadataV1, TableMetadataV2], Field(alias="format-version", discriminator="format-version")]
```

And this is very nice because then we can just remove the boilerplate, and everything is done by pydantic
```python
TableMetadata.parse_raw("{'file-format': 1, ...}")
TableMetadata. parse_obj({'file-format': 1, ...})
```

python/src/iceberg/schema.py
```diff
@@ -343,23 +339,23 @@ def _(obj: StructType, visitor: SchemaVisitor[T]) -> T:
...
     return visitor.list(obj, result)
 
 
 @visit.register(MapType)
 def _(obj: MapType, visitor: SchemaVisitor[T]) -> T:
     """Visit a MapType with a concrete SchemaVisitor"""
-    visitor.before_map_key(obj.key)
-    key_result = visit(obj.key.field_type, visitor)
-    visitor.after_map_key(obj.key)
+    visitor.before_map_key(obj.key_field)
```
Fokko, line 352: We have to change these since we call them by the dataclass alias, but with Pydantic you need to call them by the variable name, which I like since it is more explicit.

python/src/iceberg/schema.py
rdblue: No worries, we can keep it as is for now.

python/src/iceberg/utils/iceberg_base_model.py
rdblue: Thanks for explaining! That makes sense. Can you make sure that this is clear in a comment? That way future readers will also know.

Also, should the `by_alias` passed to the method be passed to the `super().json(...)` call? Right now it ignores it.

python/src/iceberg/schema.py
```diff
@@ -16,18 +16,22 @@
 # under the License.
 # pylint: disable=W0511
 
-from __future__ import annotations
```
rdblue, line 19: Is this incompatible with Pydantic?

python/src/iceberg/serializers.py
```diff
@@ -0,0 +1,75 @@
...
+class FromByteStream:
+    """A collection of methods that deserialize dictionaries into Iceberg objects"""
+
+    @staticmethod
+    def table_metadata(byte_stream: InputStream, encoding: str = "utf-8") -> TableMetadata:
+        """Instantiate a TableMetadata object from a byte stream
+
+        Args:
+            byte_stream: A file-like byte stream object
+            encoding (default "utf-8"): The byte encoder to use for the reader
```
rdblue, line 35: I don't think that we allow any other encoding, but it should be fine to leave this in.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,168 @@
...
+    default_sort_order_id: int = Field(alias="default-sort-order-id")
+    """Default sort order id of the table. Note that this could be used by
+    writers, but is not used when reading because reads use the specs stored
+     in manifest files."""
+
+
+class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
+    def __new__(cls, *_, **data):
+        # When we read a V1 format-version, we'll bump it to a V2 table right
+        # away by populating the required fields, and setting the version
```
rdblue, line 112: We shouldn't actually change the format of the table. In the library, we always want to work with v2 structures. But we still need to maintain compatibility with older readers and writers, so upgrading a table from v1 to v2 must be done by users.

What we want to do here is deserialize to a common representation so that the rest of the library doesn't need to worry about v1/v2. Then when writing metadata, this should serialize according to the version. For example, if the table is v1, then write `schema` in addition to `schemas` and `current-schema-id`. If the table version is v2, then omit `schema`.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,168 @@
...
+
+    partition_spec: dict = Field(alias="partition-spec")
+    """The table’s current partition spec, stored as only fields.
+    Note that this is used by writers to partition data, but is
+    not used when reading because reads use the specs stored in
+    manifest files. (Deprecated: use partition-specs and default-spec-id
+    instead)"""
+
+
+class TableMetadataV2(TableMetadataCommonFields, IcebergBaseModel):
```
rdblue, line 141: For serializing to JSON, I think it makes sense to have separate classes. I don't see another way to suppress fields like `schema` or `last-sequence-number`.

python/src/iceberg/types.py
rdblue: Looks like this is still here?

python/src/iceberg/types.py
```diff
@@ -48,107 +57,189 @@ class IcebergType(ABC, Singleton):
...
+    def __repr__(self) -> str:
+        return f"DecimalType(precision={self._precision}, scale={self._scale})"
 
 
-@dataclass(frozen=True)
 class NestedField(IcebergType):
     """Represents a field of a struct, a map key, a map value, or a list element.
 
     This is where field IDs, names, docs, and nullability are tracked.
-
```
rdblue, line 115: Accidental change?

python/tests/conftest.py
```diff
@@ -301,3 +319,81 @@ def all_avro_types() -> Dict[str, Any]:
...
+
+    def new_output(self, location: str):
+        return LocalOutputFile(location=location)
+
+    def delete(self, location: Union[str, InputFile, OutputFile]):
+        location = location.location if isinstance(location, (InputFile, OutputFile)) else location
+        try:
+            os.remove(location)
+        except FileNotFoundError as e:
+            raise FileNotFoundError(f"Cannot delete file, does not exist: {location}") from e
```
rdblue, line 394: If the file doesn't exist, why not just return? It doesn't seem worth throwing an exception.

python/tests/table/test_metadata.py
```diff
@@ -0,0 +1,279 @@
...
+        "schema-id": 0,
+        "fields": [
+            {"id": 1, "name": "foo", "required": True, "type": "string"},
+            {"id": 2, "name": "bar", "required": True, "type": "string"},
+            {"id": 3, "name": "baz", "required": True, "type": "string"},
+            {"id": 4, "name": "qux", "required": True, "type": "string"},
+        ],
+        "identifier-field-ids": [],
+    },
+    "schemas": [
```
rdblue, line 45: I would recommend testing v1 without these newer structures to ensure that it is correctly defaulting them. Then when you test writing v1 metadata, make sure they are added.

python/tests/table/test_metadata.py
```diff
@@ -0,0 +1,279 @@
...
+                {"id": 1, "name": "foo", "required": True, "type": "string"},
+                {"id": 2, "name": "bar", "required": True, "type": "string"},
+                {"id": 3, "name": "baz", "required": True, "type": "string"},
+                {"id": 4, "name": "qux", "required": True, "type": "string"},
+            ],
+            "identifier-field-ids": [],
+        },
+    ],
+    "current-schema-id": 0,
+    "partition-spec": {},
```
rdblue, line 58: In v1, this should be a list. Initially, this stored only the partition fields because we didn't support multiple partition specs. See https://iceberg.apache.org/spec/#table-metadata

python/tests/table/test_metadata.py
```diff
@@ -0,0 +1,279 @@
...
+            "schema-id": 0,
+        },
+    ],
+    "snapshot-log": [
+        {"timestamp-ms": 1637943123188, "snapshot-id": 7681945274687743099},
+    ],
+    "metadata-log": [
+        {
+            "timestamp-ms": 1637943123331,
+            "metadata-file": "3://foo/bar/baz/00000-907830f8-1a92-4944-965a-ff82c890e912.metadata.json",
```
rdblue, line 151: `s3`?

python/src/iceberg/utils/iceberg_base_model.py
Fokko: I missed your previous comments, sorry about that!

> Also, should the by_alias passed to the method be passed to the super().json(...) call? Right now it ignores it.

Thanks! That was a copy-paste error.

> I've added the following comment:

```
This class extends the Pydantic BaseModel to set default values by overriding them.

This is because we always want to set by_alias to True. In Python, the dash can't
be used in variable names, and this is used throughout the Iceberg spec.

The same goes for exclude_none, if a field is None we want to omit it from
serialization, for example, the doc attribute on the NestedField object.
Default non-null values will be serialized.

This is recommended by Pydantic:
https://pydantic-docs.helpmanual.io/usage/model_config/#change-behaviour-globally
```

python/src/iceberg/schema.py
Fokko: There is a bug in Pydantic in combination with Python 3.8. Somewhere in a type reference of NestedField something is off. I want to dig into this, but I first wanted to get this PR ready to get feedback on it.

python/src/iceberg/table/metadata.py
Fokko: Got it! I just updated the code, and the comment 👍🏻

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,168 @@
...
+    """A list of sort orders, stored as full sort order objects."""
+
+    default_sort_order_id: int = Field(alias="default-sort-order-id")
+    """Default sort order id of the table. Note that this could be used by
+    writers, but is not used when reading because reads use the specs stored
+     in manifest files."""
+
+
+class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
+    def __new__(cls, *_, **data):
```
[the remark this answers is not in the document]
Fokko: Based on the comment above, I'd prefer to use a Pydantic validator:

```python
@root_validator(pre=True)
def populate_schemas(cls, data: Dict[str, Any]):
    # When we read a V1 format-version, we'll make sure to populate the fields
    # for V2 as well. This makes it easier downstream because we can just
    # assume that everything is a TableMetadataV2.
    # When writing, we should stick to the same version that it was,
    # because bumping the version should be an explicit operation that is up
    # to the owner of the table.
    schema = data["schema"]
    if "schemas" not in data:
        if all([schema != other_schema for other_schema in data["schemas"]]):
            data["schemas"].append(schema)
    else:
        data["schemas"] = [schema]
    data["current-schema-id"] = schema["schema-id"]
    data["last-sequence-number"] = _INITIAL_SEQUENCE_NUMBER
    return data
```

python/src/iceberg/table/metadata.py
Fokko: Pydantic will check which version it is, and serialize the fields that are bound to the version. Of course, for V1 the schemas field is optional, and those will be set because we populate them when we initialize a class (the root_validator a comment above).

The following will produce the json with last-sequence-number:
```python
>>> table_metadata = TableMetadataV2(**EXAMPLE_TABLE_METADATA_V2)
>>> table_metadata.json()
{
	"table-uuid": "aefee669-d568-4f9c-b732-3c0cfd3bc7b0",
	"location": "s3://foo/bar/baz.metadata.json",
	"last-updated-ms": 1600000000000,
	"last-column-id": 4,
	"schemas": [{
		"fields": [{
			"id": 1,
			"name": "foo",
			"type": "string",
			"required": true
		}, {
			"id": 2,
			"name": "bar",
			"type": "string",
			"required": true
		}, {
			"id": 3,
			"name": "baz",
			"type": "string",
			"required": true
		}, {
			"id": 4,
			"name": "qux",
			"type": "string",
			"required": true
		}],
		"schema-id": 0,
		"identifier-field-ids": []
	}],
	"current-schema-id": 0,
	"partition-specs": [{
		"spec-id": 0,
		"fields": []
	}],
	"default-spec-id": 0,
	"last-partition-id": 999,
	"properties": {
		"owner": "root",
		"write.format.default": "parquet",
		"read.split.target.size": 134217728
	},
	"current-snapshot-id": 7681945274687743099,
	"snapshots": [{
		"snapshot-id": 7681945274687743099,
		"timestamp-ms": 1637943123188,
		"summary": {
			"operation": "append",
			"added-data-files": "6",
			"added-records": "237993",
			"added-files-size": "3386901",
			"changed-partition-count": "1",
			"total-records": "237993",
			"total-files-size": "3386901",
			"total-data-files": "6",
			"total-delete-files": "0",
			"total-position-deletes": "0",
			"total-equality-deletes": "0"
		},
		"manifest-list": "s3://foo/bar/baz/snap-2874264644797652805-1-9cb3c3cf-5a04-40c1-bdd9-d8d7e38cd8e3.avro",
		"schema-id": 0
	}],
	"snapshot-log": [{
		"timestamp-ms": 1637943123188,
		"snapshot-id": 7681945274687743099
	}],
	"metadata-log": [{
		"timestamp-ms": 1637943123331,
		"metadata-file": "3://foo/bar/baz/00000-907830f8-1a92-4944-965a-ff82c890e912.metadata.json"
	}],
	"sort-orders": [{
		"order-id": 0,
		"fields": []
	}],
	"default-sort-order-id": 0,
	"format-version": 2,
	"last-sequence-number": 1
}
```

python/src/iceberg/types.py
Fokko: I blame my new keyboard, thanks!

python/src/iceberg/types.py
```diff
@@ -181,29 +275,33 @@ class ListType(IcebergType):
...
-            self,
-            "element",
-            NestedField(
-                name="element",
-                required=self.element_required,
-                field_id=self.element_id,
-                field_type=self.element_type,
-            ),
+    class Config:
+        fields = {"element_field": {"exclude": True}}
```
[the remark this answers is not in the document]
Fokko: Good question! This is because `PrivateField` requires the variable name to start with an underscore, and these variables shouldn't be used outside of the class itself. Because we use the `element_field` in the visitor as well, it makes more sense to make it an actual field, and omit it when we serialize the `ListType`.

python/tests/conftest.py
```diff
@@ -301,3 +319,81 @@ def all_avro_types() -> Dict[str, Any]:
 @pytest.fixture
 def catalog() -> InMemoryCatalog:
     return InMemoryCatalog("test.in.memory.catalog", {"test.key": "test.value"})
+
+
+@pytest.fixture(scope="session")
+def simple_struct():
+    return StructType(
+        NestedField(1, "required_field", StringType(), True, "this is a doc"), NestedField(2, "optional_field", IntegerType())
```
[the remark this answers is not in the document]
Fokko: Thanks!

python/tests/conftest.py
```diff
@@ -301,3 +319,81 @@ def all_avro_types() -> Dict[str, Any]:
...
+        if parsed_location.scheme and parsed_location.scheme != "file":  # Validate that a uri is provided with a scheme of `file`
+            raise ValueError("LocalOutputFile location must have a scheme of `file`")
+        elif parsed_location.netloc:
+            raise ValueError(f"Network location is not allowed for LocalOutputFile: {parsed_location.netloc}")
+
+        super().__init__(location=location)
+        self._parsed_location = parsed_location
+
+    @property
+    def parsed_location(self) -> ParseResult:
```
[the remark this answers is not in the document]
Fokko: I shamelessly copied this from the PR of @samredai, just updated it! Thanks!

python/tests/conftest.py
Fokko: I think for the test I think we should be more strict. If some file hasn't been written, we should know that. I think we can just remove the try-catch logic since the original error will be sufficient

python/tests/table/test_metadata.py
Fokko: I wasn't aware of that, thanks!

python/src/iceberg/table/metadata.py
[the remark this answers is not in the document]
rdblue: Looks reasonable to me.

python/src/iceberg/types.py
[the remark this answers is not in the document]
rdblue: Sounds reasonable to me.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,178 @@
...
+    default_sort_order_id: int = Field(alias="default-sort-order-id")
+    """Default sort order id of the table. Note that this could be used by
+    writers, but is not used when reading because reads use the specs stored
+     in manifest files."""
+
+
+class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
+    @root_validator(pre=True)
+    def populate_schemas(cls, data: Dict[str, Any]):
+        # When we read a V1 format-version, we'll make sure to populate the fields
```
rdblue, line 118: Style (minor): avoid using personal pronouns ("I" or "we") in docs or comments because it ends up making them longer and less direct.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,178 @@
...
+    sort_orders: list = Field(alias="sort-orders")
+    """A list of sort orders, stored as full sort order objects."""
+
+    default_sort_order_id: int = Field(alias="default-sort-order-id")
+    """Default sort order id of the table. Note that this could be used by
+    writers, but is not used when reading because reads use the specs stored
+     in manifest files."""
+
+
+class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
```
rdblue, line 115: Not something we need to do now, but eventually we will want to add `refs` to these classes.

python/src/iceberg/types.py
rdblue: @Fokko, is this supposed to be left in?

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,178 @@
...
+    @root_validator(pre=True)
+    def populate_schemas(cls, data: Dict[str, Any]):
+        # When we read a V1 format-version, we'll make sure to populate the fields
+        # for V2 as well. This makes it easier downstream because we can just
+        # assume that everything is a TableMetadataV2.
+        # When writing, we should stick to the same version that it was,
+        # because bumping the version should be an explicit operation that is up
+        # to the owner of the table.
+        schema = data["schema"]
+        if "schemas" in data:
```
rdblue, line 125: I think you need the same handling for `partition-spec` and `partition-specs` / `default-spec-id` and some handling to initialize `sort-orders` / `default-sort-order-id`. You might want to check the spec and make sure this is correctly handling all of the optional v1 fields.

python/src/iceberg/types.py
Fokko: I'm not sure why I missed this comment. So, this is supposed to be still in, but I moved the actual parsing (regex) to the type itself. Because the schema is a `List[IcebergType]` it will try to validate that one first. As an example fixed[22]` It will not be passed to the Fixed validate method (because the `__root__` is dynamic).

Before I had everything in the validate method, now I call a static method on the class itself:

```python
            if v.startswith("fixed"):
                return FixedType.parse(v)
```

python/src/iceberg/table/metadata.py
Fokko: Ah, that one slipped in. Thanks 👍🏻

python/src/iceberg/table/metadata.py
Fokko: I went all-in on the spec today, and added them 👍🏻

python/src/iceberg/table/metadata.py
Fokko: Hey @rdblue I've added tests for handling the sort-order, based on this input:

- https://github.com/apache/iceberg/blob/master/core/src/test/resources/TableMetadataV1Valid.json
- https://github.com/apache/iceberg/blob/master/core/src/test/resources/TableMetadataV2Valid.json

This also shows that my previous test data was incorrect. For example, the partition spec doesn't have a spec id. I now set those to default for V1. I do believe that sort orders should work now, there are no fields to be migrated in V1/V2 if I understand correctly, it is just that it was optional before. We can also handle this as soon as we start adding the sort order classes. Let me know if I'm overlooking something here.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # When we read a V1 format-version, we'll make sure to populate the fields
+    # for V2 as well. This makes it easier downstream because we can just
+    # assume that everything is a TableMetadataV2.
+    # When writing, we should stick to the same version that it was,
+    # because bumping the version should be an explicit operation that is up
+    # to the owner of the table.
+
+    @root_validator(pre=True)
+    def set_schema_id(cls, data: Dict[str, Any]):
```
rdblue, line 136: This name doesn't quite describe what the method does right now.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+        data["schema"]["schema-id"] = DEFAULT_SCHEMA_ID
+        data["default-spec-id"] = INITIAL_SPEC_ID
+        data["last-partition-id"] = max(spec["field-id"] for spec in data["partition-spec"])
+        data["default-sort-order-id"] = 0
+        return data
+
+    @root_validator()
+    def migrate_schema(cls, data: Dict[str, Any]):
+        # Migrate schemas
+        schema = data["schema_"]
```
rdblue, line 149: Don't we use the `schema` alias for serialization? Or is that translation done in the JSON to dict mapping, so we will have `schema_` in the dict?

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+        schemas = data["schemas"]
+        if all([schema != other_schema for other_schema in schemas]):
+            data["schemas"].append(schema)
+        data["current_schema_id"] = schema.schema_id
```
rdblue, line 153: If `current-schema-id` is already set, we should use it. If you want to check against the `schema` field's ID, then I think we should make that a check that throws `ValidationError`.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+        return data
+
+    @root_validator()
+    def migrate_partition_spec(cls, data: Dict[str, Any]):
+        # This is going to be much nicer as soon as partition-spec is also migrated to pydantic
+        if partition_spec := data.get("partition_spec"):
+            data["partition_specs"] = [{**spec, "spec-id": INITIAL_SPEC_ID + idx} for idx, spec in enumerate(partition_spec)]
```
rdblue, line 160: Any spec in `partition_specs` will have a spec ID assigned. We only need to assign a default ID to the `partition-spec` field, and then only if `partition-specs` is undefined.

rdblue: I think the logic should be:

```python
if not data["partition_specs"]:
    fields = data.get("partition_spec")
    data["partition_specs"] = [{"spec-id": INITIAL_SPEC_ID, "fields": fields}]
```

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+    def check_partition_spec(cls, data: Dict[str, Any]):
+        default_spec_id = data["default_spec_id"]
+
+        for spec in data["partition_specs"]:
+            if spec["spec-id"] == default_spec_id:
+                return data
+
+        raise ValueError(f"default-spec-id {default_spec_id} can't be found")
+
+    @root_validator(skip_on_failure=True)
```
rdblue, line 208: Why skip on failure? If the order ID is defined, it should match a sort order.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+    table_uuid: UUID = Field(alias="table-uuid")
+    """A UUID that identifies the table, generated when the table is created.
+    Implementations must throw an exception if a table’s UUID does not match
+    the expected UUID after refreshing metadata."""
+
+    last_sequence_number: int = Field(alias="last-sequence-number", default=_INITIAL_SEQUENCE_NUMBER)
+    """The table’s highest assigned sequence number, a monotonically
+    increasing long that tracks the order of snapshots in a table."""
+
+    refs: Dict[str, SnapshotRef] = Field(default_factory=dict)
```
rdblue, line 235: This can be in common metadata. There's no reason why v1 tables can't also have refs. Adding refs is a forward-compatible change.

python/src/iceberg/types.py
rdblue: Okay, I see the calls to `parse`. That brings us back to my original question, which is why are we doing parsing in a validate method? Shouldn't this validate the input rather than returning the deserialized value?

I would expect this to do something like the schema validator, which would change `"fixed[12]"` into `{"__root__": "fixed", "length": 12} or something.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+        data["default-spec-id"] = INITIAL_SPEC_ID
+        data["last-partition-id"] = max(spec["field-id"] for spec in data["partition-spec"])
+        data["default-sort-order-id"] = 0
+        return data
+
+    @root_validator()
+    def migrate_schema(cls, data: Dict[str, Any]):
+        # Migrate schemas
+        schema = data["schema_"]
+        schemas = data["schemas"]
```
rdblue, line 150: This needs to be defaulted, right?

python/src/iceberg/table/metadata.py
Fokko: It used to do, but it grew a bit out of control. I renamed it to `set_v2_compatible_defaults`

python/src/iceberg/table/metadata.py
Fokko: So, this root_validator is after the aliases have applied (otherwise it would be `@root_validator(pre=True)`. This means that the reserved Pydantic `.schema` is already translated to `.schema_`. When we serilaize it, we go back to `schema`:
```json
{
	"location": "s3://bucket/test/location",
	"last-updated-ms": 1602638573874,
	"last-column-id": 3,
	"schemas": [{
		"fields": [{
			"id": 1,
			"name": "x",
			"type": "long",
			"required": true
		}, {
			"id": 2,
			"name": "y",
			"type": "long",
			"required": true,
			"doc": "comment"
		}, {
			"id": 3,
			"name": "z",
			"type": "long",
			"required": true
		}],
		"schema-id": 0,
		"identifier-field-ids": []
	}],
	"current-schema-id": 0,
	"partition-specs": [{
		"name": "x",
		"transform": "identity",
		"source-id": 1,
		"field-id": 1000,
		"spec-id": 0
	}],
	"default-spec-id": 0,
	"last-partition-id": 1000,
	"properties": {},
	"current-snapshot-id": -1,
	"snapshots": [{
		"snapshot-id": 1925,
		"timestamp-ms": 1602638573822
	}],
	"snapshot-log": [],
	"metadata-log": [],
	"sort-orders": [],
	"default-sort-order-id": 0,
	"table-uuid": "d20125c8-7284-442c-9aea-15fee620737c",
	"format-version": 1,
	"schema": {
		"fields": [{
			"id": 1,
			"name": "x",
			"type": "long",
			"required": true
		}, {
			"id": 2,
			"name": "y",
			"type": "long",
			"required": true,
			"doc": "comment"
		}, {
			"id": 3,
			"name": "z",
			"type": "long",
			"required": true
		}],
		"schema-id": 0,
		"identifier-field-ids": []
	},
	"partition-spec": [{
		"name": "x",
		"transform": "identity",
		"source-id": 1,
		"field-id": 1000
	}]
}
```

python/src/iceberg/table/metadata.py
Fokko: Makes sense. I've made it conditionally and added a `ValidationError`. It feels a bit weird that both of them can be set, and they serve the same purpose. I've added a check as well.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+
+class SnapshotRef(IcebergBaseModel):
+    snapshot_id: int = Field(alias="snapshot-id")
+    snapshot_ref_type: SnapshotRefType = Field(alias="type")
+    min_snapshots_to_keep: int = Field(alias="min-snapshots-to-keep")
+    max_snapshot_age_ms: int = Field(alias="max-snapshot-age-ms")
+    max_ref_age_ms: int = Field(alias="max-ref-age-ms")
+
+
+class TableMetadataCommonFields(IcebergBaseModel):
```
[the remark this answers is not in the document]
Fokko: Since refs are added as optional in V2:
image

I've added them as a property of the V2. Since they are optional, I set the `default_factory` to `dict`:
```
class TableMetadataV2(TableMetadataCommonFields, IcebergBaseModel):
...

    refs: Dict[str, SnapshotRef] = Field(default_factory=dict)
    """A map of snapshot references.
    The map keys are the unique snapshot reference names in the table,
    and the map values are snapshot reference objects.
    There is always a main branch reference pointing to the
    current-snapshot-id even if the refs map is null."""
```
If it isn't supplied, an empty `dict` will be set.

python/src/iceberg/table/metadata.py
Fokko: What this means is that if there are missing fields, then we don't want to run the validation of the `sort_order`, because we have other problems (maybe missing fields). Therefore I figured to only run the validation if all prior validations have succeeded.

From: https://pydantic-docs.helpmanual.io/usage/validators/#root-validators

As with field validators, root validators can have `pre=True`, in which case they're called before field validation occurs (and are provided with the raw input data), or `pre=False` (the default), in which case they're called after field validation.

Field validation will not occur if `pre=True` root validators raise an error. As with field validators, "post" (i.e. pre=False) root validators by default will be called even if prior validators fail; this behaviour can be changed by setting the `skip_on_failure=True` keyword argument to the validator. The `values` argument will be a dict containing the values which passed field validation and field defaults where applicable.

python/src/iceberg/table/metadata.py
Fokko: Fair point, I've just moved them.

python/src/iceberg/table/metadata.py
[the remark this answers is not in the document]
rdblue: Got it, that makes sense. I think we still want to add `main` if `current-snapshot-id` is set, so we probably do want to use a validator.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,259 @@
...
+    # When writing, we should stick to the same version that it was,
+    # because bumping the version should be an explicit operation that is up
+    # to the owner of the table.
+
+    @root_validator(pre=True)
+    def set_schema_id(cls, data: Dict[str, Any]):
+        # Set some sensible defaults for V1, so we comply with the schema
+        # this is in pre=True, meaning that this will be done before validation
+        # we don't want to make them optional, since we do require them for V2
+        data["schema"]["schema-id"] = DEFAULT_SCHEMA_ID
```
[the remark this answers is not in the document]
Fokko: Good one, updated!

python/src/iceberg/table/metadata.py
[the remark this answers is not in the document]
Fokko: Yes, just added that from the logic in:  https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L309

Came up with:

```python
    @root_validator(skip_on_failure=True)
    def construct_refs(cls, data: Dict[str, Any]):
        # This is going to be much nicer as soon as sort-order is an actual pydantic object
        if not data.get("refs"):
            if current_snapshot_id := data.get("current_snapshot_id"):
                if current_snapshot_id != -1:
                    data["refs"] = {
                        MAIN_BRANCH: SnapshotRef(snapshot_id=current_snapshot_id, snapshot_ref_type=SnapshotRefType.branch)
                    }
        return data
```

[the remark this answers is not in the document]
rdblue: Looks good, except for the comment about sort order.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,363 @@
...
+from iceberg.utils.iceberg_base_model import IcebergBaseModel
+
+_INITIAL_SEQUENCE_NUMBER = 0
+INITIAL_SPEC_ID = 0
+DEFAULT_SCHEMA_ID = 0
+DEFAULT_SORT_ORDER_UNSORTED = 0
+
+
+class ValidationError(Exception):
+    ...
```
rdblue, line 41: Can you move this to `exceptions.py`?

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,363 @@
...
+
+    raise ValidationError(f"default-spec-id {default_spec_id} can't be found")
+
+
+def check_sort_orders(values: Dict[str, Any]) -> Dict[str, Any]:
+    """Validator to check if the default_sort_order_id is present in sort-orders"""
+    default_sort_order_id = values["default_sort_order_id"]
+
+    # 0 == unsorted
+    if default_sort_order_id != 0:
```
rdblue, line 71: Should this handle the case where default_sort_order_id is None?

python/src/iceberg/table/metadata.py
Fokko: Ah, missed that one. Just moved it.

python/src/iceberg/table/metadata.py
Fokko: The `default-sort-order-id` can't be null since it is a required field in v2, and for v1 we set it to 0 if not set:
```python
    @root_validator(pre=True)
    def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
        """Sets default values to be compatible with the format v2

        Set some sensible defaults for V1, so we comply with the schema
        this is in pre=True, meaning that this will be done before validation.
        We don't want to make the fields optional, since they are required for V2

        Args:
            data: The raw arguments when initializing a V1 TableMetadata

        Returns:
            The TableMetadata with the defaults applied
        """
...
        if "default-sort-order-id" not in data:
            data["default-sort-order-id"] = DEFAULT_SORT_ORDER_UNSORTED
        return data
```

Fokko: To elaborate, we can't set it to 0 by default, because this would also affect V2. If there would be no sort order supplied in V2, then it would default to 0.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+    return values
+
+
+class TableMetadataCommonFields(IcebergBaseModel):
+    """Metadata for an Iceberg table as specified in the Apache Iceberg
+    spec (https://iceberg.apache.org/spec/#iceberg-table-spec)"""
+
+    @root_validator(skip_on_failure=True)
+    def construct_refs(cls, data: Dict[str, Any]):
+        # This is going to be much nicer as soon as sort-order is an actual pydantic object
```
rdblue, line 83: Nit: copy & paste not quite updated, since this talks about sort-order.

python/src/iceberg/types.py
Fokko: Doesn't look like the `parse_obj`/`parse_raw` gets invoked. The `IcebergType` is a bit of a special case, since it cannot be invoked, I think it would be okay to (mis)use the validator there. It gets called because it is part of the schema:
```
class Schema:
    fields: Tuple[IcebergType. ...]
```
When deserializing, it will first visit the IcebergType for validation. We can then handle the special cases for the decimal and fixed.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+    partition specs for the table. This is used to ensure partition fields
+    are always assigned an unused ID when evolving specs."""
+
+    properties: Dict[str, str] = Field(default_factory=dict)
+    """	A string to string map of table properties. This is used to
+    control settings that affect reading and writing and is not intended
+    to be used for arbitrary metadata. For example, commit.retry.num-retries
+    is used to control the number of commit retries."""
+
+    current_snapshot_id: Optional[int] = Field(alias="current-snapshot-id", default=-1)
```
rdblue, line 129: Non-blocker: we use -1 in the format, but if I could do it over, I would probably avoid it and use `null` / `None` instead. You may want to consider converting in a pre-validation from -1 to clean this up.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+        metadata = copy(self.dict())
+        metadata["format_version"] = 2
+        return TableMetadataV2(**metadata)
+
+    format_version: Literal[1] = Field(alias="format-version")
+    """An integer version number for the format. Currently, this can be 1 or 2
+    based on the spec. Implementations must throw an exception if a table’s
+    version is higher than the supported version."""
+
+    table_uuid: Optional[UUID] = Field(alias="table-uuid")
```
rdblue, line 287: Since this is defaulted in a pre-validator, I think you could move it to common and remove the `Optional`.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+            data: The raw data after validation, meaning that the aliases are applied
+
+        Returns:
+            The TableMetadata with the sort_orders set, if not provided
+        """
+        # This is going to be much nicer as soon as sort-order is an actual pydantic object
+        # Probably we'll just create a UNSORTED_ORDER constant then
+        if not data.get("sort_orders"):
+            data["sort_orders"] = [{"order_id": 0, "fields": []}]
+            data["default_sort_order_id"] = 0
```
rdblue, line 272: Use the constant instead of `0`?

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+    """An integer version number for the format. Currently, this can be 1 or 2
+    based on the spec. Implementations must throw an exception if a table’s
+    version is higher than the supported version."""
+
+    table_uuid: Optional[UUID] = Field(alias="table-uuid")
+    """A UUID that identifies the table, generated when the table is created.
+    Implementations must throw an exception if a table’s UUID does not match
+    the expected UUID after refreshing metadata."""
+
+    schema_: Schema = Field(alias="schema")
```
rdblue, line 292: (Not urgent) One thing we will want to watch out for here is ensuring this is set when `current_schema_id` changes.

python/src/iceberg/table/metadata.py
Fokko: I've added a pre-validator. This way we can also keep it cleaner down the line.

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+
+    raise ValidationError(f"default-spec-id {default_spec_id} can't be found")
+
+
+def check_sort_orders(values: Dict[str, Any]) -> Dict[str, Any]:
+    """Validator to check if the default_sort_order_id is present in sort-orders"""
+    default_sort_order_id = values["default_sort_order_id"]
+
+    # 0 == unsorted
+    if default_sort_order_id != 0:
```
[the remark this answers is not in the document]
Fokko: Great one, also removed the comment

python/src/iceberg/table/metadata.py
Fokko: Done! Keep in mind that we keep it nullable in Java: https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L317

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+    for schema in values["schemas"]:
+        if schema.schema_id == current_schema_id:
+            return values
+
+    raise ValidationError(f"current-schema-id {current_schema_id} can't be found in the schemas")
+
+
+def check_partition_specs(values: Dict[str, Any]) -> Dict[str, Any]:
+    """Validator to check if the default-spec-id is present in partition-specs"""
+    default_spec_id = values["default_spec_id"]
```
rdblue, line 54: (Not urgent) One thing I find slightly hard to follow, especially now that these validation methods are shared, is when the `name_with_underscores` vs `name-with-dashes` is used. I know the latter is when running a pre-validator, but it would be nice to standardize on either pre-validation or normal validation so we don't have to think about when to translate field strings by hand.

It's also a bit odd that this uses `partition_specs`, but `spec-id` at the inner level... Seems like following up to use pre-validation everywhere would help readability. What do you think?

python/src/iceberg/table/metadata.py
rdblue: Got it. So just don't run if other validators have failed. Makes sense!

python/src/iceberg/table/metadata.py
```diff
@@ -0,0 +1,360 @@
...
+        Args:
+            data: The raw data after validation, meaning that the aliases are applied
+
+        Returns:
+            The TableMetadata with the partition_specs set, if not provided
+        """
+        # This is going to be much nicer as soon as partition-spec is also migrated to pydantic
+        if not data.get("partition_specs"):
+            fields = data["partition_spec"]
+            data["partition_specs"] = [{"spec-id": INITIAL_SPEC_ID, "fields": fields}]
```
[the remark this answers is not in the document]
Fokko: Hmm, this is actually a remained of some refactoring along the way. Since this is not a pre-validator, the field already needs to be set. The `default_schema_id` has a default. I've also set a default for the partition-spec, so everything is consistent again.

python/src/iceberg/table/refs.py
```diff
@@ -0,0 +1,37 @@
...
+from pydantic import Field
+
+from iceberg.utils.iceberg_base_model import IcebergBaseModel
+
+MAIN_BRANCH = "main"
+
+
+class SnapshotRefType(str, Enum):
+    branch = "branch"
+    tag = "tag"
```
rdblue, line 29: Minor: Most of the [examples in the stdlib docs](https://docs.python.org/3/library/enum.html#enum.Enum) use upper case, which would match other constants and Java. Why not BRANCH and TAG?

rdblue (approved):

python/src/iceberg/table/metadata.py
rdblue: Yeah, we kept it nullable in Java because in the process of adding a UUID could cause weird cases where UUIDs were concurrently assigned. That would cause commits to fail because it appears to be a different table. Since UUIDs have been supported for a long time, that's no longer likely so I don't think we need to care about it for the Python library.

python/src/iceberg/table/metadata.py
Fokko: I agree, it is a bit confusing. I think we should as little as pre-validators as possible, and try to standardize the use of the underscores.

> It's also a bit odd that this uses partition_specs, but spec-id at the inner level... Seems like following up to use pre-validation everywhere would help readability. What do you think?

This will be fixed once we start moving the downstream objects to Pydantic as well. `spec["spec-id"]` will change into `spec.spec_id` since it is then an object (in the post-validation at least :)

python/src/iceberg/table/refs.py
Fokko: Good one. I see this being used both ways, since they are static, I prefer uppercase. Updated!

rdblue (approved):

rdblue:
Nice work, @Fokko! Thanks for getting this done!

PR #5012, opened 2022-06-10
Docs: use 'SET sql-client.execution.result-mode = tableau' instead of 'SET …

ymZhao1001:
'SET execution.result-mode=tableau' does not work.
use 'SET sql-client.execution.result-mode = tableau' instead of 'SET execution.result-mode=tableau';


kbendick:
Thanks @ymZhao1001! Do you know if this change occurred at a specific version (or which version you tested this with)?

It would be great if you could link to specific documentation mentioning the added `sql-client` prefix, just in case we need to mention both depending on version. 🙂

ymZhao1001:
Thank you for your reply.  @kbendick  it worked since flink 1.13 .  I will add it to the document.
use 'SET execution.result-mode=tableau;'  at flink 1.12 or before
u can see it at https://nightlies.apache.org/flink/flink-docs-release-1.12/dev/table/sqlClient.html
after flink 1.13 , use 'SET sql-client.execution.result-mode = tableau'  
u can see it at  https://nightlies.apache.org/flink/flink-docs-release-1.13/docs/dev/table/sqlclient/

kbendick (commented):
Since Flink 1.12 is no longer supported, I think we can just update this and remove the hint cc @ymZhao1001!

Once this <hint info> is removed, this LGTM. Thanks @ymZhao1001!

ymZhao1001:
> Since Flink 1.12 is no longer supported, I think we can just update this and remove the hint cc @ymZhao1001!
> 
> Once this is removed, this LGTM. Thanks @ymZhao1001!

removed. @kbendick

PR #5019, opened 2022-06-13
Verify no stream use after close in in-memory tests

findepi:
The in-memory `InputFile` and `OutputFile` implementations has an
undesirable property that returned streams allowed use after close.
Since most streams don't have such property, it should be disallowed
here too.

Extracted from https://github.com/apache/iceberg/pull/4537 per https://github.com/apache/iceberg/pull/4741#discussion_r895212259


rdblue (approved):

rdblue:
Thanks, @findepi!

PR #5020, opened 2022-06-13
LICENSE: Fix copy paste: s/parquet/avro/

Fokko:
I think we forgot to update the home page from the Parquet block.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -218,7 +218,7 @@ This product includes code from Apache Avro.
 * Conversion in DecimalWriter is based on Avro's Conversions.DecimalConversion.
 
 Copyright: 2014-2017 The Apache Software Foundation.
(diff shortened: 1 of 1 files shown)


singhpk234 (approved):

singhpk234:
@Fokko , presently the CI ran all the ut's of engines, even for an update for license. As per my understanding it should not be required, as presently we avoid running them when we update something in `readme.md` etc, have filed a [pr](https://github.com/apache/iceberg/pull/5022) for the same, Please do let me know WDYT of it ?

Fokko:
@singhpk234 that makes sense, thanks!

rdblue (approved):

Fokko:
@rdblue I checked the rest, and it looks good 👍🏻

PR #5021, opened 2022-06-13
Add API changes for statistics information in table metadata

findepi:
API for https://github.com/apache/iceberg/pull/4945
Implementation PR: https://github.com/apache/iceberg/pull/4741

Extracted from https://github.com/apache/iceberg/pull/4741 per https://github.com/apache/iceberg/pull/4741#discussion_r895213091

### DIFF
--- a/api/src/main/java/org/apache/iceberg/Table.java
+++ b/api/src/main/java/org/apache/iceberg/Table.java
@@ -264,6 +264,17 @@ public interface Table {
   DeleteFiles newDelete();
 
   /**
+   * Create a new {@link UpdateStatistics update table statistics API} to add or remove statistics
+   * files in this table.
+   *
+   * @return a new {@link UpdateStatistics}
+   */
+  default UpdateStatistics updateStatistics() {
+    throw new UnsupportedOperationException(
+        "Updating statistics is not supported by " + getClass().getName());
+  }
+
+  /**
    * Create a new {@link ExpireSnapshots expire API} to manage snapshots in this table and commit.
    *
    * @return a new {@link ExpireSnapshots}
--- a/api/src/main/java/org/apache/iceberg/Transaction.java
+++ b/api/src/main/java/org/apache/iceberg/Transaction.java
@@ -138,6 +138,17 @@ public interface Transaction {
   DeleteFiles newDelete();
 
   /**
+   * Create a new {@link UpdateStatistics update table statistics API} to add or remove statistics
+   * files in this table.
+   *
+   * @return a new {@link UpdateStatistics}
+   */
+  default UpdateStatistics updateStatistics() {
+    throw new UnsupportedOperationException(
+        "Updating statistics is not supported by " + getClass().getName());
+  }
+
+  /**
    * Create a new {@link ExpireSnapshots expire API} to manage snapshots in this table.
    *
    * @return a new {@link ExpireSnapshots}
--- a/api/src/main/java/org/apache/iceberg/UpdateStatistics.java
+++ b/api/src/main/java/org/apache/iceberg/UpdateStatistics.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg;
+
+/** API for updating statistics files in a table. */
+public interface UpdateStatistics extends PendingUpdate<Snapshot> {
+  /**
+   * Set the table's statistics file for given snapshot, replacing the previous statistics file for
+   * the snapshot if any exists.
+   *
+   * @return this for method chaining
+   */
+  UpdateStatistics setStatistics(long snapshotId, StatisticsFile statisticsFile);
+
+  /**
+   * Remove the table's statistics file for given snapshot.
+   *
+   * @return this for method chaining
+   */
+  UpdateStatistics removeStatistics(long snapshotId);
+}
--- a/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
@@ -224,6 +224,11 @@ public abstract class BaseMetadataTable implements Table, HasTableOperations, Se
   }
 
   @Override
+  public UpdateStatistics updateStatistics() {
+    throw new UnsupportedOperationException("Cannot update statistics of a metadata table");
+  }
+
+  @Override
   public ExpireSnapshots expireSnapshots() {
     throw new UnsupportedOperationException("Cannot expire snapshots from a metadata table");
   }
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -148,6 +148,12 @@ acceptedBreaks:
       new: "method org.apache.iceberg.ReplacePartitions org.apache.iceberg.ReplacePartitions::validateNoConflictingDeletes()"
       justification: "Accept all changes prior to introducing API compatibility checks"
     - code: "java.method.addedToInterface"
+      new: "method org.apache.iceberg.UpdateStatistics org.apache.iceberg.Table::updateStatistics()"
+      justification: "new API method"
+    - code: "java.method.addedToInterface"
+      new: "method org.apache.iceberg.UpdateStatistics org.apache.iceberg.Transaction::updateStatistics()"
+      justification: "new API method"
+    - code: "java.method.addedToInterface"
       new: "method org.apache.iceberg.expressions.Expression org.apache.iceberg.Scan<ThisT,\
         \ T extends org.apache.iceberg.ScanTask, G extends org.apache.iceberg.ScanTaskGroup<T\
         \ extends org.apache.iceberg.ScanTask>>::filter()"
--- a/core/src/main/java/org/apache/iceberg/BaseTransaction.java
+++ b/core/src/main/java/org/apache/iceberg/BaseTransaction.java
@@ -712,6 +712,11 @@ public class BaseTransaction implements Transaction {
     }
 
     @Override
+    public UpdateStatistics updateStatistics() {
+      return BaseTransaction.this.updateStatistics();
+    }
+
+    @Override
     public ExpireSnapshots expireSnapshots() {
       return BaseTransaction.this.expireSnapshots();
     }
--- a/core/src/main/java/org/apache/iceberg/CommitCallbackTransaction.java
+++ b/core/src/main/java/org/apache/iceberg/CommitCallbackTransaction.java
@@ -102,6 +102,11 @@ class CommitCallbackTransaction implements Transaction {
   }
 
   @Override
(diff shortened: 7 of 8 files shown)


api/src/main/java/org/apache/iceberg/Transaction.java
```diff
@@ -137,6 +137,14 @@ default AppendFiles newFastAppend() {
    */
   DeleteFiles newDelete();
 
+  /**
+   * Create a new {@link UpdateTableStatistics update table statistics API} to add or remove statistics
+   * files in this table.
+   *
+   * @return a new {@link UpdateTableStatistics}
+   */
+  UpdateTableStatistics updateTableStatistics();
```
rdblue, line 146: Usually we add new methods to Table and Transaction at the same time.

findepi: while working on https://github.com/apache/iceberg/pull/4741 I didn't realize i should have changes in Table as well. sure, will add.

findepi:
@rdblue thank you for your review. Updated accordingly.

api/src/main/java/org/apache/iceberg/StatisticsFile.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.util.List;
+
+public interface StatisticsFile {
+  String location();
```
rdblue, line 25: Should this be `path` to match the spec changes?

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
```diff
@@ -216,6 +216,11 @@ public DeleteFiles newDelete() {
     throw new UnsupportedOperationException("Cannot delete from a metadata table");
   }
 
+  @Override
+  public UpdateTableStatistics updateTableStatistics() {
+    throw new UnsupportedOperationException("Cannot update statistics of a metadata table");
```
rdblue, line 221: Could you add this as a default implementation rather than adding it in this class? That way, if anyone is using the interface it doesn't break them.

api/src/main/java/org/apache/iceberg/BlobMetadata.java
```diff
@@ -0,0 +1,35 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.List;
+import java.util.Map;
+
+public interface BlobMetadata {
```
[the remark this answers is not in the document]
findepi: Added

core/src/main/java/org/apache/iceberg/BaseMetadataTable.java
findepi: Sure, added default method and removed impl stubs from BaseTable and BaseTransaction.
Impl in BaseMetadataTable stays because it includes "a metadata table" in the exc msg (consistently with all other methods in this class).

findepi:
Currently, based on https://github.com/apache/iceberg/pull/4978

findepi:
@rdblue updated this to have one stats file for a table.

findepi:
Per offline ask from @rdblue , I extracted prep change here: https://github.com/apache/iceberg/pull/5129

findepi:
Updated after merge of https://github.com/apache/iceberg/pull/4945.

@rdblue can you please take a look?

build.gradle
```diff
@@ -224,6 +224,7 @@ project(':iceberg-api') {
 
   dependencies {
     implementation project(path: ':iceberg-bundled-guava', configuration: 'shadow')
+    compileOnly "com.google.code.findbugs:jsr305"
```
rdblue, line 227: This has license issues and is not allowed. Instead, we use `com.github.stephenc.findbugs`, which is already available.

api/src/main/java/org/apache/iceberg/Table.java
```diff
@@ -263,6 +263,17 @@ default AppendFiles newFastAppend() {
    */
   DeleteFiles newDelete();
 
+  /**
+   * Create a new {@link UpdateTableStatistics update table statistics API} to add or remove
+   * statistics files in this table.
+   *
+   * @return a new {@link UpdateTableStatistics}
+   */
+  default UpdateTableStatistics updateTableStatistics() {
```
rdblue, line 272: I think it is sufficient to call this `updateStatistics`. Is there a reason to qualify it with `table`?

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -884,6 +895,17 @@ private Builder(TableMetadata base) {
...
       this.refs = Maps.newHashMap(base.refs);
+      this.statistics =
+          base.statistics.stream()
+              .collect(
+                  Collectors.toMap(
+                      StatisticsFile::snapshotId,
+                      Function.identity(),
+                      (a, b) -> {
+                        throw new IllegalArgumentException(
+                            "Two statistics files for same snapshot");
```
rdblue, line 906: Is it a good idea to reject duplicate stats files? That seems like a way to needlessly break a table.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -1176,6 +1198,19 @@ public Builder removeBranch(String branch) {
       return this;
     }
 
+    public Builder setStatistics(String snapshotId, @Nullable StatisticsFile statisticsFile) {
+      Preconditions.checkNotNull(snapshotId, "snapshotId is null");
+      Preconditions.checkNotNull(statisticsFile, "statisticsFile is null");
+      this.statistics.put(snapshotId, statisticsFile);
+      return this;
```
rdblue, line 1205: If these methods don't add changes to the `changes` list, then the builder won't recognize that anything has been updated and will return the original metadata. This should be picked up by tests of the table metadata updates.

I think you should probably move the changes to this file (and change additions) to a separate PR. We're going to need to add `MetadataUpdate` types for this, and update the JSON parser for serializing those updates. That, and testing everything. This is probably best separated from the API changes, which can be done independently without an implementation.

core/src/main/java/org/apache/iceberg/TableMetadataParser.java
```diff
@@ -498,6 +498,7 @@ static TableMetadata fromJson(FileIO io, String metadataLocation, JsonNode node)
         entries.build(),
         metadataEntries.build(),
         refs,
+        ImmutableList.of(), /* TODO: support statistics */
```
rdblue, line 501: I think this should be done in a PR with the `TableMetadata` updates.

build.gradle
findepi: dropping

api/src/main/java/org/apache/iceberg/Table.java
findepi: ok

core/src/main/java/org/apache/iceberg/TableMetadata.java
findepi: good point

core/src/main/java/org/apache/iceberg/TableMetadata.java
findepi: moved to https://github.com/apache/iceberg/pull/5450

core/src/main/java/org/apache/iceberg/TableMetadataParser.java
findepi: moved to https://github.com/apache/iceberg/pull/5450

core/src/main/java/org/apache/iceberg/TableMetadata.java
findepi: Can you perhaps point me to existing test coverage I should follow for these?

findepi:
@rdblue thank you for a careful review. Comments applied, the code is much better now.
The potentially immature changes to TableMetadata are moved to https://github.com/apache/iceberg/pull/5450

Can you please take another look?

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: @findepi, the newer tests are going into `TestTableMetadata`. Before the builder, `TableMetadata` exposed methods like `replaceSortOrder` that are tested directly (that's why it's hard to find the tests). And quite a few things are still tested end-to-end using the table operations themselves (like `UpdateSchema`). I'd take a look at the tests for `replaceSortOrder` in `TestTableMetadata`.

findepi: Thanks @rdblue for the pointer.
FYI, per the suggestion, i backed out `TableMetadata` changes from this PR. Will address these tests in the other one.
Can I assume the tests for `TableMetadata` changes are not prerequisite for this PR being merged?

findepi:
Updated to apply https://github.com/apache/iceberg/pull/5450#discussion_r939727956.

findepi:
Build failure 
```
> Could not resolve all files for configuration ':classpath'.
   > Could not resolve com.palantir.baseline:gradle-baseline-java:4.0.0.
     Required by:
         project :
      > Could not resolve com.palantir.baseline:gradle-baseline-java:4.0.0.
         > Could not get resource 'https://plugins.gradle.org/m2/com/palantir/baseline/gradle-baseline-java/4.0.0/gradle-baseline-java-4.0.0.pom'.
            > Could not GET 'https://plugins.gradle.org/m2/com/palantir/baseline/gradle-baseline-java/4.0.0/gradle-baseline-java-4.0.0.pom'.
               > The server may not support the client's requested TLS protocol versions: (TLSv1.2, TLSv1.3). You may need to configure the client to allow other protocols to be used. See: https://docs.gradle.org/7.4.2/userguide/build_environment.html#gradle_system_properties
                  > Received fatal alert: handshake_failure
```

looks unrelated.

Let me push amended commit to restart the build.

findepi:
Currently, depends on https://github.com/apache/iceberg/pull/5450

findepi:
> Currently, depends on #5450

rebased after that one is merged.

@rdblue please take a look

api/src/main/java/org/apache/iceberg/UpdateTableStatistics.java
```diff
@@ -0,0 +1,32 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+/** API for updating statistics files in a table. */
+public interface UpdateTableStatistics extends PendingUpdate<Snapshot> {
```
rdblue, line 24: Looks like this didn't get renamed after updating the method. It should be `UpdateStatistics`, I think.

findepi: renamed, thanks

api/src/main/java/org/apache/iceberg/UpdateStatistics.java
```diff
@@ -0,0 +1,32 @@
...
+
+/** API for updating statistics files in a table. */
+public interface UpdateStatistics extends PendingUpdate<Snapshot> {
+  /**
+   * Set the table's statistics file for given snapshot, replacing the previous statistics file for
+   * the snapshot if any exists.
+   *
+   * @return this for method chaining
+   */
+  UpdateStatistics setStatisticsFile(long snapshotId, @Nullable StatisticsFile statisticsFile);
```
rdblue, line 31: Can you remove the `@Nullable` annotation?

How about naming this `setStatistics` like in `TableMetadata.Builder`?

Also, I thought we needed a `removeStatistics` method as well?

rdblue:
@findepi, this is looking close. I think we just need to finalize the `UpdateStatistics` methods. Do you have a PR for the `UpdateStatistics` implementation? We should be able to work on that in parallel.

api/src/main/java/org/apache/iceberg/UpdateStatistics.java
findepi: Done

findepi:
I have applied the requested changes. @rdblue please take another look

rdblue (approved):

findepi:
Thank you for the merge!

PR #5022, opened 2022-06-13
INFRA: Avoid running engine tests on LICENSE and NOTICE update

singhpk234:
Presently on updating LICENSE / NOTICE we trigger complete engine suite, As per my understanding it should not be required, and hence like ignored running tests README / CONTRIBUTING .md we should avoid running it for them.


rdblue (approved):

rdblue:
Thanks, @singhpk234!

PR #5024, opened 2022-06-13
Core: bump Hadoop version to 3.2.0

steveloughran:
Hadoop 3.2.0 is the first release to support java 11, among
other improvements.

Includes a fix for AliyunOSSMockLocalStore to compile by changing
to a different implementation of encodeHex on the classpath
and adding commons-io as a test-time dependency.


steveloughran:
for anyone wondering "why 3.2.0 and not 3.2.3?": you need to upgrade hive from 3.2.1+ just to get guava versions lined up.

aliyun/src/test/java/org/apache/iceberg/aliyun/oss/mock/AliyunOSSMockLocalStore.java
```diff
@@ -37,8 +37,8 @@
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import org.apache.commons.codec.binary.Hex;
```
rdblue, line 40: We try not to rely on Apache Commons for anything to avoid a direct dependency, even in tests (though it looks like that was already violated). Is there an easy way to replace this? We should be able to replace `FileUtils` with a `readFully` method in `IOUtil`

rdblue:
> you need to upgrade hive from 3.2.1+ just to get guava versions lined up.

Iceberg shades Guava to avoid conflicts. Do we need to worry about this?

aliyun/src/test/java/org/apache/iceberg/aliyun/oss/mock/AliyunOSSMockLocalStore.java
steveloughran: commons code it was on the classpath already and i just needed *any* hex.encode call. There's a couple more...let me find them and see which is preferable

steveloughran:
hive and tez turn out to have problems in the public branches; ASF hive is currently only building against 3.1 until https://github.com/apache/hive/pull/3279 is in. I'm going to avoid touching this PR for a bit while I see about getting parquet building against 3.2 first

pan3793:
is there possible to use hadoop shaded client?
https://issues.apache.org/jira/browse/HADOOP-11656

rdblue:
@pan3793, Iceberg should declare its dependencies correctly. If you want to use a shaded version downstream, you can exclude and change the dependency tree.

Fokko:
@steveloughran Parquet 1.13 has been released recently and compiles against Hadoop 3.2.3

steveloughran:
@Fokko i really am not active on this...if you want to take on the "bump things up" project, that'd be wonderful for me.

steveloughran:
giving up on the eternal movign target here. focusing on reflection based access to the cloud-friendly APIs

PR #5029, opened 2022-06-14
Flink: Use Tag or Branch to scan data.

hililiwei:

Scan data using a specified tag or branch.

## Brief change log

Add the following syntax
SQL:
```sql
* SELECT * FROM sample /*+ OPTIONS('tag'='t1')*/ ;
* SELECT * FROM sample /*+ OPTIONS('branch'='t1')*/ ;
```
API:

```java
    public Builder tag(String tag) {
      contextBuilder.useTag(tag);
      return this;
    }

    public Builder branch(String branch) {
      contextBuilder.useBranch(branch);
      return this;
    }
    public Builder startTag(String startTag) {
      contextBuilder.startTag(startTag);
      return this;
    }

    public Builder endTag(String endTag) {
      contextBuilder.endTag(endTag);
      return this;
    }
```

cc @amogh-jahagirdar

### DIFF
--- a/data/src/test/java/org/apache/iceberg/data/GenericAppenderHelper.java
+++ b/data/src/test/java/org/apache/iceberg/data/GenericAppenderHelper.java
@@ -27,6 +27,7 @@ import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DataFiles;
 import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.Files;
+import org.apache.iceberg.SnapshotRef;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.io.FileAppender;
@@ -56,10 +57,11 @@ public class GenericAppenderHelper {
     this(table, fileFormat, tmp, null);
   }
 
-  public void appendToTable(DataFile... dataFiles) {
+  public void appendToTable(String branch, DataFile... dataFiles) {
     Preconditions.checkNotNull(table, "table not set");
 
-    AppendFiles append = table.newAppend();
+    AppendFiles append =
+        table.newAppend().toBranch(branch != null ? branch : SnapshotRef.MAIN_BRANCH);
 
     for (DataFile dataFile : dataFiles) {
       append = append.appendFile(dataFile);
@@ -68,8 +70,21 @@ public class GenericAppenderHelper {
     append.commit();
   }
 
+  public void appendToTable(DataFile... dataFiles) {
+    appendToTable(null, dataFiles);
+  }
+
   public void appendToTable(List<Record> records) throws IOException {
-    appendToTable(null, records);
+    appendToTable(null, null, records);
+  }
+
+  public void appendToTable(String branch, List<Record> records) throws IOException {
+    appendToTable(null, branch, records);
+  }
+
+  public void appendToTable(StructLike partition, String branch, List<Record> records)
+      throws IOException {
+    appendToTable(branch, writeFile(partition, records));
   }
 
   public void appendToTable(StructLike partition, List<Record> records) throws IOException {
--- a/flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
+++ b/flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
@@ -34,6 +34,7 @@ import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.flink.source.split.IcebergSourceSplit;
 import org.apache.iceberg.hadoop.Util;
 import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.util.Tasks;
 
@@ -86,11 +87,23 @@ public class FlinkSplitPlanner {
       IncrementalAppendScan scan = table.newIncrementalAppendScan();
       scan = refineScanWithBaseConfigs(scan, context, workerPool);
 
+      if (context.startTag() != null && table.snapshot(context.startTag()) != null) {
+        scan = scan.fromSnapshotExclusive(table.snapshot(context.startTag()).snapshotId());
+      }
+
       if (context.startSnapshotId() != null) {
+        Preconditions.checkArgument(
+            context.startTag() == null, "START_SNAPSHOT_ID and START_TAG cannot be used both");
         scan = scan.fromSnapshotExclusive(context.startSnapshotId());
       }
 
+      if (context.endTag() != null && table.snapshot(context.endTag()) != null) {
+        scan = scan.toSnapshot(table.snapshot(context.endTag()).snapshotId());
+      }
+
       if (context.endSnapshotId() != null) {
+        Preconditions.checkArgument(
+            context.endTag() == null, "END_SNAPSHOT_ID and END_TAG cannot be used both");
         scan = scan.toSnapshot(context.endSnapshotId());
       }
 
@@ -101,6 +114,10 @@ public class FlinkSplitPlanner {
 
       if (context.snapshotId() != null) {
         scan = scan.useSnapshot(context.snapshotId());
+      } else if (context.tag() != null) {
+        scan = scan.useRef(context.tag());
+      } else if (context.branch() != null) {
+        scan = scan.useRef(context.branch());
       }
 
       if (context.asOfTimestamp() != null) {
@@ -119,7 +136,9 @@ public class FlinkSplitPlanner {
   private static ScanMode checkScanMode(ScanContext context) {
     if (context.isStreaming()
         || context.startSnapshotId() != null
-        || context.endSnapshotId() != null) {
+        || context.endSnapshotId() != null
+        || context.startTag() != null
+        || context.endTag() != null) {
       return ScanMode.INCREMENTAL_APPEND_SCAN;
     } else {
       return ScanMode.BATCH;
--- a/flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
+++ b/flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
@@ -87,7 +87,11 @@ public class FlinkSplitPlanner {
       IncrementalAppendScan scan = table.newIncrementalAppendScan();
       scan = refineScanWithBaseConfigs(scan, context, workerPool);
 
-      if (context.startTag() != null && table.snapshot(context.startTag()) != null) {
+      if (context.startTag() != null) {
+        Preconditions.checkArgument(
+            table.snapshot(context.startTag()) != null,
+            "Cannot find snapshot with tag %s",
+            context.startTag());
         scan = scan.fromSnapshotExclusive(table.snapshot(context.startTag()).snapshotId());
       }
 
@@ -97,7 +101,11 @@ public class FlinkSplitPlanner {
         scan = scan.fromSnapshotExclusive(context.startSnapshotId());
       }
 
-      if (context.endTag() != null && table.snapshot(context.endTag()) != null) {
+      if (context.endTag() != null) {
+        Preconditions.checkArgument(
+            table.snapshot(context.endTag()) != null,
+            "Cannot find snapshot with tag %s",
+            context.endTag());
         scan = scan.toSnapshot(table.snapshot(context.endTag()).snapshotId());
       }
 
--- a/flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
+++ b/flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
@@ -93,7 +93,7 @@ public class FlinkSplitPlanner {
 
       if (context.startSnapshotId() != null) {
         Preconditions.checkArgument(
-            context.startTag() == null, "START_SNAPSHOT_ID and START_TAG cannot be used both");
+            context.startTag() == null, "START_SNAPSHOT_ID and START_TAG cannot both be set");
         scan = scan.fromSnapshotExclusive(context.startSnapshotId());
       }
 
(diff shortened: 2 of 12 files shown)


amogh-jahagirdar:
This has a dependency on https://github.com/apache/iceberg/pull/4428/files, I will follow up on that PR.

hililiwei:
As the core is being modified, this PR needs to wait for #5364  and #5475  to complete.

hililiwei:
cc @amogh-jahagirdar @stevenzwu

hililiwei:
cc @amogh-jahagirdar @stevenzwu @rdblue, could you please take a look when you are available?

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -124,11 +126,33 @@ public void initializeState(FunctionInitializationContext context) throws Except
...
       LOG.info("Restoring state for the {}.", getClass().getSimpleName());
       lastSnapshotId = lastSnapshotIdState.get().iterator().next();
+    } else if (scanContext.startTag() != null) {
+      Snapshot currentSnapshot =
+          scanContext.branch() == null
+              ? table.currentSnapshot()
+              : table.snapshot(scanContext.branch());
+      Preconditions.checkNotNull(
+          currentSnapshot,
+          "Don't have any available snapshot for branch " + scanContext.branch() + " in table.");
```
amogh-jahagirdar, line 136: I think it'll help if at some point (probably options/conf default) we set the default branch to main and always assume a branch is set in this code. Then we can just do table.snapshot(branch).

Nit on log message:

"No snapshots on branch %s in table %s"

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -124,11 +126,33 @@ public void initializeState(FunctionInitializationContext context) throws Except
     if (context.isRestored()) {
       LOG.info("Restoring state for the {}.", getClass().getSimpleName());
       lastSnapshotId = lastSnapshotIdState.get().iterator().next();
+    } else if (scanContext.startTag() != null) {
+      Snapshot currentSnapshot =
+          scanContext.branch() == null
+              ? table.currentSnapshot()
+              : table.snapshot(scanContext.branch());
+      Preconditions.checkNotNull(
```
amogh-jahagirdar, line 134: I think it'll help if at some point (probably options/conf default) we set the default branch to main and always assume a branch is set in this code. Then we can just do table.snapshot(branch).

Nit on log message:

"No snapshots on branch %s in table %s"

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -124,11 +126,33 @@ public void initializeState(FunctionInitializationContext context) throws Except
...
+      lastSnapshotId = startSnapshotId;
     } else if (scanContext.startSnapshotId() != null) {
+      Snapshot currentSnapshot =
+          scanContext.branch() == null
+              ? table.currentSnapshot()
+              : table.snapshot(scanContext.branch());
       Preconditions.checkNotNull(
-          table.currentSnapshot(), "Don't have any available snapshot in table.");
+          currentSnapshot,
+          "Don't have any available snapshot for branch " + scanContext.branch() + " in table.");
```
amogh-jahagirdar, line 153: Also sorry if I missed it, but at some point shouldn't there be validation that the start tag is an ancestor of the end tag?

amogh-jahagirdar (commented):
Thanks for this @hililiwei ! left some comments/questions it makes sense to just carry this PR forward based on our discussion in https://github.com/apache/iceberg/pull/6660

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
```diff
@@ -86,10 +86,18 @@ static CloseableIterable<CombinedScanTask> planTasks(
       IncrementalAppendScan scan = table.newIncrementalAppendScan();
       scan = refineScanWithBaseConfigs(scan, context, workerPool);
 
+      if (context.startTag() != null && table.snapshot(context.startTag()) != null) {
```
jackye1995, line 89: I think we probably need to verify that user only sets one of `START_SNAPSHOT_ID` and `START_TAG`, instead of both. Also the same for end case.

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
jackye1995: +1, can we actually do that in this PR? (can we check what is done in Spark, if we decided to go with using `main` branch or using if else? I cannot remember now)

jackye1995:
@stevenzwu since you are reviewing #6660, could you also take a look at this?

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
hililiwei: spark doesn't seem to use `Main` as the default.

My main concern is that `main` is not compatible with older version iceberg table.

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
hililiwei: They are eventually converted to snapshot id and checked during the `planTasks` phase

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -125,7 +137,24 @@ private void validate() {
...
       }
+
+      Preconditions.checkArgument(
+          branch == null,
+          String.format(
+              "Cannot scan table using ref %s configured for streaming reader yet", branch));
+
+      Preconditions.checkArgument(
+          tag == null,
+          String.format("Cannot scan table using ref %s configured for streaming reader", tag));
```
hililiwei, line 148: #5984 seems to be the prerequisite for flink to implement stream incremental read.

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -125,7 +137,24 @@ private void validate() {
             startSnapshotId == null,
             "Invalid starting snapshot id for SPECIFIC_START_SNAPSHOT_ID strategy: not null");
       }
+
+      Preconditions.checkArgument(
+          branch == null,
+          String.format(
+              "Cannot scan table using ref %s configured for streaming reader yet", branch));
```
hililiwei, line 144: https://github.com/apache/iceberg/pull/5984 seems to be the prerequisite for flink to implement stream incremental read.

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -124,17 +126,27 @@ public void initializeState(FunctionInitializationContext context) throws Except
     if (context.isRestored()) {
       LOG.info("Restoring state for the {}.", getClass().getSimpleName());
       lastSnapshotId = lastSnapshotIdState.get().iterator().next();
-    } else if (scanContext.startSnapshotId() != null) {
+    } else if (scanContext.startTag() != null || scanContext.startSnapshotId() != null) {
+      Preconditions.checkArgument(
+          !(scanContext.startTag() != null && scanContext.startSnapshotId() != null),
+          "START_SNAPSHOT_ID and START_TAG cannot be used both.");
```
jackye1995, line 132: nit: cannot both be set?

jackye1995 (approved):
Mostly looks good to me, just 1 nit comment

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
amogh-jahagirdar: When we read the table metadata with the current Iceberg version we will set the main branch. https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L444 so it will be compatible with older version of tables.

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/FlinkSplitPlanner.java
```diff
@@ -86,11 +87,23 @@ static CloseableIterable<CombinedScanTask> planTasks(
...
```
stevenzwu, line 90: we should add Preconditions check for `table.snapshot(context.startTag())`. Invalid reference should result in an exception. this applies to other refs.

flink/v1.16/flink/src/main/java/org/apache/iceberg/flink/source/StreamingMonitorFunction.java
```diff
@@ -124,17 +126,27 @@ public void initializeState(FunctionInitializationContext context) throws Except
...
+      Preconditions.checkArgument(
+          scanContext.branch() == null,
+          "Cannot scan table using ref %s configured for streaming reader yet.");
       Preconditions.checkNotNull(
           table.currentSnapshot(), "Don't have any available snapshot in table.");
 
       long currentSnapshotId = table.currentSnapshot().snapshotId();
+      long startSnapshotId =
+          scanContext.startTag() != null
+              ? table.snapshot(scanContext.startTag()).snapshotId()
```
stevenzwu, line 142: nit: `table.snapshot(scanContext.startTag())` could result in unfriendly NPE. Maybe Preconditions check?

flink/v1.16/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkScan.java
```diff
@@ -229,6 +230,141 @@ public void testSnapshotReads() throws Exception {
...
+
+    table.manageSnapshots().createBranch("b1", snapshotId).commit();
+
+    List<Record> expectedRecords2 = RandomGenericData.generate(TestFixtures.SCHEMA, 1, 0L);
+    helper.appendToTable(expectedRecords2);
+
+    TestHelpers.assertRecords(
+        runWithOptions(ImmutableMap.of("branch", "b1")), expectedRecords1, TestFixtures.SCHEMA);
+
+    table.manageSnapshots().fastForwardBranch("b1", "main").commit();
```
stevenzwu, line 281: it will probably be a little more interesting test when the two branches diverge. then we can assert different results on two different branches.

flink/v1.16/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamScanSql.java
```diff
@@ -248,4 +267,54 @@ public void testConsumeFromStartSnapshotId() throws Exception {
...
+    Row row4 = Row.of(4, "ddd", "2021-01-01");
+    insertRows(table, row3, row4);
+
+    TableResult result =
+        exec(
+            "SELECT * FROM %s /*+ OPTIONS('streaming'='true', 'monitor-interval'='1s', "
+                + "'start-tag'='%s')*/",
+            TABLE, tagName);
+    try (CloseableIterator<Row> iterator = result.collect()) {
+      // The row2 in start snapshot will be excluded.
```
stevenzwu, line 296: nit: we can simply say the start snapshot is exclusive. mentioning row2  can be less clear.

flink/v1.16/flink/src/test/java/org/apache/iceberg/flink/source/TestStreamScanSql.java
```diff
@@ -30,6 +30,7 @@
 import org.apache.flink.table.api.config.TableConfigOptions;
 import org.apache.flink.types.Row;
 import org.apache.flink.util.CloseableIterator;
+import org.apache.iceberg.AssertHelpers;
```
stevenzwu, line 33: This makes me realized that we don't have the stream sql scan for the FLIP-27 source.

hililiwei: added one in TestStreamingMonitorFunction.

flink/v1.16/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkScan.java
hililiwei: updated. Please take a look, thx.

hililiwei:
@stevenzwu  addressed the round of comments, PTAL, thx.

data/src/test/java/org/apache/iceberg/data/GenericAppenderHelper.java
```diff
@@ -68,8 +70,21 @@ public void appendToTable(DataFile... dataFiles) {
     append.commit();
   }
 
+  public void appendToTable(DataFile... dataFiles) {
+    appendToTable(null, dataFiles);
+  }
+
   public void appendToTable(List<Record> records) throws IOException {
-    appendToTable(null, records);
+    appendToTable(null, null, records);
```
stevenzwu, line 78: nit: this can be `appendToTable(null, records)`, right?

stevenzwu (approved):

amogh-jahagirdar (approved):
Looks good to me, thanks @hililiwei for the contribution!

data/src/test/java/org/apache/iceberg/data/GenericAppenderHelper.java
hililiwei: Currently it won't work because we have two methods:
`appendToTable(String branch, List<Record> records) ` 
and 
`appendToTable(StructLike partition, List<Record> records)`

Unless I rearrange the order of `branch` and place it after `records`, but I prefer to leave it as it is currently, at the front.

jackye1995:
Looks like all comments are addressed and we have enough votes, thanks for the work @hililiwei , and thanks for the review @stevenzwu and @amogh-jahagirdar !

hililiwei:
Thanks for the review @jackye1995 @stevenzwu @amogh-jahagirdar

PR #5030, opened 2022-06-14
add truncate transform

jun-he:
To split the PR https://github.com/apache/iceberg/pull/3450, open a new PR for truncate transform here.


jun-he:
Found at https://github.com/python/cpython/issues/85160 that singledispatchmethod significantly slower than singledispatch. As the community is working to fix it, wondering if we should just keep singledispatchmethod for simplicity.

python/src/iceberg/utils/decimal.py
```diff
@@ -75,3 +75,16 @@ def decimal_to_bytes(value: Decimal) -> bytes:
...
+    """Get a truncated Decimal value given a decimal value and a width
+    Args:
+        value (Decimal): a decimal value
+        width (int): A width for the returned Decimal instance
+    Returns:
+        Decimal: A truncated Decimal instance
+    """
+    unscaled_value = decimal_to_unscaled(value)
+    applied_value = unscaled_value - (((unscaled_value % width) + width) % width)
+    return Decimal(f"{applied_value}e{value.as_tuple().exponent}")
```
rdblue, line 90: This should create a decimal from the truncated unscaled value, rather than parsing.

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,86 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+    def __init__(self, source_type: IcebergType, width: int):
+        super().__init__(
+            f"truncate[{width}]",
+            f"transforms.truncate(source_type={repr(source_type)}, width={width})",
+        )
+        self._type = source_type
+        self._width = width
+
+    @property
+    def width(self):
```
Fokko, line 322:
```suggestion
    def width(self) -> int:
```

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,86 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+
+    def apply(self, value: Optional[S]) -> Optional[S]:
+        return self._truncate_value(value) if value is not None else None
+
+    @singledispatchmethod
+    def _truncate_value(self, value: S) -> S:
+        raise ValueError(f"Cannot truncate value: {value}")
+
+    @_truncate_value.register(int)
+    def _(self, value):
```
Fokko, line 333:
```suggestion
    def _(self, value: int) -> int:
```

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,86 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+        """Truncate a given int value into a given width if feasible."""
+        if type(self._type) in {IntegerType, LongType}:
+            return value - value % self._width
+        else:
+            raise ValueError(f"Cannot truncate type: {self._type} for value: {value}")
+
+    @_truncate_value.register(str)
+    def _(self, value):
```
Fokko, line 341:
```suggestion
    def _(self, value: str) -> str:
```

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,86 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+        """Truncate a given string to a given width."""
+        return value[0 : min(self._width, len(value))]
+
+    @_truncate_value.register(bytes)
+    def _(self, value):
```
Fokko, line 346:
```suggestion
    def _(self, value: bytes) -> bytes:
```

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,86 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+        return self._truncate_value(value) if value is not None else None
+
+    @singledispatchmethod
+    def _truncate_value(self, value: S) -> S:
+        raise ValueError(f"Cannot truncate value: {value}")
+
+    @_truncate_value.register(int)
+    def _(self, value):
+        """Truncate a given int value into a given width if feasible."""
+        if type(self._type) in {IntegerType, LongType}:
```
Fokko, line 335: Having the validation in processing itself feels a bit weird to me, shouldn't we check this when initializing the transform?

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,86 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+    @_truncate_value.register(bytes)
+    def _(self, value):
+        """Truncate a given binary bytes into a given width."""
+        if isinstance(self._type, BinaryType):
+            return value[0 : min(self._width, len(value))]
+        else:
+            raise ValueError(f"Cannot truncate type: {self._type}")
+
+    @_truncate_value.register(Decimal)
+    def _(self, value):
```
Fokko, line 354:
```suggestion
    def _(self, value: Decimal) -> Decimal:
```

Fokko (commented):
Thanks for picking this up @jun-he Some small comments below 👍🏻

python/src/iceberg/utils/decimal.py
jun-he: changed it to use `unscaled_to_decimal(applied_value, -value.as_tuple().exponent)`

python/src/iceberg/transforms.py
jun-he: Updated.

python/src/iceberg/transforms.py
jun-he: The validation here is to catch the case that the caller passes an int into Non-IntegerType, e.g. StringType truncate transformer. But the initialization might not the value type.

python/src/iceberg/transforms.py
jun-he: @Fokko I tried it before but mypy threw an error
```
an python/src/iceberg/transforms.py:332: error: Argument 1 has incompatible type "Callable[[TruncateTransform[S], int], int]"; expected "Callable[..., S]"
```
Wondering if there is a way to solve it?

python/src/iceberg/transforms.py
jun-he: Similar error here that
```
python/src/iceberg/transforms.py:340: error: Argument 1 has incompatible type "Callable[[TruncateTransform[S], str], str]"; expected "Callable[..., S]"
```

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,91 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+    @property
+    def width(self) -> int:
+        return self._width
+
+    @property
+    def type(self) -> IcebergType:
+        return self._type
+
+    def apply(self, value: Optional[S]) -> Optional[S]:
+        return self._truncate_value(value) if value is not None else None
```
Fokko, line 330: Nit: I'm just trying to popularize the Walrus operator:
```suggestion
        return truncated if (truncated := self._truncate_value(value)) else None
```

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,91 @@ def _(self, value_type: IcebergType, value: int) -> str:
         return datetime.to_human_timestamptz(value)
 
 
+class TruncateTransform(Transform[S, S]):
+    """A transform for truncating a value to a specified width.
+    Args:
+      source_type (Type): An Iceberg Type of IntegerType, LongType, StringType, BinaryType or DecimalType
```
Fokko, line 307: Nit:
```suggestion
      source_type (IcebergType): An Iceberg Type of IntegerType, LongType, StringType, BinaryType or DecimalType
```

python/src/iceberg/transforms.py
rdblue: Why not just `return self._truncate_value(value)`? If that's `None`, then returning it will return `None`.

rdblue: Actually, I see why: the `_truncate_value` method doesn't use `Optional`. So it should only be called if `value` is not `None`. I think the original is correct.

python/src/iceberg/transforms.py
rdblue: I don't think that we should do this check. We can assume that the type matches because of how we bind expressions.

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,91 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+    @_truncate_value.register(str)
+    def _(self, value):
+        """Truncate a given string to a given width."""
+        return value[0 : min(self._width, len(value))]
+
+    @_truncate_value.register(bytes)
+    def _(self, value):
+        """Truncate a given binary bytes into a given width."""
+        if isinstance(self._type, BinaryType):
+            return value[0 : min(self._width, len(value))]
```
rdblue, line 353: This is also correct:

```python
truncate(bytes("abc\u2603de", "utf-8")) # => b'abc\xe2\x98'
```

python/src/iceberg/utils/decimal.py
```diff
@@ -75,3 +75,16 @@ def decimal_to_bytes(value: Decimal) -> bytes:
...
+def truncate_decimal(value: Decimal, width: int) -> Decimal:
+    """Get a truncated Decimal value given a decimal value and a width
+    Args:
+        value (Decimal): a decimal value
+        width (int): A width for the returned Decimal instance
+    Returns:
+        Decimal: A truncated Decimal instance
+    """
+    unscaled_value = decimal_to_unscaled(value)
+    applied_value = unscaled_value - (((unscaled_value % width) + width) % width)
```
rdblue, line 89: This expression doesn't need to be as complex. The purpose of this is to handle negative numbers in languages where `%` will return a negative value. Python returns positive values:

```python
-4 % 5                   # => 1
((-4 % 5) + 5) % 5 # => 1
```

Note that if `-4 % 5` resulted in `-4` because |-4| < 5 then we would need the more complex expression.

python/tests/test_transforms.py
```diff
@@ -177,6 +177,61 @@ def test_identity_method(type_var):
...
+@pytest.mark.parametrize("input_var,expected", [("abcdefg", "abcde"), ("abc", "abc")])
+def test_truncate_string(input_var, expected):
+    trunc = transforms.truncate(StringType(), 5)
+    assert trunc.apply(input_var) == expected
+
+
+@pytest.mark.parametrize(
+    "type_var,value,expected_human_str,expected",
+    [
+        (BinaryType(), b"\x00\x01\x02\x03", "AAECAw==", b"\x00"),
```
rdblue, line 214: This should include the test cases I posted in comments above to validate that a `bytes` is truncated by the number of bytes and `str` is truncated by the number of unicode code points.

rdblue:
Thanks, @jun-he! I left a thorough review. I think this is close. The implementations look correct, but there are a few things to fix.

python/src/iceberg/transforms.py
kbendick: +1 to not validating inside of the right loop.

I’d like to see general validation of the `source_type` (though I guess singledispatch takes care of that somewhat, it would be nice to see it a bit more explicitly but that’s just my opinion),

Also validating in the constructor or elsewhere that the `width` is greater than zero.

But this function will be called in a tight loop and it’s best to avoid expensive checks here.

python/src/iceberg/transforms.py
```diff
@@ -301,6 +301,91 @@ def _(self, value_type: IcebergType, value: int) -> str:
...
+        """Truncate a given int value into a given width if feasible."""
+        if type(self._type) in {IntegerType, LongType}:
+            return value - value % self._width
+        else:
+            raise ValueError(f"Cannot truncate type: {self._type} for value: {value}")
+
+    @_truncate_value.register(str)
+    def _(self, value):
+        """Truncate a given string to a given width."""
+        return value[0 : min(self._width, len(value))]
```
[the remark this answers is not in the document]
kbendick: Some of the tests [at]huaxingao added somewhat recently for parquet bloom filter has Chinese characters that I believe are specifically multi-byte for this very purpose.

python/src/iceberg/transforms.py
jun-he: updated it using `singledispatch` here and then use Any for value.

python/src/iceberg/transforms.py
jun-he: Wondering the benefit to use Walrus operator in this case. It seems unnecessarily add a new variable `truncated`.

python/tests/test_transforms.py
jun-he: I added both tests with slightly change (removing `abc` prefix).

python/src/iceberg/transforms.py
```diff
@@ -267,39 +272,132 @@ def satisfies_order_of(self, other: Transform) -> bool:
...
 
-    @_int_to_human_string.register(TimestampType)
-    def _(self, _: IcebergType, value: int) -> str:
-        return datetime.to_human_timestamp(value)
 
-    @_int_to_human_string.register(TimestamptzType)
-    def _(self, _: IcebergType, value: int) -> str:
-        return datetime.to_human_timestamptz(value)
+@singledispatch
+def _human_string(value: Any, _type: IcebergType) -> str:
```
Fokko, line 335: Instead of having two singledispatches, we could also turn everything into one where we match on the type first. This would simplify the logic a bit.
```suggestion
def _human_string(_type: IcebergType, value: Any) -> str:
```

Fokko (approved):
One minor comment, but looks great @jun-he. Thanks!

rdblue (approved):
Looks good! I'm going to merge this to unblock the refactor and we can clean up the rest afterwards. Thanks, @jun-he!

PR #5033, opened 2022-06-14
Nessie: Fix drop table with TableReference in identifier

ajantha-bhat:
From `NessieCatalog` (initialized to `"main"` branch),

load table works with `TableReference` (aka **`tableName@branch`** syntax),
`catalog.loadTable(TableIdentifier.parse("db1.`table1.@branch1`"))` 

but drop table will NOT work with `TableReference`,
`catalog.dropTable(TableIdentifier.parse("db1.`table1.@branch1`"))` -- return false.

Reason: For drop table, key used to look up Nessie backend will have table reference which is not an actual key in Nessie backend. 

Fix: From `TableIdentifier` with `TableReference`, need to extract valid identifier to be used as a key for the drop table.


ajantha-bhat:
cc: @nastra, @snazy

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -187,7 +185,16 @@ public List<TableIdentifier> listTables(Namespace namespace) {
...
-    return client.dropTable(identifier, purge);
+    TableReference tableReference = parseTableReference(identifier);
+    TableIdentifier tableIdentifier;
+    if (tableReference.hasReference()) {
+      tableIdentifier = TableIdentifier.of(identifier.namespace(), tableReference.getName());
+    } else {
+      tableIdentifier = identifier;
+    }
+
+    NessieIcebergClient newClient = this.client.withReference(tableReference.getReference(), tableReference.getHash());
```
nastra, line 196: nit: maybe inline this

nastra (approved):

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
ajantha-bhat: done

nastra:
@RussellSpitzer could you merge this one please?

nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
```diff
@@ -187,7 +185,16 @@ public List<TableIdentifier> listTables(Namespace namespace) {
 
   @Override
   public boolean dropTable(TableIdentifier identifier, boolean purge) {
-    return client.dropTable(identifier, purge);
+    TableReference tableReference = parseTableReference(identifier);
```
pvary, line 188: Don't we have the same issue with `renameTable`?

ajantha-bhat: Good catch. I saw it during debugging. But forgot to handle. 
Handled now.

pvary (approved):
+1 from me
I will give this some time, so if some want to chime in, they can comment, then if there are no more comments I will merge

snazy (approved):

pvary:
Merged to master.
Thanks for the work @ajantha-bhat and @snazy and @nastra for the review!

ajantha-bhat:
@pvary : Thanks for the review.

PR #5034, opened 2022-06-14
Core: Fix tag ancestor snapshot handling

rdblue:
This fixes the expiration for snapshots that are ancestors of tags. Tag ancestors should not be considered "referenced" snapshots and should be cleaned up using the global table settings.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
+++ b/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java
@@ -273,8 +273,12 @@ class RemoveSnapshots implements ExpireSnapshots {
   private Set<Long> unreferencedSnapshotsToRetain(Collection<SnapshotRef> refs) {
     Set<Long> referencedSnapshots = Sets.newHashSet();
     for (SnapshotRef ref : refs) {
-      for (Snapshot snapshot : SnapshotUtil.ancestorsOf(ref.snapshotId(), base::snapshot)) {
-        referencedSnapshots.add(snapshot.snapshotId());
+      if (ref.isBranch()) {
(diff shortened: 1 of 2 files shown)


rdblue:
@amogh-jahagirdar, can you take a look?

amogh-jahagirdar:
Thanks for catching this I missed this in https://github.com/amogh-jahagirdar/iceberg/pull/1/files

Looks like there are some checkstyle errors in TestRemoveSnapshots but other than that looks good to me!

[Error: Checkstyle] [ERROR] /home/runner/work/iceberg/iceberg/core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java:34:1: Extra separation in import group before 'java.io.IOException' [ImportOrder]
Error: eckstyle] [ERROR] /home/runner/work/iceberg/iceberg/core/src/test/java/org/apache/iceberg/TestRemoveSnapshots.java:34:1: Wrong order for 'java.io.IOException' import. [ImportOrder]

namrathamyske:
@rdblue Why do we need to set `setMaxSnapshotAgeMs("branch", Long.MAX_VALUE)` in the test case ? anyway the snapshots still referenced by a branch will be added to referenced snapshots.

rdblue:
@namrathamyske, when a ref doesn't have a max snapshot age, the branch will use the `expireOlderThan` timestamp. That timestamp is set so that the snapshot would be aged off, except that it is referenced by the branch with effectively infinite snapshot age. We could set something smaller, but since the point is to make the branch protect that snapshot, using the highest possible age seems fine.

danielcweeks (approved):
+1 LGTM

PR #5036, opened 2022-06-14
Hive: Fix concurrent transactions overwriting commits by adding hive lock heartbeats.

SinghAsDev:
When using Hive Metastore catalog, it is possible that longer transactions can take longer than [transaction timeout](https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties#ConfigurationProperties-hive.txn.timeout) of the Hive Metastore (hive.txn.timeout or metastore.txn.timeout in the newer versions). In those cases, concurrent transactions can overwrite each other leading to data corruption. This change fixes concurrent transactions overwriting commits by adding hive lock heartbeats.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -31,6 +31,9 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
@@ -74,6 +77,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableBiMap;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.relocated.com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.iceberg.util.JsonUtil;
 import org.apache.iceberg.util.Tasks;
 import org.apache.thrift.TException;
@@ -90,6 +94,10 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private static final String HIVE_ACQUIRE_LOCK_TIMEOUT_MS = "iceberg.hive.lock-timeout-ms";
   private static final String HIVE_LOCK_CHECK_MIN_WAIT_MS = "iceberg.hive.lock-check-min-wait-ms";
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
+  private static final String HIVE_LOCK_HEARTBEAT_INTERVAL_MS =
+      "iceberg.hive.lock-heartbeat-interval-ms";
+  private static final String HIVE_LOCK_HEARTBEAT_TIMEOUT_MS =
+      "iceberg.hive.lock-heartbeat-timeout-ms";
   private static final String HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES =
       "iceberg.hive.metadata-refresh-max-retries";
   private static final String HIVE_TABLE_LEVEL_LOCK_EVICT_MS =
@@ -104,6 +112,7 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private static final long HIVE_ACQUIRE_LOCK_TIMEOUT_MS_DEFAULT = 3 * 60 * 1000; // 3 minutes
   private static final long HIVE_LOCK_CHECK_MIN_WAIT_MS_DEFAULT = 50; // 50 milliseconds
   private static final long HIVE_LOCK_CHECK_MAX_WAIT_MS_DEFAULT = 5 * 1000; // 5 seconds
+  private static final long HIVE_LOCK_HEARTBEAT_INTERVAL_MS_DEFAULT = 4 * 60 * 1000; // 4 minutes
   private static final int HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES_DEFAULT = 2;
   private static final long HIVE_TABLE_LEVEL_LOCK_EVICT_MS_DEFAULT = TimeUnit.MINUTES.toMillis(10);
   private static final BiMap<String, String> ICEBERG_TO_HMS_TRANSLATION =
@@ -153,10 +162,13 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private final long lockAcquireTimeout;
   private final long lockCheckMinWaitTime;
   private final long lockCheckMaxWaitTime;
+  private final long lockHeartbeatIntervalTime;
   private final long maxHiveTablePropertySize;
   private final int metadataRefreshMaxRetries;
   private final FileIO fileIO;
   private final ClientPool<IMetaStoreClient, TException> metaClients;
+  private final ScheduledExecutorService exitingScheduledExecutorService;
+  private HiveLockHeartbeat hiveLockHeartbeat;
 
   protected HiveTableOperations(
       Configuration conf,
@@ -177,6 +189,8 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
         conf.getLong(HIVE_LOCK_CHECK_MIN_WAIT_MS, HIVE_LOCK_CHECK_MIN_WAIT_MS_DEFAULT);
     this.lockCheckMaxWaitTime =
         conf.getLong(HIVE_LOCK_CHECK_MAX_WAIT_MS, HIVE_LOCK_CHECK_MAX_WAIT_MS_DEFAULT);
+    this.lockHeartbeatIntervalTime =
+        conf.getLong(HIVE_LOCK_HEARTBEAT_INTERVAL_MS, HIVE_LOCK_HEARTBEAT_INTERVAL_MS_DEFAULT);
     this.metadataRefreshMaxRetries =
         conf.getInt(
             HIVE_ICEBERG_METADATA_REFRESH_MAX_RETRIES,
@@ -185,6 +199,12 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
         conf.getLong(HIVE_TABLE_PROPERTY_MAX_SIZE, HIVE_TABLE_PROPERTY_MAX_SIZE_DEFAULT);
     long tableLevelLockCacheEvictionTimeout =
         conf.getLong(HIVE_TABLE_LEVEL_LOCK_EVICT_MS, HIVE_TABLE_LEVEL_LOCK_EVICT_MS_DEFAULT);
+    this.exitingScheduledExecutorService =
+        Executors.newSingleThreadScheduledExecutor(
+            new ThreadFactoryBuilder()
+                .setDaemon(true)
+                .setNameFormat("iceberg-hive-lock-heartbeat-%d")
+                .build());
     initTableLevelLockCache(tableLevelLockCacheEvictionTimeout);
   }
 
@@ -243,9 +263,12 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
     // JVM process, which would result in unnecessary and costly HMS lock acquisition requests
     ReentrantLock tableLevelMutex = commitLockCache.get(fullName, t -> new ReentrantLock(true));
     tableLevelMutex.lock();
+    HiveLockHeartbeat hiveLockHeartbeat = null;
     try {
       lockId = Optional.of(acquireLock());
-      // TODO add lock heart beating for cases where default lock timeout is too low.
+      hiveLockHeartbeat =
+          new HiveLockHeartbeat(metaClients, lockId.get(), lockHeartbeatIntervalTime);
+      hiveLockHeartbeat.schedule(exitingScheduledExecutorService);
 
       Table tbl = loadHmsTable();
 
@@ -296,13 +319,33 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
       }
 
       try {
+        if (hiveLockHeartbeat.future.isCancelled()
+            || hiveLockHeartbeat.encounteredException != null) {
+          throw new CommitFailedException(
+              "Failed to heartbeat for hive lock. %s",
+              hiveLockHeartbeat.encounteredException.getMessage());
+        }
+
         persistTable(tbl, updateHiveTable);
+        if (hiveLockHeartbeat.future.isCancelled()
+            || hiveLockHeartbeat.encounteredException != null) {
+          throw new CommitStateUnknownException(
+              "Failed to heartbeat for hive lock while "
+                  + "committing changes. This can lead to a concurrent commit attempt be able to overwrite this commit. "
+                  + "Please check the commit history. If you are running into this issue, try reducing "
+                  + "iceberg.hive.lock-heartbeat-interval-ms.",
+              hiveLockHeartbeat.encounteredException);
+        }
+
         commitStatus = CommitStatus.SUCCESS;
       } catch (org.apache.hadoop.hive.metastore.api.AlreadyExistsException e) {
         throw new AlreadyExistsException(e, "Table already exists: %s.%s", database, tableName);
 
       } catch (InvalidObjectException e) {
         throw new ValidationException(e, "Invalid Hive object for %s.%s", database, tableName);
+
+      } catch (CommitFailedException | CommitStateUnknownException e) {
+        throw e;
 
       } catch (Throwable e) {
         if (e.getMessage() != null
@@ -338,6 +381,9 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
       throw new RuntimeException("Interrupted during commit", e);
 
     } finally {
+      if (hiveLockHeartbeat != null) {
+        hiveLockHeartbeat.cancel();
+      }
       cleanupMetadataAndUnlock(commitStatus, newMetadataLocation, lockId, tableLevelMutex);
     }
 
@@ -363,7 +409,8 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
     }
   }
 
-  private Table loadHmsTable() throws TException, InterruptedException {
+  @VisibleForTesting
+  Table loadHmsTable() throws TException, InterruptedException {
     try {
       return metaClients.run(client -> client.getTable(database, tableName));
     } catch (NoSuchObjectException nte) {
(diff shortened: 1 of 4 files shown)


pvary:
@SinghAsDev: Quick question: Did you face a specific issue around this, or we are just fixing a long outstanding `TODO` here?

I am asking, because I have a feeling that the need for heartbeat is more like a theoretical issue  than a concrete one. The code is like this:
```
lock
loadTable
if needed createTable
some object value settings
persistTable
```

The `acquireLock` part should heartbeat the lock until the lock is acquired, and the other parts should be fast.

I was trying to find situations where we need heartbeat, and I was able to come up with these:
- Wrong configuration: Too low heartbeat interval configured on the HMS
- JVM pause on the client, or the HMS side

For the JVM pause situations the problem is still unlikely to happen (even without the heartbeat), and the heartbeating thread is also suspended in this case, so there is no guarantees that it will help.

Since Hive 4 is around the release, we might try to add a `checkAndSet` method to the HMS API for setting the `metadata_location` atomically. This could be the ultimate solution for the issue, if we find this fix important.

All that said: using the default configuration on the HMS, the timeout is set to 300s. If we can not finish the commit in 5 mins, then we already are in serious trouble.

SinghAsDev:
Hi @pvary we did hit the issue that led to data corruption. I think a lot of platforms are still going to remain on older version of HMS, and may not want to increase the txn timeout.

Tbh, the decision to succeed the job instead of rolling back the transaction or at least failing the job with failed to unlock exception would have been better. Many users may be hitting this issue without even realizing.

SinghAsDev:
> > For instance, HMS does a fast scan for alter table operations even when stats collection is turned off
> 
> 
> 
> That's seriously bad and unnecessary. This would cause a seriously long commit times.
> 
> Which version of Hive you are using?

We are on hive 1.2. But I think the issue still exists in later versions. I agree this is more of a bug. We had fixed this for our hms deployment. Btw, forgot to mention this issue is limited to unpartitioned tables. All iceberg tables are stored as unpartitioned tables on HMS, so iceberg tables are affected by this.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -229,7 +241,8 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     tableLevelMutex.lock();
     try {
       lockId = Optional.of(acquireLock());
-      // TODO add lock heart beating for cases where default lock timeout is too low.
+      hiveLockHeartbeat = new HiveLockHeartbeat(metaClients, lockId.get(), lockCheckMaxWaitTime);
```
pvary, line 244: This is how Hive schedules the heartbeat:
```
HiveConf.getTimeVar(conf, HiveConf.ConfVars.HIVE_TXN_TIMEOUT, TimeUnit.MILLISECONDS) / 2;
```

pvary:
> > > For instance, HMS does a fast scan for alter table operations even when stats collection is turned off
> > 
> > 
> > That's seriously bad and unnecessary. This would cause a seriously long commit times.
> > Which version of Hive you are using?
> 
> We are on hive 1.2. But I think the issue still exists in later versions. I agree this is more of a bug. We had fixed this for our hms deployment. Btw, forgot to mention this issue is limited to unpartitioned tables. All iceberg tables are stored as unpartitioned tables on HMS, so iceberg tables are affected by this.

After fixing this issue, you still see long `alter table` HMS calls? I am trying to understand the situation better, and possible other places where we have to fix too

SinghAsDev:
> After fixing this issue, you still see long alter table HMS calls? I am trying to understand the situation better, and possible other places where we have to fix too

Yea, though likely it won't be as frequent anymore. I think the point I am trying to make is that we can not guarantee that Iceberg commit would be able finish within hive.txn.timeout. You and I both listed a few reasons on why that can happen. I think heartbeats on locks is a must have, or we should fail the commit/ rollback the commit if unlock fails with lock not found. I prefer the former as it would really be a waste of compute to simply fail or rollback.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
SinghAsDev: Yea, but Iceberg does not have this config, so we can't use that. The reason I used `iceberg.hive.lock-check-max-wait-ms` is that we already advise users to keep this lower than `hive.txn.timeout`. I would be open to making this even smaller to something like `lockCheckMaxWaitTime / 2` to allow room for delays in heartbeats and/ or retries. Thoughts @pvary ?

pvary: I would prefer a different config for the heartbeat period

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveMetastore.java
```diff
@@ -242,6 +242,7 @@ private void initConf(HiveConf conf, int port) {
     conf.set(HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL.varname, "false");
     conf.set(HiveConf.ConfVars.METASTORE_DISALLOW_INCOMPATIBLE_COL_TYPE_CHANGES.varname, "false");
     conf.set("iceberg.hive.client-pool-size", "2");
+    conf.set(HiveConf.ConfVars.HIVE_TXN_TIMEOUT.varname, "1s");
```
pvary, line 245: Is it intentional to change this for all of the tests?
I would keep this as a default here, and modify the config only in the `TestHiveComitLocks` test

pvary:
> > After fixing this issue, you still see long alter table HMS calls? I am trying to understand the situation better, and possible other places where we have to fix too
> 
> Yea, though likely it won't be as frequent anymore. I think the point I am trying to make is that we can not guarantee that Iceberg commit would be able finish within hive.txn.timeout. You and I both listed a few reasons on why that can happen. I think heartbeats on locks is a must have, or we should fail the commit/ rollback the commit if unlock fails with lock not found. I prefer the former as it would really be a waste of compute to simply fail or rollback.

You have convinced me here, I just trying to collect the things that we have to fix in Hive to make this work as fast as possible 😄

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -631,4 +647,40 @@ private static boolean hiveEngineEnabled(TableMetadata metadata, Configuration c
...
+
+    @Override
+    public void run() {
+      try {
+        hmsClients.run(client -> {
+          client.heartbeat(0, lockId);
+          return null;
+        });
+      } catch (TException | InterruptedException e) {
+        LOG.error("Fail to heartbeat for lock: {}", lockId, e);
```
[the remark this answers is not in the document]
SinghAsDev: Thanks, I also think we should fail the commit on heart beat failure. However, to do so we would have to add retry on heart beats as well, so that transient failures don't fail the entire job. I will add that.

For the unlock failure handling, I think that needs a bit more discussion and if it is OK with you let's do that discussion in a follow up PR.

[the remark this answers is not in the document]
pvary: RetryingMetastoreClient does the retry for us. No need to do the retries by hand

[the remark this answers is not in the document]
SinghAsDev: Oh yea, forgot about the niceties of `RetryingMetastoreClient`. Thanks!

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCommitLocks.java
```diff
@@ -72,10 +77,17 @@ public class TestHiveCommitLocks extends HiveTableBaseTest {
   LockResponse notAcquiredLockResponse = new LockResponse(dummyLockId, LockState.NOT_ACQUIRED);
 
   @BeforeClass
-  public static void initializeSpies() throws Exception {
+  public static void startMetastore() throws Exception {
+    HiveMetastoreTest.hiveConf = new HiveConf();
+    HiveMetastoreTest.hiveConf.set(HiveConf.ConfVars.HIVE_TXN_TIMEOUT.varname, "1s");
+    HiveMetastoreTest.startMetastore();
```
pvary, line 83: Could we create a:
```
HiveMetastoreTest.startMetastore(Map<String, String> override) {
}
```
in the `HiveMetastoreTest` class? It might be more readable than this?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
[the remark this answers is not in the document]
pvary: @SinghAsDev: We can do some elaborate stuff to throw `CommitFailedException` if we are before the `alter` operation, and we can throw `CommitStateUnknownException` if we called the `alter` command already, but I hope this would be a rare occurrence, so it might not worth the extra complexity. The question is how we will be able to communicate back to the original thread that the heartbeat is failed. Based on this the decision above could be an easy one, or a more complex one, and then we can decide which one to chose.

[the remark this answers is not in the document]
SinghAsDev: @pvary I agree that this added complexity does not really buy us much. All the lengthy operations I have seen that needed heartbeats are within the alter table operation, so very unlikely we will see heartbeat failure before alter operation. After alter table is done, any issue with lock, we will anyway see in the unlock operation. As such, I still think we likely don't need the added complexity. If we agree on this reasoning, I can add this info in the code as well.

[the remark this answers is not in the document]
pvary: Ok. So we would likely to need a test where the failing heartbeat will fail the commit with the `CommitStateUnknownException`. So It is fine if we have `CommitStateUnknownException` for all of these cases

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -631,4 +652,41 @@ private static boolean hiveEngineEnabled(TableMetadata metadata, Configuration c
...
+          client.heartbeat(0, lockId);
+          return null;
+        });
+      } catch (TException | InterruptedException e) {
+        throw new CommitFailedException(e, "Fail to heartbeat for lock: %d", lockId);
+      }
+    }
+
+    public void schedule(ScheduledExecutorService scheduler) {
+      future =
```
szehon-ho, line 682: Nit: I think most of the code does not newline so soon, maybe we can newline in the arguments of the scheduleAtFixedRate.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -631,4 +652,41 @@ private static boolean hiveEngineEnabled(TableMetadata metadata, Configuration c
...
+      this.intervalMs = intervalMs;
+      this.future = null;
+    }
+
+    @Override
+    public void run() {
+      try {
+        hmsClients.run(client -> {
+          client.heartbeat(0, lockId);
+          return null;
```
szehon-ho, line 674: Is the code block necessary?

 client -> client.heartbeat(0, lockId);

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveMetastoreTest.java
```diff
@@ -43,8 +43,12 @@ public abstract class HiveMetastoreTest {
   @BeforeClass
   public static void startMetastore() throws Exception {
     HiveMetastoreTest.metastore = new TestHiveMetastore();
-    metastore.start();
-    HiveMetastoreTest.hiveConf = metastore.hiveConf();
+    if (HiveMetastoreTest.hiveConf != null) {
```
szehon-ho, line 46: Trying to understand, how would this happen?

szehon-ho:
> Since Hive 4 is around the release, we might try to add a checkAndSet method to the HMS API for setting the metadata_location atomically. This could be the ultimate solution for the issue, if we find this fix important.

I think that would be awesome.

rdblue:
The 0.14.0 release is coming soon. I'll try to include this if it is done in time.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
SinghAsDev: I think it is as we need to return null. Let me know if there is a way to make it work without code block.

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveMetastoreTest.java
SinghAsDev: As this is happening before class is actually instantiated `HiveMetastoreTest.hiveConf` can be null. I actually saw this in the unit test I added as part of this PR.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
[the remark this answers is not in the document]
SinghAsDev: Added a test. Can you help take a look @pvary @szehon-ho , thanks!

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -274,14 +292,27 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+          throw new CommitFailedException("Failed to heartbeat for hive lock. %s",
+              hiveLockHeartbeat.encounteredException.getMessage());
+        }
         persistTable(tbl, updateHiveTable);
+        if (hiveLockHeartbeat.future.isCancelled() || hiveLockHeartbeat.encounteredException != null) {
+          throw new CommitStateUnknownException("Failed to heartbeat for hive lock while " +
+              "committing changes. This can lead to a concurrent commit attempt be able to overwrite this commit. " +
+              "Please check the commit history. If you are running into this issue, try reducing " +
+              "iceberg.hive.lock-heartbeat-interval-ms.", hiveLockHeartbeat.encounteredException);
+        }
```
pvary, line 305: nit: empty line after block

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -149,10 +155,14 @@ private static class WaitingForLockException extends RuntimeException {
...
   private final long lockCheckMinWaitTime;
   private final long lockCheckMaxWaitTime;
+  private final long lockHeartbeatIntervalTime;
   private final long maxHiveTablePropertySize;
   private final int metadataRefreshMaxRetries;
   private final FileIO fileIO;
   private final ClientPool<IMetaStoreClient, TException> metaClients;
+  private final ScheduledExecutorService exitingScheduledExecutorService;
+
+  private HiveLockHeartbeat hiveLockHeartbeat;
```
pvary, line 165: Could this be local to the method where it is used?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -632,4 +666,43 @@ private static boolean hiveEngineEnabled(TableMetadata metadata, Configuration c
...
+    @Override
+    public void run() {
+      try {
+        hmsClients.run(client -> {
+          client.heartbeat(0, lockId);
+          return null;
+        });
+      } catch (TException | InterruptedException e) {
+        this.encounteredException = e;
+        throw new CommitFailedException(e, "Failed to heartbeat for lock: %d", lockId);
```
pvary, line 693: Is this exception that we throw here used anywhere?

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveMetastoreTest.java
```diff
@@ -42,9 +44,21 @@ public abstract class HiveMetastoreTest {
 
   @BeforeClass
   public static void startMetastore() throws Exception {
+    startMetastore(Collections.emptyMap());
+  }
```
pvary, line 48: nit: newline

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveMetastoreTest.java
```diff
@@ -42,9 +44,21 @@ public abstract class HiveMetastoreTest {
...
-    HiveMetastoreTest.hiveConf = metastore.hiveConf();
+    if (HiveMetastoreTest.hiveConf != null) {
+      if (hiveConfOverride != null) {
+        for (Map.Entry<String, String> kv : hiveConfOverride.entrySet()) {
+          hiveConf.set(kv.getKey(), kv.getValue());
+        }
+      }
+      metastore.start(HiveMetastoreTest.hiveConf);
+    } else {
+      metastore.start();
```
pvary, line 59: We still would like to apply the `hiveConfOverride` here.
I would guess that it would be easier to read if we calculate the `hiveConf` first, and then start the metastore with the final conf

hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveCommitLocks.java
```diff
@@ -72,10 +79,15 @@ public class TestHiveCommitLocks extends HiveTableBaseTest {
   LockResponse notAcquiredLockResponse = new LockResponse(dummyLockId, LockState.NOT_ACQUIRED);
 
   @BeforeClass
-  public static void initializeSpies() throws Exception {
+  public static void startMetastore() throws Exception {
+    HiveMetastoreTest.startMetastore(Collections.singletonMap(HiveConf.ConfVars.HIVE_TXN_TIMEOUT.varname, "1s"));
```
pvary, line 83: I think in the Iceberg code we usually use: `org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap.of`

pvary:
@SinghAsDev: Could you please check why the test is failing? Also left some minor comments.

Thanks,
Peter

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
SinghAsDev: It is not, but throwing exception here will stop the future runs of this runnable.

SinghAsDev:
> @SinghAsDev: Could you please check why the test is failing? Also left some minor comments.
> 
> 
> 
> Thanks,
> 
> Peter

@pvary mind looking again, thanks!

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveMetastoreTest.java
```diff
@@ -41,10 +44,21 @@ public abstract class HiveMetastoreTest {
 
   @BeforeClass
   public static void startMetastore() throws Exception {
+    startMetastore(Collections.emptyMap());
+  }
+
+  public static void startMetastore(Map<String, String> hiveConfOverride) throws Exception {
     HiveMetastoreTest.metastore = new TestHiveMetastore();
-    metastore.start();
+    HiveConf hiveConfWithOverrides = new HiveConf(new Configuration(), TestHiveMetastore.class);
```
pvary, line 52: Is there a special reason behind not using `new HiveConf(TestHiveMetastore.class)` here?

pvary (approved):

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -90,6 +94,10 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
   private static final String HIVE_ACQUIRE_LOCK_TIMEOUT_MS = "iceberg.hive.lock-timeout-ms";
   private static final String HIVE_LOCK_CHECK_MIN_WAIT_MS = "iceberg.hive.lock-check-min-wait-ms";
   private static final String HIVE_LOCK_CHECK_MAX_WAIT_MS = "iceberg.hive.lock-check-max-wait-ms";
+  private static final String HIVE_LOCK_HEARTBEAT_INTERVAL_MS =
+      "iceberg.hive.lock-heartbeat-interval-ms";
+  private static final String HIVE_LOCK_HEARTBEAT_TIMEOUT_MS =
```
szehon-ho, line 99: Is this unused?

hive-metastore/src/test/java/org/apache/iceberg/hive/HiveMetastoreTest.java
SinghAsDev: Not really, I just moved this line from `TestHiveMetastore`. Updated to remove this though.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
SinghAsDev: Thanks for catching, it's not needed anymore. Removed.

SinghAsDev:
@pvary @szehon-ho @rdblue I think this is ready now? The one test failure is not related to this diff. I will force push to re-trigger the checks.

szehon-ho (approved):

pvary:
Thanks @SinghAsDev for the PR and @szehon-ho for the review!

PR #5039, opened 2022-06-14
Hive: Print db and table name while acquiring hive meta-store lock

krisdas:
While acquiring hive-metastore lock on an Iceberg table, after every timeout, below [log line](https://github.com/apache/iceberg/blob/98dc9fe8724375e8b932e7e8afdb925c897b4695/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L542) is printed, which doesn't have database and table name information. 

`org.apache.iceberg.hive.HiveTableOperations$WaitingForLockException: Waiting for lock.`

After exhausting all the re-try attempt, finally below [log line](https://github.com/apache/iceberg/blob/98dc9fe8724375e8b932e7e8afdb925c897b4695/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L561) prints database and table name.

`Retrying task after failure: Timed out after 180133 ms waiting for lock on database.table`

Here we are adding the database and table name in first log line to speed up investigation related with locking.

cc : @szehon-ho @dramaticlly

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -74,6 +74,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.util.JsonUtil;
+import org.apache.iceberg.util.Pair;
 import org.apache.iceberg.util.Tasks;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
@@ -512,6 +513,7 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
     LockResponse lockResponse = metaClients.run(client -> client.lock(lockRequest));
     AtomicReference<LockState> state = new AtomicReference<>(lockResponse.getState());
     long lockId = lockResponse.getLockid();
+    Pair<Long, String> lockDetails = Pair.of(lockId, String.format("%s.%s", database, tableName));
 
     final long start = System.currentTimeMillis();
     long duration = 0;
@@ -524,7 +526,7 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
         // upper bound of retries. So it is just reasonable to set a large retry count. However, if we set
         // Integer.MAX_VALUE, the above logic of `retries + 1` would overflow into Integer.MIN_VALUE. Hence,
         // the retry is set conservatively as `Integer.MAX_VALUE - 100` so it doesn't hit any boundary issues.
-        Tasks.foreach(lockId)
+        Tasks.foreach(lockDetails)
             .retry(Integer.MAX_VALUE - 100)
             .exponentialBackoff(
                 lockCheckMinWaitTime,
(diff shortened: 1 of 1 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -512,6 +513,7 @@ long acquireLock() throws UnknownHostException, TException, InterruptedException
     LockResponse lockResponse = metaClients.run(client -> client.lock(lockRequest));
     AtomicReference<LockState> state = new AtomicReference<>(lockResponse.getState());
     long lockId = lockResponse.getLockid();
+    Pair<Long, String> lockDetails = Pair.of(lockId, String.format("%s.%s", database, tableName));
```
dramaticlly, line 516: nit: can make it final

dramaticlly (approved):
LGTM

abmo-x:
Looking at https://github.com/apache/iceberg/blob/a5b2c703fb810a6d0731c18056cf287358d15e6c/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L563, it should be logging the database.tablename if lock is not acquired in a timely manner. 

is that exception getting swallowed somewhere or the msg not logged from those exception which include these details along with the state of the lock.

krisdas:
> Question:
> 
> https://github.com/apache/iceberg/blob/a5b2c703fb810a6d0731c18056cf287358d15e6c/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L563
> 
> 
> And
> https://github.com/apache/iceberg/blob/a5b2c703fb810a6d0731c18056cf287358d15e6c/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L568
> 
> These should be logging the database.tablename if lock is not acquired in a timely manner.
> 
> is that exception getting swallowed somewhere or the msg not logged from the exception, which includes these details along with the state of the lock.

These logs gets printed, just that, they are printed at the very end, after exhausting all the retries. Here we are trying to add db.table details after each re-try and get the information up in the log file. Currently with just "Waiting for lock" message, we don't get details about, which db.table, the lock is requested.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
krisdas: done

abmo-x:
> > Question:
> > https://github.com/apache/iceberg/blob/a5b2c703fb810a6d0731c18056cf287358d15e6c/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L563
> > 
> > And
> > https://github.com/apache/iceberg/blob/a5b2c703fb810a6d0731c18056cf287358d15e6c/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L568
> > 
> > These should be logging the database.tablename if lock is not acquired in a timely manner.
> > is that exception getting swallowed somewhere or the msg not logged from the exception, which includes these details along with the state of the lock.
> 
> These logs gets printed, just that, they are printed at the very end, after exhausting all the retries. Here we are trying to add db.table details after each re-try and get the information up in the log file. Currently with just "Waiting for lock" message, we don't get details about, which db.table, the lock is requested.

Ah! That makes sense, thanks for clarifying.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -512,6 +513,7 @@ long acquireLock() throws UnknownHostException, TException, InterruptedException
     LockResponse lockResponse = metaClients.run(client -> client.lock(lockRequest));
     AtomicReference<LockState> state = new AtomicReference<>(lockResponse.getState());
     long lockId = lockResponse.getLockid();
+    final Pair<Long, String> lockDetails = Pair.of(lockId, String.format("%s.%s", database, tableName));
```
pvary, line 516: I do not think we need this `Pair` object here. Could we just use database/tableName down below?

pvary: I think the `Tasks.foreach(lockDetails)` is only used to reuse the retry functionality of `Tasks`. So I think it is perfectly fine to keep the simpler implementation and using database/tableName.

pvary (approved):
+1 pending tests

szehon-ho (approved):

pvary:
Thanks @krisdas for the PR, and for the team for the review comments!

PR #5045, opened 2022-06-15
Docs: Fix Flink Connector docs with custom catalog

nastra:


docs/flink-connector.md
```diff
@@ -35,10 +35,8 @@ To create the table in Flink SQL by using SQL syntax `CREATE TABLE test (..) WIT
 
 * `connector`: Use the constant `iceberg`.
 * `catalog-name`: User-specified catalog name. It's required because the connector don't have any default value.
-* `catalog-type`: Default to use `hive` if don't specify any value. The optional values are:
-    * `hive`: The Hive metastore catalog.
-    * `hadoop`: The hadoop catalog.
-    * `custom`: The customized catalog, see [custom catalog](../custom-catalog) for more details.
+* `catalog-type`: `hive` or `hadoop` for built-in catalogs (defaults to `hive`), or left unset for custom catalog implementations using `catalog-impl`.
+* `catalog-impl`: The fully-qualified class name custom catalog implementation, must be set if `catalog-type` is unset. See also [custom catalog](../flink/flink-getting-started.md#custom-catalog) for more details.
```
Reo-LEI, line 39: I think `catalog-impl` is an optional and most of the time users don't need to set it. Should we indicate that this configuration is optional and indicate the default value?

docs/flink-connector.md
```diff
@@ -107,7 +105,6 @@ CREATE TABLE flink_table (
 ) WITH (
     'connector'='iceberg',
     'catalog-name'='custom_prod',
-    'catalog-type'='custom',
```
Reo-LEI, line 110: Why we need to delete this?

nastra: because the example as it's stated here doesn't work

Reo-LEI: You are right, user will fail for this validation. We should delete this.

https://github.com/apache/iceberg/blob/dd5b35067baac5eb004141d51d32251d2f8e808c/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalogFactory.java#L85-L90

docs/flink-connector.md
nastra: I think it should be clear from **must be set if `catalog-type` is unset**  that this is optional

kbendick: I agree it should be clear already that one or the other needs to be set from the language already.

kbendick, line 39: Nit: “The fully qualified class name _of a_ custom catalog implementation”. The addition of `of a` makes the sentence a bit more fluent.

Then I’d start a new sentence for “Must be set if `catalog-type` is unset”. But that part might just be my own opinion, but it reads a bit more clearly to me.

docs/flink-connector.md
kbendick: +1 to not having this configuration combination which can’t exist.

Should we also update the sentence on line 99 above which mentions that this “maps to the iceberg table `default_database.flink_table` managed in custom catalog.”? Specifically maybe mentioning that the table is “managed in a custom catalog of type com.my.custom.CatalogImpl” instead?

kbendick (approved):
Thanks for this @nastra.

I left a small nit and two optional things to consider, but overall this looks good to me.

Grest not to have the example be incorrect.

RussellSpitzer (approved):
I agree with @kbendick's comment on the grammar of that sentence "The fully qualified class name of a". I'd also be ok if we wanted to just start that line with "Must be set when catalog-type is unset" then go into the definition, but I think the current order is fine as well.

Once that's cleared up I'm good to merge

docs/flink-connector.md
nastra: good point, updated the sentence on L99 as you suggested

docs/flink-connector.md
nastra: makes perfect sense, I should have actually read the sentence when copy/pasting it from flink-connector.md :) I updated the sentence there as well

pvary:
Merged, as it seems to me that everyone is in agreement that we should merge this, but i has been forgotten.
Do we need this here as well, and maybe in other branches too? https://github.com/apache/iceberg-docs/blob/latest/docs/content/flink-connector.md

nastra:
Usually https://github.com/apache/iceberg-docs/blob/latest/docs/content/flink-connector.md would be updated when copying over docs during a release, but here I think it would make sense to update it manually, since the docs for the release were updated before this was merged.

PR #5046, opened 2022-06-15
AWS: support configure region

dungdm93:
Currently, to configure AWS region, we need to configure at global scope such as in environment variable `AWS_REGION=us-east-1`.

In the case Spark (or Flink) have multiple catalogs, each use different bucket in different region, there is no way specify region for each bucket now.

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
@@ -30,10 +30,12 @@ import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.awscore.client.builder.AwsClientBuilder;
 import software.amazon.awssdk.core.client.builder.SdkClientBuilder;
 import software.amazon.awssdk.http.SdkHttpClient;
 import software.amazon.awssdk.http.apache.ApacheHttpClient;
 import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import software.amazon.awssdk.services.glue.GlueClient;
 import software.amazon.awssdk.services.kms.KmsClient;
@@ -84,12 +86,16 @@ public class AwsClientFactories {
 
   static class DefaultAwsClientFactory implements AwsClientFactory {
 
+    private String s3Region;
     private String s3Endpoint;
     private String s3AccessKeyId;
     private String s3SecretAccessKey;
     private String s3SessionToken;
     private Boolean s3PathStyleAccess;
     private Boolean s3UseArnRegionEnabled;
+    private String kmsRegion;
+    private String glueRegion;
+    private String dynamoDbRegion;
     private String dynamoDbEndpoint;
     private String httpClientType;
 
@@ -101,6 +107,7 @@ public class AwsClientFactories {
       return S3Client.builder()
           .httpClientBuilder(configureHttpClientBuilder(httpClientType))
           .applyMutation(builder -> configureEndpoint(builder, s3Endpoint))
+          .applyMutation(builder -> configureRegion(builder, s3Region))
           .serviceConfiguration(s3Configuration(s3PathStyleAccess, s3UseArnRegionEnabled))
           .credentialsProvider(credentialsProvider(s3AccessKeyId, s3SecretAccessKey, s3SessionToken))
           .build();
@@ -108,12 +115,18 @@ public class AwsClientFactories {
 
     @Override
     public GlueClient glue() {
-      return GlueClient.builder().httpClientBuilder(configureHttpClientBuilder(httpClientType)).build();
+      return GlueClient.builder()
+          .httpClientBuilder(configureHttpClientBuilder(httpClientType))
+          .applyMutation(builder -> configureRegion(builder, glueRegion))
+          .build();
     }
 
     @Override
     public KmsClient kms() {
-      return KmsClient.builder().httpClientBuilder(configureHttpClientBuilder(httpClientType)).build();
+      return KmsClient.builder()
+          .httpClientBuilder(configureHttpClientBuilder(httpClientType))
+          .applyMutation(builder -> configureRegion(builder, kmsRegion))
+          .build();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


rdblue:
@jackye1995, can you take a look at this?

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -99,29 +102,40 @@ static class DefaultAwsClientFactory implements AwsClientFactory {
     public S3Client s3() {
       return S3Client.builder()
           .httpClientBuilder(configureHttpClientBuilder(httpClientType))
-          .applyMutation(builder -> configureEndpoint(builder, s3Endpoint))
+          .applyMutation(builder -> configureRegion(builder, s3Endpoint))
+          .applyMutation(builder -> configureEndpoint(builder, region))
```
rdblue, line 106: This looks suspicious. It calls `configureRegion` with `s3Endpoint` and `configureEndpoint` with `region`.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -280,6 +280,12 @@ public class AwsProperties implements Serializable {
    */
   public static final String CLIENT_ASSUME_ROLE_REGION = "client.assume-role.region";
 
+  /**
+   * The value must be one of {@link software.amazon.awssdk.regions.Region}, such as 'us-east-1'.
+   * For more details, see https://docs.aws.amazon.com/general/latest/gr/rande.html
+   */
+  public static final String CLIENT_REGION = "client.region";
```
rdblue, line 287: Why `client.region` and not just `region`?

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -99,29 +102,40 @@ static class DefaultAwsClientFactory implements AwsClientFactory {
...
           .credentialsProvider(credentialsProvider(s3AccessKeyId, s3SecretAccessKey, s3SessionToken))
           .build();
     }
 
     @Override
     public GlueClient glue() {
-      return GlueClient.builder().httpClientBuilder(configureHttpClientBuilder(httpClientType)).build();
+      return GlueClient.builder()
+          .httpClientBuilder(configureHttpClientBuilder(httpClientType))
+          .applyMutation(builder -> configureEndpoint(builder, region))
```
amogh-jahagirdar, line 116: I don't think this is right, we're calling configureEndpoint with the region, instead of the new configureRegion?

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -99,29 +102,40 @@ static class DefaultAwsClientFactory implements AwsClientFactory {
...
+          .build();
     }
 
     @Override
     public DynamoDbClient dynamo() {
-      return DynamoDbClient.builder().httpClientBuilder(configureHttpClientBuilder(httpClientType)).build();
+      return DynamoDbClient.builder()
+          .httpClientBuilder(configureHttpClientBuilder(httpClientType))
+          .applyMutation(builder -> configureEndpoint(builder, region))
```
amogh-jahagirdar, line 132: same

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
dungdm93: Sorry, my bad.

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
dungdm93: Yeah. It's my bad. fixed.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
dungdm93: @rdblue , @amogh-jahagirdar I'm ok with `client`.
But I wonder is there any cases `s3`/`glue`/`dymano` in a catalog are located in different region.
So, in that case, we need different key for each service like `s3.region`, `glue.region` and `dymano.region`.

dungdm93: I decided to create a key for each service: `s3.region`, `glue.region`, `dymano.region` and `kms.region`.

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
```diff
@@ -539,6 +563,7 @@ public void setGlueCatalogId(String id) {
   public boolean glueCatalogSkipArchive() {
     return glueCatalogSkipArchive;
   }
+
```
amogh-jahagirdar, line 566: Nit: Unnecessary new line

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
```diff
@@ -140,17 +155,22 @@ public void initialize(Map<String, String> properties) {
...
           (s3AccessKeyId == null) == (s3SecretAccessKey == null),
-          "S3 client access key ID and secret access key must be set at the same time");
+          "S3 client access key ID and secret access key must be set at the same time"
+      );
+
+      this.kmsRegion = properties.get(AwsProperties.KMS_REGION);
+      this.glueRegion = properties.get(AwsProperties.GLUE_REGION);
+      this.dynamoDbRegion = properties.get(AwsProperties.DYNAMODB_REGION);
       this.dynamoDbEndpoint = properties.get(AwsProperties.DYNAMODB_ENDPOINT);
+
```
amogh-jahagirdar, line 167: Nit: unnecessary new line

amogh-jahagirdar (commented):
Overall LGTM, while I do think having separate configurable regions is nice and would cover wider possibilities of infra setups, it does seem a bit overkill because I think most people would be these services within same region. For cross region s3 access, access points could be leveraged. Overall though the flexibility is nice, so I think it should be good. Would like to hear the communities thoughts too. Also @jackye1995 any thoughts here?

aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java
dungdm93: It's done by IntelliJ formatter. I think the previous commit forgot this one.
 
image

aws/src/main/java/org/apache/iceberg/aws/AwsClientFactories.java
dungdm93: Removed.

dungdm93:
@[amogh-jahagirdar Currently, I'm see the faction of configurations between services, e.g: `glue`, `kms`, `dynamo` missing the way to set `credentialsProvider`

WDYT about having common configuration of `endpoint`, `region`, `access-key-id`, `secret-access-key` and `session-token` for all services.
```java
String S3FILEIO_PREFIX = "s3."
String GLUE_PREFIX = "glue."

/**
* To set S3 region, use s3.region, to set Glue region set glue.region, etc...
**/
String REGION_CONFIG = "region"
/**
* To set S3 endpoint, use s3.endpoint, to set Glue region set glue.endpoint, etc...
**/
String ENDPOINT_CONFIG = "endpoint"
...
```

nastra:
fixes https://github.com/apache/iceberg/issues/5769

jackye1995:
Hi sorry I overlooked this PR for a long time, do you still plan to finish this feature?

ChristinaTech:
This has the same goal as a PR I was going to open @dungdm93. Do you plan to continue or shall I pick up from where you left things?

PR #3613, opened 2021-11-26
Spark: Fix NOT IN predicates in SparkFilters

aokolnychyi:
This PR is an attempt to fix NOT IN predicates in `SparkFilters`.

PR #5049, opened 2022-06-15
Spark: Fixes NPE when convert NOT IN filter

ConeyLiu:
This pr fixes the following problems:
``` java
Cannot create expression literal from null
java.lang.NullPointerException: Cannot create expression literal from null
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkNotNull(Preconditions.java:907)
	at org.apache.iceberg.expressions.Literals.from(Literals.java:62)
	at org.apache.iceberg.relocated.com.google.common.collect.Iterators$6.transform(Iterators.java:826)
	at org.apache.iceberg.relocated.com.google.common.collect.TransformedIterator.next(TransformedIterator.java:52)
	at org.apache.iceberg.relocated.com.google.common.collect.Iterators.addAll(Iterators.java:367)
	at org.apache.iceberg.relocated.com.google.common.collect.Lists.newArrayList(Lists.java:147)
	at org.apache.iceberg.relocated.com.google.common.collect.Lists.newArrayList(Lists.java:133)
	at org.apache.iceberg.expressions.UnboundPredicate.<init>(UnboundPredicate.java:55)
	at org.apache.iceberg.expressions.Expressions.predicate(Expressions.java:265)
	at org.apache.iceberg.expressions.Expressions.predicate(Expressions.java:254)
	at org.apache.iceberg.expressions.Expressions.notIn(Expressions.java:234)
	at org.apache.iceberg.spark.SparkFilters.convert(SparkFilters.java:179)
```

This happens when we convert `some_col NOT IN values` filter from spark to Iceberg and the `values` contains `null`.  For example: `DELETE FROM table WHERE id NOT IN (1, null)`.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkFilters.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkFilters.java
@@ -178,6 +178,7 @@ public class SparkFilters {
             In childInFilter = (In) childFilter;
             Expression notIn = notIn(unquote(childInFilter.attribute()),
                 Stream.of(childInFilter.values())
+                    .filter(Objects::nonNull)
                     .map(SparkFilters::convertLiteral)
                     .collect(Collectors.toList()));
             return and(notNull(childInFilter.attribute()), notIn);
--- a/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/TestSparkFilters.java
+++ b/spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/TestSparkFilters.java
@@ -148,4 +148,12 @@ public class TestSparkFilters {
     Expression expected = Expressions.and(Expressions.notNull("col"), Expressions.notIn("col", 1, 2));
     Assert.assertEquals("Expressions should match", expected.toString(), actual.toString());
   }
+
+  @Test
+  public void testNotInNull() {
+    Not filter = Not.apply(In.apply("col", new Integer[]{1, 2, null}));
+    Expression actual = SparkFilters.convert(filter);
+    Expression expected = Expressions.and(Expressions.notNull("col"), Expressions.notIn("col", 1, 2));
+    Assert.assertEquals("Expressions should match", expected.toString(), actual.toString());
+  }
 }


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkFilters.java
```diff
@@ -178,6 +178,7 @@ public static Expression convert(Filter filter) {
             In childInFilter = (In) childFilter;
             Expression notIn = notIn(unquote(childInFilter.attribute()),
                 Stream.of(childInFilter.values())
+                    .filter(Objects::nonNull)
```
ConeyLiu, line 181: This refers to the process in L166. I think this is because we don't support NULL literal in Iceberg. However,  I think here have changed the original filter meanings. @rdblue, should we support NULL literal in Iceberg?

ConeyLiu:
Hi, @rdblue @aokolnychyi @chenjunjiedada @jackye1995, could you help to review this? Thanks in advance.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/SparkFilters.java
rdblue: > should we support NULL literal in Iceberg?

No. `null` is purposely not allowed as a literal because that brings in differences between SQL 3-valued boolean semantics and normal semantics. Engines are responsible for interpreting and rewriting filters. Similarly, in cases like this we actually want to reject null values in IN predicates. The user or the engine needs to rewrite before passing to Iceberg.

rdblue:
I don't think this should be merged. This NPE is intended. This code can catch the NPE and throw a more friendly error, but I think it is also reasonable to keep the exception message as it is right now. The problem is that `null` should not be passed as a literal.

kbendick:
I recall now when working on the `NOT_STARTS_WITH` that we wanted to leave it up to the engine to handle 3 value boolean logic.

The example query, `DELETE FROM table WHERE id NOT IN (1, null)`, should probably be written as `DELETE FROM table WHERE id is not null and id != 1`. Spark will add the not null-check itself often times.

With 3-valued boolean logic, for comparison purposes every null is a different null and nothing equals null. For those who might want a refresher (like I often need), I like this website's explanation of 3-valued boolean logic involving nulls: https://modern-sql.com/concept/three-valued-logic

To avoid the headache that can incur, we explicitly avoid 3-valued boolean logic. There would be better filter pushdown and more advantage taken of Iceberg's statistics by writing the query with explicit `AND is [not] null` checking too.

ConeyLiu:
Thanks, @rdblue @kbendick for the detailed explanation. And this doc: https://modern-sql.com/concept/three-valued-logic is really helpful. I submit this PR because the user could do this with Spark 3.1 while failing in Spark 3.2. The current logic should be enough. Thanks again.

Another question. Should we port the https://github.com/apache/iceberg/pull/3613 to other spark versions? Currently, the other Spark versions have different behavior again Spark 3.2.

rdblue:
@ConeyLiu, sure. If you want to backport that, I think it would be useful.

rdblue:
I'm going to close this since I think the behavior is not controversial.

PR #3998, opened 2022-01-28
Flink: Support write options in FlinkSink builder

hililiwei:
Supports overwriting table configuration, such upsert \ overwirte\ file-format, when creating Flink sink.

PR #5050, opened 2022-06-15
Flink: Support write options in the in-line insert SQL comments

hililiwei:
Support write options in the in-line insert SQL comments:

```sql
insert into talbeName  /*+ OPTIONS('upsert-enabled'='true')*/ 
select 
…………

```

For properties like upsert, it is not recommended to enable it in the table props. Configuring it in table props is a bit strange.

### DIFF
--- a/docs/flink-getting-started.md
+++ b/docs/flink-getting-started.md
@@ -562,6 +562,11 @@ FlinkSink.Builder builder = FlinkSink.forRow(dataStream, SimpleDataUtil.FLINK_SC
     .set("write-format", "orc")
     .set(FlinkWriteOptions.OVERWRITE_MODE, "true");
 ```
+For Flink SQL, write options can be passed in via SQL hints like this:
+```
+INSERT INTO tableName /*+ OPTIONS('upsert-enabled'='true') */
+...
+```
 
 | Flink option           | Default                    | Description                                                                                                |
 |------------------------| -------------------------- |------------------------------------------------------------------------------------------------------------|
--- a/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
+++ b/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
@@ -106,7 +106,7 @@ public class FlinkDynamicTableFactory
   public DynamicTableSink createDynamicTableSink(Context context) {
     ObjectPath objectPath = context.getObjectIdentifier().toObjectPath();
     CatalogTable catalogTable = context.getCatalogTable();
-    Map<String, String> tableProps = catalogTable.getOptions();
+    Map<String, String> writeProps = catalogTable.getOptions();
     TableSchema tableSchema = TableSchemaUtils.getPhysicalSchema(catalogTable.getSchema());
 
     TableLoader tableLoader;
@@ -115,10 +115,10 @@ public class FlinkDynamicTableFactory
     } else {
       tableLoader =
           createTableLoader(
-              catalogTable, tableProps, objectPath.getDatabaseName(), objectPath.getObjectName());
+              catalogTable, writeProps, objectPath.getDatabaseName(), objectPath.getObjectName());
     }
 
-    return new IcebergTableSink(tableLoader, tableSchema, context.getConfiguration());
+    return new IcebergTableSink(tableLoader, tableSchema, context.getConfiguration(), writeProps);
   }
 
   @Override
--- a/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/IcebergTableSink.java
+++ b/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/IcebergTableSink.java
@@ -37,6 +37,7 @@ public class IcebergTableSink implements DynamicTableSink, SupportsPartitioning,
   private final TableLoader tableLoader;
   private final TableSchema tableSchema;
   private final ReadableConfig readableConfig;
+  private final Map<String, String> writeProps;
 
   private boolean overwrite = false;
 
@@ -45,13 +46,18 @@ public class IcebergTableSink implements DynamicTableSink, SupportsPartitioning,
     this.tableSchema = toCopy.tableSchema;
     this.overwrite = toCopy.overwrite;
     this.readableConfig = toCopy.readableConfig;
+    this.writeProps = toCopy.writeProps;
   }
 
   public IcebergTableSink(
-      TableLoader tableLoader, TableSchema tableSchema, ReadableConfig readableConfig) {
+      TableLoader tableLoader,
+      TableSchema tableSchema,
+      ReadableConfig readableConfig,
+      Map<String, String> writeProps) {
     this.tableLoader = tableLoader;
     this.tableSchema = tableSchema;
     this.readableConfig = readableConfig;
+    this.writeProps = writeProps;
   }
 
   @Override
@@ -70,6 +76,7 @@ public class IcebergTableSink implements DynamicTableSink, SupportsPartitioning,
                 .tableSchema(tableSchema)
                 .equalityFieldColumns(equalityColumns)
                 .overwrite(overwrite)
+                .setAll(writeProps)
                 .flinkConf(readableConfig)
                 .append();
   }
--- a/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
+++ b/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
@@ -166,6 +166,53 @@ public class TestFlinkUpsert extends FlinkCatalogTestBase {
   }
 
   @Test
+  public void testUpsertOptions() {
+    String tableName = "test_upsert_options";
+    LocalDate dt20220301 = LocalDate.of(2022, 3, 1);
+    LocalDate dt20220302 = LocalDate.of(2022, 3, 2);
+
+    Map<String, String> optionsUpsertProps = Maps.newHashMap(tableUpsertProps);
+    optionsUpsertProps.remove(TableProperties.UPSERT_ENABLED);
+    sql(
+        "CREATE TABLE %s(id INT NOT NULL, province STRING NOT NULL, dt DATE, PRIMARY KEY(id,province) NOT ENFORCED) "
+            + "PARTITIONED BY (province) WITH %s",
+        tableName, toWithClause(optionsUpsertProps));
+
+    try {
+      sql(
+          "INSERT INTO %s /*+ OPTIONS('upsert-enabled'='true')*/  VALUES "
+              + "(1, 'a', DATE '2022-03-01'),"
+              + "(2, 'b', DATE '2022-03-01'),"
+              + "(1, 'b', DATE '2022-03-01')",
+          tableName);
+
+      sql(
+          "INSERT INTO %s /*+ OPTIONS('upsert-enabled'='true')*/  VALUES "
+              + "(4, 'a', DATE '2022-03-02'),"
+              + "(5, 'b', DATE '2022-03-02'),"
+              + "(1, 'b', DATE '2022-03-02')",
+          tableName);
+
+      List<Row> rowsOn20220301 =
+          Lists.newArrayList(Row.of(2, "b", dt20220301), Row.of(1, "a", dt20220301));
+      TestHelpers.assertRows(
+          sql("SELECT * FROM %s WHERE dt < '2022-03-02'", tableName), rowsOn20220301);
+
+      List<Row> rowsOn20220302 =
+          Lists.newArrayList(
+              Row.of(1, "b", dt20220302), Row.of(4, "a", dt20220302), Row.of(5, "b", dt20220302));
+      TestHelpers.assertRows(
+          sql("SELECT * FROM %s WHERE dt = '2022-03-02'", tableName), rowsOn20220302);
+
+      TestHelpers.assertRows(
+          sql("SELECT * FROM %s", tableName),
+          Lists.newArrayList(Iterables.concat(rowsOn20220301, rowsOn20220302)));
+    } finally {
+      sql("DROP TABLE IF EXISTS %s.%s", flinkDatabase, tableName);
+    }
+  }
+
+  @Test
   public void testPrimaryKeyEqualToPartitionKey() {
     // This is an SQL based reproduction of TestFlinkIcebergSinkV2#testUpsertOnDataKey
     String tableName = "upsert_on_data_key";


kbendick:
> For properties like upsert, it is not recommended to enable it in the table props. Configuring it in table props is a bit strange.

kbendick (commented):
This is great to see worked on @hililiwei!

Can you provide a test of some sort to ensure that this works?

hililiwei:
> This is great to see worked on @hililiwei!
> 
> Can you provide a test of some sort to ensure that this works?

thx. Added UT

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkSink.java
```diff
@@ -231,6 +231,11 @@ public Builder writeParallelism(int newWriteParallelism) {
       return this;
     }
 
+    public Builder properties(Map<String, String> properties) {
+      upsert(PropertyUtil.propertyAsBoolean(properties, UPSERT_ENABLED, UPSERT_ENABLED_DEFAULT));
+      return this;
+    }
```
kbendick, line 237: As a follow up, we should consider supporting more options from in-line SQL comments.

Is this `properties` needed for any SQL in-line comment based configuration or just for our own custom configurations that Flink doesn't know about?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkSink.java
```diff
@@ -231,6 +231,11 @@ public Builder writeParallelism(int newWriteParallelism) {
       return this;
     }
 
+    public Builder properties(Map<String, String> properties) {
```
kbendick, line 234: Nit: Can we add a javadoc here similar to the other ones? This might help me understand the flow better for non-SQL users. Users shouldn't pass in the run-time properties map, correct?

We might want to be somewhat more specific about which `properties` should come here.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkSink.java
```diff
@@ -231,6 +231,11 @@ public Builder writeParallelism(int newWriteParallelism) {
...
+      upsert(PropertyUtil.propertyAsBoolean(properties, UPSERT_ENABLED, UPSERT_ENABLED_DEFAULT));
```
kbendick, line 235: Nit: Please add a `Preconditions` check that the map isn't null (unless null needs to be allowed).

chenjunjiedada:
+1 to honor job level option. Shall we use the table property or Flink-specific option definition as in #4670? cc @stevenzwu  as well.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkSink.java
hililiwei: Yes, in addition to `upsert`, we have more configuration that can be used in inline SQL , such as write file size and so on.
For more configuration that can be supported, see FlinkWriteConf.java of  #3998.
I'd prefer to wait for #3998 to finish.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkSink.java
hililiwei: Adding a javadoc is a good note.
As described above, we have another PR #3998 processing non-SQL case, that is, Java api.

hililiwei:
I rewrote the PR based on #3998, which can make this clearer. Thanks for @kbendick 's advice.

hililiwei:
cc @rdblue @openinx could you please take a look when you are available?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
```diff
@@ -160,6 +160,49 @@ public void testUpsertAndQuery() {
...
+    LocalDate dt20220301 = LocalDate.of(2022, 3, 1);
+    LocalDate dt20220302 = LocalDate.of(2022, 3, 2);
+
+    tableUpsertProps.remove(TableProperties.UPSERT_ENABLED);
+    sql("CREATE TABLE %s(id INT NOT NULL, province STRING NOT NULL, dt DATE, PRIMARY KEY(id,province) NOT ENFORCED) " +
+            "PARTITIONED BY (province) WITH %s",
+        tableName, toWithClause(tableUpsertProps));
+
+    try {
+      sql("INSERT INTO %s /*+ OPTIONS('upsert-enabled'='true')*/  VALUES " +
```
chenjunjiedada, line 175: Is this syntax also supported in previous flink versions?

hililiwei: yes.

hililiwei:
cc @rdblue @openinx could you please take a look when you are available?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/IcebergTableSink.java
```diff
@@ -37,6 +37,7 @@ public class IcebergTableSink implements DynamicTableSink, SupportsPartitioning,
   private final TableLoader tableLoader;
   private final TableSchema tableSchema;
   private final ReadableConfig readableConfig;
+  private final Map<String, String> tableProps;
```
stevenzwu, line 40: let's not call it `tableProps` to avoid confusing with Iceberg table properties. Is `writeProps` more accurate? it would be consistent with the Javadoc of `setAll` from `FlinkSink#Builder`.

```
    /**
     * Set the write properties for Flink sink. View the supported properties in {@link
     * FlinkWriteOptions}
     */
    public Builder setAll(Map<String, String> properties) {
      writeOptions.putAll(properties);
      return this;
    }
```

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
```diff
@@ -165,6 +165,52 @@ public void testUpsertAndQuery() {
     }
   }
 
+  @Test
+  public void testUpsertOptions() {
+    String tableName = "test_upsert_options";
+    LocalDate dt20220301 = LocalDate.of(2022, 3, 1);
+    LocalDate dt20220302 = LocalDate.of(2022, 3, 2);
+
+    tableUpsertProps.remove(TableProperties.UPSERT_ENABLED);
```
stevenzwu, line 174: we shouldn't modify the `tableUpsertProps`, as the side effect can affect other tests. It is better to construct a different map in this test method. I am a little surprised and confused why the CI tests passed.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
```diff
@@ -165,6 +165,52 @@ public void testUpsertAndQuery() {
...
+          "INSERT INTO %s /*+ OPTIONS('upsert-enabled'='true')*/  VALUES "
+              + "(4, 'a', DATE '2022-03-02'),"
+              + "(5, 'b', DATE '2022-03-02'),"
+              + "(1, 'b', DATE '2022-03-02')",
+          tableName);
+
+      List<Row> rowsOn20220301 =
+          Lists.newArrayList(Row.of(2, "b", dt20220301), Row.of(1, "a", dt20220301));
+      TestHelpers.assertRows(
+          sql("SELECT * FROM %s WHERE dt < '2022-03-02'", tableName), rowsOn20220301);
```
stevenzwu, line 198: nit: is it intentionally to be `dt <` not `dt =`?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
```diff
@@ -165,6 +165,52 @@ public void testUpsertAndQuery() {
...
+      sql(
+          "INSERT INTO %s /*+ OPTIONS('upsert-enabled'='true')*/  VALUES "
+              + "(4, 'a', DATE '2022-03-02'),"
+              + "(5, 'b', DATE '2022-03-02'),"
+              + "(1, 'b', DATE '2022-03-02')",
+          tableName);
+
+      List<Row> rowsOn20220301 =
+          Lists.newArrayList(Row.of(2, "b", dt20220301), Row.of(1, "a", dt20220301));
+      TestHelpers.assertRows(
```
stevenzwu, line 197: not very familiar with upsert. for my own learning, why are the two queries resulted in different output? the primary key is the combination/concatenation of (id, province).

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
```diff
@@ -118,7 +118,7 @@ public DynamicTableSink createDynamicTableSink(Context context) {
               catalogTable, tableProps, objectPath.getDatabaseName(), objectPath.getObjectName());
     }
 
-    return new IcebergTableSink(tableLoader, tableSchema, context.getConfiguration());
+    return new IcebergTableSink(tableLoader, tableSchema, context.getConfiguration(), tableProps);
```
stevenzwu, line 121: similarly, can we rename `tableProps` here to `writeProps`?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
hililiwei: In UPSERT mode, data with the same primary key is overwritten by new data. In this case, after two insertion operations, the data in the table is:
```
(1, 'a', DATE '2022-03-01')
(2, 'b', DATE '2022-03-01')
(4, 'a', DATE '2022-03-02')
(5, 'b', DATE '2022-03-02')
(1, 'b', DATE '2022-03-02')
```

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/IcebergTableSink.java
hililiwei: done

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkDynamicTableFactory.java
hililiwei: done

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
hililiwei: That makes sense. I changed it. :joy:

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
stevenzwu: I realized where my confusion came from. I assumed the table is partitioned by date. Hence couldn't get my head around why the upsert (composite primary key) didn't work for day 1.

docs/flink-getting-started.md
```diff
@@ -553,14 +553,18 @@ env.execute("Test Iceberg DataStream");
...
 
 ```
 FlinkSink.Builder builder = FlinkSink.forRow(dataStream, SimpleDataUtil.FLINK_SCHEMA)
     .table(table)
     .tableLoader(tableLoader)
     .set("write-format", "orc")
     .set(FlinkWriteOptions.OVERWRITE_MODE, "true");
+    
+INSERT INTO talbeName  /*+ OPTIONS('upsert-enabled'='true')*/ 
+SELECT
```
stevenzwu, line 566: nit: either complete `select` part or just leave it as `...`

stevenzwu (approved):
LGTM except for a nit comment in doc

docs/flink-getting-started.md
stevenzwu: also it is probably more clear to start the SQL hints as a separate paragraph after the Java API code snippet. e.g.

For Flink SQL, write options can be passed as hints like this
```
INSERT INTO tableName  /*+ OPTIONS('upsert-enabled'='true')*/  
...
```

docs/flink-getting-started.md
```diff
@@ -553,14 +553,18 @@ env.execute("Test Iceberg DataStream");
...
+Flink write options are passed when configuring the FlinkSink or Flink SQL hints, like this:
 
 ```
 FlinkSink.Builder builder = FlinkSink.forRow(dataStream, SimpleDataUtil.FLINK_SCHEMA)
     .table(table)
     .tableLoader(tableLoader)
     .set("write-format", "orc")
     .set(FlinkWriteOptions.OVERWRITE_MODE, "true");
+    
+INSERT INTO talbeName  /*+ OPTIONS('upsert-enabled'='true')*/
```
rdblue, line 565: Typo: `tableName`

docs/flink-getting-started.md
kbendick: Nit: Is it possible to have a space before the terminal `*/`? If not, no big deal but it would be nice if we could.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
kbendick: We should consider refactoring these in a follow up PR so the table is partitioned by date.

I've encountered this confusion as well before.

rdblue (approved):

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
kbendick: Non-blocking though =)

rdblue:
Looks good other than the issues that @stevenzwu pointed out for docs. Thanks, @hililiwei!

kbendick (approved):
A few nits but overall this looks good to me.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
hililiwei: 😂  It's easy to misunderstand.

stevenzwu (approved):

hililiwei:
Doc updated.

rdblue:
Thanks, @hililiwei!

hililiwei:
Thanks @rdblue @stevenzwu @chenjunjiedada @kbendick 
I will port it to 13&14. 

I also created an [issuse ](https://github.com/apache/iceberg/issues/5418) for https://github.com/apache/iceberg/pull/5050#discussion_r933426976, which i think it's good for beginners to start with. If you can, maybe we could tag an 'beginner' for it? @rdblue

PR #5052, opened 2022-06-15
[CORE] Add SerDe tests for CreateTableRequest

kbendick:
This adds tests for the serialization / deserialization of `CreateTableRequest`.

This class is already being used quite a bit in other tests, specifically `CatalogTests` and `TestRESTCatalog`. It also relied pretty heavily on existing serializers inside of `RESTSerializers` for things like `Schema` etc, which simply call out to our existing parsers.

Thus, I kept the tests light in terms of combinations of schemas etc as this is a combination of several tested components.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/requests/CreateTableRequest.java
+++ b/core/src/main/java/org/apache/iceberg/rest/requests/CreateTableRequest.java
@@ -43,7 +43,7 @@ public class CreateTableRequest implements RESTRequest {
   private UnboundPartitionSpec spec;
   private UnboundSortOrder order;
   private Map<String, String> properties;
-  private Boolean stageCreate;
+  private Boolean stageCreate = false;
 
   public CreateTableRequest() {
     // Needed for Jackson Deserialization.
--- a/core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
+++ b/core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
@@ -82,15 +82,24 @@ public class TestCreateTableRequest extends RequestResponseTestBase<CreateTableR
 
     assertRoundTripSerializesEquallyFrom(fullJsonRaw, req);
 
-    // The same JSON but using existing parsers for clarity
-    String fullJson = String.format(
+    // The same JSON but using existing parsers for clarity and staging the request instead of committing
+    String jsonStagedReq = String.format(
         "{\"name\":\"%s\",\"location\":\"%s\",\"schema\":%s,\"spec\":%s," +
             "\"order\":%s,\"properties\":%s,\"stageCreate\":%b}",
             SAMPLE_NAME, SAMPLE_LOCATION, SchemaParser.toJson(SAMPLE_SCHEMA),
             PartitionSpecParser.toJson(SAMPLE_SPEC.toUnbound()), SortOrderParser.toJson(SAMPLE_WRITE_ORDER.toUnbound()),
-            mapper().writeValueAsString(SAMPLE_PROPERTIES), false);
+            mapper().writeValueAsString(SAMPLE_PROPERTIES), true);
 
-    assertRoundTripSerializesEquallyFrom(fullJson, req);
+    CreateTableRequest stagedReq  = CreateTableRequest.builder()
+        .withName(SAMPLE_NAME)
+        .withLocation(SAMPLE_LOCATION)
+        .withSchema(SAMPLE_SCHEMA)
+        .setProperties(SAMPLE_PROPERTIES)
+        .withPartitionSpec(SAMPLE_SPEC)
+        .withWriteOrder(SAMPLE_WRITE_ORDER)
+        .build();
+
+    assertRoundTripSerializesEquallyFrom(jsonStagedReq, req);
 
     // Partition spec and write order can be null or use PartitionSpec.unpartitioned() and SortOrder.unsorted()
     String jsonWithExplicitUnsortedUnordered = String.format(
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/iceberg/rest/requests/CreateTableRequest.java
```diff
@@ -43,7 +43,7 @@ public class CreateTableRequest implements RESTRequest {
   private UnboundPartitionSpec spec;
   private UnboundSortOrder order;
   private Map<String, String> properties;
-  private Boolean stageCreate;
+  private Boolean stageCreate = false;
```
kbendick, line 46: I defaulted this to `false` so that users didn't need to explicitly set it if writing JSON. Because it's a boxed `Boolean`, if it's not present in the JSON, it gets set as `null`.

I did this because `CreateTableRequest` does not have `stageCreate` in its list of required fields in the OpenAPI spec.

As there is an assertion that `stageCreate` is not `null` in the `validate` method of this class, I would also be comfortable with requiring that `stageCreate` be set in the OpenAPI spec, but I prefer this as it matches the builders behavior as well (if `.stageCreate()` isn't called on the builder, `false` is used).

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
```diff
@@ -0,0 +1,290 @@
...
+      .bucket("id", 16)
+      .build();
+  private static final SortOrder SAMPLE_WRITE_ORDER = SortOrder.builderFor(SAMPLE_SCHEMA)
+      .asc("data", NullOrder.NULLS_LAST)
+      .build();
+
+  @Test
+  // Test cases that are JSON that can be created via the Builder
+  public void testRoundTripSerDe() throws JsonProcessingException {
+    String fullJsonRaw =
```
kbendick, line 65: As most of the complex fields (all of the classes we define, such as `Schema` etc) are serialized and deserialized via the existing `*Parser`, I only put one test where the whole JSON is defined.

Formatting it well across lines was somewhat difficult, as the `assertRoundTripSerializesEquallyFrom` method _requires_ that the JSON matches letter for letter.

I can split this out with newlines and then replace them if we'd like though.

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
```diff
@@ -0,0 +1,290 @@
...
+        .withPartitionSpec(SAMPLE_SPEC)
+        .withWriteOrder(SAMPLE_WRITE_ORDER)
+        .build();
+
+    assertRoundTripSerializesEquallyFrom(fullJsonRaw, req);
+
+    // The same JSON but using existing parsers for clarity
+    String fullJson = String.format(
+        "{\"name\":\"%s\",\"location\":\"%s\",\"schema\":%s,\"spec\":%s," +
+            "\"order\":%s,\"properties\":%s,\"stageCreate\":%b}",
```
kbendick, line 88: For strings that continue onto the next line, my editor (with our configuration applied) wanted the next string line to be indented. I can remove that though.

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
```diff
@@ -0,0 +1,293 @@
...
+        .stageCreate()
+        .build();
+  }
+
+  @Override
+  public void assertEquals(CreateTableRequest actual, CreateTableRequest expected) {
+    Assert.assertEquals("Name should be the same", expected.name(), actual.name());
+    Assert.assertEquals("Location should be the same if provided", expected.location(), actual.location());
+    Assert.assertTrue("Schemas should be equivalent and have same schema id",
+        expected.schema().sameSchema(actual.schema()) && expected.schema().schemaId() == actual.schema().schemaId());
```
kbendick, line 280: I originally had this as two assertions (one for `sameSchema` and one for equality checking on `schemaId`).

If we prefer, I can revert to that.

kbendick:
cc @danielcweeks @rdblue @Fokko @nastra and @singhpk234 who has helped contribute to some areas of the REST spec.

Kept this one relatively simple given that most of the heavy lifting is done by existing Parsers.

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
```diff
@@ -0,0 +1,294 @@
...
+
+    // The same JSON but using existing parsers for clarity and staging the request instead of committing
+    String jsonStagedReq = String.format(
+        "{\"name\":\"%s\",\"location\":\"%s\",\"schema\":%s,\"spec\":%s," +
+            "\"order\":%s,\"properties\":%s,\"stageCreate\":%b}",
+            SAMPLE_NAME, SAMPLE_LOCATION, SchemaParser.toJson(SAMPLE_SCHEMA),
+            PartitionSpecParser.toJson(SAMPLE_SPEC.toUnbound()), SortOrderParser.toJson(SAMPLE_WRITE_ORDER.toUnbound()),
+            mapper().writeValueAsString(SAMPLE_PROPERTIES), true);
+
+    CreateTableRequest stagedReq  = CreateTableRequest.builder()
```
singhpk234, line 93: nit : extra space
```suggestion
    CreateTableRequest stagedReq = CreateTableRequest.builder()
```

singhpk234 (approved):
LGTM, Thanks @kbendick !!

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #5055, opened 2022-06-15
Python: Don't use a metaclass for the Singleton

Fokko:
Follow up PR of https://github.com/apache/iceberg/pull/5008

Metaclasses in Python don't mix very well, and when they cause conflicts then it breaks at initialization.
An example of metaclasses is the ABC, but also the Pydantic BaseModel. I tried to circumvent this by extending the Singleton from ABC but it is likely that we use different metaclasses in the future, and then it will bring us sadness and misery.


samredai (approved):
LGTM 🚀

danielcweeks (approved):

Issue #4713, opened 2022-05-06
[Spark 3.3] : Support Apache Spark 3.3 in iceberg


singhpk234:
Hello All Members,

As we might be aware, Spark [v3.3.0-rc1](https://github.com/apache/spark/releases/tag/v3.3.0-rc1) is cut and being actively voted upon.

If someone else is not already working on this, I am happy to pick this up , also happy to collaborate if it's already being worked upon. 

--- 

cc @aokolnychyi @rdblue

hililiwei:
+1

abmo-x:
@singhpk234 
checking to see if you already started working on this? Let me know, happy to work together.

singhpk234:
@abmo-x 
I have started working on it, resolving conflicts due to [SPARK-38085](https://issues.apache.org/jira/browse/SPARK-38085), will update an initial patch soon.

singhpk234:
are you suggesting doing this : 
```
mv spark/v3.2 spark/v3.3
cp -r spark/v3.3 spark/v3.2
```

@hililiwei

singhpk234:
Thanks @hililiwei for highlighting this, this is a good proposal. will make sure I follow this.

singhpk234:
Resolved via https://github.com/apache/iceberg/pull/5094

PR #5056, opened 2022-06-15
Spark: Support Spark 3.3

singhpk234:
Closes https://github.com/apache/iceberg/issues/4713

- based on https://github.com/singhpk234/iceberg/pull/1

---- 
### About Change

-  Introduced a new Node `ReplaceIcebergData` to be used in place of `ReplaceData`, as it extends command and commands are eagerly executed, which was causing problems with `Update` (in case we allow allowScanDuplication) & `RowLevelCommandDynamicPruning` (filters not being injected)
- Changed the Rule name `RewriteDeleteFromTable` -> `RewriteDeleteFromIcebergTable`, `OptimizeMetadataOnlyDeleteFromTable` -> `OptimizeMetadataOnlyDeleteFromIcebergTable` to avoid conflicting rule in upstream which doesn't support DeltaWrites and uses ReplaceData

-----
### Other Notes : 

Have marked some UT's ignored for complete suite to run: 
[1] `TestAddFilesProcedure` failed due to bug in spark upstream, have fixed it in upstream but since the snapshot is last based of 7th May it still failing hence ignored it.
JIRA : https://issues.apache.org/jira/browse/SPARK-39417
[2] Investigating the RC for the below failures : 
    TestMerge#testMergeWithInvalidAssignments
    TestDeleteFrom#testDeleteFromUnpartitionedTable, TestDeleteFrom#testDeleteFromPartitionedTable

----

cc @rdblue @aokolnychyi @jackye1995 @RussellSpitzer @kbendick @amogh-jahagirdar @rajarshisarkar @flyrain


pan3793:
"copy 3.2 files from 3.3" should be in another PR, otherwise after squash, the git will not keep commit history.

singhpk234:
makes sense, thanks @pan3793, will do it in a separate PR.

spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestMerge.java
```diff
@@ -0,0 +1,1816 @@
...
+            });
+
+        AssertHelpers.assertThrows("Should complain about writing invalid data types",
+            AnalysisException.class, "Cannot safely cast",
+            () -> {
+              sql("MERGE INTO %s t USING source s " +
+                  "ON t.id == s.c1 " +
+                  "WHEN MATCHED THEN " +
+                  "  UPDATE SET t.s.n1 = s.c3", tableName);
+            });
```
singhpk234, line 1592: These test are failing as the Analysis Exception is happening due to 
https://github.com/apache/iceberg/blob/13db47e2755d8a885043a7ec62eee2dcab43a9bc/spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/AssignmentAlignmentSupport.scala#L174-L176

On debugging found that `s.c2` despite being `not null` by table definition has `nullable=true` when assignments are resolved. 

This was happening as of spark 3.3 , it introduced in https://github.com/apache/spark/commit/6689b97ec76abe5bab27f02869f8f16b32530d1a, now spark by default will not respect nullability in schema provided via 

```
`DataFrameReader.schema(schema).json(jsonDataset)`, `DataFrameReader.schema(schema).csv(csvDataset)` 
```

added this conf true in https://github.com/apache/iceberg/blob/master/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkExtensionsTestBase.java

as we use this in our test utils here we are using it to create our temp view

singhpk234:
UPDATE : 
spark 3.3 yesterday got published on maven repo : https://mvnrepository.com/artifact/org.apache.spark/spark-core
have removed the pr dependency on snapshot's repo
[1] Have reverted the ignored AddFilesProcedure UT's as well, since the upstream now having fix is published.
[2] Have reverted the Test#Merge ut failure as well, found spark hid the behaviour we were relying on behind a conf, have enabled in our session.


I am investigating the RC for the two remaining UT's failure in `TestDeleteFrom`, will update the fixes the shortly.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestDeleteFrom.java
```diff
@@ -0,0 +1,168 @@
...
+    long snapshotId = validationCatalog.loadTable(tableIdent).currentSnapshot().snapshotId();
+    String prefix = "snapshot_id_";
+    AssertHelpers.assertThrows("Should not be able to delete from a table at a specific snapshot",
+        IllegalArgumentException.class, "Cannot delete from table at a specific snapshot",
+        () -> sql("DELETE FROM %s.%s WHERE id < 4", tableName, prefix + snapshotId));
+  }
+
+  // revisit
+  @Ignore
+  @Test
```
singhpk234, line 101: These test's were failing stating  : 
```
Field _file not found in source schema
java.lang.IllegalArgumentException: Field _file not found in source schema
	at org.apache.iceberg.types.ReassignIds.field(ReassignIds.java:104)
	at org.apache.iceberg.types.ReassignIds.field(ReassignIds.java:28)
       ....
       	at org.apache.iceberg.spark.SparkSchemaUtil.convert(SparkSchemaUtil.java:192)
	at org.apache.iceberg.spark.source.SparkWriteBuilder.validateOrMergeWriteSchema(SparkWriteBuilder.java:247)
```

`_file` not found this was happening because in the upstream we were relying on `_ _metadata_col` flag in metadata attribute to filter these out : 
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/v2Commands.scala#L221-L225

But when we were doing spark schema conversion we were not adding this in attribute metadata, hence when it was trying validating with source schema was failing. 

Added a fix for same, (added for [3.2](https://github.com/apache/iceberg/pull/5075) as well) and adapted the UT to respect the upstream changes.

singhpk234:
> "copy 3.2 files from 3.3" should be in another PR, otherwise after squash, the git will not keep commit history.

AC, opened a new [PR](https://github.com/apache/iceberg/pull/5094) for "copy 3.2 files from 3.3"

singhpk234:
updated this pr also with 3.3 only changes just in case we decide to merge 3.3 only first.

singhpk234:
superceded by https://github.com/apache/iceberg/pull/5094

Thank you everyone for your awesome reviews :) !!

saswata-dutta (commented):
LGTM

PR #5057, opened 2022-06-15
Checkstyle: Add checkstyle rule for %d in Preconditions.checkArgument

### LINKED ISSUE #4618: Add checkstyle rule for %d in Preconditions.checkArgument
Reference: https://github.com/apache/iceberg/pull/4558, Preconditions.checkArgument only works with %s, as %d does not parse and returns a bad error message.

We can make a checkstyle rule to prevent this, example of a checkstyle rule:  https://github.com/apache/iceberg/blob/master/.baseline/checkstyle/checkstyle.xml#L54-L57

It should probably use multi-line regex rule as some of these are across multiple lines:  https://checkstyle.sourceforge.io/config_regexp.html#RegexpMultiline (thanks @RussellSpitzer for pointer)

xrl1:
Closes #4618 

From the issue:
`Preconditions.checkArgument` only works with %s, as %d does not parse and returns a bad error message.

Created a multiline regex to prevent this.

The regex is as follows:
```regex
Preconditions\.checkArgument\([^;]+%d[^;]+\);
```
The logic behind it: Find the function call, then any characters (except `;`) until `%d`, and finally search until the function closes and a ';'.
* Used `[^;]*` instead of `.*`, because this is a multiline search and the dotall regex param (enabled by `matchAcrossLines` config) is aggressive and this is the best way to break its search.

Also, fixed %d usages that the Checkstyle rule found

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -71,6 +71,12 @@
         <property name="format" value="new LinkedHashSet&lt;&gt;\(.*\)"/>
         <property name="message" value="Prefer using Sets.newLinkedHashSet() instead."/>
     </module>
+    <module name="RegexpMultiline">
+        <property name="fileExtensions" value="java"/>
+        <property name="matchAcrossLines" value="true"/>
+        <property name="format" value="Preconditions\.checkArgument\([^;]+%d[^;]+\);"/>
+        <property name="message" value="Preconditions.checkArgument does not support %d. use %s instead"/>
+    </module>
     <module name="SuppressionFilter"> <!-- baseline-gradle: README.md -->
         <property name="file" value="${config_loc}/checkstyle-suppressions.xml"/>
     </module>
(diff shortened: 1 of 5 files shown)


szehon-ho (approved):

szehon-ho:
Looks like some violation in Spark 3.0?  Also fyi , @kbendick

xrl1:
@szehon-ho I committed a fixup to the violation (missed it while running the checkstyle locally in the IDE for some reason)

xrl1:
Updated the PR again, and now also tested the build locally to pass, sorry for that.

szehon-ho:
@xrl1 thanks for the explanation for those points.  How about the last point:

> Added the quote and comma to avoid matching the first argument, but I can simplify the regex and not handle the edge cases where there is "%d" in the first argument

Do you think we should simplify this?  I don't see much possibility of %d in the first argument of Precondition (which is evaluating a boolean), other than a modulo operation but in that case there is a space after %.  Or let me know if I missed what you mean.

xrl1:
@szehon-ho Yeah you are right, it is really unlikely to encounter `%d` in the first argument;
 simplified it to:
```regex
Preconditions\.checkArgument\([^;]+%d[^;]+\);
```

kbendick (approved):
LGTM! Thank you!

kbendick:
> @szehon-ho Yeah you are right, it is really unlikely to encounter `%d` in the first argument I simplified it to:
> 
> ```regexp
> Preconditions\.checkArgument\([^;]+%d[^;]+\);
> ```

Could you update the PR description with the explanation of this regex (so we don't have to look it up in the conversation) @xrl1? I believe the original regex is still there.

xrl1:
@kbendick Yeah sure, updated.

szehon-ho:
Thanks @xrl1  and @kbendick for additional review

PR #5059, opened 2022-06-16
Dell: Fix bugs during documenting

wang-x-xia:
1. Make `DellClientFactory` serializable
2. To avoid the NPE when reading the table/namespace object which doesn't have the version field.


singhpk234 (approved):

openinx (approved):
LGTM,  Thanks @wang-x-xia for the contribution !

PR #4904, opened 2022-05-30
Flink: new sink base on the unified sink API

hililiwei:
Co-authored-by: Kyle Bendickson kjbendickson@gmail.com

## What is the purpose of the change
There is a preliminary proposal:
https://docs.google.com/document/d/1G4O6JidAoKgbIdy8Ts73OfG_KBEMpsW-LkXIb89I5k8/edit?usp=sharing

## Motivation & Structure

In Flink 1.12 the community focused on implementing a unified Data Sink API ([FLIP-143](https://cwiki.apache.org/confluence/display/FLINK/FLIP-143%3A+Unified+Sink+API)). The new abstraction introduces a write/commit protocol and a more modular interface where the individual components are transparently exposed to the framework.
A Sink implementor will have to provide the what and how: a [SinkWriter](https://nightlies.apache.org/flink/flink-docs-release-1.12/api/java/org/apache/flink/api/connector/sink/SinkWriter.html) that writes data and outputs what needs to be committed (i.e. committables); and a [Committer](https://nightlies.apache.org/flink/flink-docs-release-1.12/api/java/org/apache/flink/api/connector/sink/Committer.html) and [GlobalCommitter](https://nightlies.apache.org/flink/flink-docs-release-1.12/api/java/org/apache/flink/api/connector/sink/GlobalCommitter.html) that encapsulate how to handle the committables. The framework is responsible for the when and where: at what time and on which machine or process to commit.

image

[GlobalCommitter](https://nightlies.apache.org/flink/flink-docs-release-1.12/api/java/org/apache/flink/api/connector/sink/GlobalCommitter.html) can complete the action of submitting snapshots in Iceberg.
There is a QA related to Iceberg in its design documentation.

> Is the sink an operator or a topology?
> From the discussion in the long run we should give the sink developer the ability of building “arbitrary” topologies. But for Flink-1.12 we should be more focused on only satisfying the S3/HDFS/Iceberg sink.

The current Writing process of Iceberg Flink Sink is as follows:

image

In the figure above, data is written by multiple write operators, and then snapshot is submitted by an operator with 1 parallelism. Finally, an empty Sink node is added to end the whole process.
Based on the FLIP 143, the above flow can be modified to:
image

The Sink node writes data, and the GlobalCommiter operator commits the snapshot.

[...]

PR #5061, opened 2022-06-16
Flink: upsert table join failed

hililiwei:
### Reappear 
```sql
CREATE TABLE upsert_test (
  `id`  STRING UNIQUE COMMENT 'unique id',
  `data` STRING NOT NULL,
 PRIMARY KEY(`id`) NOT ENFORCED
) with ('format-version'='2', 'write.upsert.enabled'='true');

insert into upsert_test values ('1','20220507'), ('2','20220504'),('6','20220505');


CREATE TABLE upsert_sample (
  `id`  STRING UNIQUE COMMENT 'unique id',
  `data` STRING NOT NULL,
 PRIMARY KEY(`id`) NOT ENFORCED
) with ('format-version'='2', 'write.upsert.enabled'='true');

insert into upsert_sample values ('1','20220607'),('2','20220503'),('3','20220505');

insert into upsert_test
select
 t1.id as id,
 t1.data as data
from upsert_sample t1 left join upsert_test t2
on t1.id = t2.id 
where t1.data > t2.data or t2.data is null;
```

Expected Results
```
id                      data
 1                  20220607
 3                  20220505
 6                  20220505
 2                  20220504
```

Actual Results

```
     id                      data
      1                  20220607
      3                  20220505
      6                  20220505

```

or 

```
     id                      data
      1                  20220607
      3                  20220505
      3                  20220505
      6                  20220505

```
………………

**Note: This occurs occasionally.**

### Reason:
When we write data in this way, we get not only the result data, but all the cdc data.
the above sql obviously didn't mean that, the CDC data caused the table result to be abnormal.

Let's look at what happens when the ID2 data disappears.

look following SQL first:
```sql
select  t1.id as id ,t1.data as data from upsert_sample t1 left join upsert_test t2 on t1.id = t2.id  where t1.data > t2.data or t2.data is null
```
image

upsert_test data file:
```

**upsert_test/metadata**

$ java -jar /d/tools/iceberg-tools-1.0-SNAPSHOT.jar manifest2json 28dfde75-110d-409a-93fb-39dbf0daf99e-m1.avro v3.metadata.json
log4j:WARN No appenders could be found for logger (org.apache.hadoop.metrics2.lib.MutableMetricsFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
[
{"status":1,"snapshot_id":{"long":5799541658712377255},"sequence_number":null,"data_file":{"content":2,"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00000-0-f39bdce9-96cc-46cf-adb8-9e84a0577ef2-00002.avro","file_format":"AVRO","partition":{},"record_count":2,"file_size_in_bytes":317,"column_sizes":null,"value_counts":null,"null_value_counts":null,"nan_value_counts":null,"lower_bounds":null,"upper_bounds":null,"key_metadata":null,"split_offsets":null,"equality_ids":{"array":[1]},"sort_order_id":{"int":0}}},
{"status":1,"snapshot_id":{"long":5799541658712377255},"sequence_number":null,"data_file":{"content":1,"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00000-0-f39bdce9-96cc-46cf-adb8-9e84a0577ef2-00003.avro","file_format":"AVRO","partition":{},"record_count":1,"file_size_in_bytes":801,"column_sizes":null,"value_counts":null,"null_value_counts":null,"nan_value_counts":null,"lower_bounds":null,"upper_bounds":null,"key_metadata":null,"split_offsets":null,"equality_ids":null,"sort_order_id":null}},
{"status":1,"snapshot_id":{"long":5799541658712377255},"sequence_number":null,"data_file":{"content":2,"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00002.avro","file_format":"AVRO","partition":{},"record_count":2,"file_size_in_bytes":317,"column_sizes":null,"value_counts":null,"null_value_counts":null,"nan_value_counts":null,"lower_bounds":null,"upper_bounds":null,"key_metadata":null,"split_offsets":null,"equality_ids":{"array":[1]},"sort_order_id":{"int":0}}},
{"status":1,"snapshot_id":{"long":5799541658712377255},"sequence_number":null,"data_file":{"content":1,"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00003.avro","file_format":"AVRO","partition":{},"record_count":1,"file_size_in_bytes":800,"column_sizes":null,"value_counts":null,"null_value_counts":null,"nan_value_counts":null,"lower_bounds":null,"upper_bounds":null,"key_metadata":null,"split_offsets":null,"equality_ids":null,"sort_order_id":null}}
]


$ java -jar /d/tools/iceberg-tools-1.0-SNAPSHOT.jar manifest2json 28dfde75-110d-409a-93fb-39dbf0daf99e-m0.avro v3.metadata.json
log4j:WARN No appenders could be found for logger (org.apache.hadoop.metrics2.lib.MutableMetricsFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
[
{"status":1,"snapshot_id":{"long":5799541658712377255},"sequence_number":null,"data_file":{"content":0,"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00000-0-f39bdce9-96cc-46cf-adb8-9e84a0577ef2-00001.avro","file_format":"AVRO","partition":{},"record_count":2,"file_size_in_bytes":413,"column_sizes":null,"value_counts":null,"null_value_counts":null,"nan_value_counts":null,"lower_bounds":null,"upper_bounds":null,"key_metadata":null,"split_offsets":null,"equality_ids":null,"sort_order_id":{"int":0}}},
{"status":1,"snapshot_id":{"long":5799541658712377255},"sequence_number":null,"data_file":{"content":0,"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00001.avro","file_format":"AVRO","partition":{},"record_count":2,"file_size_in_bytes":415,"column_sizes":null,"value_counts":null,"null_value_counts":null,"nan_value_counts":null,"lower_bounds":null,"upper_bounds":null,"key_metadata":null,"split_offsets":null,"equality_ids":null,"sort_order_id":{"int":0}}}
]


**upsert_test/data**
$  java -jar /d/tools/avro-tools-1.11.0.jar tojson 00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00003.avro
{"file_path":"file:/C:/Users/L00561~1/AppData/Local/Temp/junit1644645413478420150/default/db/upsert_test/data/00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00001.avro","pos":1}

$ java -jar /d/tools/avro-tools-1.11.0.jar tojson  00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00002.avro
{"id":"3"}
{"id":"2"}

$ java -jar /d/tools/avro-tools-1.11.0.jar tojson  00001-0-683475d0-066a-48f2-96c9-52e4cb481fca-00001.avro
{"id":"3","data":"20220505"}
{"id":"2","data":"20220503"}


```

https://github.com/apache/iceberg/blob/547152cf3dfa5e399351b851c48edd8f916e9c48/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/FlinkSink.java#L418-L421

https://github.com/apache/iceberg/blob/547152cf3dfa5e399351b851c48edd8f916e9c48/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/BaseDeltaTaskWriter.java#L76-L101

### Resolves

We are currently using the new Sink API to fix it temporarily, FYI #4904 (add a configuration that specifies whether to enable the new version sink), that use `sink` instead of `transform`, but in this way, we can't get the CDC data.
In the above case, CDC conflicts with upsert. Do we have a better way to deal with it?

### Questions:
Maybe we didn't pinpoint the real problem.
We also wanted to find a simple way to solve it.

### DIFF
--- a/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
+++ b/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/TestFlinkUpsert.java
@@ -243,6 +243,43 @@ public class TestFlinkUpsert extends FlinkCatalogTestBase {
   }
 
   @Test
+  public void testUpsertTableJoin() {
+    try {
+      sql("CREATE TABLE upsert_test ( " +
+              "  `id`  STRING UNIQUE COMMENT 'unique id', " +
+              "  `data` STRING NOT NULL, " +
+              " PRIMARY KEY(`id`) NOT ENFORCED " +
+              " ) WITH %s", toWithClause(tableUpsertProps));
+
+      sql("insert into upsert_test values ('1','20220507'), ('2','20220504'),('6','20220505')");
+
+      sql("CREATE TABLE upsert_sample ( " +
+              "  `id`  STRING UNIQUE COMMENT 'unique id', " +
+              "  `data` STRING NOT NULL, " +
+              " PRIMARY KEY(`id`) NOT ENFORCED " +
+              " ) WITH %s", toWithClause(tableUpsertProps));
+
+      sql("insert into upsert_sample values ('1','20220607'),('2','20220503'),('3','20220505')");
+
+      sql("insert into upsert_test " +
+              "select " +
+              "t1.id as id, " +
+              "t1.data as data " +
+              "from upsert_sample t1 left join upsert_test t2 " +
+              "on t1.id = t2.id  " +
+              "where t1.data > t2.data or t2.data is null");
+
+      TestHelpers.assertRows(
+              sql("SELECT * FROM upsert_test"),
+              Lists.newArrayList(Row.of("1", "20220607"), Row.of("2", "20220504"), Row.of("3", "20220505"),
+                      Row.of("6", "20220505")));
+    } finally {
+      sql("DROP TABLE IF EXISTS %s.%s", flinkDatabase, "upsert_test");
+      sql("DROP TABLE IF EXISTS %s.%s", flinkDatabase, "upsert_sample");
+    }
+  }
+
+  @Test
   public void testPrimaryKeyFieldsAtEndOfTableSchema() {
     // This is the same test case as testPrimaryKeyFieldsAtBeginningOfSchema, but the primary key fields
     // are located at the end of the flink schema.


wuwenchi:
in your Expected Results:
```
id                      data
 1                  20220607
 3                  20220505
 6                  20220505
 2                  20220504    -------> should 20220503 ?
```

hililiwei:
In our latest tests, we found this problem not only with upsert table. Tables with no-upsert have similar problems.

rdblue:
To me, it looks like the problem is that the join is producing CDC records. A join would normally not produce output records that don't match the filter, so it is strange to have `+I[2, ...]` and `-D[2, ...]` in the join output.

I'm not sure how to solve this, but it looks like a problem with Flink. Flink shouldn't produce CDC rows for a straightforward join. It should just produce the regular SQL output set, which is:

```
id | data
---+---------
 1 | 20220607
 3 | 20220505
```

> We are currently using the new Sink API to fix it temporarily, FYI #4904 (add a configuration that specifies whether to enable the new version sink), that use sink instead of transform, but in this way, we can't get the CDC data.

Do you mean that the new sink doesn't pass the `RowKind` and you only get in inserted rows?

hililiwei:
In Flink SQL, data is transferred from one operator to another as a Chanlog Stream. Changelog Stream can be translated as either a table or a Stream.

Iceberg should be more like a Dynamic Table. In our case above, I think it should materialize changes rather than store changes in the table as they were.

> Do you mean that the new sink doesn't pass the RowKind and you only get in inserted rows?

Sink will not just receive Inserted Rows. With Flink CDC, it can receive change log normally.

wuwenchi:
For the data with id=2
1. Originally there is a piece of data, the operation is insert, which is `+I`,
2. The two data in the CDC data are `+I` and `-D`

If the cdc data is regarded as a whole, and the cdc is inserted first and then deleted, the original piece of data should be left in the end.  But in fact, on the whole, all operations on the data are `+I, +I, -D`, and the second `+I` was wrongly processed as an update to the original data, which eventually led to the disappearance of this data. (In fact, the second `+I` is an insert into the intermediate result set)

So, in this case, should batch mode be used? Generate the final result directly for `select` in batch mode:
```
select  t1.id as id ,t1.data as data from upsert_sample t1 left join upsert_test t2 on t1.id = t2.id  where t1.data > t2.data or t2.data is null  :
+----+--------------------------------+--------------------------------+
| op |                             id |                           data |
+----+--------------------------------+--------------------------------+
| +I |                              1 |                       20220607 |
| +I |                              3 |                       20220505 |
+----+--------------------------------+--------------------------------+
```

This is just a temporary solution that I can think of. But in stream mode, there is indeed a problem with the result processed by iceberg now.

hililiwei:
@wuwenchi, Thank you first.

In batch case, this problem does not occur. Unfortunately, in our production, data is streaming. Batch tasks are executed using Spark. But the current trend (at least within our department) is that more and more tasks are moving from batch to stream.

Personally, if there is no blocking, then it seems like a good time to use the new Sink API.

openinx:
Few questions: 

1.  Why the expected results is  the following ?  Does the record `<id=6, data=20220505>` really to be the expected record in the output for a __LEFT JOIN__ ? 

```bash
id                      data
 1                  20220607
 3                  20220505
 6                  20220505
 2                  20220504
```

2. As we apache iceberg format v2 don't support incremental streaming reading, then how could we accomplish this streaming read and streaming join ?  Shouldn't we solve the incremental streaming read semantics firstly and then handle the complex join semantic ?

hililiwei:
1. condition :`where t1.data > t2.data or t2.data is null`
 <id=6, data=20220505> hit the `t2.data is null` It will be considered new data.

2. In general, we don't need the two join tables to be iceberg tables. In fact, in our case, the two join tables are Kafka and a dim table, the target is an iceberg table.

rdblue:
> In Flink SQL, data is transferred from one operator to another as a Chanlog Stream.

The stream of changes from the debugger screenshot appears incorrect to me. Even if join outputs are modeled this way, there should not be a delete and an insert for ID 2. Why would an unmatched row produce anything from the join?

kbendick:
> Note: This occurs occasionally.

Does this mean that sometimes you get one set of results from the `select *` query and sometimes you get the other set of results listed in your “actual results”?

The result set with id = 3 in it twice is concerning to me. When I’ve seen that before, it’s usually a clear sign of an issue. Often times it has occurred because one of the rows has a different kind in it, but it’s coming out on read with the schema without rowkind so then there is duplicate data.

I cannot confirm that is what is happening here, but I’ve seen that before in other occurrences.

kbendick:
If the INSERT INTO statements are all run one by one, is the result more stable?

Also, can you provide the intermediate result set for the LEFT JOIN and then the final result of `upsert_test` (the target table that records are being inserted into?

It seems some people are somewhat hung up on the LEFT JOIN result vs the final result.

kbendick:
Also, for clarity, in the SELECT statement before the insert, did you mean to use `select upsert_sample.id, upsert_sample.data`?

kbendick:
> > In Flink SQL, data is transferred from one operator to another as a Chanlog Stream.
> 
> The stream of changes from the debugger screenshot appears incorrect to me. Even if join outputs are modeled this way, there should not be a delete and an insert for ID 2. Why would an unmatched row produce anything from the join?

I believe we get a +I and -D row in the changelog because a row for ID=2 entered the join and then was not selected for the join. 

As we’re selecting from `upsert_sample` as the left side of the join (aka t1), that has the row `('2','20220503')` in it. I am wondering if the `unique` keyword is adding an operator to the DAG to cause the left join to have to consider `('2','20220503')` during the join (and thus a temporary +I record) because of the enforced uniqueness constraint.

So id=2 is inserted due to its existence in upsert_sample and then deleted because it’s not an output of the join. But it is an input of the join. Is that not what’s happening in the debugger output?

Also @hililiwei what happens if you remove the `UNIQUE` constraint? Is that even possible for `id` to the the primary key? And what happens if you `select *` and not specify t1.id and t1.data?

Reo-LEI:
image
@hililiwei I reproduced this case in my local env, the `id=2` record will be deleted is Flink's behavior, has nothing to do with iceberg. And I think this record should be delete becasue the record `('2','20220503')` does not meet this filter condition `t1.data > t2.data`. So the expected final result should be `('1','20220607'), ('3','20220505')` same as the batch result.    
<br/>

image
> so it is strange to have +I[2, ...] and -D[2, ...] in the join output.
@rdblue If we adjust the env setting, the +I[2, ...] and -D[2, ...] will disappear occasionally. I think the occurrence of these records is related to the timing of checkpoint and the distribution of data.
<br/>

> As we apache iceberg format v2 don't support incremental streaming reading, then how could we accomplish this streaming read and streaming join ?
@openinx Because this test only setting the env as `StreamTableEnvironment`, but there not setting the `ScanContext.isStreaming` as true(that is false by default). Therefor, the iceberg table will be scan by batch but not streaming.

hililiwei:
@Reo-LEI The two tables that join can be non-Iceberg tables. But when your target table has ID2 data, the data that should not be deleted will be deleted.

hililiwei:
@kbendick When I remove the `UNIQUE`, it's going to show up the same way. 
But aone thing, when I use temporal Join, I have to have a primary key field.
>Also, can you provide the intermediate result set for the LEFT JOIN and then the final result of upsert_test (the target table that records are being inserted into?
When I run left Join in Flink SQL, its result set is:
```
id                      data
 1                  20220607
 3                  20220505
```
It's the same as @Reo-LEI

PR #5062, opened 2022-06-16
Core: In HadoopTableOperations, replace Util.getFs call with getFileSystem

chakradhar2:
In HadoopTableOperations, we use getFileSystem method to get FileSystem, which internally calls  Util.getFs but it was missing in some methods.
Using getFileSystem at all places in HadoopFileSystem allows to overriding method.


ajantha-bhat (approved):
Thanks for the contribution @chakradhar2

pvary:
Thanks @chakradhar2 for the fix and @ajantha-bhat for the review!

PR #4818, opened 2022-05-19
Python: Fix issue related to Pyenchant with executing linters on Apple Silicons

dhruv-pratap:
If you are using an M1 Mac to run tox, there is currently [an issue with pyenchant](https://github.com/pyenchant/pyenchant/issues/265) that you would run into while running `linters`. This happens because python distributions installed by `pyenv` are in the default M1 architecture (aarch64), and can't find a compatible enchant library. As a workaround, you can follow the below steps. 

Install enchant with Homebrew 
```shell
brew install enchant
```
Set environment variable `PYENCHANT_LIBRARY_PATH` to point to the installed enchant library before running `tox`.
```shell
export PYENCHANT_LIBRARY_PATH = /opt/homebrew/lib/libenchant-2.dylib
```
You can set this environment variable in your `~/.zshrc` or in your IDE to avoid doing it everytime.

PR #5066, opened 2022-06-16
Python: Renable pylint

Fokko:
For some reason, pylint fell through the cracks when migrating from tox to pre-commit


Fokko:
I've removed the spellcheck because it relies on a C lib enchant that doesn't play very well with virtual-env. It can't find the binding anymore. I spend quite some time to get it to work, but no luck. Also looking at https://github.com/apache/iceberg/pull/4818 I've removed it for now since it is also not very friendly for new people to ask them to install system packages.

samredai (commented):
Needs to be rebased but otherwise LGTM

python/tests/io/test_pyarrow.py
```diff
@@ -14,6 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+# pylint: disable=W0212,W0613
```
dramaticlly, line 17: Nit: curious on how pylint update its code to warning mapping, from understanding perspective, it's easier to read `disable=protected-access`

dramaticlly (approved):
LGTM, sad to see dictionary going away

python/tests/io/test_pyarrow.py
Fokko: Thanks, I like that a lot. Just updated the code.

danielcweeks (approved):

PR #5068, opened 2022-06-16
Python: Replace vars with cached_property decorator

Fokko:
Instead of doing the caching ourselves, functools has a nice decorator that can do that for us. I also removed the type ignore to avoid having an accidental None.


Fokko:
@samredai would you have time to look at this? Since you're the original author 👍🏻

danielcweeks (approved):

PR #5075, opened 2022-06-17
Spark: Add `__metadata_col` metadata in metadata columns when doing Schema Conversion

singhpk234:
### About the change : 

Presently when doing schema conversion we were setting `metadata` for all the the columns, but we should add `____metadata_col` in metadata column which can be used by spark to check if the column is `metadata` col and drop it if required like, 
here : 
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/v2Commands.scala#L221-L225

Note : `MetadataAttribute` extractor uses the above key in attribute meta data to find if the attribute is metadata attribute or not.

This PR includes : 
(i) Fix for above 
(ii) Fix an existing minor typo in `TestSparkSchemaUtil`

---- 
### Testing Done
Added an UT for the same.

cc @rdblue @aokolnychyi @jackye1995 @RussellSpitzer

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
@@ -20,6 +20,7 @@
 package org.apache.iceberg.spark;
 
 import java.util.List;
+import org.apache.iceberg.MetadataColumns;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.types.Type;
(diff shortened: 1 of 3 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
```diff
@@ -59,8 +61,11 @@ public DataType struct(Types.StructType struct, List<DataType> fieldResults) {
     for (int i = 0; i < fields.size(); i += 1) {
       Types.NestedField field = fields.get(i);
       DataType type = fieldResults.get(i);
-      StructField sparkField = StructField.apply(
-          field.name(), type, field.isOptional(), Metadata.empty());
+      Metadata metadata = Metadata.empty();
+      if (MetadataColumns.isMetadataColumn(field.name())) {
+        metadata = new MetadataBuilder().putBoolean(MetadataColumns.METADATA_COL_ATTR_KEY, true).build();
+      }
```
rdblue, line 67: Could this be refactored into a separate method? I think that would be cleaner. And, this should identify metadata columns using column IDs, since they are available in Iceberg schemas.

singhpk234: makes sense, made it consistent to 3.3 review comments as well.

rdblue (approved):

rdblue:
Thanks, @singhpk234!

PR #5077, opened 2022-06-17
API: Refactor FileScanTask

aokolnychyi:


api/src/main/java/org/apache/iceberg/InputSplit.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.iceberg;
+
+import java.util.Collection;
+
+/**
+ * A scan task that may include partial input files, multiple input files or both.
+ *
+ * @param <T> the type of scan tasks
+ */
+public interface InputSplit<T extends ScanTask> extends ScanTask {
```
aokolnychyi, line 29: This is a parent interface added to avoid breaking `CombinedScanTask`.

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -29,7 +29,7 @@
  * Scan objects are immutable and can be shared between threads. Refinement methods, like
  * {@link #select(Collection)} and {@link #filter(Expression)}, create new TableScan instances.
  */
-public interface Scan<T extends Scan<T>> {
+public interface Scan<ThisT, T extends ScanTask, S extends InputSplit<T>> {
```
aokolnychyi, line 32: I am debating whether we need to attach a boundary to `ThisT`. It won't harm but we don't do that in any other places and it makes the definition a bit more cumbersome as `Scan` has 3 params now.

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -113,26 +113,23 @@
...
+   * Plan input split tasks for this scan and balance the work.
    * <p>
-   * Tasks created by this method may read partial input files, multiple input files, or both.
+   * Tasks created by this method may read partial input files, multiple input files or both.
    *
-   * @return an Iterable of tasks for this scan
+   * @return an Iterable of input splits required by this scan
    */
-  CloseableIterable<CombinedScanTask> planTasks();
+  CloseableIterable<S> planTasks();
```
aokolnychyi, line 132: The plan for other scans like CDC to use `InputSplit<T>` and `BaseInputSplit` instead of `CombinedScanTask`.

Alternatively, we could add something like `planInputSplits` instead of parameterizing `Scan` with `S extends InputSplit<T>` and keep `planTasks` only in `TableScan` for compatibility. Not sure it is any better.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +83,57 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
             splitFiles),
         BaseCombinedScanTask::new);
   }
+
+  public static <T extends SplittableScanTask<T>> CloseableIterable<InputSplit<T>> planInputSplits(
```
aokolnychyi, line 87: The existing logic we had for splitting and combining expressed in terms of `SplittableScanTask`.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
```diff
@@ -174,7 +176,7 @@ private CloseableIterable<FileScanTask> appendFilesFromSnapshots(List<Snapshot>
   }
 
   /**
-   * This method doesn't perform validation, which is already done by the caller {@link #planFiles()}
+   * This method doesn't perform validation, which is already done by the caller {@link #planFiles()} ()}
```
rdblue, line 179: Looks like a typo.

api/src/main/java/org/apache/iceberg/BaseInputSplit.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+
+public class BaseInputSplit<T extends ScanTask> implements InputSplit<T> {
+  private final List<T> tasks;
+
+  public BaseInputSplit(List<T> tasks) {
+    Preconditions.checkNotNull(tasks, "tasks cannot be null");
+    this.tasks = Collections.unmodifiableList(Lists.newArrayList(tasks));
```
aokolnychyi, line 33: I need to double check Kryo serialization. Maybe, I should return an unmodifiable wrapper only in `files()`.

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
```diff
@@ -217,6 +218,18 @@ public Iterable<FileScanTask> split(long splitSize) {
       throw new UnsupportedOperationException("Cannot split a task which is already split");
     }
 
+    @Override
+    public boolean isAdjacent(FileScanTask other) {
+      return file().equals(other.file()) && offset + len == other.start();
```
aokolnychyi, line 223: I copied this logic from an existing place but we have to remember that files will be equal only if references are equal, which I think is true here. Our data and delete file implementations don't override equals.

api/src/main/java/org/apache/iceberg/BaseInputSplit.java
rdblue: Yeah, for serialization we usually want to go with a Java array.

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
rdblue: I think this is correct.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
```diff
@@ -0,0 +1,60 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+public interface SplittableScanTask<ThisT> extends ScanTask {
```
rdblue, line 22: I like the addition of `SplittableScanTask`. This looks pretty good.

api/src/main/java/org/apache/iceberg/InputSplit.java
rdblue: I'm not sure about this name. We purposely didn't use the name `InputSplit` initially because we didn't want to conflict with Hadoop and other uses. Now, I'm afraid that it will duplicate those classes, and that it will be confusing to have both tasks and splits, with an unclear relationship between the two.

It also isn't clear why an `InputSplit` would have files, but a scan task would not. What about `CommonFileTask` or something similar?

rdblue:
Overall I think this looks good. Are you happy with the direction this is going, @aokolnychyi?

api/src/main/java/org/apache/iceberg/InputSplit.java
aokolnychyi: Oh, I tried at least 3 different names and went back and forth with them. Note this is a task that is usually referenced by an input split in query engines. The ideal name is `CombinedScanTask` as it is a task created by combining smaller tasks into a bigger one. However, we can't parameterize `CombinedScanTask` to keep compatibility.

I am afraid `CommonFileTask` is very generic and does not imply it may consist of multiple smaller tasks. Maybe, something with `Packed`? We need a reasonable synonym for `CombinedScanTask`. Any ideas are welcome.

I am also not sure `files()` is very appropriate. I inherited it from `CombinedScanTask` but it makes more sense there.

aokolnychyi:
@rdblue, I spent quite some time looking at it so I am no longer sure :) We need to sort out some details but it does unlock having tasks that are not `FileScanTask` and also parametrizes our scans. It is probably worth it.

api/src/main/java/org/apache/iceberg/InputSplit.java
aokolnychyi: Maybe, something like `CompoundScanTask<T>`?
Or maybe drop it at all and define `CombinedChangelogScanTask` that would extend `ScanTask` directly?

aokolnychyi: I'll think more about this but this is the biggest open question at the moment. Ideas are welcome.

api/src/main/java/org/apache/iceberg/ContentScanTask.java
```diff
@@ -0,0 +1,63 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import org.apache.iceberg.expressions.Expression;
+
+public interface ContentScanTask<F extends ContentFile<F>> extends ScanTask {
```
RussellSpitzer, line 24: I think we should have a short description here of what the purpose of this interface is? Maybe

"A Scan task which reads the rows out of an Iceberg non-metadata file"? 

Maybe that is too generic to be useful?

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Splits this scan task into several smaller scan tasks, each of {@code splitSize} size.
+   *
+   * @param splitSize the target size of each new scan task
+   * @return an Iterable of smaller tasks
+   */
+  Iterable<ThisT> split(long splitSize);
+
+  /**
+   * Checks if this task is adjacent to another.
```
RussellSpitzer, line 32: I think this may need a bit more explanation, "Checks if this task is adjacent to another. Two tasks are adjacent if they are scanning the same file and one task begins where exactly where the other ends." ?

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * @param other another task
+   * @return whether this task is adjacent to another
+   */
+  boolean isAdjacent(ThisT other);
+
+  /**
+   * Combines two tasks into one.
```
RussellSpitzer, line 40: Is this only allowed for Adjacent Tasks? Or are we leaving that up to the implementation to decide?

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * The total number of bytes that should be read by this scan task.
+   *
+   * @return the total number of bytes to read
+   */
+  long totalSizeBytes();
+
+  /**
+   * The total number of files that should be open by this scan task.
```
RussellSpitzer, line 55: "open" -> "opened"

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
RussellSpitzer: In the other code we assume there is an ordering of the files. 

1 is adjacent to 2
2 is not adjacent to 1

We may want to change the api definition just to note that it is not commutative

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
RussellSpitzer, line 22: Do we need this to be SplittableScanTask since I think our only usage is in combining? I do like the idea here though I just wonder if we will actually need to re-split or if we should always be producing minimally sized splits (one row group) and then combining them?

Avro is still an issue I guess ...

api/src/main/java/org/apache/iceberg/InputSplit.java
aokolnychyi: I went with `ScanTaskGroup` for now. It acts simply like a group of tasks now. Let me know what you think.

api/src/main/java/org/apache/iceberg/Scan.java
aokolnychyi: This is still open.

api/src/main/java/org/apache/iceberg/Scan.java
aokolnychyi: This is still open.

core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java
aokolnychyi: Fixed.

api/src/main/java/org/apache/iceberg/BaseInputSplit.java
aokolnychyi: Doesn't Kryo support mutable collections? I remember having issues with immutable collections only.

I was reluctant to use an array because of generics. I will most likely will need to pass `Class<T>` around.

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
aokolnychyi: What about updating the code to cover both in the future?

api/src/main/java/org/apache/iceberg/ContentScanTask.java
aokolnychyi: Added a note. The note is generic but the class is generic too :)

api/src/main/java/org/apache/iceberg/FileScanTask.java
```diff
@@ -20,64 +20,37 @@
 package org.apache.iceberg;
 
 import java.util.List;
-import org.apache.iceberg.expressions.Expression;
 
 /**
- * A scan task over a range of a single file.
+ * A scan task over a range of a single data file.
```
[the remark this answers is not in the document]
aokolnychyi: Slightly changed it.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
aokolnychyi: Added a sentence below.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
```diff
@@ -0,0 +1,60 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+public interface SplittableScanTask<ThisT> extends ScanTask {
+  /**
+   * Splits this scan task into several smaller scan tasks, each of {@code splitSize} size.
```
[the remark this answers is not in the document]
aokolnychyi: Added a sentence clarifying the behavior.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
aokolnychyi: Hm, good question. I initially though about leaving that up to implementations but I am no longer sure. Maybe, we can change the doc to state that the passed task is adjacent. Then we won't have to check it again in classes that can only stitch adjacent tasks.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
aokolnychyi: Changed.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
aokolnychyi: We actually leverage `split` before combining in `TableScanUtil`.

Avro is still splittable, right? I am not sure whether we persist split offsets in the metadata so planning may not be as optimal as for Parquet.

szehon-ho (commented):
Generally looks good, added a few comments.

There may be quite some code to change to adopt these new ways to split/combine tasks?

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
flyrain: Minor suggestion: Rename it to `combineAdjacentTask` if it is only for adjacent tasks?

api/src/main/java/org/apache/iceberg/CombinedScanTask.java
```diff
@@ -24,13 +24,18 @@
 /**
  * A scan task made of several ranges from files.
  */
-public interface CombinedScanTask extends ScanTask {
+public interface CombinedScanTask extends ScanTaskGroup<FileScanTask> {
   /**
    * Return the {@link FileScanTask tasks} in this combined task.
    * @return a Collection of FileScanTask instances.
    */
   Collection<FileScanTask> files();
```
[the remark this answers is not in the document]
aokolnychyi: I am fine either way. This naming probably makes sense given that we return `FileScanTask`s. I just wanted to have a more flexible naming in `ScanTaskGroup` so I added a new method.

api/src/main/java/org/apache/iceberg/BaseInputSplit.java
aokolnychyi: Added tests to cover mutable lists but I generally don't mind passing `Class<T>` to switch to arrays.

api/src/main/java/org/apache/iceberg/SplittableScanTask.java
aokolnychyi: Renamed and changed the doc.

api/src/main/java/org/apache/iceberg/ContentScanTask.java
```diff
@@ -0,0 +1,68 @@
...
+   * The number of bytes to scan from the {@link #start()} position in the file.
+   *
+   * @return the length of this scan range in bytes
+   */
+  long length();
+
+  /**
+   * Returns the residual expression that should be applied to rows in this file scan.
+   * <p>
+   * The residual expression for a file is a filter expression created from the scan's filter, inclusive
```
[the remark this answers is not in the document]
aokolnychyi: Rewrote this sentence.

api/src/main/java/org/apache/iceberg/Scan.java
```diff
@@ -113,26 +113,23 @@
...
-   * Plan the {@link FileScanTask files} that will be read by this scan.
+   * Plan tasks for this scan without trying to balance the work.
    * <p>
-   * Each file has a residual expression that should be applied to filter the file's rows.
-   * <p>
-   * This simple plan returns file scans for each file from position 0 to the file's length. For
-   * planning that will combine small files, split large files, and attempt to balance work, use
-   * {@link #planTasks()} instead.
+   * Use {@link #planTasks()} for planning that will attempt to balance the work
+   * by combining small or splitting large files.
```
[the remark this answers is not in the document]
aokolnychyi: Changed.

flyrain:
LGTM

flyrain (approved):

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +83,57 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
...
+      CloseableIterable<T> tasks,
+      long splitSize, int lookback, long openFileCost) {
+
+    Preconditions.checkArgument(splitSize > 0, "Invalid split size (negative or 0): %s", splitSize);
+    Preconditions.checkArgument(lookback > 0, "Invalid split planning lookback (negative or 0): %s", lookback);
+    Preconditions.checkArgument(openFileCost >= 0, "Invalid file open cost (negative): %s", openFileCost);
+
+    // capture manifests which can be closed after scan planning
+    CloseableIterable<T> splitTasks = CloseableIterable.combine(
+        FluentIterable.from(tasks).transformAndConcat(input -> input.split(splitSize)),
```
rdblue, line 97: This could check for `SplittableScanTask` and call split, or just pass the task through.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +83,57 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
...
+        tasks);
+
+    Function<T, Long> weightFunc = task -> Math.max(task.totalSizeBytes(), task.totalFilesCount() * openFileCost);
```
rdblue, line 100: Should `totalSizeBytes` and `totalFilesCount` be part of `ScanTask` instead?

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +83,57 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
...
+    }
+
+    List<T> combinedTasks = Lists.newArrayList();
+    T lastTask = null;
+
+    for (T task : tasks) {
+      if (lastTask != null) {
+        if (lastTask.isAdjacent(task)) {
+          // merge with the last task
+          lastTask = lastTask.combineWithAdjacentTask(task);
```
rdblue, line 121: Maybe change `combineWithAdjacentTask(ThisT)` to `combineWithAdjacentTask(ScanTask)`. Then it is up to the implementation whether it can combine.

api/src/main/java/org/apache/iceberg/CombinableScanTask.java
```diff
@@ -0,0 +1,45 @@
...
+ * @param <ThisT> the child Java API class
+ */
+public interface CombinableScanTask<ThisT> extends ScanTask {
+  /**
+   * Checks if this task can be combined with a given task.
+   *
+   * @param other another task
+   * @return whether the tasks can be combined
+   */
+  boolean isCombinableWith(ScanTask other);
```
rdblue, line 34: `canCombineWith`? The word "combinable" is hard to understand.

rdblue: Also, default to `false`?

api/src/main/java/org/apache/iceberg/CombinedScanTask.java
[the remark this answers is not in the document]
rdblue: I'd probably leave it.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +86,71 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
...
+  public static <T extends ScanTask> List<T> combineTasks(List<T> tasks) {
+    if (tasks.isEmpty()) {
+      return tasks;
+    }
+
+    List<T> combinedTasks = Lists.newArrayList();
+    CombinableScanTask<? extends T> lastCombinableTask = null;
+
+    for (T task : tasks) {
+      if (task instanceof CombinableScanTask<?>) {
```
rdblue, line 129: I don't think this is correct. It doesn't matter if the next task is combineable. It only matters if the last task was. And we can't keep around the last _combineable_ task because then we would possibly combine tasks out of order.

I think `lastCombineableTask` should be `lastTask` and this should check whether `lastTask` is a combineable in order to try combining with the current task. The new task, if not combined, should always be set as the new `lastTask`.

api/src/main/java/org/apache/iceberg/BaseScanTaskGroup.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+
+public class BaseScanTaskGroup<T extends ScanTask> implements ScanTaskGroup<T> {
+  private final List<T> tasks;
+
+  public BaseScanTaskGroup(List<T> tasks) {
+    Preconditions.checkNotNull(tasks, "tasks cannot be null");
+    this.tasks = Lists.newArrayList(tasks);
```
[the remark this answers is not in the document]
rdblue: We could just use `Object[]` then?

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +86,71 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
...
+
+    return CloseableIterable.transform(
+        CloseableIterable.combine(
+            new BinPacking.PackingIterable<>(splitTasks, splitSize, lookback, weightFunc, true),
+            splitTasks),
+        combinedTasks -> new BaseScanTaskGroup<>(combineTasks(combinedTasks)));
+  }
+
+  @SuppressWarnings("unchecked")
+  public static <T extends ScanTask> List<T> combineTasks(List<T> tasks) {
```
aokolnychyi, line 120: I've struggled quite a bit trying to support both split and non-split scan tasks in this method. I had to accept arbitrary scan tasks in `isAdjacent` and `combineWithAdjacent`, which made it look weird. That's why I introduced a separate interface `CombinableScanTask` but that doesn't really make it great.

api/src/main/java/org/apache/iceberg/CombinableScanTask.java
aokolnychyi: I am not a big fan of this interface but I created it trying to support arbitrary scan tasks. This is something that is only implemented by our existing `SplitScanTask`, which is not public. That's why I did not add default methods. We don't break anything.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
aokolnychyi: You are right. The old logic assumed we can combine only tasks of the same type so this logic worked. If we decide to go this route and have `CombinableScanTask`, I'll adapt. I didn't want to change this logic completely before discussing `CombinableScanTask`.

api/src/main/java/org/apache/iceberg/BaseScanTaskGroup.java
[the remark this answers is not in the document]
aokolnychyi: Yeah, we can use an object array.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
aokolnychyi: This is done.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
aokolnychyi: Done.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
aokolnychyi: Done.

api/src/main/java/org/apache/iceberg/CombinableScanTask.java
aokolnychyi: No longer applies.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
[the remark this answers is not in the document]
aokolnychyi: Changed.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
aokolnychyi: I updated.

api/src/main/java/org/apache/iceberg/MergeableScanTask.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.apache.iceberg;
+
+/**
+ * A scan task that can be potentially merged with other scan tasks.
+ *
+ * @param <ThisT> the child Java API class
+ */
+public interface MergeableScanTask<ThisT> extends ScanTask {
```
aokolnychyi, line 27: I kept it separate from `SplittableScanTask` as `FileScanTask` is splittable but `SplitScanTask` is mergable.

core/src/test/java/org/apache/iceberg/util/TestTableScanUtil.java
```diff
@@ -88,4 +94,109 @@ public void testPlanTaskWithDeleteFiles() {
           expectedCombinedTasks.get(i).files(), combinedScanTasks.get(i).files());
     }
   }
+
+  @Test
+  public void testTaskGroupPlanning() {
```
aokolnychyi, line 99: These are tests where only some tasks are splittable and mergable.

api/src/main/java/org/apache/iceberg/BaseScanTaskGroup.java
```diff
@@ -0,0 +1,62 @@
...
+  public BaseScanTaskGroup(Collection<T> tasks) {
+    Preconditions.checkNotNull(tasks, "tasks cannot be null");
+    this.tasks = tasks.toArray();
+  }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public Collection<T> tasks() {
+    if (taskList == null) {
+      synchronized (this) {
```
rdblue, line 42: I don't have a problem with this, but do you really expect this to be accessed from different threads after construction?

rdblue (approved):

rdblue:
Looks good to me! Merge when you're ready.

Do we also want to deprecate `CombinedScanTask` so we can just use the group after 1.0?

api/src/main/java/org/apache/iceberg/BaseScanTaskGroup.java
aokolnychyi: I did it just in case given how hard it would be debug such issues. Better be safe than sorry :)

aokolnychyi:
I haven't made my mind on deprecating `CombinedScanTask` cause it is so widely used...
We will also have to break `TableScan` if we switch to `ScanTaskGroup`.

aokolnychyi:
Thanks for reviewing, @RussellSpitzer @szehon-ho @flyrain @rdblue! I know it was a tricky one.

rdblue:
> I haven't made my mind on deprecating CombinedScanTask cause it is so widely used...
> We will also have to break TableScan if we switch to ScanTaskGroup.

Let's not then. Not worth it.

core/src/main/java/org/apache/iceberg/util/TableScanUtil.java
```diff
@@ -78,4 +86,65 @@ public static CloseableIterable<CombinedScanTask> planTasks(CloseableIterable<Fi
...
+        FluentIterable.from(tasks).transformAndConcat(task -> {
+          if (task instanceof SplittableScanTask<?>) {
+            return ((SplittableScanTask<? extends T>) task).split(splitSize);
+          } else {
+            return ImmutableList.of(task);
+          }
+        }),
+        tasks);
+
+    Function<T, Long> weightFunc = task -> Math.max(task.sizeBytes(), task.filesCount() * openFileCost);
```
sumedhsakdeo, line 110: This is resulting in unbalanced bin-packing. Interestingly previous function planTasks was handling it correctly.

See https://github.com/apache/iceberg/pull/5077/changes#diff-0b0affef002de35215ed5459cf6963fd0d5bf6d5403e67440b7ef97a35189454L70

Was there any particular reason to deviate for planTaskGroups?
CC: @teamurko

PR #5083, opened 2022-06-20
support update/delete/merge into on hadoopTables using Spark

CodingCat:
when upgrading to Spark 3.2, I found an issue that we cannot use update/delete/merge into with tables we created with `HadoopTables.create()`, this issue doesn't exist in 3.0 - 3.1. 

Previously, we use Spark to read and merge/delete/update hadoop tables with the following approach

```scala
val hadoopTableDF = spark.read.format("iceberg").load(path)
hadoopTableDF.createOrReplaceTempView("target")
newDF.createOrReplaceTempView("source")
spark.sql("MERGE INTO target using source on target.id = source.id WHEN MATCHED THEN... WHEN NOT MATCHED THEN ...")
```

this doesn't work because the analyzer rules in iceberg with Spark 3.2 only recognize tables already registered in catalog (it will throw something like `MERGE INTO is not supported` even we set extensions already)

While I am aware of that catalog usage is recommended in production, I think this compatibility is still needed

* any breaking change from Spark 3.0/3.1 to Spark 3.2 is not desired anyway and the incompatibility between APIs of iceberg is not desired either, in this case, tables created with HadoopTables cannot be read by Spark
* strategically, I think many users in Delta Lake have been used to "path tables" (like ourselves), this incompatibility is a barrier for us to completely move over
* using pure path table is a lightweight approach for experiments, we do not need to clean those temp tables names in catalog but only creating tables in some s3 bucket with TTL
* personally, we do have many tables already created as hadoop table , move everything to catalog is an undesired part but imposed by iceberg when we upgrade to Spark 3.2


CodingCat:
@kbendick, @rdblue would you mind giving a review? thanks!

rdblue:
@aokolnychyi, can you take a look? You're most familiar with this part of the code.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
```diff
@@ -49,25 +51,42 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
...
-          val operation = buildRowLevelOperation(tbl, DELETE)
-          val table = RowLevelOperationTable(tbl, operation)
-          val rewritePlan = operation match {
-            case _: SupportsDelta =>
-              buildWriteDeltaPlan(r, table, cond)
-            case _ =>
-              buildReplaceDataPlan(r, table, cond)
+          rewriteDeleteFromTable(d, r, tbl)
+        case v: View =>
+          val relations = v.children.collect { case r: DataSourceV2Relation if r.table.isInstanceOf[SparkTable] =>
```
rdblue, line 59: I think the logic here needs to be much more strict. I think this should validate that `v.child` is the `DataSourceV2Relation`. Almost any other nodes in the logical plan would make this invalid. For example, if you had registered `table.select(...)` instead of just `table`, then it is unclear what deleting from that view should do. We only want to support this if the plan is only a relation.

CodingCat: this is a great point, updated

CodingCat:
@aokolnychyi would you please giving a review when you have a chance? thanks!

CodingCat:
ping @kbendick, @jackye1995, @nastra, @aokolnychyi, @danielcweeks, @rdblue @singhpk234 @RussellSpitzer  for a review , thanks!

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
```diff
@@ -49,25 +51,42 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
...
-          val table = RowLevelOperationTable(tbl, operation)
-          val rewritePlan = operation match {
-            case _: SupportsDelta =>
-              buildWriteDeltaPlan(r, table, cond)
-            case _ =>
-              buildReplaceDataPlan(r, table, cond)
+          rewriteDeleteFromTable(d, r, tbl)
+        case v: View =>
+          val viewChild = v.child
+          val icebergTableView = v.children.size == 1 && viewChild.isInstanceOf[DataSourceV2Relation] &&
```
kbendick, line 60: Do we have to use `View`? Do you know what changed?

I’m by no means a catalyst expert, but this seems like an abuse of `View`. Maybe I’m mistaken and this is not uncommon pattern.

kbendick (commented):
Is there nothing we can change or mix into the Hadoop catalog to support this change instead?

And what about query plans? Have we compared before and after this change? Given the code changed, I’d be interested to see that.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
CodingCat: the major reason to use view is that we have no ways to refer to a table created with HadoopTables.create()..which is usually a pure directory path.... so we have to load the table and then register it as a view, and then we can refer it with the view name in the statement like MERGE INTO, etc.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
```diff
@@ -49,25 +51,42 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
...
+          if (icebergTableView) {
+            val dataSourceV2Relation = viewChild.asInstanceOf[DataSourceV2Relation]
+            rewriteDeleteFromTable(d, dataSourceV2Relation,
+              dataSourceV2Relation.table.asInstanceOf[SupportsRowLevelOperations])
+          } else {
+            throw new AnalysisException(s"$v is not an Iceberg table")
           }
-          // keep the original relation in DELETE to try deleting using filters
-          DeleteFromIcebergTable(r, Some(cond), Some(rewritePlan))
-
```
rdblue, line 65: Nit: I think this whitespace was helpful.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
```diff
@@ -49,25 +51,42 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
...
-          val rewritePlan = operation match {
-            case _: SupportsDelta =>
-              buildWriteDeltaPlan(r, table, cond)
-            case _ =>
-              buildReplaceDataPlan(r, table, cond)
+          rewriteDeleteFromTable(d, r, tbl)
+        case v: View =>
+          val viewChild = v.child
+          val icebergTableView = v.children.size == 1 && viewChild.isInstanceOf[DataSourceV2Relation] &&
+            viewChild.asInstanceOf[DataSourceV2Relation].table.isInstanceOf[SparkTable]
```
rdblue, line 61: Couldn't this be done in a match expression?

I think it would make sense to do this:

```scala
  // omit if EliminateSubqueryAliases works on the whole tree, not just the current node
  EliminateSubqueryAliases(v.child) match {
    case rel @ DataSourceV2Relation(table, Some(_), None) =>
      rewriteDeleteFromTable(d, rel, table)
  }
```

rdblue: Or maybe combine this with the outer case and just have a different way of getting the relation, wrapped by a view.

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
```diff
@@ -49,25 +51,42 @@ import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
...
+          if (icebergTableView) {
+            val dataSourceV2Relation = viewChild.asInstanceOf[DataSourceV2Relation]
+            rewriteDeleteFromTable(d, dataSourceV2Relation,
+              dataSourceV2Relation.table.asInstanceOf[SupportsRowLevelOperations])
+          } else {
+            throw new AnalysisException(s"$v is not an Iceberg table")
```
rdblue, line 67: In this case, I think we should simply return the original plan unmodified. Refactoring to use more specific match expressions would help here.

CodingCat:
> Is there nothing we can change or mix into the Hadoop catalog to support this change instead?
> 
> And what about query plans? Have we compared before and after this change? Given the code changed, I’d be interested to see that.

@kbendick , I don't think we can make it to work with Hadoop catalog which requires a fixed path for all tables but HadoopTables.create() allows arbitrary paths

regarding the query plan, without the change, we cannot generate any query plan since the table cannot be recognized, but with the change, basically we will see the analyzed plan like

```
AppendData RelationV2[id#1069] /var/folders/q3/rsbny1cj6vsc7nkydr30vv2w0000gn/T/TestUpdate8997120534842900007, false
+- Project [id#1067 AS id#1073]
   +- Join LeftAnti, (id#1069 = id#1067)
      :- SubqueryAlias source
      :  +- View (`source`, [id#1067])
      :     +- Project [value#1064 AS id#1067]
      :        +- LocalRelation [value#1064]
      +- RelationV2[id#1069] /var/folders/q3/rsbny1cj6vsc7nkydr30vv2w0000gn/T/TestUpdate8997120534842900007
```

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
CodingCat: sure, updated

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
CodingCat: updated

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteDeleteFromTable.scala
CodingCat: updated

CodingCat:
ping for another review @kbendick, @jackye1995, @nastra, @aokolnychyi, @danielcweeks, @rdblue @singhpk234 @RussellSpitzer thanks!

CodingCat:
ping for another review @kbendick, @jackye1995, @nastra, @aokolnychyi, @danielcweeks, @rdblue @singhpk234 @RussellSpitzer thanks!

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
```diff
@@ -80,91 +82,136 @@ object RewriteMergeIntoTable extends RewriteRowLevelCommand {
...
-          // when there are no MATCHED actions, use a left anti join to remove any matching rows
-          // and switch to using a regular append instead of a row-level merge
-          // only unmatched source rows that match the condition are appended to the table
-          val joinPlan = Join(filteredSource, r, LeftAnti, Some(cond), JoinHint.NONE)
-
-          val outputExprs = insertAction.assignments.map(_.value)
-          val outputColNames = r.output.map(_.name)
-          val outputCols = outputExprs.zip(outputColNames).map { case (expr, name) =>
-            Alias(expr, name)()
+          val icebergTableView = relations.nonEmpty && relations.size == 1
```
hililiwei, line 95: just doubt: Why do we need `relations.size == 1` ? Can there be more than one?

hililiwei: So, is it enough to just use `relations.nonEmpty`?

CodingCat: hmmm...giving it a second thought, I think we still need relations.size == 1 as it is possible to have multiple relations involved here, e.g., we register a view with a join result from two iceberg tables, and tried to merge into it...it is hard to define what's the expected behavior (if any), so we still need relations.size == 1 to limit the scenario to `register the hadoopTable directly as a view`

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
```diff
@@ -80,91 +82,136 @@ object RewriteMergeIntoTable extends RewriteRowLevelCommand {
...
 
       EliminateSubqueryAliases(aliasedTable) match {
         case r: DataSourceV2Relation =>
-          // NOT MATCHED conditions may only refer to columns in source so they can be pushed down
-          val insertAction = notMatchedActions.head.asInstanceOf[InsertAction]
-          val filteredSource = insertAction.condition match {
-            case Some(insertCond) => Filter(insertCond, source)
-            case None => source
+          rewriteMergeIntoTableWithSingleNotMatchedAction(m, r)
+        case v: View =>
```
kbendick, line 91: Under what situations could it be a view outside of Hadoop tables? Can we be a bit more tight on the type here or not?

spark/v3.2/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteUpdateTable.scala
```diff
@@ -56,24 +58,41 @@ object RewriteUpdateTable extends RewriteRowLevelCommand {
...
-          val rewritePlan = operation match {
-            case _: SupportsDelta =>
-              buildWriteDeltaPlan(r, table, assignments, updateCond)
-            case _ if SubqueryExpression.hasSubquery(updateCond) =>
-              buildReplaceDataWithUnionPlan(r, table, assignments, updateCond)
-            case _ =>
-              buildReplaceDataPlan(r, table, assignments, updateCond)
+          rewriteUpdateTable(u, r, tbl)
+        case v: View =>
+          val relations = v.children.collect { case r: DataSourceV2Relation if r.table.isInstanceOf[SparkTable] =>
```
kbendick, line 63: I think there's mAh e  types of table within Spark than just `SparkTable` now but I'm not sure if that matters (cached tables used for some procédures I thubj).

PR #5084, opened 2022-06-20
REST: Set table format version for create table transactions

bryanck:
This PR adds the table format version in the list of metadata updates when starting a create table transaction with the REST catalog client.


kbendick (commented):
This is great. Thank you @bryanck!

Left one comment.

We should add a test as well, ideally / likely within `CatalogTests`, that when a V2 table is created we get a V2 table on load.

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
```diff
@@ -556,6 +556,8 @@ private LoadTableResponse stageCreate() {
   private static List<MetadataUpdate> createChanges(TableMetadata meta) {
     ImmutableList.Builder<MetadataUpdate> changes = ImmutableList.builder();
 
+    changes.add(new MetadataUpdate.UpgradeFormatVersion(meta.formatVersion()));
```
rdblue, line 559: I think this should be set only if the format version is not the default?

bryanck: I think the server-side default and the client-side default may not match? That is the issue I originally encountered, the Trino plugin had v2 as the default and set the initial sequence number to 1, but the server didn't receive a format version so treated it as v1, and failed because the sequence number was > 0.

rdblue: Okay, I see what you mean.

rdblue (approved):

core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java
kbendick: +1. But I too see what you mean.

I could have sworn I left an almost identical comment when I reviewed this before. I can possibly help with adding a unit test for this sequence in `CatalogTests` if we’d like. Feel free to reach out on Slack too.

kbendick (approved):
This looks good to me. Thank you @bryanck!

bryanck:
I added a test for creating a V2 table via a create transaction, which fails without the change.

PR #5089, opened 2022-06-20
Include actual class name in class mismatch exception

findepi:
Include requested and actual class name in the exception message thrown
when `PartitionData` detects type mismatch.


findepi:
Thanks @danielcweeks

PR #5093, opened 2022-06-20
Open-api: RenameTableRequest props are required

Fokko:
Currently, the properties are optional, which doesn't really make sense.

Also, in the Java code it throws an exception:
https://github.com/apache/iceberg/blob/master/core/src/test/java/org/apache/iceberg/rest/requests/TestRenameTableRequest.java#L106-L121

Therefore I would suggest making those required in the spec as well


danielcweeks:
Good catch.  Thanks @Fokko

PR #5094, opened 2022-06-20
Spark: Add Spark 3.2 copy on top of Spark 3.3

singhpk234:
About the change :

PR to address https://github.com/apache/iceberg/pull/5056#issuecomment-1157515292, 

This PR contains commits in following order : 
1.  https://github.com/apache/iceberg/pull/5056
  a) move spark/v3.2 to spark/v3.3
  b) changes to make spark 3.3 work
2. copy spark/v3.3 files to spark/v3.2 (done after 1.a)
3. ci changes to make spark/v3.2 and spark/v3.3 work together

This is based on proposal by @hililiwei : https://lists.apache.org/thread/3qk85db08rvdon15fyfvyf0y4cd3d8sq

issue: 
- https://github.com/apache/iceberg/issues/4713

--- 

cc @rdblue @aokolnychyi @jackye1995 @RussellSpitzer @kbendick @amogh-jahagirdar @rajarshisarkar @flyrain @hililiwei @pan3793


rdblue:
@singhpk234, can you post a gist of the diff between 3.2 and 3.3? I see there's an addition for metadata columns.

rdblue:
Here's a gist with the diff: https://gist.github.com/rdblue/73c575f7ada567c3fb3b997918f3bf9d

spark/v3.3/build.gradle
```diff
@@ -0,0 +1,279 @@
...
+    compileOnly "com.google.errorprone:error_prone_annotations"
+    compileOnly "org.apache.avro:avro"
+    compileOnly("org.apache.spark:spark-hive_${scalaVersion}:${sparkVersion}") {
+      exclude group: 'org.apache.avro', module: 'avro'
+      exclude group: 'org.apache.arrow'
+      exclude group: 'org.apache.parquet'
+      // to make sure io.netty.buffer only comes from project(':iceberg-arrow')
+      exclude group: 'io.netty', module: 'netty-buffer'
+      exclude group: 'org.roaringbitmap'
+      exclude group: 'com.fasterxml'
```
rdblue, line 71: Why was this added?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaOperation.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  public DeltaWriteBuilder newWriteBuilder(LogicalWriteInfo info) {
+    if (lazyWriteBuilder == null) {
+      // don't validate the scan is not null as if the condition evaluates to false,
+      // the optimizer replaces the original scan relation with a local relation
+      lazyWriteBuilder = new SparkPositionDeltaWriteBuilder(
+          spark, table, command, configuredScan,
+          isolationLevel, (ExtendedLogicalWriteInfo) info);
```
rdblue, line 86: Should this check the cast? Or at least add `@SuppressWarnings("unchecked")`?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
```diff
@@ -0,0 +1,651 @@
...
+    } catch (org.apache.iceberg.exceptions.NoSuchTableException e) {
+      throw new NoSuchTableException(ident);
+    }
+  }
+
+  @Override
+  public SparkTable loadTable(Identifier ident, long timestamp) throws NoSuchTableException {
+    try {
+      Pair<Table, Long> icebergTable = load(ident);
+      return new SparkTable(icebergTable.first(), SnapshotUtil.snapshotIdAsOfTime(icebergTable.first(), timestamp),
```
rdblue, line 162: Can you move `snapshotIdAsOfTime` to a separate line? This makes line wrapping weird so it would be better to save it in a local variable.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
```diff
@@ -0,0 +1,651 @@
...
+    } catch (org.apache.iceberg.exceptions.NoSuchTableException e) {
+      throw new NoSuchTableException(ident);
+    }
+  }
+
+  @Override
+  public SparkTable loadTable(Identifier ident, String version) throws NoSuchTableException {
+    try {
+      Pair<Table, Long> icebergTable = load(ident);
+      return new SparkTable(icebergTable.first(), Long.parseLong(version), !cacheEnabled);
```
rdblue, line 152: I think this should not ignore the snapshot ID from `load`. It should check that the one returned from load is null and throw an exception if it is set. Otherwise, this would allow conflicting table clauses, like `table@snapshot_id_1 AS OF VERSION 2`

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
rdblue: I think we also want to follow up with tests for the new load table clauses.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
```diff
@@ -0,0 +1,651 @@
...
+          !cacheEnabled);
+    } catch (org.apache.iceberg.exceptions.NoSuchTableException e) {
+      throw new NoSuchTableException(ident);
+    }
+  }
+
+  @Override
+  public SparkTable createTable(Identifier ident, StructType schema,
+      Transform[] transforms,
+      Map<String, String> properties) throws TableAlreadyExistsException {
```
rdblue, line 172: Can you revert the whitespace changes in this file? It looks like something autoformatted these and removed the alignment. These should be aligned with the start of `Identifier ident, ...`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
```diff
@@ -0,0 +1,333 @@
...
+  public String[][] listNamespaces(String[] namespace) throws NoSuchNamespaceException {
+    return getSessionCatalog().listNamespaces(namespace);
+  }
+
+  @Override
+  public Map<String, String> loadNamespaceMetadata(String[] namespace) throws NoSuchNamespaceException {
+    try {
+      return getSessionCatalog().loadNamespaceMetadata(namespace);
+    } catch (Exception e) {
+      // work around because it's now throwing NoSuchDbException
```
rdblue, line 99: We avoid catching `Exception` because it is too broad. Can you make this catch clause more specific?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
```diff
@@ -0,0 +1,333 @@
...
+  public void invalidateTable(Identifier ident) {
+    // We do not need to check whether the table exists and whether
+    // it is an Iceberg table to reduce remote service requests.
+    icebergCatalog.invalidateTable(ident);
+    getSessionCatalog().invalidateTable(ident);
+  }
+
+  @Override
+  public Table createTable(Identifier ident, StructType schema, Transform[] partitions,
+      Map<String, String> properties)
```
rdblue, line 145: Can you also fix the argument alignment changes in this file?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
```diff
@@ -0,0 +1,333 @@
...
+  @Override
+  public Catalog icebergCatalog() {
+    Preconditions.checkArgument(icebergCatalog instanceof HasIcebergCatalog,
+        "Cannot return underlying Iceberg Catalog, wrapped catalog does not contain an Iceberg Catalog");
+    return ((HasIcebergCatalog) icebergCatalog).icebergCatalog();
+  }
+
+  @Override
+  public Identifier[] listFunctions(String[] namespace) throws NoSuchNamespaceException {
+    throw new NoSuchNamespaceException(namespace);
```
rdblue, line 326: Should this return an empty array instead?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
```diff
@@ -0,0 +1,131 @@
...
+  @Override
+  public DataType struct(Types.StructType struct, List<DataType> fieldResults) {
+    List<Types.NestedField> fields = struct.fields();
+
+    List<StructField> sparkFields = Lists.newArrayListWithExpectedSize(fieldResults.size());
+    for (int i = 0; i < fields.size(); i += 1) {
+      Types.NestedField field = fields.get(i);
+      DataType type = fieldResults.get(i);
+      Metadata metadata = Metadata.empty();
+      if (MetadataColumns.isMetadataColumn(field.name())) {
```
rdblue, line 65: I think this should check the field ID rather than the field name, since we have reserved IDs for metadata columns.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
```diff
@@ -0,0 +1,131 @@
...
+
+  @Override
+  public DataType struct(Types.StructType struct, List<DataType> fieldResults) {
+    List<Types.NestedField> fields = struct.fields();
+
+    List<StructField> sparkFields = Lists.newArrayListWithExpectedSize(fieldResults.size());
+    for (int i = 0; i < fields.size(); i += 1) {
+      Types.NestedField field = fields.get(i);
+      DataType type = fieldResults.get(i);
+      Metadata metadata = Metadata.empty();
```
rdblue, line 64: Can you move this into a `private` method? I think that would be a bit cleaner. That way we don't have a lot of inline logic.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/actions/TestCreateActions.java
```diff
@@ -0,0 +1,898 @@
...
+                                        false,
+                                        Metadata.empty())
+                                }), false), true, Metadata.empty())});
+
+    // even though this list looks like three level list, it is actually a 2-level list where the items are
+    // structs with 1 field.
+    String expectedParquetSchema =
+        "message spark_schema {\n" +
+            "  optional group col1 (LIST) {\n" +
+            "    repeated group bag {\n" +
```
rdblue, line 662: This change isn't correct. This was validating that Spark actually wrote a two-level list. But this is now validating a 3-level list. We'll need to find out why Spark is no longer writing the legacy format. Can you change this back and set the test to `@Ignore` instead?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReader.java
```diff
@@ -0,0 +1,204 @@
...
+            });
+    List<InternalRow> rows = Lists.newArrayList(RandomData.generateSpark(schema, 10, 0L));
+
+    try (FileAppender<InternalRow> writer =
+        new ParquetWriteAdapter<>(
+            new NativeSparkWriterBuilder(outputFile)
+                .set("org.apache.spark.sql.parquet.row.attributes", sparkSchema.json())
+                .set("spark.sql.parquet.writeLegacyFormat", "false")
+                .set("spark.sql.parquet.outputTimestampType", "INT96")
+                .set("spark.sql.parquet.fieldId.write.enabled", "true")
```
rdblue, line 150: Why was this needed?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestDeleteFrom.java
```diff
@@ -0,0 +1,163 @@
...
+        new SimpleRecord(3, "c")
+    );
+    Dataset<Row> df = spark.createDataFrame(records, SimpleRecord.class);
+    df.coalesce(1).writeTo(tableName).append();
+
+    assertEquals("Should have expected rows",
+        ImmutableList.of(row(1L, "a"), row(2L, "b"), row(3L, "c")),
+        sql("SELECT * FROM %s ORDER BY id", tableName));
+
+    sql("DELETE FROM %s WHERE id < 2", tableName);
```
rdblue, line 64: This changed because Spark 3.3 now supports row-level deletes using an eager rewrite strategy (copy-on-write). I think this is fine, but we should probably deconflict the DELETE implementation in Iceberg.

FYI @aokolnychyi.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkSchemaUtil.java
```diff
@@ -0,0 +1,83 @@
...
+    Assert.assertEquals("estimateSize handles overflow", Long.MAX_VALUE, tableSize);
+  }
+
+  @Test
+  public void testEstimateSize() throws IOException {
+    long tableSize = SparkSchemaUtil.estimateSize(SparkSchemaUtil.convert(TEST_SCHEMA), 1);
+    Assert.assertEquals("estimateSize matches with expected approximation", 24, tableSize);
+  }
+
+
```
rdblue, line 67: Nit: unnecessary newline.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
```diff
@@ -0,0 +1,387 @@
...
+        case p =>
+          throw new AnalysisException(s"$p is not an Iceberg table")
+      }
+
+    case m @ MergeIntoIcebergTable(aliasedTable, source, cond, matchedActions, notMatchedActions, None)
+        if m.resolved && m.aligned =>
+
+      EliminateSubqueryAliases(aliasedTable) match {
+        case r @ DataSourceV2Relation(tbl: SupportsRowLevelOperations, _, _, _, _) =>
+          val table = buildOperationTable(tbl, MERGE, CaseInsensitiveStringMap.empty())
```
rdblue, line 153: Why did this change? Are you using utilities from Spark now?

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
```diff
@@ -0,0 +1,387 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.Join
+import org.apache.spark.sql.catalyst.plans.logical.JoinHint
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.plans.logical.MergeAction
+import org.apache.spark.sql.catalyst.plans.logical.MergeIntoIcebergTable
+import org.apache.spark.sql.catalyst.plans.logical.MergeRows
+import org.apache.spark.sql.catalyst.plans.logical.NO_BROADCAST_HASH
+import org.apache.spark.sql.catalyst.plans.logical.NoStatsUnaryNode
+import org.apache.spark.sql.catalyst.plans.logical.Project
+import org.apache.spark.sql.catalyst.plans.logical.ReplaceIcebergData
```
rdblue, line 53: Can we not use the one from Spark?

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
rdblue, line 153: This should use `r.options`

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
rdblue: Nevermind. I see they're slightly different. We'll need to deconflict, but I agree with using the Iceberg one for now.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/expressions/ExtendedV2ExpressionUtils.scala
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  def toCatalyst(expr: V2Expression, query: LogicalPlan): Expression = {
+    expr match {
+      case SortValue(child, direction, nullOrdering) =>
+        val catalystChild = toCatalyst(child, query)
+        SortOrder(catalystChild, toCatalyst(direction), toCatalyst(nullOrdering), Seq.empty)
+      case IdentityTransform(ref) =>
+        resolveRef[NamedExpression](ref, query)
+      case t :Transform if BucketTransform.unapply(t).isDefined =>
```
rdblue, line 69: Why not update to use the matcher? It looks like Spark added an argument?

spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkExtensionsTestBase.java
```diff
@@ -0,0 +1,67 @@
...
+
+    SparkTestBase.spark = SparkSession.builder()
+        .master("local[2]")
+        .config("spark.testing", "true")
+        .config(SQLConf.PARTITION_OVERWRITE_MODE().key(), "dynamic")
+        .config("spark.sql.extensions", IcebergSparkSessionExtensions.class.getName())
+        .config("spark.hadoop." + METASTOREURIS.varname, hiveConf.get(METASTOREURIS.varname))
+        .config("spark.sql.shuffle.partitions", "4")
+        .config("spark.sql.hive.metastorePartitionPruningFallbackOnException", "true")
+        .config("spark.sql.legacy.respectNullabilityInTextDatasetConversion", "true")
```
rdblue, line 59: Why was adding this needed?

rdblue:
@singhpk234, thanks for putting this together! I did a round of review. Hopefully we can get this in soon!

It would be great if @aokolnychyi could take a look as well.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
singhpk234: This was discussed in supporting delete pr in spark as well https://github.com/apache/spark/pull/35395#discussion_r826611125

>These are options passed into newRowLevelOperationBuilder and I thought they should come from the SQL operation. For example, if Spark adds a clause OPTIONS to its SQL for DELETE, UPDATE, MERGE, then these values will be propagated here.

based on the rationale above by @aokolnychyi  I decided to keep them same here as well.

spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkExtensionsTestBase.java
singhpk234: This was added as we in our UT's were providing, schema via json via `DataFrameReader.schema(schema).json(jsonDataset)` which was not respecting the nullability in the schema provided and hence an UT in TestMerge was failing, https://github.com/apache/iceberg/pull/5056#discussion_r899896356

Hence to preserve this behaviour in our helper I turned this on. Do you recommend refactoring our [Test Utils](https://github.com/apache/iceberg/blob/master/spark/v3.2/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkRowLevelOperationsTestBase.java#L194-L198) 

This was a breaking change and spark has documented this here as well : https://spark.apache.org/docs/latest/sql-migration-guide.html#upgrading-from-spark-sql-32-to-33

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
singhpk234: I tried specifying `NoSuchDatabaseException` but since loadNamespaceMetadata doesn't specify throwing `NoSuchDatabaseException` in it's signature my ide started complaining and hence I had to swtich to catching generic exception. Your thoughts ?

I think we must need to do additional checks here for  `catalog.databaseExists(db)` then load db : https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/V2SessionCatalog.scala#L246


I though of filing an issue in upstream but deferred for now since it could be worked around.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/expressions/ExtendedV2ExpressionUtils.scala
singhpk234: BucketTransform extractor now expects the unapply to be of type `Transform` post this [commit](https://github.com/apache/spark/commit/2ed827a8169a0ad098c5ee4f6101bd003f8c6de2). since expr is of type `V2Expression` on specifying the matcher ide complains with no corresponding unapply() found.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReader.java
singhpk234: This was a newly introduced conf in spark 3.3, since we directly pass conf from here to ParquetWriter ->  ParquetWriteSupport -> SparkToParquetSchemaConverter . then this tries to get the value of this [conf](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala#L481) : 
```
  def this(conf: Configuration) = this(
    writeLegacyParquetFormat = conf.get(SQLConf.PARQUET_WRITE_LEGACY_FORMAT.key).toBoolean,
    outputTimestampType = SQLConf.ParquetOutputTimestampType.withName(
      conf.get(SQLConf.PARQUET_OUTPUT_TIMESTAMP_TYPE.key)),
    useFieldId = conf.get(SQLConf.PARQUET_FIELD_ID_WRITE_ENABLED.key).toBoolean)
```

now if this conf is not set in the config we passed it will fail with NPE. trying to call to get boolean value of a null value.

spark/v3.3/build.gradle
singhpk234: in my frok when I was running the UT with the spark 3.3 snapshot, my ut's were failing due to conflicts in jar, was 
getting : https://github.com/singhpk234/iceberg/runs/6792894997?check_suite_focus=true
> org.apache.iceberg.spark.sql.TestTimestampWithoutZone > testCreateNewTableShouldHaveTimestampWithoutZoneIcebergType[catalogName = spark_catalog, implementation = org.apache.iceberg.spark.SparkSessionCatalog, config = {type=hive, default-namespace=default, parquet-enabled=true, cache-enabled=false}] FAILED
    java.lang.NoClassDefFoundError: Could not initialize class org.apache.spark.sql.catalyst.util.RebaseDateTime$

even after forcing the `com.fasterxml` verison to 2.12.3

I removed this with released 3.3 and see ut's which were failing earlier now succeeding , seems like problem was in un-published 3.3 i.e snapshot, removed it from this.

singhpk234:
Many thanks @rdblue, for the feedbacks, have addressed most of them and reverted on the rest.

> It would be great if @aokolnychyi could take a look as well.

+1, it would be awesome :) !!!

Here is a gist post addressing feedbacks : https://gist.github.com/singhpk234/640953c2658a6b16fa94c60c8066c4c0

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
singhpk234: I did some digging and found we can work this around is we override the namespaceExists and delegate it to V2SessionCatalog#namespaceExists, default implementation present in SupportNameSpace uses loadNamespaceMetadata which is causing this issue.

Never the less the behavior of V2SessionCatalog#loadNamespaceMetadata seemed a off as compared to exsiting catalogs in spark for ex. JDBC catalog, have filed a [jira](https://github.com/apache/spark/pull/36948) and a [PR](https://github.com/apache/spark/pull/36948) in spark for same as well.

aokolnychyi:
I'll be able to take a look today. Sorry for the delay!

aokolnychyi:
I am a little bit confused. Shouldn't #5056 be reviewed and merged first?

hililiwei:
Oh,  it seems that #5056 should be merged first so that history can be keeped.

aokolnychyi:
Okay, I'll review this PR for changes regarding 3.3 support and then we can either remove the 3.2 part from this PR or add appropriate changes to #5056.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteOperation.java
```diff
@@ -0,0 +1,100 @@
...
+
+    return lazyWriteBuilder;
+  }
+
+  @Override
+  public NamedReference[] requiredMetadataAttributes() {
+    NamedReference file = Expressions.column(MetadataColumns.FILE_PATH.name());
+    NamedReference pos = Expressions.column(MetadataColumns.ROW_POSITION.name());
+
+    if (command == Command.DELETE || command == Command.UPDATE) {
```
aokolnychyi, line 94: Can we adapt `checkstyle.xml` to include the new class instead of qualifying? This should reduce the amount of changes.

We have the following entry in `checkstyle.xml`:
```
org.apache.spark.sql.connector.iceberg.write.RowLevelOperation.Command.*,
```

I'd simply add the Spark class now.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaOperation.java
```diff
@@ -0,0 +1,108 @@
...
+      };
+    }
+
+    return lazyScanBuilder;
+  }
+
+  @Override
+  public DeltaWriteBuilder newWriteBuilder(LogicalWriteInfo info) {
+    if (lazyWriteBuilder == null) {
+      Preconditions.checkArgument(info instanceof ExtendedLogicalWriteInfo,
```
aokolnychyi, line 83: What about using a shorter message to stay on one line?

```
Preconditions.checkArgument(info instanceof ExtendedLogicalWriteInfo, "info must be ExtendedLogicalWriteInfo");
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java
```diff
@@ -0,0 +1,665 @@
...
+
+        rowDelta.validateDataFilesExist(referencedDataFiles);
+
+        if (scan.snapshotId() != null) {
+          // set the read snapshot ID to check only snapshots that happened after the table was read
+          // otherwise, the validation will go through all snapshots present in the table
+          rowDelta.validateFromSnapshot(scan.snapshotId());
+        }
+
+        if (command == Command.UPDATE || command == Command.MERGE) {
```
aokolnychyi, line 194: If we adapt `checkstyle.xml`, changes like this won't be needed in this class.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkDistributionAndOrderingUtil.java
```diff
@@ -0,0 +1,221 @@
...
+          return Distributions.ordered(ordering);
+        }
+
+      default:
+        throw new IllegalArgumentException("Unexpected distribution mode: " + distributionMode);
+    }
+  }
+
+  public static SortOrder[] buildCopyOnWriteOrdering(Table table, Command command, Distribution distribution) {
+    if (command == Command.DELETE || command == Command.UPDATE) {
```
aokolnychyi, line 119: nit: same in this class

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
```diff
@@ -0,0 +1,333 @@
...
+  @Override
+  public Catalog icebergCatalog() {
+    Preconditions.checkArgument(icebergCatalog instanceof HasIcebergCatalog,
+        "Cannot return underlying Iceberg Catalog, wrapped catalog does not contain an Iceberg Catalog");
+    return ((HasIcebergCatalog) icebergCatalog).icebergCatalog();
+  }
+
+  @Override
+  public Identifier[] listFunctions(String[] namespace) {
+    return new Identifier[0];
```
aokolnychyi, line 326: @rdblue, should we delegate to the Spark catalog in new function-related methods?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestDeleteFrom.java
aokolnychyi: Well, I am not sure this is correct as the built-in implementation is currently sub-optimal and does not have dynamic file filtering. Let me take a closer look.

aokolnychyi (commented):
I did a quick look over the Iceberg code. I'll take a look at the extensions tomorrow.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteMergeIntoTable.scala
aokolnychyi: Using empty options seems correct to me.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestDeleteFrom.java
aokolnychyi: Okay, this test does not enable SQL extensions. As a result, we don't replace Spark commands with Iceberg alternatives. Then the updated behavior makes sense.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteRowLevelDeltaCommand.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.spark.sql.catalyst.expressions.ExtendedV2ExpressionUtils
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.catalyst.util.WriteDeltaProjections
+import org.apache.spark.sql.connector.iceberg.write.SupportsDelta
+import org.apache.spark.sql.connector.write.RowLevelOperation
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
+import org.apache.spark.sql.types.StructType
+
+trait RewriteRowLevelDeltaCommand extends Rule[LogicalPlan] {
```
aokolnychyi, line 35: I think we use this for delta and copy-on-write plans. What about `RewriteRowLevelIcebergCommand`?

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteUpdateTable.scala
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Assigns a rewrite plan for v2 tables that support rewriting data to handle UPDATE statements.
+ *
+ * This rule assumes the commands have been fully resolved and all assignments have been aligned.
+ * That's why it must be run after AlignRowLevelCommandAssignments.
+ *
+ * This rule also must be run in the same batch with DeduplicateRelations in Spark.
+ */
+object RewriteUpdateTable extends RewriteRowLevelDeltaCommand with RewriteRowLevelCommand {
```
aokolnychyi, line 54: Do we actually rely on `RewriteRowLevelCommand`? We may want to maybe reuse some of the methods from it but I'd probably just keep  `RewriteRowLevelIcebergCommand`.

aokolnychyi: We can do the migration in a follow-up PR.

aokolnychyi: Or you can extend `RewriteRowLevelCommand` in `RewriteRowLevelIcebergCommand`.

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/expressions/ExtendedV2ExpressionUtils.scala
```diff
@@ -0,0 +1,103 @@
...
+    expr match {
+      case SortValue(child, direction, nullOrdering) =>
+        val catalystChild = toCatalyst(child, query)
+        SortOrder(catalystChild, toCatalyst(direction), toCatalyst(nullOrdering), Seq.empty)
+      case IdentityTransform(ref) =>
+        resolveRef[NamedExpression](ref, query)
+      case t: Transform if BucketTransform.unapply(t).isDefined =>
+        t match {
+            // sort columns will be empty for bucket.
+          case BucketTransform(numBuckets, cols, _) =>
```
aokolnychyi, line 72: Why can't we use the extractor right away?

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ReplaceRewrittenRowLevelCommand.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.plans.logical.RowLevelCommand
+import org.apache.spark.sql.catalyst.rules.Rule
+
+/**
+ * Replaces operations such as DELETE and MERGE with the corresponding rewrite plans.
+ */
+object ReplaceRewrittenRowLevelCommand extends Rule[LogicalPlan] {
+
```
aokolnychyi, line 30: nit: unnecessary new line?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
```diff
@@ -0,0 +1,136 @@
...
+      case DECIMAL:
+        Types.DecimalType decimal = (Types.DecimalType) primitive;
+        return DecimalType$.MODULE$.apply(decimal.precision(), decimal.scale());
+      default:
+        throw new UnsupportedOperationException(
+            "Cannot convert unknown type to Spark: " + primitive);
+    }
+  }
+
+  private Metadata fieldMetadata(int fieldId) {
```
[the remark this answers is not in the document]
singhpk234: Yes, for the deletes that doesn't uses extension, we require this. 
Basically now when we filter-out attributes from `query` of ReplaceData the query now contains attributes created here and the [filtering now checks if an attribute is MetadataAtrribute or not](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/v2Commands.scala#L221-L225), which without this change will not be filtered. Hence we face an issue like https://github.com/apache/iceberg/pull/5056#discussion_r900040404 without this change.

core/src/main/java/org/apache/iceberg/MetadataColumns.java
```diff
@@ -31,6 +31,8 @@ public class MetadataColumns {
   private MetadataColumns() {
   }
 
+  public static final String METADATA_COL_ATTR_KEY = "__metadata_col";
```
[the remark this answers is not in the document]
singhpk234: agree with you, moved this to `TypeToSparkType`, just made it public as will be needing it in tests.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java
```diff
@@ -0,0 +1,136 @@
...
+  @Override
+  public DataType struct(Types.StructType struct, List<DataType> fieldResults) {
+    List<Types.NestedField> fields = struct.fields();
+
+    List<StructField> sparkFields = Lists.newArrayListWithExpectedSize(fieldResults.size());
+    for (int i = 0; i < fields.size(); i += 1) {
+      Types.NestedField field = fields.get(i);
+      DataType type = fieldResults.get(i);
+      StructField sparkField = StructField.apply(
+          field.name(), type, field.isOptional(), fieldMetadata(field.fieldId()));
```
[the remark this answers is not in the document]
singhpk234: defined a temp var, thanks !

spark/v3.3/spark-extensions/src/main/scala/org/apache/spark/sql/catalyst/analysis/RewriteUpdateTable.scala
singhpk234: nice idea, made the changes, thanks !

singhpk234:
> I am a little bit confused. Shouldn't https://github.com/apache/iceberg/pull/5056 be reviewed and merged first?

It was what I had in mind originally as well : 
1. we would first merge https://github.com/apache/iceberg/pull/5056, which will make master have only 3.3 and not 3.2
2. then we would merge this pr adding back 3.2 and make it work with 3.3.

since we started with this pr, I think post finalizing 3.3 changes will make this pr 3.3 only and move 3.2 to https://github.com/apache/iceberg/pull/5056.

which @aokolnychyi  suggested as well.

singhpk234:
Thanks @aokolnychyi for reviewing this PR, have addressed the feedbacks from round 1 of your review.

Here is the updated gist : https://gist.github.com/singhpk234/640953c2658a6b16fa94c60c8066c4c0

.github/workflows/spark-ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     strategy:
       matrix:
         jvm: [8, 11]
-        spark: ['3.0', '3.1', '3.2']
+        spark: ['3.0', '3.1', '3.3']
```
RussellSpitzer: Should add in 3.3 here and not replace 3.2

dev/stage-binaries.sh
```diff
@@ -20,12 +20,12 @@
 
 SCALA_VERSION=2.12
 FLINK_VERSIONS=1.13,1.14,1.15
-SPARK_VERSIONS=2.4,3.0,3.1,3.2
+SPARK_VERSIONS=2.4,3.0,3.1,3.3
```
RussellSpitzer: Should add 3.3 and not replace 3.2

dev/stage-binaries.sh
```diff
@@ -20,12 +20,12 @@
...
 HIVE_VERSIONS=2,3
 
 ./gradlew -Prelease -DscalaVersion=$SCALA_VERSION -DflinkVersions=$FLINK_VERSIONS -DsparkVersions=$SPARK_VERSIONS -DhiveVersions=$HIVE_VERSIONS publishApachePublicationToMavenRepository
 
 # Also publish Scala 2.13 Artifacts for versions that support it.
 # Flink does not yet support 2.13 (and is largely dropping a user-facing dependency on Scala). Hive doesn't need a Scala specification.
-./gradlew -Prelease -DscalaVersion=2.13 :iceberg-spark:iceberg-spark-3.2_2.13:publishApachePublicationToMavenRepository :iceberg-spark:iceberg-spark-extensions-3.2_2.13:publishApachePublicationToMavenRepository :iceberg-spark:iceberg-spark-runtime-3.2_2.13:publishApachePublicationToMavenRepository
+./gradlew -Prelease -DscalaVersion=2.13 :iceberg-spark:iceberg-spark-3.3_2.13:publishApachePublicationToMavenRepository :iceberg-spark:iceberg-spark-extensions-3.3_2.13:publishApachePublicationToMavenRepository :iceberg-spark:iceberg-spark-runtime-3.3_2.13:publishApachePublicationToMavenRepository
```
RussellSpitzer, line 30: Should add 3.3 publish, this removes 3.2 publish. We probably want both

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
RussellSpitzer: It may sense to move these sorts of changes out into another PR? Try to keep this one to just getting all test passing and the build working

.github/workflows/spark-ci.yml
singhpk234: I believe this is showing up in this [commit](https://github.com/apache/iceberg/pull/5094/commits/38eb20e7d1c7fcde22ec6ed79679b7a21ef12896) diff, In the final file diff including all the commits the 3.3 is added after 3.2 [complete diff](https://github.com/apache/iceberg/pull/5094/files#diff-cab7395ccb3f096361a9a1a6af739dad164d06383eae8742d26e11d7495c93d8). 

I replaced 3.2 to 3.3 in that [commit](https://github.com/apache/iceberg/pull/5094/commits/38eb20e7d1c7fcde22ec6ed79679b7a21ef12896)  as I was thinking this would be merge sequence https://github.com/apache/iceberg/pull/5094#issuecomment-1162859934 . Hence this showed up in this way.

dev/stage-binaries.sh
singhpk234: same as https://github.com/apache/iceberg/pull/5094#discussion_r904705128

dev/stage-binaries.sh
singhpk234: sure, added !

spark/v3.3/build.gradle
```diff
@@ -31,8 +31,7 @@ configure(sparkProjects) {
   configurations {
     all {
       resolutionStrategy {
-        force "com.fasterxml.jackson.module:jackson-module-scala_${scalaVersion}:2.12.3"
-        force 'com.fasterxml.jackson.module:jackson-module-paranamer:2.12.3'
+        force 'com.fasterxml.jackson.*:*:2.12.3'
```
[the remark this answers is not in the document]
singhpk234: Nice catch !, you are right, read more about it here : https://github.com/gradle/gradle/commit/7446af4e28cc7c14f3342315796629fab20beb78 , https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ResolutionStrategy.html. This made me think if this force resolution is required in the first place, turns out it's not, removed that as well.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
singhpk234: > V2SessionCatalog#loadNamespaceMetadata seemed a off as compared to other exisiting catalogs in spark for ex. JDBC catalog

The fix for this has been merged in the upstream, never the less the work around to override `namespaceExists` seems good me, will mark this comment resolved if everyone is onboard.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
singhpk234: +1, can add it in a follow-up pr, if everyone is onboard.

singhpk234:
Thanks @RussellSpitzer  for reviewing this PR, have addressed the feedbacks from round 1 of your review.

Here is the updated gist post addressing feedbacks: https://gist.github.com/singhpk234/640953c2658a6b16fa94c60c8066c4c0

spark/v3.3/spark/src/jmh/java/org/apache/iceberg/spark/source/parquet/IcebergSourceFlatParquetDataFilterBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+ * A benchmark that evaluates the file skipping capabilities in the Spark data source for Iceberg.
+ *
+ * This class uses a dataset with a flat schema, where the records are clustered according to the
+ * column used in the filter predicate.
+ *
+ * The performance is compared to the built-in file source in Spark.
+ *
+ * To run this benchmark for spark-3.2:
+ * <code>
+ *   ./gradlew -DsparkVersions=3.2 :iceberg-spark:iceberg-spark-3.2_2.12:jmh
```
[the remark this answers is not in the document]
singhpk234: Nice catch, made the changes, thanks @flyrain !

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkParquetReader.java
rdblue: Okay. Sounds like a Spark bug, but I think it's reasonable to set it in our tests. Also, since this is writing extra data it is good to ensure that we can consume the files that were produced.

spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/SparkExtensionsTestBase.java
rdblue: There's probably a cleanup, but this seems fine for now.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSessionCatalog.java
rdblue: After this is in, yes.

rdblue (approved):
@singhpk234, I think this is ready to commit. What PR should we commit first? This one?

rdblue:
Actually, I think the idea is to rebase and merge this rather than squash and merge. Unfortunately, rebase and merge doesn't work. @singhpk234, can you rebase and fix the problems and then I can rebase & merge without conflicts tomorrow?

singhpk234:
> @singhpk234, I think this is ready to commit. What PR should we commit first? This one?

I was initially thinking this could be first one to get in having just 3.3 without 3.2 and immediately we would merge a followup pr to addback a 3.2 with making 3.3 and 3.2 ci work together. 

> I think the idea is to rebase and merge this rather than squash and merge

+1, This is a good idea, if we don't squash we will be able to preserve the history. 

> @singhpk234, can you rebase and fix the problems and then

Thanks @rdblue, have rebased the changes, here is the updated gist post rebase : https://gist.github.com/singhpk234/640953c2658a6b16fa94c60c8066c4c0

singhpk234:
> Can you check for any updates to Spark 3.2 that have gone in since this was started so we can keep 3.3 up to date?

confirmed last 3 commits that went to 3.2 (master) is also present in both 3.2 / 3.3
1. revert - https://github.com/apache/iceberg/commit/612fa2d070c0e8d4cbdd608573f413ecf68e9c46
2. https://github.com/apache/iceberg/pull/5134
3. https://github.com/apache/iceberg/pull/5075

I did the following as well : 
1. resetted  this branch to master 
2. re-did mv & copy and applied / squashed review comments in `make spark 3.3 work` 
3. verified the diff of v3.2 with v3.3 against approved diff [gist's](https://gist.github.com/singhpk234/640953c2658a6b16fa94c60c8066c4c0).

singhpk234:
Thank you everyone for your awesome reviews :) !!!

I learned a lot from this.

PR #5096, opened 2022-06-20
Add interface for FileIO prefix operations and implementations 

danielcweeks:
This adds an interface for FileIO implementations to support prefix based operations for listing and deleting.

The primary motivation is to enable supporting maintenance activities (like cleaning path directories or listing table locations) without the need to fall back to Hadoop FileSystem. 

There are some notable behavioral differences between directory based and object based storage systems (e.g. for directory based storage, a the prefix must denote a directory).

### DIFF
--- a/api/src/main/java/org/apache/iceberg/io/FileInfo.java
+++ b/api/src/main/java/org/apache/iceberg/io/FileInfo.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.io;
+
+import java.time.Instant;
+
+public class FileInfo {
+  private final String location;
+  private final long size;
+  private final Instant created;
+
+  public FileInfo(String location, long size, Instant created) {
+    this.location = location;
+    this.size = size;
+    this.created = created;
+  }
+
+  public String getLocation() {
+    return location;
+  }
+
+  public long getSize() {
+    return size;
+  }
+
+  public Instant getCreated() {
+    return created;
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java
+++ b/api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.io;
+
+import java.util.stream.Stream;
+
+/**
+ * This interface is intended as an extension for FileIO implementations
+ * to provide additional prefix based operations that may be useful in
+ * performing supporting operations.
+ */
+public interface SupportsPrefixOperations {
+
+  /**
+   * Return a stream of all files under a prefix.
+   *
+   * Hierarchical file systems (e.g. HDFS) may impose additional restrictions
+   * like the prefix mush fully match a directory whereas key/value object
+   * stores may allow for arbitrary prefixes.
+   *
+   * @param prefix prefix to list
+   * @return stream of file information
+   */
+  Stream<FileInfo> listPrefix(String prefix);
+
+  /**
+   * Delete all files under a prefix.
+   *
+   * Hierarchical file systems (e.g. HDFS) may impose additional restrictions
+   * like the prefix mush fully match a directory whereas key/value object
+   * stores may allow for arbitrary prefixes.
+   *
+   * @param prefix prefix to delete
+   */
+  void deletePrefix(String prefix);
+}
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.iceberg.aws.AwsClientFactories;
 import org.apache.iceberg.aws.AwsClientFactory;
 import org.apache.iceberg.aws.AwsProperties;
@@ -33,9 +34,11 @@ import org.apache.iceberg.common.DynConstructors;
 import org.apache.iceberg.io.BulkDeletionFailureException;
 import org.apache.iceberg.io.CredentialSupplier;
 import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.io.FileInfo;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.io.OutputFile;
 import org.apache.iceberg.io.SupportsBulkOperations;
+import org.apache.iceberg.io.SupportsPrefixOperations;
 import org.apache.iceberg.metrics.MetricsContext;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
@@ -54,11 +57,13 @@ import software.amazon.awssdk.services.s3.model.DeleteObjectsRequest;
 import software.amazon.awssdk.services.s3.model.DeleteObjectsResponse;
 import software.amazon.awssdk.services.s3.model.GetObjectTaggingRequest;
 import software.amazon.awssdk.services.s3.model.GetObjectTaggingResponse;
+import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
 import software.amazon.awssdk.services.s3.model.ObjectIdentifier;
 import software.amazon.awssdk.services.s3.model.PutObjectTaggingRequest;
 import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.awssdk.services.s3.model.Tag;
 import software.amazon.awssdk.services.s3.model.Tagging;
+import software.amazon.awssdk.services.s3.paginators.ListObjectsV2Iterable;
 
 /**
  * FileIO implementation backed by S3.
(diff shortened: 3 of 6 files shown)


api/src/main/java/org/apache/iceberg/io/FileInfo.java
```diff
@@ -0,0 +1,46 @@
...
+  private final long size;
+  private final Instant created;
+
+  public FileInfo(String location, long size, Instant created) {
+    this.location = location;
+    this.size = size;
+    this.created = created;
+  }
+
+  public String getLocation() {
```
rdblue, line 35: Can you remove `get` from these method names? We typically omit them.

api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * This interface is intended as an extension for FileIO implementations
+ * to provide additional prefix based operations that may be useful in
+ * performing supporting operations.
+ */
+public interface SupportsPrefixOperations {
+
+  /**
+   * Return a stream of all files under a prefix.
+   *
```
rdblue, line 33: Javadoc: if you want a paragraph break in the rendered version, you have to add `<p>` to the empty lines.

api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java
```diff
@@ -0,0 +1,53 @@
...
+   * Hierarchical file systems (e.g. HDFS) may impose additional restrictions
+   * like the prefix mush fully match a directory whereas key/value object
+   * stores may allow for arbitrary prefixes.
+   *
+   * @param prefix prefix to list
+   * @return stream of file information
+   */
+  Stream<FileInfo> listPrefix(String prefix);
```
rdblue, line 41: We don't usually expose `Stream` in our APIs. Normally this would be an `Iterable<FileInfo>`. Is there a reason to use a stream instead?

api/src/main/java/org/apache/iceberg/io/FileInfo.java
danielcweeks: Yeah, sorry, those were IDE generated.

api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java
danielcweeks: We probably shouldn't expose `Iterable` here because this should really be "traversable once" and it gets overly complicated to fully support the contract.  Hadoop uses `Iterator`, but has its own implementation as you can see from the adaptor.  If your goal is to simply keep streams out of the API, we should probably fallback to `Iterator`.

api/src/main/java/org/apache/iceberg/io/FileInfo.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.iceberg.io;
+
+import java.time.Instant;
+
+public class FileInfo {
+  private final String location;
+  private final long size;
+  private final Instant created;
+
+  public FileInfo(String location, long size, Instant created) {
```
rdblue, line 29: In the rest of the code, we use a plain long value in milliseconds for timestamps like this rather than using `Instant`. Can you switch to `long` or is there a reason to use `Instant` instead?

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -241,6 +246,52 @@ private List<String> deleteObjectsInBucket(String bucket, Collection<String> obj
     return Lists.newArrayList();
   }
 
+  @Override
+  public Stream<FileInfo> listPrefix(String prefix) {
+    S3URI s3uri = new S3URI(prefix, awsProperties.s3BucketToAccessPointMapping());
+
+    return internalListPrefix(s3uri.bucket(), s3uri.key()).stream()
+        .flatMap(r -> r.contents().stream())
+        .map(o -> new FileInfo(o.key(), o.size(), o.lastModified()));
```
rdblue, line 255: I think this needs to reconstruct a URI rather than returning just the key. Otherwise, it isn't really a "location" as we think of them in Iceberg. For example, when listing `"s3://bucket/prefix"`, this might produce `[FileInfo("prefix/file1.parquet", 10, ...), ...]`

api/src/main/java/org/apache/iceberg/io/FileInfo.java
danielcweeks: No specific need (other than it removes confusion about the unit).

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
danielcweeks: Good catch, the bucket is actually not returned as part of the object, but I think it can be reconstructed from the original path.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -241,6 +246,52 @@ private List<String> deleteObjectsInBucket(String bucket, Collection<String> obj
...
+   * they fail, but will log any individual objects that are not deleted as part
+   * of the bulk operation.
+   *
+   * @param prefix prefix to delete
+   */
+  @Override
+  public void deletePrefix(String prefix) {
+    S3URI s3uri = new S3URI(prefix, awsProperties.s3BucketToAccessPointMapping());
+
+    internalListPrefix(s3uri.bucket(), s3uri.key()).stream().parallel().forEach(listing -> {
```
rdblue, line 272: Rather than using `parallel`, can you use `Tasks` and a thread pool? Tasks is how we prefer to parallelize operations because it gives you much better control over error handling:

```java
Tasks.foreach(internalListPrefix(s3uri.bucket(), s3uri.key()).stream())
    .suppressFailureWhenFinished()
    .executeWith(executorService)
    .retry(5)
    .exponentialBackoff(10, 60_000, 600_000, 2)
    .onlyRetryOn(ThrottledException.class, TooManyRequestsException.class)
    .onFailure((task, exc) -> LOG.warn("Failed to delete S3 key: %s", task.key(), exc))
    .run(o -> {
      ObjectIdentifier obj = ObjectIdentifier.builder().key(o.key()).build();
      ...
    });
```

core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java
```diff
@@ -89,4 +97,62 @@ public Configuration getConf() {
...
     this.hadoopConf = confSerializer.apply(getConf());
   }
+
+  @Override
+  public Stream<FileInfo> listPrefix(String prefix) {
+    Path prefixToList = new Path(prefix);
+    FileSystem fs = Util.getFs(prefixToList, hadoopConf.get());
+
+    try {
+      return Streams.stream(new AdaptingIterator<>(fs.listFiles(prefixToList, true)))
```
rdblue, line 105: Minor: when passing boolean arguments, we prefer to add an inline comment to explain them: `fs.listFiles(prefixToList, true /* recursive *);`

core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java
```diff
@@ -89,4 +97,62 @@ public Configuration getConf() {
...
+    }
+  }
+
+  @Override
+  public void deletePrefix(String prefix) {
+    Path prefixToDelete = new Path(prefix);
+    FileSystem fs = Util.getFs(prefixToDelete, hadoopConf.get());
+
+    try {
+      fs.delete(prefixToDelete, true);
```
rdblue, line 119: Same here: `fs.delete(prefixToDelete, true /* recursive */);`

api/src/main/java/org/apache/iceberg/io/FileInfo.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public String location() {
+    return location;
+  }
+
+  public long size() {
+    return size;
+  }
+
+  public long created() {
```
rdblue, line 41: What about `createdAtMillis`? That way we specify the expected unit (millis) and it is clear that this is a timestamp?

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -241,6 +246,52 @@ private List<String> deleteObjectsInBucket(String bucket, Collection<String> obj
...
+   * given prefix.
+   *
+   * Bulk delete operations are used and no reattempt is made for deletes if
+   * they fail, but will log any individual objects that are not deleted as part
+   * of the bulk operation.
+   *
+   * @param prefix prefix to delete
+   */
+  @Override
+  public void deletePrefix(String prefix) {
```
rdblue, line 269: After looking at the bulk delete code, I think we may want to delegate to it, or at least refactor to share.

Deletes now support tag-based deletion, so this needs to detect whether tags should be used. Also, S3 supports bulk deletes, which would make this go much more quickly.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
danielcweeks: Updated and made sure the original scheme was returned.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
danielcweeks: I'll take a look at consolidating the two delete paths.  It's a little messy because the bulk delete is not streaming and will pull everything into memory, so doesn't work well for large operations.

I'll see if I can find a combination / compromise here.

rdblue: The bulk delete accepts an Iterable, so I think it should work with streaming.

nastra (commented):
mostly LGTM once we switch to `Tasks.foreach`, everything else were just a few nits

api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * This interface is intended as an extension for FileIO implementations
+ * to provide additional prefix based operations that may be useful in
+ * performing supporting operations.
+ */
+public interface SupportsPrefixOperations {
+
+  /**
+   * Return a stream of all files under a prefix.
```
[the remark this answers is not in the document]
danielcweeks: Thanks!

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
danielcweeks: This is now handled by the existing batch delete call.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -241,6 +246,54 @@ private List<String> deleteObjectsInBucket(String bucket, Collection<String> obj
...
+
+      client().deleteObjects(request).errors().forEach((s3Error -> {
+        LOG.warn("Error occurred during delete operation. {}: {}. {}", s3Error.code(), s3Error.message(),
+            s3Error.key());
+      }));
+    });
+  }
+
+  private ListObjectsV2Iterable internalListPrefix(String bucket, String keyPrefix) {
+    ListObjectsV2Request request = ListObjectsV2Request.builder().bucket(bucket).prefix(keyPrefix).build();
```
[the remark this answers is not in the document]
danielcweeks: This was removed.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -241,6 +245,33 @@ private List<String> deleteObjectsInBucket(String bucket, Collection<String> obj
...
 
+  @Override
+  public Iterable<FileInfo> listPrefix(String prefix) {
+    S3URI s3uri = new S3URI(prefix, awsProperties.s3BucketToAccessPointMapping());
+    ListObjectsV2Request request = ListObjectsV2Request.builder().bucket(s3uri.bucket()).prefix(s3uri.key()).build();
+
+    return () -> client().listObjectsV2Paginator(request).stream()
+        .flatMap(r -> r.contents().stream())
+        .map(o -> new FileInfo(
+            String.format("%s://%s/%s", s3uri.scheme(), s3uri.bucket(), o.key()),
```
rdblue, line 256: Minor: The bucket returned by `S3URI` may actually be an access point reference, which could break URI parsing if this were to embed it in a location.

I think the solution is to move the S3 access point mapping out of `S3URI`, since that class should remain simple and report what was parsed and avoid this kind of issue. Let's not fix it here, but we should revisit this.

FYI @jackye1995.

PR #5097, opened 2022-06-20
[WIP] Migrate Delta Lake tables to Iceberg

ericlgoodman:
Work is in progress, but putting this out here for general feedback.

The idea will be to leverage the [Delta Standalone](https://docs.delta.io/latest/delta-standalone.html) library, rather than the DeltaCatalog, to read schema/partitioning/table metadata from the existing Delta Lake table. This simplifies the experience as users won't need to configure both a Delta Lake and Iceberg catalog as well as specify the mapping thereof.

A new action will be created to decouple the code behind this from the existing `migrate` action, which makes lots of assumptions about the incoming/exported catalog and relies on the `SparkSessionCatalog` to retrieve metadata about the state of the source table.


ericlgoodman:
See https://github.com/apache/iceberg/pull/5331

Issue #4312, opened 2022-03-11
Flink: Rewrite all data files( upsert mode, v2 table) lead to taskmanager out of memory


Shane-Yu:
I rewrite small files( whole file) with flink api on un-partitioned table,  got the failed job and error log as follows.

 The environment like this:
      
     

> TM's memory       16G
      numFiles              717
      numRows             176450547
 

As we can see from the source code, it seems that the Eq-Delelte file should be loaded into memory when applyEqDeletes with records.  Is that too many  Eq-Delete files lead to taskmanager OOM?

https://github.com/apache/iceberg/blob/197325f3e31193054ecdd860ac0fbc2af4b64748/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java#L149-L159

- error log:

```
Caused by: java.lang.RuntimeException: java.lang.OutOfMemoryError: Java heap space
	at org.apache.iceberg.flink.source.RowDataRewriter$RewriteMap.map(RowDataRewriter.java:150) ~[iceberg-flink-runtime-1.12-0.13.1.jar:?]
	at org.apache.iceberg.flink.source.RowDataRewriter$RewriteMap.map(RowDataRewriter.java:91) ~[iceberg-flink-runtime-1.12-0.13.1.jar:?]
	at org.apache.flink.streaming.api.operators.StreamMap.processElement(StreamMap.java:41) ~[flink-dist_2.11-1.12.0.jar:1.12.0]
	at org.apache.flink.streaming.runtime.tasks.OneInputStreamTask$StreamTaskNetworkOutput.emitRecord(OneInputStreamTask.java:193) ~[flink-dist_2.11-1.12.0.jar:1.12.0]
	at org.apache.flink.streaming.runtime.io.StreamTaskNetworkInput.processElement(StreamTaskNetworkInput.java:179) ~[flink-dist_2.11-1.12.0.jar:1.12.0]
	at org.apache.flink.streaming.runtime.io.StreamTaskNetworkInput.emitNext(StreamTaskNetworkInput.java:152) ~[flink-dist_2.11-1.12.0.jar:1.12.0]
	at org.apache.flink.streaming.runtime.io.StreamOneInputProcessor.processInput(StreamOneInputProcessor.java:67) ~[flink-dist_2.11-1.12.0.jar:1.12.0]
	at org.apache.flink.streaming.runtime.tasks.StreamTask.processInput(StreamTask.java:372) ~[flink-dist_2.11-1.12.0.jar:1.12.0]
[...]

PR #5100, opened 2022-06-21
Data: delete compaction optimization by bloom filter

Zhangg7723:
**Purpose**
 V2 table support equality deletes for row level delete, delete rows are loaded into memory when rewrite or query job is running, but this will cause OOM with too many delete rows in a real scenario, especially in flink upsert mode. As issue #4312 mentioned, flink rewrite jobs caused out of memory exception which also happen with spark or other engine support v2 table.
image
delete rows in hash set occupy most of the heap memory.

**Goal**
Reduce delete rows loaded in memory, optimize the performance of delete compaction by bloom filter, just for parquet format in this PR, thanks for the pull request of @huaxingao about the parquet bloom filter support #4831, and we are working on orc format.

**How**
image
Before reading the equality delete data, load the bloom filter of current data file, then filter delete rows not in this data file.

**Verification**
We verified the performance improvement by a test case.
Environment:
  Job: spark rewrite job with 300 million data rows and 30 million delete rows
  executor num：2 
  executor memory：2G
  executor core：8

Before optimization:
image
JVM did full gc frequently，this job failed in the end.

After optimization:
image
Obviously, the memory pressure reduced, and the job finished successfully.

### DIFF
--- a/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
+++ b/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
@@ -19,6 +19,8 @@
 
 package org.apache.iceberg.data;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -195,7 +197,13 @@ public abstract class DeleteFilter<T> {
       Predicate<T> isInDeleteSet = record -> deleteSet.contains(projectRow.wrap(asStructLike(record)));
       isInDeleteSets.add(isInDeleteSet);
     }
-
+    try {
+      if (parquetReader != null) {
+        parquetReader.close();
+      }
+    } catch (IOException e) {
+      throw new UncheckedIOException("failed to close parquet file reader!", e);
+    }
     return isInDeleteSets;
   }
 
--- a/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
+++ b/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
@@ -36,11 +36,15 @@ import org.apache.iceberg.data.orc.GenericOrcReader;
 import org.apache.iceberg.data.parquet.GenericParquetReaders;
 import org.apache.iceberg.deletes.Deletes;
 import org.apache.iceberg.deletes.PositionDeleteIndex;
+import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.expressions.Literal;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.orc.ORC;
 import org.apache.iceberg.parquet.Parquet;
+import org.apache.iceberg.parquet.ParquetBloomRowGroupFilter;
+import org.apache.iceberg.parquet.ParquetUtil;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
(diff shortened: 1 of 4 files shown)


Zhangg7723:
@rdblue can you give a review? thank you.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -133,10 +144,19 @@ public CloseableIterable<T> filter(CloseableIterable<T> records) {
...
+    Map<BlockMetaData, BloomFilterReader> parquetBloomFilterReader = Maps.newHashMap();
+    ParquetFileReader parquetReader = null;
+    Predicate<Record> isInBloomFilter = null;
     if (eqDeletes.isEmpty()) {
       return isInDeleteSets;
     }
 
+    // load bloomfilter readers from data file
+    if (filePath.endsWith(".parquet")) {
+      parquetReader = ParquetUtil.openFile(getInputFile(filePath));
```
chenjunjiedada, line 156: Can we use `try-with-resource` here?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -148,6 +168,12 @@ private List<Predicate<T>> applyEqDeletes() {
 
       Schema deleteSchema = TypeUtil.select(requiredSchema, ids);
 
+      if (filePath.endsWith(".parquet") && parquetReader != null) {
```
chenjunjiedada, line 171: I think we can change the ctor parameter from `String` to `DataFile` and then here we can check file.format().

Zhangg7723: yes, you are right, but dataFile is not passed into DeleteFilter as parameter, it was changed to filePath in #4381

chenjunjiedada: I see, how about adding another `ctor` to accept `DataFile` parameter?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -133,10 +144,19 @@ public CloseableIterable<T> filter(CloseableIterable<T> records) {
...
     List<Predicate<T>> isInDeleteSets = Lists.newArrayList();
+    Map<BlockMetaData, BloomFilterReader> parquetBloomFilterReader = Maps.newHashMap();
+    ParquetFileReader parquetReader = null;
+    Predicate<Record> isInBloomFilter = null;
     if (eqDeletes.isEmpty()) {
       return isInDeleteSets;
     }
 
+    // load bloomfilter readers from data file
+    if (filePath.endsWith(".parquet")) {
```
chenjunjiedada, line 155: Do we want to check whether the bloom filter is turned on to avoid reading the footer if it is not?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
Zhangg7723: the change for constructor parameters is only for Trino supporting mor，Trino wrapped a dummy fileScanTask for data file currently, the author wants to remove fileScanTask implemented in Trino, and use the filePath parameter. If we change it to dataFile, compatibility is a problem.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
Zhangg7723: You mean we check the bloom filter by table properties, right? but the bloom filter properties may be updated, the bloom filter in current file is unmatched with table properties.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
Zhangg7723: Maybe it's a big change , I want to keep this reader open during the iteration of delete files, and considering orc format, the delete iteration need to be encapsulated in a new function, any advise for this change?

jiamin13579:
We also encountered the same problem

PR #5105, opened 2022-06-21
API: Access deleted and added delete files in Snapshot

aokolnychyi:
This PR uses the work in #4873 as an opportunity to support accessing delete files in `Snapshot`. That PR added new methods for accessing data files with an explicit `FileIO`, which haven't been released yet. I propose to rename the newly added methods to `addedDataFiles` and `deletedDataFiles` and also add similar methods for delete files. I feel it is important to do before 1.0.0.


api/src/main/java/org/apache/iceberg/Snapshot.java
```diff
@@ -133,50 +133,76 @@ public interface Snapshot extends Serializable {
...
    * <p>
    * The files returned include the following columns: file_path, file_format, partition,
    * record_count, and file_size_in_bytes. Other columns will be null.
    *
    * @param io a {@link FileIO} instance used for reading files from storage
-   * @return all files added to the table in this snapshot.
+   * @return all data files added to the table in this snapshot.
    */
-  Iterable<DataFile> addedFiles(FileIO io);
+  Iterable<DataFile> addedDataFiles(FileIO io);
```
aokolnychyi, line 156: Renamed methods haven't been released yet so we won't break anyone.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -235,55 +239,101 @@ public List<ManifestFile> deleteManifests() {
...
+    Preconditions.checkArgument(fileIO != null, "Cannot cache delete file changes: FileIO is null");
+
+    ImmutableList.Builder<DeleteFile> adds = ImmutableList.builder();
+    ImmutableList.Builder<DeleteFile> deletes = ImmutableList.builder();
+
+    Iterable<ManifestFile> changedManifests = Iterables.filter(deleteManifests(fileIO),
+        manifest -> Objects.equal(manifest.snapshotId(), snapshotId));
+
+    for (ManifestFile manifest : changedManifests) {
+      try (ManifestReader<DeleteFile> reader = ManifestFiles.readDeleteManifest(manifest, fileIO, null)) {
```
aokolnychyi, line 312: If I remember correctly, `ManifestReader` will use the Avro header metadata to parse the schema and spec. While it is generally not reliable as the schema may be old, I think it should be fine as long as we don't do any binding or filtering (like in this case).

aokolnychyi:
cc @kbendick @rdblue @flyrain @karuppayya @szehon-ho @RussellSpitzer

api/src/main/java/org/apache/iceberg/Snapshot.java
```diff
@@ -133,50 +133,76 @@ public interface Snapshot extends Serializable {
...
+  /**
+   * Return all delete files deleted from the table in this snapshot.
+   * <p>
+   * The files returned include the following columns: file_path, file_format, partition,
+   * record_count, and file_size_in_bytes. Other columns will be null.
+   *
+   * @param io a {@link FileIO} instance used for reading files from storage
+   * @return all delete files deleted from the table in this snapshot
+   */
+  default Iterable<DeleteFile> deletedDeleteFiles(FileIO io) {
```
rdblue, line 203: Should we use `removed` instead of `deleted` so we don't have `deleted-delete-...`?

rdblue (approved):
+1. Merge when you're ready, but I had a question on a method name.

api/src/main/java/org/apache/iceberg/Snapshot.java
aokolnychyi: This question came up a few times. I went with `deleted` for consistency cause we use that for data files in this class and other places like `deletedPositionDeleteFilesCount`. We do use `removed` in snapshot summary but that seems to be the only place.

If we decide to use `removed`, I think we should update the naming for data file methods too. They haven't been released yet. I'd say let's either keep things as they are or switch to this:

```
removedDataFiles(FileIO io);
removedDeleteFiles(FileIO io);
```

aokolnychyi: @rdblue, what do you think?

rdblue: I prefer removed.

aokolnychyi: Sounds good, I'll update both.

aokolnychyi:
Thanks, @rdblue!

PR #5106, opened 2022-06-21
Spark 3.2: Expose actual action classes in SparkActions

aokolnychyi:
This PR exposes actual classes in `SparkActions` so that users don't have to do weird explicit casts to access methods that expose or accept `DataFrame`s.


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
```diff
@@ -19,165 +19,21 @@
...
+ * An action that deletes reachable files from a given metadata location.
+ *
+ * @deprecated since 0.14.0, will be removed in 1.0.0;
+ *             use {@link SparkActions} and {@link DeleteReachableFilesSparkAction} instead.
  */
-@SuppressWarnings("UnnecessaryAnonymousClass")
-public class BaseDeleteReachableFilesSparkAction
-    extends BaseSparkAction<DeleteReachableFiles, DeleteReachableFiles.Result> implements DeleteReachableFiles {
+@Deprecated
+public class BaseDeleteReachableFilesSparkAction extends DeleteReachableFilesSparkAction {
```
aokolnychyi, line 31: I initially simply renamed the old actions but this may break folks who did explicit casts in the past as these base action classes were kind of public. That's why I extended the renamed classes for compatibility with a deprecation warning.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -19,346 +19,19 @@
...
- * be deleted, using the same {@link Table#location()} and {@link #olderThan(long)} filtering as above.
- * <p>
- * <em>Note:</em> It is dangerous to call this action with a short retention interval as it might corrupt
- * the state of the table if another operation is writing at the same time.
+ * An action to delete orphan files.
+ *
+ * @deprecated since 0.14.0, will be removed in 1.0.0;
+ *             use {@link SparkActions} and {@link DeleteOrphanFilesSparkAction} instead.
  */
-public class BaseDeleteOrphanFilesSparkAction
```
aokolnychyi, line 93: Instead of exposing actual classes, we could add Spark-specific interfaces and expose them. However, I don't think it is a good idea. Methods we would add to those interfaces usually reflect internal implementation details that are not generic. In addition, we probably don’t want to complicate the actions API by adding another layer of interfaces. Exposing classes seems to be the most straightforward option but I'd prefer to rename the classes in that case.

PR #5107, opened 2022-06-21
Core: Check for valid identifiers in REST catalog

bryanck:
This PR adds checks for valid identifiers before performing certain catalog operations in the REST catalog. Currently, the pathing for most REST requests includes the namespace as part of the path. This adds checks that the namespace is not empty to prevent the REST catalog from building a path with `//`. Using double slashes in the path is considered a security vulnerability by some frameworks so seems best to avoid.


danielcweeks (approved):
Thanks @bryanck

PR #5108, opened 2022-06-21
Spark 3.2: Expose actual action classes in SparkActions

aokolnychyi:
This PR exposes actual classes in `SparkActions` so that users don't have to do weird explicit casts to access methods that expose or accept `DataFrame`s.


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteOrphanFilesSparkAction.java
```diff
@@ -19,346 +19,52 @@
...
-    }
-  };
-
-  private String location = null;
-  private long olderThanTimestamp = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(3);
-  private Dataset<Row> compareToFileList;
-  private Consumer<String> deleteFunc = defaultDelete;
-  private ExecutorService deleteExecutorService = null;
+@Deprecated
+public class BaseDeleteOrphanFilesSparkAction extends DeleteOrphanFilesSparkAction {
```
aokolnychyi, line 36: I initially simply renamed the old actions. It was a much smaller change but could break folks who did explicit casts in the past as these base action classes were kind of public. That's why I extended the renamed classes for compatibility with a deprecation warning.

aokolnychyi, line 36: Instead of exposing actual classes, we could add Spark-specific interfaces and expose them. However, I don't think it is a good idea. Methods we would add to those interfaces usually reflect internal implementation details that are not generic. In addition, we probably don’t want to complicate the actions API by adding another layer of interfaces. Exposing classes seems to be the most straightforward option but I'd prefer to do a rename in that case.

rdblue:
@aokolnychyi, should this go into the 0.14.0 release?

aokolnychyi:
Closing in favor of #5257.

PR #4712, opened 2022-05-06
Flink 1.15: Migrate back to use ReusableArrayData

yittg:
Migrate back to use ReusableArrayData for Flink 1.15

part of #4624

This change is already covered by the existing test: `TestFlinkScan#testCustomizedFlinkDataTypes`, for detail:
https://github.com/apache/iceberg/blob/f79da868f48c3364762eae25ee13f31d6f620eb7/flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/source/TestFlinkScan.java#L310-L321

PR #5109, opened 2022-06-21
Flink: FLIP-27 Iceberg source and builder that puts everything together

stevenzwu:


flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -157,7 +157,7 @@ public Long snapshotId() {
     return snapshotId;
   }
 
-  public StreamingStartingStrategy startingStrategy() {
+  public StreamingStartingStrategy streamingStartingStrategy() {
```
stevenzwu, line 160: ScanContext is an internal class. this change shouldn't break any users.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/data/FlinkParquetReaders.java
```diff
@@ -482,8 +484,10 @@ protected void addElement(ReusableArrayData reused, E element) {
 
     @Override
     protected ArrayData buildList(ReusableArrayData list) {
-      list.setNumElements(writePos);
-      return list;
+      // Since ReusableArrayData is not accepted by Flink, use GenericArrayData temporarily to walk around it.
```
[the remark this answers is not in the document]
stevenzwu: Just realized the `ReusableArrayData` is actually from Iceberg code `FlinkParquetReaders`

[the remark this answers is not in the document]
yittg: I don't really get it, i think the fix in Flink is included in 1.15.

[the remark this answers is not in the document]
stevenzwu: FLIP-27 source reader uses this util method to clone the RowData as it batches records for thread handover. It depends on `ArrayDataSerializer` to clone the field. With the change from PR #4712, `ReusableArrayData` object is not cloned and reused, which corrupts the batched RowData array.

```
  public static RowData clone(RowData from, RowData reuse, RowType rowType, TypeSerializer[] fieldSerializers) {
    GenericRowData ret;
    if (reuse instanceof GenericRowData) {
      ret = (GenericRowData) reuse;
    } else {
      ret = new GenericRowData(from.getArity());
    }
    ret.setRowKind(from.getRowKind());
    for (int i = 0; i < rowType.getFieldCount(); i++) {
      if (!from.isNullAt(i)) {
        RowData.FieldGetter getter = RowData.createFieldGetter(rowType.getTypeAt(i), i);
        ret.setField(i, fieldSerializers[i].copy(getter.getFieldOrNull(from)));
      } else {
        ret.setField(i, null);
      }
    }
    return ret;
  }
```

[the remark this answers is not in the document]
yittg: @stevenzwu Do you have any failed test case? Let me have a look or reproduce it locally. I'm still confused why it still can not work.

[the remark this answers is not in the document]
stevenzwu: @yittg you can check out the dev branch for this PR and run the `TestIcebergSourceBounded`. `testCustomizedFlinkDataTypes` method should fail, because the array field has the same value for all records. 

For easier read of the diff, you can change the record count from 10 to 2.
```
List<Record> records = RandomGenericData.generate(schema, 10, 0L);
```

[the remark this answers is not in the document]
yittg: Thanks @stevenzwu,
I finally got the point, the ArrayDataSerializer in Flink should be renew each time because it reuse the BinaryArrayData  internally. 'Think we can change the signature of RowDataUtil#clone to accept a supplier of serializer to walk around it for now.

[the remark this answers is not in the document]
yittg: I looked into the Flink for why map data works but array data not with similar implementation, found that the MapDataSerializer#toBinaryMap always contains copy semantics implicitly but not for ArrayDataSerializer#toBinaryArray.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSource.java
```diff
@@ -0,0 +1,289 @@
...
+    this.readerFunction = readerFunction;
+    this.assignerFactory = assignerFactory;
+  }
+
+  private static Table loadTable(TableLoader tableLoader) {
+    tableLoader.open();
+    try (TableLoader loader = tableLoader) {
+      return loader.loadTable();
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to close table loader", e);
```
rdblue, line 82: Should this be `UncheckedIOException` instead of generic `RuntimeException`?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSource.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  @Override
+  public Boundedness getBoundedness() {
+    return scanContext.isStreaming() ? Boundedness.BOUNDED : Boundedness.CONTINUOUS_UNBOUNDED;
```
rdblue, line 88: This is backwards. Good catch, @stevenzwu!

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSource.java
```diff
@@ -0,0 +1,289 @@
...
+      this.contextBuilder.planParallelism(planParallelism);
+      return this;
+    }
+
+    public Builder properties(Map<String, String> properties) {
+      contextBuilder.fromProperties(properties);
+      return this;
+    }
+
+    public IcebergSource<T> build() {
```
rdblue, line 274: You could use `public <T> IcebergSource<T> build()` here to allow a bit easier type customization.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSource.java
```diff
@@ -0,0 +1,289 @@
...
+      // For now, <table name>-<random UUID> is used as the unique thread pool name.
+      ContinuousSplitPlanner splitPlanner = new ContinuousSplitPlannerImpl(
+          table, scanContext, table.name() + "-" + UUID.randomUUID());
+      return new ContinuousIcebergEnumerator(enumContext, assigner, scanContext, splitPlanner, enumState);
+    } else {
+      return new StaticIcebergEnumerator(enumContext, assigner, table, scanContext, enumState);
+    }
+  }
+
+  public static <T> Builder<T> builder() {
```
rdblue, line 147: You could use `builderForRowData` here also.

stevenzwu: When I tried to implement the RowData factory method, I found it kind of need to duplicate 3 of the ScanContext configs for the RowDataReaderFunction.

```
  public static Builder<RowData> builderForRowData(Configuration readConfig, Table table, ScanContext context) {
   ReaderFunction<RowData> readerFunction = new RowDataReaderFunction(readConfig, table.schema(), context.project(),
        context.nameMapping(), context.caseSensitive(), table.io(), table.encryption());
    return new Builder<RowData>()
        .readerFunction(readerFunction);
  }
```

stevenzwu: Personally, I actually like to make `ScanContext` public and exposed to users. Then it is ok to have method like `Builder<RowData> builderForRowData(Configuration readConfig, Table table, ScanContext context)`. We can also avoid duplicate more than a dozen of methods from `ScanContext `to `IcebergSource$Builder`. Users can construct ScanContext once.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSource.java
```diff
@@ -0,0 +1,290 @@
...
+      this.splitAssignerFactory = assignerFactory;
+      return this;
+    }
+
+    public Builder<T> readerFunction(ReaderFunction<T> newReaderFunction) {
+      this.readerFunction = newReaderFunction;
+      return this;
+    }
+
+    public Builder caseSensitive(boolean newCaseSensitive) {
```
stevenzwu, line 180: @rdblue I like to revisit the discussion of exposing `ScanContext` directly to users (instead of replicating the `ScanContext` methods in the source builder here). Not only we reduced code duplication. We also avoid potential out-of-sync problem. In the past, I have seen the case where we added new methods to `ScanContext` but forgot to add them to source builder.

stevenzwu: @yittg @kbendick what's your take?

rdblue: I think this is making the API more complicated for the convenience of developers, so I don't think that we should expose it.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/IcebergSource.java
rdblue: We can probably do better here. Why not initialize `readerFunction` in the `build` method if it is `null`?

rdblue (approved):

rdblue:
Thanks, @stevenzwu!

I know we still need to discuss the builder that makes reading as RowData easy, but I went ahead and merged this because we can add that later.

stevenzwu:
@klam-shop @zoucao This is the PR for source builder that puts everything together. You can try the MVP version out from the master branch. Also your feedbacks on the source builder/construction are welcomed.

zoucao:
> 

Thanks for your great work, @stevenzwu, and we will try it as soon as possible.

klam-shop:
Thank you @stevenzwu !

PR #5113, opened 2022-06-22
Fix NPE in BoundPredicate.toString for IS_NULL/NOT_NULL on an optional struct column

huaxingao:
If we have a Predicate `IS_NULL/NOT_NULL` on an optional Struct field, `BoundPredicate.toString` throws NPE.
For example, the following code throws NPE.
```
    UnboundPredicate<?> unbound1 = new UnboundPredicate<>(IS_NULL, ref("struct"));
    Expression expr1 = unbound1.bind(struct);
    BoundPredicate<?> bound1 = assertAndUnwrap(expr1);
```
This PR fixes the NPE.

I also feel that we probably should not allow predicate on a top column of nested column, except `IS_NULL/NOT_NULL`, so I add a check to throw Exception.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
@@ -109,6 +109,10 @@ public class UnboundPredicate<T> extends Predicate<T, UnboundTerm<T>> implements
       return bindUnaryOperation(bound);
     }
 
+    if (bound.type().isNestedType()) {
+      throw new UnsupportedOperationException("Non-Unary Predicate is not supported for top column of nested types");
+    }
+
     if (op() == Operation.IN || op() == Operation.NOT_IN) {
       return bindInOperation(bound);
     }
--- a/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
@@ -109,9 +109,8 @@ public class UnboundPredicate<T> extends Predicate<T, UnboundTerm<T>> implements
       return bindUnaryOperation(bound);
     }
 
-    if (bound.type().isNestedType()) {
-      throw new UnsupportedOperationException("Non-Unary Predicate is not supported for top column of nested types");
-    }
+    Preconditions.checkState(!bound.type().isNestedType(),
+        "Non-Unary Predicate is not supported for top column of nested types");
 
     if (op() == Operation.IN || op() == Operation.NOT_IN) {
       return bindInOperation(bound);
--- a/api/src/test/java/org/apache/iceberg/expressions/TestPredicateBinding.java
+++ b/api/src/test/java/org/apache/iceberg/expressions/TestPredicateBinding.java
@@ -619,7 +619,7 @@ public class TestPredicateBinding {
       lt.bind(struct);
       Assert.fail("Binding should fail");
     } catch (UnsupportedOperationException e) {
-      Assert.assertTrue("Validation should complain about Non-Unary predicate is not supported for top column of nested types",
+      Assert.assertTrue("Validation should complain about Non-Unary predicate is not supported for nested types",
           e.getMessage().contains("Non-Unary Predicate is not supported for top column of nested types"));
     }
 
@@ -701,7 +701,7 @@ public class TestPredicateBinding {
       lt.bind(struct);
       Assert.fail("Binding should fail");
     } catch (UnsupportedOperationException e) {
-      Assert.assertTrue("Validation should complain about Non-Unary predicate is not supported for top column of nested types",
+      Assert.assertTrue("Validation should complain about Non-Unary predicate is not supported for nested types",
           e.getMessage().contains("Non-Unary Predicate is not supported for top column of nested types"));
     }
 
@@ -722,7 +722,7 @@ public class TestPredicateBinding {
     Assert.assertTrue("Should be a unary predicate", bound2.isUnaryPredicate());
 
     UnboundPredicate<?> unbound3 = new UnboundPredicate<>(IS_NAN, ref("list"));
-     try {
+    try {
       unbound3.bind(struct);
       Assert.fail("Binding should fail");
     } catch (ValidationException e) {
@@ -751,7 +751,7 @@ public class TestPredicateBinding {
       lt.bind(struct);
       Assert.fail("Binding should fail");
     } catch (UnsupportedOperationException e) {
-      Assert.assertTrue("Validation should complain about Non-Unary predicate is not supported for top column of nested types",
+      Assert.assertTrue("Validation should complain about Non-Unary predicate is not supported for nested types",
           e.getMessage().contains("Non-Unary Predicate is not supported for top column of nested types"));
     }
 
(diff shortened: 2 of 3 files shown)


api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
```diff
@@ -109,6 +109,10 @@ public Expression bind(StructType struct, boolean caseSensitive) {
       return bindUnaryOperation(bound);
     }
 
+    if (bound.type().isNestedType()) {
```
szehon-ho, line 112: Wondering if this is necessary.  Today will hit ValidationException when it tries to make a Literal?  

If necessary, we can use "Preconditions".

api/src/test/java/org/apache/iceberg/expressions/TestPredicateBinding.java
```diff
@@ -605,4 +605,188 @@ public void testNotInPredicateBindingConversionToExpression() {
...
+  public void testStructFields() {
+    StructType structFieldType =
+        Types.StructType.of(Types.NestedField.optional(11, "float_field", Types.FloatType.get()));
+    StructType struct = StructType.of(
+        optional(10, "struct", structFieldType)
+    );
+
+    UnboundPredicate<Float> lt = new UnboundPredicate<>(LT, ref("struct"), 1.234f);
+    try {
+      lt.bind(struct);
```
szehon-ho, line 619: For this and other instance of validation exception, we can use "AssertHelpers.assertThrowsCause"

szehon-ho (commented):
I think this is a good find.  (accessor is null for nested types)

szehon-ho:
By the way, was playing around and was wondering, are comparison for struct literal expected to be supported in Spark/Iceberg?  

Today in Spark SQL:

select * from table where table.struct_field = struct(10)
> org.apache.spark.sql.AnalysisException: cannot resolve '(table.struct_field = struct(10))' due to data type mismatch: differing types in '(table.struct_field = struct(1))' (struct<nested:int> and struct<col1:int>).; line 1 pos 39;


select * from table where table.struct_field in (struct(10))
```
java.lang.IllegalArgumentException: Cannot create expression literal from org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema: [1]
  at org.apache.iceberg.expressions.Literals.from(Literals.java:87)
  at org.apache.iceberg.expressions.UnboundPredicate.<init>(UnboundPredicate.java:40)
  at org.apache.iceberg.expressions.Expressions.equal(Expressions.java:175)
  at org.apache.iceberg.spark.SparkFilters.handleEqual(SparkFilters.java:239)
  at org.apache.iceberg.spark.SparkFilters.convert(SparkFilters.java:152)
  at org.apache.iceberg.spark.source.SparkScanBuilder.pushFilters(SparkScanBuilder.java:106)
  at org.apache.spark.sql.execution.datasources.v2.PushDownUtils$.pushFilters(PushDownUtils.scala:69)
  at org.apache.spark.sql.execution.datasources.v2.V2ScanRelationPushDown$$anonfun$pushDownFilters$1.applyOrElse(V2ScanRelationPushDown.scala:60)
  at org.apache.spark.sql.execution.datasources.v2.V2ScanRelationPushDown$$anonfun$pushDownFilters$1.applyOrElse(V2ScanRelationPushDown.scala:47)
```

api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
huaxingao: If the predicate value is a complex type, for example
```
    Float[] value = {1.23f, 2.34f};
    UnboundPredicate<Float[]> lt = new UnboundPredicate<>(LT, ref("list"), value);
```
The Exception will be thrown at `Literals.from(value)` at this [line](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java#L40)

If the predicate value is a primitive, for example

```
    UnboundPredicate lt = new UnboundPredicate<>(LT, ref("list"), 1.23f);
```
It will throw Exception at this [line](https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java#L116). So this `if (bound.type().isNestedType())` might be needed when a primitive value is accidentally used for Predicate on a nested type. 

I changed this to `Preconditions`.

api/src/test/java/org/apache/iceberg/expressions/TestPredicateBinding.java
huaxingao: Changed. Thanks!

huaxingao:
> are comparison for struct literal expected to be supported in Spark?

It's not supported in Spark.

huaxingao:
@szehon-ho Thank you very much for your review! I have addressed the comments. Could you please take one more look when you have a chance?

api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
```diff
@@ -109,6 +109,9 @@ public Expression bind(StructType struct, boolean caseSensitive) {
       return bindUnaryOperation(bound);
     }
 
+    Preconditions.checkState(!bound.type().isNestedType(),
+        "Non-Unary Predicate is not supported for top column of nested types");
```
szehon-ho, line 113: Sentence is a bit complex, maybe 'Non-unary predicate is not supported for complex types'?  Or even 'Only unary predicates are supported for complex types'.

api/src/test/java/org/apache/iceberg/expressions/TestPredicateBinding.java
```diff
@@ -605,4 +605,161 @@ public void testNotInPredicateBindingConversionToExpression() {
     Expression expr = unbound.bind(struct);
     Assert.assertEquals("Should change NOT_IN to alwaysTrue expression", Expressions.alwaysTrue(), expr);
   }
+
+  @Test
+  public void testStructFields() {
```
szehon-ho, line 610: Sorry to make more work, but do you think we can split these two big test methods ?  I feel there's a lot of independent tests here.

szehon-ho (commented):
Thanks, some more comments

api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java
szehon-ho: OK , makes sense to be more up front in what we support.

szehon-ho:
> It's not supported in Spark.

Do you know if that is documented somewhere?  Actually I tried on Spark non-Iceberg table and it seems to support it.  Now, I am just wondering whether we should somehow for now skip pushing these filters into Iceberg in the first place so we don't block these kind of queries from being evaluated on Spark.

```
spark.sql("select * from test_struct_non_iceberg where struct_field in(struct(10))").show
+------------+
|struct_field|
+------------+
|        {10}|
+------------+


scala> spark.sql("select * from test_struct_non_iceberg where struct_field = struct(10)").show
+------------+
|struct_field|
+------------+
|        {10}|
+------------+
```

Note: it's all orthogonal to this PR.

huaxingao:
Sorry, I was wrong, Spark actually supports nested types comparison. I didn't construct the struct literal correctly when I tried the struct comparison earlier. I checked again just now and it works OK for me.

I need to rethink about this PR. I will close this for now and reopen when I am ready. Thanks a lot for helping me on this! @szehon-ho

szehon-ho:
No problem.  Actually I do think this change is still valid, as it's Iceberg API and it's better to protect the caller, if you want to proceed.

I made another issue about what we are talking about: https://github.com/apache/iceberg/issues/5132, maybe we can fix it in the filter pushdown part by not pushing down the complex predicates.

szehon-ho:
Note: for this particular NPE, it may change once this is in: https://github.com/apache/iceberg/pull/5067

PR #5118, opened 2022-06-22
CORE - Load Table Response Test

kbendick:
This PR adds serde tests for `LoadTableResponse`.

It also removes `metadata-location` as a required field for `LoadTableResponse`, as `LoadTableResponse` is returned in response to a `CreateTableRequest` for staging a table for creation without committing it during a transaction. There is not yet necessarily a metadata location as the table has not been fully created.

To avoid making a number of things public, I'm reading in existing `TableMetadata` JSON files and testing `LoadTableResponse` with those (as `LoadTableResponse` is what's being tested, not `TableMetadata`).

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
+++ b/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
@@ -28,6 +28,7 @@ import java.util.stream.Collectors;
 import org.apache.iceberg.BaseMetadataTable;
 import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.BaseTransaction;
+import org.apache.iceberg.HasTableOperations;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SortOrder;
@@ -170,16 +171,19 @@ public class CatalogHandlers {
     properties.putAll(request.properties());
 
     String location;
+    String metadataLocation = null;
     if (request.location() != null) {
       location = request.location();
     } else {
-      location = catalog.buildTable(ident, request.schema())
+      Table table = catalog.buildTable(ident, request.schema())
           .withPartitionSpec(request.spec())
           .withSortOrder(request.writeOrder())
           .withProperties(properties)
           .createTransaction()
-          .table()
-          .location();
+          .table();
+
+      location = table.location();
+      metadataLocation = metadataLocationFor(table);
     }
 
     TableMetadata metadata = TableMetadata.newTableMetadata(
@@ -188,6 +192,12 @@ public class CatalogHandlers {
         request.writeOrder() != null ? request.writeOrder() : SortOrder.unsorted(),
         location,
         properties);
+
+    if (metadataLocation != null) {
+      metadata = TableMetadata.buildFrom(metadata)
+          .withMetadataLocation(metadataLocation)
+          .build();
+    }
 
     return LoadTableResponse.builder()
         .withTableMetadata(metadata)
@@ -335,5 +345,9 @@ public class CatalogHandlers {
     }
 
     return ops.current();
+  }
+
+  private static String metadataLocationFor(Table table) {
+    return ((HasTableOperations) table).operations().current().metadataFileLocation();
   }
 }
--- a/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
+++ b/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
@@ -28,7 +28,6 @@ import java.util.stream.Collectors;
 import org.apache.iceberg.BaseMetadataTable;
 import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.BaseTransaction;
-import org.apache.iceberg.HasTableOperations;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SortOrder;
@@ -171,19 +170,16 @@ public class CatalogHandlers {
     properties.putAll(request.properties());
 
     String location;
-    String metadataLocation = null;
     if (request.location() != null) {
       location = request.location();
     } else {
-      Table table = catalog.buildTable(ident, request.schema())
+      location = catalog.buildTable(ident, request.schema())
           .withPartitionSpec(request.spec())
           .withSortOrder(request.writeOrder())
           .withProperties(properties)
           .createTransaction()
-          .table();
-
-      location = table.location();
-      metadataLocation = metadataLocationFor(table);
+          .table()
+          .location();
     }
 
     TableMetadata metadata = TableMetadata.newTableMetadata(
@@ -192,12 +188,6 @@ public class CatalogHandlers {
         request.writeOrder() != null ? request.writeOrder() : SortOrder.unsorted(),
         location,
         properties);
-
-    if (metadataLocation != null) {
-      metadata = TableMetadata.buildFrom(metadata)
-          .withMetadataLocation(metadataLocation)
-          .build();
-    }
 
     return LoadTableResponse.builder()
         .withTableMetadata(metadata)
@@ -345,9 +335,5 @@ public class CatalogHandlers {
     }
 
     return ops.current();
-  }
-
-  private static String metadataLocationFor(Table table) {
-    return ((HasTableOperations) table).operations().current().metadataFileLocation();
   }
 }
--- a/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
+++ b/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
@@ -53,6 +53,7 @@ public class LoadTableResponse implements RESTResponse {
 
   @Override
   public void validate() {
+    Preconditions.checkNotNull(metadata, "Invalid metadata: null");
   }
 
   public String metadataLocation() {
--- a/core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
+++ b/core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
@@ -104,39 +104,42 @@ public class TestLoadTableResponse extends RequestResponseTestBase<LoadTableResp
   }
 
   @Test
-  public void testRoundTripSerde() throws Exception {
+  public void testRoundTripSerdeWithV1TableMetadata() throws Exception {
     // Default fields are missing in this JSON
-    String tableMetadataV1Json = readTableMetadataInputFile("TableMetadataV1Valid.json");
-    TableMetadata metadataV1 = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, tableMetadataV1Json);
+    String tableMetadataJson = readTableMetadataInputFile("TableMetadataV1Valid.json");
+    TableMetadata v1Metadata = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, tableMetadataJson);
     // Convert the TableMetadata JSON from the file to an object and then back to JSON so that missing fields
     // are filled in with their default values.
-    String json1 = String.format(
+    String json = String.format(
         "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
-        TEST_METADATA_LOCATION, TableMetadataParser.toJson(metadataV1));
-    LoadTableResponse resp1 = LoadTableResponse.builder()
-        .withTableMetadata(metadataV1)
+        TEST_METADATA_LOCATION, TableMetadataParser.toJson(v1Metadata));
+    LoadTableResponse resp = LoadTableResponse.builder()
+        .withTableMetadata(v1Metadata)
         .addAllConfig(CONFIG)
         .build();
-    assertRoundTripSerializesEquallyFrom(json1, resp1);
+    assertRoundTripSerializesEquallyFrom(json, resp);
+  }
 
-    String tableMetadataV2Json = readTableMetadataInputFile("TableMetadataV2Valid.json");
-    TableMetadata metadataV2 = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, tableMetadataV2Json);
+  @Test
+  public void testRoundTripSerdeWithV2TableMetadata() throws Exception {
+    String tableMetadataJson = readTableMetadataInputFile("TableMetadataV2Valid.json");
+    TableMetadata v2Metadata = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, tableMetadataJson);
     // Convert the TableMetadata JSON from the file to an object and then back to JSON so that missing fields
     // are filled in with their default values.
-    String json2 = String.format(
+    String json = String.format(
         "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
-        TEST_METADATA_LOCATION, TableMetadataParser.toJson(metadataV2));
-    LoadTableResponse resp2 = LoadTableResponse.builder()
-        .withTableMetadata(metadataV2)
+        TEST_METADATA_LOCATION, TableMetadataParser.toJson(v2Metadata));
+    LoadTableResponse resp = LoadTableResponse.builder()
+        .withTableMetadata(v2Metadata)
         .addAllConfig(CONFIG)
         .build();
-    assertRoundTripSerializesEquallyFrom(json2, resp2);
+    assertRoundTripSerializesEquallyFrom(json, resp);
   }
 
   @Test
   public void testCanDeserializeWithoutDefaultValues() throws Exception {
     String metadataJson = readTableMetadataInputFile("TableMetadataV1Valid.json");
-    // `config` is missing in the json
+    // `config` is missing in the JSON
     String json = String.format("{\"metadata-location\":\"%s\",\"metadata\":%s}", TEST_METADATA_LOCATION, metadataJson);
     TableMetadata metadata = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, metadataJson);
     LoadTableResponse actual = deserialize(json);
--- a/core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
+++ b/core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
@@ -1,0 +1,218 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest.responses;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import org.apache.iceberg.AssertHelpers;
+import org.apache.iceberg.NullOrder;
+import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SortOrder;
+import org.apache.iceberg.TableMetadata;
+import org.apache.iceberg.TableMetadataParser;
+import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.rest.RequestResponseTestBase;
+import org.apache.iceberg.types.Types;
+import org.junit.Assert;
+import org.junit.Test;
+
+import static org.apache.iceberg.TestHelpers.assertSameSchemaList;
+
+public class TestLoadTableResponse extends RequestResponseTestBase<LoadTableResponse> {
+
+  private static final String TEST_METADATA_LOCATION = "s3://bucket/test/location/metadata/v1.metadata.json";
+
+  private static final String TEST_TABLE_LOCATION = "s3://bucket/test/location";
+
+  private static final Schema SCHEMA_7 = new Schema(7,
+      Types.NestedField.required(1, "x", Types.LongType.get()),
+      Types.NestedField.required(2, "y", Types.LongType.get(), "comment"),
+      Types.NestedField.required(3, "z", Types.LongType.get())
+  );
+
+  private static final PartitionSpec SPEC_5 = PartitionSpec.builderFor(SCHEMA_7).withSpecId(5).build();
+
+  private static final SortOrder SORT_ORDER_3 = SortOrder.builderFor(SCHEMA_7)
+      .withOrderId(3)
+      .asc("y", NullOrder.NULLS_FIRST)
+      .desc(Expressions.bucket("z", 4), NullOrder.NULLS_LAST)
+      .build();
+
+  private static final Map<String, String> TABLE_PROPS = ImmutableMap.of(
+      "format-version", "1",
+      "owner", "hank");
+
+  private static final Map<String, String> CONFIG = ImmutableMap.of("foo", "bar");
+
+  @Override
+  public String[] allFieldsFromSpec() {
+    return new String[] { "metadata-location", "metadata", "config" };
+  }
+
+  @Override
+  public LoadTableResponse createExampleInstance() {
+    TableMetadata metadata =
+        TableMetadata
+            .buildFrom(
+                TableMetadata.newTableMetadata(SCHEMA_7, SPEC_5, SORT_ORDER_3, TEST_TABLE_LOCATION, TABLE_PROPS))
+            .withMetadataLocation(TEST_METADATA_LOCATION)
+            .build();
+
+    return LoadTableResponse.builder()
+        .withTableMetadata(metadata)
+        .addAllConfig(CONFIG)
+        .build();
+  }
+
+  @Override
+  public LoadTableResponse deserialize(String json) throws JsonProcessingException {
+    LoadTableResponse resp = mapper().readValue(json, LoadTableResponse.class);
+    resp.validate();
+    return resp;
+  }
+
+  @Test
+  public void testFailures() {
+    AssertHelpers.assertThrows(
+        "Table metadata should be required",
+        NullPointerException.class,
+        "Invalid metadata: null",
+        () -> LoadTableResponse.builder().build());
+  }
+
+  @Test
+  public void testRoundTripSerde() throws Exception {
+    // Default fields are missing in this JSON
+    String tableMetadataV1Json = readTableMetadataInputFile("TableMetadataV1Valid.json");
+    TableMetadata metadataV1 = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, tableMetadataV1Json);
+    // Convert the TableMetadata JSON from the file to an object and then back to JSON so that missing fields
+    // are filled in with their default values.
+    String json1 = String.format(
+        "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
+        TEST_METADATA_LOCATION, TableMetadataParser.toJson(metadataV1));
+    LoadTableResponse resp1 = LoadTableResponse.builder()
+        .withTableMetadata(metadataV1)
+        .addAllConfig(CONFIG)
+        .build();
+    assertRoundTripSerializesEquallyFrom(json1, resp1);
+
+    String tableMetadataV2Json = readTableMetadataInputFile("TableMetadataV2Valid.json");
+    TableMetadata metadataV2 = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, tableMetadataV2Json);
+    // Convert the TableMetadata JSON from the file to an object and then back to JSON so that missing fields
+    // are filled in with their default values.
+    String json2 = String.format(
+        "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
+        TEST_METADATA_LOCATION, TableMetadataParser.toJson(metadataV2));
+    LoadTableResponse resp2 = LoadTableResponse.builder()
+        .withTableMetadata(metadataV2)
+        .addAllConfig(CONFIG)
+        .build();
+    assertRoundTripSerializesEquallyFrom(json2, resp2);
+  }
+
+  @Test
+  public void testCanDeserializeWithoutDefaultValues() throws Exception {
+    String metadataJson = readTableMetadataInputFile("TableMetadataV1Valid.json");
+    String json1 = String.format(
+        "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
+        TEST_METADATA_LOCATION, metadataJson);
+
+    TableMetadata metadataV1 = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, metadataJson);
+    LoadTableResponse resp1 = LoadTableResponse.builder()
+        .withTableMetadata(metadataV1)
+        .addAllConfig(CONFIG)
+        .build();
+    assertEquals(deserialize(json1), resp1);
+  }
+
+  @Override
+  public void assertEquals(LoadTableResponse actual, LoadTableResponse expected) {
+    Assert.assertEquals("Should have the same configuration", expected.config(), actual.config());
+    assertEqualTableMetadata(actual.tableMetadata(), expected.tableMetadata());
+    Assert.assertEquals("Should have the same metadata location",
+        expected.metadataLocation(), actual.metadataLocation());
+  }
+
+  private void assertEqualTableMetadata(TableMetadata actual, TableMetadata expected) {
+    Assert.assertEquals("Format version should match",
+        expected.formatVersion(), actual.formatVersion());
+    Assert.assertEquals("Table UUID should match",
+        expected.uuid(), actual.uuid());
+    Assert.assertEquals("Table location should match",
+        expected.location(), actual.location());
+    Assert.assertEquals("Last column id", expected.lastColumnId(), actual.lastColumnId());
+    Assert.assertEquals("Schema should match", expected.schema().asStruct(), actual.schema().asStruct());
+    assertSameSchemaList(expected.schemas(), actual.schemas());
+    Assert.assertEquals("Current schema id should match", expected.currentSchemaId(), actual.currentSchemaId());
+    Assert.assertEquals("Schema should match", expected.schema().asStruct(), actual.schema().asStruct());
+    Assert.assertEquals("Last sequence number should match",
+        expected.lastSequenceNumber(), actual.lastSequenceNumber());
+    Assert.assertEquals("Partition spec should match",
+        expected.spec().toString(), actual.spec().toString());
+    Assert.assertEquals("Default spec ID should match",
+        expected.defaultSpecId(), actual.defaultSpecId());
+    Assert.assertEquals("PartitionSpec map should match",
+        expected.specs(), actual.specs());
+    Assert.assertEquals("Default Sort ID should match",
+        expected.defaultSortOrderId(), actual.defaultSortOrderId());
+    Assert.assertEquals("Sort order should match",
+        expected.sortOrder(), actual.sortOrder());
+    Assert.assertEquals("Sort order map should match",
+        expected.sortOrders(), actual.sortOrders());
+    Assert.assertEquals("Properties should match",
+        expected.properties(), actual.properties());
+    Assert.assertEquals("Snapshots should match",
+        Lists.transform(expected.snapshots(), Snapshot::snapshotId),
+        Lists.transform(actual.snapshots(), Snapshot::snapshotId));
+    Assert.assertEquals("History should match", expected.snapshotLog(), actual.snapshotLog());
+    Snapshot expectedCurrentSnapshot = expected.currentSnapshot();
+    Snapshot actualCurrentSnapshot = actual.currentSnapshot();
+    Assert.assertTrue("Both expected and actual current snapshot should either be null or non-null",
+        (expectedCurrentSnapshot != null && actualCurrentSnapshot != null) ||
+            (expectedCurrentSnapshot == null && actualCurrentSnapshot == null));
+    if (expectedCurrentSnapshot != null) {
+      Assert.assertEquals("Current snapshot ID should match",
+          expected.currentSnapshot().snapshotId(), actual.currentSnapshot().snapshotId());
+      Assert.assertEquals("Parent snapshot ID should match",
+          expected.currentSnapshot().parentId(), actual.currentSnapshot().parentId());
+      Assert.assertEquals("Schema ID for current snapshot should match",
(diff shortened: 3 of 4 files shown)


core/src/test/java/org/apache/iceberg/LocalTableOperations.java
```diff
@@ -27,13 +27,13 @@
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.junit.rules.TemporaryFolder;
 
-class LocalTableOperations implements TableOperations {
+public class LocalTableOperations implements TableOperations {
```
kbendick, line 30: This is in test so I don't personally have concerns making it public, but others might feel differently.

core/src/main/java/org/apache/iceberg/TableMetadataParser.java
```diff
@@ -83,30 +83,30 @@ private TableMetadataParser() {
...
+  public static final String SORT_ORDERS = "sort-orders";
+  public static final String PROPERTIES = "properties";
+  public static final String CURRENT_SNAPSHOT_ID = "current-snapshot-id";
+  public static final String REFS = "refs";
+  public static final String SNAPSHOTS = "snapshots";
+  public static final String SNAPSHOT_ID = "snapshot-id";
+  public static final String TIMESTAMP_MS = "timestamp-ms";
+  public static final String SNAPSHOT_LOG = "snapshot-log";
+  public static final String METADATA_FILE = "metadata-file";
+  public static final String METADATA_LOG = "metadata-log";
```
kbendick, line 109: I don't have concern making these public personally as they are generally speaking usable by engines if needeed, but open to other opinions.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -133,7 +133,8 @@ public static class SnapshotLogEntry implements HistoryEntry {
     private final long timestampMillis;
     private final long snapshotId;
 
-    SnapshotLogEntry(long timestampMillis, long snapshotId) {
+    // VisibleForTesting
+    public SnapshotLogEntry(long timestampMillis, long snapshotId) {
```
kbendick, line 137: This is one I could make public via reflection, as the class itself is `public static`.

api/src/main/java/org/apache/iceberg/PartitionSpec.java
```diff
@@ -102,7 +102,8 @@ public boolean isUnpartitioned() {
     return !isPartitioned();
   }
 
-  int lastAssignedFieldId() {
+  // visible for testing
+  public int lastAssignedFieldId() {
```
kbendick, line 106: This is in the api module, so we might choose simply to not test this instead (or access the data by other means).

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
```diff
@@ -32,7 +32,8 @@
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 
-class BaseSnapshot implements Snapshot {
+// visible for testing
+public class BaseSnapshot implements Snapshot {
```
kbendick, line 36: I don't love these changes, but they are in `core`.

kbendick (commented):
Left some notes on what can be done to reduce the scope of things made public or my thoughts on each.

We can also use shims in `test` code etc to access some values (like `lastAssignedFieldId()`, but it's just for `TestLoadTableResult` so that seems like overkill.

cc @rdblue

kbendick:
Not sure why the Flink CI suite was cancelled after many hours, so I’m going to close and reopen to have CI run again.

kbendick:
Since we’re only testing `LoadTableResponse` here and not `TableMetadata` specifically, it might be better to write the JSON out as external files and use `TableMetadataParser` after reading them in.

We’d still be building the `LoadTableResponse` via the builder and testing that, but we wouldn’t need to create the `TableMetadata` instances programmaticallly.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+  @Override
+  public void assertEquals(LoadTableResponse actual, LoadTableResponse expected) {
+    // For printing JSON
+    // try {
+    //   Assert.assertEquals("Should have the same string representation",
+    //       mapper().writeValueAsString(expected), mapper().writeValueAsString(actual));
+    // } catch (JsonProcessingException e) {
+    //   LOG.error("\tRAUL Received exceptiom when comparing actual = {}, expected = {}", actual, expected, e);
+    //   LOG.error("\tRAUL exception is: ", e);
+    // }
```
rdblue, line 305: Can you remove print statements or comments used for debugging?

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+  private static final Map<String, String> CONFIG = ImmutableMap.of("foo", "bar");
+
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  public TableOperations ops = new LocalTableOperations(temp);
+
+  @Override
+  public String[] allFieldsFromSpec() {
+    return new String[] { "metadataLocation", "metadata", "config" };
```
rdblue, line 95: This doesn't match the spec. The spec uses `metadata-location`, not `metadataLocation`. It looks like this needs to be fixed in the serialization.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+      .asc("y", NullOrder.NULLS_FIRST)
+      .desc(Expressions.bucket("z", 4), NullOrder.NULLS_LAST)
+      .build();
+
+  private static final long SEQ_NO = 34;
+  private static final int LAST_ASSIGNED_COLUMN_ID = 3;
+
+  private static final long PREVIOUS_SNAPSHOT_ID = System.currentTimeMillis() - new Random(1234).nextInt(3600);
+
+  private static final String TEST_UUIID = "01073077-d2fd-4132-b86c-09d5107e4747";
```
rdblue, line 80: Typo: `UUIID` should be `UUID`.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+
+  @Override
+  public LoadTableResponse deserialize(String json) throws JsonProcessingException {
+    LoadTableResponse resp = mapper().readValue(json, LoadTableResponse.class);
+    resp.validate();
+    return resp;
+  }
+
+  @Test
+  public void testFailures() {
```
rdblue, line 121: Metadata location is also required according to the spec.

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
rdblue: Do you mean that these are in `org.apache.iceberg` and the test is in `org.apache.iceberg.rest.responses`? All of the classes are in the core module.

I don't think there is a need for this change. Because we delegate to the `TableMetadataParser`, we just need to make sure that metadata makes it through correctly. We don't need to make a complicated `TableMetadata` instance. Instead, rely on `TableMetadataParser` to be correct. The metadata in should match the metadata coming out, but we don't want to do a ton of work to simply exercise `TableMetadataParser`.

api/src/main/java/org/apache/iceberg/PartitionSpec.java
rdblue: I don't think that we need this change. This isn't something that we need to validate because it is derived from the fields for each `PartitionSpec`.

core/src/main/java/org/apache/iceberg/TableMetadata.java
rdblue: Let's just omit this from the validation.

core/src/main/java/org/apache/iceberg/TableMetadataParser.java
rdblue: I think we can revert this. No need to test `TableMetadataParser`, just `LoadTableResponse`.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+        "  \"config\":{\"foo\":\"bar\"}}";
+
+    assertRoundTripSerializesEquallyFrom(fullExampleInstanceJson.replaceAll("\\s", ""), respV2);
+
+  }
+
+
+  private TableMetadata buildV2TableMetadataWithTagsAndSnapshotHistory() {
+    Snapshot previousSnapshot = new BaseSnapshot(
+        ops.io(), PREVIOUS_SNAPSHOT_ID, null, PREVIOUS_SNAPSHOT_ID, null, null, null, ImmutableList.of(
```
rdblue, line 261: I thought Snapshots no longer needed to have IO passed in?

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+
+  @Test
+  public void testRoundTripSerDe() throws JsonProcessingException {
+
+    // Build table metadata so that its uuid and last updated millis can be reused.
+    TableMetadata metaV1 = TableMetadata.buildFrom(
+        TableMetadata.newTableMetadata(
+            TEST_SCHEMA, SPEC_5, SORT_ORDER_3, TEST_TABLE_LOCATION, TABLE_PROPS))
+        .withMetadataLocation(TEST_METADATA_LOCATION)
+        .build();
```
rdblue, line 137: I don't think that we need to check v1 or v2. Just one round-trip test should be fine.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+        "      \"snapshot-id\":" + currentSnapshot.snapshotId() +
+        "  }]," +
+        "  \"metadata-log\":[]}," +
+        "  \"config\":{\"foo\":\"bar\"}}";
+
+    assertRoundTripSerializesEquallyFrom(fullExampleInstanceJson.replaceAll("\\s", ""), respV2);
+
+  }
+
+
```
rdblue, line 258: Style: only one empty line is needed between methods.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+        Types.NestedField.required(10, "x", Types.StringType.get()));
+
+    Map<String, SnapshotRef> refs = ImmutableMap.of(
+        "main", SnapshotRef.branchBuilder(currentSnapshotId).build(),
+        "previous", SnapshotRef.tagBuilder(PREVIOUS_SNAPSHOT_ID).build(),
+        "test", SnapshotRef.branchBuilder(PREVIOUS_SNAPSHOT_ID).build()
+    );
+
+    // TODO - Use builder to get here instead - possibly even just writing out to a .json file and reading in from
+    //  there.
```
rdblue, line 283: I agree with this. You can use `TableMetadata.buildFrom` and `TableMetadata.newTableMetadata` instead of making the constructor public.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,371 @@
...
+    TableMetadata metaV2 = buildV2TableMetadataWithTagsAndSnapshotHistory();
+    LoadTableResponse respV2 = LoadTableResponse.builder()
+        .withTableMetadata(metaV2)
+        .addAllConfig(CONFIG)
+        .build();
+    Snapshot currentSnapshot = metaV2.snapshots().get(1);
+    Snapshot previousSnapshot = metaV2.snapshots().get(0);
+    String fullExampleInstanceJson = "{" +
+        "  \"metadataLocation\":\"s3://bucket/test/location/metadata/v1.metadata.json\"," +
+        "  \"metadata\":{" +
```
rdblue, line 193: Could you just use `TableMetadataParser` and build a string around it? There are similar tests in `TestCreateTableRequest` and that works great. I don't think it's very valuable to have complicated inline JSON like this.

core/src/main/java/org/apache/iceberg/TableMetadataParser.java
rdblue: Also, I just checked and it doesn't look like `SNAPSHOT_ID` is used in the tests. The string is embedded instead, so it should be even easier to make this change.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: Removed

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: It's not quite required by the builder though. I'll update the builder as well 👍

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: They aren't. My bad. I'll remove them.

api/src/main/java/org/apache/iceberg/PartitionSpec.java
kbendick: Reverted this and am no longer testing it.

core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java
```diff
@@ -39,6 +40,7 @@ static ObjectMapper mapper() {
         if (!isInitialized) {
           MAPPER.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
           MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+          MAPPER.setPropertyNamingStrategy(new PropertyNamingStrategy.KebabCaseStrategy());
```
kbendick, line 43: This will come from https://github.com/apache/iceberg/pull/5135

core/src/main/java/org/apache/iceberg/BaseSnapshot.java
kbendick: I did mean that they are all in `core`. As opposed to having this be in `api` where we have tests and where stability is guaranteed.

I agree with your statement though, and I will update the tests to read in a JSON file with a `TableMetadata` instance and then rely on `TableMetadataParser`, as we're testing `LoadTableResponse`.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: Updating the builder to require it is causing a number of failures in `TestRESTCatalog`.

I'll uncomment it and push and collect the stacktrace.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: This has been updated

kbendick:
There's currently one failure `TestCreateTableRequest#testReoundTripSerDe`, that will be fixed once either https://github.com/apache/iceberg/pull/5135 or https://github.com/apache/iceberg/pull/5133 is merged.

```
TestCreateTableRequest > testRoundTripSerDe FAILED
    org.junit.ComparisonFailure: The serialized version of the Java record should match the original JSON expected:<...wner":"Hank"},"stage[-c]reate":false}> but was:<...wner":"Hank"},"stage[C]reate":false}>
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.apache.iceberg.rest.RequestResponseTestBase.assertRoundTripSerializesEquallyFrom(RequestResponseTestBase.java:90)
        at org.apache.iceberg.rest.requests.TestCreateTableRequest.testRoundTripSerDe(TestCreateTableRequest.java:83)
```

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: Updated.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: Updated to use `TableMetadataParser` as well as read in from existing external JSON files.

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
```diff
@@ -69,7 +69,7 @@ public void testRoundTripSerDe() throws JsonProcessingException {
         "\"fields\":[{\"name\":\"id_bucket\",\"transform\":\"bucket[16]\",\"source-id\":1,\"field-id\":1000}]}," +
         "\"order\":{\"order-id\":1,\"fields\":" +
         "[{\"transform\":\"identity\",\"source-id\":2,\"direction\":\"asc\",\"null-order\":\"nulls-last\"}]}," +
-        "\"properties\":{\"owner\":\"Hank\"},\"stageCreate\":false}";
+        "\"properties\":{\"owner\":\"Hank\"},\"stage-create\":false}";
```
kbendick, line 72: There's currently one failure `TestCreateTableRequest#testReoundTripSerDe`, that will be fixed once either https://github.com/apache/iceberg/pull/5135 or https://github.com/apache/iceberg/pull/5133 is merged.

```
TestCreateTableRequest > testRoundTripSerDe FAILED
    org.junit.ComparisonFailure: The serialized version of the Java record should match the original JSON expected:<...wner":"Hank"},"stage[-c]reate":false}> but was:<...wner":"Hank"},"stage[C]reate":false}>
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.apache.iceberg.rest.RequestResponseTestBase.assertRoundTripSerializesEquallyFrom(RequestResponseTestBase.java:90)
        at org.apache.iceberg.rest.requests.TestCreateTableRequest.testRoundTripSerDe(TestCreateTableRequest.java:83)
```

I've updated this for now so that this test failure won't cause issues and then will rebase when one of those PRs are merged.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -225,6 +225,16 @@ public static LoadTableResponse loadTable(Catalog catalog, TableIdentifier ident
...
+      // java.lang.AssertionError: Table should have correct number of previous metadata locations expected:<1> but was
+      //   :<2>
+      // TableMetadata metadataWithoutLocation = ((BaseTable) table).operations().current();
+      // String metadataLocation = ((BaseTable) table).operations().metadataFileLocation("metadataV1.json");
+      // TableMetadata metadata = TableMetadata.buildFrom(metadataWithoutLocation)
+      //     .withMetadataLocation(metadataLocation)
+      //     .build();
+      // String metadataLocation = ((BaseTable) table).operations().current().previousFiles();
       return LoadTableResponse.builder()
           .withTableMetadata(((BaseTable) table).operations().current())
```
kbendick, line 239: This `TableMetadata` has no metadataLocation value on it. Tracking down where this is built and will update it.

kbendick: It actually does. There's some internal usage of `LoadTableResponse` where the `metadataLocation` isn't known (e.g. during `stageCreate`).

Removing the requirement that the result of the builder have `metadataLocation` at all times fixes this.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -336,4 +348,8 @@ private static TableMetadata commit(TableOperations ops, UpdateTableRequest requ
 
     return ops.current();
   }
+
+  private static String metadataLocation(Table tbl) {
+    return ((HasTableOperations) tbl).operations().current().metadataFileLocation();
+  }
```
kbendick, line 354: Currently unused - was using to check the metadata file location in some of the failing tests. It's only intermediate results that have a `LoadTableResponse` without a metadataLocation (e.g. if `stageCreate` is being called).

So maybe `metadata-location` should be nullable as `LoadTableResponse` can be returned without it? Or we need to be adding it somewhere in the responses that we're not inside of `RESTSessionCatalog`.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -53,6 +53,8 @@ private LoadTableResponse(String metadataLocation, TableMetadata metadata, Map<S
 
   @Override
   public void validate() {
+    Preconditions.checkNotNull(metadata, "Invalid metadata: null");
+    Preconditions.checkNotNull(metadataLocation, "Invalid metadata location: null");
```
kbendick, line 57: Since `validate` is for ensuring that all requirements are met from results received externally, it makes sense to have the `metadata-location` check here.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -106,6 +113,8 @@ public Builder addAllConfig(Map<String, String> properties) {
 
     public LoadTableResponse build() {
       Preconditions.checkNotNull(metadata, "Invalid metadata: null");
+      // This is causing errors in TestRESTCatalog via CatalogHandlers
+      // Preconditions.checkNotNull(metadataLocation, "Invalid metadata location: null");
```
kbendick, line 117: There's a call to the builder somewhere that's used during the `CatalogTests` that is creating a `LoadTableResponse` without a metadata location.

I think it's somewhere related to here https://github.com/apache/iceberg/blob/e0ff3c6e9876ce04c890908e375c876d14240c3e/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java#L160-L195, or some other create transaction that doesn't commit right away. Will investigate further.

kbendick: It's this `LoadTableResponse` https://github.com/apache/iceberg/blob/e0ff3c6e9876ce04c890908e375c876d14240c3e/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L483-L484, which is called in the failing tests via `catalog.buildTable(TABLE, OTHER_SCHEMA).createTransaction()` which then fails the builder because there's no metadata location yet.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -189,8 +194,12 @@ public static LoadTableResponse stageTableCreate(Catalog catalog, Namespace name
         location,
         properties);
 
+    TableMetadata metadataWithLocation = TableMetadata.buildFrom(metadata)
+        .withMetadataLocation(metadataLocation)
+        .build();
+
     return LoadTableResponse.builder()
-        .withTableMetadata(metadata)
+        .withTableMetadata(metadataWithLocation)
```
kbendick, line 202: This change causes the number of tests in `TestRESTCatalog` that fail if we require `metadataLocation` be non-null via the builder to go from 16 to 13.

If we remove that requirement from the builder, then all tests pass. Given that we use `LoadTableResponse` in various places, I think we should drop the requirement that `metadataLocation` be present at all times and allow for it to be `null`.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -106,6 +113,8 @@ public Builder addAllConfig(Map<String, String> properties) {
 
     public LoadTableResponse build() {
       Preconditions.checkNotNull(metadata, "Invalid metadata: null");
+      // This is causing errors in TestRESTCatalog via CatalogHandlers
+      Preconditions.checkNotNull(metadataLocation, "Invalid metadata location: null");
```
kbendick, line 117: This is the check that is causing 16 test failures in `TestRESTCatalog` - which has been reduced to 13 by the changes in `CatalogHandlers`.

Without this check, all tests in `TestRESTCatalog` pass. See notes here too

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -94,6 +96,11 @@ public Builder withTableMetadata(TableMetadata tableMetadata) {
       return this;
     }
 
+    public Builder withMetadataLocation(String metaLocation) {
+      this.metadataLocation = metaLocation;
+      return this;
+    }
```
kbendick, line 102: This isn't used presently, but I've found myself having to rebuild the `TableMetadata` in a few places to get `metadataLocation` to be non-null. 

Given that the accessor in `LoadTableResponse` _also_ calls the builder and adds `metadataLocation` to it, this could be useful.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
kbendick: Nevermind. We've determined that `metadata-location` is not a required field. See the PR description for explanation.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
kbendick: We've determined that `metadata-location` is not required (can be null), due to this very case.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -170,16 +171,20 @@ public static LoadTableResponse stageTableCreate(Catalog catalog, Namespace name
     properties.putAll(request.properties());
 
     String location;
+    String metadataLocation;
     if (request.location() != null) {
       location = request.location();
+      metadataLocation = location + "/metadata";
     } else {
-      location = catalog.buildTable(ident, request.schema())
+      Table tbl = catalog.buildTable(ident, request.schema())
```
rdblue, line 179: In Iceberg, we don't avoid abbreviations like `tbl`. It's no easier to type and is harder to read. Use `table` instead.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -170,16 +171,20 @@ public static LoadTableResponse stageTableCreate(Catalog catalog, Namespace name
     properties.putAll(request.properties());
 
     String location;
+    String metadataLocation;
     if (request.location() != null) {
       location = request.location();
+      metadataLocation = location + "/metadata";
```
rdblue, line 177: Why does this hard-code `/metadata`?

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -89,8 +90,16 @@ private Builder() {
...
-      this.metadataLocation = tableMetadata.metadataFileLocation();
       this.metadata = tableMetadata;
+      if (tableMetadata.metadataFileLocation() != null) {
+        this.metadataLocation = tableMetadata.metadataFileLocation();
+      }
+
+      return this;
+    }
+
+    public Builder withMetadataLocation(String metaLocation) {
```
rdblue, line 101: Why is changing this builder necessary?

core/src/test/java/org/apache/iceberg/LocalTableOperations.java
```diff
@@ -27,13 +27,13 @@
...
   private final TemporaryFolder temp;
   private final FileIO io;
 
   private final Map<String, String> createdMetadataFilePaths = Maps.newHashMap();
 
-  LocalTableOperations(TemporaryFolder temp) {
+  public LocalTableOperations(TemporaryFolder temp) {
```
rdblue, line 36: Is this still needed?

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
kbendick: Is there a particular constant you’d like me to use or to skip adding metadata location for this section where only `location` is known?

open-api/rest-catalog-open-api.yaml
```diff
@@ -1464,19 +1464,23 @@ components:
...
       description:
         Result used when a table is successfully loaded.
-        
-        
-        The table metadata JSON is returned in the `metadata` field. The corresponding file location of table metadata must be returned in the `metadata-location` field. Clients can check whether metadata has changed by comparing metadata locations.
-        
-        
+
+
+        The table metadata JSON is returned in the `metadata` field. The corresponding file location of table metadata should be returned in the `metadata-location` field, unless it hasn't been determined yet - such as when a transaction begins to stage a table for creation but has not commit.
```
rdblue, line 1469: Instead of adding ` - ` and extending the sentence, start a new one with "For example".

It should be something like "unless the metadata is not yet committed. For example, a create transaction may return metadata that is staged but not committed."

open-api/rest-catalog-open-api.yaml
```diff
@@ -1464,19 +1464,23 @@ components:
...
+
         The `config` map returns table-specific configuration for the table's resources, including its HTTP client and FileIO. For example, config may contain a specific FileIO implementation class for the table depending on its underlying storage.
       type: object
       required:
-        - metadata-location
         - metadata
       properties:
         metadata-location:
           type: string
+          nullable: true
```
rdblue, line 1482: I don't think that we use `nullable: true`. I thought we only used `required`.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
rdblue: Yeah, please remove any unnecessary changes.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
kbendick: I'll just skip location in this case as it doesn't particularly matter as we're not making it a required field.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
kbendick: Changed to table.

open-api/rest-catalog-open-api.yaml
```diff
@@ -1464,19 +1464,23 @@ components:
...
+          description: Current location of the table's metadata file. Should only be null if the table has been staged for creation as part of a transaction but does not yet exist.
```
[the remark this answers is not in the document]
kbendick: image

Looks better with the shorter description given that the whole result has a thorough description.

core/src/test/java/org/apache/iceberg/LocalTableOperations.java
kbendick: Nope. Reverting now.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  @Test
+  public void testCanDeserializeWithoutDefaultValues() throws Exception {
+    String metadataJson = readTableMetadataInputFile("TableMetadataV1Valid.json");
+    String json1 = String.format(
+        "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
+        TEST_METADATA_LOCATION, metadataJson);
+
+    TableMetadata metadataV1 = TableMetadataParser.fromJson(ops.io(), metadataJson);
```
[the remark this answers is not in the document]
kbendick: Changed to null and removed `ops` from the class entirely.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,246 @@
...
+
+  @Override
+  public LoadTableResponse deserialize(String json) throws JsonProcessingException {
+    LoadTableResponse resp = mapper().readValue(json, LoadTableResponse.class);
+    resp.validate();
+    return resp;
+  }
+
+  @Override
+  public void testHasOnlyKnownFields() {
```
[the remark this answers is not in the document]
kbendick: Rebased and updated.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -89,8 +90,11 @@ private Builder() {
     }
 
     public Builder withTableMetadata(TableMetadata tableMetadata) {
-      this.metadataLocation = tableMetadata.metadataFileLocation();
       this.metadata = tableMetadata;
+      if (tableMetadata.metadataFileLocation() != null) {
+        this.metadataLocation = tableMetadata.metadataFileLocation();
+      }
```
kbendick, line 96: As we've removed the `withMetadataLocation` method from the builder, should I revert this change?

I do like that it shows the nullability a bit further to the reader, but it is arguably an unnecessary change.

core/src/test/java/org/apache/iceberg/LocalTableOperations.java
kbendick: No longer public.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
rdblue: Yes, please remove unnecessary changes.

core/src/test/java/org/apache/iceberg/rest/RequestResponseTestBase.java
```diff
@@ -104,7 +103,7 @@ protected void assertRoundTripSerializesEquallyFrom(String json, T expected) thr
     T actual = deserialize(json);
     assertEquals(actual, expected);
 
-    // Check that the deserialized value serializes back into the original JSON
-    Assertions.assertThat(serialize(expected)).isEqualTo(json);
+    Assert.assertEquals("The serialized version of the Java record should match the original JSON",
+        serialize(expected), json);
```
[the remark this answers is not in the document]
rdblue: Let's do this separately. I don't think this is needed to finish the tests, right?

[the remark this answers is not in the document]
kbendick: Nope. This is just a beneficial upgrade. I'll revert it now.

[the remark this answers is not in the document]
kbendick: I'll open this change in another PR (or just leave it be). But it's been reverted.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
kbendick: This has been reverted.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -28,6 +28,7 @@
 import org.apache.iceberg.BaseMetadataTable;
 import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.BaseTransaction;
+import org.apache.iceberg.HasTableOperations;
```
rdblue, line 31: Why has this class changed? Is it still needed?

kbendick: Nope. Sorry. I had asked on here but I think it got lost. I just reverted it though.

I can open those changes in another PR. They're not strictly needed for these tests (all tests still pass - but we should consider making those changes later as they are more correct that way).

rdblue: Why are those changes more correct?

kbendick: They've been reverted so once the tests pass, this is the minimum change set for these tests.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,218 @@
...
+    String json1 = String.format(
+        "{\"metadata-location\":\"%s\",\"metadata\":%s,\"config\":{\"foo\":\"bar\"}}",
+        TEST_METADATA_LOCATION, metadataJson);
+
+    TableMetadata metadataV1 = TableMetadataParser.fromJson(null, TEST_METADATA_LOCATION, metadataJson);
+    LoadTableResponse resp1 = LoadTableResponse.builder()
+        .withTableMetadata(metadataV1)
+        .addAllConfig(CONFIG)
+        .build();
+    assertEquals(deserialize(json1), resp1);
```
rdblue, line 148: This is the first half of what `assertRoundTripSerializesEquallyFrom` will check and the case is identical to the one above in `testRoundTripSerde`.

From the test name, I think what you intended to test is that all values that can be omitted will be deserialized to defaults. Should this omit "config" from the JSON string and validate that the response is equal to the one created without calling `addAllConfig`?

Also, there's no need for `json1` and `resp1` names since this only has one JSON string and one response object.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
```diff
@@ -0,0 +1,218 @@
...
+  public void testFailures() {
+    AssertHelpers.assertThrows(
+        "Table metadata should be required",
+        NullPointerException.class,
+        "Invalid metadata: null",
+        () -> LoadTableResponse.builder().build());
+  }
+
+  @Test
+  public void testRoundTripSerde() throws Exception {
```
rdblue, line 107: There's no need for this to put 2 different test cases in the same method. Can you test v1 and v2 separately?

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: > From the test name, I think what you intended to test is that all values that can be omitted will be deserialized to defaults. Should this omit "config" from the JSON string and validate that the response is equal to the one created without calling addAllConfig?

Yeah that's what this is testing. It's a little unclear from the way it's written admittedly.

I'll try removing `config` and then ensuring we get the default back.

And update the names.

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
kbendick: Yeah. Split it into two tests 👍

core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java
rdblue: > Yeah that's what this is testing. It's a little unclear from the way it's written admittedly.

I'm not sure I understand. What was this testing in the version I was looking at?

Initially, I was going to comment that it duplicates the first part of the round-trip test for v1. Then I realized that it was intended to test something else.

kbendick: In the version you were looking at, this was testing that the JSON parses into the class as expected, even if default values are not present.

Because default values are not present in the external JSON, I used those to represent missing values (which was the source of confusion).

For the `round trip` tests, it tests that the JSON deserializes into the object _as well as that the JSON matches exactly as expected the serialiazed output of the object_

That's the `round trip` in the `assertRoundTripSerialzesEquallyFrom`: https://github.com/apache/iceberg/blob/22298818cb7eba13294eda3b2dd696b773e0f170/core/src/test/java/org/apache/iceberg/rest/RequestResponseTestBase.java#L102-L109

kbendick: Do let me know if you think it would be better some other way (for this test suite at least). Though I'm anxious to get it in.

rdblue (approved):

rdblue:
Thanks, @kbendick

PR #5120, opened 2022-06-23
Flink: add sql partition functions

wuwenchi:
Add a partition function, you can easily view the partition where a value is located.
cc @kbendick thanks!


flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -619,12 +631,17 @@ public void alterPartition(ObjectPath tablePath, CatalogPartitionSpec partitionS
...
   @Override
   public List<String> listFunctions(String dbName) throws CatalogException {
-    return Collections.emptyList();
+    return Lists.newArrayList(partitionFunctions.keySet());
   }
 
   @Override
   public CatalogFunction getFunction(ObjectPath functionPath) throws FunctionNotExistException, CatalogException {
-    throw new FunctionNotExistException(getName(), functionPath);
+    CatalogFunction catalogFunction = partitionFunctions.get(functionPath.getObjectName());
```
kbendick, line 639: Are there any situations where this might get used for other reasons than just partition UDFs? Like accessing some other UDF that the user might have registered?

From the existing code, it looks fine. But for my own understanding, I’m still wondering if it’s possible for a user to register their own UDF that would be associated with this Catalog, and if so are we currently making it not possible to do that?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.iceberg.expressions.Literals;
+import org.apache.iceberg.transforms.Transform;
+import org.apache.iceberg.transforms.Transforms;
+import org.apache.iceberg.types.Type;
+import org.apache.iceberg.types.TypeUtil;
+
+public class PartitionTransformUdf {
+
+  public static class Truncate extends ScalarFunction {
+    public String eval(int num, @DataTypeHint(inputGroup = InputGroup.ANY) Object obj) {
```
kbendick, line 34: Is there any way for us to be more specific about what types this function can take? Does Flink use these annotations to check the users SQL before running the query?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,50 @@
...
+      Type type = TypeUtil.fromJavaType(obj);
+      Transform<Object, Object> truncate = Transforms.truncate(type, num);
+      Object value = truncate.apply(Literals.fromJavaType(obj).to(type).value());
```
kbendick, line 37: Do we need to do any more further type evaluation? What will happen if the type passed in is not supported?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,50 @@
...
+      return truncate.toHumanString(value);
+    }
+  }
+
+  public static class Bucket extends ScalarFunction {
+    public String eval(int num, @DataTypeHint(inputGroup = InputGroup.ANY) Object obj) {
+      Type type = TypeUtil.fromJavaType(obj);
+      Transform<Object, Integer> bucket = Transforms.bucket(type, num);
```
kbendick, line 45: Question for my own understanding:

Is it possible to make a typed version of the function? Is that common / desired generally speaking for Flink UDFs?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
```diff
@@ -0,0 +1,224 @@
...
+public class TestPartitionUdf {
+
+  public static TableEnvironment tEnv;
+
+  @BeforeClass
+  public static void before() {
+    StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+    tEnv = StreamTableEnvironment.create(env);
+    tEnv.createTemporarySystemFunction("buckets", PartitionTransformUdf.Bucket.class);
+    tEnv.createTemporarySystemFunction("truncates", PartitionTransformUdf.Truncate.class);
```
kbendick, line 49: Question: Can we provide a single registration option so things are uniform? The `bucket` and `truncate` functions are universally needed when working with Iceberg in my opinion (though not in all situations).

If we automatically register the functions for users when the catalog is instantiated, does that have any performance impact or downsides?

Also, for naming, is it possible that we might clash with some other name (if we registered the functions for people)?

I’m working on function registration in Spark so we should ideally get our names somewhat stabilized.

I would propose `iceberg_bucket` but need to better understand how it is used.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
```diff
@@ -0,0 +1,224 @@
...
+    Assert.assertEquals(expectHumanString, sql.get(0).getField(0));
+  }
+
+  @Test
+  public void testBucketLong() {
+    Transform<Object, Integer> bucket = Transforms.bucket(Types.LongType.get(), 4);
+
+    long num = 10;
+    Integer obj = bucket.apply(num);
+    String expectHumanString = bucket.toHumanString(obj);
```
kbendick, line 92: Nit: Can we have a test where the expected human string is tested? Eg with a hard coded string value as the expected human string and then the actual value is the result of this call?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
```diff
@@ -0,0 +1,224 @@
...
+  @Test
+  public void testTruncateByteBuffer() {
+    Transform<Object, Object> truncate = Transforms.truncate(Types.BinaryType.get(), 4);
+
+    Object obj = truncate.apply(ByteBuffer.wrap(new byte[] {1, 2, 3, 4, 5, 6}));
+    String expectHumanString = truncate.toHumanString(obj);
+
+    List<Row> sql = sql("SELECT truncates(4, x'010203040506')");
+    Assert.assertEquals(expectHumanString, sql.get(0).getField(0));
+  }
```
kbendick, line 223: I’m still reviewing these tests, but for any types that are _not_ supported, can we add a test checking what happens if those types are passed in?

Assuming an exception results, then `AssertHelpers` has some methods that should be useful.

kbendick (commented):
Thanks @wuwenchi!

I left some initial feedback based on my first pass through. Im still digesting this but had some questions and some feedback. 👍

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -107,11 +112,18 @@ public FlinkCatalog(
     this.catalogLoader = catalogLoader;
     this.baseNamespace = baseNamespace;
     this.cacheEnabled = cacheEnabled;
+    this.partitionFunctions = new LinkedHashMap<>();
```
kbendick, line 115: Nit: Does this have to be a LinkedHashMap? Normally we use the relocated Guava map constructors like `Maps.newHashMap()` or maybe the variant with expected size.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: > Are there any situations where this might get used for other reasons than just partition UDFs? Like accessing some other UDF that the user might have registered?

Here, we mainly use these udfs to create, modify, and delete partitionSpec through flink.

> From the existing code, it looks fine. But for my own understanding, I’m still wondering if it’s possible for a user to register their own UDF that would be associated with this Catalog, and if so are we currently making it not possible to do that?

Currently, it is not supported for users to register UDF under the flink catalog. But in the next step (just next PR), I will add this feature.

But after we support users to add UDFs, I still want to add these two functions, so that users do not need to manually add them again when they want to query related partitions, I would like to provide them by default. What do you think about this?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: Because I used `InputGroup.ANY`, in terms of parameter types, flink will pass any data type and finally rely on the original process to check the data type. The specific check is in the `get` function of the corresponding transform, for example
``` java
  static <T> Truncate<T> get(Type type, int width) {
    Preconditions.checkArgument(width > 0,
        "Invalid truncate width: %s (must be > 0)", width);

    switch (type.typeId()) {
      case INTEGER:
        return (Truncate<T>) new TruncateInteger(width);
      case LONG:
         .......
      default:
            "Cannot truncate type: " + type);
    }
  }
```

Both the bucket function and the truncate function can support many parameter types. If they are all split into functions with specific parameter types, it will be redundant, because only the input parameter types are different, and the function processing process is exactly the same.
Of course, we can know which types these functions support through the case branch above. This may not seem very straightforward...

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: Because `InputGroup.ANY` is used, any data type can be received. If a type that is not supported by a function is passed in, an unsupported exception will be reported in the corresponding transform, for example
```
SELECT truncates(4, TIMESTAMP '2022-05-20 10:12:55.038194');
...
  java.lang.UnsupportedOperationException: Cannot truncate type: timestamp
...
```

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: I once wrote a version, but they look redundant, because only the input parameter types are different, and the function processing process is exactly the same...

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
wuwenchi: Looks good, it will be clearer. I will modify it.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
wuwenchi: OK, I will add test cases for different types.

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
wuwenchi: Flink supports 4 properties categories of functions:
1. Temporary system functions
2. System functions
3. Temporary catalog functions
4. Catalog functions

If user is using Ambiguous Function Reference, The resolution order is:

1. Temporary system function
2. System function
3. Temporary catalog function, in the current catalog and current database of the session
4. Catalog function, in the current catalog and current database of the session

For details https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/table/functions/overview/


In the code, these two functions are registered in the catalog, so 
1. If the user uses iceberg's catalog, they can use these two functions directly. 
2. If the user does not use iceberg's catalog, they need to register these two functions before using them (eg written in test).

wuwenchi: eg:
```java
  // first, create iceberg catalog
  StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
  tEnv = StreamTableEnvironment.create(env);
  tEnv.executeSql("" +
          "CREATE CATALOG hive_catalog WITH (\n" +
          "    'type'='iceberg',\n" +
          "    'catalog-type'='hive',\n" +
          .....
          "  )");
    
    // second, call function by Precise Function Reference
    tEnv.executeSql("select hive_catalog.db.buckets(200, 12341)");        // you can use function directly by Precise Function Reference under hive_catalog

    // third, or call function by Ambiguous Function Reference
    tEnv.executeSql("use catalog hive_catalog");
    tEnv.executeSql("select buckets(200, 12341)");        // you can use function directly by Ambiguous Function Reference under hive_catalog
```

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
kbendick: Yeah i don't see any point in adding that redundancy. If the function throws, will we get `null` or will the user's program terminate?

I know there's been some debate about this with other functions. I believe it was the introduction of `try_cast` which uses old `cast` semantics, though I don't recall the details.

kbendick: Follow up question: can we define our own `InputGroup`? And if we did, would the SQL compiler fail to build the program or would it still just fail at runtime?

I guess if the annotation is tightened, what will flink do when it goes to compile a program out of SQL that passes the wrong type to a `DataHint`?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
kbendick: For the ambiguous function reference here, for clarification, this will work?

Once this is merged, this might be something we want to document, or at least point user's to the Flink documentation for function resolution when we mention these UDFs.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -619,12 +630,17 @@ public void alterPartition(ObjectPath tablePath, CatalogPartitionSpec partitionS
 
   @Override
   public List<String> listFunctions(String dbName) throws CatalogException {
-    return Collections.emptyList();
+    return Lists.newArrayList(partitionFunctions.keySet());
```
rdblue, line 633: I don't think it is a good idea to make the functions accessible from any database. Instead, I think you should choose where you want to expose these functions.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.iceberg.transforms.Transform;
+import org.apache.iceberg.transforms.Transforms;
+import org.apache.iceberg.types.Type;
+import org.apache.iceberg.types.TypeUtil;
+
+public class PartitionTransformUdf {
+
+  public static class Truncate extends ScalarFunction {
+    public String eval(int num, @DataTypeHint(inputGroup = InputGroup.ANY) Object obj) {
+      Type type = TypeUtil.fromJavaType(obj);
```
rdblue, line 35: This shouldn't look up the correct function implementation on every call. Instead, this should define a function for every supported input type and keep a specific reference to the Iceberg transform.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,50 @@
...
+      Transform<Object, Object> truncate = Transforms.truncate(type, num);
+      Object value = truncate.apply(Literals.fromJavaType(obj).to(type).value());
+      return truncate.toHumanString(value);
```
rdblue, line 38: This should produce the original data type, not a string. Otherwise, the result is useless.

api/src/main/java/org/apache/iceberg/expressions/Literals.java
```diff
@@ -88,6 +88,22 @@ static <T> Literal<T> from(T value) {
         "Cannot create expression literal from %s: %s", value.getClass().getName(), value));
   }
 
+  public static Literal<Object> fromJavaType(Object value) {
```
rdblue, line 91: I don't think this should be added to the public API. We avoided adding support for these because Iceberg is not responsible for time zone conversions.

It would be fine to revisit that decision, but I think this is only used so that the implementation is kept simple, although it will have poor performance. Instead, I think this should define real functions that are typed, instead of accepting any input type and dynamically calling functions.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
rdblue: Dynamically calling `bucket` based on the Java type is really dangerous. If the type is wrong, then there are going to be correctness issues. This should define a function per supported type.

rdblue (changes requested):
Thanks for working on this, @wuwenchi! I think the approach is a little off, but we should be able to get this working soon.

api/src/main/java/org/apache/iceberg/expressions/Literals.java
kbendick: +1. Particularly about keeping this out of `api` where we have compatibility guarantees.

kbendick: Also calling `Literal.of(value.toString())` for every input seems like a performance concern.

Though I agree with Ryan that with some minor changes this should be a good addition overall :)

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: > can we define our own InputGroup

We can customize input and output by overriding `getTypeInference` method:
```java
@Override
public TypeInference getTypeInference(DataTypeFactory typeFactory) {
    return TypeInference.newBuilder()
            .inputTypeStrategy(...)       // custom input
            .outputTypeStrategy(...)      // custom outpout
            .build();
}
```

> And if we did, would the SQL compiler fail to build the program or would it still just fail at runtime?

If you throw an exception in `inferInputTypes`, then flink will throw exception and terminate the program:
```java
@Override
public TypeInference getTypeInference(DataTypeFactory typeFactory) {
    return TypeInference.newBuilder()
        .inputTypeStrategy(new InputTypeStrategy() {
        ...
        @Override
        public Optional<List<DataType>> inferInputTypes(CallContext callContext, boolean throwOnFailure) {
            throw new UnsupportedOperationException("exception");
        }
        ...
        })
        .build();
}

org.apache.flink.table.api.ValidationException: SQL validation failed.
...
Caused by: java.lang.UnsupportedOperationException: exception
```




> I guess if the annotation is tightened, what will flink do when it goes to compile a program out of SQL that passes the wrong type to a DataHint?

Yes, if parameter type is not matched `eval`'s parameter type , flink will throw an exception directly:
```
org.apache.flink.table.api.ValidationException: SQL validation failed. Invalid function call
```

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
wuwenchi: It will work. Because if `use catalog hive_catalog` is not used, then the current catalog is the `default_catalog`, and will can't find the `buckets` function.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: For `bucket`, the original type (int) is directly returned, and the user can find the corresponding partition according to the int. But for `truncate`, if the input parameter type is binary, the original data type returned is 'ByteBuffer', and iceberg will convert it into a string after getting the buffer (https://github.com/wuwenchi/iceberg/blob/35b8558d7c0b5564fb3d47f92721534f45850cd6/api/src/main/java/org/apache/iceberg/transforms/Truncate.java#L459-L461). So, in this case, should we use string?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: This interface will show all the functions under the current iceberg catalog. My idea is that these functions are registered under the iceberg catalog level, and all databases under the current iceberg catalog can be used. If the user uses a non-iceberg catalog, these functions are not visible.   I think the access boundary of these functions is the catalog level, so is it possible to not need to display or limit the database?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
rdblue: `toHumanString` should not be used, so that all truncate functions return the same type that was passed in. The `toHumanString` method is for producing strings that can be embedded in paths, not for coercing values like this.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
kbendick: It seems like maybe adding the TypeInference might be a good way to go.

In addition to adding a specific function for each type.

That way, if I'm not mistaken, the compiler will be able to reject any invalid input. Then, as Ryan mentioned, we should have a function for each possible type and the correct function should be picked ahead of time based on type from the compiler.

This would avoid the runtime reflection used on each function call which has large overheard and should be avoided whenever possible.

wuwenchi: Now it has been changed to the way that each parameter corresponds to a function implementation.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: Now it has been changed to the way that each parameter corresponds to a function implementation.

api/src/main/java/org/apache/iceberg/expressions/Literals.java
wuwenchi: Yes, this way is really not suitable. After the change, this function is no longer used. Deleted.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,159 @@
...
+      Transform<String, String> truncate = Transforms.truncate(Types.StringType.get(), num);
+      return truncate.apply(value);
+    }
+
+    public byte[] eval(int num, byte[] value) {
+      ByteBuffer wrap = ByteBuffer.wrap(value);
+      Transform<ByteBuffer, ByteBuffer> truncate = Transforms.truncate(Types.BinaryType.get(), num);
+      ByteBuffer byteBuffer = truncate.apply(wrap);
+      byte[] out = new byte[byteBuffer.remaining()];
+      byteBuffer.get(out, 0, out.length);
```
rdblue, line 65: This should use `ByteBuffers.toByteArray(buf)`.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,159 @@
...
+    public int eval(int num, @DataTypeHint("DATE") LocalDate value) {
+      Transform<Object, Integer> bucket = Transforms.bucket(Types.DateType.get(), num);
+      int days = (int) value.toEpochDay();
+      return bucket.apply(days);
+    }
+
+    public int eval(int num, @DataTypeHint("TIME") LocalTime value) {
+      Types.TimeType type = Types.TimeType.get();
+      Transform<Object, Integer> bucket = Transforms.bucket(type, num);
+      long micros = TimeUnit.NANOSECONDS.toMicros(value.toNanoOfDay());
```
rdblue, line 130: Can you use the conversion functions in `DateTimeUtil`?

flink/v1.15/flink/src/test/java/org/apache/iceberg/flink/sink/TestPartitionUdf.java
rdblue: Can you separate the tests for registering these as UDFs from the tests for calling them through the catalog? I want to make sure we have separate tests for those two cases.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,159 @@
...
+      return truncate.apply(value);
+    }
+
+    public long eval(int num, long value) {
+      Transform<Long, Long> truncate = Transforms.truncate(Types.LongType.get(), num);
+      return truncate.apply(value);
+    }
+
+    public String eval(int num, String value) {
+      Transform<String, String> truncate = Transforms.truncate(Types.StringType.get(), num);
```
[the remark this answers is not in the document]
wuwenchi: Do you mean like this?
```java
public class PartitionTransformUdf {

 public static Type stringType = Types.StringType.get();

  public static class Truncate extends ScalarFunction {
    public String eval(int num, String value) {
      Transform<String, String> truncate = Transforms.truncate(stringType, num);
      return truncate.apply(value);
    }
 }
```

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: done

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: done

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
rdblue: I'm not sure whether Flink allows you to omit the database name or not, but that's up to Flink. Iceberg should not expose these functions as though they exist in all databases, though.

wuwenchi: What you said is more reasonable. It is more appropriate for this interface to display the functions registered by the user.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -624,7 +635,12 @@ public List<String> listFunctions(String dbName) throws CatalogException {
 
   @Override
   public CatalogFunction getFunction(ObjectPath functionPath) throws FunctionNotExistException, CatalogException {
-    throw new FunctionNotExistException(getName(), functionPath);
+    CatalogFunction catalogFunction = partitionFunctions.get(functionPath.getObjectName());
```
rdblue, line 638: I see that you've removed the `listFunctions` method, but that doesn't actually solve the problem -- it introduces a new one. The problem is that these functions should not be defined in every database, but that's what this `getFunction` implementation does because it ignores `functionPath.getDatabaseName()`.

Whatever this method does, `listFunctions` should be consistent with its behavior. If this will load a function, then `listFunctions` should show it.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.iceberg.util.Pair;
+
+public class PartitionTransformUdf {
+
+  public static class Truncate extends ScalarFunction {
+
+    public static final Cache<Tuple2<Integer, Type>, Transform<Object, Object>> TRUNCATE_CACHE =
+        Caffeine.newBuilder().build();
+
+    public static Transform<Object, Object> getTruncateTransform(int num, Type type) {
```
rdblue, line 53: There's no need for this method to be public. Can you make it private instead? Same with the cache.

Also, Iceberg does not typically use `get` in method names because it isn't helping. There is usually a better verb, or it can be omitted.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
wuwenchi: My original idea was that these functions belong to the catalog (it may not appropriate now) and that the functions are registered when the catalog is created so that users can use them directly. So I provide the function in `getFunctions` and show the function in `listFunctions`. If the function is not displayed to the user now, and these functions cannot be get directly (also functions will not be directly registered in the catalog), should we mention these functions in the documentation, and provide an example for register and use function ?

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/PartitionTransformUdf.java
wuwenchi: done

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -161,19 +167,33 @@ public List<String> listDatabases() throws CatalogException {
...
+        new CatalogFunctionImpl(PartitionTransformUdf.Bucket.class.getName(), FunctionLanguage.JAVA));
+    partitionFunctions.putIfAbsent(databaseName + ".truncates",
+        new CatalogFunctionImpl(PartitionTransformUdf.Truncate.class.getName(), FunctionLanguage.JAVA));
+  }
+
+  public void deregisterPartitionFunction(String databaseName) {
+    partitionFunctions.remove(databaseName + ".buckets");
+    partitionFunctions.remove(databaseName + ".truncates");
+  }
+
```
wuwenchi, line 181: If we register the function under the database, we need to first get the database where the user is currently located.  When the user enters `use database` or `create database` in sql, some interfaces will be called. If the database exists, we can register the function under the specified database.

But there is a problem here, as long as the database actually exists, even if the iceberg table does not exist in the database, these functions will be registered. This database may be a hive database, the user may have specified the wrong database, resulting in the use of a wrong database, but we will still register these functions under this database, because we cannot judge whether this database is the iceberg table, or under the database, there are not only iceberg tables, but also hive tables.


My idea is that since we can't distinguish whether there is an iceberg table under the current database, when the user specifies a database, these functions can be registered. Or, is there a good way to distinguish the above cases?

wuwenchi: @rdblue

wuwenchi: If the timing of registering functions is appropriate, these functions will be displayed in the listfunction later, and some test cases will be added.

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
pvary: How Flink users/catalogs handled? Is it possible for one user to access/see other user's functions/catalogs?

pvary: Reading the doc linked below: https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/table/functions/overview/, I think I understand it now.

Basically we are registering temporary catalog functions, so the queries will use it anyway, and will not clash with other implementations already registered on the catalog. What happens if the user already registered a `bucket` temporary catalog function?

Why do we use catalog functions instead of system functions? I think the highest priority is the temporary system function, so if we use them we can be reasonably sure that we do not call any user define functions by mistake.

Another issue I see is that if we already has a function registered with the same name, then it might be surprising for the user that the implementation is changed for iceberg tables (not the old catalog level implementation is used, but the one registered by the Iceberg code). Using a more specific name (`iceberg_bucket`), like the one suggested by @kbendick would help to reduce the likelihood of name clashes.

pvary: I have found the answers for my deleted questions 😄

flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/FlinkCatalog.java
```diff
@@ -162,20 +168,33 @@ public List<String> listDatabases() throws CatalogException {
...
       } else {
+        registerPartitionFunction(databaseName);
         return new CatalogDatabaseImpl(Maps.newHashMap(), "");
       }
     } else {
       try {
         Map<String, String> metadata =
             Maps.newHashMap(asNamespaceCatalog.loadNamespaceMetadata(toNamespace(databaseName)));
         String comment = metadata.remove("comment");
+        registerPartitionFunction(databaseName);
```
rdblue, line 197: @wuwenchi, these functions should not be registered in every known database. The problem before was not how you were tracking functions, it was the behavior. You need to change the behavior.

Here's my last comment:

> The problem is that this is effectively registering the functions in every database. Instead, choose one database where you want the functions to live. For stored procedures, we use `system`. I think that's a good idea here.

Please remove the registration code that isn't needed and update the function lookup and listing to only return these functions if the database is `system`.

wuwenchi: About the `system` database, is this database just a logical database? Is it that users cannot create tables under this library? In sql, is it not possible to execute `use iceberg_catalog.system`?  Because this is related to the list of functions, if we you cannot use `use system` to switch the currently used database, we cannot list functions under the `system`, because `show functions` can only display the functions under the current database.

rdblue: I don't see why `system` would be different than any other database. It should be a normal database.

wuwenchi: If it is a normal database, should we create it first when open catalog  like the `default` database? Because if it is not created, the user cannot see the database, but can use the functions under the `system` database, which feels a little strange.

PR #5124, opened 2022-06-23
Python: Add sort order fields

Fokko:
This is required when we start (de)serializing the TableMetadata from the REST catalog.


python/src/iceberg/table/sorting.py
```diff
@@ -0,0 +1,60 @@
...
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Optional
+
+from iceberg.transforms import Transform
+
+
+class SortDirection(Enum):
+    asc = "asc"
+    desc = "desc"
```
rdblue, line 26: ALL CAPS?

python/src/iceberg/table/sorting.py
```diff
@@ -0,0 +1,60 @@
...
+
+@dataclass(frozen=True)
+class SortField:
+    source_id: int
+    transform: Transform
+    direction: SortDirection
+    null_order: NullOrder
+
+
+@dataclass(frozen=True)
```
rdblue, line 42: Should these be Pydantic classes?

python/tests/table/test_sorting.py
```diff
@@ -0,0 +1,53 @@
...
+    NullOrder,
+    SortDirection,
+    SortField,
+    SortOrder,
+)
+from pyiceberg.transforms import IdentityTransform
+from pyiceberg.types import IntegerType, StringType
+
+
+def test_serialize_sort_order():
```
rdblue, line 28: Does deserialization also work?

python/pyiceberg/table/metadata.py
```diff
@@ -270,7 +271,7 @@ def set_sort_orders(cls, data: Dict[str, Any]):
         # This is going to be much nicer as soon as sort-order is an actual pydantic object
         # Probably we'll just create a UNSORTED_ORDER constant then
```
kbendick, line 272: Do these two lines still apply now that we have the constants and are using `IcebergBaseModel`?

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,84 @@
...
+    NULLS_LAST = "nulls-last"
+
+
+class SortField(IcebergBaseModel):
+    def __init__(
+        self,
+        source_id: int,
+        transform: Transform,
+        direction: SortDirection = SortDirection.ASC,
+        null_order: NullOrder = NullOrder.NULLS_LAST,
```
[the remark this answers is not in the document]
Fokko: That makes a lot of sense, thanks!

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,84 @@
...
+    source_id: int = Field(alias="source-id")
+    transform: Transform = Field()
+    direction: SortDirection = Field()
+    null_order: NullOrder = Field(alias="null-order")
+
+
+class SortOrder(IcebergBaseModel):
+    """
+    Users can sort their data within partitions by columns to gain performance.
+    The information on how the data is sorted can be declared per data or delete file, by a sort order.
```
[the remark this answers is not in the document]
Fokko: Done!

python/tests/table/test_sorting.py
Fokko: I still need to implement that one, mostly because of the transform. `bucket[4]` cannot be deserialized to the implementation without knowing which type the source-id refers to.

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,84 @@
...
+        data["order-id"] = order_id
+        if fields:
+            data["fields"] = fields
+        super().__init__(**data)
+
+    order_id: Optional[int] = Field(alias="order-id")
+    fields: List[SortField] = Field(default_factory=list)
+
+
+UNPARTITIONED_SORT_ORDER_ID = 0
```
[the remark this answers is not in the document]
Fokko: Working at two things at the same time lol, thanks!

python/tests/table/test_sorting.py
rdblue: In Java, we recently introduced `UnboundSortOrder` and `UnboundPartitionSpec` to fix this problem. We may want to do the same here, where the transform is not instantiated until the order is bound to a schema.

python/pyiceberg/table/metadata.py
Fokko: I'll remove this right away, thanks for the sharp eye @kbendick

python/tests/table/test_sorting.py
Fokko: I was working on something similar, thanks!

Fokko: Just updated this with a `UnboundTransform` that deserializes properly back and forth, and can be bound using the `bind()` method. Also added a lot of tests 👍🏻

python/pyiceberg/table/metadata.py
```diff
@@ -62,21 +62,25 @@ def check_partition_specs(values: Dict[str, Any]) -> Dict[str, Any]:
...
+        raise ValidationError(f"default-sort-order-id {default_sort_order_id} can't be found in {sort_orders}")
     return values
 
 
 class TableMetadataCommonFields(IcebergBaseModel):
     """Metadata for an Iceberg table as specified in the Apache Iceberg
     spec (https://iceberg.apache.org/spec/#iceberg-table-spec)"""
 
+    def current_schema(self) -> Schema:
+        return next(schema for schema in self.schemas if schema.schema_id == self.current_schema_id)
```
rdblue, line 82: We may want to index these for quick lookup, but it seems minor. I'm not really sure if it is worth it.

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,133 @@
...
+    DESC = "desc"
+
+
+class NullOrder(Enum):
+    NULLS_FIRST = "nulls-first"
+    NULLS_LAST = "nulls-last"
+
+
+class SortField(IcebergBaseModel):
+    """A field that provides information on how the table is sorted
```
rdblue, line 48: This doesn't necessarily have information on how the table is sorted. These are defined sort orders. There's not necessarily any data in the sort order in a given table.

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Args:
+      source_id (int): Source column id from the table’s schema
+      transform (str): Transform that is used to produce values to be sorted on from the source column.
+                       This is the same transform as described in partition transforms.
+      source_id (SortDirection): Sort direction, that can only be either asc or desc
```
rdblue, line 54: Should be `direction` and `null_order`

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,133 @@
...
+            data["transform"] = transform
+        if direction is not None:
+            data["direction"] = direction
+        if null_order is not None:
+            data["null-order"] = null_order
+        super().__init__(**data)
+
+    @root_validator(pre=True)
+    def set_null_order(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        values["direction"] = values["direction"] if values.get("direction") else SortDirection.ASC
```
rdblue, line 81: This doesn't need to produce a `str` value? (Assuming since the type is `Any` it works, but this is unexpected for me)

python/pyiceberg/transforms.py
```diff
@@ -64,6 +70,31 @@ class Transform(IcebergBaseModel, ABC, Generic[S, T]):
...
+    def validate(cls, v: Any):
+        # When Pydantic is unable to determine the subtype
+        # In this case we'll help pydantic a bit by parsing the transform type ourselves
+        if isinstance(v, str):
+            if v == "identity":
+                bind_func = identity
+            elif v == "void":
+                bind_func = always_null  # type: ignore
+            elif v.startswith("bucket"):
+                bind_func = functools.partial(bucket, num_buckets=BaseBucketTransform.parse(v))
```
rdblue, line 90: In types.py, `parse` is used to return an instance of the type, not to parse part of the string form:

```python
FixedType.parse("fixed[16]") // -> FixedType(length=16)
```

If possible, I'd keep the expectations for those methods the same and return a `BucketTransform` from `BucketTransform.parse`.

python/pyiceberg/transforms.py
```diff
@@ -96,6 +127,38 @@ def dedup_name(self) -> str:
     def __str__(self) -> str:
         return self.__root__
 
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, Transform):
+            return self.__root__ == other.__root__
+        return False
+
+
+class UnboundTransform(Transform):
```
rdblue, line 136: Looking at the Java side a little more, I think there's an alternative to making this `UnboundTransform` change. The Java implementation really only needs the source type for the `apply` method. I think you could solve this problem by changing the `Transform` API. Instead of having `apply` on `Transform`, you could put `apply` on a `Transformer` class, then add `forType(t: IcebergType) -> Transformer` to `Transform`. That may be cleaner in the long term, although I think this works for now.

python/pyiceberg/table/metadata.py
```diff
@@ -92,6 +96,14 @@ def construct_refs(cls, data: Dict[str, Any]):
             data["refs"] = {MAIN_BRANCH: SnapshotRef(snapshot_id=current_snapshot_id, snapshot_ref_type=SnapshotRefType.BRANCH)}
         return data
 
+    def bind(self):
```
rdblue, line 99: The main issue I have with this commit is that binding is done to transform existing state. I think that's going to cause problems later.

For example, if I load a table and bind all of the sort orders to the schema, what happens when I then update the schema? I think we would normally want to be able to copy all of the sort orders into a new `TableMetadata` object, but we would actually have to "unbind" them and rebind them to the new schema. This in-place binding is hard to work with.

Instead, I'd recommend having separate lists of unbound and bound things. Then when you want to get a sort order for a particular write, you can get the unbound order, bind it to the write schema, and know it is going to work.

rdblue (changes requested):
This is looking great! The main issue is that I don't think binding should modify objects. It will be easier in the long run if we bind when necessary.

python/pyiceberg/table/metadata.py
Fokko: This will stop the iteration when it finds it. We could also cache it. When we have more lookups on a specific id I think it makes sense to index it.

python/pyiceberg/table/sorting.py
Fokko: Copy/paste, thanks! I need to set a checker on this some day

python/pyiceberg/table/sorting.py
Fokko: Of course, you can put everything in there. This will be checked when the validator does its job, but that hasn't been done at the `pre_validator` step.

python/pyiceberg/transforms.py
Fokko: If I understand correctly, this is what we already have for each of the classes of Transforms:
```python
def bucket(source_type: IcebergType, num_buckets: int) -> BaseBucketTransform:
    if type(source_type) in {IntegerType, LongType, DateType, TimeType, TimestampType, TimestamptzType}:
        return BucketNumberTransform(source_type, num_buckets)
    elif isinstance(source_type, DecimalType):
        return BucketDecimalTransform(source_type, num_buckets)
    elif isinstance(source_type, StringType):
        return BucketStringTransform(source_type, num_buckets)
    elif isinstance(source_type, BinaryType):
        return BucketBytesTransform(source_type, num_buckets)
    elif isinstance(source_type, FixedType):
        return BucketBytesTransform(source_type, num_buckets)
    elif isinstance(source_type, UUIDType):
        return BucketUUIDTransform(source_type, num_buckets)
    else:
        raise ValueError(f"Cannot bucket by type: {source_type}")
```
This is actually called on the `.bind()` method by the transform:
```python
bind_func = functools.partial(bucket, num_buckets=BaseBucketTransform.parse(v))
```
As soon as the source_type is known, we can invoke the method and get the actual implementation.

python/pyiceberg/table/metadata.py
```diff
@@ -159,10 +163,14 @@ def construct_refs(cls, data: Dict[str, Any]):
...
     recent entries after a commit."""
 
-    sort_orders: List[Dict[str, Any]] = Field(alias="sort-orders", default_factory=list)
+    sort_orders: List[SortOrder] = Field(alias="sort-orders", default_factory=list)
     """A list of sort orders, stored as full sort order objects."""
 
-    default_sort_order_id: int = Field(alias="default-sort-order-id", default=DEFAULT_SORT_ORDER_UNSORTED)
+    @property
+    def bound_sort_orders(self) -> List[SortOrder]:
+        return [sort_order.bind(self.current_schema()) for sort_order in self.sort_orders]
```
rdblue, line 171: For now, I'd probably just make one accessor that binds the default sort order to the current schema. I'm not sure if it will be valuable to have all of the known sort orders bound to a particular schema.

python/pyiceberg/table/sorting.py
```diff
@@ -0,0 +1,133 @@
...
+    transform: Transform = Field()
+    direction: SortDirection = Field()
+    null_order: NullOrder = Field(alias="null-order")
+
+    def bind(self, schema: Schema) -> "SortField":
+        return SortField(
+            source_id=self.source_id,
+            transform=(
+                self.transform.bind(schema.find_type(self.source_id))
+                if isinstance(self.transform, UnboundTransform)
```
rdblue, line 92: There's no need to make a new `SortField` if this one is already bound, right? This could be:

```python
if isinstance(self.transform, UnboundTransform):
    return SortField(..., transform=self.transform.bind(...))
else:
    return self
```

python/pyiceberg/transforms.py
rdblue: Yes, this definitely works.

My point was just that we could structure transforms slightly differently, so that we don't need to bind in most situations. The only time we actually need to get the concrete source type for a transform is when we are calling `apply`. The other method signatures pass a type in. For example, `can_transform(IcebergType)` doesn't need to be specific to a `LongType` -- we can write a version for `BucketTransform` instead of for `BucketNumberTransform`.

In that case, we'd change the code so that instead of calling `apply` on `BucketNumberTransform`, we'd call `BucketTransform.bind` to get `Function[int, int]` that implements the hash and bucketing logic.

That refactor may end up being easier to work with in the end because there would be far fewer places where we'd need to bind the transform to a specific type.

rdblue: It may not be something that we want to do in this PR. We can follow up to change it, probably

Fokko: I have a working version, which make the code much simpler, which I like. I think it makes sense to do it in this PR, since it will remove 80% of the tests I already wrote :) I think the code is much nicer because we can omit the bind() stuff, including all the copying of the object.

python/pyiceberg/table/sorting.py
Fokko: Good call, but this will be obsolete as well with just passing down the hashing functions.

python/pyiceberg/transforms.py
rdblue: Sounds great!

python/pyiceberg/transforms.py
```diff
@@ -64,11 +72,32 @@ class Transform(IcebergBaseModel, ABC, Generic[S, T]):
...
+                return BucketTransform(num_buckets=BUCKET_PARSER.match(v))
+            elif v.startswith(TRUNCATE):
+                return TruncateTransform(width=BUCKET_PARSER.match(v))
+            else:
+                return UnknownTransform(transform=v)
+        return v
 
     @abstractmethod
-    def apply(self, value: Optional[S]) -> Optional[T]:
+    def hash_function(self, source: IcebergType) -> Callable[[Optional[S]], Optional[T]]:
```
rdblue, line 100: While `hash_function` makes sense for the bucket transform, I think we want a more generic name for the base class. What about `transformer`? That way a `Transform` is generic and `transformer` is specific to a type.

python/pyiceberg/transforms.py
```diff
@@ -312,6 +293,28 @@ def source_type(self) -> IcebergType:
...
+        source_type = type(source)
+        if source_type in {IntegerType, LongType}:
+
+            def truncate_func(v):
+                return v - v % self._width
+
+        elif source_type in {StringType, BinaryType}:
+
+            def truncate_func(v):
+                return v[0 : min(self._width, len(v))]
```
rdblue, line 306: Nice that this reduces code duplication as well.

python/pyiceberg/transforms.py
```diff
@@ -418,69 +392,39 @@ class UnknownTransform(Transform):
...
-        self._source_type = source_type
         self._transform = transform
 
-    def apply(self, value: Optional[S]):
+    def hash_function(self, source: IcebergType) -> Callable[[Optional[S]], Optional[T]]:
         raise AttributeError(f"Cannot apply unsupported transform: {self}")
 
     def can_transform(self, source: IcebergType) -> bool:
-        return self._source_type == source
+        return False
```
rdblue, line 403: We may need to change this later. I'm not sure what the implication will be.

rdblue (changes requested):
This looks great! I think the only remaining thing is the `hash_function` name. Since it isn't a hash, we should find a better name for the method that returns the actual transform implementation.

python/pyiceberg/transforms.py
Fokko: Good call, I've updated the code

python/pyiceberg/transforms.py
Fokko: Yes, like it a lot

python/pyiceberg/transforms.py
Fokko: I had the same feeling, but the `UnknownTransform` feels like a transform that you don't want to use anyway. Returning `False`, felt like the right thing to do.

Fokko:
@rdblue good call, I've updated it to `transform`, which makes a ton of sense to me in the context

rdblue (approved):

rdblue:
Thanks @Fokko! Good to have this in. And I'm excited to see how much cleaner things are with the new approach to Transform.

PR #5125, opened 2022-06-23
Reuse rest client for catalog and table operations

danielcweeks:


rdblue (approved):

rdblue:
This looks good to me.

For anyone looking at this later, the context is that we no longer need to create a client per table now that the Authorization HTTP header is handled by sessions rather than by client. That was the purpose of adding sessions, since we don't need to create a different client (and connection pool) for every auth session.

rdblue:
Thanks, @danielcweeks!

PR #5126, opened 2022-06-23
Parquet: Release the compressor when calling close in ParquetWriter

hmit:


hmit:
I discovered this after switching to `ParquetWriter` from iceberg project.
Without this, we see [this log message](https://github.com/rdblue/brotli-codec/blob/master/src/main/java/org/apache/hadoop/io/compress/brotli/BrotliCompressor.java#L249) when using Brotli compression to write data files to iceberg.

jzhuge (approved):
The commit message should have prefix "Parquet: " instead of "Core: ".

kbendick (approved):
+1. Good catch @hmit!

I tried to find a place where this could be tested, but there didn't seem to be anywhere that was already testing whether or not things were closed so I think it's fine without it.

hmit:
> +1. Good catch @hmit!
> 
> I tried to find a place where this could be tested, but there didn't seem to be anywhere that was already testing whether or not things were closed so I think it's fine without it.

Yeah, I didn't find a good unit-test to add this to. However, we did test it by deploying (with brotli compression) and it seems to do the trick.

szehon-ho (approved):

szehon-ho:
Thanks @hmit , and @kbendick @jzhuge for review

hmitnflx:
Thanks all for the quick review and merge!

jzhuge:
Thanks @hmit for discovering the issue and providing a fix. Thanks @kbendick for review. Thanks @szehon-ho for review and merge!

PR #5127, opened 2022-06-24
A few small fixes to docs markdown files

samredai:
- Remove no longer needed files
- Remove double quote typo
- Replace Iceberg logo with `#Documentation`
- Add front-matter for view-spec and puffin-spec docs pages


docs/configuration.md
```diff
@@ -1,4 +1,5 @@
 ---
+title: "Configuration"
```
samredai, line 2: @rdblue the issue you were seeing with the search results container was not actually because it was behind the top navbar. It was caused by this `configuration.md` file not having a title set. This should fix it!

rdblue (approved):

rdblue:
Thanks, @samredai!

PR #5129, opened 2022-06-24
Add source snapshot info to Puffin Blob metadata

findepi:
Extracted from https://github.com/apache/iceberg/pull/5021


core/src/main/java/org/apache/iceberg/puffin/FileMetadataParser.java
```diff
@@ -41,6 +41,8 @@ private FileMetadataParser() {
 
   private static final String TYPE = "type";
   private static final String FIELDS = "fields";
+  private static final String SOURCE_SNAPSHOT_ID = "source-snapshot-id";
+  private static final String SOURCE_SNAPSHOT_SEQUENCE_NUMBER = "source-snapshot-sequence-number";
```
rdblue, line 45: What is the value of `source` here? What about just using `snapshot-id`?

Also, it should just be `sequence-number` and not `snapshot-sequence-number`.

rdblue:
Mostly looks good. Just need to agree on and fix the names.

core/src/main/java/org/apache/iceberg/puffin/FileMetadataParser.java
findepi: Updated

findepi:
AC

rdblue (approved):

findepi:
@rdblue  Now that we have this on the blob metadata level, do we still need to have `org.apache.iceberg.StatisticsFile#snapshotId` field?

cc @ajantha-bhat

PR #5130, opened 2022-06-24
Python: Add flake8-comprehensions

Fokko:
It’s slower to call e.g. `dict()` than using the empty literal `{}`, because the name `dict` must be looked up in the global scope in case it has been rebound.


rdblue (approved):

rdblue:
Thanks, @Fokko!

PR #5131, opened 2022-06-24
Core: Fix partition clustering to produce table sort order

rdblue:
This updates `SortOrderUtil.buildSortOrder` to produce a correct clustering when building a final sort order. This method creates a new sort order that clusters rows for a given partition spec and produces the ordering that was passed in for rows within the same partition.

Previously, this was done by checking whether the partition source fields were included in the sort order, and that the transform applied to the source data satisfied the partition field's sort order. This simple approach left a few cases where the clustering was not actually satisfied.

For example: when partitioning by `[ day(ts), category ]` and sorting by `[ ts, category ]`, the original order would be used because `ts` satisfies the order of `days(ts)` and `category` satisfies the order for `category`. However, this fails for the following data:

```text
2022-06-24T00:00:00.000000, a    -> Partition(2022-06-24, a)
2022-06-24T00:00:00.000000, b    -> Partition(2022-06-24, b)
2022-06-24T00:00:00.000001, a    -> Partition(2022-06-24, a)    <-- fails when writing from Spark
2022-06-24T00:00:00.000001, b    -> Partition(2022-06-24, b)
```

If the sort order were reversed, `[ category, ts ]` then the sort order does not fail.

```text
2022-06-24T00:00:00.000000, a    -> Partition(2022-06-24, a)
2022-06-24T00:00:00.000001, a    -> Partition(2022-06-24, a)
2022-06-24T00:00:00.000000, b    -> Partition(2022-06-24, b)
2022-06-24T00:00:00.000001, b    -> Partition(2022-06-24, b)
```

The problem is that a sort field that satisfies a partition field's ordering but is more granular can only be used as the last field to produce the partition clustering. If another partition field comes after, the clustering is no longer valid.

This PR fixes the problem by changing how the partition clustering prefix is constructed:
1. Build a map of the required clustering fields
2. For each sort field, check if it matches the transform/source of a required clustering field
3. For fields that match a clustering field, remove that field because it is satisfied by the ordering
4. Once a sort field that does not match a clustering field is found, stop considering sort fields
5. For the last field that does not match a clustering field, check if it satisfies the ordering of a clustering field and, if so, remove the clustering field
6. Add any remaining clustering fields to the start of the final order

For the example above with `[ days(ts), category ]` and sort order `[ ts, category ]`:
1. The required clustering fields are `days(ts)` and `category`.
2. There are no fields at the start of the sort order that exactly match a required clustering field
3. The sort field `ts` satisfies the order of `days(ts)`, so that is removed
4. The remaining required clustering field is `category`
5. `category` is added to the beginning of the sort order
6. The final sort order is `[ category, ts, category ]`, which satisfies the required partition clustering and retains the incoming ordering within partitions.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/util/SortOrderUtil.java
+++ b/core/src/main/java/org/apache/iceberg/util/SortOrderUtil.java
@@ -66,10 +66,46 @@ public class SortOrderUtil {
     }
 
     // make a map of the partition fields that need to be included in the clustering produced by the sort order
-    Map<Pair<Transform<?, ?>, Integer>, PartitionField> partitionFields = Maps.newLinkedHashMap();
+    Map<Pair<Transform<?, ?>, Integer>, PartitionField> requiredClusteringFields = requiredClusteringFields(spec);
+
+    // remove any partition fields that are clustered by the sort order by iterating over a prefix in the sort order.
+    // this will stop when a non-partition field is found, or when the sort field only satisfies the partition field.
+    for (SortField sortField : sortOrder.fields()) {
+      Pair<Transform<?, ?>, Integer> sourceAndTransform = Pair.of(sortField.transform(), sortField.sourceId());
+      if (requiredClusteringFields.containsKey(sourceAndTransform)) {
+        requiredClusteringFields.remove(sourceAndTransform);
+        continue; // keep processing the prefix
+      }
+
+      // if the field satisfies the order of any partition fields, also remove them before stopping
+      // use a set to avoid concurrent modification
+      for (PartitionField field : spec.fields()) {
+        if (sortField.sourceId() == field.sourceId() && sortField.transform().satisfiesOrderOf(field.transform())) {
+          requiredClusteringFields.remove(Pair.of(field.transform(), field.sourceId()));
+        }
+      }
+
+      break;
+    }
+
+    // build a sort prefix of partition fields that are not already in the sort order's prefix
+    SortOrder.Builder builder = SortOrder.builderFor(schema);
+    for (PartitionField field : requiredClusteringFields.values()) {
+      String sourceName = schema.findColumnName(field.sourceId());
+      builder.asc(Expressions.transform(sourceName, field.transform()));
+    }
+
+    // add the configured sort to the partition spec prefix sort
+    SortOrderVisitor.visit(sortOrder, new CopySortOrderFields(builder));
+
+    return builder.build();
+  }
+
+  private static Map<Pair<Transform<?, ?>, Integer>, PartitionField> requiredClusteringFields(PartitionSpec spec) {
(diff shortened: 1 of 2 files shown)


rdblue:
FYI @bryanck and @ajreid21

aokolnychyi:
Let me take a look.

core/src/main/java/org/apache/iceberg/util/SortOrderUtil.java
```diff
@@ -74,6 +101,28 @@ public static SortOrder buildSortOrder(Schema schema, PartitionSpec spec, SortOr
     return builder.build();
   }
 
+  private static Map<Pair<Transform<?, ?>, Integer>, PartitionField> requiredClusteringFields(PartitionSpec spec) {
+    Map<Pair<Transform<?, ?>, Integer>, PartitionField> requiredClusteringFields = Maps.newLinkedHashMap();
+    for (PartitionField partField : spec.fields()) {
+      if (!partField.transform().toString().equals("void")) {
```
aokolnychyi, line 107: What's our preferred way to check if its a void transform? Here is what we have in `Partitioning`:

```
field.transform().equals(Transforms.alwaysNull())
```

core/src/main/java/org/apache/iceberg/util/SortOrderUtil.java
```diff
@@ -74,6 +101,28 @@ public static SortOrder buildSortOrder(Schema schema, PartitionSpec spec, SortOr
...
+        requiredClusteringFields.put(Pair.of(partField.transform(), partField.sourceId()), partField);
+      }
+    }
+
+    // remove any partition fields that are satisfied by another partition field, like days(ts) and hours(ts)
+    for (PartitionField partField : spec.fields()) {
+      for (PartitionField field : spec.fields()) {
```
aokolnychyi, line 114: nit: `field/otherField` or `partField/otherPartField` instead of `partField/field`?

aokolnychyi (approved):
It took a bit of time to wrap my head around this but this looks correct to me.
Thanks, @rdblue. It has been hanging open for a while.

core/src/main/java/org/apache/iceberg/util/SortOrderUtil.java
rdblue: That's probably better. I'll update to that.

core/src/main/java/org/apache/iceberg/util/SortOrderUtil.java
rdblue: Yeah, I can update. It was more for consistency with the other loop, but I think that got moved to a different method because of the complexity.

rdblue:
Thanks for reviewing, @aokolnychyi!

PR #5133, opened 2022-06-25
Core: Fix REST field name case strategy

rdblue:
This fixes serialization for `CreateTableRequest` and `LoadTableResponse`, which have multi-word fields. Those fields need to be serialized using `kebab-case` in Jackson rather than camelCase. This also updates the tests for `CreateTableRequest`.


danielcweeks (approved):

kbendick:
Thanks @rdblue!

I have a branch that will update all of the tests to use this. Let me push that and open it as well =)

kbendick:
I have a PR, https://github.com/apache/iceberg/pull/5135, that can replace this PR and also updates the base class `RequestResponseTestBase`, to always check the serialized fields.

RIght now, the base class is testing that the java object has the correctly named fields. It's just by chance that the names match up with the spec. Now it's enforced with https://github.com/apache/iceberg/pull/5135 cc @rdblue @danielcweeks

rdblue:
Thanks for reviewing, @kbendick and @danielcweeks!

PR #5135, opened 2022-06-25
Core: Fix REST field name case strategy and ensure field name tests are for serialized versions

kbendick:
Previously, we were testing whether or not each RESTCatalog request / response had the expected fields as named in the Java class.

However, we want to test that the correct names, as defined in the spec are used.

This PR updates:
- updates the request / response base test for `hasOnlyKnownFields` to check the serialized JSON
- updates the RESTObjectMapper to allow for kebab-case
- updates the CreateTableRequest tests to use `stage-create` for serialized versions, the name that is defined in the spec, vs `stageCreate`, the Java class formatted name.

This PR can replace https://github.com/apache/iceberg/pull/5133 entirely (or be merged in after that one and rebasing).


core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateNamespacePropertiesRequest.java
```diff
@@ -206,7 +206,7 @@ public void testBuilderDoesNotCreateInvalidObjects() {
 
   @Override
   public String[] allFieldsFromSpec() {
-    return new String[] { "updates", "removals" };
+    return new String[] { "removals", "updates" };
```
kbendick, line 209: Updated to be in the order defined by the spec... can revert this change if we don't find it necessary.

core/src/test/java/org/apache/iceberg/rest/responses/TestOAuthTokenResponse.java
```diff
@@ -67,24 +67,6 @@ public String serialize(OAuthTokenResponse response) throws JsonProcessingExcept
...
-        Assert.assertTrue("Should have field: " + field, node.has(field));
-      }
-
-      for (String field : ((Iterable<? extends String>) node::fieldNames)) {
-        Assert.assertTrue("Should not have field: " + field, fieldsFromSpec.contains(field));
-      }
-    } catch (JsonProcessingException e) {
-      throw new RuntimeException(e);
-    }
-  }
```
kbendick, line 86: This is now the version of the test that exists in the base class `RequestResponseTestBase`, and thus does not need to be overridden anymore.

kbendick:
cc @rdblue as this PR can replace https://github.com/apache/iceberg/pull/5133

core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateNamespacePropertiesRequest.java
rdblue: I think it would be best to revert this.

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
```diff
@@ -257,7 +255,7 @@ public void testBuilderDoesNotBuildInvalidRequests() {
 
   @Override
   public String[] allFieldsFromSpec() {
-    return new String[] {"name", "location", "schema", "spec", "order", "properties", "stageCreate"};
+    return new String[] {"name", "location", "schema", "spec", "order", "stage-create", "properties"};
```
rdblue, line 258: Since we caught that this wasn't correctly testing against the spec with `stageCreate`, I just [checked the spec](https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L1514-L1517) and it looks like there are more incorrect fields: `partition-spec` and `write-order`. Can you also fix those fields?

rdblue:
@kbendick, can you please update this since #5133 was merged and the other fields need to be corrected?

core/src/test/java/org/apache/iceberg/rest/requests/TestUpdateNamespacePropertiesRequest.java
kbendick: Reverted

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
kbendick: Sure

kbendick:
> @kbendick, can you please update this since #5133 was merged and the other fields need to be corrected?

Sure. Probably the easiest way to fix the other names is to rename the class fields. Will do that.

core/src/test/java/org/apache/iceberg/rest/requests/TestCreateTableRequest.java
kbendick: Fixed.

kbendick:
> > @kbendick, can you please update this since #5133 was merged and the other fields need to be corrected?
> 
> Sure. Probably the easiest way to fix the other names is to rename the class fields. Will do that.

This is done. If we merge this, then https://github.com/apache/iceberg/pull/5118 can be rebased and considered for merging afterwards.

kbendick:
There's a failing bloom filter related test in Java CI
```
org.apache.iceberg.parquet.TestBloomRowGroupFilter > testBytesEq FAILED
    java.lang.AssertionError: Should not read: cannot match a new generated binary
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertFalse(Assert.java:65)
        at org.apache.iceberg.parquet.TestBloomRowGroupFilter.testBytesEq(TestBloomRowGroupFilter.java:587)
```

rdblue (approved):

PR #5143, opened 2022-06-28
Spark: fix regression from scan refactor

bryanck:
A performance regression was introduced with the refactoring of the Spark batch classes. One problem that surfaced was that `BatchScanExec` does an equality check on the batch object. However with the refactoring, a new batch object was created for each call to `toBatch()` so the equality check returned false for the same scan instance (because the batch objects were different). The end result is that filters weren't being pushed down in some cases.

Creating only one batch object per scan partially fixed the performance problems, but other issues remained. So this PR makes the scan class implement the Batch interface as before the refactor, for now, until all of the issues can be addressed.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatch.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatch.java
@@ -20,6 +20,7 @@
 package org.apache.iceberg.spark.source;
 
 import java.util.List;
+import java.util.function.Supplier;
 import org.apache.iceberg.CombinedScanTask;
 import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.Schema;
@@ -43,17 +44,17 @@ class SparkBatch implements Batch {
   private final JavaSparkContext sparkContext;
   private final Table table;
   private final SparkReadConf readConf;
-  private final List<CombinedScanTask> tasks;
+  private final Supplier<List<CombinedScanTask>> taskSupplier;
   private final Schema expectedSchema;
   private final boolean caseSensitive;
   private final boolean localityEnabled;
 
   SparkBatch(JavaSparkContext sparkContext, Table table, SparkReadConf readConf,
-             List<CombinedScanTask> tasks, Schema expectedSchema) {
+             Supplier<List<CombinedScanTask>> taskSupplier, Schema expectedSchema) {
     this.sparkContext = sparkContext;
     this.table = table;
     this.readConf = readConf;
-    this.tasks = tasks;
+    this.taskSupplier = taskSupplier;
     this.expectedSchema = expectedSchema;
     this.caseSensitive = readConf.caseSensitive();
     this.localityEnabled = readConf.localityEnabled();
@@ -65,6 +66,7 @@ class SparkBatch implements Batch {
     Broadcast<Table> tableBroadcast = sparkContext.broadcast(SerializableTable.copyOf(table));
     String expectedSchemaString = SchemaParser.toJson(expectedSchema);
 
+    List<CombinedScanTask> tasks = taskSupplier.get();
     InputPartition[] readTasks = new InputPartition[tasks.size()];
 
     Tasks.range(readTasks.length)
@@ -83,16 +85,17 @@ class SparkBatch implements Batch {
   }
 
   private int batchSize() {
-    if (parquetOnly() && parquetBatchReadsEnabled()) {
+    List<CombinedScanTask> tasks = taskSupplier.get();
+    if (parquetOnly(tasks) && parquetBatchReadsEnabled()) {
       return readConf.parquetBatchSize();
-    } else if (orcOnly() && orcBatchReadsEnabled()) {
+    } else if (orcOnly(tasks) && orcBatchReadsEnabled(tasks)) {
       return readConf.orcBatchSize();
     } else {
       return 0;
     }
   }
 
-  private boolean parquetOnly() {
+  private boolean parquetOnly(List<CombinedScanTask> tasks) {
     return tasks.stream().allMatch(task -> !task.isDataTask() && onlyFileFormat(task, FileFormat.PARQUET));
   }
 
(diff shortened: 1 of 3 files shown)


spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatch.java
```diff
@@ -65,18 +63,24 @@ public InputPartition[] planInputPartitions() {
     Broadcast<Table> tableBroadcast = sparkContext.broadcast(SerializableTable.copyOf(table));
     String expectedSchemaString = SchemaParser.toJson(expectedSchema);
 
-    InputPartition[] readTasks = new InputPartition[tasks.size()];
+    InputPartition[] readTasks = new InputPartition[tasks().size()];
```
kbendick, line 66: Question: is there any concern with calling `tasks()` repeatedly like this?

zinking: ja, in some Batch like fileBatch, tasks are recomputed

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatch.java
```diff
@@ -65,18 +63,24 @@ public InputPartition[] planInputPartitions() {
...
 
     Tasks.range(readTasks.length)
         .stopOnFailure()
         .executeWith(localityEnabled ? ThreadPools.getWorkerPool() : null)
         .run(index -> readTasks[index] = new ReadTask(
-            tasks.get(index), tableBroadcast, expectedSchemaString,
+            tasks().get(index), tableBroadcast, expectedSchemaString,
```
singhpk234, line 72: [minor] task() is synchronized, and we might be accessing here with a pool, should remove synchronized from function def and make it like : 

```
private volatile List<CombinedScanTask> tasks = null

protected synchronized List<CombinedScanTask> tasks() {
    if (tasks == null) {
      synchronized(this) {
        if (tasks == null) {
            task = // assign stuff to task
         }
       }
    }
    return tasks;
  }
```

I think now we might also want to make `SparkFilesScan#task()` handle multi-threading.

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
```diff
@@ -101,16 +99,14 @@ protected List<Expression> filterExpressions() {
     return filterExpressions;
   }
 
-  protected abstract List<CombinedScanTask> tasks();
-
   @Override
   public Batch toBatch() {
-    return new SparkBatch(sparkContext, table, readConf, tasks(), expectedSchema);
+    return this;
```
singhpk234, line 104: Nice !, this is in sync with built in v2 file Sources: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/FileScan.scala#L197

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatch.java
bryanck: The subclasses all lazily instantiate tasks only once AFAIK

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatch.java
bryanck: It looks like we weren't synchronizing before in 0.13, but I can add that check if you think it's needed

singhpk234: I see, as per my understanding, was thinking, we would be requiring it, as we have a possibility to call tasks() via mt (as we were in a thread pool), Though I see now in L66 above we already would have called tasks() which  would have populated this `task` anyhow, since we have made the function itself [synchronized](https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkCopyOnWriteScan.java#L134), each of this access will now be sequential even for the case we have task as not null.

All in all would say I don't have a strong reason for it, would say it's just something that caught my eye :) !

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScan.java
bryanck: My suspicion is that there is some check that the scan implements Batch, or some other equality check someplace, that is required for filter pushdown.

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestExpireSnapshotsAction.java
```diff
@@ -1147,7 +1147,7 @@ public void testUseLocalIterator() {
       checkExpirationResults(1L, 0L, 0L, 1L, 2L, results);
 
       Assert.assertEquals("Expected total number of jobs with stream-results should match the expected number",
-          5L, jobsRunDuringStreamResults);
+          4L, jobsRunDuringStreamResults);
```
bryanck, line 1150: Spark is planning the scan more optimally by reusing BatchScanExecs, so there are only 4 tasks. However if anyone has more in depth knowledge of this assertion, it would be good to double check this is correct.

rdblue (approved):

rdblue:
@bryanck, do we need to port this to 3.3 as well?

rdblue:
Thanks, @bryanck! Looks great to me.

bryanck:
Yes we should add this to 3.3 also.

bryanck:
I'll create a PR for that...

bryanck:
The PR for Spark 3.3 is [here](https://github.com/apache/iceberg/pull/5155).

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/actions/TestExpireSnapshotsAction.java
antiii-tech: could you explain what the 4 task is? I'm really confused, thanks a lot

PR #5148, opened 2022-06-28
StreamingDelete becomes public class

gustavoatt:
Make `StreamingDelete` a public class so that it can be used on packages outside of `org.apache.iceberg`.

This is useful for whenever users want to subclass the default implementation for monitoring or to add custom snapshot properties.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/StreamingDelete.java
+++ b/core/src/main/java/org/apache/iceberg/StreamingDelete.java
@@ -27,7 +27,7 @@ import org.apache.iceberg.expressions.Expression;
  * <p>
  * This implementation will attempt to commit 5 times before throwing {@link CommitFailedException}.
  */
-class StreamingDelete extends MergingSnapshotProducer<DeleteFiles> implements DeleteFiles {
+public class StreamingDelete extends MergingSnapshotProducer<DeleteFiles> implements DeleteFiles {
   StreamingDelete(String tableName, TableOperations ops) {
     super(tableName, ops);
   }


rdblue:
@gustavoatt, to add custom snapshot properties, you should be able to call `set` on the public interface. As for monitoring, should we add that into the class rather than making this public?

gustavoatt:
@rdblue let me provide more context. What we are doing at Airbnb, is that we are trying to fully integrate Airflow partition sensing with Airflow tables. One problem we have encountered when doing this is keeping track of empty partitions added explicitly via the command:

```sql
INSERT OVERWRITE $TBL
PARTITION (parKey=partVal,...)
SELECT *
FROM $EMPTY_SELECTION
```

Currently there is no way for us to know whether such an insert took place for sensing purposes. We have decided to keep track of these by adding custom properties on `Snapshot::summary` via a subclass of `BaseOvewriteFiles` that captures the `overwriteByRowFilter` passed to find out whether we were inserting a single partition.

We want to do the same for `DeleteFiles` so that we know when an empty partition "gets deleted", if that makes sense. Unfortunately the default implementation of `DeleteFiles` is protected which means we cannot subclass it outside of the iceberg package.

Note that we are hoping to keep track of empty partitions without having to do any changes to the way we write to Spark tables, so injecting our custom `OverwriteFiles` and `DeleteFiles` via our custom catalog is the most transparent way we have found to keep track of these expressions without having to change user jobs code.

gustavoatt:
Thank you Ryan!

PR #5149, opened 2022-06-28
Python: Make the VoidTransform a singleton

Fokko:
We want to reuse VoidTransform since it doesn't carry any state


rdblue (approved):

rdblue:
Thanks, @Fokko!

PR #5150, opened 2022-06-28
 Spark Integration to read from Snapshot ref

namrathamyske:
Issue adressed: https://github.com/apache/iceberg/issues/3899

This PR provides a way to spark query using snapshot ref


namrathamyske:
@amogh-jahagirdar Took a pull from your PR. let me know if my commit https://github.com/apache/iceberg/pull/5150/commits/8132d205f895208f36344dcc8dcc2794052146ab looks good.

namrathamyske:
@hililiwei @rdblue Can someone pls review this ?

spark/v3.2/spark/src/test/java/org/apache/iceberg/spark/source/TestSnapshotSelection.java
```diff
@@ -226,4 +226,62 @@ public void testSnapshotSelectionBySnapshotIdAndTimestamp() throws IOException {
         .hasMessageContaining("Cannot specify both snapshot-id")
         .hasMessageContaining("and as-of-timestamp");
   }
+
+  @Test
+  public void testSnapshotSelectionByRef() throws IOException {
+    String tableLocation = temp.newFolder("iceberg-table").toString();
```
namrathamyske, line 232: Still need to polish these testcases!

spark/v3.1/spark/src/test/java/org/apache/iceberg/spark/source/TestSnapshotSelection.java
```diff
@@ -226,4 +226,62 @@ public void testSnapshotSelectionBySnapshotIdAndTimestamp() throws IOException {
         .hasMessageContaining("Cannot specify both snapshot-id")
         .hasMessageContaining("and as-of-timestamp");
   }
+
+  @Test
+  public void testSnapshotSelectionByRef() throws IOException {
+    String tableLocation = temp.newFolder("iceberg-table").toString();
```
namrathamyske, line 232: Still need to polish these testcases!

amogh-jahagirdar:
@namrathamyske Thanks for this! I think we should separate the API and the Spark integration changes. Also at the API level, I think it makes sense to separate useBranch and useTag, rather than having one useSnapshotRef , because branching can be combined with time travel, but tagging cannot be. Let me know what you think.

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -49,6 +49,16 @@ public interface TableScan extends Scan<TableScan, FileScanTask, CombinedScanTas
...
 
+  /**
+   * Create a new {@link TableScan} from this scan's configuration that will use the most recent
+   * snapshot as of the given snapshot ref.
+   *
+   * @param snapshotRef a snapshot Ref
+   * @return a new scan based on this with the given snapshot Ref
+   * @throws IllegalArgumentException if the snapshot cannot be found
+   */
+  TableScan useSnapshotRef(String snapshotRef);
```
rdblue, line 60: I think this this should be `useRef(String branchOrTagName)`. The term `SnapshotRef` is internal and I don't think it should be exposed.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -94,6 +94,13 @@ public TableScan useSnapshot(long scanSnapshotId) {
         tableOps(), table(), tableSchema(), context().useSnapshotId(scanSnapshotId));
   }
 
+  @Override
+  public TableScan useSnapshotRef(String snapshotRef) {
+    Preconditions.checkArgument(table().snapshot(snapshotRef) != null,
+        "Cannot find ref with name %s", snapshotRef);
+    return useSnapshot(table().snapshot(snapshotRef).snapshotId());
```
rdblue, line 101: The logic here needs to be a bit more complicated because you can time-travel within a branch:

```java
table.newScan().useRef("test-branch").asOfTime(t).planFiles()
```

`useSnapshot` is the most specific and can't be used with anything else. `useRef` with a tag name is the same way. `useRef` with a branch name is compatible with `asOfTime`, just like it is with `main` implicitly.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/SparkReadConf.java
```diff
@@ -88,6 +88,8 @@ public Long endSnapshotId() {
     return confParser.longConf().option(SparkReadOptions.END_SNAPSHOT_ID).parseOptional();
   }
 
+  public String snapshotRef() {return confParser.stringConf().option(SparkReadOptions.SNAPSHOT_REF).parseOptional();}
```
rdblue, line 91: Why is this on one line?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/SparkConfParser.java
```diff
@@ -132,6 +132,10 @@ public String parse() {
       Preconditions.checkArgument(defaultValue != null, "Default value cannot be null");
       return parse(Function.identity(), defaultValue);
     }
+
+    public String parseOptional() {
```
rdblue, line 136: Can you please make Spark changes to the 3.3 branch only and then backport once the initial commit is done?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/SparkReadOptions.java
```diff
@@ -35,6 +35,9 @@ private SparkReadOptions() {}
   // A timestamp in milliseconds; the snapshot used will be the snapshot current at this time.
   public static final String AS_OF_TIMESTAMP = "as-of-timestamp";
 
+  // Snapshot Ref of the table snapshot to read from
+  public static final String SNAPSHOT_REF = "snapshot-ref";
```
rdblue, line 39: Since this is user-facing, it should be two properties, `branch` and `tag`

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -61,22 +62,32 @@ class SparkBatchQueryScan extends SparkBatchScan {
 
     this.snapshotId = readConf.snapshotId();
     this.asOfTimestamp = readConf.asOfTimestamp();
+    this.snapshotRef = readConf.snapshotRef();
+
```
rdblue, line 66: Style: unnecessary whitespace.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -61,22 +62,32 @@ class SparkBatchQueryScan extends SparkBatchScan {
...
 
     if (snapshotId != null && asOfTimestamp != null) {
       throw new IllegalArgumentException(
           "Cannot scan using both snapshot-id and as-of-timestamp to select the table snapshot");
+    } else if (snapshotId != null && snapshotRef != null) {
+      throw new IllegalArgumentException(
+          "Cannot scan using both snapshot-id and snapshot-ref to select the table snapshot");
+    } else if(asOfTimestamp!= null && snapshotRef != null) {
+      throw new IllegalArgumentException(
+          "Cannot scan using both as-of-timestamp and snapshot-ref to select the table snapshot");
```
rdblue, line 76: This is possible. I also think we should probably let the core context handle throwing exceptions, rather than checking here.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -61,22 +62,32 @@ class SparkBatchQueryScan extends SparkBatchScan {
...
       if (startSnapshotId != null || endSnapshotId != null) {
         throw new IllegalArgumentException(
             "Cannot specify start-snapshot-id and end-snapshot-id to do incremental scan when either "
                 + SparkReadOptions.SNAPSHOT_ID
                 + " or "
                 + SparkReadOptions.AS_OF_TIMESTAMP
-                + " is specified");
+                + " or "
+                + SparkReadOptions.SNAPSHOT_REF
+                + " is specified ");
```
rdblue, line 90: Can you rewrite this as a format string?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
amogh-jahagirdar: +1 time travel is supported for branches. 

@rdblue @namrathamyske how do you feel about having a separate scan API useBranchAsOfTime https://github.com/apache/iceberg/pull/5364/files#diff-68f831bb319578a644bd1024f51d85a09284bb34ebc5d9b7608362b24caf441aR73 like in this PR? 

Pro:

Explicit API for time travel on a branch which seems more obvious to me than 

`scan.useBranch("branch").asOfTime(someTimeMs)`

Cons:

It's another API and perhaps some users find the builder more obvious. validation could be handled in the scan context as Ryan pointed out.

api/src/main/java/org/apache/iceberg/TableScan.java
amogh-jahagirdar: I think we need to separate the useBranch and useTag APIs. As you said, refs are internal. From a Spark user perspective we also want to only expose the branch/tag terms; imo I think the same case could be applied to the API level. Also considering branches can be combined with time travel we could do a separate API for that ; although there's an argument to be made to just combine useBranch + as Of Time.

namrathamyske:
@rdblue @amogh-jahagirdar Thanks for your review! working on changes for this!

api/src/main/java/org/apache/iceberg/TableScan.java
rdblue: Yeah, I considered that as well. The problem is that the caller doesn't know whether the ref is a tag or a branch before calling the method. That's determined when we look at table metadata and we don't want to force the caller to do that.

There may be a better name than "ref" for `useRef`. That seems like the problem to me. Maybe we could simplify it to `use`? I'm not sure that's obvious enough.

@aokolnychyi, do you have any thoughts on the name here?

namrathamyske: @rdblue @amogh-jahagirdar I agree that we can use a common API for tag or branch like `useRef`.

We have two signatures:

`useRef(String refName)`

`useRef(String refName, Long timeStampMillis)` -> will throw exception for tag type, since we cant do time travel for tag.

amogh-jahagirdar: Sure. The only thing is  I think if we do useRef (or if we come up with a better name) then we would not want to have the `useRef(String refName, Long timeStampMillis)`. A user would chain it with the existing useTimestamp and then the validation would happen in the scan context. I don't think we would want the extra method because time travel would only apply for branches so having useRef in that case doesn't make sense to me.

If we have consensus on this, then I can update https://github.com/apache/iceberg/pull/5364/files with the updated approach. Then this, PR could be focused on the Spark side of the integration

jackye1995: > will throw exception for tag type, since we cant do time travel for tag.

In that case I would suggest:

- useRef(String refName)
- useBranchAsOfTime(String branchName, Long timeStampMillis)

jackye1995: Oh I see the alternative is just `.useRef(refName).asOfTime(timestampMillis)`. That also works, in that case +1 for `useRef(String refName)`

rdblue: Sounds like there is consensus for `useRef`.

namrathamyske: @amogh-jahagirdar  After these changes are incorporated in https://github.com/apache/iceberg/pull/5364/files, Only then this PR can be merged.

rdblue:
@namrathamyske, it looks like some of these changes went in with #5364. Can you rebase on that?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -244,6 +248,13 @@ public Statistics estimateStatistics() {
       Snapshot snapshot = table().snapshot(snapshotIdAsOfTime);
       return estimateStatistics(snapshot);
 
+    } else if (branch != null) {
+      Snapshot snapshot = table().snapshot(branch);
+      return estimateStatistics(snapshot);
+
+    } else if (tag != null) {
+      Snapshot snapshot = table().snapshot(tag);
+      return estimateStatistics(snapshot);
```
rdblue, line 257: Nit: missing empty line at the end of the block like others have.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -182,13 +183,28 @@ private Schema schemaWithMetadataColumns() {
...
+        SparkReadOptions.TAG);
 
     Preconditions.checkArgument(
         snapshotId == null || asOfTimestamp == null,
         "Cannot set both %s and %s to select which table snapshot to scan",
         SparkReadOptions.SNAPSHOT_ID,
         SparkReadOptions.AS_OF_TIMESTAMP);
 
+    String snapshotRef = branch != null ? branch : tag;
+    Preconditions.checkArgument(
```
rdblue, line 202: I don't think that we need checks here. The `TableScan` will check for consistency and throw exceptions, so we should allow it to fail there rather than duplicating checks. We can probably remove both this and the one checking for both tag and branch set. We still need tests for this behavior though.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -225,6 +241,12 @@ public Scan build() {
       scan = scan.asOfTime(asOfTimestamp);
     }
 
+    if (branch != null) {
+      scan.useSnapshotRef(branch);
+    } else if (tag != null) {
```
rdblue, line 246: I think this should not use `else`. Just set the ref and let the `TableScan` handle ensuring there is only one call.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/source/TestSnapshotSelection.java
```diff
@@ -226,4 +226,88 @@ public void testSnapshotSelectionBySnapshotIdAndTimestamp() throws IOException {
         .hasMessageContaining("Cannot specify both snapshot-id")
         .hasMessageContaining("and as-of-timestamp");
   }
+
+  @Test
+  public void testSnapshotSelectionByTag() throws IOException {
```
rdblue, line 231: I think we also need tests to show that branch and tag options can't be used at the same time, and tests to validate what happens when snapshot or timestamp are set along with branch or tag. It should be easy to make a few failing tests.

rdblue (commented):
Thanks, @namrathamyske! This looks close.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkReadOptions.java
```diff
@@ -35,6 +35,12 @@ private SparkReadOptions() {}
   // A timestamp in milliseconds; the snapshot used will be the snapshot current at this time.
   public static final String AS_OF_TIMESTAMP = "as-of-timestamp";
 
+  // branch ref of the table snapshot to read from
+  public static final String BRANCH = "branch";
+
+  // tag ref of the table snapshot to read from
```
amogh-jahagirdar, line 41: Just a nit: I think for the comments we can leave off the "table snapshot" and "ref" part and the comment could look something like 

"Tag to read from"
"Branch to read from"

namrathamyske:
I came across duplicate implementations of using timestamp option in https://github.com/apache/iceberg/blob/5688d59f6b8ea4784de1f45c9386e13618803673/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java#L185 and https://github.com/apache/iceberg/blob/5688d59f6b8ea4784de1f45c9386e13618803673/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java#L629. I am unsure of how to proceed. Just making changes in SparkScanBuilder worked before for previous versions of spark - 3.1, 3.2. But for 3.3 , properties get overridden in SparkCatalog.java before reaching SparkScanBuilder.java. I am just unsure of how to proceed here or if both are different flows altogether. @rdblue @amogh-jahagirdar any thoughts appreciated.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -226,6 +222,12 @@ public Scan build() {
       scan = scan.asOfTime(asOfTimestamp);
     }
 
+    if (branch != null) {
+      scan.useRef(branch);
+    } else if (tag != null) {
+      scan.useRef(tag);
+    }
```
amogh-jahagirdar, line 229: I think the issue is that scan.useRef(tag) by itself doesn't update the snapshot in the context; it's a builder like pattern so it needs to be `scan = scan.useRef(ref)`. That's why the tests are failing; the context snapshot isn't set, so by default the scan reads main

amogh-jahagirdar (commented):
@namrathamyske I checked out your branch and debugged, see my comment for what the bug is; after the fix the tests should start passing.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
```diff
@@ -270,6 +282,8 @@ && readSchema().equals(that.readSchema())
         && Objects.equals(startSnapshotId, that.startSnapshotId)
         && Objects.equals(endSnapshotId, that.endSnapshotId)
         && Objects.equals(asOfTimestamp, that.asOfTimestamp);
+    //        && Objects.equals(branch, that.branch)
+    //        && Objects.equals(tag, that.tag);
   }
```
namrathamyske, line 287: have to uncomment this , but a checkstyle cyclomatic complexity error.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
namrathamyske: my bad missed this! Thanks for pointing out!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkBatchQueryScan.java
amogh-jahagirdar: Considering it's required for a correct equals implementation of SparkBatchQueryScan, I think it makes the most sense just to suppress the warnings on the method  `@SuppressWarnings("checkstyle:CyclomaticComplexity")`

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
amogh-jahagirdar: No problem!

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/source/TestSnapshotSelection.java
```diff
@@ -226,4 +226,148 @@ public void testSnapshotSelectionBySnapshotIdAndTimestamp() throws IOException {
...
+    List<SimpleRecord> currentSnapshotRecords =
+        currentSnapshotResult.orderBy("id").as(Encoders.bean(SimpleRecord.class)).collectAsList();
+    List<SimpleRecord> expectedRecords = Lists.newArrayList();
+    expectedRecords.addAll(firstBatchRecords);
+    Assert.assertEquals(
+        "Current snapshot rows should match", expectedRecords, currentSnapshotRecords);
+  }
+
+  @Test
+  public void testSnapshotSelectionByBranchAndTag() throws IOException {
```
amogh-jahagirdar, line 301: Nit: In test names if they are expected to fail I feel that it's nice to suffix the test name with "fails" so "testSnapshotSelectionByBranchAndTagFails".  Non blocking

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/source/TestSnapshotSelection.java
```diff
@@ -226,4 +226,148 @@ public void testSnapshotSelectionBySnapshotIdAndTimestamp() throws IOException {
...
+    table.manageSnapshots().createBranch("branch", table.currentSnapshot().snapshotId()).commit();
+
+    // produce the second snapshot
+    List<SimpleRecord> secondBatchRecords =
+        Lists.newArrayList(
+            new SimpleRecord(4, "d"), new SimpleRecord(5, "e"), new SimpleRecord(6, "f"));
+    Dataset<Row> secondDf = spark.createDataFrame(secondBatchRecords, SimpleRecord.class);
+    secondDf.select("id", "data").write().format("iceberg").mode("append").save(tableLocation);
+
+    // verify records in the current snapshot by tag
```
amogh-jahagirdar, line 289: Nit: Rename tag to branch in the comment

amogh-jahagirdar (approved):
Overall looks great to me, thanks for contributing this @namrathamyske ! cc @rdblue

rdblue:
> I am unsure of how to proceed for branches and tags usecase. Just making changes to read from branch/tag in SparkScanBuilder worked before for previous versions of spark - 3.1, 3.2. But for 3.3 , properties get overridden in SparkCatalog.java before reaching SparkScanBuilder.java

The current implementation looks fine to me. If those code paths are taken, it indicates that Spark was passed syntax like `TIMESTAMP AS OF '...'`, which should be incompatible with `branch` or `tag` options. This PR already implements that because the snapshot passed to `SparkTable` is added to options.

rdblue (approved):

rdblue:
This looks good to me. I'm rerunning CI since the failures don't look related to this.

PR #5151, opened 2022-06-28
Spec: Add string ↔ binary promotion

Fokko:
Inspired by Avro and Ryan: https://github.com/apache/iceberg/pull/5116#discussion_r907858442

Screenshot 2022-06-28 at 22 02 50

By promoting string to bytes back and forth, we can just read the string as bytes. I've added a test for Python, to make sure that the hashing functions are equal. It might be a bit overkill, but just to make sure. I don't think this is necessary for Java since it is already tested: https://github.com/apache/iceberg/blob/master/api/src/test/java/org/apache/iceberg/transforms/TestBucketing.java#L207-L216

### DIFF
--- a/format/spec.md
+++ b/format/spec.md
@@ -218,6 +218,8 @@ Valid type promotions are:
 * `int` to `long`
 * `float` to `double`
 * `decimal(P, S)` to `decimal(P', S)` if `P' > P` -- widen the precision of decimal types.
+* `bytes` to `string`
+* `string` to `bytes`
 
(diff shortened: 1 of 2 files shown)


format/spec.md
```diff
@@ -218,6 +218,8 @@ Valid type promotions are:
 * `int` to `long`
 * `float` to `double`
 * `decimal(P, S)` to `decimal(P', S)` if `P' > P` -- widen the precision of decimal types.
+* `bytes` to `string`
+* `string` to `bytes`
```
rdblue, line 222: We're going to need to ensure this is implemented in Java, and add a note about compatibility. This isn't a forward-compatible change.

PR #5154, opened 2022-06-28
API: Add java type to composite TypeID enums

dimas-b:
Previously TypeID.STRUCT, LIST and MAP exposed Void as their
java type. It seems reasonable to expose more specific java
types according to existing code patterns in Accessors and
Comparators and simplify their if/switch statements.

ManifestFileUtil has non-trivial logic around the value returned
from TypeID.javaClass() but it works only on PrimitiveType, so it
is not affected by this change.

BaseDataReader (for Spark) also has non-trivial logic for processing
TypeID.javaClass() in its convertConstant() method, but this change
appears to actually improve the processing of values in that case.

PartitionSpec appears to be unaffected by this change as it would
normally use TypeID.javaClass() only for primitive values. In any
case this change should not break its processing logic.

### DIFF
--- a/api/src/test/java/org/apache/iceberg/types/TestComparators.java
+++ b/api/src/test/java/org/apache/iceberg/types/TestComparators.java
@@ -1,0 +1,151 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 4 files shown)


api/src/test/java/org/apache/iceberg/types/TestComparators.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  @Test
+  public void testBoolean() {
+    assertComparesCorrectly(Comparators.forType(Types.BooleanType.get()), false, true);
+  }
+
+  @Test
+  public void testInt() {
+    assertComparesCorrectly(Comparators.forType(Types.IntegerType.get()), 0, 1);
```
nastra, line 48: nit: should we also have comparisons where both values are the same across all the test methods?

nastra (approved):

rdblue (approved):

rdblue:
Thanks, @dimas-b!

api/src/test/java/org/apache/iceberg/types/TestComparators.java
dimas-b: `assertComparesCorrectly` does that already... or did you mean something else?

PR #5159, opened 2022-06-29
Drop table's root folder if the table owns it

gaborkaszab:
Introducing a new Catalog level flag to indicate whether dropping a
table from HiveCatalog should also drop the whole location of the
table.

Testing:
  - Created a new test that verifies that the location of the table
    exists or not based on how the new flag is set.
  - Manually loaded the changed jars into Apache Impala and checked
    that setting the new flag will result the table location being
    dropped.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -345,4 +345,11 @@ public class TableProperties {
 
   public static final String UPSERT_ENABLED = "write.upsert.enabled";
   public static final boolean UPSERT_ENABLED_DEFAULT = false;
+
+  /**
+   * Table properties to be used when dropping a table. Contains a list of locations that are owned
+   * by the table and should also be dropped.
+   */
+  public static final String LOCATIONS_OWNED_BY_TABLE = "locations-owned-by-table";
+  public static final String LOCATIONS_OWNED_BY_TABLE_DEFAULT = null;
 }
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
@@ -40,6 +40,7 @@ import org.apache.iceberg.CatalogUtil;
 import org.apache.iceberg.ClientPool;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableOperations;
+import org.apache.iceberg.TableProperties;
 import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.SupportsNamespaces;
 import org.apache.iceberg.catalog.TableIdentifier;
@@ -48,6 +49,7 @@ import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.hadoop.HadoopFileIO;
 import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.io.SupportsPrefixOperations;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
@@ -177,20 +179,17 @@ public class HiveCatalog extends BaseMetastoreCatalog implements SupportsNamespa
     }
 
     try {
-      clients.run(
-          client -> {
-            client.dropTable(
-                database,
-                identifier.name(),
-                false /* do not delete data */,
-                false /* throw NoSuchObjectException if the table doesn't exist */);
-            return null;
-          });
-
+      clients.run(client -> {
+        client.dropTable(database, identifier.name(),
+            false /* do not delete data */,
+            false /* throw NoSuchObjectException if the table doesn't exist */);
+        return null;
+      });
       if (purge && lastMetadata != null) {
         CatalogUtil.dropTableData(ops.io(), lastMetadata);
-      }
 
+        dropLocationsFromTableProperty(identifier, lastMetadata);
+      }
       LOG.info("Dropped table: {}", identifier);
       return true;
 
@@ -204,6 +203,22 @@ public class HiveCatalog extends BaseMetastoreCatalog implements SupportsNamespa
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
       throw new RuntimeException("Interrupted in call to dropTable", e);
+    }
+  }
+
+  private void dropLocationsFromTableProperty(TableIdentifier identifier, TableMetadata metadata) {
+    if (fileIO instanceof SupportsPrefixOperations) {
+      String locationsOwnedParam = metadata.property(
+          TableProperties.LOCATIONS_OWNED_BY_TABLE,
+          String.valueOf(TableProperties.LOCATIONS_OWNED_BY_TABLE_DEFAULT));
+      if (locationsOwnedParam != null) {
+        String[] locations = locationsOwnedParam.split(",", 0);
+        for (String location : locations) {
+          ((SupportsPrefixOperations) fileIO).deletePrefix(location);
+
+          LOG.debug("Location {} deleted for table: {}", location, identifier);
+        }
+      }
     }
   }
 
--- a/mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
+++ b/mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
@@ -244,6 +244,62 @@ public class TestHiveIcebergStorageHandlerNoScan {
     }
   }
 
+  /**
+   * This test exercises the table property level 'locations-owned-by-table' setting to see if it has an effect on
+   * dropping folders associated with the table.
+   */
+  @Test
+  public void testDropTableFolderOwnershipForHiveCatalog()
+      throws TException, IOException, InterruptedException {
+    Assume.assumeTrue("Only relevant for HiveCatalog",
+        testTableType.equals(TestTables.TestTableType.HIVE_CATALOG));
+
+    String tableName = "customers";
+    testTables.createTable(shell, tableName, HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
+        PartitionSpec.unpartitioned(), FileFormat.PARQUET, null);
+    TableIdentifier identifier = TableIdentifier.of("default", tableName);
+
+    org.apache.hadoop.hive.metastore.api.Table hmsTable = shell.metastore().getTable("default", "customers");
+    Path hmsTableLocation = new Path(hmsTable.getSd().getLocation());
+
+    // Drop the Iceberg table using HiveCatalog with the location ownership turned off.
+    Properties dropProperties = new Properties();
+    dropProperties.put(Catalogs.NAME, identifier.toString());
+    dropProperties.put(Catalogs.LOCATION, hmsTableLocation);
+    Catalogs.dropTable(shell.getHiveConf(), dropProperties);
+
+    FileSystem fs = Util.getFs(hmsTableLocation, shell.getHiveConf());
+    Assert.assertTrue("Table root directory should remain when the table doesn't own its location",
+        fs.exists(hmsTableLocation));
+
+    // Re-create the table with setting a single owned location in the properties to drop the root folder.
+    Map<String, String> props = Maps.newHashMap();
+    props.put(TableProperties.LOCATIONS_OWNED_BY_TABLE, hmsTableLocation.toString());
+    testTables.createTable(shell, tableName, HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
+        PartitionSpec.unpartitioned(), FileFormat.PARQUET, null, props);
+
+    Catalogs.dropTable(shell.getHiveConf(), dropProperties);
+
+    // Check that now dropTable() also drops the root directory of the table.
+    Assert.assertTrue("Table root directory should be dropped",
+        !fs.exists(hmsTableLocation));
+
+    // Re-create the table with setting an invalid and a valid owned location in the properties to
+    // drop the root folder.
+    props = Maps.newHashMap();
+    StringBuilder locations = new StringBuilder();
+    locations.append("Some invalid location").append(",").append(hmsTableLocation.toString());
+    props.put(TableProperties.LOCATIONS_OWNED_BY_TABLE, locations.toString());
+    testTables.createTable(shell, tableName, HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
+        PartitionSpec.unpartitioned(), FileFormat.PARQUET, null, props);
+
+    Catalogs.dropTable(shell.getHiveConf(), dropProperties);
+
+    // Check that now dropTable() also drops the root directory of the table.
+    Assert.assertTrue("Table root directory should be dropped",
+        !fs.exists(hmsTableLocation));
+  }
+
   @Test
   public void testCreateDropTableNonDefaultCatalog() throws TException, InterruptedException {
     TableIdentifier identifier = TableIdentifier.of("default", "customers");
--- a/mr/src/test/java/org/apache/iceberg/mr/hive/TestTables.java
+++ b/mr/src/test/java/org/apache/iceberg/mr/hive/TestTables.java
@@ -214,34 +214,49 @@ abstract class TestTables {
       PartitionSpec spec,
       FileFormat fileFormat,
       List<Record> records) {
+    return createTable(shell, tableName, schema, spec, fileFormat, records, null);
+  }
+
+  /**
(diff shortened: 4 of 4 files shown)


api/src/main/java/org/apache/iceberg/io/FileIO.java
```diff
@@ -61,6 +61,13 @@ default void deleteFile(OutputFile file) {
     deleteFile(file.location());
   }
 
+  /**
+   * Delete a folder at a given path.
+   */
+  default void deleteFolder(String path) {
+    deleteFile(path);
+  }
+
```
pvary, line 70: Do we need this?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -171,9 +172,16 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
-
       if (purge && lastMetadata != null) {
         CatalogUtil.dropTableData(ops.io(), lastMetadata);
+
+        boolean tableOwnsLocation = PropertyUtil.propertyAsBoolean(catalogProperties,
+            CatalogProperties.DROP_FULL_TABLE_FOLDER, CatalogProperties.DROP_FULL_TABLE_FOLDER_DEFAULT);
+        if (tableOwnsLocation) {
+          fileIO.deleteFolder(lastMetadata.location());
+
+          LOG.info("Root directory deleted for table: {}", identifier);
```
pvary, line 183: Maybe debug level log should be enough

core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java
```diff
@@ -66,10 +66,19 @@ public OutputFile newOutputFile(String path) {
...
+  public void deleteFolder(String path) {
+    deleteImpl(path, true);
+  }
+
+  private void deleteImpl(String path, boolean recursive) {
     Path toDelete = new Path(path);
     FileSystem fs = Util.getFs(toDelete, hadoopConf.get());
     try {
-      fs.delete(toDelete, false /* not recursive */);
+      fs.delete(toDelete, recursive);
```
pvary, line 81: Is the directory empty here?
If not, could we first remove the files in it?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -171,9 +172,16 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
         return null;
       });
-
       if (purge && lastMetadata != null) {
         CatalogUtil.dropTableData(ops.io(), lastMetadata);
+
+        boolean tableOwnsLocation = PropertyUtil.propertyAsBoolean(catalogProperties,
+            CatalogProperties.DROP_FULL_TABLE_FOLDER, CatalogProperties.DROP_FULL_TABLE_FOLDER_DEFAULT);
+        if (tableOwnsLocation) {
+          fileIO.deleteFolder(lastMetadata.location());
```
pvary, line 181: Maybe we just remove the empty `data` and `metadata` folders, and then remove the table folder?

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
```diff
@@ -212,6 +216,46 @@ public void testCreateDropTable() throws TException, IOException, InterruptedExc
     }
   }
 
+  // This test exercises the catalog level 'drop-full-table-folder' setting to see if it has an effect on
+  // dropping the full table location or not.
```
pvary, line 220: Maybe a proper javadoc?

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
```diff
@@ -212,6 +216,46 @@ public void testCreateDropTable() throws TException, IOException, InterruptedExc
...
+  @Test
+  public void testDropTableFolderOwnershipForHiveCatalog()
```
pvary, line 222: Maybe another test case when we do not drop the dirs?

pvary: If it is not tested somewhere else?

api/src/main/java/org/apache/iceberg/io/FileIO.java
gaborkaszab: Currently only empty folders could be dropped as the 'isRecursive' flag is hardcoded to false in HadoopFileIO.
If we want to keep FileIO intact, we can then change new logic in HiveCatalog.dropTable() to check if the FileIO is of HadoopFileIO and use its interface to invoke dropFolder() (or could be named dropRecursive()) instead of through FileIO.

core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java
gaborkaszab: For that we would need to list the files in the directory so that we can drop them one-by-one. I'd go for simple invoke a recursive drop here.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
gaborkaszab: That would work, I also gave it a try. But we can't assume that only this table puts files into these folders, so if there is something else non-related to this table inside these tables than the delete operation would fail. I found it more errorproof to recursively drop the root folder of the table.

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
gaborkaszab: This test also covers when the setting is off and we don't drop the table dir.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
gaborkaszab: That's true. There is one thing I wasn't sure about. I saw some old issues (also related to dropTable) where the root table dir not only contained a data/ and metadata/ folder but also the different partitions had some own directories under data/.
I'm not sure this could still happen, and I didn't manage to repro such a case but if in fact there are tables with such a directory structure the proposed approach wouldn't be able to drop the root dir then.

gaborkaszab:
@pvary I uploaded a new patch where I got rid of the FileIO changes. In turn I had to check the type for 'fileIO' member of HiveCatalog so that I can call deleteRecusrive()
If you are not a fan of this approach we can simply call deleteFile() for data/, metadata/ and the table's location but this seems error-prone for me as we can't really guarantee that those 2 folders will be the only content of a table's directory. A recursive delete seems simpler and error-proof.

rdblue:
I think we need to formalize the idea of ownership. What paths does a table own and how do we know that? I think we should probably have something in the table metadata to track this.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
pvary: For the record: we should check partitioned tables as well. I expect that the partition directories will remain as well

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
gaborkaszab: done

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
gaborkaszab: done

api/src/main/java/org/apache/iceberg/io/FileIO.java
pvary: Maybe `SupportsPrefixOperations.deletePrefix` could help us here.
See the `HadoopFileIO.deletePrefix` implementation.

gaborkaszab: Aaah, this code is relatively new, I didn't have it in my local copy. HadoopFileIO.deletePrefix() is actually the same I implemented now, so apparently someone also needed this but was a bit faster than me :)

gaborkaszab:
@rdblue I'm not sure I got your preference from you comment but I changed the code to have a table-level property instead of a catalog-level flag to control this functionality. I hope this is what you had in mind.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -171,9 +172,17 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
       });
-
       if (purge && lastMetadata != null) {
         CatalogUtil.dropTableData(ops.io(), lastMetadata);
+
+        Boolean shouldDropFullTableFolder = Boolean.parseBoolean(lastMetadata.property(
+            TableProperties.DROP_FULL_TABLE_FOLDER,
+            String.valueOf(TableProperties.DROP_FULL_TABLE_FOLDER_DEFAULT)));
+        if (shouldDropFullTableFolder) {
+          ((HadoopFileIO) fileIO).deletePrefix(lastMetadata.location());
```
pvary, line 182: I this is not something we can build upon.
See:
```
    this.fileIO = fileIOImpl == null ? new HadoopFileIO(conf) : CatalogUtil.loadFileIO(fileIOImpl, properties, conf);
```

I would do something like this:
```
if (fileIO instanceof SupportsPrefixOperations) {
    ((SupportsPrefixOperations) fileIO).deletePrefix(lastMetadata.location());
}
```

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
```diff
@@ -112,6 +112,8 @@ public class TestHiveIcebergStorageHandlerNoScan {
 
   private static final Set<String> IGNORED_PARAMS = ImmutableSet.of("bucketing_version", "numFilesErasureCoded");
 
+  private static final String DEFAULT_CATALOG_NAME = "default_iceberg";
+
```
pvary, line 116: Is this needed?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -171,9 +172,17 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
             false /* throw NoSuchObjectException if the table doesn't exist */);
         return null;
       });
-
       if (purge && lastMetadata != null) {
         CatalogUtil.dropTableData(ops.io(), lastMetadata);
```
pvary, line 176: If we own the full table dir - can we expect that all of the files are in the given dir?
We might not need this `CatalogUtil.dropTableData(ops.io(), lastMetadata)` call when `shouldDropFullTableFolder` is `true`

mr/src/test/java/org/apache/iceberg/mr/hive/TestTables.java
```diff
@@ -189,16 +189,37 @@ public Table createTable(TestHiveShell shell, String tableName, Schema schema, F
...
+   * @param tableName The name of the test table
+   * @param schema The schema used for the table creation
+   * @param spec The partition specification for the table
+   * @param fileFormat The file format used for writing the data
+   * @param records The records with which the table is populated
+   * @param properties Additional TblProperties to pass to the create table statement
+   * @return The created table
+   * @throws IOException If there is an error writing data
+   */
+  public Table createTable(TestHiveShell shell, String tableName, Schema schema, PartitionSpec spec,
```
pvary, line 208: Could we base this change the already existing hive code?
https://github.com/apache/hive/blob/90428cc5f594bd0abb457e4e5c391007b2ad1cb8/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestTables.java#L338-L356

pvary:
@gaborkaszab I think the main question is if we set the table property for owning the full directory then do we expect that all of the files are in the given directory, and there is no files outside of this directory. If we accept/expect this then we can simplify the delete process

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
gaborkaszab: I think we can't expect that all of the files related to a table are in it's own directory. At least it's not enforced after setting the new tableproperty to keep all the files in metadata.location(). Additionally this tableproperty could be set and unset without checking the locations of the existing files.

mr/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStorageHandlerNoScan.java
gaborkaszab: Not anymore. Dropped.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
gaborkaszab: Indeed, thx!

mr/src/test/java/org/apache/iceberg/mr/hive/TestTables.java
gaborkaszab: You mean I should put this test framework change into the Hive repo as well?
I can do so, I think the timing would be better after this PR is merged. Unless if there is any restriction that changes like this should land in Hive first.

rdblue:
@gaborkaszab, by adding a table property you're defining ownership. The problem is, you're adding an ownership concept that we know is going to change (we will need multiple roots) and doing it with very specific support (as @pvary mentioned, the assumption is that the directory is owned AND that no files are outside of it). As I said, I think the solution is to add a real concept of location ownership rather than adding temporary work-arounds.

What about updating the spec to have a list of owned locations and using that to fix this?

gaborkaszab:
@rdblue thanks for the review!
I changed the patch to have a tableproperty that is a list of locations that the table owns, and that are dropped when the table is dropped. I'm looking forward to hear you opinion!

danielcweeks:
The concept of location ownership and catalog behaviors around it are something that I feel we need to be more specific about.  While this may work for specific uses of HiveCatalog, it very likely wouldn't be consistent across catalog implementations and even has some potential side-effects (e.g. renamed tables may end up sharing location and result in dropping data incorrectly).  

These behaviors should really be owned by the Catalog (not the table) and should be tied to specific guarantees of the catalog implementation (i.e. de-conflicted/unique locations). 

I agree we want to be able to support this behavior, but I don't feel like this approach is comprehensive enough to address the problem.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -177,20 +179,17 @@ public boolean dropTable(TableIdentifier identifier, boolean purge) {
...
-                false /* throw NoSuchObjectException if the table doesn't exist */);
-            return null;
-          });
-
+      clients.run(client -> {
+        client.dropTable(database, identifier.name(),
+            false /* do not delete data */,
+            false /* throw NoSuchObjectException if the table doesn't exist */);
+        return null;
+      });
```
wypoon, line 187: This seems to be a purely formatting change. It is unnecessary.

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -345,4 +345,11 @@ private TableProperties() {}
 
   public static final String UPSERT_ENABLED = "write.upsert.enabled";
   public static final boolean UPSERT_ENABLED_DEFAULT = false;
+
+  /**
+   * Table properties to be used when dropping a table. Contains a list of locations that are owned
+   * by the table and should also be dropped.
+   */
+  public static final String LOCATIONS_OWNED_BY_TABLE = "locations-owned-by-table";
```
wypoon, line 353: How would this property be populated? Does the user have to specify this explicitly per table? Or will a catalog auto-populate it?

wypoon (commented):
In the case of the HiveCatalog, my understanding is that each table will have a location: either the user specifies a location in the CREATE TABLE statement, or a default location in the warehouse is assigned (if location is not specified). If the catalog can guarantee that a location is only used by a table, then when the table is dropped and data removed, it is safe to remove that location (for file systems where that makes sense). In this case, I don't think that having another table property, locations-owned-by-table, is necessary. If the catalog is not able to guarantee that locations specified in a locations-owned-by-table property are used exclusively by the table, then having such a property is not useful anyway.
I believe this is what @danielcweeks means.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
gaborkaszab: Thanks @wypoon for your comment!
Yes, I meanwhile I also learnt that this PR was a bit naive and it involves a much higher level topic about Catalogs being responsible to guarantee that dropping table location is safe.
I think I'll close this PR, then.

PR #5160, opened 2022-06-29
Spark 3.3: [FOLLOWUP] Add TODO for TimeTravel via DataframeReader using timestamp in seconds and CatalogOptions

singhpk234:
### About the changes

- Document limitation of TT via DataframeReader using timestamp in seconds and CatalogOptions.

### Testing Done

- Add an UT (in ignored mode) which would pass only when change for https://issues.apache.org/jira/browse/SPARK-39633 is available.

cc @rdblue

### DIFF
--- a/core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
+++ b/core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
@@ -88,8 +88,4 @@ public class DateTimeUtil {
   public static String formatTimestampMillis(long millis) {
     return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC));
   }
-
-  public static String formatTimestampMillisWithLocalTime(long millis) {
-    return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.systemDefault()));
-  }
 }
--- a/core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
+++ b/core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
@@ -88,4 +88,8 @@ public class DateTimeUtil {
   public static String formatTimestampMillis(long millis) {
     return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC));
   }
+
+  public static String formatTimestampMillisWithLocalTime(long millis) {
+    return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.systemDefault()));
+  }
 }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
```diff
@@ -88,4 +88,8 @@ public static long microsFromTimestamptz(OffsetDateTime dateTime) {
   public static String formatTimestampMillis(long millis) {
     return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC));
   }
+
+  public static String formatTimestampMillisWithLocalTime(long millis) {
+    return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.systemDefault()));
```
rdblue, line 93: Iceberg does not produce zone-specific time formats. This should not go in a common class like `DateTimeUtil`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/IcebergSource.java
```diff
@@ -173,7 +174,19 @@ public Optional<String> extractTimeTravelVersion(CaseInsensitiveStringMap option
...
+    String timestampAsOf = PropertyUtil.propertyAsString(options, "timestampAsOf", null);
+    if (timestampAsOf == null) {
+      return Optional.empty();
+    }
+
+    try {
+      // timestamp provided should be at a seconds precision.
+      // TODO: remove once https://issues.apache.org/jira/browse/SPARK-39633 is resolved
+      long timestampAsOfAsLong = Long.parseLong(timestampAsOf);
+      return Optional.of(DateTimeUtil.formatTimestampMillisWithLocalTime(timestampAsOfAsLong * 1000));
```
rdblue, line 186: This needs to be handled by Spark. Iceberg should not translate to a zone-specific time because that is what Spark expects..

singhpk234: Makes sense, calling spark api's via iceberg will also not be ok i think then as this now indirectly puts the translation to iceberg (as which spark api to use). Should I  add a TODO in the code with the issue linking it and add the UT added in this PR in ignore as soon as we upgrade to 3.3.1 we will have it (The PR for [SPARK-39633](https://issues.apache.org/jira/browse/SPARK-39633) is merged in upstream). 

Meanwhile we have existing ways to TT via dataframe options we can specify [as-of-timestamp](https://iceberg.apache.org/docs/latest/spark-queries/#time-travel) in milliseconds.

rdblue: @singhpk234, sounds good to me.

rdblue:
Should we close this since it won't be fixed in Iceberg?

singhpk234:
Sure, I just thought it would nice to document this atleast, so that we don't miss this when we upgrade to 3.3.1, but I think we should be fine either ways !

Thanks @rdblue for the review :) !! I learned about timestamp handling in iceberg / spark from this.

PR #5162, opened 2022-06-29
[WIP][BUILD] Add Build Attributes to Generated Manifests for Jar Output

kbendick:
For debugging purposes, it would be beneficial to have the build information inside of each generated JAR.

This way, a JAR that is used on a cluster or within an application can be evaluated for its build SHA or true package name to help debug issues.

This adds some build metadata to the jars generated from projects in build.gradle and Spark jars etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -188,6 +188,15 @@ project(':iceberg-bundled-guava') {
   }
 
   jar {
+    manifest {
+      attributes(
+        "Implementation-Title": project.name,
+        "Implementation-Version": project.version,
+        "Build-Timestamp": java.time.Instant.now().toString(),
+        "Build-JDK": "${System.properties['java.version']} (${System.properties['java.vendor']} ${System.properties['java.vm.version']})",
+        "Build-OS": "${System.properties['os.name']} ${System.properties['os.arch']} ${System.properties['os.version']}"
+      )
+    }
     archiveClassifier = 'empty'
   }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -200,15 +200,6 @@ project(':iceberg-bundled-guava') {
   }
 
   jar {
-    manifest {
-      attributes(
-        "Implementation-Title": project.name,
-        "Implementation-Version": project.version,
-        "Build-Timestamp": java.time.Instant.now().toString(),
-        "Build-JDK": "${System.properties['java.version']} (${System.properties['java.vendor']} ${System.properties['java.vm.version']})",
-        "Build-OS": "${System.properties['os.name']} ${System.properties['os.arch']} ${System.properties['os.version']}"
-      )
-    }
     archiveClassifier = 'empty'
   }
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -149,6 +153,19 @@ subprojects {
...
+      attributes(
+        "Implementation-Title": project.name,
+        "Implementation-Version": project.version,
+        "Build-SHA": project.buildSha,
+        "Build-Timestamp": java.time.Instant.now().toString(),
+        "Build-JDK": "${System.properties['java.version']} (${System.properties['java.vendor']} ${System.properties['java.vm.version']})",
+        "Build-OS": "${System.properties['os.name']} ${System.properties['os.arch']} ${System.properties['os.version']}"
+      )
+    }
+  }
```
kbendick, line 168: This doesn't apply to entries in other `build.gradle` files unfortunately.

kbendick:
Example of what gets added using a `displayManifest` script I added.

```
$ ./displayManifest.groovy hive3-orc-bundle/build/libs/iceberg-hive3-orc-bundle-0.14.0-SNAPSHOT.jar
Build-SHA: 08d58f514ddfb2a5ccfb1667440e7c17889e71d2
Implementation-Title: iceberg-hive3-orc-bundle
Implementation-Version: 0.14.0-SNAPSHOT
Build-Timestamp: 2022-06-29T22:38:55.132Z
Build-OS: Mac OS X aarch64 12.2.1
Build-JDK: 1.8.0_332 (Amazon.com Inc. 25.332-b08)
Manifest-Version: 1.0
```

build.gradle
```diff
@@ -636,6 +653,10 @@ String getVersionFromFile() {
   return file('version.txt').text.trim()
 }
 
+String gitBuildSha() {
```
nastra, line 656: I don't think this is actually needed. We can just use `versionDetails().gitHash` from the `gradle-git-version` plugin

build.gradle
```diff
@@ -149,6 +153,19 @@ subprojects {
       exceptionFormat "full"
     }
   }
+
+  jar {
+    manifest {
+      attributes(
+        "Implementation-Title": project.name,
+        "Implementation-Version": project.version,
+        "Build-SHA": project.buildSha,
```
nastra, line 162: you can directly use `versionDetails().gitHash` here

build.gradle
```diff
@@ -149,6 +153,19 @@ subprojects {
       exceptionFormat "full"
     }
   }
+
```
nastra, line 156: placing this entire block under `subprojects` right after the `test` block in https://github.com/apache/iceberg/blob/b4ed29b2af16aa88b5da842c64a8cc3630583ada/build.gradle#L151 worked for me

nastra (commented):
added some comments. I think this can be simplified and centrally handled in the root `build.gradle`

PR #5166, opened 2022-06-30
Close resources underlying various classes

NeQuissimus:
The underlying objects implement `close()`, which should be called
when HiveCatalog is discarded.


core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java
```diff
@@ -73,8 +73,7 @@ public OutputFile newOutputFile(String path) {
   @Override
   public void deleteFile(String path) {
     Path toDelete = new Path(path);
-    FileSystem fs = Util.getFs(toDelete, hadoopConf.get());
-    try {
+    try (FileSystem fs = Util.getFs(toDelete, hadoopConf.get())) {
```
pvary, line 76: This will remove the FileSystem from the cache.
If we delete multiple files on S3, closing the FS every time will become costly

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -547,6 +547,11 @@ protected Map<String, String> properties() {
     return catalogProperties == null ? ImmutableMap.of() : catalogProperties;
   }
 
+  public void close() {
+    fileIO.close();
+    clients.close();
```
pvary, line 552: The current practice is that we use `CatalogUtil.loadCatalog` to create the Catalog every time when we load a table. Opening and closing the clientPool (HMSClients) every time would be a costly operation, and we would likely lose everything that we win by using a pool.

Basically the expectation is that we have a global catalog cache and a thin Catalog wrapper using it.

NeQuissimus: This is why I did not link this method into existing code paths and left it completely optional.
For users of `HiveCatalog` there needs to be a way to close the resources.

NeQuissimus: So when I have a `HiveCatalog`, I can just throw it away? This seems strange with resources inside it that seemingly need to have `close`.

pvary: Took some time to wrap my head around this, but that's the current situation

PR #5170, opened 2022-06-30
Python: Move Transforms to Pydantic

Fokko:
This is being used by the partition-spec.

### DIFF
--- a/python/pyiceberg/transforms.py
+++ b/python/pyiceberg/transforms.py
@@ -23,12 +23,14 @@ from functools import singledispatch
 from typing import (
     Any,
     Generic,
+    Literal,
     Optional,
     TypeVar,
 )
 from uuid import UUID
 
 import mmh3  # type: ignore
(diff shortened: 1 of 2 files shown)


python/pyiceberg/transforms.py
```diff
@@ -403,35 +408,38 @@ def _(value: Decimal, _width: int) -> Decimal:
...
 
     def can_transform(self, source: IcebergType) -> bool:
-        return self._type == source
+        return self._source_type == source
 
     def result_type(self, source: IcebergType) -> IcebergType:
         return StringType()
 
+    def __repr__(self) -> str:
+        return f"transforms.UnknownTransform(source_type={repr(self._source_type)}, transform={repr(self._transform)})"
```
rdblue, line 436: Some of the other `__repr__` implementations use factory methods, like `transforms.truncate`. Should we do the same here?

rdblue (approved):

rdblue:
Looks good overall. Thanks, @Fokko!

PR #5171, opened 2022-06-30
Python: Reenable mypy

Fokko:
Mypy was so quiet, that I got a bit suspicious.

The path to the config file was incorrect, therefore it was not working.

I fixed all the violations, but also ran into some other issues:

It looks like that open() returns an incompatible type:
```
python/tests/conftest.py:833: error: Incompatible return value type (got "BufferedWriter", expected "OutputStream")
python/tests/conftest.py:833: note: Following member(s) of "BufferedWriter" have conflicts:
python/tests/conftest.py:833: note:     Expected:
python/tests/conftest.py:833: note:         def write(self, b: bytes) -> None
python/tests/conftest.py:833: note:     Got:
python/tests/conftest.py:833: note:         def write(self, Union[bytes, Union[bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer]]) -> int
```

When fixing the type to the awkward signature, and converting `closed()` to `closed` (a property). I got the following:
```
python/tests/conftest.py:831: error: Only protocols that don't have non-method members can be used with issubclass()
python/tests/conftest.py:831: note: Protocol "OutputStream" has non-method member(s): closed
```

Which basically tells us that we can't include `closed` in the protocol because I had to change it to a property to conform it to `IOBase`.

My suggestion would be to get rid of the InputStream/OutputStream and just use the internal IOBase which is a bit richer, but we don't have to implement everything.

I also stumbled upon this when trying to integrate the Avro library, but this wasn't an issue in the end because we vendor'ed it because we replaced the Avro schema with the iceberg one. Therefore we just could tailor the interfaces to our liking.

WDYT?

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -828,9 +828,9 @@ class LocalOutputFile(OutputFile):
 
     def create(self, overwrite: bool = False) -> OutputStream:
         output_file = open(self._path, "wb" if overwrite else "xb")
-        if not isinstance(output_file, OutputStream):
-            raise TypeError("Object returned from LocalOutputFile.create(...) does not match the OutputStream protocol.")
-        return output_file
+        # if not issubclass(type(output_file), OutputStream):
+        #    raise TypeError("Object returned from LocalOutputFile.create(...) does not match the OutputStream protocol.")
+        return output_file  # type: ignore
 
 
 class LocalFileIO(FileIO):
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -828,9 +828,9 @@ class LocalOutputFile(OutputFile):
 
     def create(self, overwrite: bool = False) -> OutputStream:
         output_file = open(self._path, "wb" if overwrite else "xb")
-        # if not issubclass(type(output_file), OutputStream):
-        #    raise TypeError("Object returned from LocalOutputFile.create(...) does not match the OutputStream protocol.")
-        return output_file  # type: ignore
+        if not issubclass(type(output_file), OutputStream):
+            raise TypeError("Object returned from LocalOutputFile.create(...) does not match the OutputStream protocol.")
+        return output_file
 
 
 class LocalFileIO(FileIO):
--- a/python/tests/io/test_io_base.py
+++ b/python/tests/io/test_io_base.py
@@ -64,9 +64,9 @@ class LocalInputFile(InputFile):
 
     def open(self) -> InputStream:
         input_file = open(self.parsed_location.path, "rb")
-        if not isinstance(input_file, InputStream):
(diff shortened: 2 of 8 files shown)


python/tests/conftest.py
```diff
@@ -828,9 +828,9 @@ def to_input_file(self):
 
     def create(self, overwrite: bool = False) -> OutputStream:
         output_file = open(self._path, "wb" if overwrite else "xb")
-        if not isinstance(output_file, OutputStream):
-            raise TypeError("Object returned from LocalOutputFile.create(...) does not match the OutputStream protocol.")
-        return output_file
+        # if not issubclass(type(output_file), OutputStream):
```
samredai, line 831: Should we just remove these two commented lines?

python/tests/io/test_io_base.py
```diff
@@ -64,9 +64,9 @@ def exists(self):
 
     def open(self) -> InputStream:
         input_file = open(self.parsed_location.path, "rb")
-        if not isinstance(input_file, InputStream):
-            raise TypeError("Object returned from LocalInputFile.open() does not match the OutputStream protocol.")
-        return input_file
+        # if not isinstance(input_file, InputStream):
```
samredai, line 67: Same here, these can be removed right?

python/tests/conftest.py
Fokko: We actually need this line for mypy to assert that it is an outputstream. I now commented it out because it would never pass.

python/tests/io/test_io_base.py
Fokko: We actually need this line for mypy to assert that it is an outputstream. I now commented it out because it would never pass. By reinstating these lines, we can remove the `type: ignore` below

Fokko:
@kbendick the signature was suggested by mypy:

```python
python/tests/conftest.py:833: error: Incompatible return value type (got "BufferedWriter", expected "OutputStream")
python/tests/conftest.py:833: note: Following member(s) of "BufferedWriter" have conflicts:
python/tests/conftest.py:833: note:     Expected:
python/tests/conftest.py:833: note:         def write(self, b: bytes) -> None
python/tests/conftest.py:833: note:     Got:
python/tests/conftest.py:833: note:         def write(self, Union[bytes, Union[bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer]]) -> int
```

Turns out a less akward signature is also okay, the issue was in returning a `None` instead of an `int`.

@rdblue I've removed the `.closed` property, and I both had the update the `seek()` and `read()` signature by returning an `int` instead of a `None`.

python/tests/io/test_io_base.py
samredai: Ah I see, ok I missed that.

rdblue (approved):

rdblue:
Thanks, @Fokko!

PR #5177, opened 2022-07-01
Spec: Sort-order order-id is mandatory

Fokko:
Looking at the code, we expect this field:

https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/SortOrderParser.java#L154

And we always write it on the Java side.

@rdblue Can you confirm this? I saw that you recently did some work on this.

### DIFF
--- a/open-api/rest-catalog-open-api.yaml
+++ b/open-api/rest-catalog-open-api.yaml
@@ -1068,6 +1068,7 @@ components:
     SortOrder:
       type: object
       required:
+        - order-id
         - fields
       properties:
         order-id:


rdblue (approved):

rdblue:
We do require this in the parser and it is probably best if we require it all the time.

There are a couple of places where we don't technically need it because the order will be reassigned. For example, when interacting with a REST-based catalog, the catalog service will add the sort order to the current metadata and reassign the ID to ensure it doesn't conflict with concurrent operations. But it's always okay to have an order ID so I think we should go ahead and require it.

rdblue:
Thanks, @Fokko!

PR #5178, opened 2022-07-01
API: Add generic FileIO JSON serialization

rdblue:
This adds a JSON parser to serialize FileIO instances to and from JSON. The format is an object with an `io-impl` and configuration `properties`. Deserialization uses `CatalogUtil` to re-create an instance from an environment Hadoop Configuration (optional), the IO impl class, and the configuration used to create the original `FileIO`.

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
@@ -79,6 +79,7 @@ public class S3FileIO implements FileIO, SupportsBulkOperations, SupportsPrefixO
   private String credential = null;
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
+  private Map<String, String> properties = null;
   private transient volatile S3Client client;
   private MetricsContext metrics = MetricsContext.nullMetrics();
   private final AtomicBoolean isResourceClosed = new AtomicBoolean(false);
@@ -142,6 +143,11 @@ public class S3FileIO implements FileIO, SupportsBulkOperations, SupportsPrefixO
         DeleteObjectRequest.builder().bucket(location.bucket()).key(location.key()).build();
 
     client().deleteObject(deleteRequest);
+  }
+
+  @Override
+  public Map<String, String> properties() {
+    return properties;
   }
 
   /**
@@ -302,12 +308,13 @@ public class S3FileIO implements FileIO, SupportsBulkOperations, SupportsPrefixO
   }
 
   @Override
-  public void initialize(Map<String, String> properties) {
-    this.awsProperties = new AwsProperties(properties);
+  public void initialize(Map<String, String> props) {
+    this.awsProperties = new AwsProperties(props);
+    this.properties = props;
 
     // Do not override s3 client if it was provided
     if (s3 == null) {
-      AwsClientFactory clientFactory = AwsClientFactories.from(properties);
+      AwsClientFactory clientFactory = AwsClientFactories.from(props);
       if (clientFactory instanceof CredentialSupplier) {
         this.credential = ((CredentialSupplier) clientFactory).getCredential();
       }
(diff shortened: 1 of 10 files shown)


aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
```diff
@@ -79,6 +79,7 @@ public class S3FileIO implements FileIO, SupportsBulkOperations, SupportsPrefixO
   private String credential = null;
   private SerializableSupplier<S3Client> s3;
   private AwsProperties awsProperties;
+  private Map<String, String> properties = null;
```
danielcweeks, line 82: Originally we didn't want to include all properties as part of the serialized object (this would include all catalog properties), which is why AwsProperties is a slimmed down version.  Maybe we could trim this to just what the FileIO needs?

core/src/main/java/org/apache/iceberg/io/FileIOParser.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+import org.apache.iceberg.CatalogUtil;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.util.JsonUtil;
+
+public class FileIOParser {
+  private FileIOParser() {
+  }
+
+  private static final String FILE_IO_IMPL = "io-impl";
```
danielcweeks, line 34: I think we could reference `CatalogProperties.FILE_IO_IMPL` instead of redefining here.

danielcweeks: I guess that isn't really necessary since this is just the internal JSON representation.  Either way.

core/src/main/java/org/apache/iceberg/util/JsonUtil.java
```diff
@@ -50,6 +52,55 @@ public static ObjectMapper mapper() {
...
+   * Helper for writing JSON with a JsonGenerator.
+   *
+   * @param toJson a function to produce JSON using a JsonGenerator
+   * @param pretty whether to pretty-print JSON for readability
+   * @return a JSON string produced from the generator
+   */
+  public static String generate(ToJson toJson, boolean pretty) {
+    try {
+      StringWriter writer = new StringWriter();
+      JsonGenerator generator = JsonUtil.factory().createGenerator(writer);
```
danielcweeks, line 70: The JsonGenerator is closable, so we should probably be using TWR or closing with finally.

aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
danielcweeks: Ok, I think I understand why you're doing this now, which is basically that we need to re-initialize an S3/GCS FileIO (as opposed to the java/kryo serialization, which captures the aws/gcp properties for deserialization).  However, this pulls in far more than we would want to include.

We could have the `properties()` return the aws properties configuration:

```java
  @Override
  public Map<String, String> properties() {
    return awsProperties.toMap();
  }
```

This is a little painful for S3FileIO due to the number of properties, but I feel like we don't want to send all of the config properties (including sensitive properties) everywhere with the FileIO.

danielcweeks: I guess this would be necessary for any custom aws client factories as well, since they may need properties not enumerated.  

I guess this is probably the best we can do for now.

rdblue: Yeah, I debated this for quite a while, too. I think this is the cleanest way to do it for now.

core/src/main/java/org/apache/iceberg/io/FileIOParser.java
rdblue: Yeah, I copied it since this is internal. If we make a change to this, we don't want it to affect `CatalogProperties`.

danielcweeks (approved):

PR #5181, opened 2022-07-02
Core, API: RowDelta changes to branch Impl

namrathamyske:
issue addressed from: https://github.com/apache/iceberg/issues/3896

PR to address: https://github.com/apache/iceberg/pull/4926#issuecomment-1161499158

This PR implements the performing RowDelta operations on a particular branch ref. Throws unsupported operations with other operations.

PR continued from:
https://github.com/apache/iceberg/pull/4926
https://github.com/apache/iceberg/pull/5010


namrathamyske:
@amogh-jahagirdar @rdblue As per https://github.com/apache/iceberg/pull/4926#issuecomment-1161499158 I have changed validation checks when we commit to a branch. Right now I have implemented for BaseRowDelta usecase. My future plan is to extend for others. 

My thought process is every-time we check for validations, the start snapshot should start from an ancestor of the branch. If not we can throw a validation exception. Start snapshot should ideally never go before the snapshot with which branch was created. But we don't store the commit for which the branch was created in metadata. So i am quite confused how to go ahead. Any thoughts appreciated!

namrathamyske:
@rdblue @aokolnychyi @amogh-jahagirdar can you review this ?

PR #5184, opened 2022-07-03
Python: Pin bugbear in Python legacy

Fokko:
We got a new release on June 1st:
https://pypi.org/project/flake8-bugbear/22.7.1/

This introduced a new rule that actually broke the CI.

It is best to not do global lookups because of scoping and performance,
but I don't feel like rewriting the legacy code either.

### DIFF
--- a/python_legacy/tox.ini
+++ b/python_legacy/tox.ini
@@ -51,7 +51,7 @@ skip_install = true
 deps =
     flake8>=3.8.4
     flake8-import-order>=0.9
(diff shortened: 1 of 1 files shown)


samredai (approved):

kbendick (approved):
LGTM.

Was going to suggest leaving a comment but this is legacy CI so not a big deal.

danielcweeks (approved):

PR #5185, opened 2022-07-03
Spark: Parse sort orders with zorder expressions

rdblue:
This updates the `RewriteDataFilesProcedure` to directly call the parser to parse sort orders, instead of creating an ALTER TABLE statement and extracting the sort order.

This also makes a few updates to allow creating a `Term` from a zorder expression when parsed by Spark, so that the parser can pass back a zorder sort order representation (using `RawSortField` because `SortOrder` cannot contain a Zorder expression).


rdblue:
@ajantha-bhat, this demonstrates how to use the Spark extensions parser to parse a Zorder expression. This also updates the current procedure to use the parser directly. You can use this for your update to `rewrite_data_files` to support Zorder.

ajantha-bhat:
@rdblue : Thank you so much. I will update my PR with this change and add more test cases and a  co-authored by message.

api/src/main/java/org/apache/iceberg/expressions/Zorder.java
```diff
@@ -17,24 +17,19 @@
...
-import org.apache.iceberg.NullOrder
-import org.apache.iceberg.Schema
-import org.apache.iceberg.SortDirection
-import org.apache.iceberg.SortOrder
-import org.apache.iceberg.expressions.Term
+import java.util.Arrays;
+import java.util.List;
 
-class SortOrderParserUtil {
+public class Zorder implements Term {
```
ajantha-bhat, line 25: should this implement `UnboundTerm`?

Because `SortOrder` only accepts `UnboundTerm`. https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/SortOrder.java#L258

Also, Now `sortOrder` needs some extra array or list that says which of the fields are Zordered in the sortOrder?

ajantha-bhat:
I have updated my PR as per this.
https://github.com/apache/iceberg/pull/4902

api/src/main/java/org/apache/iceberg/expressions/Zorder.java
rdblue: No. `UnboundTerm` expects to have a single ref.

rdblue:
Closing this in favor of #4902.

PR #5186, opened 2022-07-03
[Build/Core] - Add a Gradle task to create a git version properties file for core module

kbendick:
This PR adds a gradle task `createProperties`, which generates a properties file that is intended for access from `org.apache.iceberg.util.VersionPropertiesUtil`.

The git build SHA (both short and full) and other useful information is accessible from that file.

This information is useful for debugging, and can also be sent as a header or request property with various catalogs for debugging catalog interactions with external resources.

The generated `version.properties` file is added to the runtime and test resources from within the `createProperties` task.

**If anybody knows a cleaner way to do that so as not to pollute the git sources, please let me know**


kbendick:
`version.properties` is added to core jar and thus runtime jars:
```bash
$ jar -tvf  spark/v3.3/spark-runtime/build/libs/iceberg-spark-runtime-3.3_2.12-0.14.0-SNAPSHOT.jar | grep version.properties
   291 Fri Feb 01 00:00:00 PST 1980 org/apache/iceberg/util/version.properties
   903 Fri Feb 01 00:00:00 PST 1980 org/apache/hc/client5/version.properties
   839 Fri Feb 01 00:00:00 PST 1980 org/apache/hc/core5/version.properties
```

`version.properties` is accessible at runtime:
```scala
$ spark-shell --jars spark/v3.3/spark-runtime/build/libs/iceberg-spark-runtime-3.3_2.12-0.14.0-SNAPSHOT.jar

scala> import org.apache.iceberg.util.VersionPropertiesUtil;
import org.apache.iceberg.util.VersionPropertiesUtil

scala> VersionPropertiesUtil.gitHashFull()
res0: String = aceda8990a62e0b36a895cb5f5d49c4ee8d82d6f

scala> VersionPropertiesUtil.gitHash()
res1: String = aceda8990a
```

nastra:
In general the approach makes sense to me. However, when looking at the spark-runtime jar I noticed that there's a `git.properties` file with a bunch of information being added (looks like this actually comes from the `apache-arrow` lib). Here's the content of that file:
```
#Generated by Git-Commit-Id-Plugin
#Sat Jan 29 05:11:45 UTC 2022
git.build.user.email=
git.build.host=Mac-1643432868411.local
git.dirty=true
git.remote.origin.url=https\://github.com/apache/arrow
git.closest.tag.name=apache-arrow-7.0.0
git.commit.id.describe-short=apache-arrow-7.0.0-0-dirty
git.commit.user.email=szucs.krisztian@gmail.com
git.commit.time=29.01.2022 @ 00\:08\:24 UTC
git.commit.message.full=[Release] Update versions for 7.0.0
git.build.version=7.0.0
git.commit.message.short=[Release] Update versions for 7.0.0
git.commit.id.abbrev=e90472e
git.branch=e90472e35b40f58b17d408438bb8de1641bfe6ef
git.build.user.name=
git.closest.tag.commit.count=0
git.commit.id.describe=apache-arrow-7.0.0-0-ge90472e-dirty
git.commit.id=e90472e35b40f58b17d408438bb8de1641bfe6ef
git.tags=apache-arrow-7.0.0
git.build.time=29.01.2022 @ 05\:11\:45 UTC
git.commit.user.name=Krisztián Szűcs
```

I looked at https://github.com/n0mer/gradle-git-properties and was wondering whether we could actually use that plugin to generate a `git.properties` file for Iceberg. It seems the plugin lets you control which info goes into that file. Also note that we could include the Iceberg version via a `customProperty`.

For completeness, here's how Arrow generates that file: https://github.com/apache/arrow/blob/master/java/pom.xml#L249-L290

kbendick:
> In general the approach makes sense to me. However, when looking at the spark-runtime jar I noticed that there's a `git.properties` file with a bunch of information being added (looks like this actually comes from the `apache-arrow` lib). Here's the content of that file:
> 
> ```
> #Generated by Git-Commit-Id-Plugin
> #Sat Jan 29 05:11:45 UTC 2022
> git.build.user.email=
> git.build.host=Mac-1643432868411.local
> git.dirty=true
> git.remote.origin.url=https\://github.com/apache/arrow
> git.closest.tag.name=apache-arrow-7.0.0
> git.commit.id.describe-short=apache-arrow-7.0.0-0-dirty
> git.commit.user.email=szucs.krisztian@gmail.com
> git.commit.time=29.01.2022 @ 00\:08\:24 UTC
> git.commit.message.full=[Release] Update versions for 7.0.0
> git.build.version=7.0.0
> git.commit.message.short=[Release] Update versions for 7.0.0
> git.commit.id.abbrev=e90472e
> git.branch=e90472e35b40f58b17d408438bb8de1641bfe6ef
> git.build.user.name=
> git.closest.tag.commit.count=0
> git.commit.id.describe=apache-arrow-7.0.0-0-ge90472e-dirty
> git.commit.id=e90472e35b40f58b17d408438bb8de1641bfe6ef
> git.tags=apache-arrow-7.0.0
> git.build.time=29.01.2022 @ 05\:11\:45 UTC
> git.commit.user.name=Krisztián Szűcs
> ```
> 
> I looked at https://github.com/n0mer/gradle-git-properties and was wondering whether we could actually use that plugin to generate a `git.properties` file for Iceberg. It seems the plugin lets you control which info goes into that file. Also note that we could include the Iceberg version via a `customProperty`.
> 
> For completeness, here's how Arrow generates that file: https://github.com/apache/arrow/blob/master/java/pom.xml#L249-L290

This is great. I'll try with this, as I can't find a clean way to deal with the additional file that's generated from the git tree.

rdblue:
@nastra, @kbendick, the Iceberg release process builds the binaries that are pushed to maven central from the source release tarball, where there is no git directory. To use a plugin like this, we will need to make sure we have some process for generating the file and probably putting it in the source tree as part of the process, like how we [add the `version.txt` file](https://github.com/apache/iceberg/blob/master/dev/source-release.sh#L90-L91).

Also, I couldn't tell from a brief look at the README, can that plugin produce a `git.properties` file that is not in the root of the Jar? We want to make sure we don't have a name conflict, so I'd prefer to have `org/apache/iceberg/git.properties` or something similar.

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
```diff
@@ -0,0 +1,108 @@
...
+  private static String projectVersion;
+
+  /**
+   * Loads the version.properties file for this module.
+   */
+  public static void loadBuildInfo() {
+    InputStream inputStream = null;
+
+    try {
+      inputStream = readResource(VERSION_PROPERTIES_FILE);
```
rdblue, line 55: This should use a try-with-resources block to ensure that it is closed.

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
```diff
@@ -0,0 +1,108 @@
...
+public class VersionPropertiesUtil {
+
+  private VersionPropertiesUtil() {
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(VersionPropertiesUtil.class);
+  private static final String VERSION_PROPERTIES_FILE = "version.properties";
+
+  private static boolean isLoaded = false;
+  private static Properties versionProperties = new Properties();
```
rdblue, line 41: Could this be a local variable?

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
```diff
@@ -0,0 +1,108 @@
...
+        versionProperties.load(inputStream);
+      } catch (Exception e) {
+        LOG.error("Could not load version.properties", e);
+      }
+    }
+
+    gitHash = versionProperties.getProperty("gitHash", "git-hash-not-found");
+    gitHashFull = versionProperties.getProperty("gitHashFull", "git-hash-full-not-found");
+    projectName = versionProperties.getProperty("projectName", "iceberg-core");
+    projectVersion = versionProperties.getProperty("projectVersion", "project-version-not-found");
```
rdblue, line 71: For defaults, how about a more friendly string, like "unknown"?

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
kbendick: Yeah. I made everything `static` because most things in `util` are static and because the generated properties file is relative to this path.

But this could be a local variable if we make `VersionPropertiesUtil` have a constructor etc. Maybe we should move it out of `util`?

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
kbendick: Updated to a constant of "uknown".

kbendick:
> @nastra, @kbendick, the Iceberg release process builds the binaries that are pushed to maven central from the source release tarball, where there is no git directory. To use a plugin like this, we will need to make sure we have some process for generating the file and probably putting it in the source tree as part of the process, like how we [add the `version.txt` file](https://github.com/apache/iceberg/blob/master/dev/source-release.sh#L90-L91).
> 
> Also, I couldn't tell from a brief look at the README, can that plugin produce a `git.properties` file that is not in the root of the Jar? We want to make sure we don't have a name conflict, so I'd prefer to have `org/apache/iceberg/git.properties` or something similar.

I'm not sure about the plugin. The way I'm doing it now definitely adds the file to the source tree (though I'd like to only do that for the release process). For now, I'm going to remove the `.gitignore` entry for this file and then possibly add steps in the release process to ensure `version.properties` is committed.

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
```diff
@@ -0,0 +1,99 @@
...
+public class VersionPropertiesUtil {
+
+  private VersionPropertiesUtil() {
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(VersionPropertiesUtil.class);
+  private static final String VERSION_PROPERTIES_FILE = "version.properties";
+  private static final String UNKNOWN_DEFAULT = "unknown";
+
+  private static boolean isLoaded = false;
```
danielcweeks, line 41: Even with double-check locking, I believe this should be marked `volatile` to prevent cache incoherence.

core/src/main/java/org/apache/iceberg/util/VersionPropertiesUtil.java
```diff
@@ -0,0 +1,99 @@
...
+      LOG.warn("Failed to load version properties from {}. Will use default values.", VERSION_PROPERTIES_FILE, e);
+    }
+
+    gitHash = versionProperties.getProperty("gitHash", UNKNOWN_DEFAULT);
+    gitHashFull = versionProperties.getProperty("gitHashFull", UNKNOWN_DEFAULT);
+    projectName = versionProperties.getProperty("projectName", UNKNOWN_DEFAULT);
+    projectVersion = versionProperties.getProperty("projectVersion", UNKNOWN_DEFAULT);
+  }
+
+  public static String gitHash() {
```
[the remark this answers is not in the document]
rdblue: With the current plan to add the properties file to the release tarball, we can include these all the time.

kbendick:
> @kbendick, can you either open a separate PR or update this one to only include the gradle task that drops a file into `core/build/iceberg-build.properties`? I'd like to get that reviewed quickly, without lots of unrelated things.

I saved everything in this branch elsewhere, but I’ll open a new PR with just the task.

nastra (approved):
LGTM

kbendick:
Closed in favor of https://github.com/apache/iceberg/pull/5228.

Might come back to this task to add into our manifests at a later date.

PR #5190, opened 2022-07-04
[API] Fix ErrorProne warnings in API module

kbendick:
The number of error prone warnings has slowly gotten somewhat higher.

This PR fixes all of the relatively simple error prone warnings for the API module, which will have API / ABI compatibility guarantees come the 1.0 release (and so the JavaDocs should hopefully be stable from then on).


kbendick:
The 1 remaining error prone warning in API module requires a bit more caution when / if we address it as quite a lot of code uses `ExceptionUtil`.

```
iceberg/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java:124: warning: [Finally] If you return or throw from a finally, then values returned or thrown from the try-catch block will be ignored. Consider using try-with-resources instead.
            throw new RuntimeException("Unknown exception in finally block", failure);
            ^
    (see https://errorprone.info/bugpattern/Finally)
```

dimas-b:
Re: https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java#L124 it does look like the most recent exception (`e` in code) is lost. 

Perhaps that line should do `throw new RuntimeException("Unknown exception in finally block", e)` (since `failure` is certainly `null` there).

kbendick:
Looks like this can be closed as these were fixed in: https://github.com/apache/iceberg/pull/5200

Thanks all!

kbendick:
> Looks like this can be closed as these were fixed in: #5200
> 
> Thanks all!

My mistake. These two were not fixed in #5200. I'll switch from `failure` to `e` as mentioned by @nastra.

In master, I still see these warnings:
```
> Task :iceberg-api:compileJava
/Users/kylebendickson/repos/iceberg/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java:124: warning: [Finally] If you return or throw from a finally, then values returned or thrown from the try-catch block will be ignored. Consider using try-with-resources instead.
            throw new RuntimeException("Unknown exception in finally block", failure);
            ^
    (see https://errorprone.info/bugpattern/Finally)
/Users/kylebendickson/repos/iceberg/api/src/main/java/org/apache/iceberg/io/FileIO.java:65: warning: [MissingSummary] A summary line is required on public/protected Javadocs.
   * @return the property map used to configure this FileIO
     ^
    (see https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment)
/Users/kylebendickson/repos/iceberg/api/src/main/java/org/apache/iceberg/io/CredentialSupplier.java:31: warning: [MissingSummary] A summary fragment is required; consider using the value of the @return block as a summary fragment instead.
   * @return the credential string
     ^
    (see https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment)
  Did you mean '*Returns the credential string.'?
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
3 warnings
```

kbendick:
Switching from throwing `e` instead of `finally` still gives the exception message. But I agree that at line 124, we know that `failure` is `null` so we should probably throw `e` instead as it's likely non-null.

And we're trying to throw `e` in the 4 lines above 127 (as we're not suppressing there): https://github.com/apache/iceberg/blob/d563d6db0a56199306c82187027fb76e2bf41be2/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java#L117-L124

So it looks like we need to change to throwing `e` at line 127 (where we've tried throwing `e` as several different types in the preceding lines) and then also add the suppression. WDYT?

api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java
```diff
@@ -116,12 +116,17 @@ public static <R, E1 extends Exception, E2 extends Exception, E3 extends Excepti
           LOG.warn("Suppressing failure in finally block", e);
           if (failure != null) {
             failure.addSuppressed(e);
+            tryThrowAs(failure, e1Class);
+            tryThrowAs(failure, e2Class);
+            tryThrowAs(failure, e3Class);
+            tryThrowAs(failure, RuntimeException.class);
+            throw new RuntimeException("Unknown throwable", failure);
```
kbendick, line 123: This seems like it will fix what the warning is about, that throwing from a finally block will swallow whatever is returned or thrown inside of the try block.

So this repeats lines 105-109, where we attempt to throw `failure` after running the catch block. As the warning message from error prone says that it will get swallowed if we throw inside of the finally, but we did try to throw it inside of the catch block.

Though I think this also changes the semantics of the function as intended. So maybe we should just add the suppression?

kbendick:
> If there are multiple exceptions, then I think we should add the one that isn't thrown as a suppressed exception on the one that we do throw.

This seems to happen already on line 118, but we're not rethrowing it from within the finally block (which is what the warning message is mentioning if I'm not mistaken): 

https://github.com/apache/iceberg/blob/d563d6db0a56199306c82187027fb76e2bf41be2/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java#L118

kbendick:
You know I just noticed that _none_ of these `runSafely` functions are used anywhere other than `TestExceptionUtil`.

kbendick:
I reverted the changes to `ExceptionUtils` and opened a PR to deprecate the unused functions that are giving the warnings here https://github.com/apache/iceberg/pull/5205

nastra (approved):

dimas-b (approved):

PR #5192, opened 2022-07-04
Python: Move PartitionSpec to Pydantic

Fokko:
~~Blocked by https://github.com/apache/iceberg/pull/5124~~

~~This also changes the logic a bit. Instead of passing the schema when initializing the PartitionSpec, we pass it to the `.bind(schema: Schema)` function. This will check the source fields, and initialize the transforms.~~

PartitionSpec as Pydantic for easy json (de)serialization and validation

Rebased against the latest master. Removed the bind method that we solely used to check if the column is there. Probably we have to move this down the line.


python/pyiceberg/transforms.py
```diff
@@ -91,7 +91,7 @@ def validate(cls, v: Any):
             elif v.startswith(BUCKET):
                 return BucketTransform(num_buckets=BUCKET_PARSER.match(v))
             elif v.startswith(TRUNCATE):
-                return TruncateTransform(width=BUCKET_PARSER.match(v))
+                return TruncateTransform(width=TRUNCATE_PARSER.match(v))
```
Fokko, line 94: Oops!

Fokko:
@rdblue Feel like some more Python?

python/pyiceberg/table/partitioning.py
```diff
@@ -57,24 +81,33 @@ class PartitionSpec:
...
-                raise ValueError(f"Cannot find source column: {partition_field.source_id}")
-            existing = source_id_to_fields_map.get(partition_field.source_id, [])
-            existing.append(partition_field)
-            source_id_to_fields_map[partition_field.source_id] = existing
-        object.__setattr__(self, "source_id_to_fields_map", source_id_to_fields_map)
-
-    def __eq__(self, other):
+    spec_id: int = Field(alias="spec-id")
+    fields: Tuple[PartitionField, ...] = Field()
+    last_assigned_field_id: int = Field(alias="last-assigned-field-id", default=_PARTITION_DATA_ID_START)
```
rdblue, line 86: We probably don't need this. It is just a high watermark of the field IDs in this spec. We can just define a property that returns it:

```python
    @property
    def last_assigned_field_id(self):
        return max(pf.field_id for pf in self.fields)
```

python/tests/table/test_metadata.py
```diff
@@ -205,13 +207,13 @@ def test_updating_metadata():
...
...ation", "table-uuid": "d20125c8-7284-442c-9aea-15fee620737c", "last-updated-ms": 1602638573874, "last-column-id": 3, "schemas": [{"fields": [{"id": 1, "name": "x", "type": "long", "required": true}, {"id": 2, "name": "y", "type": "long", "required": true, "doc": "comment"}, {"id": 3, "name": "z", "type": "long", "required": true}], "schema-id": 0, "identifier-field-ids": []}], "current-schema-id": 0, "partition-specs": [{"spec-id": 0, "fields": [{"source-id": 1, "field-id": 1000, "transform": "identity", "name": "x"}], "last-assigned-field-id": 1000}], "default-spec-id": 0, "last-partition-id": 1000, "properties": {}, "snapshots": [{"snapshot-id": 1925, "timestamp-ms": 1602638573822}], "snapshot-log": [], "metadata-log": [], "sort-orders": [{"order-id": 0, "fields": []}], "default-sort-order-id": 0, "refs": {}, "format-version": 1, "schema": {"fields": [{"id": 1, "name": "x", "type": "long", "required": true}, {"id": 2, "name": "y", "type": "long", "required": true, "doc": "comment"}, {"id": 3, "name": "z", "type": "long", "required": true}], "schema-id": 0, "identifier-field-ids": []}, "partition-spec": [{"name": "x", "transform": "identity", "source-id": 1, "field-id": 1000}]}"""
```
rdblue, line 210: Serialization should not include `last-partition-id`.

Fokko:
@rdblue thanks, I've removed it as a field, and added it as a property:

```python
    @property
    def last_assigned_field_id(self) -> int:
        if self.fields:
            return max(pf.field_id for pf in self.fields)
        return _PARTITION_DATA_ID_START
```

rdblue (approved):

PR #5194, opened 2022-07-04
Flink: Port #3998 to Flink 1.13 & 1.14

hililiwei:
Port #3998 to Flink 1.13 & 1.14


kbendick:
Hi @hililiwei! Can you rebase this off of latest master so its easier to diff?

I think I found some files that might not have been backported to 1.13, but I'm not sure if they're related to this PR or not (I don't think so but it would be nice to check).

PR #5195, opened 2022-07-04
Make predicates of delete only initialize once

shidayang:
DeleteFilter initialize 'delete rows' multiple times  when call DeleteFilter#filter method multiple times,  This cause performance of MOR is very low in trino.
In my case, It spend 8 minutes when I query a table, after optimize it only spend 20 seconds.

### DIFF
--- a/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
+++ b/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
@@ -70,7 +70,8 @@ public abstract class DeleteFilter<T> {
   private final int isDeletedColumnPosition;
 
   private PositionDeleteIndex deleteRowPositions = null;
-  private List<Predicate<T>> eqDeletePredicates = null;
+  private List<Predicate<T>> isInDeleteSets = null;
+  private Predicate<T> eqDeleteRows = null;
 
   protected DeleteFilter(String filePath, List<DeleteFile> deletes, Schema tableSchema, Schema requestedSchema) {
     this.setFilterThreshold = DEFAULT_SET_FILTER_THRESHOLD;
@@ -132,12 +133,12 @@ public abstract class DeleteFilter<T> {
   }
 
   private List<Predicate<T>> applyEqDeletes() {
-    if (eqDeletePredicates != null) {
-      return eqDeletePredicates;
+    if (isInDeleteSets != null) {
+      return isInDeleteSets;
     }
-    List<Predicate<T>> isInDeleteSets = Lists.newArrayList();
+
+    isInDeleteSets = Lists.newArrayList();
     if (eqDeletes.isEmpty()) {
-      this.eqDeletePredicates = isInDeleteSets;
       return isInDeleteSets;
     }
 
@@ -170,7 +171,7 @@ public abstract class DeleteFilter<T> {
       Predicate<T> isInDeleteSet = record -> deleteSet.contains(projectRow.wrap(asStructLike(record)));
       isInDeleteSets.add(isInDeleteSet);
     }
-    this.eqDeletePredicates = isInDeleteSets;
+
     return isInDeleteSets;
   }
 
@@ -196,10 +197,13 @@ public abstract class DeleteFilter<T> {
   }
 
   public Predicate<T> eqDeletedRowFilter() {
-    return applyEqDeletes().stream()
-            .map(Predicate::negate)
-            .reduce(Predicate::and)
-            .orElse(t -> true);
+    if (eqDeleteRows == null) {
+      eqDeleteRows = applyEqDeletes().stream()
+          .map(Predicate::negate)
+          .reduce(Predicate::and)
+          .orElse(t -> true);
+    }
+    return eqDeleteRows;
   }
 
   public PositionDeleteIndex deletedRowPositions() {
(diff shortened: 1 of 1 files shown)


data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -192,13 +196,10 @@ protected void markRowDeleted(T item) {
   }
 
   public Predicate<T> eqDeletedRowFilter() {
-    if (eqDeleteRows == null) {
```
chenjunjiedada, line 195: Do we need to change this?

shidayang:
@chenjunjiedada In trino query engine every "page"  will call DeleteFilter#filter, The "The DeleteFilter#filter is called once for each scan task" is not guaranteed。

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
shidayang: Yes,The function to avoid initialize 'delete row' of eqDeleteRows is the same as eqDeletePredicates.

chenjunjiedada (approved):
LGTM

lhofhansl:
Note that the same can be done in Trino without any API change in Iceberg. See https://github.com/trinodb/trino/pull/13112.
(But I do not care where we fix it.)

Always caching delete filters might be detrimental to Spark, which does not require the filters to be cached.

amogh-jahagirdar:
@lhofhansl For my understanding, could you elaborate on the cases where caching delete filters might be detrimental to Spark? Do you mean in terms of retaining the filters in memory?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -70,7 +70,7 @@
   private final int isDeletedColumnPosition;
 
   private PositionDeleteIndex deleteRowPositions = null;
-  private Predicate<T> eqDeleteRows = null;
```
rdblue, line 73: I don't see a need to remove this. Can you add it back?

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -132,8 +132,12 @@ public CloseableIterable<T> filter(CloseableIterable<T> records) {
   }
 
   private List<Predicate<T>> applyEqDeletes() {
+    if (eqDeletePredicates != null) {
+      return eqDeletePredicates;
+    }
```
rdblue, line 137: Please add an empty newline after this block to follow style guidelines.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
rdblue: I don't think this needs to change. The unified predicate and the list of predicates can coexist.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -192,13 +196,10 @@ protected void markRowDeleted(T item) {
...
-      eqDeleteRows = applyEqDeletes().stream()
-          .map(Predicate::negate)
-          .reduce(Predicate::and)
-          .orElse(t -> true);
-    }
-    return eqDeleteRows;
+    return applyEqDeletes().stream()
+            .map(Predicate::negate)
+            .reduce(Predicate::and)
+            .orElse(t -> true);
```
rdblue, line 202: Please fix indentation. Continuation indents are 4 spaces. Rolling back this change should fix it.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -132,8 +132,12 @@ public CloseableIterable<T> filter(CloseableIterable<T> records) {
   }
 
   private List<Predicate<T>> applyEqDeletes() {
+    if (eqDeletePredicates != null) {
+      return eqDeletePredicates;
+    }
     List<Predicate<T>> isInDeleteSets = Lists.newArrayList();
     if (eqDeletes.isEmpty()) {
+      this.eqDeletePredicates = isInDeleteSets;
```
rdblue, line 140: I think a better solution is to make `isInDeleteSets` a private field in place of `eqDeletePredicates`. You should just rename the new `eqDeletePredicates` field to `isInDeleteSets` and set it on the line above instead of creating a local variable.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
shidayang: I get you

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
shidayang: ok

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
shidayang: Bacause if encounter some exception when loading delete file, it is difficult to judge weather loaded delete file complete when next retry

lhofhansl:
> could you elaborate on the cases where caching delete filters might be detrimental to Spark? Do you mean in terms of retaining the filters in memory?

Yep. We'd be losing out on the opportunity to not materialize the complete filter as a set in memory.
(see: https://github.com/apache/iceberg/blob/6cc4a198c56c05ff103e6ecdf75fe50004af19da/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java#L231)

An alternative is to add the same API I added to TrinoDeleteFilter (which extends DeleteFilter<TrinoRow>) in the Trino PR, namely a `filterCached` method (could call it cacheAndFilter as well). That way we can leave the Spark code in place and Trino can use the caching method.

shidayang:
Thanks for your review, @rdblue

PR #5198, opened 2022-07-04
Arrow: Fix for dictionary encoded fixed length binary decimals

bryanck:
This PR fixes the vectorized reader for decimals that are fixed length binary and dictionary-encoded. Before these decimals would be downcast to 8 byte (long) precision. This only affects Parquet V2, as fixed length binary decimals aren't dictionary-encoded in Parquet V1. The Spark vectorized reader test for Parquet V2 was modified so this is now tested.

Also included is a minor refactor of the big endian padding to share some common logic and make it easier to enhance later.

In addition, the Spark benchmark test for the vectorized reader was enhanced to include both long encoded and fixed width binary encoded decimals.


rdblue (approved):

rdblue:
This looks good to me. Thanks for fixing it, @bryanck!

PR #5199, opened 2022-07-04
REST: Create commit catalog handler fix

bryanck:
Currently in the REST catalog handler, during a CTAS, the files will be written and associated with the partition spec ID from the initial create request. When the commit occurs, an empty table metadata object is created and updates are applied to that. However, applying the updates will increment the partition spec ID, so the files will be associated with the empty spec ID, and thus partition pruning will not be available for these files.

This PR changes the create commit so that the initial table metadata is constructed from the commit request information rather than using an empty table metadata object.

It appears this only impacts the tests for now.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
+++ b/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
@@ -263,6 +263,7 @@ public class CatalogHandlers {
     } else {
       finalMetadata = commitUpdate(catalog, ident, request);
     }
+
     return LoadTableResponse.builder()
         .withTableMetadata(finalMetadata)
         .build();
@@ -296,15 +297,19 @@ public class CatalogHandlers {
   private static TableMetadata commitCreate(Catalog catalog, TableIdentifier ident, UpdateTableRequest request) {
     // this is a hacky way to get TableOperations for an uncommitted table
     Transaction transaction = catalog.buildTable(ident, EMPTY_SCHEMA).createOrReplaceTransaction();
-    if (!(transaction instanceof BaseTransaction)) {
-      throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTransaction");
-    }
+    Preconditions.checkState(transaction instanceof BaseTransaction,
+        "Cannot wrap catalog that does not produce BaseTransaction");
     TableOperations ops = ((BaseTransaction) transaction).underlyingOps();
 
     // the only valid requirement is that the table will be created
     request.requirements().forEach(requirement -> requirement.validate(ops.current()));
 
     CreateCommitInfo info = extractCreateCommitInfo(request);
+
+    if (info.schema == null) {
+      throw new BadRequestException("No schema specified for create commit");
+    }
+
     PartitionSpec partitionSpec =
         info.unboundPartitionSpec == null ? PartitionSpec.unpartitioned() : info.unboundPartitionSpec.bind(info.schema);
     SortOrder sortOrder =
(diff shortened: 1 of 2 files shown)


nastra (commented):
changes LGTM, but I think it would be good to have some tests that show/verify this behavior (since the PR mentions that it currently only affects tests)

bryanck:
It is pretty well tested by TestRESTCatalog (via RESTCatalogAdapter), I feel like adding additional tests is a somewhat redundant, but I can add that if folks feel it is worthwhile.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -236,37 +252,22 @@ public static LoadTableResponse loadTable(Catalog catalog, TableIdentifier ident
...
-      Table table = catalog.loadTable(ident);
-      if (table instanceof BaseTable) {
-        TableOperations ops = ((BaseTable) table).operations();
-        finalMetadata = commit(ops, request);
-      } else {
-        throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
-      }
+      finalMetadata = commitUpdate(catalog, ident, request);
     }
-
```
rdblue, line 260: Nit: we like to keep a newline separating control flow blocks from the following statements.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -282,20 +283,77 @@ private static boolean isCreate(UpdateTableRequest request) {
...
+    } else {
+      throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+    }
+  }
+
+  private static TableMetadata commitCreate(Catalog catalog, TableIdentifier ident, UpdateTableRequest request) {
+    // this is a hacky way to get TableOperations for an uncommitted table
+    Transaction transaction = catalog.buildTable(ident, EMPTY_SCHEMA).createOrReplaceTransaction();
+    if (!(transaction instanceof BaseTransaction)) {
+      throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTransaction");
```
rdblue, line 300: Looks like this could be cleaner if it were rewritten to use `Preconditions.checkState` instead.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -282,20 +283,77 @@ private static boolean isCreate(UpdateTableRequest request) {
...
+      TableOperations ops = ((BaseTable) table).operations();
+      return commit(ops, request.requirements(), request.updates());
+    } else {
+      throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+    }
+  }
+
+  private static TableMetadata commitCreate(Catalog catalog, TableIdentifier ident, UpdateTableRequest request) {
+    // this is a hacky way to get TableOperations for an uncommitted table
+    Transaction transaction = catalog.buildTable(ident, EMPTY_SCHEMA).createOrReplaceTransaction();
```
rdblue, line 298: I think this should just build the transaction directly, instead of starting with a `newTableMetadata`:

```java
CreateCommitInfo info = extractCreateCommitInfo(request);
TableBuilder builder = catalog.buildTable(info.schema)
if (info.unboundPartitionSpec != null) {
  builder.withPartitionSpec(info.unboundPartitionSpec.bind(info.schema);
}
...
Transaction transaction = builder.createOrReplaceTransaction();
TableOperations ops = ((TransactionTable) transaction.table()).operations();
TableMetadata start = ops.current();
TableMetadata.Builder metaBuilder = TableMetadata.buildFrom(start)
info.oherUpdates.forEach(update -> update.applyTo(metaBuilder));
ops.commit(start, builder.build());
```

Then we should be able to remove the `underlyingOps` and `startMetadata` methods that were added to make this work!

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
```diff
@@ -282,20 +283,77 @@ private static boolean isCreate(UpdateTableRequest request) {
...
 
     return ops.current();
   }
 
-  private static TableMetadata commit(TableOperations ops, UpdateTableRequest request) {
+  private static CreateCommitInfo extractCreateCommitInfo(UpdateTableRequest request) {
+    CreateCommitInfo result = new CreateCommitInfo();
+    for (MetadataUpdate update : request.updates()) {
+      if (update instanceof MetadataUpdate.AddSchema) {
+        result.schema = ((MetadataUpdate.AddSchema) update).schema();
```
rdblue, line 331: I don't think it is safe to do this for every update. This should only consume the first updates that were added by the client to initialize the table. Otherwise, there could be an initial PartitionSpec that is later updated in the create transaction. This would end up using the second PartitionSpec and would have the same bug that you're trying to fix here.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
bryanck: This is done.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
bryanck: I made this change.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
bryanck: I attempted this approach but it wasn't working for me. The ops looks like it was a temporary ops that didn't perform a commit to the delegate catalog.

bryanck:
> changes LGTM, but I think it would be good to have some tests that show/verify this behavior (since the PR mentions that it currently only affects tests)

I added an assertion that checks that the file spec ID matches the table spec ID.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
rdblue: Ah, yeah. You'd probably need to use `commitTransaction()` at the end.

bryanck: I made this change, though still access `underlyingOps()` to check the requirements. Alternatively I could call `builder.createTransaction()` which will throw the same exception, but the message is different (and causes a test assertion to fail).

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
bryanck: I changed this so all updates after a second `AddSchema` will be applied on top of the initial table.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
bryanck: I still need to fix something, also I'm adding a test...

bryanck: I made this change. Filtering out the setter types with ID of -1 wasn't working, I changed that to filter those out only if an associated add type is pending. I also added a test that adds 2 files, each associated with a different schema, spec, and sort order.

core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java
bryanck: I made another change. Filtering out the setter types with ID of -1 wasn't working, I changed that to filter those out only if an associated add type is not pending. I also added a test that adds 2 files in a create commit, with each file associated with a different schema, spec, and sort order.

bryanck:
This has gotten a little complicated, trying to infer the create table (I tried to keep the change scoped to CatalogHandlers). There is a simpler approach but it requires some changes to other classes (I'm testing it out now).

bryanck:
Here is the [alternative approach](https://github.com/bryanck/iceberg/pull/1/files). This changes the TableMetadata builder to allow starting from empty (rather than requiring a base). Also, it changes the REST create to always include a partition spec and sort order.

bryanck:
I'm closing this, in favor of https://github.com/apache/iceberg/pull/5235

PR #5200, opened 2022-07-05
Core: Fix ErrorProne Warnings

nastra:
The only warning that is left is this one from https://github.com/apache/iceberg/blob/2f8a09a8f81b27b8ba4025efaa328244ae0fc2a6/core/src/main/java/org/apache/iceberg/util/ZOrderByteUtils.java#L182-L183
which seems legit as we're doing a lossy cast that would fail compilation if I rewrite it (https://errorprone.info/bugpattern/NarrowingCompoundAssignment):

```
> Task :iceberg-core:compileJava
/home/nastra/Development/workspace/iceberg/core/src/main/java/org/apache/iceberg/util/ZOrderByteUtils.java:182: warning: [NarrowingCompoundAssignment] Compound assignments from int to byte hide lossy casts
      interleavedBytes[interleaveByte] |=
                                       ^
    (see https://errorprone.info/bugpattern/NarrowingCompoundAssignment)
  Did you mean 'interleavedBytes[interleaveByte] = (byte) (interleavedBytes[interleaveByte] | (columnsBinary[sourceColumn][sourceByte] & 1 << sourceBit) >>> sourceBit << interleaveBit);'?
```

### DIFF
--- a/core/src/main/java/org/apache/iceberg/ManifestFiles.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestFiles.java
@@ -226,6 +226,7 @@ public class ManifestFiles {
     }
   }
 
+  @SuppressWarnings("Finally")
   private static ManifestFile copyManifestInternal(int formatVersion, ManifestReader<DataFile> reader,
                                                    OutputFile outputFile, long snapshotId,
                                                    SnapshotSummary.Builder summaryBuilder,
--- a/core/src/main/java/org/apache/iceberg/MetricsModes.java
+++ b/core/src/main/java/org/apache/iceberg/MetricsModes.java
@@ -22,7 +22,6 @@ package org.apache.iceberg;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.Locale;
-import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
@@ -134,7 +133,7 @@ public class MetricsModes {
 
     @Override
     public int hashCode() {
-      return Objects.hash(length);
+      return Integer.hashCode(length);
     }
   }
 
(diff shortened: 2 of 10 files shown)


rdblue (approved):

RussellSpitzer:
> The only warning that is left is this one from
> 
> https://github.com/apache/iceberg/blob/2f8a09a8f81b27b8ba4025efaa328244ae0fc2a6/core/src/main/java/org/apache/iceberg/util/ZOrderByteUtils.java#L182-L183
> 
> 
> which seems legit as we're doing a lossy cast that would fail compilation if I rewrite it (https://errorprone.info/bugpattern/NarrowingCompoundAssignment):
> ```
> > Task :iceberg-core:compileJava
> /home/nastra/Development/workspace/iceberg/core/src/main/java/org/apache/iceberg/util/ZOrderByteUtils.java:182: warning: [NarrowingCompoundAssignment] Compound assignments from int to byte hide lossy casts
>       interleavedBytes[interleaveByte] |=
>                                        ^
>     (see https://errorprone.info/bugpattern/NarrowingCompoundAssignment)
>   Did you mean 'interleavedBytes[interleaveByte] = (byte) (interleavedBytes[interleaveByte] | (columnsBinary[sourceColumn][sourceByte] & 1 << sourceBit) >>> sourceBit << interleaveBit);'?
> ```

So this is as intended 

The code is a bit complicated but a description would be
``` java
 interleavedBytes[interleaveByte] |=  //  if the target bit isn't set it
         (columnsBinary[sourceColumn][sourceByte] & 1 << sourceBit) >>> sourceBit  // Get the bit we are checking and move it all the way into the 1's place
<< interleaveBit;  // Move the bit back into our target position
```

So 
```java
(columnsBinary[sourceColumn][sourceByte] & 1 << sourceBit) 
```
Can only produce values between 0 and 128 (00000000), (00000001), (00000010), .... (10000000)

Then

```
>>> sourceBit 
```
Can only produce values between 0 and 1 

then

```
<< interleaveBit
```

Can only produce values again between 0 and 128

Now if interleave bit or source bit were greater than 7 we would be in trouble but those are locked down

nastra:
thanks for the clarification @RussellSpitzer. I suppressed the warning for [NarrowingCompoundAssignment](https://errorprone.info/bugpattern/NarrowingCompoundAssignment) in https://github.com/apache/iceberg/pull/5212

PR #5201, opened 2022-07-05
Python: Move Snapshot to Pydantic

Fokko:


python/pyiceberg/utils/iceberg_base_model.py
```diff
@@ -38,6 +42,7 @@ class Config:
         keep_untouched = (cached_property,)
         allow_population_by_field_name = True
         frozen = True
+        json_encoders = {"Summary": test_encoder}
```
rdblue, line 45: What's happening here?

python/pyiceberg/table/snapshots.py
```diff
@@ -0,0 +1,94 @@
...
+
+class Snapshot(IcebergBaseModel):
+    snapshot_id: int = Field(alias="snapshot-id")
+    parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id")
+    sequence_number: Optional[int] = Field(alias="sequence-number", default=None)
+    timestamp_ms: int = Field(alias="timestamp-ms")
+    manifest_list: Optional[str] = Field(
+        alias="manifest-list", description="Location of the snapshot's manifest list file", default=None
+    )
+    manifests: List[str] = Field(default_factory=list, repr=False, exclude=True)
```
rdblue, line 92: I would probably not include this. Manifests were originally listed in snapshot, but that caused problems and we quickly moved to manifest lists. It's really unlikely that there are tables still listing manifests in the snapshot itself, so I'd probably just drop this.

python/pyiceberg/table/snapshots.py
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+
+class Snapshot(IcebergBaseModel):
+    snapshot_id: int = Field(alias="snapshot-id")
+    parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id")
+    sequence_number: Optional[int] = Field(alias="sequence-number", default=None)
+    timestamp_ms: int = Field(alias="timestamp-ms")
+    manifest_list: Optional[str] = Field(
+        alias="manifest-list", description="Location of the snapshot's manifest list file", default=None
```
rdblue, line 90: If you remove `manifests`, then there's no need to make this `Optional`

python/pyiceberg/table/snapshots.py
```diff
@@ -0,0 +1,94 @@
...
+            return operation
+        else:
+            # Should not happen
+            raise ValueError(f"Unknown type of operation: {operation}")
+
+    @property
+    def additional_properties(self) -> Dict[str, str]:
+        return {
+            k: v for k, v in self.__root__.items() if k != OPERATION  # type: ignore # We know that they are all string, and we don't want to check
+        }
```
rdblue, line 81: Might be better to keep a copy of this in `__init__`?

python/tests/table/test_snapshots.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_serialize_summary_with_properties():
+    assert Summary(Operation.APPEND, property="yes").json() == """{"operation": "append", "property": "yes"}"""
+
+
+def test_serialize_snapshot():
+    snapshot = Snapshot(
+        snapshot_id=25,
+        parent_snapshot_id=19,
+        sequence_number=200,
```
rdblue, line 32: You may want a test without a sequence number?

python/pyiceberg/utils/iceberg_base_model.py
Fokko: Ah, that was some spielerij.  Instead of using a custom encoder, I ended up making the root of the Summary a Dict, which solved it in a much nicer way.

Fokko: I've removed this

python/pyiceberg/table/snapshots.py
Fokko: Thanks for the historical context. Sounds like an excellent idea. I'll drop this one, and make the manifest-list mandatory.

python/pyiceberg/table/snapshots.py
Fokko: Done!

python/tests/table/test_snapshots.py
Fokko: Good catch, I've added a test. I noticed that in Java we set the sequence number to zero. I tried to do this in Python as well, but then we don't want to write the value. Turns out that setting it to `None` makes it easier since we don't write `None` keys in the json object (similar to doc in the NestedField).

python/pyiceberg/table/snapshots.py
```diff
@@ -0,0 +1,90 @@
...
+    def additional_properties(self) -> Dict[str, str]:
+        return self._additional_properties
+
+
+class Snapshot(IcebergBaseModel):
+    snapshot_id: int = Field(alias="snapshot-id")
+    parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id")
+    sequence_number: Optional[int] = Field(alias="sequence-number", default=None)
+    timestamp_ms: int = Field(alias="timestamp-ms")
+    manifest_list: Optional[str] = Field(alias="manifest-list", description="Location of the snapshot's manifest list file")
```
rdblue, line 88: I think this can be required

PR #5205, opened 2022-07-05
API - Deprecate unused runSafely functions

kbendick:
When cleaning up some error prone warnings, I noticed that the `ExceptionUtil.runSafely` methods are not used anywhere other than from their own tests.

Given that we are going to have API stability guarantees come 1.0, we should consider deprecating these for 0.14.0 and then removing them at 1.0 so we're not giving API / ABI compatibility guarantees for code that isn't used within the library and gives error prone warnings that are difficult to reason about.

Original PR where the discussion of these functions took place: https://github.com/apache/iceberg/pull/5190


kbendick:
I'd be ok with removing these now, though they are part of API so we might want to deprecate for at least one release.

api/src/test/java/org/apache/iceberg/util/TestExceptionUtil.java
```diff
@@ -1,170 +0,0 @@
...
-package org.apache.iceberg.util;
-
-import java.io.IOException;
-import org.assertj.core.api.Assertions;
-import org.junit.Assert;
-import org.junit.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public class TestExceptionUtil {
```
dimas-b, line 29: Why remove the test if the tested code is still in the codebase? Even if deprecated, it'd be nice to validate that it works. These tests do not look too expensive to run :thinking:

kbendick: Happy to add them back!

dimas-b: thx :)

dimas-b: nit: update the PR'd summary line?

kbendick: Updated. I'm going to add a note about removing in 1.0 as well.

nastra (approved):

kbendick:
cc @RussellSpitzer I'm not sure if you guys use these functions internally.

They've always given somewhat difficult to reason about error prone warnings, and I discovered they aren't used at all (outside of in their associated tests). So I deprecated them but wanted to give you a heads up.

RussellSpitzer:
We do not use these, +1

kbendick:
Ryan mentioned that since this code is tested and works, we might want to have it still.

So I'm opening a PR to fix a small bug in the thrown combined `Throwable` and suppress the warning.

PR #5207, opened 2022-07-05
Core: Add length arg to FileIO.newInputFile

rdblue:
While investigating #5206, I found that a lot of time is spent in `InputFile#getLength()` for manifest files during job planning. This isn't necessary because `ManifestFile` stored in a manifest list has the manifest file length. This PR avoids round-trip calls to get object lengths by passing the length into `FileIO.newInputFile`. This implements the improvement for S3, GCP, and Hadoop.

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
@@ -70,7 +70,7 @@ public class S3OutputFile extends BaseS3File implements OutputFile, NativelyEncr
 
   @Override
   public InputFile toInputFile() {
-    return new S3InputFile(client(), uri(), awsProperties(), metrics());
+    return new S3InputFile(client(), uri(), null, awsProperties(), metrics());
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


gcp/src/main/java/org/apache/iceberg/gcp/gcs/GCSOutputFile.java
```diff
@@ -67,6 +67,6 @@ public PositionOutputStream createOrOverwrite() {
 
   @Override
   public InputFile toInputFile() {
-    return new GCSInputFile(storage(), blobId(), gcpProperties(), metrics());
+    return new GCSInputFile(storage(), blobId(), null, gcpProperties(), metrics());
```
danielcweeks, line 70: Could we just use `GCSInputFile.fromLocation(...)` to avoid calling the constructor with null?

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputFile.java
```diff
@@ -70,7 +70,7 @@ public PositionOutputStream createOrOverwrite() {
 
   @Override
   public InputFile toInputFile() {
-    return new S3InputFile(client(), uri(), awsProperties(), metrics());
+    return new S3InputFile(client(), uri(), null, awsProperties(), metrics());
```
danielcweeks, line 73: Similar to GCS, could we just call `S3InputFile.fromLocation(...)` to avoid calling the constructor with null?

danielcweeks (approved):
+1 Minor comment/suggestion.  Looks good either way.

gcp/src/main/java/org/apache/iceberg/gcp/gcs/GCSOutputFile.java
rdblue: I just didn't want to change very much here. We probably could.

PR #5210, opened 2022-07-06
Build: Upgrade test dependencies to latest version

XN137:
- upgrade junit5 to 5.8.2
- upgrade assertj to 3.23.1
- upgrade mockserver to 5.13.2
- upgrade mockito to 4.6.1
- org.mockito.Matchers class was removed

### DIFF
--- a/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
@@ -35,7 +35,6 @@ import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.types.Types;
 import org.junit.Assert;
 import org.junit.Test;
-import org.mockito.Matchers;
 import org.mockito.Mockito;
 import software.amazon.awssdk.services.glue.model.AccessDeniedException;
 import software.amazon.awssdk.services.glue.model.ConcurrentModificationException;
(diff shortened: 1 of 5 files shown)


nastra (approved):

XN137:
rebased due to merge conflict

szehon-ho:
It makes sense to upgrade to me, but do you think we can leave out the ArgumentMatchers => Mockito refactor?  It doesn't seem that valuable unless its enforced going forward, one way or the other.  As any() is defined on ArgumentMatchers, I was actually thinking it makes sense to call it from where it is defined.

XN137:
> As any() is defined on ArgumentMatchers, I was actually thinking it makes sense to call it from where it is defined.

this was the situation on master:
```
~/code/iceberg$ rg "Mockito.any" | wc -l
45
~/code/iceberg$ rg "ArgumentMatchers.any" | wc -l
9
```
so using the any-methods from Mockito seems to be preferred across the codebase.
thus i replaced the now deleted `Matchers` with `Mockito`
and refactored `ArgumentMatchers` usage as well for consistency.

i think generally it is not unusual to use wildcard imports with Mockito as an exception:
`import static org.mockito.Mockito.*;`
or just import `Mockito` and have all related methods discoverable from a single entry point (i.e. auto-complete).

that being said, i am fine to revert this part of the change if you still want me to?

szehon-ho:
Yea , I do understand the desire to make it the same, but there's definitely many correct ways unless there is some definite rule, so I'd say let's revert those changes to focus this pr?  Other changes look good to me.

XN137:
@szehon-ho ok i have removed the refactoring around `ArgumentMatchers` now

szehon-ho (approved):

szehon-ho:
Missed the message, thanks, looks good to me.  Will wait a little bit to see any other comments, otherwise will merge soon.

szehon-ho:
Merged, thanks @XN137 for change and @nastra for review

PR #5212, opened 2022-07-06
Arrow/AWS/Core/Hive: Fix ErrorProne warnings

nastra:


core/src/main/java/org/apache/iceberg/util/ZOrderByteUtils.java
```diff
@@ -166,7 +166,8 @@ static byte[] interleaveBits(byte[][] columnsBinary, int interleavedSize) {
    * @param interleavedSize the number of bytes to use in the output
    * @return the columnbytes interleaved
    */
-  @SuppressWarnings("ByteBufferBackingArray")
+  // NarrowingCompoundAssignment is intended here. See https://github.com/apache/iceberg/pull/5200#issuecomment-1176226163
```
kbendick, line 169: 👍

kbendick (approved):
LGTM. Thanks @nastra!

Fokko (approved):

core/src/main/java/org/apache/iceberg/util/ZOrderByteUtils.java
RussellSpitzer: The only reason I don't have this documented more is that I was hoping we can get a better implementation in :) 

I think we should write in some optimized versions for 2, 3, 4 columns since I think most cases will benefit in that range. Having more than that many Z order dimensions is probably a waste

RussellSpitzer (approved):

PR #5213, opened 2022-07-06
Core: Add copy(withStats) to ContentFile

nastra:
This allows eliminating a few if/else statements around the code which
would either call `copy()` or `copyWithoutStats()`


rdblue (approved):

rdblue:
Nice cleanup!

PR #5214, opened 2022-07-06
Spark: Add the query ID to file names

rdblue:
Spark writes should use the query ID in filenames, but it was not correctly passed through, causing each writer to use a different UUID. Adding the UUID to filenames should allow identifying files from the same write.


rdblue:
@aokolnychyi, FYI

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -622,6 +625,7 @@ public DataWriter<InternalRow> createWriter(int partitionId, long taskId, long e
       FileIO io = table.io();
 
       OutputFileFactory fileFactory = OutputFileFactory.builderFor(table, partitionId, taskId)
+          .operationId(queryId)
```
singhpk234, line 628: [question] should we also do it for [SparkPositionDeltaWrite](https://github.com/apache/iceberg/blob/master/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java#L321-L326) used in mor.

szehon-ho (approved):
Change makes sense to me

nastra:
Closing this one as it's been superseded by https://github.com/apache/iceberg/pull/6569

PR #3959, opened 2022-01-24
Disable metrics for very wide tables to reduce metadata size

szehon-ho:
Giving a try to the suggestion in the last community sync to address very-wide-table.

This changes the default MetricsMode from "truncate(16)" to "none" if there are over 100 columns (should it be counts?).  This is overriden if the user explicitly sets a default, and of course if a user explicitly sets metrics on a particular column.

PR #5215, opened 2022-07-06
Core: Update MetricsConfig to use a default for first 32 columns

rdblue:
#3959 updated `MetricsConfig` to stop writing metrics by default for tables with more than 32 columns. The intent was to avoid having too much metadata stored as stats in manifest files, but the implementation stopped writing metrics for all columns after a table reached 32 columns, which caused a large change in table performance after a table grows to 32+ columns.

This updates the logic so that the first 32 columns still have metrics, but no new metrics are stored for columns after that point unless the table has a global default set or columns are individually configured.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/MetricsConfig.java
+++ b/core/src/main/java/org/apache/iceberg/MetricsConfig.java
@@ -38,6 +38,8 @@ import org.slf4j.LoggerFactory;
 
 import static org.apache.iceberg.TableProperties.DEFAULT_WRITE_METRICS_MODE;
 import static org.apache.iceberg.TableProperties.DEFAULT_WRITE_METRICS_MODE_DEFAULT;
+import static org.apache.iceberg.TableProperties.METRICS_MAX_INFERRED_COLUMN_DEFAULTS;
+import static org.apache.iceberg.TableProperties.METRICS_MAX_INFERRED_COLUMN_DEFAULTS_DEFAULT;
 import static org.apache.iceberg.TableProperties.METRICS_MODE_COLUMN_CONF_PREFIX;
 
 @Immutable
@@ -140,9 +142,7 @@ public final class MetricsConfig implements Serializable {
    * @return metrics configuration
    */
   private static MetricsConfig from(Map<String, String> props, Schema schema, SortOrder order) {
-    int maxInferredDefaultColumns = PropertyUtil.propertyAsInt(props,
-        TableProperties.METRICS_MAX_INFERRED_COLUMN_DEFAULTS,
-        TableProperties.METRICS_MAX_INFERRED_COLUMN_DEFAULTS_DEFAULT);
+    int maxInferredDefaultColumns = maxInferredColumnDefaults(props);
     Map<String, MetricsMode> columnModes = Maps.newHashMap();
 
     // Handle user override of default mode
@@ -157,7 +157,7 @@ public final class MetricsConfig implements Serializable {
       defaultMode = DEFAULT_MODE;
 
     } else {
-      // a inferred default mode is applied to the first few columns, up to the limit
+      // an inferred default mode is applied to the first few columns, up to the limit
       Schema subSchema = new Schema(schema.columns().subList(0, maxInferredDefaultColumns));
       for (Integer id : TypeUtil.getProjectedIds(subSchema)) {
         columnModes.put(subSchema.findColumnName(id), DEFAULT_MODE);
@@ -173,11 +173,10 @@ public final class MetricsConfig implements Serializable {
     sortedCols.forEach(sc -> columnModes.put(sc, sortedColDefaultMode));
 
     // Handle user overrides of defaults
-    MetricsMode finalDefaultModeVal = defaultMode;
     for (String key : props.keySet()) {
       if (key.startsWith(METRICS_MODE_COLUMN_CONF_PREFIX)) {
         String columnAlias = key.replaceFirst(METRICS_MODE_COLUMN_CONF_PREFIX, "");
-        MetricsMode mode = parseMode(props.get(key), finalDefaultModeVal, "column " + columnAlias);
+        MetricsMode mode = parseMode(props.get(key), defaultMode, "column " + columnAlias);
         columnModes.put(columnAlias, mode);
       }
     }
@@ -195,6 +194,19 @@ public final class MetricsConfig implements Serializable {
       return MetricsModes.Truncate.withLength(16);
     } else {
       return defaultMode;
+    }
+  }
+
+  private static int maxInferredColumnDefaults(Map<String, String> properties) {
+    int maxInferredDefaultColumns = PropertyUtil.propertyAsInt(properties,
+        METRICS_MAX_INFERRED_COLUMN_DEFAULTS, METRICS_MAX_INFERRED_COLUMN_DEFAULTS_DEFAULT);
+    if (maxInferredDefaultColumns < 0) {
+      LOG.warn("Invalid value for {} (negative): {}, falling back to {}",
+          METRICS_MAX_INFERRED_COLUMN_DEFAULTS, maxInferredDefaultColumns,
+          METRICS_MAX_INFERRED_COLUMN_DEFAULTS_DEFAULT);
+      return METRICS_MAX_INFERRED_COLUMN_DEFAULTS_DEFAULT;
+    } else {
+      return maxInferredDefaultColumns;
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/MetricsConfig.java
+++ b/core/src/main/java/org/apache/iceberg/MetricsConfig.java
@@ -29,6 +29,8 @@ import org.apache.iceberg.exceptions.ValidationException;
 import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.types.TypeUtil;
+import org.apache.iceberg.util.PropertyUtil;
 import org.apache.iceberg.util.SerializableMap;
 import org.apache.iceberg.util.SortOrderUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -267,6 +267,10 @@ private TableProperties() {
   public static final String METADATA_DELETE_AFTER_COMMIT_ENABLED = "write.metadata.delete-after-commit.enabled";
   public static final boolean METADATA_DELETE_AFTER_COMMIT_ENABLED_DEFAULT = false;
 
+  public static final String METRICS_MAX_INFERRED_COLUMN_DEFAULTS =
```
danielcweeks, line 270: nit: `INFERRED` doesn't really make sense to me in this context.  We're not inferring, we're actually explicit here.

rdblue: I debated what to call this. The problem is that we are defaulting the default. You can explicitly set a default using `write.metadata.metrics.default`, and that will apply to all columns. If you don't _set_ a default we _infer_ one for you, but only for the first 32 columns (at least, after this PR). That's why I used "inferred" -- I thought it was better than `max-defaulted-default-columns`.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -136,50 +131,58 @@ public static MetricsConfig forPositionDelete(Table table) {
...
+      // a user-configured default mode is applied for all columns
+      defaultMode = parseMode(configuredDefault, DEFAULT_MODE, "default");
+
+    } else if (schema == null || schema.columns().size() <= maxInferredDefaultColumns) {
+      // there are less than the inferred limit, so the default is used everywhere
+      defaultMode = DEFAULT_MODE;
+
+    } else {
+      // a inferred default mode is applied to the first few columns, up to the limit
+      Schema subSchema = new Schema(schema.columns().subList(0, maxInferredDefaultColumns));
```
danielcweeks, line 161: Wouldn't this still cause an issue with metrics for nested types?  Once we go cross the column threshold we will actually drop nested stats and only include top-level column stats.

rdblue: No, we keep the first 32 columns, but we add a metrics mode for each sub-type. That's why the loop below iterates over `TypeUtil.getProjectedIds(subSchema)`

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -195,6 +198,16 @@ private static MetricsMode sortedColumnDefaultMode(MetricsMode defaultMode) {
     }
   }
 
+  private static MetricsMode parseMode(String modeString, MetricsMode fallback, String context) {
+    try {
+      return MetricsModes.fromString(modeString);
+    } catch (IllegalArgumentException err) {
+      // User override was invalid, log the error and use the default
+      LOG.warn("Ignoring invalid metrics mode ({}): {}", context, modeString, err);
```
szehon-ho, line 206: This may not log err?  (It looks like the parameterized version of log.warn)

core/src/main/java/org/apache/iceberg/MetricsConfig.java
szehon-ho: OK so we may have still more than 32 type metrics (if chosen 32 columns are nested for example).

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -136,50 +131,58 @@ public static MetricsConfig forPositionDelete(Table table) {
...
    * @param props will be read for metrics overrides (write.metadata.metrics.column.*) and default
    *              (write.metadata.metrics.default)
+   * @param schema table schema
    * @param order sort order columns, will be promoted to truncate(16)
-   * @param defaultMode default, if not set by user property
    * @return metrics configuration
    */
-  private static MetricsConfig from(Map<String, String> props, SortOrder order, String defaultMode) {
+  private static MetricsConfig from(Map<String, String> props, Schema schema, SortOrder order) {
+    int maxInferredDefaultColumns = PropertyUtil.propertyAsInt(props,
```
szehon-ho, line 143: Add precondition that its >= 0?

core/src/main/java/org/apache/iceberg/TableProperties.java
szehon-ho: Agree with this table property, initially I had made one but it was taken out during the discussions.  Indeed it's a bit of a confusing config, but I dont see any other great option.

szehon-ho (approved):
Yea this makes sense, thanks for making the change (should have considered this case earlier).  Added some small comments

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -136,50 +131,58 @@ public static MetricsConfig forPositionDelete(Table table) {
...
+        TableProperties.METRICS_MAX_INFERRED_COLUMN_DEFAULTS,
```
aokolnychyi, line 144: nit: I think we have static imports for all other table properties in this class.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -136,50 +131,58 @@ public static MetricsConfig forPositionDelete(Table table) {
...
+    if (configuredDefault != null) {
+      // a user-configured default mode is applied for all columns
+      defaultMode = parseMode(configuredDefault, DEFAULT_MODE, "default");
+
+    } else if (schema == null || schema.columns().size() <= maxInferredDefaultColumns) {
+      // there are less than the inferred limit, so the default is used everywhere
+      defaultMode = DEFAULT_MODE;
+
+    } else {
+      // a inferred default mode is applied to the first few columns, up to the limit
```
aokolnychyi, line 160: nit: `a` -> `an`?

core/src/main/java/org/apache/iceberg/MetricsConfig.java
aokolnychyi, line 161: Quick warning: `subList` would return a view on top of the original list and any subsequent changes would be reflected in both. This does not seem to cause issues here but I better mention.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -136,50 +131,58 @@ public static MetricsConfig forPositionDelete(Table table) {
...
-          try {
-            mode = MetricsModes.fromString(props.get(key));
-          } catch (IllegalArgumentException err) {
-            // Mode was invalid, log the error and use the default
-            LOG.warn("Ignoring invalid metrics mode for column {}: {}", columnAlias, props.get(key), err);
-            mode = finalDefaultModeVal;
-          }
-          columnModes.put(columnAlias, mode);
-        });
+    MetricsMode finalDefaultModeVal = defaultMode;
```
aokolnychyi, line 176: Do we still need `finalDefaultModeVal`? Can we use `defaultMode` instead now?

core/src/main/java/org/apache/iceberg/MetricsConfig.java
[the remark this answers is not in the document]
rdblue: Yes, but this is the current behavior. We use the top-level columns for the current check.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
rdblue: If the last argument is a Throwable, the logger does the right thing.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
rdblue: Fixed.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
rdblue: Fixed.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
rdblue: This is a temporary schema so it should be fine.

core/src/main/java/org/apache/iceberg/MetricsConfig.java
rdblue: Fixed.

core/src/main/java/org/apache/iceberg/TableProperties.java
rdblue: The problem with "persist" and similar is that it misses the distinction between an explicit default (when `write.metadata.metrics.default` is set) and an implicit default that comes from Iceberg. I think the right behavior is to preserve what we currently do, which is to use the explicit default for all columns. But that means that this property should obviously not apply to the explicit default. That's why I used "inferred default".

What about changing this to include `unconfigured` or `missing`? Something like `missing-mode-limit`?

aokolnychyi: Let's just keep it as-is then. I don't think it is a big deal.

szehon-ho (approved):

bryanck:
This is already merged, but I thought I'd leave feedback anyway, in case it is useful.

As a data engineer, many tables I have maintained have more than 32 top-level columns. Often columns used for partitioning, sorting, auditing, and so forth are put at the end of a table schema, but these are some of the most frequently used in filtering. Also, additional columns are generally added at the end of the schema. The assumption that the first columns in a table schema are the most important to have stats on is not always accurate. 

In testing 0.14, I ran into missing stats on tables, which was confusing and difficult to debug. I image those new to Iceberg and who are most likely to leave settings at the default, it would be even more confusing.

I feel a more sensible default is to leave it the same as previous Iceberg versions (i.e. no column limit). Then an option could be introduced to limit the number of columns so those that prefer can set it on their tables, e.g. "first(32)". I feel it is better to err on the side of too many stats and dial that back as needed.

szehon-ho:
I think that's an interesting enhancement idea, if I understand, to make write.metadata.metrics.max-inferred-column-defaults to also accept first or last.  An organization might be able to set last(32) as default on the catalog level (ie, https://github.com/apache/iceberg/pull/4011).  On the other hand, sort and partition columns are already auto-promoted to have stats today, curious do you see a good impact of having stats on other columns?

core/src/main/java/org/apache/iceberg/MetricsConfig.java
```diff
@@ -45,9 +49,8 @@ public final class MetricsConfig implements Serializable {
   private static final Joiner DOT = Joiner.on('.');
 
   // Disable metrics by default for wide tables to prevent excessive metadata
```
findinpath, line 51: @danielcweeks  Is this a left-over comment?

PR #5222, opened 2022-07-07
Don't throw from finally block

nastra:
Any exception coming from the try-catch block will be swallowed and the user will only see the exception from the finally block.
Note that `cleanupMetadataAndUnlock()` is being called from a `finally` block.

It seems enough to just log the error but not throw an exception when
metadata cleanup failed.

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/dynamodb/DynamoDbTableOperations.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/dynamodb/DynamoDbTableOperations.java
@@ -135,8 +135,7 @@ class DynamoDbTableOperations extends BaseMetastoreTableOperations {
           io().deleteFile(newMetadataLocation);
         }
       } catch (RuntimeException e) {
-        LOG.error("Fail to cleanup metadata file at {}", newMetadataLocation, e);
-        throw e;
+        LOG.error("Failed to cleanup metadata file at {}", newMetadataLocation, e);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


aws/src/main/java/org/apache/iceberg/aws/dynamodb/DynamoDbTableOperations.java
```diff
@@ -135,8 +135,7 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
           io().deleteFile(newMetadataLocation);
         }
       } catch (RuntimeException e) {
-        LOG.error("Fail to cleanup metadata file at {}", newMetadataLocation, e);
-        throw e;
+        LOG.error("Failed to cleanup metadata file at {}", newMetadataLocation, e);
```
dimas-b, line 138: Would you consider adding `e` as a "suppressed" exception to any previous exception?

It's a bit more refactoring, but often having more information in the thrown exception is more useful than logs.

rdblue (approved):

aws/src/main/java/org/apache/iceberg/aws/dynamodb/DynamoDbTableOperations.java
rdblue: I think this is fine. If it were easier, I'd opt to suppress the exception (maybe we should consider using `runSafely` for that) but since this is just warning that a file wasn't cleaned up, the important thing is not to throw.

PR #5225, opened 2022-07-07
Spark: Provide size estimate for table broadcast

bryanck:
During query planning, Iceberg broadcasts table data, and Spark will run its size estimator tool on the object as part of the broadcast. This size estimation can be an expensive operation in some cases, for example, when the table uses `S3FileIO`, as the object graph being analyzed is large even if the data isn't being serialized. Ultimately this can cause performance problems in query planning. Also, the size estimate is not correct as it includes data that will not be serialized. For example, a table with an S3FileIO reference was being estimated at 16MB in size when the serialized size was only ~32KB.

This PR creates a subclass of `SerializableTable` that implements Spark's `KnownSizeEstimation` trait and uses that for broadcasts. By doing this, the expensive size estimation calculation is bypassed. The size is set to 32KB, as during testing the size of the serialized data was very roughly in this ballpark.

One side note - it appears as if the same table is being broadcast multiple times during query planning, so there are further opportunities for optimization in this area.

Screen Shot 2022-07-07 at 4 03 28 PM

### DIFF
--- a/core/src/main/java/org/apache/iceberg/SerializableTable.java
+++ b/core/src/main/java/org/apache/iceberg/SerializableTable.java
@@ -67,7 +67,7 @@ public class SerializableTable implements Table, Serializable {
   private transient volatile Map<Integer, PartitionSpec> lazySpecs = null;
   private transient volatile SortOrder lazySortOrder = null;
 
-  private SerializableTable(Table table) {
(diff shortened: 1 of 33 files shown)


nastra (approved):
really nice find, LGTM

PR #5230, opened 2022-07-08
Core,API: Add table property transform to support stateful property update.

jfz:
This PR adds a transform API to table properties update, and the transforms will be re-applied based on refreshed properties when there's a conflict, and this make it possible for truly stateful and transactional property updates.

Currently, the table property update commit is basically "last commit wins", and the property value update happens out side of iceberg library so it will not be re-applied during commit retries when there's a conflict, that can cause some commits "effectively" being overwritten when the property update is supposed to be based on existing value.

Consider below example use case: 
Table property "subscribers" is used to track who should be notified whenever there's a new data write, and "subscribers" is a set of users and we do add/remove to manage users.

Steps for a value update loss:
1. initially there's only 1 user, subscribers value = \<u1>
2. commit A attempts to add user u2 to subscribers: set value to  <u1,u2>
3. commit B attempts to add user u3 to subscribers: set value to  <u1,u3>
4. commit B succeeded first, now subscribers value is: <u1,u3>
5. Commit A failed, and retry succeeded, setting new value to: <u1,u2>
6. now both A and B succeeded, final value: <u1,u2>    =>   commit B is effectively lost.

With transform API, commits can request a transform like "add user u2 to existing set" instead of "set it to <u1,u2>", and no commit will be lost because transform is re-applied for commit retries.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/PropertiesUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/PropertiesUpdate.java
@@ -68,7 +68,7 @@ class PropertiesUpdate implements UpdateProperties {
     Preconditions.checkNotNull(key, "Key cannot be null");
     Preconditions.checkNotNull(transformFunc, "Transform function cannot be null");
     Preconditions.checkArgument(!updates.containsKey(key),
-        "Cannot updates and transform the same key: %s", key);
+        "Cannot update and transform the same key: %s", key);
     Preconditions.checkArgument(!removals.contains(key),
         "Cannot remove and transform the same key: %s", key);
     transforms.put(key, transformFunc);
--- a/core/src/main/java/org/apache/iceberg/PropertiesUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/PropertiesUpdate.java
@@ -21,6 +21,7 @@ package org.apache.iceberg;
 
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
@@ -39,6 +40,7 @@ import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS_DEFA
 class PropertiesUpdate implements UpdateProperties {
   private final TableOperations ops;
   private final Map<String, String> updates = Maps.newHashMap();
+  private final Map<String, Function<String, String>> transforms = Maps.newHashMap();
   private final Set<String> removals = Sets.newHashSet();
   private TableMetadata base;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/PropertiesUpdate.java
```diff
@@ -53,18 +55,34 @@ public UpdateProperties set(String key, String value) {
...
 
     return this;
   }
 
+  @Override
+  public UpdateProperties transform(String key, Function<String, String> transformFunc) {
+    Preconditions.checkNotNull(key, "Key cannot be null");
+    Preconditions.checkNotNull(transformFunc, "Transform function cannot be null");
+    Preconditions.checkArgument(!updates.containsKey(key),
+        "Cannot updates and transform the same key: %s", key);
```
rdblue, line 71: Nit: should be "update" rather than "updates".

jfz: Fixed.

jfz: @rdblue @kbendick can you take a look at this?

rdblue:
I'm not sure about this. At first, it seems like a reasonable thing to support. But, Iceberg table properties are not intended to be used for arbitrary user metadata like this. They are intended for information to configure reading or writing a table. The existing behavior is that set is idempotent, so it is a stretch to say that Iceberg should support a new feature so you can use table properties for a different purpose.

There is also an implementation problem with this. Not all catalogs can guarantee the behavior introduced in this PR. For example, the REST protocol's `SetPropertiesUpdate` is an idempotent change. There's no way to communicate what the catalog service should do if there is a conflict, or even to signal that it should fail and retry in the event of a conflict. At a minimum, we would need to add a validation for table properties that enabled failing the server-side commit.

I'm not sure that those changes are worth it. Have you considered using a prefix instead of a single option? Instead of `subscribers=a,b,c` you could set `subscribers.a.enabled=true`, `subscribers.b.enabled=true`, etc. That gives you the ability to customize each one as well.

jfz:
@rdblue thanks for looking into this and the insights!

> They are intended for information to configure reading or writing a table.

Time to evolve? :) I feel this can be useful and unlock new use cases that require stateful/transactional update to any table specific meta info.

> implementation challenges, e.g. REST Cagalog

I think this can be a generic challenge for passing any sort of transformation to server that's worth looking into, a quick thought would be either serialize/deserialize or predefined transforms like `partition transforms`. Or we can choose not to support in REST Catalog to start with.

> using a prefix instead of a single option

I think it will reduce but not eliminate the chance of overwrite on commit conflicts.

rdblue:
> > They are intended for information to configure reading or writing a table.
> 
> Time to evolve? :)

I think this is a great demonstration of why we want to be strict. This would be a convenient place to stash properties, but then you run into situations like this one where properties don't provide the behavior guarantees that you want for another use case. If we think about implementing those behaviors, then we need to consider things like how to make assertions on the state of table properties in the REST API so that concurrent updates will fail and the client can retry. That's a lot of work and it makes the REST spec more difficult to implement.

I think it is best to continue with the current behavior, especially given that there is a reasonable work-around.

PR #5233, opened 2022-07-09
Rest-catalog: Add flexibility to paths in rest catalog

Fokko:
While generating Python code from the open-api spec to bootstrap the Python rest catalog implementation I noticed that the URLs were constructed differently than expected.

For our service the warehouse uuid is encoded as part of the path:
```
https://api.iceberg.io/ws/v1/oss/warehouses/8bcb0838-50fc-472d-9ddb-8feb89ef5f1e/namespaces
```

However, the open-api spec states that the `/v1/` comes after the base path:
```
/v1/namespaces/{namespace}/properties:
```
This would lead to:
```
https://api.iceberg.io/ws/oss/warehouses/8bcb0838-50fc-472d-9ddb-8feb89ef5f1e/v1/namespaces
```

where the base path is part of the server's definition:
```yaml
servers:
  - url: "{scheme}://{host}/{basePath}"
    description: Server URL when the port can be inferred from the scheme
    variables:
      scheme:
        description: The scheme of the URI, either http or https.
        default: https
      host:
        description: The host address for the specified server
        default: localhost
      basePath:
        description: Optional prefix to be appended to all routes
        default: ""
```
This will never allow us to construct an url with a prefix (the warehouse identifier). Therefore I suggest adding the prefix for the metadata-related endpoints (all except token and config). This way we can add v2 endpoints in the future to the same spec, and we also keep backward compatibility (just leaving the prefix empty)

### DIFF
--- a/.github/workflows/open-api.yml
+++ b/.github/workflows/open-api.yml
@@ -1,0 +1,48 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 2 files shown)


rdblue:
@Fokko, the base path and the prefix are different. The base path is the URL of the endpoint, where you can find `v1/confg` and other `v1` paths. The `prefix` is a separate configurable option that adds a prefix after `v1`. The distinction isn't clear from the spec, which uses "prefix" to describe the base URL. We should clean up the base URL wording, and document the use of `prefix`.

rdblue:
Thanks, @Fokko! Good to have this documented.

PR #5234, opened 2022-07-10
Core, API: BaseRowDelta, BaseOverwrite,BaseReplacePartitions, BaseRewrite to branch Impl 

namrathamyske:
issue addressed from: https://github.com/apache/iceberg/issues/3896

PR to address: https://github.com/apache/iceberg/pull/4926#issuecomment-1161499158

This PR implements the performing RowDelta operations on a particular branch ref. Throws unsupported operations with other operations.

PR continued from:
https://github.com/apache/iceberg/pull/4926


namrathamyske:
@amogh-jahagirdar @rdblue As per https://github.com/apache/iceberg/pull/4926#issuecomment-1161499158 I have changed validation checks when we commit to a branch. Right now I have implemented for BaseRowDelta usecase. My future plan is to extend for others.

My thought process is every-time we check for validations, the start snapshot should start from an ancestor of the branch. If not we can throw a validation exception. Start snapshot should ideally never go before the snapshot with which branch was created. But we don't store the commit for which the branch was created in metadata. So i am quite confused how to go ahead. Any thoughts appreciated!

core/src/main/java/org/apache/iceberg/BaseRowDelta.java
```diff
@@ -95,24 +97,46 @@ public RowDelta validateNoConflictingDeleteFiles() {
...
+    }
+
+    for (Snapshot ancestor : SnapshotUtil.ancestorsOf(snapshot.snapshotId(), base::snapshot)) {
+      if (ancestor.snapshotId() == this.startingSnapshotId) {
+        return;
+      }
+    }
+    throw new ValidationException(
+        "Snapshot %s is not an ancestor of %s", startingSnapshotId, snapshot.snapshotId());
+  }
```
amogh-jahagirdar, line 118: We can use SnapshotUtil.isAncestorOf for this.

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -349,30 +354,34 @@ public void commit() {
...
+                  // snapshot to an ID that is already current. note that this check uses identity.
+                  return;
+                }
+
+                // if the table UUID is missing, add it here. the UUID will be re-created each time
+                // this operation retries
+                // to ensure that if a concurrent operation assigns the UUID, this operation will
+                // not fail.
+                taskOps.commit(base, updated.withUUID());
+              });
```
amogh-jahagirdar, line 384: I think we can undo this.

amogh-jahagirdar (commented):
Awesome, I forgot you had this draft PR already so I think we can just continue on here. I closed out my PR https://github.com/apache/iceberg/pull/5595

core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java
```diff
@@ -133,19 +133,19 @@ protected void validate(TableMetadata base, Snapshot snapshot) {
     }
 
     if (validateNewDataFiles) {
-      validateAddedDataFiles(base, startingSnapshotId, dataConflictDetectionFilter());
+      validateAddedDataFiles(base, startingSnapshotId, dataConflictDetectionFilter(), snapshot);
```
rdblue, line 136: Can we name this `parent` instead of `snapshot`? Snapshot is not very clear because there are a lot of snapshots.

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
```diff
@@ -304,9 +304,12 @@ protected void validateAddedDataFiles(
    * @param conflictDetectionFilter an expression used to find new conflicting data files
    */
   protected void validateAddedDataFiles(
-      TableMetadata base, Long startingSnapshotId, Expression conflictDetectionFilter) {
+      TableMetadata base,
+      Long startingSnapshotId,
+      Expression conflictDetectionFilter,
+      Snapshot snapshot) {
```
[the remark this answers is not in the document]
rdblue: I think that these should all be called `parent` so it is clear that this is the parent of the snapshot to be committed.

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -1414,18 +1415,69 @@ public void testRowDeltaCaseSensitivity() {
   }
 
   @Test
-  public void testRowDeltaToBranchUnsupported() {
-    AssertHelpers.assertThrows(
-        "Should reject committing row delta to branch",
-        UnsupportedOperationException.class,
-        "Cannot commit to branch someBranch: org.apache.iceberg.BaseRowDelta does not support branch commits",
-        () ->
+  public void testBranchConflictingDeletes() {
```
rdblue, line 1418: I think that this class either needs to copy all of the tests and validate against a branch (like `testValidateDataFilesExistDefaults`), or be parameterized. I prefer parameterized.

If the tests are parameterized for `main` and `some_branch`, you could refactor all of the `commit` calls to something like this:

```
  public void commitOperation(SnapshotProducer producer) {
    if (branch.equals("main")) {
      producer.commit();
    } else {
      producer.toBranch(branch).commit();
    }
  }
```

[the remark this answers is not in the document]
namrathamk: My bad, this test is still a work in progress. Making more changes.

namrathamyske:
In PR https://github.com/apache/iceberg/pull/5618 MergeAppend and TableTestBase commit and apply methods are implemented. Will make use of those methods after its merged.

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -1270,11 +1332,16 @@ public void testConcurrentNonConflictingRowDeltaAndRewriteFilesWithSequenceNumbe
...
             .validateFromSnapshot(baseSnapshot.snapshotId());
 
-    rowDelta.commit();
-    rewriteFiles.commit();
+    commit(table, rowDelta, branch);
+    commit(table, rewriteFiles, branch);
 
-    validateTableDeleteFiles(table, deleteFile1);
-    validateTableFiles(table, dataFile2);
+    if (branch == "testBranch") {
```
amogh-jahagirdar, line 1338: equals() instead of ==

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -429,29 +470,31 @@ public void testValidateNoConflictsFromSnapshot() {
         statuses(Status.ADDED));
   }
 
-  @Test
+  // @Test
```
amogh-jahagirdar, line 473: Any reason this test is commented out?

core/src/test/java/org/apache/iceberg/TableTestBase.java
```diff
@@ -458,6 +458,18 @@ void validateTableFiles(Table tbl, DataFile... expectedFiles) {
     Assert.assertEquals("Files should match", expectedFilePaths, actualFilePaths);
   }
 
+  void validateTableFilesWithRef(Table tbl, String ref, DataFile... expectedFiles) {
```
amogh-jahagirdar, line 461: validateBranchDataFiles?

core/src/test/java/org/apache/iceberg/TableTestBase.java
```diff
@@ -472,6 +484,20 @@ void validateTableDeleteFiles(Table tbl, DeleteFile... expectedFiles) {
     Assert.assertEquals("Delete files should match", expectedFilePaths, actualFilePaths);
   }
 
+  void validateTableDeleteFilesWithRef(Table tbl, String ref, DeleteFile... expectedFiles) {
```
amogh-jahagirdar, line 487: validateBranchDeleteFiles?

core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -267,7 +267,26 @@ private Map<String, String> summary(TableMetadata previous) {
...
 
     Map<String, String> previousSummary;
-    if (previous.currentSnapshot() != null) {
+    if (!targetBranch.equals(SnapshotRef.MAIN_BRANCH)) {
+      if (previous.ref(targetBranch) != null) {
+        if (previous.snapshot(previous.ref(targetBranch).snapshotId()).summary() != null) {
+          previousSummary = previous.snapshot(previous.ref(targetBranch).snapshotId()).summary();
+        } else {
+          previousSummary = ImmutableMap.of();
+        }
```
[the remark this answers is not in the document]
amogh-jahagirdar: Also I think we could probably simplify all the if/else logic here. the branch will either be main or not, and in both cases it's the same logic. Regardless of main or not main: There is either not a prev snapshot in which case default to 0, or the prev snapshot had no summary use an empty summary. or the prev snapshot had a summary so use it.

amogh-jahagirdar (commented):
Hey Namratha thanks for all the contributions here, just had some comments.

core/src/test/java/org/apache/iceberg/TestRowDelta.java
amogh-jahagirdar: Also just realized we don't need to differentiate the two cases here, I think we could just have a single method which takes in a branch (and the branch could just be main or not, it doesn't really matter)

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -1449,18 +1519,48 @@ public void testRowDeltaCaseSensitivity() {
...
+        .addDeletes(FILE_A_DELETES)
+        .validateFromSnapshot(ancestorSnapshot)
+        .conflictDetectionFilter(Expressions.alwaysTrue())
+        .validateNoConflictingDeleteFiles()
+        .commit();
+
+    int branchSnapshot = 2;
+
+    Assert.assertEquals(table.currentSnapshot().snapshotId(), 1);
+    Assert.assertEquals(table.ops().current().ref("branch").snapshotId(), branchSnapshot);
```
amogh-jahagirdar, line 1564: Sorry not getting why we need this test case, won't it be covered by the others?

namrathamyske: Cool, will be removing these.

namrathamyske:
@jackye1995 @amogh-jahagirdar Finished for the above producer implementations. I am yet to commit tests for BaseReplacePartitions. Let me know what you think!

amogh-jahagirdar:
Great work @namrathamyske, just some minor comments but at least the core of it around the tests looks good to me. When you get a chance to update the ReplacePartition tests that would be great, since we can then see all the CI tests pass (right now it looks to be failing since the ReplacePartition tests haven't been updated yet).

amogh-jahagirdar:
Raised a PR to @namrathamyske branch for replace partition and a bug fix for validation during overwrite https://github.com/namrathamyske/iceberg/pull/20/files

namrathamyske:
@rdblue Appreciate your feedback on this!

core/src/test/java/org/apache/iceberg/TableTestBase.java
```diff
@@ -458,6 +458,18 @@ void validateTableFiles(Table tbl, DataFile... expectedFiles) {
...
+    Set<CharSequence> expectedFilePaths = Sets.newHashSet();
+    for (DataFile file : expectedFiles) {
+      expectedFilePaths.add(file.path());
+    }
+    Set<CharSequence> actualFilePaths = Sets.newHashSet();
+    for (FileScanTask task : tbl.newScan().useRef(ref).planFiles()) {
+      actualFilePaths.add(task.file().path());
+    }
+    Assert.assertEquals("Files should match", expectedFilePaths, actualFilePaths);
+  }
```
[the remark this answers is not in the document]
namrathamyske: Hi, `validateTableFiles` has 30 usages in in TestOverwriteWithValidation

core/src/test/java/org/apache/iceberg/TableTestBase.java
```diff
@@ -472,6 +484,20 @@ void validateTableDeleteFiles(Table tbl, DeleteFile... expectedFiles) {
...
+      expectedFilePaths.add(file.path());
+    }
+    Set<CharSequence> actualFilePaths = Sets.newHashSet();
+    for (FileScanTask task : tbl.newScan().useRef(branch).planFiles()) {
+      for (DeleteFile file : task.deletes()) {
+        actualFilePaths.add(file.path());
+      }
+    }
+    Assert.assertEquals("Delete files should match", expectedFilePaths, actualFilePaths);
+  }
```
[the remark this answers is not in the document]
namrathamyske: Will go ahead edit the whole class too.

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
```diff
@@ -337,15 +340,20 @@ private CloseableIterable<ManifestEntry<DataFile>> addedDataFiles(
       TableMetadata base,
       Long startingSnapshotId,
       Expression dataFilter,
-      PartitionSet partitionSet) {
+      PartitionSet partitionSet,
+      Snapshot parent) {
     // if there is no current table state, no files have been added
```
amogh-jahagirdar, line 345: My bad, missed this when I raised the change to your branch. I think we should update this comment to say "If there is no parent, no files have been added"

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
```diff
@@ -337,15 +340,20 @@ private CloseableIterable<ManifestEntry<DataFile>> addedDataFiles(
...
-    if (base.currentSnapshot() == null) {
+    if (parent == null) {
```
amogh-jahagirdar, line 346: My bad, missed this when I raised the change to your branch. I think we should update this comment to say "If there is no parent, no files have been added"

.palantir/revapi.yml
```diff
@@ -261,6 +261,235 @@ acceptedBreaks:
         \ T) throws java.io.IOException, com.fasterxml.jackson.core.JacksonException\
         \ @ org.apache.iceberg.rest.RESTSerializers.UpdateRequirementDeserializer"
       justification: "False positive - JacksonException is a subclass of IOException"
+    - code: "java.method.inheritedMovedToClass"
```
rdblue, line 264: When the review is closer to approval, we should see how many of these breaking changes can be solved. For example, this one about moving the method we may be able to take care of by overriding rather than replacing the method.

core/src/test/java/org/apache/iceberg/TableTestBase.java
```diff
@@ -458,13 +458,25 @@ void validateTableFiles(Table tbl, DataFile... expectedFiles) {
     Assert.assertEquals("Files should match", expectedFilePaths, actualFilePaths);
   }
 
-  void validateTableDeleteFiles(Table tbl, DeleteFile... expectedFiles) {
+  void validateBranchFiles(Table tbl, String ref, DataFile... expectedFiles) {
+    Set<CharSequence> expectedFilePaths = Sets.newHashSet();
+    for (DataFile file : expectedFiles) {
+      expectedFilePaths.add(file.path());
+    }
+    Set<CharSequence> actualFilePaths = Sets.newHashSet();
```
rdblue, line 466: Nit: block spacing.

core/src/main/java/org/apache/iceberg/BaseReplacePartitions.java
```diff
@@ -79,23 +79,32 @@ public ReplacePartitions validateNoConflictingData() {
...
   @Override
   public void validate(TableMetadata currentMetadata, Snapshot snapshot) {
     if (validateConflictingData) {
       if (dataSpec().isUnpartitioned()) {
-        validateAddedDataFiles(currentMetadata, startingSnapshotId, Expressions.alwaysTrue());
+        validateAddedDataFiles(
+            currentMetadata, startingSnapshotId, Expressions.alwaysTrue(), snapshot);
       } else {
-        validateAddedDataFiles(currentMetadata, startingSnapshotId, replacedPartitions);
+        validateAddedDataFiles(currentMetadata, startingSnapshotId, replacedPartitions, snapshot);
```
rdblue, line 95: Minor: I think it makes more sense to use `parent` rather than `snapshot` for the parent snapshot variable. Not a blocker, but it makes it easier to understand what's going on.

core/src/main/java/org/apache/iceberg/BaseRowDelta.java
```diff
@@ -96,23 +97,37 @@ public RowDelta validateNoConflictingDeleteFiles() {
...
+  @Override
+  protected void validate(TableMetadata base, Snapshot parent) {
+    if (parent != null) {
+      if (startingSnapshotId != null) {
+        Preconditions.checkArgument(
+            SnapshotUtil.isAncestorOf(parent.snapshotId(), startingSnapshotId, base::snapshot),
+            "Snapshot %s is not an ancestor of %s",
+            startingSnapshotId,
+            parent.snapshotId());
+      }
```
rdblue, line 114: Nit: whitespace after this block.

core/src/main/java/org/apache/iceberg/BaseRowDelta.java
```diff
@@ -96,23 +97,37 @@ public RowDelta validateNoConflictingDeleteFiles() {
...
+    return this;
+  }
+
+  @Override
+  protected void validate(TableMetadata base, Snapshot parent) {
+    if (parent != null) {
+      if (startingSnapshotId != null) {
+        Preconditions.checkArgument(
+            SnapshotUtil.isAncestorOf(parent.snapshotId(), startingSnapshotId, base::snapshot),
+            "Snapshot %s is not an ancestor of %s",
```
rdblue, line 111: I think a bit more context would be helpful if you ever encounter this error. `"Cannot validate changes: starting snapshot %s is not an ancestor of parent snapshot %s (branch %s)"`

core/src/test/java/org/apache/iceberg/TestOverwrite.java
```diff
@@ -164,40 +173,43 @@ public void testOverwriteFailsDelete() {
...
-        "Should not create a new snapshot", baseId, table.currentSnapshot().snapshotId());
+        "Should not create a new snapshot", baseId, latestSnapshot(base, branch).snapshotId());
   }
 
   @Test
   public void testOverwriteWithAppendOutsideOfDelete() {
     TableMetadata base = TestTables.readMetadata(TABLE_NAME);
-    long baseId = base.currentSnapshot().snapshotId();
+    long baseId =
+        latestSnapshot(base, branch) == null ? -1 : latestSnapshot(base, branch).snapshotId();
```
rdblue, line 186: Minor: prefer to keep ternary expressions as simple as possible. If you're calling a method twice with the same arguments, you generally want to use a variable:

```java
Snapshot latest = latestSnapshot(base, branch);
long baseId = latest == null ? -1 : latest.snapshotId();
```

It's also a bit odd that the current snapshot is always defined, but here we handle the case where the branch does not yet exist. Why is that? Are we committing to main and then implicitly branching somewhere?

core/src/test/java/org/apache/iceberg/TestOverwrite.java
```diff
@@ -164,40 +173,43 @@ public void testOverwriteFailsDelete() {
...
-        .overwriteByRowFilter(equal("date", "2018-06-08"))
-        .addFile(FILE_10_TO_14) // in 2018-06-09, NOT in 2018-06-08
-        .commit();
+    commit(
+        table,
+        table
+            .newOverwrite()
+            .overwriteByRowFilter(equal("date", "2018-06-08"))
+            .addFile(FILE_10_TO_14),
+        branch); // in 2018-06-09, NOT in 2018-06-08
```
rdblue, line 194: Minor: Can the comment remain on the `FILE_10_TO_14` line?

core/src/test/java/org/apache/iceberg/TestReplacePartitions.java
```diff
@@ -114,20 +122,22 @@ public void testReplaceAndMergeOnePartition() {
     // ensure the overwrite results in a merge
     table.updateProperties().set(TableProperties.MANIFEST_MIN_MERGE_COUNT, "1").commit();
 
-    table.newFastAppend().appendFile(FILE_A).appendFile(FILE_B).commit();
+    table.newFastAppend().appendFile(FILE_A).appendFile(FILE_B).toBranch(branch).commit();
```
rdblue, line 125: Why not use the `commit` method in these tests?

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -39,18 +39,32 @@
...
-        .newRowDelta()
-        .addRows(FILE_A)
-        .addDeletes(FILE_A_DELETES)
-        .addDeletes(FILE_B_DELETES)
-        .commit();
+    SnapshotUpdate rowDelta =
+        table.newRowDelta().addRows(FILE_A).addDeletes(FILE_A_DELETES).addDeletes(FILE_B_DELETES);
 
-    Snapshot snap = table.currentSnapshot();
+    commit(table, rowDelta, branch);
```
rdblue, line 66: Odd spacing here. I'd expect the commit to come after rowDelta insteadd of being combined with the latestSnapshot call.

core/src/test/java/org/apache/iceberg/TestRowDelta.java
```diff
@@ -81,155 +95,171 @@ public void testAddDeleteFile() {
...
 
     Assert.assertEquals(
         "Table should not have any delete manifests",
         0,
-        table.currentSnapshot().deleteManifests(table.io()).size());
+        latestSnapshot(table, branch).deleteManifests(table.io()).size());
   }
 
-  @Test
+  // @Test
```
rdblue, line 238: @amogh-jahagirdar, why is this test disabled?

rdblue: This passes locally. I'm going to assume this was an accident and restore the test to get CI running it.

amogh-jahagirdar: yeah shouldn't be disabled, thanks!

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
```diff
@@ -397,8 +405,10 @@ protected void validateNoNewDeletesForDataFiles(
       TableMetadata base,
       Long startingSnapshotId,
       Expression dataFilter,
-      Iterable<DataFile> dataFiles) {
-    validateNoNewDeletesForDataFiles(base, startingSnapshotId, dataFilter, dataFiles, false);
+      Iterable<DataFile> dataFiles,
+      Snapshot parent) {
```
rdblue, line 409: Looks like this doesn't update Javadoc...

core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java
```diff
@@ -460,8 +471,8 @@ private void validateNoNewDeletesForDataFiles(
    * @param dataFilter an expression used to find new conflicting delete files
    */
   protected void validateNoNewDeleteFiles(
-      TableMetadata base, Long startingSnapshotId, Expression dataFilter) {
-    DeleteFileIndex deletes = addedDeleteFiles(base, startingSnapshotId, dataFilter, null);
+      TableMetadata base, Long startingSnapshotId, Expression dataFilter, Snapshot parent) {
```
rdblue, line 474: I think a lot of the `revapi` problems are probably related to adding `parent` to these methods. We should add the older version back that passes `base.currentSnapshot()` and mark it deprecated, to be removed in 1.3.0. That will take care of at least some of the breaking changes.

rdblue (commented):
@namrathamyske and @amogh-jahagirdar, this is looking great! There are no major blockers, just minor comments. I think that means we should move forward with fixing the revapi failures (add back methods and deprecate them). After that, we can get this in.

Thanks!

amogh-jahagirdar:
Thanks for the reviews @rdblue ! @namrathamyske I raised a PR to your branch for deprecating the old validation methods and updating rev API https://github.com/namrathamyske/iceberg/pull/21. When you get a chance please take a look!

core/src/main/java/org/apache/iceberg/BaseRowDelta.java
namrathamyske: @amogh-jahagirdar Is this even required anymore ? We are anyway doing ancestor check in https://github.com/apache/iceberg/blob/a1d994c131dc5b5f42e1693e92e058493382593e/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L742

rdblue (approved):

PR #5235, opened 2022-07-10
REST: Create commit catalog handler fix

bryanck:
Currently in the REST catalog handler, during a CTAS, the files will be written and associated with the partition spec ID from the initial create request. When the commit occurs, an empty table metadata object is created and updates are applied to that. However, applying the updates will increment the partition spec ID, so the files will be associated with the empty spec ID, and thus partition pruning will not be available for these files.

This PR adds the ability to build on top of an empty TableMetadata so the initial IDs will match the initial files. It also moves `TableMetadata.newTableMetadata` to use the builder to instantiate a new instance so it can reuse functionality from the builder and to be consistent with `buildReplacement()`.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
@@ -127,7 +127,6 @@ public class TableMetadata implements Serializable {
         .setDefaultSortOrder(freshSortOrder)
         .setLocation(location)
         .setProperties(properties)
-        .discardChanges()
         .build();
   }
 
@@ -1245,7 +1244,7 @@ public class TableMetadata implements Serializable {
           ImmutableList.copyOf(newSnapshotLog),
           ImmutableList.copyOf(metadataHistory),
           ImmutableMap.copyOf(refs),
-          discardChanges ? ImmutableList.of() : ImmutableList.copyOf(changes)
+          discardChanges || base == null ? ImmutableList.of() : ImmutableList.copyOf(changes)
       );
     }
 
--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
@@ -1012,10 +1012,12 @@ public class TableMetadata implements Serializable {
         return this;
       }
 
+      ValidationException.check(!schemas.isEmpty(), "Attempting to add a snapshot before a schema is added");
+      ValidationException.check(!specs.isEmpty(), "Attempting to add a snapshot before a partition spec is added");
+      ValidationException.check(!sortOrders.isEmpty(), "Attempting to add a snapshot before a sort order is added");
       ValidationException.check(!snapshotsById.containsKey(snapshot.snapshotId()),
           "Snapshot already exists for id: %s",
           snapshot.snapshotId());
-
       ValidationException.check(formatVersion == 1 || snapshot.sequenceNumber() > lastSequenceNumber,
           "Cannot add snapshot with sequence number %s older than last sequence number %s",
           snapshot.sequenceNumber(), lastSequenceNumber);
--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
@@ -120,13 +120,15 @@ public class TableMetadata implements Serializable {
     // break existing tables.
     MetricsConfig.fromProperties(properties).validateReferencedColumns(schema);
 
-    return new TableMetadata(null, formatVersion, UUID.randomUUID().toString(), location,
-        INITIAL_SEQUENCE_NUMBER, System.currentTimeMillis(),
-        lastColumnId.get(), freshSchema.schemaId(), ImmutableList.of(freshSchema),
-        freshSpec.specId(), ImmutableList.of(freshSpec), freshSpec.lastAssignedFieldId(),
-        freshSortOrderId, ImmutableList.of(freshSortOrder),
-        ImmutableMap.copyOf(properties), -1, ImmutableList.of(),
-        ImmutableList.of(), ImmutableList.of(), ImmutableMap.of(), ImmutableList.of());
+    return new Builder()
+        .upgradeFormatVersion(formatVersion)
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -985,10 +1012,12 @@ public Builder addSnapshot(Snapshot snapshot) {
         return this;
       }
 
+      ValidationException.check(!schemas.isEmpty(), "Attempting to add a snapshot before a schema is added");
+      ValidationException.check(!specs.isEmpty(), "Attempting to add a snapshot before a partition spec is added");
+      ValidationException.check(!sortOrders.isEmpty(), "Attempting to add a snapshot before a sort order is added");
```
bryanck, line 1017: When building metadata from empty, the table specs aren't initialized. These assertions ensure they are set before adding files, so files added during a transaction line up with the table specs.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -985,10 +1012,12 @@ public Builder addSnapshot(Snapshot snapshot) {
...
       ValidationException.check(!snapshotsById.containsKey(snapshot.snapshotId()),
           "Snapshot already exists for id: %s",
           snapshot.snapshotId());
-
```
rdblue, line 991: Nit: unnecessary change that could cause commit conflicts.

core/src/main/java/org/apache/iceberg/TableMetadata.java
```diff
@@ -1212,7 +1246,7 @@ public TableMetadata build() {
           ImmutableList.copyOf(newSnapshotLog),
           ImmutableList.copyOf(metadataHistory),
           ImmutableMap.copyOf(refs),
-          discardChanges ? ImmutableList.of() : ImmutableList.copyOf(changes)
+          discardChanges || base == null ? ImmutableList.of() : ImmutableList.copyOf(changes)
```
rdblue, line 1249: I'm not sure about this. If base is `null`, aren't all of the changes still valid?

Tests pass without this change, so I would probably not include it. What was the reason for adding this?

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -1213,6 +1213,67 @@ public void testCompleteCreateTransaction() {
...
+
+    create.commitTransaction();
+
+    Assert.assertTrue("Table should exist after append commit", catalog.tableExists(TABLE));
+    Table table = catalog.loadTable(TABLE);
+
+    Assert.assertEquals("Table schema should match the new schema",
+        newSchema.asStruct(), table.schema().asStruct());
+    Assert.assertEquals("Table should have create partition spec", newSpec.fields(), table.spec().fields());
+    Assert.assertEquals("Table should have create sort order", newSortOrder.fields(), table.sortOrder().fields());
```
rdblue, line 1267: Minor: I think it would make more sense to describe the new schema, spec, and sort order as "updated" rather than "create" because these are actually different than the created ones.

rdblue: Also, do we need to validate the IDs of each of these?

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
```diff
@@ -1213,6 +1213,67 @@ public void testCompleteCreateTransaction() {
...
+    Assert.assertEquals("Table properties should be a superset of the requested properties",
+        properties.entrySet(),
+        Sets.intersection(properties.entrySet(), table.properties().entrySet()));
+    if (!overridesRequestedLocation()) {
+      Assert.assertEquals("Table location should match requested", "file:/tmp/ns/table", table.location());
+    }
+    assertFiles(table, FILE_A, anotherFile);
+    assertFilePartitionSpec(table, FILE_A, 0);
+    assertFilePartitionSpec(table, anotherFile, 1);
```
rdblue, line 1276: I think it would be better to use `TABLE_SPEC.specId()` and `newSpec.specId()` rather than hard-coding the IDs.

core/src/main/java/org/apache/iceberg/TableMetadata.java
bryanck: `TestLoadTableResponse.testHasOnlyKnownFields()` was failing, so my intent was to make it behave like `newTableMetadata` which was setting changes to empty, though probably my misunderstanding. I'll remove this and fix what's going on with that test.

core/src/main/java/org/apache/iceberg/TableMetadata.java
bryanck: OK I'll put that back, was making spacing consistent with the validation checks

core/src/main/java/org/apache/iceberg/TableMetadata.java
bryanck: I made this change

core/src/main/java/org/apache/iceberg/TableMetadata.java
bryanck: This is done

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
bryanck: The spec ID isn't incremented at this point, so `newSpec.specId()` is zero

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
bryanck: I updated the message to `updated`. and added ID validation.

core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java
bryanck: I created constants for the expected IDs.

rdblue (approved):

rdblue:
Thanks, @bryanck! Looks great. And thanks for finding a better solution!

PR #5237, opened 2022-07-10
API: Add IcebergBuild info

rdblue:
This is based on `VersionProperties` from an old version of #5228 by @kbendick. This updates the class to use iceberg-build.properties in the root `build` folder, renames the class to `IcebergBuild`, and adds tests.

Co-authored-by: Kyle Bendickson <kyle@tabular.io>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,15 +45,6 @@ plugins {
 try {
   // apply these plugins in a try-catch block so that we can handle cases without .git directory
   apply plugin: 'com.palantir.git-version'
-  apply plugin: 'com.gorylenko.gradle-git-properties'
-  // git properties file for the root project for adding to the source tarball
-  gitProperties {
-    gitPropertiesName = 'iceberg-build.properties'
-    gitPropertiesResourceDir = file("${rootDir}/build")
-    extProperty = 'gitProps'
-    failOnNoGitDirectory = true
-  }
-  generateGitProperties.outputs.upToDateWhen { false }
 } catch (Exception e) {
   project.logger.error(e.getMessage())
 }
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -66,6 +57,28 @@ if (JavaVersion.current() == JavaVersion.VERSION_1_8) {
   throw new GradleException("This build must be run with JDK 8 or 11 but was executed with JDK " + JavaVersion.current())
 }
 
+apply plugin: 'com.gorylenko.gradle-git-properties'
```
kbendick, line 60: Ooo I see you were able to get rid of the try-catch. Good job! 👍

build.gradle
```diff
@@ -45,15 +45,6 @@ plugins {
...
   apply plugin: 'com.palantir.git-version'
-  apply plugin: 'com.gorylenko.gradle-git-properties'
-  // git properties file for the root project for adding to the source tarball
-  gitProperties {
-    gitPropertiesName = 'iceberg-build.properties'
-    gitPropertiesResourceDir = file("${rootDir}/build")
-    extProperty = 'gitProps'
-    failOnNoGitDirectory = true
-  }
-  generateGitProperties.outputs.upToDateWhen { false }
```
rdblue, line 56: This is moved out of the try/catch block because it is safe to add the plugin when `.git` is not present. The `fialOnNoGitDirectory` flag determines whether `generateGitProperties` will succeed if it isn't present. It is set to true here because the build should fail if there is no root `iceberg-build.properties` file and `.git` is not present.

build.gradle
```diff
@@ -66,6 +57,28 @@ if (JavaVersion.current() == JavaVersion.VERSION_1_8) {
...
+  tasks.register('buildInfo', Exec) {
+    project.logger.info('Using build info from iceberg-build.properties')
+    commandLine 'cp', "${rootDir}/iceberg-build.properties", 'build/iceberg-build.properties'
+  }
+} else {
+  tasks.register('buildInfo') {
+    project.logger.info('Generating iceberg-build.properties from git')
+    dependsOn generateGitProperties
+  }
+}
```
rdblue, line 80: The `buildInfo` task was moved here because API tests now depend on it.

kbendick (approved):
This LGTM.

build.gradle
kbendick: 👍  Good find. 

When I tried cloning with depth=1 and then `rm -rf .git/` tasks failed, but glad to see you were able to make it work.

nastra (approved):

PR #5238, opened 2022-07-11
Core - Add build information to all requests sent by HTTPClient as headers

kbendick:
The HTTPClientFactory is used to build the HTTP Client that will be used by the REST Catalog.

This PR injects the following two headers into each request, for debugging purposes:
-   X-Client-Version  
-   X-Client-Git-Commit-Short

To test for these headers, `TestHTTPClient` was updated to use the `HTTPClientFactory`. The headers, with their expected values, are placed on each expected request, which is an assertion.

Notice that we don't have to add the headers onto the request that is sent, as they are default headers for each request.

This PR needs relies on work in #5237 and needs to be rebased once that is merged.

### DIFF
--- a/core/src/test/java/org/apache/iceberg/rest/TestHTTPClient.java
+++ b/core/src/test/java/org/apache/iceberg/rest/TestHTTPClient.java
@@ -28,6 +28,8 @@ import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import org.apache.iceberg.AssertHelpers;
+import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.IcebergBuild;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.rest.responses.ErrorResponse;
 import org.apache.iceberg.rest.responses.ErrorResponseParser;
@@ -53,13 +55,17 @@ public class TestHTTPClient {
   private static final String URI = String.format("http://127.0.0.1:%d", PORT);
   private static final ObjectMapper MAPPER = RESTObjectMapper.mapper();
 
+  private static String icebergBuildGitCommitShort;
+  private static String icebergBuildFullVersion;
   private static ClientAndServer mockServer;
   private static RESTClient restClient;
 
   @BeforeClass
   public static void beforeClass() {
     mockServer = startClientAndServer(PORT);
-    restClient = HTTPClient.builder().uri(URI).build();
+    restClient = new HTTPClientFactory().apply(ImmutableMap.of(CatalogProperties.URI, URI));
+    icebergBuildGitCommitShort = IcebergBuild.gitCommitShortId();
+    icebergBuildFullVersion = IcebergBuild.fullVersion();
   }
 
   @AfterClass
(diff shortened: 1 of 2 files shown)


kbendick:
Here's a sample of the headers sent for each request, per the mock server tests:

Relevant portions:
`X-Client-Git-Commit-Short`: `"a850705"
`X-Client-Version`: `""Apache Iceberg 0.14.0-SNAPSHOT (commit a8507057f778efba6f1750ae814749f52a4cd7bc)"`

We can update these header names or which values are used (as well as where they're placed and how, but I think `HTTPClientFactory` makes sense).

The headers can technically be overridden via the config `header.X-Client-Git-Commit-Short` in the configuration, although we can code against that (make it a reserved property) if desired.

```
   "headers" : {
      "Accept" : [ "application/json" ],
      "Content-Type" : [ "application/json" ],
      "X-Client-Git-Commit-Short" : [ "a850705" ],
      "X-Client-Version" : [ "Apache Iceberg 0.14.0-SNAPSHOT (commit a8507057f778efba6f1750ae814749f52a4cd7bc)" ],
      "Authorization" : [ "Bearer auth_token" ],
      "Accept-Encoding" : [ "gzip, x-gzip, deflate" ],
      "Host" : [ "127.0.0.1:1080" ],
      "Connection" : [ "keep-alive" ],
      "User-Agent" : [ "Apache-HttpClient/5.1 (Java/11.0.15)" ],
      "content-length" : [ "0" ]
    }
```

core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java
```diff
@@ -110,7 +115,10 @@ public <T extends RESTResponse> T execute(RESTCatalogAdapter.HTTPMethod method,
 
     this.restCatalog = new RESTCatalog(context, (config) -> adaptor);
     restCatalog.setConf(conf);
-    restCatalog.initialize("prod", ImmutableMap.of(CatalogProperties.URI, "ignored", "credential", "catalog:12345"));
+    restCatalog.initialize("prod", ImmutableMap.of(
+        CatalogProperties.URI, "ignored", "credential", "catalog:12345",
+        String.format("header.%s", HTTPClientFactory.CLIENT_GIT_COMMIT_SHORT_HEADER), IcebergBuild.gitCommitShortId(),
+        String.format("header.%s", HTTPClientFactory.CLIENT_VERSION_HEADER), IcebergBuild.version()));
```
kbendick, line 121: These don't really need to be set because `HTTPClientFactory` isn't used here - the underlying JDBC client is what builds this.

We can however still override the version etc by sending back a configuration property of `header.X-Client-Version` with the desired value to be used as the value (though I don't see any reason to do that).

core/src/test/java/org/apache/iceberg/rest/TestHTTPClient.java
```diff
@@ -166,7 +172,9 @@ private static String addRequestTestCaseAndGetPath(HttpMethod method, Item body,
     String asJson = body != null ? MAPPER.writeValueAsString(body) : null;
     HttpRequest mockRequest = request("/" + path)
         .withMethod(method.name().toUpperCase(Locale.ROOT))
-        .withHeader("Authorization", "Bearer " + BEARER_AUTH_TOKEN);
+        .withHeader("Authorization", "Bearer " + BEARER_AUTH_TOKEN)
+        .withHeader(HTTPClientFactory.CLIENT_VERSION_HEADER, icebergBuildFullVersion)
+        .withHeader(HTTPClientFactory.CLIENT_GIT_COMMIT_SHORT_HEADER, icebergBuildGitCommitShort);
```
kbendick, line 177: We add the headers that we expect to see as assertions on the mock request. The actual `HTTPClient` is used, generated by the `HTTPClientFactory`, which injects the headers.

If requests are not generated which have these headers, they will not match and the assertion of this mock request will fail.

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #5242, opened 2022-07-11
Add InternalRecordWrapperFactory to generate InternalRecordWrapper

shidayang:
The cost of creating InternalRecord is very high, So the performance of MOR is low when have many delete files.


rdblue:
@shidayang, I fixed this in #5249 and made you a co-author. Thanks for identifying this!

shidayang:
@rdblue OK

PR #5244, opened 2022-07-11
Add StructLikeWrapperFactory to generate StructLikeWrapper

shidayang:
The cost of creating StructLikeWrapper is very high, So the performance of MOR is very low when have many delete files


shidayang:
@rdblue OK

PR #5249, opened 2022-07-11
Core: Fix wrapper performance for delete filters

rdblue:
This is an alternative implementation of #5242 and #5244. Rather than adding factory classes, this adds a `copyFor` method to `StructLikeWrapper` and `InternalRecordWrapper`. This is smaller than adding a factory and keeps the logic in one place (rather than duplicating it in a factory).

This updates cases where the wrappers are used to copy wrappers rather than create new ones. This avoids analyzing the type that the wrappers are created for and should speed up execution.


rdblue:
Thanks, @aokolnychyi!

PR #5255, opened 2022-07-12
Core: Fix REST catalog metadata table loading.

rdblue:
This fixes metadata table loading. Although the metadata table is correctly loaded, the `TableOperations` fails when it is first used because the path is based on the full table identifier (e.g. db.table.files) rather than the loaded table identifier (db.table). This fixes both the path and the table name.


bijanhoule (approved):
Thanks, @rdblue!

PR #5257, opened 2022-07-12
Spark 3.3: Expose action classes in SparkActions

aokolnychyi:
This is a newer version of #5108.


spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/BaseSnapshotUpdateSparkAction.java
```diff
@@ -20,20 +20,17 @@
 package org.apache.iceberg.spark.actions;
 
 import java.util.Map;
-import org.apache.iceberg.actions.SnapshotUpdate;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.spark.sql.SparkSession;
 
-abstract class BaseSnapshotUpdateSparkAction<ThisT, R>
-    extends BaseSparkAction<ThisT, R> implements SnapshotUpdate<ThisT, R> {
+abstract class BaseSnapshotUpdateSparkAction<ThisT> extends BaseSparkAction<ThisT> {
```
aokolnychyi, line 26: I had to drop interfaces like `SnapshotUpdate` from parent actions so that `ThisT` can be an action class.

aokolnychyi:
It is easier to review each commit in this PR individually.
We may even want to rebase and merge to keep the commit  history.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/BaseDeleteReachableFilesSparkAction.java
```diff
@@ -19,165 +19,21 @@
...
 /**
- * An implementation of {@link DeleteReachableFiles} that uses metadata tables in Spark
- * to determine which files should be deleted.
+ * An action that deletes reachable files from a given metadata file.
+ *
+ * @deprecated since 0.14.0, will be removed in 1.0.0;
+ *             use {@link SparkActions} and {@link DeleteReachableFilesSparkAction} instead.
  */
-@SuppressWarnings("UnnecessaryAnonymousClass")
-public class BaseDeleteReachableFilesSparkAction
```
rdblue, line 53: @aokolnychyi, Since 3.3 is unreleased, do we need to preserve the base classes?

aokolnychyi: Well, good question. It depends how we treat our Spark libs. If we treat them as independent, I think we can skip the old classes. Let me do that and keep the base classes for 3.2.

rdblue (approved):

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/procedures/ExpireSnapshotsProcedure.java
bknbkn, line 1: It seem should use ExpireSnapshotsSparkAction in code's 96 line?  I also find it in spark 3.5

PR #5258, opened 2022-07-12
More Python Expressions

CircArgs:
Adding EQ, LT, GT, Null, NaN - bound and unbound versions

Predicate literals validation e.g. ensure predicates that take a single literal ultimately end up as tuple of literals as `.literals`, make sure unary predicates have no literals.


python/pyiceberg/expressions/base.py
```diff
@@ -172,13 +178,33 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
     term: Reference[T]
-    literals: Tuple[Literal[T], ...]
+    literals: Union[Tuple[Literal[T], ...], Literal[T]]
+    literals_len_lb: int = field(init=False, repr=False, hash=False, compare=False, default=1)
+    literals_len_ub: Optional[int] = field(init=False, repr=False, hash=False, compare=False, default=1)
+
+    def __post_init__(self):
+        if not isinstance(self.literals, tuple):
+            object.__setattr__(self, "literals", (self.literals,))
+        if not (self.literals_len_lb <= len(self.literals) <= (self.literals_len_ub or float("inf"))):  # type: ignore
```
samredai, line 204: For the `literals_len_lb` and `literals_len_ub`, are there scenarios where we need this specificity or is it the case that there will be two kinds of predicates: those that take a single literal and those that take a set of literals of any size? If it's the latter, should we just handle that in the `__post_init__` methods for the concrete predicate classes?

```py
@dataclass(frozen=True)
class In(UnboundPredicate[T]):

    def __post_init__(self):
        if isinstance(self.literals, Literal):
            object.__setattr__(self, "literals", set(self.literals,))
```

...and do the reverse for predicates that require a single literal:

```py
@dataclass(frozen=True)
class EQ(UnboundPredicate[T]):

    def __post_init__(self):
        if isinstance(self.literals, Set):
            if len(self.literals) > 1:
                raise TypeError("...")
            object.__setattr__(self, "literals", self.literals[0])
```

python/pyiceberg/expressions/base.py
```diff
@@ -172,13 +178,33 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
+                f"{self.__class__} expects a number of literals between at least {self.literals_len_lb} and at most {self.literals_len_ub or float('inf')}"
+            )
 
 
 @dataclass(frozen=True)  # type: ignore[misc]
 class UnboundPredicate(Unbound[T, BooleanExpression], BooleanExpression):
     term: Reference[T]
-    literals: Tuple[Literal[T], ...]
+    literals: Union[Tuple[Literal[T], ...], Literal[T]]
+    literals_len_lb: int = field(init=False, repr=False, hash=False, compare=False, default=1)
```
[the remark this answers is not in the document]
CircArgs: I figured `Optional` would imply it could be `None` (would be equivalent to `int | None` which I don't think makes sense

python/pyiceberg/expressions/base.py
CircArgs: yeah maybe I was playing too much as I tend to

python/pyiceberg/expressions/base.py
```diff
@@ -327,6 +347,54 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundIn[T]:
...
+
+
+@dataclass(frozen=True)
+class BoundLT(BoundPredicate[T]):
+    def __invert__(self):
+        raise TypeError("LT expressions do not support negation.")
+
+
+@dataclass(frozen=True)
+class LT(UnboundPredicate[T]):
```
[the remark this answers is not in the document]
CircArgs: still open to this. currently changed to `Lt` and the like

python/pyiceberg/expressions/base.py
CircArgs: ditched this idea in favor of an obfuscated `_validate_literals` method that defaults to the common pattern of checking for having exactly one literal otherwise children override the validation

python/pyiceberg/expressions/base.py
```diff
@@ -169,16 +171,64 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
         return BoundReference(field=field, accessor=accessor)
 
 
-@dataclass(frozen=True)  # type: ignore[misc]
 class BoundPredicate(Bound[T], BooleanExpression):
```
Fokko, line 174: For readability and type checking it is best to also define the variables in the class itself:
```python
class BoundPredicate(Bound[T], BooleanExpression):
    _term: BoundTerm[T]
    _literals: Tuple[Literal[T]]
```
This will also remove the ignore below:
```python
@property
def literals(self) -> tuple[Literal[T], ...]:
    return self._literals  # type: ignore
```

python/pyiceberg/expressions/base.py
```diff
@@ -169,16 +171,64 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
-    term: BoundReference[T]
-    literals: Tuple[Literal[T], ...]
+    def __init__(self, term: BoundTerm[T], literals: tuple[Literal[T], ...] | Literal[T] | None = None):
```
Fokko, line 175: What do you think of just accumulating the positional arguments into a list of literals?
```suggestion
    def __init__(self, term: BoundTerm[T], *literals: Literal[T]):
```

Fokko (changes requested):
Thanks @CircArgs for working on this, much appreciated! I do have some comments on the code.

python/pyiceberg/expressions/base.py
CircArgs: Done!

python/pyiceberg/expressions/base.py
CircArgs: I like it! I may be misremembering. I thought I had gotten pushback in a similar situation using `*args` but let me have another go

python/pyiceberg/expressions/base.py
```diff
@@ -169,16 +171,64 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
+        self._term = term
+        self._literals = literals if isinstance(literals, tuple) else (literals and (literals,))
+        self._validate_literals()
+
+    def _validate_literals(self):
+        if self.literals is None or len(self.literals) != 1:
```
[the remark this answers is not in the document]
CircArgs: @Fokko  Originally, this was not optional, but was made so to account for the Unary expressions for example `IsNaN`

python/pyiceberg/expressions/base.py
```diff
@@ -169,16 +171,64 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
+        return f"{self.__class__.__name__}({repr(self.term)}{self.literals and ', '+repr(self.literals)})"
+
+    def __eq__(self, other) -> bool:
+        return id(self) == id(other) or (
+            type(self) == type(other) and self.term == other.term and self.literals == other.literals
+        )
 
 
-@dataclass(frozen=True)  # type: ignore[misc]
 class UnboundPredicate(Unbound[T, BooleanExpression], BooleanExpression):
```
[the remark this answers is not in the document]
CircArgs: does `UnboundPredicate` need it? I think it's abstract still. It lacks `__invert__` too

python/pyiceberg/expressions/base.py
```diff
@@ -169,16 +171,64 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
+    def literals(self) -> tuple[Literal[T], ...]:
+        return self._literals  # type: ignore
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}({str(self.term)}{self.literals and ', '+str(self.literals)})"
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({repr(self.term)}{self.literals and ', '+repr(self.literals)})"
+
+    def __eq__(self, other) -> bool:
```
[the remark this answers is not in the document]
CircArgs: I originally used dataclasses throughout a lot of this code, but had difficulties getting things to play nice with the validation of literals. I think the trouble comes from leveraging `__post_init__` in a similar way to how `__init__` is currently, but it seems child classes don't care about the parent's `__post_init__` hence why I switched to `__init__` and just work things with `super`

python/pyiceberg/expressions/base.py
Fokko: I think it would be nice for passing in the literal(s). We also use it in other places: https://github.com/apache/iceberg/blob/master/python/pyiceberg/schema.py#L63-L67 So I think we should be okay here.

python/pyiceberg/expressions/base.py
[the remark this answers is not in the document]
Fokko: Then we should mark it as an abstract class, or we can create stubs for the methods, and raise NotImplementedError: https://docs.python.org/3/library/exceptions.html#NotImplementedError. PyCharm is bugging me that the classes are missing :)

python/pyiceberg/expressions/base.py
CircArgs: @Fokko Done!

python/pyiceberg/expressions/base.py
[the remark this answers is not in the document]
Fokko: A missed that one, thanks! Again, I would suggest changing this into:
```python
    def _validate_literals(self):  # pylint: disable=W0238
        if self.literals is not None:
            raise AttributeError("IsNaN is a unary predicate and takes no Literals.")
```
into:
```
    def _validate_literals(self):  # pylint: disable=W0238
        if self.literals:
            raise AttributeError("IsNaN is a unary predicate and takes no Literals.")
```
This way we raise the error on an empty list.

python/pyiceberg/expressions/base.py
```diff
@@ -172,13 +178,33 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
 @dataclass(frozen=True)  # type: ignore[misc]
 class UnboundPredicate(Unbound[T, BooleanExpression], BooleanExpression):
     term: Reference[T]
-    literals: Tuple[Literal[T], ...]
+    literals: Union[Tuple[Literal[T], ...], Literal[T]]
+    literals_len_lb: int = field(init=False, repr=False, hash=False, compare=False, default=1)
+    literals_len_ub: Optional[int] = field(init=False, repr=False, hash=False, compare=False, default=1)
+
+    def __post_init__(self):
+        if not isinstance(self.literals, tuple):
```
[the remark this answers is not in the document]
CircArgs: no longer relevant

python/pyiceberg/expressions/base.py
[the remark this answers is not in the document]
CircArgs: I did this while I was changing literals to an *args pattern :p

CircArgs:
@Fokko please have another look when you get a chance. Thanks!

python/pyiceberg/expressions/base.py
```diff
@@ -169,16 +171,70 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundReference[T]:
...
+    @property
+    def term(self) -> BoundTerm[T]:
+        return self._term
+
+    @property
+    def literals(self) -> tuple[Literal[T], ...]:
+        return self._literals
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}({str(self.term)}{self.literals and ', '+str(self.literals)})"
```
rdblue, line 196: When I test this and `literals` is an empty list, I get `x[]` (where `x` is the term)

python/pyiceberg/expressions/base.py
```diff
@@ -209,14 +265,14 @@ def right(self) -> BooleanExpression:
...
-    def __invert__(self) -> "Or":
+    def __invert__(self) -> Or:
         return Or(~self.left, ~self.right)
 
     def __repr__(self) -> str:
         return f"And({repr(self.left)}, {repr(self.right)})"
 
     def __str__(self) -> str:
-        return f"({self.left} and {self.right})"
+        return f"And({str(self.left)}, {str(self.right)})"
```
rdblue, line 275: I don't think this is correct. This is basically just `repr` output. I think this should match what Java produces, which is a readable infix string.

python/pyiceberg/expressions/base.py
```diff
@@ -282,49 +338,239 @@ def __repr__(self) -> str:
...
+class IsNull(UnboundPredicate[T]):
+    def __invert__(self) -> NotNull:
+        return NotNull(self.term)
+
+    def _validate_literals(self):  # pylint: disable=W0238
+        if self.literals is not None:
+            raise AttributeError("Null is a unary predicate and takes no Literals.")
+
+    def bind(self, schema: Schema, case_sensitive: bool) -> BoundIsNull[T]:
+        bound_ref = self.term.bind(schema, case_sensitive)
```
rdblue, line 369: A better name would be `bound_term` since this is not necessarily a ref.

python/pyiceberg/expressions/base.py
```diff
@@ -282,49 +338,239 @@ def __repr__(self) -> str:
...
+        return BoundNotIn(bound_ref, *tuple(lit.to(bound_ref.field.field_type) for lit in self.literals))  # type: ignore
+
+
+class BoundEq(BoundPredicate[T]):
+    def __invert__(self) -> BoundNotEq[T]:
+        return BoundNotEq(self.term, *self.literals)
+
+
+class Eq(UnboundPredicate[T]):
```
rdblue, line 497: Since we are using expression classes directly, rather than using factory methods, I think that these classes should use full names, like `Equals` and `LessThan` (or `BoundEqual` and `BoundLessThan`).

rdblue (approved):

rdblue:
Thanks, @CircArgs! I think there are some changes that would help readability, but those are minor. I merged this so that other work can continue in parallel.

CircArgs:
> Thanks, @CircArgs! I think there are some changes that would help readability, but those are minor. I merged this so that other work can continue in parallel.

Thanks @rdblue. I'll open another PR soon with the changes hopefully

PR #5259, opened 2022-07-12
[API] Fix throwing null exception and add error prone suppression to runSafely functions

kbendick:
The `runSafely` function gives an error prone warning for throwing from a finally block.

Originally, we fixed this via suppression in https://github.com/apache/iceberg/pull/5190, but I noticed this function was unused and so chose to deprecate it instead in https://github.com/apache/iceberg/pull/5205.

@rdblue wanted to keep the function, as it's tested and could be useful in the future.

So this PR:
- fixes the issue discovered in #5190 of throwing `failure` when it's null
- attempts to throw `failure` after adding `e` as a suppressed error when it's non-null
- adds the `@Finally` suppression to remove the error prone log from the build.

Warning that is logged on build:
```
iceberg/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java:124: warning: [Finally] If you return or throw from a finally, then values returned or thrown from the try-catch block will be ignored. Consider using try-with-resources instead.
            throw new RuntimeException("Unknown exception in finally block", failure);
            ^
    (see https://errorprone.info/bugpattern/Finally)
```

### DIFF
--- a/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java
+++ b/api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java
@@ -78,6 +78,7 @@ public class ExceptionUtil {
     return runSafely(block, catchBlock, finallyBlock, e1Class, e2Class, RuntimeException.class);
   }
 
+  @SuppressWarnings("Finally")
   public static <R, E1 extends Exception, E2 extends Exception, E3 extends Exception> R runSafely(
       Block<R, E1, E2, E3> block,
       CatchBlock catchBlock,
@@ -113,15 +114,15 @@ public class ExceptionUtil {
         try {
           finallyBlock.run();
         } catch (Exception e) {
-          LOG.warn("Suppressing failure in finally block", e);
           if (failure != null) {
+            LOG.warn("Suppressing failure in finally block", e);
             failure.addSuppressed(e);
           } else {
             tryThrowAs(e, e1Class);
             tryThrowAs(e, e2Class);
             tryThrowAs(e, e3Class);
             tryThrowAs(e, RuntimeException.class);
-            throw new RuntimeException("Unknown exception in finally block", failure);
+            throw new RuntimeException("Unknown exception in finally block", e);
           }
         }
       }


api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java
```diff
@@ -113,15 +114,20 @@ public static <R, E1 extends Exception, E2 extends Exception, E3 extends Excepti
         try {
           finallyBlock.run();
         } catch (Exception e) {
-          LOG.warn("Suppressing failure in finally block", e);
           if (failure != null) {
+            LOG.warn("Suppressing failure in finally block", e);
```
kbendick, line 118: I moved this log because we're not suppressing `e` in the other block.

api/src/main/java/org/apache/iceberg/util/ExceptionUtil.java
```diff
@@ -113,15 +114,20 @@ public static <R, E1 extends Exception, E2 extends Exception, E3 extends Excepti
...
             failure.addSuppressed(e);
+            tryThrowAs(failure, e1Class);
```
dimas-b, line 120: This is probably not necessary and actually goes against the ErrorProne guideline of not throwing from `finally`.

The only way `failure` can be non-null here is if it was thrown from the main `try` block, in which case the generated bytecode will ensure it is re-thrown when `finally` completes.

rdblue: In this case, `failure` is already thrown because it is thrown in the `catch` block. `tryThrowAs` needs to be removed here. The logic was correct before, except for the wrong exception in the next block.

kbendick: TIL. Will remove. Thanks guys.

kbendick: Oh actually yeah I understand. The `failure` would have already been thrown, just the finally block gets run and we add one more suppression to it. So we don't need to throw it again. Thanks for the explanation.

rdblue (approved):

rdblue:
Thanks, @kbendick!

PR #5264, opened 2022-07-13
Cache PositionIndex

shidayang:
Like #5195  PositionIndex also can be cache

### DIFF
--- a/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
+++ b/data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
@@ -222,17 +222,17 @@ public abstract class DeleteFilter<T> {
       return records;
     }
 
-    List<CloseableIterable<Record>> deletes = Lists.transform(posDeletes, this::openPosDeletes);
-
     // if there are fewer deletes than a reasonable number to keep in memory, use a set
     if (posDeletes.stream().mapToLong(DeleteFile::recordCount).sum() < setFilterThreshold) {
       if (deleteRowPositions == null) {
+        List<CloseableIterable<Record>> deletes = Lists.transform(posDeletes, this::openPosDeletes);
         deleteRowPositions = Deletes.toPositionIndex(filePath, deletes);
       }
       Predicate<T> isDeleted = record -> deleteRowPositions.isDeleted(pos(record));
       return createDeleteIterable(records, isDeleted);
     }
 
+    List<CloseableIterable<Record>> deletes = Lists.transform(posDeletes, this::openPosDeletes);
     return hasIsDeletedColumn ?
         Deletes.streamingMarker(records, this::pos, Deletes.deletePositions(filePath, deletes), this::markRowDeleted) :
         Deletes.streamingFilter(records, this::pos, Deletes.deletePositions(filePath, deletes));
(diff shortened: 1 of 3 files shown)


shidayang:
@rdblue cc

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,8 +226,10 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
 
     // if there are fewer deletes than a reasonable number to keep in memory, use a set
     if (posDeletes.stream().mapToLong(DeleteFile::recordCount).sum() < setFilterThreshold) {
-      PositionDeleteIndex positionIndex = Deletes.toPositionIndex(filePath, deletes);
-      Predicate<T> isDeleted = record -> positionIndex.isDeleted(pos(record));
+      if (deleteRowPositions == null) {
+        deleteRowPositions = Deletes.toPositionIndex(filePath, deletes);
+      }
+      Predicate<T> isDeleted = record -> deleteRowPositions.isDeleted(pos(record));
```
kbendick, line 232: If `deleteRowPositions` is null, is this predicate at risk of an NPE?

shidayang: No NPE. The return of " Deletes.toPositionIndex(filePath, deletes)" must not be null.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,8 +226,10 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
 
     // if there are fewer deletes than a reasonable number to keep in memory, use a set
     if (posDeletes.stream().mapToLong(DeleteFile::recordCount).sum() < setFilterThreshold) {
```
lhofhansl, line 228: What about the else case? In that case we will not cache.

rdblue: Yes, it won't cache because the deletes are streamed from files.

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
```diff
@@ -226,8 +226,10 @@ private CloseableIterable<T> applyPosDeletes(CloseableIterable<T> records) {
...
-      PositionDeleteIndex positionIndex = Deletes.toPositionIndex(filePath, deletes);
-      Predicate<T> isDeleted = record -> positionIndex.isDeleted(pos(record));
+      if (deleteRowPositions == null) {
```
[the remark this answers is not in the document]
shidayang: I get you

data/src/main/java/org/apache/iceberg/data/DeleteFilter.java
shidayang: Yes, Because it use sort merge to avoid loading too many delete rows in memory.

shidayang: But, They should't reopen file and read file from head, they should read file continue。I will do some change here

PR #5268, opened 2022-07-13
API/Core: Initial Table Scan Reporting support

nastra:
The idea here is that we collect metrics during Table scans so that we
can better understand what exactly happens during a Table scan.

This is just an initial implementation that gathers/records a few metrics. The idea is to get early feedback on this approach.

## Overview of proposed changes

* a `ScanReport` that contains all relevant metrics from a Scan
* a `ScanReporter` interface that can be used to collect and report different metrics during a Table scant. Engines can implement this API to customize how a `ScanReport` is being reported
* a `LoggingScanReporter` is introduced that logs a `ScanReport` to the log file. We can think about defaulting to that one or using a **no-op** scan reporter
* a new Catalog Property `scan-reporter-impl` is introduced, allowing a Catalog to specify which `ScanReporter` to use
* a `ScanMetrics` that carries all relevant metrics during a Table scan. This class should be seen as something that is only used during a scan. At the end of a scan we'll want to create a `ScanReport` based on the collected metrics
* a `Timer` and a `DefaultTimer` were introduced so that we can measure spent time for a table scan
* some other changes, such as `DefaultScanMetrics` that use native Java Integer/Long counters and a `DefaultTimer`

### DIFF
--- a/api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
+++ b/api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
@@ -26,6 +26,7 @@ import java.util.NoSuchElementException;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import org.apache.iceberg.exceptions.RuntimeIOException;
+import org.apache.iceberg.metrics.MetricsContext;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
 
@@ -73,6 +74,36 @@ public interface CloseableIterable<T> extends Iterable<T>, Closeable {
     };
   }
 
+  /**
+   * Will run the given runnable when {@link CloseableIterable#close()} has been called.
+   *
+   * @param iterable The underlying {@link CloseableIterable} to iterate over
+   * @param onCompletionRunnable The runnable to run after the underlying iterable was closed
+   * @param <E> The type of the underlying iterable
+   * @return A new {@link CloseableIterable} where the runnable will be executed as the final step
+   *     after {@link CloseableIterable#close()} has been called
+   */
+  static <E> CloseableIterable<E> whenComplete(
+      CloseableIterable<E> iterable, Runnable onCompletionRunnable) {
+    Preconditions.checkNotNull(
+        onCompletionRunnable, "Cannot execute a null Runnable after completion");
+    return new CloseableIterable<E>() {
+      @Override
+      public void close() throws IOException {
+        try {
+          iterable.close();
+        } finally {
+          onCompletionRunnable.run();
+        }
+      }
+
+      @Override
+      public CloseableIterator<E> iterator() {
+        return CloseableIterator.withClose(iterable.iterator());
+      }
+    };
+  }
+
   static <E> CloseableIterable<E> filter(CloseableIterable<E> iterable, Predicate<E> pred) {
     return combine(
         () ->
@@ -83,6 +114,65 @@ public interface CloseableIterable<T> extends Iterable<T>, Closeable {
               }
             },
         iterable);
+  }
+
+  /**
+   * Filters the given {@link CloseableIterable} and counts the number of elements that do not match
+   * the predicate by incrementing the {@link MetricsContext.Counter}.
+   *
+   * @param skipCounter The {@link MetricsContext.Counter} instance to increment on each skipped
+   *     item during filtering.
+   * @param iterable The underlying {@link CloseableIterable} to filter.
+   * @param <E> The underlying type to be iterated.
+   * @return A filtered {@link CloseableIterable} where the given skipCounter is incremented
+   *     whenever the predicate does not match.
+   */
+  static <E> CloseableIterable<E> filter(
+      MetricsContext.Counter<?> skipCounter, CloseableIterable<E> iterable, Predicate<E> pred) {
+    Preconditions.checkArgument(null != skipCounter, "Invalid counter: null");
+    Preconditions.checkArgument(null != iterable, "Invalid iterable: null");
+    Preconditions.checkArgument(null != pred, "Invalid predicate: null");
+    return combine(
+        () ->
+            new FilterIterator<E>(iterable.iterator()) {
+              @Override
+              protected boolean shouldKeep(E item) {
+                boolean matches = pred.test(item);
+                if (!matches) {
+                  skipCounter.increment();
+                }
+                return matches;
+              }
+            },
+        iterable);
+  }
+
+  /**
+   * Counts the number of elements in the given {@link CloseableIterable} by incrementing the {@link
+   * MetricsContext.Counter} instance for each {@link Iterator#next()} call.
+   *
+   * @param counter The {@link MetricsContext.Counter} instance to increment on each {@link
+   *     Iterator#next()} call.
+   * @param iterable The underlying {@link CloseableIterable} to count
+   * @param <T> The underlying type to be iterated.
+   * @return A {@link CloseableIterable} that increments the given counter on each {@link
+   *     Iterator#next()} call.
+   */
+  static <T> CloseableIterable<T> count(
+      MetricsContext.Counter<?> counter, CloseableIterable<T> iterable) {
+    Preconditions.checkArgument(null != counter, "Invalid counter: null");
+    Preconditions.checkArgument(null != iterable, "Invalid iterable: null");
+    return new CloseableIterable<T>() {
+      @Override
+      public CloseableIterator<T> iterator() {
+        return CloseableIterator.count(counter, iterable.iterator());
+      }
+
+      @Override
+      public void close() throws IOException {
+        iterable.close();
+      }
+    };
   }
 
   static <I, O> CloseableIterable<O> transform(
--- a/api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
@@ -27,6 +27,25 @@ import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
  * {@link Integer} to count events.
  */
 class IntCounter implements MetricsContext.Counter<Integer> {
+  static final IntCounter NOOP =
+      new IntCounter("NOOP", MetricsContext.Unit.UNDEFINED) {
+        @Override
+        public void increment() {}
+
+        @Override
+        public void increment(Integer amount) {}
+
+        @Override
+        public Optional<Integer> count() {
+          return Optional.of(value());
+        }
+
+        @Override
+        public Integer value() {
+          return 0;
+        }
+      };
+
   private final AtomicInteger counter;
   private final String name;
   private final MetricsContext.Unit unit;
--- a/api/src/main/java/org/apache/iceberg/metrics/LoggingScanReporter.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/LoggingScanReporter.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A default {@link ScanReporter} implementation that logs the {@link ScanReport} to the log file.
+ */
+public class LoggingScanReporter implements ScanReporter {
+  private static final Logger LOG = LoggerFactory.getLogger(LoggingScanReporter.class);
+
+  @Override
+  public void reportScan(ScanReport scanReport) {
+    Preconditions.checkArgument(null != scanReport, "Invalid scan report: null");
+    LOG.info("Completed scan planning: {}", scanReport);
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/metrics/LongCounter.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/LongCounter.java
@@ -27,6 +27,25 @@ import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
  * {@link Long} to count events.
  */
 class LongCounter implements MetricsContext.Counter<Long> {
+  static final LongCounter NOOP =
+      new LongCounter("NOOP", MetricsContext.Unit.UNDEFINED) {
+        @Override
+        public void increment() {}
+
+        @Override
+        public void increment(Long amount) {}
+
+        @Override
+        public Optional<Long> count() {
+          return Optional.of(value());
+        }
+
+        @Override
+        public Long value() {
+          return 0L;
+        }
+      };
+
   private final AtomicLong counter;
   private final String name;
   private final MetricsContext.Unit unit;
--- a/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
@@ -135,13 +135,16 @@ public interface MetricsContext extends Serializable {
 
       @Override
       public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit unit) {
-        return new Counter<T>() {
-          @Override
-          public void increment() {}
+        if (Integer.class.equals(type)) {
+          return (Counter<T>) IntCounter.NOOP;
+        }
 
-          @Override
-          public void increment(T amount) {}
-        };
+        if (Long.class.equals(type)) {
+          return (Counter<T>) LongCounter.NOOP;
+        }
+
+        throw new IllegalArgumentException(
+            String.format("Counter for type %s is not supported", type.getName()));
       }
     };
   }
--- a/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
@@ -1,0 +1,216 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+import java.io.Serializable;
+import java.util.concurrent.TimeUnit;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.metrics.MetricsContext.Counter;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/** A Table Scan report that contains all relevant information from a Table Scan. */
+public class ScanReport implements Serializable {
+
+  private final String tableName;
+  private final long snapshotId;
+  private final Expression filter;
+  private final Schema projection;
+  private final ScanMetrics scanMetrics;
+
+  private ScanReport(
+      String tableName,
+      long snapshotId,
+      Expression filter,
+      Schema projection,
+      ScanMetrics scanMetrics) {
+    this.tableName = tableName;
+    this.snapshotId = snapshotId;
+    this.filter = filter;
+    this.projection = projection;
+    this.scanMetrics = scanMetrics;
+  }
+
+  public String tableName() {
+    return tableName;
+  }
+
+  public long snapshotId() {
+    return snapshotId;
+  }
+
+  public Expression filter() {
+    return filter;
+  }
+
+  public Schema projection() {
+    return projection;
+  }
+
+  public ScanMetrics scanMetrics() {
+    return scanMetrics;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("tableName", tableName)
+        .add("snapshotId", snapshotId)
+        .add("filter", filter)
+        .add("projection", projection)
+        .add("scanMetrics", scanMetrics)
+        .toString();
+  }
+
+  public static class Builder {
+    private String tableName;
+    private long snapshotId = -1L;
+    private Expression filter;
+    private Schema projection;
+    private ScanMetrics scanMetrics;
+
+    private Builder() {}
+
+    public Builder withTableName(String newTableName) {
+      this.tableName = newTableName;
+      return this;
+    }
+
+    public Builder withSnapshotId(long newSnapshotId) {
+      this.snapshotId = newSnapshotId;
+      return this;
+    }
+
+    public Builder withFilter(Expression newFilter) {
+      this.filter = newFilter;
+      return this;
+    }
+
+    public Builder withProjection(Schema newProjection) {
+      this.projection = newProjection;
+      return this;
+    }
+
+    public Builder fromScanMetrics(ScanMetrics newScanMetrics) {
+      this.scanMetrics = newScanMetrics;
+      return this;
+    }
+
+    public ScanReport build() {
+      Preconditions.checkArgument(null != tableName, "Invalid table name: null");
+      Preconditions.checkArgument(null != filter, "Invalid expression filter: null");
+      Preconditions.checkArgument(null != projection, "Invalid schema projection: null");
+      Preconditions.checkArgument(null != scanMetrics, "Invalid scan metrics: null");
+      return new ScanReport(tableName, snapshotId, filter, projection, scanMetrics);
+    }
+  }
+
+  /** Carries all metrics for a particular scan */
+  public static class ScanMetrics {
+    public static final ScanMetrics NOOP = new ScanMetrics(MetricsContext.nullMetrics());
+    private final Timer totalPlanningDuration;
+    private final Counter<Integer> resultDataFiles;
+    private final Counter<Integer> resultDeleteFiles;
+    private final Counter<Integer> totalDataManifests;
+    private final Counter<Integer> totalDeleteManifests;
+    private final Counter<Integer> scannedDataManifests;
+    private final Counter<Integer> skippedDataManifests;
+    private final Counter<Long> totalFileSizeInBytes;
+    private final Counter<Long> totalDeleteFileSizeInBytes;
+
+    public ScanMetrics(MetricsContext metricsContext) {
+      Preconditions.checkArgument(null != metricsContext, "Invalid metrics context: null");
+      this.totalPlanningDuration =
+          metricsContext.timer("totalPlanningDuration", TimeUnit.NANOSECONDS);
+      this.resultDataFiles =
+          metricsContext.counter("resultDataFiles", Integer.class, MetricsContext.Unit.COUNT);
+      this.resultDeleteFiles =
+          metricsContext.counter("resultDeleteFiles", Integer.class, MetricsContext.Unit.COUNT);
+      this.scannedDataManifests =
+          metricsContext.counter("scannedDataManifests", Integer.class, MetricsContext.Unit.COUNT);
+      this.totalDataManifests =
+          metricsContext.counter("totalDataManifests", Integer.class, MetricsContext.Unit.COUNT);
+      this.totalDeleteManifests =
+          metricsContext.counter("totalDeleteManifests", Integer.class, MetricsContext.Unit.COUNT);
+      this.totalFileSizeInBytes =
+          metricsContext.counter("totalFileSizeInBytes", Long.class, MetricsContext.Unit.BYTES);
+      this.totalDeleteFileSizeInBytes =
+          metricsContext.counter(
+              "totalDeleteFileSizeInBytes", Long.class, MetricsContext.Unit.BYTES);
+      this.skippedDataManifests =
+          metricsContext.counter("skippedDataManifests", Integer.class, MetricsContext.Unit.COUNT);
+    }
+
+    public Timer totalPlanningDuration() {
+      return totalPlanningDuration;
+    }
+
+    public Counter<Integer> resultDataFiles() {
+      return resultDataFiles;
+    }
+
+    public Counter<Integer> resultDeleteFiles() {
+      return resultDeleteFiles;
+    }
+
+    public Counter<Integer> scannedDataManifests() {
+      return scannedDataManifests;
+    }
+
+    public Counter<Integer> totalDataManifests() {
+      return totalDataManifests;
+    }
+
+    public Counter<Integer> totalDeleteManifests() {
+      return totalDeleteManifests;
+    }
+
+    public Counter<Long> totalFileSizeInBytes() {
+      return totalFileSizeInBytes;
+    }
+
+    public Counter<Long> totalDeleteFileSizeInBytes() {
+      return totalDeleteFileSizeInBytes;
+    }
+
+    public Counter<Integer> skippedDataManifests() {
+      return skippedDataManifests;
+    }
+
+    @Override
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .addValue(totalPlanningDuration)
+          .addValue(resultDataFiles)
+          .addValue(resultDeleteFiles)
+          .addValue(scannedDataManifests)
+          .addValue(skippedDataManifests)
+          .addValue(totalDataManifests)
+          .addValue(totalDeleteManifests)
+          .addValue(totalFileSizeInBytes)
+          .addValue(totalDeleteFileSizeInBytes)
+          .toString();
+    }
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/metrics/ScanReporter.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/ScanReporter.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+import org.apache.iceberg.metrics.ScanReport.ScanMetrics;
+
+/**
+ * This interface defines the basic API for a Table Scan Reporter that can be used to report
+ * different metrics after a Table scan is done.
+ */
+@FunctionalInterface
+public interface ScanReporter {
+
+  /**
+   * Indicates that a Scan is done by reporting a {@link ScanReport}. A {@link ScanReport} is
+   * usually directly derived from a {@link ScanMetrics} instance.
+   *
+   * @param scanReport The {@link ScanReport} to report.
+   */
+  void reportScan(ScanReport scanReport);
+}
--- a/api/src/test/java/org/apache/iceberg/io/TestCloseableIterable.java
+++ b/api/src/test/java/org/apache/iceberg/io/TestCloseableIterable.java
@@ -19,12 +19,16 @@
 package org.apache.iceberg.io;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
 import org.apache.iceberg.AssertHelpers;
 import org.apache.iceberg.io.TestableCloseableIterable.TestableCloseableIterator;
+import org.apache.iceberg.metrics.DefaultMetricsContext;
+import org.apache.iceberg.metrics.MetricsContext;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.assertj.core.api.Assertions;
@@ -94,6 +98,71 @@ public class TestCloseableIterable {
   }
 
   @Test
+  public void testWithCompletionRunnable() throws IOException {
+    AtomicInteger completionCounter = new AtomicInteger(0);
+    List<Integer> items = Lists.newArrayList(1, 2, 3, 4, 5);
+    Assertions.assertThatThrownBy(
+            () -> CloseableIterable.whenComplete(CloseableIterable.combine(items, () -> {}), null))
+        .isInstanceOf(NullPointerException.class)
+        .hasMessage("Cannot execute a null Runnable after completion");
+
+    try (CloseableIterable<Integer> iter =
+        CloseableIterable.whenComplete(
+            CloseableIterable.combine(items, () -> {}), completionCounter::incrementAndGet)) {
+      iter.forEach(val -> Assertions.assertThat(completionCounter.get()).isEqualTo(0));
+    }
+    Assertions.assertThat(completionCounter.get()).isEqualTo(1);
+  }
+
+  @Test
+  public void testWithCompletionRunnableAndEmptyIterable() throws IOException {
+    AtomicInteger completionCounter = new AtomicInteger(0);
+    CloseableIterable<Integer> empty = CloseableIterable.empty();
+    try (CloseableIterable<Integer> iter =
+        CloseableIterable.whenComplete(
+            CloseableIterable.combine(empty, () -> {}), completionCounter::incrementAndGet)) {
+      iter.forEach(val -> Assertions.assertThat(completionCounter.get()).isEqualTo(0));
+    }
+    Assertions.assertThat(completionCounter.get()).isEqualTo(1);
+  }
+
+  @Test
+  public void testWithCompletionRunnableAndUnclosedIterable() {
+    AtomicInteger completionCounter = new AtomicInteger(0);
+    List<Integer> items = Lists.newArrayList(1, 2, 3, 4, 5);
+    CloseableIterable<Integer> iter =
+        CloseableIterable.whenComplete(
+            CloseableIterable.combine(items, () -> {}), completionCounter::incrementAndGet);
+    iter.forEach(val -> Assertions.assertThat(completionCounter.get()).isEqualTo(0));
+    // given that we never close iter, the completionRunnable is never called
+    Assertions.assertThat(completionCounter.get()).isEqualTo(0);
+  }
+
+  @Test
+  public void testWithCompletionRunnableWhenIterableThrows() {
+    AtomicInteger completionCounter = new AtomicInteger(0);
+    List<Integer> items = Lists.newArrayList(1, 2, 3, 4, 5);
+
+    Assertions.assertThatThrownBy(
+            () -> {
+              try (CloseableIterable<Integer> iter =
+                  CloseableIterable.whenComplete(
+                      CloseableIterable.combine(
+                          items,
+                          () -> {
+                            throw new RuntimeException("expected");
+                          }),
+                      completionCounter::incrementAndGet)) {
+                iter.forEach(val -> Assertions.assertThat(completionCounter.get()).isEqualTo(0));
+              }
+            })
+        .isInstanceOf(RuntimeException.class)
+        .hasMessage("expected");
+
+    Assertions.assertThat(completionCounter.get()).isEqualTo(1);
+  }
+
+  @Test
   public void testConcatWithEmpty() {
     AtomicInteger counter = new AtomicInteger(0);
     CloseableIterable.concat(Collections.emptyList()).forEach(c -> counter.incrementAndGet());
@@ -130,5 +199,65 @@ public class TestCloseableIterable {
       concat.forEach(c -> c++);
     }
     Assertions.assertThat(counter.get()).isEqualTo(items.size());
+  }
+
+  @Test
+  public void count() {
+    MetricsContext.Counter<Integer> counter =
+        new DefaultMetricsContext().counter("x", Integer.class, MetricsContext.Unit.COUNT);
+    CloseableIterable<Integer> items =
+        CloseableIterable.count(
+            counter, CloseableIterable.withNoopClose(Arrays.asList(1, 2, 3, 4, 5)));
+    Assertions.assertThat(counter.value()).isEqualTo(0);
+    items.forEach(item -> {});
+    Assertions.assertThat(counter.value()).isEqualTo(5);
+  }
+
+  @Test
+  public void countSkipped() {
+    MetricsContext.Counter<Integer> counter =
+        new DefaultMetricsContext().counter("x", Integer.class, MetricsContext.Unit.COUNT);
+    CloseableIterable<Integer> items =
+        CloseableIterable.filter(
+            counter,
+            CloseableIterable.withNoopClose(Arrays.asList(1, 2, 3, 4, 5)),
+            x -> x % 2 == 0);
+    Assertions.assertThat(counter.value()).isEqualTo(0);
+    items.forEach(item -> {});
+    Assertions.assertThat(counter.value()).isEqualTo(3);
+  }
+
+  @Test
+  public void countNullCheck() {
+    Assertions.assertThatThrownBy(() -> CloseableIterable.count(null, CloseableIterable.empty()))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Invalid counter: null");
+
+    MetricsContext.Counter<Integer> counter =
+        new DefaultMetricsContext().counter("x", Integer.class, MetricsContext.Unit.COUNT);
+    Assertions.assertThatThrownBy(() -> CloseableIterable.count(counter, null))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Invalid iterable: null");
+  }
+
+  @Test
+  public void countSkippedNullCheck() {
+    Assertions.assertThatThrownBy(
+            () ->
+                CloseableIterable.filter(null, CloseableIterable.empty(), Predicate.isEqual(true)))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Invalid counter: null");
+
+    MetricsContext.Counter<Integer> counter =
+        new DefaultMetricsContext().counter("x", Integer.class, MetricsContext.Unit.COUNT);
+    Assertions.assertThatThrownBy(
+            () -> CloseableIterable.filter(counter, null, Predicate.isEqual(true)))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Invalid iterable: null");
+
+    Assertions.assertThatThrownBy(
+            () -> CloseableIterable.filter(counter, CloseableIterable.empty(), null))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Invalid predicate: null");
   }
 }
--- a/core/src/main/java/org/apache/iceberg/BaseTableScan.java
+++ b/core/src/main/java/org/apache/iceberg/BaseTableScan.java
@@ -25,6 +25,9 @@ import org.apache.iceberg.events.ScanEvent;
 import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.expressions.ExpressionUtil;
 import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.metrics.DefaultMetricsContext;
+import org.apache.iceberg.metrics.ScanReport;
+import org.apache.iceberg.metrics.Timer;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.util.DateTimeUtil;
@@ -37,6 +40,7 @@ import org.slf4j.LoggerFactory;
 abstract class BaseTableScan extends BaseScan<TableScan, FileScanTask, CombinedScanTask>
     implements TableScan {
   private static final Logger LOG = LoggerFactory.getLogger(BaseTableScan.class);
+  private ScanReport.ScanMetrics scanMetrics;
 
   protected BaseTableScan(TableOperations ops, Table table, Schema schema) {
     this(ops, table, schema, new TableScanContext());
@@ -68,6 +72,14 @@ abstract class BaseTableScan extends BaseScan<TableScan, FileScanTask, CombinedS
   }
 
   protected abstract CloseableIterable<FileScanTask> doPlanFiles();
+
+  protected ScanReport.ScanMetrics scanMetrics() {
+    if (scanMetrics == null) {
+      this.scanMetrics = new ScanReport.ScanMetrics(new DefaultMetricsContext());
+    }
+
+    return scanMetrics;
+  }
 
   @Override
   public Table table() {
@@ -121,9 +133,22 @@ abstract class BaseTableScan extends BaseScan<TableScan, FileScanTask, CombinedS
           ExpressionUtil.toSanitizedString(filter()));
 
       Listeners.notifyAll(new ScanEvent(table().name(), snapshot.snapshotId(), filter(), schema()));
+      Timer.Timed scanDuration = scanMetrics().totalPlanningDuration().start();
 
-      return doPlanFiles();
-
+      return CloseableIterable.whenComplete(
+          doPlanFiles(),
+          () -> {
+            scanDuration.stop();
+            ScanReport scanReport =
+                ScanReport.builder()
+                    .withFilter(ExpressionUtil.sanitize(filter()))
+                    .withProjection(schema())
+                    .withTableName(table().name())
+                    .withSnapshotId(snapshot.snapshotId())
+                    .fromScanMetrics(scanMetrics())
+                    .build();
+            context().scanReporter().reportScan(scanReport);
+          });
     } else {
       LOG.info("Scanning empty table {}", table());
       return CloseableIterable.empty();
--- a/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java
+++ b/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java
@@ -40,6 +40,7 @@ import org.apache.iceberg.expressions.ManifestEvaluator;
 import org.apache.iceberg.expressions.Projections;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.metrics.ScanReport;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
@@ -363,6 +364,7 @@ class DeleteFileIndex {
     private PartitionSet partitionSet = null;
     private boolean caseSensitive = true;
     private ExecutorService executorService = null;
+    private ScanReport.ScanMetrics scanMetrics = ScanReport.ScanMetrics.NOOP;
 
     Builder(FileIO io, Set<ManifestFile> deleteManifests) {
       this.io = io;
@@ -401,6 +403,11 @@ class DeleteFileIndex {
 
     Builder planWith(ExecutorService newExecutorService) {
       this.executorService = newExecutorService;
+      return this;
+    }
+
+    Builder scanMetrics(ScanReport.ScanMetrics newScanMetrics) {
+      this.scanMetrics = newScanMetrics;
       return this;
     }
 
--- a/core/src/main/java/org/apache/iceberg/ManifestGroup.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestGroup.java
@@ -36,6 +36,7 @@ import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.CloseableIterator;
 import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.metrics.ScanReport;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
@@ -60,6 +61,7 @@ class ManifestGroup {
   private List<String> columns;
   private boolean caseSensitive;
   private ExecutorService executorService;
+  private ScanReport.ScanMetrics scanMetrics;
 
   ManifestGroup(FileIO io, Iterable<ManifestFile> manifests) {
     this(
@@ -83,6 +85,7 @@ class ManifestGroup {
     this.caseSensitive = true;
     this.manifestPredicate = m -> true;
     this.manifestEntryPredicate = e -> true;
+    this.scanMetrics = ScanReport.ScanMetrics.NOOP;
   }
 
   ManifestGroup specsById(Map<Integer, PartitionSpec> newSpecsById) {
@@ -116,6 +119,11 @@ class ManifestGroup {
   ManifestGroup filterManifestEntries(
       Predicate<ManifestEntry<DataFile>> newManifestEntryPredicate) {
     this.manifestEntryPredicate = manifestEntryPredicate.and(newManifestEntryPredicate);
+    return this;
+  }
+
+  ManifestGroup scanMetrics(ScanReport.ScanMetrics metrics) {
+    this.scanMetrics = metrics;
     return this;
   }
 
@@ -171,7 +179,7 @@ class ManifestGroup {
                   return ResidualEvaluator.of(spec, filter, caseSensitive);
                 });
 
-    DeleteFileIndex deleteFiles = deleteIndexBuilder.build();
+    DeleteFileIndex deleteFiles = deleteIndexBuilder.scanMetrics(scanMetrics).build();
 
     boolean dropStats = ManifestReader.dropStats(dataFilter, columns);
     if (!deleteFiles.isEmpty()) {
@@ -184,7 +192,7 @@ class ManifestGroup {
                 specId -> {
                   PartitionSpec spec = specsById.get(specId);
                   ResidualEvaluator residuals = residualCache.get(specId);
-                  return new TaskContext(spec, deleteFiles, residuals, dropStats);
+                  return new TaskContext(spec, deleteFiles, residuals, dropStats, scanMetrics);
                 });
 
     Iterable<CloseableIterable<T>> tasks =
@@ -239,11 +247,14 @@ class ManifestGroup {
       evaluator = null;
     }
 
-    Iterable<ManifestFile> matchingManifests =
+    CloseableIterable<ManifestFile> closeableDataManifests =
+        CloseableIterable.withNoopClose(dataManifests);
+    CloseableIterable<ManifestFile> matchingManifests =
         evalCache == null
-            ? dataManifests
-            : Iterables.filter(
-                dataManifests,
+            ? closeableDataManifests
+            : CloseableIterable.filter(
+                scanMetrics.skippedDataManifests(),
+                closeableDataManifests,
                 manifest -> evalCache.get(manifest.partitionSpecId()).eval(manifest));
 
     if (ignoreDeleted) {
@@ -251,7 +262,8 @@ class ManifestGroup {
       // remove any manifests that don't have any existing or added files. if either the added or
       // existing files count is missing, the manifest must be scanned.
       matchingManifests =
-          Iterables.filter(
+          CloseableIterable.filter(
+              scanMetrics.skippedDataManifests(),
               matchingManifests,
               manifest -> manifest.hasAddedFiles() || manifest.hasExistingFiles());
     }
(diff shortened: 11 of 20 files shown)


api/src/main/java/org/apache/iceberg/metrics/ScanMetricsContext.java
```diff
@@ -0,0 +1,128 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.metrics;
+
+import java.util.StringJoiner;
+
+public class ScanMetricsContext implements MetricsContext {
```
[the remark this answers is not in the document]
nastra: that makes perfect sense @danielcweeks, thanks. I'll adjust the implementation accordingly

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,136 @@
...
+   * The total duration that was recorded.
+   *
+   * @return The total duration that was recorded.
+   */
+  Duration totalDuration();
+
+  /**
+   * Starts the timer
+   */
+  void startTimer();
```
rdblue, line 49: What about using just `start` and `stop`? Isn't it clear that these are timer methods?

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,136 @@
...
+
+import java.time.Duration;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * Generalized Timer interface for creating telemetry related instances for measuring duration of operations.
+ */
+public interface Timer {
```
rdblue, line 30: A lot of other frameworks implement `AutoCloseable` so that you can use try-with-resources to stop the timer:

```java
  List<DataFile> files;
  try (Timer t = context.timer("manfiest-read", TimeUnit.MICROSECONDS).start()) {
    this.files = ManifestFiles.read(...);
  }
```

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,136 @@
...
+  default void record(Duration duration) {
+    record(duration.toNanos(), TimeUnit.NANOSECONDS);
+  }
+
+  /**
+   * Executes and measures the given {@link Runnable} instance.
+   *
+   * @param runnable The {@link Runnable} to execute and measure.
+   */
+  void record(Runnable runnable);
```
rdblue, line 78: What about using `time` as the verb? I think `record` makes sense for adding intervals timed in some other way, but I would expect `time` to be used when passing a `Runnable` or `Callable`.

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Executes and measures the given {@link Callable} and returns its result.
+   *
+   * @param callable The {@link Callable} to execute and measure.
+   * @param <T>      The type of the {@link Callable}
+   * @return The result of the underlying {@link Callable}.
+   * @throws Exception In case the {@link Callable} fails.
+   */
+  <T> T recordCallable(Callable<T> callable) throws Exception;
```
rdblue, line 88: An alternative that I tend to like is to have a custom `Task` interface rather than using `Callable` that allows you to specify the exception type:

```java
  public interface TimedTask<R, E extends Exception> {
    R call() throws E;
  }
```

That way, you can customize the exception type and don't need an outer `try` block if it only throws `RuntimeException`.

api/src/main/java/org/apache/iceberg/metrics/LongCounter.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public Optional<Long> count() {
+    return Optional.of(counter.get());
+  }
+
+  @Override
+  public String toString() {
+    return count().orElse(-1L).toString();
```
rdblue, line 53: This will never be empty, right? Why not just print the value of counter directly?

api/src/main/java/org/apache/iceberg/metrics/LongCounter.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A default {@link org.apache.iceberg.metrics.MetricsContext.Counter} implementation that uses a {@link Long} to count
+ * events.
+ */
+class LongCounter implements MetricsContext.Counter<Long> {
+  private final AtomicLong counter;
+
+  LongCounter() {
+    counter = new AtomicLong(0L);
```
rdblue, line 33: When setting instance fields, we use `this.` to show that it is modifying state outside of the local context.

api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
```diff
@@ -75,6 +75,31 @@ public CloseableIterator<E> iterator() {
...
+   * @return A new {@link CloseableIterable} where the runnable will be executed
+   * as the final step after {@link CloseableIterable#close()} has been called
+   */
+  static <E> CloseableIterable<E> whenComplete(CloseableIterable<E> iterable, Runnable onCompletionRunnable) {
+    Preconditions.checkNotNull(onCompletionRunnable, "Cannot execute a null Runnable after completion");
+    return new CloseableIterable<E>() {
+      @Override
+      public void close() throws IOException {
+        iterable.close();
+        onCompletionRunnable.run();
```
kbendick, line 93: Does the `onCompletionRunnable` need to happen in a `try-finally` block in case `iterable.close()` throws?

Might be something you want to add a test for.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  @Override
+  public String toString() {
+    return new StringJoiner(", ", DefaultTimer.class.getSimpleName() + "[", "]")
+        .add("count=" + count)
+        .add("duration=" + duration)
+        .toString();
+  }
```
kbendick, line 114: Style / Non-blocking: I'm a big fan of using Guava's `MoreObjects.toStringHelper` like seen here:

https://github.com/apache/iceberg/blob/90fe0edf1a671095e587a53adeb31cc84d01fb89/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java#L72-L78

It more or less gives you the same result but is a lot less mental overhead in my opinion. Up to you if you use it or not though.

nastra: agreed, that looks much nicer

api/src/main/java/org/apache/iceberg/metrics/LongCounter.java
nastra: makes sense, updated

api/src/main/java/org/apache/iceberg/metrics/LongCounter.java
nastra: makes sense, updated

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: good point, I updated this to just start() / stop()

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: I was broadly following how Micrometer defines their [Timer](https://www.javadoc.io/doc/io.micrometer/micrometer-core/1.0.3/io/micrometer/core/instrument/Timer.html) interface and I had the exact same thought as you mentioned here. 
That being said, `time()` makes also sense to me as a verb for measuring a Runnable/Callable/Supplier, but I think `record(long amount, TimeUnit unit)` / `record(Duration duration)` reads nicer than `time(long amount, TimeUnit unit)` / `time(Duration duration)` and in order to keep method names aligned I decided to go with `record()` across the board.

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: I like that idea. Would you like me to add this as part of this PR or rather later?

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,141 @@
...
+
+  @Override
+  public Duration totalDuration() {
+    return duration;
+  }
+
+  @Override
+  public void startTimer() {
+    boolean updated = current.compareAndSet(null, Stopwatch.createStarted());
+    Preconditions.checkState(updated, "startTimer() called multiple times");
```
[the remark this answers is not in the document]
nastra: you're absolutely right about this. That's also how Micrometer handles timing state. I'll update accordingly

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: I think it would also make sense to have something like a `TimedRunnable` / `TimedSupplier`

api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
nastra: good catch, that makes definitely sense. Updated and added a test

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: Given that we're tracking state, using `AutoCloseable` (bf2178c818508152dce7a3b910a998a5fba9d2b5) makes the API less readable imo. I've done it in a separate commit, so we can still decide whether we want to have it or not

nastra:
I applied all the review feedback and splitted out the introduction of a `DefaultMetricsContext` and the `Timer` API into https://github.com/apache/iceberg/pull/5286. Once https://github.com/apache/iceberg/pull/5286 is merged, I'll rebase this PR so that it only contains Table Scan related changes

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: I left the `AutoCloseable` changes in the commit mentioned above but left it out from https://github.com/apache/iceberg/pull/5286 for now. Let me know if you'd like to include it in https://github.com/apache/iceberg/pull/5286

api/src/main/java/org/apache/iceberg/metrics/Timer.java
rdblue: I agree with using `record` for direct additions, but I don't think I would make the names the same across something that is timed by the timer and a time that is provided by the caller.

api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
```diff
@@ -0,0 +1,44 @@
...
+ * A default {@link MetricsContext} implementation that uses native Java counters/timers.
+ */
+public class DefaultMetricsContext implements MetricsContext {
+
+  @Override
+  public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit unit) {
+    if (Integer.class.equals(type)) {
+      return (Counter<T>) new IntCounter();
+    }
+    if (Long.class.equals(type)) {
```
rdblue, line 34: Style: we add an empty newline between control flow blocks and the following statement.

api/src/main/java/org/apache/iceberg/metrics/Timer.java
rdblue: I'd probably replace what's here with two variations, one that returns a value and one that has no return. Then you've covered all of the cases and can use method references to call with a Runnable, Supplier, or Callable.

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,156 @@
...
+   * @param <T>      The type of the {@link Supplier}.
+   * @return The result of the underlying {@link Supplier}.
+   */
+  <T> T record(Supplier<T> supplier);
+
+  /**
+   * A timing sample that carries internal state about the Timer's start position. The timing can be completed by
+   * calling {@link Sample#stop(Timer)}.
+   */
+  interface Sample {
```
rdblue, line 102: When I think of sampling in metrics, I think of randomly deciding whether to time or use a no-op, to avoid overhead in tight loops. It looks like the intent here is to make this something that reports when stopped, so a different name is probably more clear (I had no idea why this had sampling).

How about calling this `Timed` or something?

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  Timer NOOP = new Timer() {
+    @Override
+    public Sample start() {
+      return Sample.NOOP;
+    }
+
+    @Override
+    public CloseableSample startCloseable() {
```
rdblue, line 124: I think this is probably unnecessary given that there's already a `start` method.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,181 @@
...
+  public Sample start() {
+    boolean updated = stopwatchRef.compareAndSet(null, Stopwatch.createStarted());
+    Preconditions.checkState(updated, "startTimer() called multiple times");
+    return new DefaultSample(stopwatchRef, defaultTimeUnit);
+  }
+
+  @Override
+  public CloseableSample startCloseable() {
+    boolean updated = stopwatchRef.compareAndSet(null, Stopwatch.createStarted());
+    Preconditions.checkState(updated, "startTimer() called multiple times");
```
rdblue, line 67: I think this needs to be updated. Now that it is using a `Sample` to hold the stopwatch, there's no need to fail if it is called concurrently.

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+
+  public int totalDataManifestsRead() {
+    return totalDataManifestsRead;
+  }
+
+  public int addedDataFiles() {
+    return addedDataFiles;
+  }
+
+  public int deletedDataFiles() {
```
rdblue, line 98: What is `deletedDataFiles` reporting?

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+    return deletedDataFiles;
+  }
+
+  public Duration totalScanDuration() {
```
rdblue, line 102: This is tracking scan planning, not a scan. Values like this should be `totalPlanningDuration`.

We also typically use `long` to return durations in milliseconds. Is there a reason to use `Duration` instead? Are millis not granular enough?

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+
+  public int matchingDataManifests() {
+    return matchingDataManifests;
+  }
+
+  public int totalDataManifestsRead() {
+    return totalDataManifestsRead;
+  }
+
+  public int addedDataFiles() {
```
rdblue, line 94: When would a scan add data files?

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+
+  public Schema projection() {
+    return projection;
+  }
+
+  public int matchingDataFiles() {
+    return matchingDataFiles;
+  }
+
+  public int matchingDataManifests() {
```
rdblue, line 86: I think we want a metric for total data manifests as well so that we can track the % of data manifests that were scanned.

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+      return this;
+    }
+
+    public ScanReport build() {
+      Preconditions.checkArgument(null != tableName, "TableName must be non-null");
+      Preconditions.checkArgument(null != filter, "Expression filter must be non-null");
+      Preconditions.checkArgument(null != projection, "Schema projection must be non-null");
+      Preconditions.checkArgument(null != scanMetrics, "ScanMetrics must be non-null");
+      return new ScanReport(tableName, snapshotId, filter, projection,
+          scanMetrics.matchingDataFilesCount().count().orElse(-1),
```
rdblue, line 173: Why would a count not default to 0?

api/src/main/java/org/apache/iceberg/metrics/ScanReporter.java
```diff
@@ -0,0 +1,84 @@
...
+   * @return The current {@link ScanMetrics} that can be used during a table scan to add metrics to.
+   */
+  ScanMetrics scanMetrics();
+
+  /**
+   * Can be used to provide additional properties to the underlying {@link ScanReporter} implementation.
+   *
+   * @param properties The properties to pass to the underlying {@link ScanReporter} implementation.
+   */
+  void initialize(Map<String, String> properties);
```
rdblue, line 62: I would probably remove customization for now. We can add this later.

core/src/main/java/org/apache/iceberg/BaseScan.java
```diff
@@ -59,6 +76,10 @@ protected TableScanContext context() {
     return context;
   }
 
+  protected ScanReporter scanReporter() {
```
rdblue, line 79: Does this need to be exposed publicly?

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -116,9 +133,20 @@ public CloseableIterable<FileScanTask> planFiles() {
           ExpressionUtil.toSanitizedString(filter()));
 
       Listeners.notifyAll(new ScanEvent(table().name(), snapshot.snapshotId(), filter(), schema()));
-
-      return doPlanFiles();
-
+      scanReporter().startScan();
+      Timer.Sample sample = scanReporter().scanMetrics().totalScanDuration().start();
```
rdblue, line 137: Tying the lifecycle of scan metrics to the scan reporter means that all of the scans created through this chain will have the same metrics collector, which is a correctness problem.

In the Java library, the `TableScan` interface uses a refinement pattern. Each `TableScan` is independent and you get a new table scan that has additional options when you call the refinement methods, like `select` or `filter`. For example,

```java
TableScan fullTableScan = table.newScan();
TableScan yesterday = fullTableScan.filter(Expressions.equal("date_col", yesterday()));
TableScan loginsYesterday = yesterday.filter(Expressions.equal("event_type", "login"));

CloseableIterable<FileScanTask> yesterdayTasks = yesterday.planFiles();
CloseableIterable<FileScanTask> loginTasks = loginsYesterday.planFiles();
```

Two scans are planned, but the metrics for both would use the same `ScanMetrics`.

Above, I suggested separating `ScanMetrics` from the reporter, which would solve this problem. Not all scans are planned, so you wouldn't want to eagerly create metrics, but the ones that are used should have a new one:

```java
  protected ScanMetrics metrics() {
    if (scanMetrics == null) {
      this.scanMetrics = scanReporter.newScanMetrics();
    }

    return scanMetrics;
  }

  public CloseableIterable<FileScanTask> planFiles(...) {
    ScanMetrics metrics = metrics();
    ...
  }
```

core/src/main/java/org/apache/iceberg/IncrementalDataTableScan.java
```diff
@@ -34,7 +35,7 @@
 class IncrementalDataTableScan extends DataTableScan {
 
   IncrementalDataTableScan(TableOperations ops, Table table, Schema schema, TableScanContext context) {
-    super(ops, table, schema, context.useSnapshotId(null));
+    super(ops, table, schema, context.useSnapshotId(null), new LoggingScanReporter());
```
rdblue, line 38: Rather than changing so many method signatures, why not add `ScanReporter` to the `TableScanContext`?

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -215,8 +229,15 @@ private <T> Iterable<CloseableIterable<T>> entries(
       evaluator = null;
     }
 
+    if (evalCache == null) {
+      scanMetrics.totalDataManifestsReadCount().increment(dataManifests.size());
+    }
```
rdblue, line 234: Style: missing newline.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -163,20 +171,26 @@ public CloseableIterable<FileScanTask> planFiles() {
...
     DeleteFileIndex deleteFiles = deleteIndexBuilder.build();
-
     boolean dropStats = ManifestReader.dropStats(dataFilter, columns);
     if (!deleteFiles.isEmpty()) {
       select(ManifestReader.withStatsColumns(columns));
     }
 
     Iterable<CloseableIterable<FileScanTask>> tasks = entries((manifest, entries) -> {
+      scanMetrics.addedDataFiles().increment(manifest.addedFilesCount());
+      scanMetrics.deletedDataFiles().increment(manifest.deletedFilesCount());
```
rdblue, line 181: I think I see what you're doing with these counters now. These end up tracking the total number of data or delete files in manifests that are scanned, right?

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -163,20 +171,26 @@ public CloseableIterable<FileScanTask> planFiles() {
...
       int specId = manifest.partitionSpecId();
       PartitionSpec spec = specsById.get(specId);
       String schemaString = SchemaParser.toJson(spec.schema());
       String specString = PartitionSpecParser.toJson(spec);
       ResidualEvaluator residuals = residualCache.get(specId);
-      return CloseableIterable.transform(entries, e -> new BaseFileScanTask(
-          e.file().copy(!dropStats), deleteFiles.forEntry(e), schemaString, specString, residuals));
+      return CloseableIterable.transform(entries, e -> {
+        scanMetrics.matchingDataFilesCount().increment();
+        scanMetrics.totalFileSizeInBytes().increment(e.file().fileSizeInBytes());
```
rdblue, line 190: I think a more reliable place to gather these by wrapping the iterator returned by `planFiles`. That way you don't have to worry about future changes to code futher filtering the iterator that you're wrapping here.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -163,20 +171,26 @@ public CloseableIterable<FileScanTask> planFiles() {
     });
 
     DeleteFileIndex deleteFiles = deleteIndexBuilder.build();
-
```
rdblue, line 166: Nit: unnecessary whitespace change.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -234,7 +255,13 @@ private <T> Iterable<CloseableIterable<T>> entries(
           manifest -> manifest.hasAddedFiles() || manifest.hasDeletedFiles());
     }
 
-    matchingManifests = Iterables.filter(matchingManifests, manifestPredicate::test);
+    matchingManifests = Iterables.filter(matchingManifests, manifest -> {
```
rdblue, line 258: Rather than changing the filtering, why not make a new Iterable that updates a counter? Then you'd have less custom code:

```java
  class <T> Iterable<T> count(Counter counter, Iterable<T> iterable) {
    return () -> new Iterator<T> {
      Iterator<T> iter = iterable.iterator();

      public boolean hasNext() {
        return iter.hasNext();
      }

      public T next() {
        T next = iter.next();
        counter.increment();
        return next;
      }
    }
  }

  ...

  matchingManifests = count(scanMetrics.matchingDataManifests(), Iterables.filter(...));
```

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: `Timed` sounds good to me. Updated

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,156 @@
...
+  interface Sample {
+    /**
+     * Stops the timer and records the total duration up until {@link Timer#start()} was called.
+     */
+    void stop(Timer timer);
+
+    Sample NOOP = timer -> { };
+  }
+
+  interface CloseableSample extends AutoCloseable {
```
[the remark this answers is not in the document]
nastra: the only reason I figured this might be a good idea is the IDE complaining about `Timer#start()` being used without a `try-with-resources` statement everywhere, but I guess the API is less readable so it's probably better to make `Timed` extend `AutoCloseable`

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * A timing sample that carries internal state about the Timer's start position. The timing can be completed by
+   * calling {@link Sample#stop(Timer)}.
+   */
+  interface Sample {
+    /**
+     * Stops the timer and records the total duration up until {@link Timer#start()} was called.
+     */
+    void stop(Timer timer);
```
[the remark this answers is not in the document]
nastra: agreed, fixed

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: ok renamed those 3 methods to `time(...)`

api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
nastra: fixed

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,156 @@
...
+    public Duration totalDuration() {
+      return Duration.ZERO;
+    }
+
+    @Override
+    public void record(long amount, TimeUnit unit) {
+    }
+
+    @Override
+    public void record(Runnable runnable) {
```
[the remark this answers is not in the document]
nastra: fixed

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
nastra: updated

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,181 @@
...
+  @Override
+  public Sample start() {
+    boolean updated = stopwatchRef.compareAndSet(null, Stopwatch.createStarted());
+    Preconditions.checkState(updated, "startTimer() called multiple times");
+    return new DefaultSample(stopwatchRef, defaultTimeUnit);
+  }
+
+  @Override
+  public CloseableSample startCloseable() {
+    boolean updated = stopwatchRef.compareAndSet(null, Stopwatch.createStarted());
```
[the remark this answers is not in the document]
nastra: updated

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private DefaultSample(AtomicReference<Stopwatch> stopwatchRef, TimeUnit defaultTimeUnit) {
+      this.stopwatchRef = stopwatchRef;
+      this.defaultTimeUnit = defaultTimeUnit;
+    }
+
+    public void stop(Timer timer) {
+      Preconditions.checkArgument(null != timer, "Timer must be non-null");
+      Stopwatch stopwatch = stopwatchRef.getAndSet(null);
+      Preconditions.checkState(null != stopwatch, "startTimer() was not called");
```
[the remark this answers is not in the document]
nastra: fixed

api/src/main/java/org/apache/iceberg/metrics/Timer.java
nastra: fixed now

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: -1 means here that this wasn't tracked/available, but 0 means that this was tracked and was actually 0

api/src/test/java/org/apache/iceberg/io/TestCloseableIterable.java
```diff
@@ -82,4 +85,61 @@ public void testConcateWithEmptyIterables() {
...
   }
+
+  @Test
+  public void testWithCompletionRunnable() throws IOException {
+    AtomicInteger completionCounter = new AtomicInteger(0);
+    List<Integer> items = Lists.newArrayList(1, 2, 3, 4, 5);
+    Assertions.assertThatThrownBy(() ->
+                    CloseableIterable.whenComplete(CloseableIterable.combine(items, () -> { }), null))
+            .isInstanceOf(NullPointerException.class)
+            .hasMessage("Cannot execute a null Runnable after completion");
```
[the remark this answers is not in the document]
nastra: fixed

api/src/main/java/org/apache/iceberg/metrics/ScanReporter.java
nastra: ok removed it

core/src/main/java/org/apache/iceberg/BaseScan.java
nastra: subclasses need access to this and checkstyle is complaining that this needs to be exposed via a method for subclasses

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -346,4 +347,42 @@ public static void configureHadoopConf(Object maybeConfigurable, Object conf) {
...
+   * The implementation must have a no-arg constructor. {@link ScanReporter#initialize(Map properties)} is called to
+   * complete the initialization.
+   *
+   * @param impl       full class name of a custom {@link ScanReporter} implementation
+   * @param properties used to initialize the {@link ScanReporter} implementation
+   * @return An initialized {@link ScanReporter}.
+   * @throws IllegalArgumentException if class path not found or right constructor not found or the loaded class cannot
+   *                                  be cast to the given interface type
+   */
+  public static ScanReporter loadScanReporter(String impl, Map<String, String> properties) {
```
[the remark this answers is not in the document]
nastra: ok removed it

core/src/main/java/org/apache/iceberg/IncrementalDataTableScan.java
nastra: I haven't considered that, but that makes perfect sense. Updated the code

core/src/main/java/org/apache/iceberg/BaseScan.java
nastra: moved the `ScanReporter` into the `TableScanContext`, so this change isn't required anymore

core/src/main/java/org/apache/iceberg/ManifestGroup.java
nastra: yes correct. The naming is probably bad in `ScanMetrics`, so I'm completely open to suggestions on naming and also on what metrics to track and where to retrieve them from. The initial idea was that there are just **some metrics** gathered and reported

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+
+  public int matchingDataFiles() {
+    return matchingDataFiles;
+  }
+
+  public int matchingDataManifests() {
+    return matchingDataManifests;
+  }
+
+  public int totalDataManifestsRead() {
```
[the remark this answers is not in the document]
nastra: yes correct

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: this is tracked as part of `totalDataManifestsRead()`

core/src/main/java/org/apache/iceberg/BaseTableScan.java
nastra: thanks for bringing this up. I haven't thought about that. I'll rewrite the implementation and add some tests to make sure that each independent table scan produce their own metrics/report

nastra: I double-checked and the old implementation was behaving correctly, because a new `ScanMetrics` instance was created when `ScanReporter#startScan()` was called.
However, I updated the `ScanReporter` implementation to how you suggested, as that is cleaner

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,124 @@
...
+    private DefaultTimed(Timer timer, TimeUnit defaultTimeUnit) {
+      this.timer = timer;
+      this.defaultTimeUnit = defaultTimeUnit;
+      stopwatchRef.compareAndSet(null, Stopwatch.createStarted());
+    }
+
+    @Override
+    public void stop() {
+      Stopwatch stopwatch = stopwatchRef.getAndSet(null);
+      Preconditions.checkState(null != stopwatch, "startTimer() was not called");
```
rdblue, line 120: There is no `startTimer` method anymore. If `stopwatch` is null, that indicates that `stop` was called more than once.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public long count() {
+    return count.longValue();
+  }
+
+  @Override
+  public Duration totalDuration() {
+    return Duration.ofNanos(totalTime.longValue());
```
rdblue, line 52: Why is `totalTime` recorded in nanos instead of `defaultTimeUnit`?

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
nastra: ah yeah I need to adjust the error message to say `stop() called multiple times`

rdblue (changes requested):
I merged the CloseableIterable fix and the addition of Timer. Can you rebase? Thanks!

core/src/main/java/org/apache/iceberg/ManifestGroup.java
nastra: added that functionality

api/src/main/java/org/apache/iceberg/metrics/LoggingScanReporter.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class LoggingScanReporter implements ScanReporter {
+  private static final Logger LOG = LoggerFactory.getLogger(LoggingScanReporter.class);
+
+  @Override
+  @SuppressWarnings("Slf4jConstantLogMessage")
+  public void reportScan(ScanReport scanReport) {
+    Preconditions.checkArgument(null != scanReport, "ScanReport must be non-null");
+    LOG.info("Ending table scan");
+    LOG.info(scanReport.toString());
```
rdblue, line 37: Why use two info messages rather than one to report the result? I'd expect something like `"Completed scan planning: %s", report`

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+      return this;
+    }
+
+    public ScanReport build() {
+      Preconditions.checkArgument(null != tableName, "TableName must be non-null");
+      Preconditions.checkArgument(null != filter, "Expression filter must be non-null");
+      Preconditions.checkArgument(null != projection, "Schema projection must be non-null");
+      Preconditions.checkArgument(null != scanMetrics, "ScanMetrics must be non-null");
+      return new ScanReport(tableName, snapshotId, filter, projection,
+          scanMetrics.matchingDataFiles().count().orElse(-1),
```
rdblue, line 173: Following up on this comment, I

api/src/main/java/org/apache/iceberg/metrics/LoggingScanReporter.java
nastra: fixed

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+        .add("matchingDataManifests", matchingDataManifests)
+        .add("totalDataManifestsRead", totalDataManifestsRead)
+        .add("addedDataFiles", addedDataFiles)
+        .add("deletedDataFiles", deletedDataFiles)
+        .add("totalScanDuration", totalPlanningDuration)
+        .add("totalFileSizeBytes", totalFileSizeBytes)
+        .toString();
+  }
+
+  @SuppressWarnings("HiddenField")
```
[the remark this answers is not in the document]
nastra:
```
> Task :iceberg-api:checkstyleMain
[ant:checkstyle] [ERROR] /home/nastra/Development/workspace/iceberg/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java:141:41: 'tableName' hides a field. [HiddenField]
[ant:checkstyle] [ERROR] /home/nastra/Development/workspace/iceberg/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java:146:40: 'snapshotId' hides a field. [HiddenField]
[ant:checkstyle] [ERROR] /home/nastra/Development/workspace/iceberg/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java:151:42: 'filter' hides a field. [HiddenField]
[ant:checkstyle] [ERROR] /home/nastra/Development/workspace/iceberg/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java:156:42: 'projection' hides a field. [HiddenField]
[ant:checkstyle] [ERROR] /home/nastra/Development/workspace/iceberg/api/src/main/java/org/apache/iceberg/metrics/ScanReport.java:161:48: 'scanMetrics' hides a field. [HiddenField]
```

this is because the parameter names in the Builder methods are named like the field names in the Builder itself

api/src/main/java/org/apache/iceberg/metrics/LoggingScanReporter.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A default {@link ScanReporter} implementation that logs the {@link ScanReport} to the log file.
+ */
+public class LoggingScanReporter implements ScanReporter {
+  private static final Logger LOG = LoggerFactory.getLogger(LoggingScanReporter.class);
+
+  @Override
+  @SuppressWarnings("Slf4jConstantLogMessage")
+  public void reportScan(ScanReport scanReport) {
+    Preconditions.checkArgument(null != scanReport, "ScanReport must be non-null");
```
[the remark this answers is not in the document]
nastra: updated

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: one of the reasons I added this distinction between `ScanMetrics` and the final `ScanReport` is so that we wouldn't need to serialize Counters/Timers. Additionally, it seemed cleaner to me to have metrics that are being updated wrapped in one class, where we would derive a final `ScanReport` from that is immutable (well kind-of). 
The `ScanReport` is just a  Data holder that is serializable, but it doesn't carry the baggage of whatever a particular `Counter`/`Timer` implementation would have.

I understand the reasoning behind making `Counter#count()` return an `Optional` but I was actually wondering whether it wouldn't be better to make this return an actual value. It seems overly restrictive to have this be an `Optional` for a special case, such as Hadoop. Also you generally don't see such a restriction in the Counter API of Micrometer. Maybe in such a case the `Counters` that are being produced by `HadoopMetricsContext` would just return -1 for `count()` if they can't fetch whatever they are counting from `FileSystem.Statistics`.
@danielcweeks thoughts on this?

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,251 @@
...
+  private final long snapshotId;
+  private final Expression filter;
+  private final Schema projection;
+  private final int matchingDataFiles;
+  private final int matchingDataManifests;
+  private final int totalDataManifestsRead;
+  private final int addedDataFiles;
+  private final int deletedDataFiles;
+  private final Duration totalPlanningDuration;
+  private final long totalFileSizeBytes;
```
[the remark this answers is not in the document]
nastra: that makes perfect sense, thanks

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: regarding `Counter#count()` being `Optional` I talked to @danielcweeks and we eventually concluded that it would probably be best to actually report the count and throw an exception in the Hadoop case. I opened https://github.com/apache/iceberg/pull/5328 for that

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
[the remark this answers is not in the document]
nastra: I think instead of `filteredX` we might rather want to use `skippedX`. Depending on the context, `filtered` could mean something is included or excluded but `skipped` should be unambiguous.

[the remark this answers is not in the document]
rdblue: `skipped` sounds reasonable to me!

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
rdblue: I don't think that we need to serialize timers or counters if we don't have separate `ScanReport` and `ScanMetrics` classes. Scan planning should happen only on the coordinator, driver, or planner node. The resulting tasks are serialized to workers, but the scan itself should not be (and `TableScanContext` is not `Serializable`).

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,371 @@
...
+          scanMetrics.skippedDeleteManifests().value(),
+          scanMetrics.skippedDataFiles().value(),
+          scanMetrics.skippedDeleteFiles.value());
+    }
+  }
+
+  /**
+   * Carries all metrics for a particular scan
+   */
+  public static class ScanMetrics {
```
rdblue, line 244: I think this should be part of `ScanReport` rather than a separate class. I don't see much value in a separate class.

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
```diff
@@ -126,15 +126,14 @@ public Timer timer(String name, TimeUnit unit) {
...
-          }
-        };
+        if (Integer.class.equals(type)) {
+          return (Counter<T>) IntCounter.NOOP;
+        }
+
+        if (Long.class.equals(type)) {
+          return (Counter<T>) LongCounter.NOOP;
+        }
+        throw new IllegalArgumentException(String.format("Counter for type %s is not supported", type.getName()));
```
rdblue, line 136: Style: there should be a newline between a control flow block and the follow statement.

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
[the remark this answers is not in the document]
rdblue: With builders, we typically don't suppress the warning, since it is easy to omit `this.` and not catch that in a review. Instead, it is normal to use argument names like `newTableName` instead of `tableName`.

api/src/main/java/org/apache/iceberg/metrics/ScanReporter.java
```diff
@@ -0,0 +1,57 @@
...
+   * @param scanReport The {@link ScanReport} to report.
+   */
+  void reportScan(ScanReport scanReport);
+
+  /**
+   * Creates a new {@link ScanMetrics} instance that can be used during a table scan to add metrics to.
+   *
+   * @return A new {@link ScanMetrics} instance that can be used during a table scan to add metrics to.
+   */
+  ScanMetrics newScanMetrics();
```
rdblue, line 43: I think this should be `newReportBuilder` and the report should function as `ScanMetrics`.

api/src/main/java/org/apache/iceberg/io/IterableUtils.java
```diff
@@ -0,0 +1,85 @@
...
+   * Counts the number of elements in the given {@link Iterable} by incrementing the {@link Counter}
+   * instance for each {@link Iterator#next()} call.
+   *
+   * @param counter The {@link Counter} instance to increment on each {@link Iterator#next()} call.
+   * @param iterable The underlying {@link Iterator} to count
+   * @param <T> The underlying type to be iterated.
+   * @return An {@link Iterable} that increments the given counter on each {@link Iterator#next()}
+   *     call.
+   */
+  public static <T> Iterable<T> count(Counter<?> counter, Iterable<T> iterable) {
```
rdblue, line 43: It would be safer in general if these methods were moved to `CloseableIterable`. Then if the argument is `Iterable` we can wrap it with a noop close and if the argument is `CloseableIterable` we handle it correctly. Adding versions that dont' support close could easily cause trouble with later refactors.

api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
```diff
@@ -84,6 +113,20 @@ protected boolean shouldKeep(E item) {
     }, iterable);
   }
 
+  static <E> CloseableIterable<E> filter(MetricsContext.Counter<?> counter, CloseableIterable<E> iterable,
+      Predicate<E> pred) {
```
rdblue, line 117: Should there be a version that increments a `skip` counter?

I'd also prefer to add the counter support directly in `FilterIterator` rather than using a predicate. There is no guarantee that the predicate is called only once per value. Putting the increment logic inside `FilterIterator#advance` is better.

rdblue: If you want, we can also make a subclass that has callbacks for incrementing, so that we don't need to use NOOP counters when counters aren't supplied.

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
nastra: fixed

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: I'm rather thinking of the case where someone would want to send the `ScanReport` over the wire. For example, with the REST catalog we might want to send the `ScanReport` to an endpoint via a `POST`. Having actual `Counters` / `Timers` in the `ScanReport` makes things difficult. Does that make sense?

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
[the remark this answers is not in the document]
nastra: ok updated, but to be honest I don't feel like having the `HiddenField` checkstyle rule is actually that useful :)

api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
nastra: this version is actually incrementing a `skip` counter. I just forgot adding some javadocs to it.

Based on the current contract of [FilterIterator](https://github.com/apache/iceberg/blob/bbcc2fbf6db6f5386fa9e6d169a4cc19d7883ef2/api/src/main/java/org/apache/iceberg/io/FilterIterator.java#L67), it seems that `shouldKeep` is only called once. I tried to keep the changes for counting skipped elements to a minimum, so is this somethind we'd want to do in a follow-up PR?

api/src/main/java/org/apache/iceberg/io/IterableUtils.java
nastra: Is this somethind we'd want to do as a follow-up? As it would require to change a few places and I didn't want to pollute this PR with semi-related changes

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,370 @@
...
+      int resultDataFiles, int totalDataManifests, int totalDeleteManifests, int scannedDataManifests,
+      int scannedDeleteManifests, int totalDataFiles, int totalDeleteFiles, Duration totalPlanningDuration,
+      long totalFileSizeInBytes, int indexedDeleteFiles, int resultDeleteFiles,
+      int skippedDataManifests, int skippedDeleteManifests,
+      int skippedDataFiles, int skippedDeleteFiles) {
+    this.tableName = tableName;
+    this.snapshotId = snapshotId;
+    this.filter = filter;
+    this.projection = projection;
+    this.resultDataFiles = resultDataFiles;
```
danielcweeks, line 67: I had a conversation with Ryan about what to do with ScanReport and ScanMetrics.  I think we actually want to remove the final values from the scan report here.  I get that will make it easy to serialize a pojo result, but there's a separation of concerns that I think we can achieve rather easily.

Rather than collect all of the metric results in the ScanReport, just keep a reference to the ScanMetrics with the actual counters/timers.  This will allow the ScanReporter to manage the structure of the response.

For example:

```java
ScanReport:
  - table identifier
  - snapshot id
  - filter
  - projection
  - ScanMetrics:
      - <various metrics>
```

The scan reporter can then construct the payload dynamically based off of the Scan report (espcially if we make a way to iterate over the metrics):

```json
{
  "table-identifier": "a.b.c",
  "snapshot-id": 123,
  "filter": "foo=bar",
  "projection": "c1, c2",
  "metrics": {
    "<name>": {
      "type": "counter",   # we probably type metric type info and 
      "value": "val"
    },
    "<name>": {
      "type": "timer",
      "value": "val",
      "histogram": { "p50":  213, "p99":  18983 },
      "min": 23,
      "max": 20199
    }
  }
}
```

This allows us to separate the way we track metrics and the way we construct the payload so that if a future metric (say from micrometer) has a histogram we could then just update the reporter without have to create a new object structure for the values in ScanReport.  As you can see from what the json might look like, the metric info can get more complicated than we probably want to track here.

api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
```diff
@@ -75,6 +76,34 @@ public CloseableIterator<E> iterator() {
     };
   }
 
+  /**
+   * Will run the given runnable when {@link CloseableIterable#close()} has been called.
+   *
+   * @param iterable             The underlying {@link CloseableIterable} to iterate over
+   * @param onCompletionRunnable The runnable to run after the underlying iterable was closed
+   * @param <E>                  The type of der underlying iterable
```
danielcweeks, line 84: "der" -> "the"

api/src/main/java/org/apache/iceberg/metrics/LoggingScanReporter.java
```diff
@@ -0,0 +1,43 @@
...
+ * A default {@link ScanReporter} implementation that logs the {@link ScanReport} to the log file.
+ */
+public class LoggingScanReporter implements ScanReporter {
+  private static final Logger LOG = LoggerFactory.getLogger(LoggingScanReporter.class);
+
+  @Override
+  @SuppressWarnings("Slf4jConstantLogMessage")
+  public void reportScan(ScanReport scanReport) {
+    Preconditions.checkArgument(null != scanReport, "Invalid scan report: null");
+    LOG.info("Completed scan planning: {}", scanReport);
```
danielcweeks, line 36: If we update the ScanReport, we can still use this as long as the `ScanReport::toString` just iterates over the metrics and reports the value.

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
```diff
@@ -126,15 +126,15 @@ public Timer timer(String name, TimeUnit unit) {
...
-        return new Counter<T>() {
-          @Override
-          public void increment() {
-          }
-
-          @Override
-          public void increment(T amount) {
-          }
-        };
+        if (Integer.class.equals(type)) {
```
danielcweeks, line 129: Not sure I follow why we need these NOOP implementations.  Seems like once you create a counter, you don't really need to know if it's a long/int/etc when using it.

api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
nastra: I accidentally leaked german into the code, good catch :)

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: I like that idea, updated the code to reflect that

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
nastra: this is particuarly for the case where one would use `ScanMetrics NOOP = new ScanMetrics(MetricsContext.nullMetrics())`, which would then use `MetricsContext.nullMetrics()`. I don't think we'd want any access of `.value()` to throw an exception when `MetricsContext.nullMetrics()` is used, so we need NOOP counters for those (we can't make a generic NOOP counter, as `.value()` needs to return `0` for `Integer.class` and `0L` for `Long.class`). The below test verifies that exact behavior

https://github.com/apache/iceberg/blob/3ed173b87909c221d66a6b09d70c8f8d4b562e51/api/src/test/java/org/apache/iceberg/metrics/TestScanReport.java#L60-L81

danielcweeks (approved):
+1 This looks good to me.  There are a number of follow on PRs (add logging metrics to other catalogs, add REST spec and reporter, other metrics), but this is really great.

api/src/main/java/org/apache/iceberg/io/IterableUtils.java
rdblue: I think as long as we're introducing a `count` method, having it work with `CloseableIterable` makes the most sense, rather than doing it in a follow-up.

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,260 @@
...
+      this.filter = newFilter;
+      return this;
+    }
+
+    public Builder withProjection(Schema newProjection) {
+      this.projection = newProjection;
+      return this;
+    }
+
+    public Builder fromScanMetrics(ScanMetrics newScanMetrics) {
```
rdblue, line 116: Shouldn't this be created by the builder, rather than passed in?

api/src/main/java/org/apache/iceberg/io/IterableUtils.java
nastra: ok I removed `IterableUtils` and moved relevant things to `CloseableIterable`

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
nastra: It feels a bit odd to me to use the `ScanReport.Builder` to retrieve `ScanMetrics`. To me the `ScanReport` feels like something that is final and you'd only want to create at the very end. 

However, I think we can still simplify `ScanReporter` and make it a `FunctionalInterface` that only has `void reportScan(ScanReport scanReport)` and nothing else. 
`BaseTableScan` doesn't need the `ScanReporter` to init the metrics. It could just do
```
protected ScanReport.ScanMetrics scanMetrics() {
    if (scanMetrics == null) {
      this.scanMetrics = new ScanReport.ScanMetrics(new DefaultMetricsContext());
    }

    return scanMetrics;
  }
```
where it can specify whatever metrics context should be used.

Then we just build the `ScanReport` from `ScanMetrics` (and other info) and send it off via `context().scanReporter().reportScan(scanReport)`.

api/src/main/java/org/apache/iceberg/metrics/Timer.java
```diff
@@ -0,0 +1,141 @@
...
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+
+/**
+ * Generalized Timer interface for creating telemetry related instances for measuring duration of operations.
+ */
+public interface Timer {
```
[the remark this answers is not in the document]
nastra: @danielcweeks do you mean https://github.com/apache/iceberg/blob/master/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java#L105-L107? I think we should be good here as the timer creation goes through the `MetricsContext`

api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
```diff
@@ -55,6 +73,6 @@ public Integer value() {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this).add("counter", counter).toString();
+    return value().toString();
```
rdblue, line 76: Minor: I think it would be strange to print this and see just the value, but it's probably okay.

core/src/main/java/org/apache/iceberg/DeleteFileIndex.java
```diff
@@ -502,6 +509,9 @@ DeleteFileIndex build() {
         }
       }
 
+      if (null != globalDeletes) {
+        scanMetrics.indexedDeleteFiles().increment(globalDeletes.length);
+      }
```
rdblue, line 514: I don't think this is correct. It should be the number of delete files that were indexed for all partitions, not just the root partition. I think you should count the entries that are processed in the first loop, or add up the number of existing and added entries in the matching delete files as they are read (near `readDeleteManifest`).

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -184,12 +192,14 @@ public <T extends ScanTask> CloseableIterable<T> plan(CreateTasksFunction<T> cre
...
                   ResidualEvaluator residuals = residualCache.get(specId);
-                  return new TaskContext(spec, deleteFiles, residuals, dropStats);
+                  return new TaskContext(spec, deleteFiles, residuals, dropStats, scanMetrics);
                 });
 
     Iterable<CloseableIterable<T>> tasks =
         entries(
             (manifest, entries) -> {
+              scanMetrics.totalDataFiles().increment(manifest.addedFilesCount());
+              scanMetrics.totalDeleteFiles().increment(manifest.deletedFilesCount());
```
rdblue, line 202: Delete files and deleted files are distinct. This is the number of deleted entries in the manifest, not the number of delete files.

In addition, existing data files will be used in the scan, not just added data files. Since this is counting inside of `ManifestGroup`, we need to make sure that all of the returned files are accounted for. There are two options:
1. Put these updates in `entries` where `ignoreDeleted` and `ignoreExisting` are checked
2. Add the total data files counter to the manifest reader, like you currently have for skipped files

I prefer option 2. I would opt to have simpler hooks instead of trying to count in bulk like this using metadata counts at the right place. That's probably going to be the most reliable way to get this data.

api/src/main/java/org/apache/iceberg/metrics/ScanReport.java
```diff
@@ -0,0 +1,265 @@
...
+          metricsContext.counter("resultDataFiles", Integer.class, MetricsContext.Unit.COUNT);
+      this.scannedDataManifests =
+          metricsContext.counter("scannedDataManifests", Integer.class, MetricsContext.Unit.COUNT);
+      this.scannedDeleteManifests =
+          metricsContext.counter(
+              "scannedDeleteManifests", Integer.class, MetricsContext.Unit.COUNT);
+      this.totalDataFiles =
+          metricsContext.counter("totalDataFiles", Integer.class, MetricsContext.Unit.COUNT);
+      this.totalDeleteFiles =
+          metricsContext.counter("totalDeleteFiles", Integer.class, MetricsContext.Unit.COUNT);
```
rdblue, line 164: I think this needs to be implemented. Currently, the code counts deleted files rather than live files in delete manifests. You should be able to update this in `DeleteFileIndex` by passing `scanMetrics` to the `ManifestReader`. I think that will also fix `skippedDeleteFiles`.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -325,6 +351,9 @@ private static CloseableIterable<FileScanTask> createFileScanTasks(
         entry -> {
           DataFile dataFile = entry.file().copy(ctx.shouldKeepStats());
           DeleteFile[] deleteFiles = ctx.deletes().forEntry(entry);
+          ctx.scanMetrics().totalFileSizeInBytes().increment(dataFile.fileSizeInBytes());
```
rdblue, line 354: Should we have a similar metrics for total delete file size?

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,210 @@
...
+    record4.setField("x", "51");
+    DataFile dataFile = writeParquetFile(table, Arrays.asList(record, record3));
+    DataFile dataFile2 = writeParquetFile(table, Arrays.asList(record2));
+    DataFile dataFile3 = writeParquetFile(table, Arrays.asList(record4));
+    table.newFastAppend().appendFile(dataFile).appendFile(dataFile2).appendFile(dataFile3).commit();
+    table.refresh();
+  }
+
+  @Test
+  public void testScanPlanningWithReport() throws IOException {
```
rdblue, line 94: I think that we need to validate that each counter is being accumulated correctly. That's a lot of work, so we may want to scope this PR down and add just a few high-level counters and then get more specific. That way we can add counters and tests together to make sure we get all the cases.

How about adding just the manifest counters now and we can expand after that? Really whatever subset you have done and want to get tested is a good start. Could also be result counters.

api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
nastra: this is superseded now by https://github.com/apache/iceberg/pull/5386 where the name of the counter is also included

core/src/main/java/org/apache/iceberg/ManifestGroup.java
nastra: good point, added

core/src/main/java/org/apache/iceberg/ManifestGroup.java
nastra: ah that makes sense, thanks for double checking this. I added option 2 and I'm counting total data/delete files in the `ManifestReader`

core/src/main/java/org/apache/iceberg/DeleteFileIndex.java
nastra: good catch as I wasn't sure if that one is correct or not. I'm counting the entries that are processed in the first loop now

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: that sounds good to me. Let me remove some counters and push a new version

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,261 @@
...
+    ScanReport scanReport = reporter.lastReport();
+    assertThat(scanReport).isNotNull();
+    assertThat(scanReport.tableName()).isEqualTo("scan-planning-with-deletes");
+    assertThat(scanReport.snapshotId()).isEqualTo(2L);
+    assertThat(scanReport.scanMetrics().totalPlanningDuration().totalDuration())
+        .isGreaterThan(Duration.ZERO);
+    assertThat(scanReport.scanMetrics().resultDataFiles().value()).isEqualTo(3);
+    assertThat(scanReport.scanMetrics().resultDeleteFiles().value()).isEqualTo(2);
+    assertThat(scanReport.scanMetrics().scannedDataManifests().value()).isEqualTo(1);
+    assertThat(scanReport.scanMetrics().skippedDataManifests().value()).isEqualTo(0);
```
rdblue, line 211: I think we are going to need a test for skipped data manifests. I think you could probably achieve that with the current tests by splitting the write across 2 manifests and using a truncate(50) partition transform or just duplicating the test and truncating id instead of x. You may have to adjust values in the files a bit, but it's pretty close.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,261 @@
...
+            .createWriterFunc(GenericParquetWriter::buildWriter)
+            .build();
+    try {
+      appender.addAll(records);
+    } finally {
+      appender.close();
+    }
+
+    PartitionKey partitionKey = new PartitionKey(tbl.spec(), tbl.schema());
+    partitionKey.partition(records.get(0));
```
rdblue, line 233: The table doesn't have any partition fields, so this isn't quite right.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,261 @@
...
+            tableDir,
+            "scan-planning-with-deletes",
+            SCHEMA,
+            SPEC,
+            SortOrder.unsorted(),
+            formatVersion,
+            reporter);
+
+    tbl.newAppend().appendFile(FILE_A).appendFile(FILE_B).appendFile(FILE_C).commit();
+    tbl.newRowDelta().addDeletes(FILE_A_DELETES).addDeletes(FILE_B_DELETES).commit();
```
rdblue, line 195: You could use this setup to easily test the manifest file counting. I think these are partitioned files, so you'd just have to write them to separate manifests.

nastra: ah good point, I wasn't sure initially how to test that but after looking at some other tests I was able to write up a test that has multiple data manifests and data files, where we can verify how many data manifests we scan & skip

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: added this as described in https://github.com/apache/iceberg/pull/5268#discussion_r936430695

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: you're right, this seems like rather a leftover from a copy/paste. removed it

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,334 @@
...
+    assertThat(scanReport.scanMetrics().totalFileSizeInBytes().value()).isGreaterThan(0L);
+    assertThat(scanReport.scanMetrics().totalDeleteFileSizeInBytes().value()).isEqualTo(0L);
+  }
+
+  private Table createTableWithCustomRecords(String tableName) throws IOException {
+    Schema schema =
+        new Schema(
+            required(1, "id", Types.IntegerType.get()), required(2, "x", Types.StringType.get()));
+
+    Table tbl =
```
rdblue, line 151: In Iceberg, we use full names, rather than names that leave out vowels. It's typically no harder to type `table` than `tbl` and it's much more readable.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,334 @@
...
+    assertThat(scanReport.filter()).isEqualTo(Expressions.alwaysTrue());
+    assertThat(scanReport.scanMetrics().totalPlanningDuration().totalDuration())
+        .isGreaterThan(Duration.ZERO);
+    assertThat(scanReport.scanMetrics().resultDataFiles().value()).isEqualTo(3);
+    assertThat(scanReport.scanMetrics().resultDeleteFiles().value()).isEqualTo(0);
+    assertThat(scanReport.scanMetrics().scannedDataManifests().value()).isEqualTo(1);
+    assertThat(scanReport.scanMetrics().skippedDataManifests().value()).isEqualTo(0);
+    assertThat(scanReport.scanMetrics().totalDataManifests().value()).isEqualTo(1);
+    assertThat(scanReport.scanMetrics().totalDeleteManifests().value()).isEqualTo(0);
+    assertThat(scanReport.scanMetrics().totalFileSizeInBytes().value()).isGreaterThan(0L);
```
rdblue, line 79: I prefer the tests that add up the file size, rather than just asserting that it's non-zero.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,334 @@
...
+    assertThat(scanReport.scanMetrics().totalDeleteFileSizeInBytes().value()).isEqualTo(0L);
+
+    // should be two files
```
rdblue, line 82: I don't think there's a need for all 3 cases. How about testing all 3 files and just 1 file? I don't think that there is much of a difference between the 2 and 1 cases.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,334 @@
...
+    record2.setField("x", "30");
+    GenericRecord record3 = GenericRecord.create(schema);
+    record3.setField("id", 3);
+    record3.setField("x", "45");
+    GenericRecord record4 = GenericRecord.create(schema);
+    record4.setField("id", 3);
+    record4.setField("x", "51");
+    DataFile dataFile = writeParquetFile(tbl, Arrays.asList(record, record3));
+    DataFile dataFile2 = writeParquetFile(tbl, Arrays.asList(record2));
+    DataFile dataFile3 = writeParquetFile(tbl, Arrays.asList(record4));
```
rdblue, line 174: Can you add some whitespace here to make this more readable?

Maybe break up record creation from writes. Also, if you use Iceberg's `GenericRecord`, it's usually easier to create records. You just create one from the schema and call `copy(Map<String, Object>)` to make a new one with data values.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
```diff
@@ -0,0 +1,334 @@
...
+    ManifestFile manifestTwo =
+        writeManifest(
+            "manifest-file-2.avro",
+            manifestEntry(Status.EXISTING, snapshotIdTwo, FILE_C),
+            manifestEntry(Status.EXISTING, snapshotIdTwo, FILE_D));
+
+    RewriteManifests rewriteManifests = tbl.rewriteManifests();
+    manifests.forEach(rewriteManifests::deleteManifest);
+    rewriteManifests.addManifest(manifestOne).addManifest(manifestTwo);
+    rewriteManifests.commit();
```
rdblue, line 250: There's no need to use `rewriteManifests`. Each fast append will create a new manifest file, so I'd just use that instead.

We also need a test case that filters manifests and results in `skippedDataManifests` with a non-zero value.

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: yeah the initial version of the test was mainly for me to better understand how scan planning works. Reduced it to 3 files and 1 file

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: done

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: updated. Also the second part in that test method has `assertThat(scanReport.scanMetrics().skippedDataManifests().value()).isEqualTo(1)`

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: thanks for the tip. updated. I left record creation and writes together for now, but I'm planning to re-visit and improving it later (when adding & testing new metrics)

data/src/test/java/org/apache/iceberg/TestScanPlanningAndReporting.java
nastra: I probably accidentally named it this way to not interfere with the `table` in the super class. should be fixed now

rdblue (approved):

rdblue:
Thanks @nastra! I just merged this.

findepi:
Just FYI, this seems to have inflated logs quite noticeably (https://github.com/trinodb/trino/issues/15492).
I wonder whether INFO-level logging should be the default behavior.
Wouldn't it be better to have these logged at DEBUG by default?

PR #5254, opened 2022-07-12
log warning when filter conversion/bind expression fail

huaxingao:
Log a warning when filter conversion/bind expression fail to address this comment https://github.com/apache/iceberg/pull/5227#discussion_r917098256

PR #5274, opened 2022-07-14
only log error message(not the stack trace) if filter conversion/bind expression fail

huaxingao:
This is a follow up to https://github.com/apache/iceberg/pull/5254. When filter conversion/bind expression fail, we will only log the error message instead of the whole stack trace.

### DIFF
--- a/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
+++ b/spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
@@ -108,7 +108,7 @@ public class SparkScanBuilder implements ScanBuilder, SupportsPushDownFilters, S
         expr = SparkFilters.convert(filter);
       } catch (IllegalArgumentException e) {
         // converting to Iceberg Expression failed, so this expression cannot be pushed down
-        LOG.info("Failed to convert filter to Iceberg expression, skipping push down for this expression: {}.{}",
+        LOG.info("Failed to convert filter to Iceberg expression, skipping push down for this expression: {}. {}",
             filter, e.getMessage());
       }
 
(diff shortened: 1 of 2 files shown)


huaxingao:
cc @szehon-ho @bryanck @kbendick

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -108,8 +108,8 @@ public Filter[] pushFilters(Filter[] filters) {
         expr = SparkFilters.convert(filter);
       } catch (IllegalArgumentException e) {
         // converting to Iceberg Expression failed, so this expression cannot be pushed down
-        LOG.warn("Failed to convert filter to Iceberg expression, skipping push down for this expression: {}",
-            filter, e);
+        LOG.info("Failed to convert filter to Iceberg expression, skipping push down for this expression: {} {}",
```
kbendick, line 111: I’d maybe put a `.` or `-` in between the two templated values to delineate the two sentences but I’m ok with it as is as well.

kbendick (approved):
This LGTM.

Thanks @huaxingao for the quick fix!

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
huaxingao: added a `.` :)

spark/v3.2/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -108,8 +108,8 @@ public Filter[] pushFilters(Filter[] filters) {
         expr = SparkFilters.convert(filter);
       } catch (IllegalArgumentException e) {
         // converting to Iceberg Expression failed, so this expression cannot be pushed down
-        LOG.warn("Failed to convert filter to Iceberg expression, skipping push down for this expression: {}",
-            filter, e);
+        LOG.info("Failed to convert filter to Iceberg expression, skipping push down for this expression: {}.{}",
```
szehon-ho, line 111: Nit, could you add one space between the two {}, so that it's not jumbled into one word?

huaxingao: Done. Thanks!

szehon-ho:
Merged, thanks @huaxingao @kbendick @bryanck for the quick turnaround!

huaxingao:
Thank you all very much! @szehon-ho @kbendick @bryanck

PR #5275, opened 2022-07-14
Write metadata-location instead of metadataLocation

Fokko:
While talking to the REST API I noticed that the API returns `metadataLocation`, instead of `metadata-location` as defined in the Open API Spec: https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L1486

### DIFF
--- a/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
+++ b/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
@@ -102,12 +102,18 @@ public class LoadTableResponse implements RESTResponse {
       while (configNames.hasNext()) {
         String configKey = configNames.next();
         JsonNode configValue = config.get(configKey);
-        Preconditions.checkArgument(configValue.isTextual(), "Cannot parse config %s from non-string: %s", configKey, configValue);
+        Preconditions.checkArgument(
+                configValue.isTextual(), "Cannot parse config %s from non-string: %s", configKey, configValue);
         configEntries.put(configKey, configValue.textValue());
       }
     }
 
-    return LoadTableResponse.builder().withMetadataLocation(metadataLocation).withTableMetadata(metadata).addAllConfig(configEntries.build()).build();
+    return LoadTableResponse
+            .builder()
+            .withMetadataLocation(metadataLocation)
+            .withTableMetadata(metadata)
+            .addAllConfig(configEntries.build())
+            .build();
   }
 
   private String metadataLocation;
@@ -162,10 +168,10 @@ public class LoadTableResponse implements RESTResponse {
     private Builder() {
     }
 
-      public Builder withMetadataLocation(String metadataLocation) {
-          this.metadataLocation = metadataLocation;
-          return this;
-      }
+    public Builder withMetadataLocation(String location) {
+      this.metadataLocation = location;
+      return this;
+    }
 
     public Builder withTableMetadata(TableMetadata tableMetadata) {
       if (this.metadataLocation == null) {
--- a/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
+++ b/core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
@@ -22,15 +22,18 @@ package org.apache.iceberg.rest.responses;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JsonNode;
 import java.io.IOException;
+import java.io.StringWriter;
 import java.util.Iterator;
 import java.util.Map;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableMetadataParser;
+import org.apache.iceberg.exceptions.RuntimeIOException;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -89,8 +144,15 @@ public static class Builder {
     private Builder() {
     }
 
+      public Builder withMetadataLocation(String metadataLocation) {
+          this.metadataLocation = metadataLocation;
+          return this;
+      }
```
kbendick, line 150: +1 to this. I find that sometimes, the `TableMetadata` that gets passed has `null` for `metadataLocation`, in which case we need to use the `TableMetadataBuilder` to add it which isn't very convenient.  👍

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -37,6 +42,56 @@
...
+    generator.writeStartObject();
+
+    generator.writeStringField(METADATA_LOCATION, loadTable.metadataLocation);
+
+    generator.writeFieldName(METADATA);
+    TableMetadataParser.toJson(loadTable.metadata, generator);
+
+    generator.writeObjectFieldStart(CONFIG);
+    for (Map.Entry<String, String> keyValue : loadTable.config.entrySet()) {
+      generator.writeStringField(keyValue.getKey(), keyValue.getValue());
```
kbendick, line 59: Do we need a null check here? I believe this throws if `keyValue.getValue()` is null.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
```diff
@@ -37,6 +42,56 @@
...
+
+  public static LoadTableResponse fromJson(JsonNode node) {
+    Preconditions.checkArgument(node.isObject(),
+            "Cannot parse table response from a non-object: %s", node);
+
+    String metadataLocation = null;
+    JsonNode  metadataLocationNode = node.get(METADATA_LOCATION);
+    if(metadataLocationNode != null) {
+      metadataLocation = metadataLocationNode.asText();
+    }
```
kbendick, line 74: You can use `String metadataLocation = JsonUtil.getStringOrNull(METADATA_LOCATION, node)` for convenience here.

kbendick (commented):
Technically this isn't presently needed, though the current strategy imposes the requirement that people use the `RESTObjectMapper` which has [kebab-case naming strategy.](https://github.com/apache/iceberg/blob/6a06cb3bd24c101d52e7987f55039b6b9692831d/core/src/main/java/org/apache/iceberg/rest/RESTObjectMapper.java#L43)

But this does make the requirement that people use the RESTObjectMapper or otherwise create one with the same characteristics, which is a limitation we should ideally remove as enforcing specific configuration on an ObjectMapper is hard to do in some frameworks without polluting things globally.

For reference, we can see that the read->write->read loop is tested in TestLoadTableResponse both [here in the round trip serialization comparison of the JSON string to itself](https://github.com/apache/iceberg/blob/6a06cb3bd24c101d52e7987f55039b6b9692831d/core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java#L120). All of the request / response object tests also [check that the serialized JSON form only has the fields from the spec](https://github.com/apache/iceberg/blob/6a06cb3bd24c101d52e7987f55039b6b9692831d/core/src/test/java/org/apache/iceberg/rest/RequestResponseTestBase.java#L73-L96).


The fields from the spec come [from the test subclasses here (for LoadTableResponse)](https://github.com/apache/iceberg/blob/6a06cb3bd24c101d52e7987f55039b6b9692831d/core/src/test/java/org/apache/iceberg/rest/responses/TestLoadTableResponse.java#L71-L73). The test checks for both expected and unexpected fields, so that if something is added and not added to the test, we'll notice.

At present, we'd also need a serializer / deserializer for `CreateTableRequest` which is the other request / response object with multi-word keys at the moment that relies on the kebab-case naming strategy.

So overall I am +1 to this if we can get some simple tests for it and clean up the checkstyle issues 🙂

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
Fokko: I copied this from the `TableMetadataParser` 🙈  Browsing through the code, I don't see any null checks.

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
Fokko: Ooh, that's nice. Thanks!

core/src/main/java/org/apache/iceberg/rest/responses/LoadTableResponse.java
rdblue: I don't think there should be any null values.

rdblue:
@Fokko, do you think we should move forward with this? As Kyle mentioned, if you use the included `RESTObjectMapper.mapper()` then the fields are sent and parsed correctly (as `metadata-location`). Did that solve the issue you ran into?

We were thinking that it wouldn't be necessary to add hand-coded parsers for each request/response class if we can ensure that the right mapper is always used. We do that in at least the client code.

Fokko:
@rdblue I missed your comment here. I agree that we don't need the custom encoder, and I'll close this one. The bug was on our impl that has been fixed a while ago 👍🏻

PR #5276, opened 2022-07-14
Python: Add type to the schema

Fokko:
Since the schema is an allOf the structType, we should also inherit the type: https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L1006-L1017

Also, I've added a check to avoid a nasty NPE


python/tests/utils/test_bin_packing.py
```diff
@@ -82,17 +81,3 @@ def weight_func(x):
         return x
 
     assert list(PackingIterator(splits, target_weight, lookback, weight_func, largest_bin_first)) == expected_lists
-
-
-def test_serialize_schema(table_schema_simple: Schema):
```
Fokko, line 87: Ended up in the wrong file

core/src/main/java/org/apache/iceberg/SchemaParser.java
```diff
@@ -254,8 +254,8 @@ private static Types.MapType mapFromJson(JsonNode json) {
   }
 
   public static Schema fromJson(JsonNode json) {
-    Type type  = typeFromJson(json);
-    Preconditions.checkArgument(type.isNestedType() && type.asNestedType().isStructType(),
+    Type type = typeFromJson(json);
+    Preconditions.checkArgument(type != null && type.isNestedType() && type.asNestedType().isStructType(),
```
rdblue, line 258: Could you open a separate PR for this change?

Fokko: Ah, of course. The Python changes are related, so I figured it would be okay to do it in one PR. I've opened https://github.com/apache/iceberg/pull/5291 with solely the Java change. I'll close this one and include the Python changes in the REST Catalog PR.

PR #3511, opened 2021-11-09
Core: Add DELETE/UPDATE/MERGE distribution mode properties

aokolnychyi:
This PR adds table properties for controlling DELETE/UPDATE/MERGE distribution mode. Right now, Spark uses only one distribution mode per table. In some cases, regular writes and row-level commands have very different characteristics and would benefit from a different distribution.

PR #5280, opened 2022-07-15
Docs: add missing table properties for update and merge write distrib…

dramaticlly:
Add missing documentation for `write.update.distribution-mode`

I believe
- `write.delete.distribution-mode`
- `write.update.distribution-mode`
- `write.merge.distribution-mode` (not super sure about this one, deducted from https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L228-L240)

are all introduced in #3511 but only one of them is shown up on documentation site. 

CC @aokolnychyi @samredai

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -71,6 +71,8 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.delete.target-file-size-bytes| 67108864 (64 MB)   | Controls the size of delete files generated to target about this many bytes |
 | write.distribution-mode            | none               | Defines distribution of write data: __none__: don't shuffle rows; __hash__: hash distribute by partition key ; __range__: range distribute by partition key or sort key if table has an SortOrder |
 | write.delete.distribution-mode     | hash               | Defines distribution of write delete data          |
+| write.update.distribution-mode     | hash               | Defines distribution of write update data          |
+| write.merge.distribution-mode      | none               | Defines distribution of write merge data           |
 | write.wap.enabled                  | false              | Enables write-audit-publish writes |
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -71,6 +71,8 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.delete.target-file-size-bytes| 67108864 (64 MB)   | Controls the size of delete files generated to target about this many bytes |
 | write.distribution-mode            | none               | Defines distribution of write data: __none__: don't shuffle rows; __hash__: hash distribute by partition key ; __range__: range distribute by partition key or sort key if table has an SortOrder |
 | write.delete.distribution-mode     | hash               | Defines distribution of write delete data          |
+| write.update.distribution-mode     | hash               | Defines distribution of write update data          |
+| write.merge.distribution-mode      | none               | Defines distribution of write merge data           |
```
szehon-ho, line 75: Looks like the default here is a bit complicated.  From my reading of the code, it picks 'write.distribution-mode' if it is set, else "range" if sorted, "none" if unsorted.  Do you see that as well, and should we mention that?

I feel we also should add something to spark-writes about distribution mode.

szehon-ho (commented):
Great to see this being documented, I also saw it missing and was thinking to take a look, then saw this pr.

aokolnychyi:
Oops, I missed this PR and merged #6683 instead as that one popped up earlier in my feed. Sorry, @dramaticlly!
I am going to close this one as that PR covers this.

We will need to add a section with examples from `TestSparkDistributionAndOrderingUtil` once #6679 is resolved.

PR #5247, opened 2022-07-11
Spark 3.3: Add Spark catalog for loading tables from cache

aokolnychyi:
This PR adds a new internal Spark catalog capable of loading tables from a static table cache, which allows us to reliably load tables in actions without passing a fully quoted Spark identifier.

PR #5284, opened 2022-07-15
[WIP] Spark-3.3: Handle dynamic catalog configurations

ajantha-bhat:
Some catalogs like Nessie supports dynamic catalog configurations (like changing the branch name).
But catalogs obtained from spark always uses the original configuration (from SQLConf) instead of active session configuration. Hence, it needs an update.

Nessie end to end testcases (with Nessie spark extensions) are not maintained in the Iceberg repo. Hence, not added a testcase here. Will be added in the Nessie repo.

original issue from Nessie: https://github.com/projectnessie/nessie/issues/4589

### DIFF
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java
@@ -462,6 +462,14 @@ public class SparkCatalog extends BaseCatalog {
 
   @Override
   public final void initialize(String name, CaseInsensitiveStringMap options) {
+    SparkSession sparkSession = SparkSession.active();
+    // Some catalogs can have dynamic catalog configurations (like Nessie catalog).
+    // Hence, update to the latest catalog configurations.
+    Map<String, String> updatedOptions = Maps.newHashMap();
+    Arrays.stream(sparkSession.sparkContext().conf()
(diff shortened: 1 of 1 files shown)


rdblue (changes requested):
I think this is incorrect.

In Spark, v2 catalogs are tracked in `sparkSession.sessionState.catalogManager`, so the set of catalogs is specific to a session. That's on purpose because you can define catalogs in Spark config when creating a new session.

This attempts to allow a catalog to use the active session instead of the one that created it, which is not correct.

ajantha-bhat:
> I think this is incorrect.
This attempts to allow a catalog to use the active session instead of the one that created it, which is not correct.

@rdblue: I see that other catalogs doesn't support the Dynamic catalog properties. So, the base problem is Nessie supporting configuring reference name dynamically as a spark catalog conf?

Any suggestion to handle this?

ajantha-bhat:
Somehow this problem disappeared.
Iceberg 0.14.0 spark3.2 and spark3.3 doesn't have this issue anymore.

Only spark 3.1 has an issue.

I still need to figure out which PR fixed this.

ajantha-bhat:
Found it. 
Fixed by https://github.com/apache/iceberg/pull/5247 by not using SparkCatalog for the rewriteDataFIles, which was using SQLConf instead of SessionConf for building the IcebergCatalog.

PR #5286, opened 2022-07-15
API: Introduce DefaultMetricsContext and Timer interface

nastra:
The idea here is that a `DefaultMetricsContext` can be used with native
Java Counters/Timers to measure things.

Additionally, this is introducing an API and a default implementation
for a Timer, which can be used to measure execution time.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.metrics;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A default {@link MetricsContext} implementation that uses native Java counters/timers.
+ */
+public class DefaultMetricsContext implements MetricsContext {
+
+  @Override
+  public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit unit) {
+    if (Integer.class.equals(type)) {
+      return (Counter<T>) new IntCounter();
+    }
+
+    if (Long.class.equals(type)) {
+      return (Counter<T>) new LongCounter();
+    }
+    throw new IllegalArgumentException(String.format("Counter for type %s is not supported", type.getName()));
+  }
+
+  @Override
+  public Timer timer(String name, TimeUnit unit) {
+    return new DefaultTimer(unit);
+  }
+}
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
```diff
@@ -0,0 +1,44 @@
...
+      return (Counter<T>) new IntCounter();
+    }
+    if (Long.class.equals(type)) {
+      return (Counter<T>) new LongCounter();
+    }
+    throw new IllegalArgumentException(String.format("Counter for type %s is not supported", type.getName()));
+  }
+
+  @Override
+  public Timer timer(String name, TimeUnit unit) {
```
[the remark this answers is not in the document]
danielcweeks: I think that's fine.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  /**
+   * Converts this {@code TimeUnit} to the equivalent {@code ChronoUnit}.
+   *
+   * @return the converted equivalent ChronoUnit
+   */
+  private static ChronoUnit toChronoUnit(TimeUnit timeUnit) {
+    switch (timeUnit) {
```
danielcweeks, line 135: It seems like there should be a better way to do this with the native libraries, but I see that wasn't introduced until Java 9.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,154 @@
...
+  @Override
+  public Sample start() {
+    boolean updated = stopwatchRef.compareAndSet(null, Stopwatch.createStarted());
+    Preconditions.checkState(updated, "startTimer() called multiple times");
+    return new DefaultSample(stopwatchRef, defaultTimeUnit);
+  }
+
+  @Override
+  public void record(long amount, TimeUnit unit) {
+    if (amount >= 0) {
```
danielcweeks, line 66: We might want to make this a precondition check. I would assume if somebody is recording a negative time there's actually a problem.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,154 @@
...
+      this.duration = this.duration.plus(amount, toChronoUnit(unit));
```
danielcweeks, line 67: I feel like there might be some synchronization missing here. Duration is being reset potentially for multiple threads.

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public long count() {
+    return count.get();
+  }
+
+  @Override
+  public Duration totalDuration() {
+    return duration;
```
danielcweeks, line 54: We might want to consider throwing if there are any running samples. It seems all too easy that something might be running in a thread and have not completed when total duration gets called. Then it wouldn't be included in the total.

rdblue (changes requested):
Sorry, I didn't see this until I left a lot of comments on the combined PR. Could you take a look at my review comments from that one?

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
nastra: yeah this is only available in JDK9+, so I had to write up something for this

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
nastra: should be fixed now

api/src/main/java/org/apache/iceberg/metrics/DefaultTimer.java
nastra: done

rdblue (approved):

rdblue:
Looks great. Thanks for adding this, @nastra!

PR #5287, opened 2022-07-15
Python: First version of the rest catalog

Fokko:


python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,365 @@
...
+
+    def _handle_non_200_response(self, exc: HTTPError, error_handler: Dict[int, Type[Exception]]):
+        try:
+            response = ErrorResponse(**exc.response.json())
+        except Exception:  # pylint: disable=W0012,W0703
+            # In the case we don't have a proper response
+            response = ErrorResponse(
+                error=ErrorResponseMessage(
+                    message=f"Could not decode json payload: {exc.response.text}",
+                    type="UnhandledError",
```
kbendick, line 220: Nit: In the Java HTTPClient, when we can't parse the response, we use `"RESTException"` for the type.

https://github.com/apache/iceberg/blob/8155e5e8129365ed8d33b8cced2e851734a302c5/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java#L92-L103

core/src/main/java/org/apache/iceberg/SchemaParser.java
```diff
@@ -254,8 +254,8 @@ private static Types.MapType mapFromJson(JsonNode json) {
   }
 
   public static Schema fromJson(JsonNode json) {
-    Type type  = typeFromJson(json);
-    Preconditions.checkArgument(type.isNestedType() && type.asNestedType().isStructType(),
+    Type type = typeFromJson(json);
+    Preconditions.checkArgument(type != null && type.isNestedType() && type.asNestedType().isStructType(),
```
rdblue, line 258: Can you open a separate PR for Java fixes? This looks good to me, but we try not to mix changes like this into other PRs to avoid commit conflicts when cherry picking to maintain branches.

python/pyiceberg/catalog/base.py
```diff
@@ -57,7 +59,8 @@ def create_table(
         identifier: str | Identifier,
         schema: Schema,
         location: str | None = None,
-        partition_spec: PartitionSpec | None = None,
+        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
+        sort_order: SortOrder = UNSORTED_SORT_ORDER,
         properties: Properties | None = None,
```
rdblue, line 64: Should we similarly default `properties` to `{}`?

python/tests/table/test_metadata.py
```diff
@@ -515,7 +516,7 @@ def test_v1_write_metadata_for_v2():
     ]
     assert metadata_v2["default-spec-id"] == 0
     assert metadata_v2["last-partition-id"] == 1000
-    assert metadata_v2["sort-orders"] == [{"order-id": 0, "fields": []}]
+    assert metadata_v2["sort-orders"] == []
```
rdblue, line 519: Why isn't this the unsorted order?

python/pyiceberg/table/metadata.py
```diff
@@ -347,10 +345,10 @@ class TableMetadata:
 
     @staticmethod
     def parse_obj(data: dict) -> Union[TableMetadataV1, TableMetadataV2]:
-        if "format-version" not in data:
+        if "format-version" not in data and "format_version" not in data:
```
rdblue, line 348: This seems suspicious to me. When would this be called with `format_version` and not `format-version`?

python/pyiceberg/table/base.py
```diff
@@ -15,17 +15,17 @@
...
 
 
-class Table(ABC):
-    """Placeholder for Table managed by the Catalog that points to the current Table Metadata.
-
-    To be implemented by https://github.com/apache/iceberg/issues/3227
-    """
-
-    identifier: str | Identifier
+class Table(IcebergBaseModel):
```
rdblue, line 27: Does `Table` need to be serializable? I'm curious why Table extends `IcebergBaseModel`. It seems like it should be a regular class to me.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,365 @@
...
+        self, namespace: Union[str, Identifier], removals: Optional[Set[str]] = None, updates: Optional[Properties] = None
+    ) -> UpdateNamespacePropertiesResponse:
+        namespace = NAMESPACE_SEPARATOR.join(self.identifier_to_tuple(namespace))
+        payload = {"removals": list(removals or []), "updates": updates}
+        response = requests.post(self.url(Endpoints.update_properties, namespace=namespace), json=payload, headers=self.headers)
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError})
+        return UpdateNamespacePropertiesResponse(**response.json())
```
rdblue, line 365: Why update the method's type to be able to return an internal response class?

core/src/main/java/org/apache/iceberg/SchemaParser.java
Fokko: I've actually cherry-picked those changes from https://github.com/apache/iceberg/pull/5276 in this branch. Once that one is in, I'll rebase this branch 👍🏻

rdblue: Sounds good to fix this one, but I think the Java changes should still go in a separate PR from the python ones.

Fokko: Ah since the Python changes are related, I figured it would make sense to do it in a single PR. I just created https://github.com/apache/iceberg/pull/5291 for the Java change.

python/pyiceberg/catalog/base.py
Fokko: I would like that, but that's actually a code smell in Python (or a quirk in the language). The default `{}` will be a reference to a single object, if you mutate that one, the next time the default value is being assigned, it will give the reference to the same object. More info here: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/ So the recommended way is just to set it to null, and then do `properties or {}` in the code. This `{}` will then always initialize an empty dict.

python/pyiceberg/table/metadata.py
Fokko: Yea, this is supspicious. I'm going to remove all the `pre_validators` because you can enter there the normal field name (`format_version`) and the alias (`format-version`) which makes it very confusing, clutters up the code, and makes it harder to comprehend the code.

python/pyiceberg/catalog/rest.py
Fokko: Ah, thanks! I like it as it is more descriptive. I've changed it into `RESTError` to keep it a Pythonic.

python/pyiceberg/catalog/rest.py
Fokko: I've created a new class next to the interface. I agree that returning to an internal response class isn't the best idea, and meant to clean that up. The REST API returns which fields are being removed, updated, or missing. That's something that we definitely want to expose in a CLI or in orchestrators.

python/pyiceberg/table/base.py
Fokko: Yes, it needs to be serializable, and it is the equivalent of the response of the REST API: https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L1481-L1503

The `CreateTableResponse` is an alias of the `LoadTableResponse`.

python/tests/table/test_metadata.py
Fokko: Thanks, that was actually a bug

python/pyiceberg/table/metadata.py
```diff
@@ -196,28 +197,6 @@ class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
     # because bumping the version should be an explicit operation that is up
     # to the owner of the table.
 
-    @root_validator(pre=True)
```
Fokko, line 199: Removing the pre-validators because they aren't very convenient. We have to check everything that goes into it, before any aliases are applied, which makes them messy.

python/pyiceberg/table/metadata.py
Fokko: Okay, this is actually the only instance that we need this. Because we don't know if the aliases are already applied or not.

python/tests/utils/test_bin_packing.py
```diff
@@ -82,17 +81,3 @@ def weight_func(x):
         return x
 
     assert list(PackingIterator(splits, target_weight, lookback, weight_func, largest_bin_first)) == expected_lists
-
-
-def test_serialize_schema(table_schema_simple: Schema):
```
Fokko, line 87: These tests also live in `test_schema.py`.

python/.ipynb_checkpoints/Untitled-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
```
rdblue, line 6: Was this added by accident or is it needed?

Fokko: Oops, that was by accident. I used a notebook to test the rest catalog and go over some of the endpoints.

python/pyiceberg/table/base.py
rdblue: I don't think the `LoadTableResponse` should be the same class as Table. Table is part of the public API and will expose a lot of methods and operations to users. `LoadTableResponse` is specific to the REST catalog and includes fields that aren't part of the table (`config`).

python/pyiceberg/catalog/rest.py
rdblue: Does this need to return anything at all? I'm okay either way, I think.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,371 @@
...
+NAMESPACE_SEPARATOR = b"\x1F".decode("UTF-8")
+
+
+class CreateTableRequest(IcebergBaseModel):
+    name: str = Field()
+    location: Optional[str] = Field()
+    table_schema: Schema = Field(alias="schema")
+    partition_spec: Optional[PartitionSpec] = Field(alias="partition-spec")
+    write_order: Optional[SortOrder] = Field(alias="write-order")
+    stage_creation: bool = Field(alias="stage-creation", default=False)
```
rdblue, line 90: I think this should be `stage-create` rather than `stage-creation`.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,371 @@
...
+    properties: Dict[str, str] = Field(default_factory=dict)
+
+
+class TokenResponse(IcebergBaseModel):
+    access_token: str = Field()
+    token_type: str = Field()
+    expires_in: int = Field()
+    warehouse_id: str = Field()
```
rdblue, line 98: `region` and `warehouse_id` aren't in the spec. Maybe this is something returned by an implementation? We probably don't want to include them here.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,371 @@
...
+    region: str = Field()
+    issued_token_type: str = Field()
+
+
+class ConfigResponse(IcebergBaseModel):
+    defaults: Dict[str, str] = Field()
```
rdblue, line 104: What about using `Properties` here as well?

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+class CreateTableRequest(IcebergBaseModel):
+    name: str = Field()
+    location: Optional[str] = Field()
+    table_schema: Schema = Field(alias="schema")
+    partition_spec: Optional[PartitionSpec] = Field(alias="partition-spec")
+    write_order: Optional[SortOrder] = Field(alias="write-order")
+    stage_creation: bool = Field(alias="stage-creation", default=False)
+    properties: Dict[str, str] = Field(default_factory=dict)
```
rdblue, line 91: Use `Properties` here?

python/pyiceberg/table/metadata.py
```diff
@@ -84,12 +85,12 @@ class TableMetadataCommonFields(IcebergBaseModel):
     def current_schema(self) -> Schema:
         return next(schema for schema in self.schemas if schema.schema_id == self.current_schema_id)
 
-    @root_validator(pre=True)
+    @root_validator()
     def cleanup_snapshot_id(cls, data: Dict[str, Any]):
-        if data.get("current-snapshot-id") == -1:
+        if data.get("current_snapshot_id") == -1:
```
rdblue, line 90: Can the changes to remove prevalidators be done in a separate PR?

python/pyiceberg/table/metadata.py
rdblue: Why don't we know if the aliases have already been applied?

Fokko: This is before anything pydantic, but the data might come from actual `json`, or from a `dict` that already has been parsed. In that case, the `-` will be replaced by `_`, and pydantic will handle this correctly. That said, I don't like it very much, and let me revert this.

python/pyiceberg/table/base.py
Fokko: This is defined by the Catalog: https://github.com/apache/iceberg/blob/84951419a9ab0aa39f9304b3dce91a179f4e643b/python/pyiceberg/catalog/base.py#L54-L77

How about subclassing the table:
```
class RestTable(Table):
    config: Dict[str, str] = Field(default_factory=dict)
```
And moving the specific items to there.

python/pyiceberg/catalog/rest.py
Fokko: Yes, I would like that, for the abovementioned reason 👍🏻

python/pyiceberg/catalog/rest.py
Fokko: Thanks, not sure how that ended up there.

python/pyiceberg/catalog/rest.py
Fokko: Looks like it is returned by an implementation ;)

python/pyiceberg/catalog/rest.py
Fokko: I like that a lot, thanks!

python/pyiceberg/catalog/rest.py
Fokko: Great idea!

python/tests/conftest.py
```diff
@@ -76,7 +76,7 @@ def table_schema_simple():
         NestedField(field_id=2, name="bar", field_type=IntegerType(), required=True),
         NestedField(field_id=3, name="baz", field_type=BooleanType(), required=False),
         schema_id=1,
-        identifier_field_ids=[1],
+        identifier_field_ids=[2],
```
Fokko, line 79: Field 1 is not required, and this makes it an invalid identifier field

python/pyiceberg/table/metadata.py
Fokko: I've removed those 👍🏻

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+
+        Returns:
+            The base url of the rest catalog
+        """
+
+        url = self.host
+        url = url if url.endswith("/") else url + "/"
+
+        if prefixed:
+            url += self.config.overrides.get(PREFIX, "")
```
rdblue, line 186: The config response shouldn't be used like this. The prefix can be configured in defaults, catalog config, or overrides. It doesn't matter where it comes from and the client shouldn't assume that certain settings are set through a specific path.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+    code: int = Field()
+
+
+class ErrorResponse(IcebergBaseModel):
+    error: ErrorResponseMessage = Field()
+
+
+class RestCatalog(Catalog):
+    token: TokenResponse
+    config: ConfigResponse
```
rdblue, line 146: I think this should be tracked a string map, not a `ConfigResponse`. See my comment in `url`

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+        if prefixed:
+            url += self.config.overrides.get(PREFIX, "")
+            url = url if url.endswith("/") else url + "/"
+
+        return url + endpoint.format(**kwargs)
+
+    def _fetch_access_token(self, token: str) -> TokenResponse:
+        client, secret = self._split_token(token)
+        data = {GRANT_TYPE: CLIENT_CREDENTIALS, CLIENT_ID: client, CLIENT_SECRET: secret, SCOPE: CATALOG_SCOPE}
+        response = requests.post(self.url(Endpoints.get_token, prefixed=False), data=data)
```
rdblue, line 194: Could you add a comment that this will use `application/x-www-form-urlencoded` by default?

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+        """Constructs the endpoint
+
+        Args:
+            prefixed: If the prefix return by the config needs to be appended
+
+        Returns:
+            The base url of the rest catalog
+        """
+
+        url = self.host
```
rdblue, line 182: It looks like `host` doesn't correctly describe the use of this setting, since it must include `v1`. I think that Python should use the same config settings that are used in most of the Java library, which to set the catalog URI using `uri`. And to match Java, that should not include `v1/`, which should be added to the path in this method.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+        url = url if url.endswith("/") else url + "/"
```
rdblue, line 183: `url = url + "v1/" if url.endsWith("/") else url + "/v1/"`

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+                )
+            )
+
+        code = exc.response.status_code
+        if code in error_handler:
+            raise error_handler[code](response) from exc
+        elif code == 400:
+            raise BadRequestError(response.error.message) from exc
+        elif code == 401:
+            raise UnauthorizedError(response.error.message) from exc
```
rdblue, line 234: I think this should also handle the other standard errors, like the [Java default handler does](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/rest/ErrorHandlers.java#L106-L126).

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+        elif code == 503:
+            raise ServiceUnavailableError(response.error.message) from exc
+        elif 500 <= code < 600:
+            raise ServerError(response.error.message) from exc
+        else:
+            raise RESTError(response.error.message) from exc
```
rdblue, line 240: I think that this should be `response` so that the entire response context is included in the trace.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+                error=ErrorResponseMessage(
+                    message=f"Could not decode json payload: {exc.response.text}",
+                    type="RESTError",
+                    code=exc.response.status_code,
+                )
+            )
+
+        code = exc.response.status_code
+        if code in error_handler:
+            raise error_handler[code](response) from exc
```
rdblue, line 230: I think this should be `response.error.message` like the other cases. That's also what Java does.

python/pyiceberg/exceptions.py
```diff
@@ -34,3 +38,27 @@ class AlreadyExistsError(Exception):
 
 class ValidationError(Exception):
     """Raises when there is an issue with the schema"""
+
+
+class BadCredentialsError(Exception):
```
rdblue, line 43: Should the REST exceptions all inherit from `RESTError`? That gives callers more flexibility to catch REST-related problems in a single `except`.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+        return {"namespace": NAMESPACE_SEPARATOR.join(identifier[:-1]), "table": identifier[-1]}
+
+    def _split_namespace_and_name(self, identifier: Union[str, Identifier]) -> Dict[str, Union[Tuple[str, ...], str]]:
+        identifier = self.identifier_to_tuple(identifier)
+        return {"namespace": identifier[:-1], "name": identifier[-1]}
+
+    def _handle_non_200_response(self, exc: HTTPError, error_handler: Dict[int, Type[Exception]]):
+        try:
+            response = ErrorResponse(**exc.response.json())
+        except Exception:  # pylint: disable=W0012,W0703
```
rdblue, line 218: `request.json()` says that it will raise `requests.exceptions.JSONDecodeError`. Is it a good idea to catch `Exception` instead of the specific one?

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def create_table(
+        self,
+        identifier: Union[str, Identifier],
+        schema: Schema,
+        location: Optional[str] = None,
+        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
+        sort_order: SortOrder = UNSORTED_SORT_ORDER,
+        properties: Optional[Properties] = None,
+    ) -> RestTable:
```
rdblue, line 250: I think this should return `Table`, not `RestTable`. There should be no difference between catalogs. The config in the `LoadTableResponse` is used to configure the catalog's interaction with the service for that table, not to configure the table itself.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+CLIENT_CREDENTIALS = "client_credentials"
+CREDENTIAL = "credential"
+GRANT_TYPE = "grant_type"
+SCOPE = "scope"
+TOKEN_EXCHANGE = "urn:ietf:params:oauth:grant-type:token-exchange"
+
+NAMESPACE_SEPARATOR = b"\x1F".decode("UTF-8")
+
+
+class RestTable(Table):
```
rdblue, line 84: I think this needs a `LoadTableResponse` rather than a `RestTable`. The response is the correct place to use `config`. The table should be a `Table` no matter what catalog loaded it.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+    UnauthorizedError,
+)
+from pyiceberg.schema import Schema
+from pyiceberg.table.base import Table
+from pyiceberg.table.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
+from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
+from pyiceberg.utils.iceberg_base_model import IcebergBaseModel
+
+
+class Endpoints:
```
rdblue, line 52: This is okay, but there are several duplicated routes. I think it would make sense to use `Resources` and name the variables after the resource. I think that's easier to map to REST when reading, because to create a namespace you POST to `Resources.namespaces`. I don't feel too strongly about this, but I do keep coming back to this table to make sure the path string is correct.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+            self.url(Endpoints.create_table, namespace=namespace_and_table["namespace"]),
+            data=serialized_json,
+            headers=self.headers,
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {409: TableAlreadyExistsError})
+
+        return RestTable(identifier=identifier, **response.json())
```
rdblue, line 272: `identifier` should include the catalog name. Otherwise you can't tell where the table was loaded from if there are multiple catalogs.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+    def list_tables(self, namespace: Optional[Union[str, Identifier]] = None) -> List[Identifier]:
+        namespace_concat = NAMESPACE_SEPARATOR.join(self.identifier_to_tuple(namespace or ""))
+        response = requests.get(
+            self.url(Endpoints.list_tables, namespace=namespace_concat),
+            headers=self.headers,
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchTableError})
```
rdblue, line 283: 404 here should be `NoSuchNamespaceError` because there isn't a specific table.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+            self.url(Endpoints.drop_table, prefixed=True, purge=purge_requested, **self._split_namespace_and_table(identifier)),
+            headers=self.headers,
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchTableError})
+
+    def purge_table(self, identifier: Union[str, Identifier]) -> None:
+        raise NotImplementedError("Not implemented")
```
rdblue, line 309: I think this should call `self.drop_table(identifier, purge_requested=True)`.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+            self._handle_non_200_response(exc, {401: BadCredentialsError})
+
+        return TokenResponse(**response.json())
+
+    def _fetch_config(self) -> ConfigResponse:
+        response = requests.get(self.url(Endpoints.get_config, prefixed=False), headers=self.headers)
+        response.raise_for_status()
+        return ConfigResponse(**response.json())
+
+    def _split_namespace_and_table(self, identifier: Union[str, Identifier]) -> Properties:
```
rdblue, line 207: The difference between `_split_namespace_and_table` and `_split_namespace_and_name` is not very obvious, but it matters a lot which one you use! Is there a way to ensure that we are using the right one for JSON vs paths? Maybe we should have "path" in the name, like `_split_identifier_for_path` and `_split_identifier_for_json`?

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+            "source": self._split_namespace_and_name(from_identifier),
+            "destination": self._split_namespace_and_name(to_identifier),
+        }
+        response = requests.post(self.url(Endpoints.rename_table), json=payload, headers=self.headers)
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchTableError, 409: TableAlreadyExistsError})
+
+    def create_namespace(self, namespace: Union[str, Identifier], properties: Optional[Properties] = None) -> None:
```
rdblue, line 322: Should this send `{}` instead of `None`? I'm not sure which one is better.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def drop_namespace(self, namespace: Union[str, Identifier]) -> None:
+        namespace = NAMESPACE_SEPARATOR.join(self.identifier_to_tuple(namespace))
+        response = requests.delete(self.url(Endpoints.drop_namespace, namespace=namespace), headers=self.headers)
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError})
+
+    def list_namespaces(self) -> List[Identifier]:
```
rdblue, line 338: This will also need to eventually accept a `parent: Union[Tuple[str, ...], str` to list nested namespaces. It's okay for now though.

python/pyiceberg/table/base.py
rdblue: I think we will want to evolve this in the future, but it is okay for now.

We will need to think about whether we want to use the same `TableOperations` abstraction that Java uses. That allows us to have a base table implementation that can work with any catalog. In that design, the table wraps a `TableMetadata` that is handled (refreshed, committed) by `TableOperations`.

python/pyiceberg/table/metadata.py
```diff
@@ -195,27 +196,21 @@ class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
     # When writing, we should stick to the same version that it was,
     # because bumping the version should be an explicit operation that is up
     # to the owner of the table.
-
     @root_validator(pre=True)
     def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
         """Sets default values to be compatible with the format v2
-
```
rdblue, line 202: Can you revert some of the unnecessary whitespace changes?

python/tests/table/test_metadata.py
```diff
@@ -515,7 +516,7 @@ def test_v1_write_metadata_for_v2():
     ]
     assert metadata_v2["default-spec-id"] == 0
     assert metadata_v2["last-partition-id"] == 1000
-    assert metadata_v2["sort-orders"] == [{"order-id": 0, "fields": []}]
+    assert metadata_v2["sort-orders"] == [{"fields": [], "order-id": 0}]
```
rdblue, line 519: Can some of these changes be separated into different PRs so that we don't carry them forward as extra changes while the REST catalog is reviewed?

python/tests/catalog/test_rest.py
```diff
@@ -0,0 +1,670 @@
...
+)
+
+TEST_HOST = "https://iceberg-test-catalog/"
+TEST_TOKEN = "client:secret"
+
+
+def test_token_200():
+    with requests_mock.Mocker() as m:
+        m.post(
+            f"{TEST_HOST}oauth/tokens",
```
rdblue, line 53: This is missing `v1/`

python/tests/catalog/test_rest.py
```diff
@@ -0,0 +1,670 @@
...
+from pyiceberg.transforms import IdentityTransform, TruncateTransform
+from pyiceberg.types import (
+    BooleanType,
+    IntegerType,
+    NestedField,
+    StringType,
+)
+
+TEST_HOST = "https://iceberg-test-catalog/"
+TEST_TOKEN = "client:secret"
```
rdblue, line 47: This is a `credential`, not a `token`. Tokens are typically JWT tokens.

python/tests/catalog/test_rest.py
```diff
@@ -0,0 +1,670 @@
...
+def test_token_200():
+    with requests_mock.Mocker() as m:
+        m.post(
+            f"{TEST_HOST}oauth/tokens",
+            json={
+                "access_token": TEST_TOKEN,
+                "token_type": "Bearer",
+                "expires_in": 86400,
+                "warehouse_id": "8bcb0838-50fc-472d-9ddb-8feb89ef5f1e",
+                "region": "us-west-2",
```
rdblue, line 59: `warehouse_id` and `region` aren't in the spec and should probably be omitted in these tests.

python/tests/catalog/test_rest.py
```diff
@@ -0,0 +1,670 @@
...
+        status_code=200,
+    )
+
+
+def test_list_tables_200():
+    with requests_mock.Mocker() as m:
+        _configure_token_and_config_endpoint(m)
+        namespace = "examples"
+        m.get(
+            f"{TEST_HOST}oss/warehouses/8bcb0838-22vo-472d-1925-8feb89lidf1e/namespaces/{namespace}/tables",
```
rdblue, line 125: I think the tests should primarily use an empty prefix. There's no need for an arbitrary prefix in these cases.

This is one reason why the Java side uses a class to build URLs. We can test the prefix behavior of that class without testing whether every catalog method supports `prefix` correctly.

python/tests/catalog/test_rest.py
```diff
@@ -0,0 +1,670 @@
...
+
+def test_list_namespaces_200():
+    with requests_mock.Mocker() as m:
+        _configure_token_and_config_endpoint(m)
+        m.get(
+            f"{TEST_HOST}oss/warehouses/8bcb0838-22vo-472d-1925-8feb89lidf1e/namespaces",
+            json={"namespaces": [["default"], ["examples"], ["fokko"], ["system"]]},
+            status_code=200,
+        )
+        assert RestCatalog("rest", {}, TEST_HOST, TEST_TOKEN).list_namespaces() == [
```
rdblue, line 161: Most tests don't need to use `TEST_TOKEN`, right? If it isn't specific to what you're testing, I'd prefer to leave it out to keep code simpler.

rdblue: Removing this and the unnecessary prefix would mean you no longer need `_configure_token_and_config_endpoint`.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+    def update_namespace_properties(
+        self, namespace: Union[str, Identifier], removals: Optional[Set[str]] = None, updates: Optional[Properties] = None
+    ) -> PropertiesUpdateSummary:
+        namespace = NAMESPACE_SEPARATOR.join(self.identifier_to_tuple(namespace))
+        payload = {"removals": list(removals or []), "updates": updates}
+        response = requests.post(self.url(Endpoints.update_properties, namespace=namespace), json=payload, headers=self.headers)
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError})
```
rdblue, line 367: This also needs to handle 422, `UnprocessableError` for when the request attempts to delete and set the same property.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def list_namespaces(self) -> List[Identifier]:
+        response = requests.get(self.url(Endpoints.list_namespaces), headers=self.headers)
+        response.raise_for_status()
+        namespaces = ListNamespaceResponse(**response.json())
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {})
```
rdblue, line 345: This should handle `404` when `parent` is supported because the parent namespace may not exist.

python/pyiceberg/catalog/rest.py
rdblue: In Java, we don't have a special exception class and just use `RESTException` for this.

python/pyiceberg/catalog/base.py
rdblue: There's no immutable dict?

Fokko: Not by default, there is a package, or we can implement it ourselves.

python/pyiceberg/catalog/rest.py
Fokko: Done!

python/pyiceberg/catalog/rest.py
Fokko: Good one! I assumed that it was part of the host, but it shouldn't

python/pyiceberg/catalog/rest.py
Fokko: In the lines above we try to parse the ErrorResponse, if that doesn't work, we just return the plain response:
```
try:
    response = ErrorResponse(**exc.response.json())
except Exception:  # pylint: disable=W0012,W0703
    # In the case we don't have a proper response
    response = ErrorResponse(
        error=ErrorResponseMessage(
            message=f"Could not decode json payload: {exc.response.text}",
            type="RESTError",
            code=exc.response.status_code,
        )
    )
```

python/pyiceberg/catalog/rest.py
Fokko: Added them, thanks!

python/pyiceberg/catalog/rest.py
Fokko: Ah, great catch

python/pyiceberg/exceptions.py
Fokko: Yes, I like it, and updated this in the code

python/pyiceberg/catalog/rest.py
Fokko: Yes, I was also thinking of that. However, since this is already a sad path, I figured to catch everything. I don't have a very strong opinion on this, but when another error bubbles up, it might not be caught.

Fokko: I've changed this to `JSONDecodeError`

python/pyiceberg/catalog/rest.py
Fokko: Alright, but where do you want to keep the config?
```
"config": {
    "client.factory": "io.tabular.iceberg.catalog.TabularAwsClientFactory",
    "region": "us-west-2",
},
```
I can imagine that this is information that you would like to keep with the table.

python/pyiceberg/catalog/rest.py
Fokko: But then we're exposing the internal API?

python/pyiceberg/catalog/rest.py
Fokko: Ah, wasn't aware of that. Added those to create, load and list tables

python/pyiceberg/catalog/rest.py
Fokko: Got it.

python/pyiceberg/catalog/rest.py
Fokko: I'll update that in the code. I was assuming that it would only truncate the table.

python/pyiceberg/catalog/rest.py
Fokko: Yes, that's much better indeed

python/pyiceberg/catalog/rest.py
Fokko: This is because the properties aren't required by the spec: https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml

I think `{}` is better than `None`.

Fokko: I've updated the code: https://github.com/apache/iceberg/pull/5358

python/pyiceberg/catalog/rest.py
Fokko: Ah, I see. I've updated this by taking the defaults and updating it with the properties that are provided through the constructor and then applying the overrides.

python/pyiceberg/catalog/rest.py
Fokko: Changed this into `Properties`

python/pyiceberg/catalog/rest.py
Fokko: This feels slightly weird to me. Thinking of the CLI, should we also return a name when printing them to stdout?

python/pyiceberg/table/metadata.py
Fokko: Not sure why this happened.  This should be caught by the linter

python/tests/catalog/test_rest.py
Fokko: When initializing the rest-catalog, we fetch the token and credential right away, therefore we need to set those.

python/tests/catalog/test_rest.py
Fokko: I've removed the prefix.

python/pyiceberg/catalog/rest.py
Fokko: Good one, added!

python/pyiceberg/catalog/rest.py
Fokko: I've added it explicitly.

python/tests/catalog/test_rest.py
Fokko: And they are gone!

python/tests/catalog/test_rest.py
Fokko: Good one! I've renamed all the tokens to credentials 👍🏻

python/tests/catalog/test_rest.py
Fokko: Good one, I've updated all the tests

python/tests/table/test_metadata.py
Fokko: Unsure why this changed in the first place 🤔

python/pyiceberg/catalog/rest.py
Fokko: I've fixed this in the line below:
```python
def create_namespace(self, namespace: Union[str, Identifier], properties: Optional[Properties] = None) -> None:
    payload = {"namespace": self.identifier_to_tuple(namespace), "properties": properties or {}}
```

python/pyiceberg/catalog/base.py
rdblue: Maybe later, but this is fine for now.

python/pyiceberg/catalog/rest.py
rdblue: That config is for the catalog.

For example, the `client.factory` is passed to configure the `S3FileIO` instance that is passed into the table. The Catalog is responsible for injecting things like `FileIO`. That's how most customization gets injected. When we start supplying a `FileIO` from `Table`, we will want to do the same thing and pass this bundle of config to the `FileIO`.

python/pyiceberg/catalog/rest.py
rdblue: No, I mean that we should create a `Table` for the `TableMetadata` that comes back through `LoadTableResponse`.

python/pyiceberg/catalog/rest.py
rdblue: Actually, it is just the table's identifier that includes catalog name. When accepting or producing an identifier, the catalog's name will not be present. If you look up `db` in some catalog instance and see `t1` and `t2`, you don't want to strip the catalog name off of the returned result. So `[db, t1]` and `[db, t2]` are correct API results.

But when we create a table, we pass in a string version of the identifier (at least, in Java) so that the table can be identified by name in logs or other situations. It is reasonable to track that as an identifier here, but in that context I think we want to preserve information about the catalog that loaded the table. That's why I think we should add it here, but not elsewhere.

python/tests/catalog/test_rest.py
rdblue: But if there are no credentials, we should skip the fetch token call, right?

python/pyiceberg/catalog/rest.py
Fokko: Ah, like that. Updated!

python/pyiceberg/catalog/rest.py
Fokko: Got it. It is now part of the response, but not of the table.

python/pyiceberg/catalog/rest.py
rdblue: Can you rename this to `uri` to match the standard catalog config options?

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,435 @@
...
+        config = config_response.defaults
+        config.update(properties)
+        config.update(config_response.overrides)
+        return config
+
+    def _split_identifier_for_path(self, identifier: Union[str, Identifier]) -> Properties:
+        identifier = self.identifier_to_tuple(identifier)
+        return {"namespace": NAMESPACE_SEPARATOR.join(identifier[:-1]), "table": identifier[-1]}
+
+    def _split_identifier_for_json(self, identifier: Union[str, Identifier]) -> Dict[str, Union[Tuple[str, ...], str]]:
```
rdblue, line 252: Nit: The return type could be `Dict[str, Union[str, Identifier]]` instead of using `Tuple[str, ...]`.

python/pyiceberg/catalog/rest.py
rdblue: What I meant was that in the event that we aren't translating to a specific error type, I think it would make sense to pass `type` through from the `ErrorResponse`. This is minor, though.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,435 @@
...
+            self.url(Endpoints.list_tables, namespace=namespace_concat),
+            headers=self.headers,
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError})
+        response = ListTablesResponse(**response.json())
+
+        return [(self.name, *entry.namespace, entry.name) for entry in response.identifiers]
```
rdblue, line 341: I don't think this should include `self.name` because we want to be able to pass any of the returned identifiers back into the catalog. The only place where we want to add the catalog name for context is when we create the `Table`.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,435 @@
...
+CREDENTIAL = "credential"
+GRANT_TYPE = "grant_type"
+SCOPE = "scope"
+TOKEN_EXCHANGE = "urn:ietf:params:oauth:grant-type:token-exchange"
+
+NAMESPACE_SEPARATOR = b"\x1F".decode("UTF-8")
+
+
+class TableResponse(IcebergBaseModel):
+    metadata_location: Optional[str] = Field(alias="metadata-location", default=None)
```
rdblue, line 89: I see that this is optional in the spec, but I think we want to make it required in the spec. Otherwise, you wouldn't be able to tell when the metadata changes. The `refresh` call will call the load route again and update the metadata if it has changed.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def drop_table(self, identifier: Union[str, Identifier], purge_requested: bool = False) -> None:
+        response = requests.delete(
+            self.url(Endpoints.drop_table, prefixed=True, purge=purge_requested, **self._split_identifier_for_path(identifier)),
+            headers=self.headers,
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError})
```
rdblue, line 368: I think this should be `NoSuchTableException`. If you try to drop a table and get a 404, then the table doesn't exist.

rdblue: https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L673

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,435 @@
...
+    def rename_table(self, from_identifier: Union[str, Identifier], to_identifier: Union[str, Identifier]):
+        payload = {
+            "source": self._split_identifier_for_json(from_identifier),
+            "destination": self._split_identifier_for_json(to_identifier),
+        }
+        response = requests.post(self.url(Endpoints.rename_table), json=payload, headers=self.headers)
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError, 409: TableAlreadyExistsError})
```
rdblue, line 382: This should also be `NoSuchTableError`: https://github.com/apache/iceberg/blob/master/open-api/rest-catalog-open-api.yaml#L752-L753

Looks like the spec has an incorrect example there, but the 404 for this route means there was no such table to rename.

python/pyiceberg/table/base.py
```diff
@@ -15,17 +15,16 @@
...
+    identifier: Union[str, Identifier] = Field()
```
rdblue, line 28: Minor This should always be an identifier, right? When the catalog loads a table, its name will be parsed into parts.

python/tests/catalog/test_base.py
```diff
@@ -78,14 +75,14 @@ def load_table(self, identifier: Union[str, Identifier]) -> Table:
         try:
             return self.__tables[identifier]
         except KeyError as error:
-            raise NoSuchTableError(f"Table does not exist: {identifier}") from error
+            raise NoSuchNamespaceError(f"Table does not exist: {identifier}") from error
```
rdblue, line 78: I don't think this change is correct.

python/tests/catalog/test_base.py
```diff
@@ -78,14 +75,14 @@ def load_table(self, identifier: Union[str, Identifier]) -> Table:
...
 
     def drop_table(self, identifier: Union[str, Identifier]) -> None:
         identifier = Catalog.identifier_to_tuple(identifier)
         try:
             self.__tables.pop(identifier)
         except KeyError as error:
-            raise NoSuchTableError(f"Table does not exist: {identifier}") from error
+            raise NoSuchNamespaceError(f"Table does not exist: {identifier}") from error
```
rdblue, line 85: I pointed this out in the REST catalog, but this should still be `drop_table`. I think you were updating this based on my comment for the `list_tables` method. In that method, 404 means `NoSuchNamespaceError`, but for methods that identify a specific table, we want to use `NoSuchTableError`.

python/tests/catalog/test_rest.py
Fokko: I'm sorry, I meant config instead of credentials. I've created a small fixture for this:
```python
@pytest.fixture
def rest_mock(requests_mock: Mocker):
    """Takes the default requests_mock and adds the config endpoint to it

    This endpoint is called when initializing the rest catalog
    """
    requests_mock.get(
        f"{TEST_HOST}v1/config",
        json={"defaults": {}, "overrides": {}},
        status_code=200,
    )
    return requests_mock
```
Regardless if we use credentials or an existing token, we should fetch the config (this contains information such as the prefix, that we need for subsequent calls).

python/pyiceberg/catalog/rest.py
Fokko: Nice one, thanks!

python/pyiceberg/catalog/rest.py
Fokko: Got it, I've reverted this change.

python/pyiceberg/catalog/rest.py
Fokko: Having no metadata-location doesn't make much sense to me:

This is from the spec:

> The table metadata JSON is returned in the `metadata` field. The corresponding file location of table metadata should be returned in the `metadata-location` field, unless the metadata is not yet committed. For example, a create transaction may return metadata that is staged but not committed.

Also, the location is in the metadata itself, where it is a required field, so I don't really see why this should be optional.

python/pyiceberg/catalog/rest.py
Fokko: I can see that, but the catalog name is arbitrary and just passed in through the constructor of the catalog. I agree that we want to return the fully qualified name, including the database/schema.

python/pyiceberg/catalog/rest.py
Fokko: I actually came up with this by just creating a faulty call in the Java impl

python/pyiceberg/table/base.py
Fokko: Great catch, that's true!

python/tests/catalog/test_base.py
Fokko: I went over them, and fixed them according to the spec

python/tests/catalog/test_base.py
Fokko: Yes, I went over the list, sorry for the confusion

python/pyiceberg/catalog/rest.py
Fokko: I'm happy to create a mapping from the RESTException as defined in the catalog to our PythonError. We could also do this later to not add to this PR.

python/pyiceberg/exceptions.py
```diff
@@ -16,12 +16,8 @@
 #  under the License.
 
 
-class NoSuchTableError(Exception):
```
rdblue, line 19: I don't think that we want to change these errors, since they are independent of the REST catalog. I'd leave the existing ones unchanged and just make the REST-specific errors extend `RESTError`

python/pyiceberg/catalog/rest.py
rdblue: Yeah, 404 is context-specific. My comment was for the `list_tables` method, where a 404 indicates that the namespace does not exist and can't be listed. That shouldn't raise `NoSuchTableError` because 404 doesn't mean the namespace was empty -- that would result in an empty list of identifiers -- 404 means that the namespace itself doesn't exist.

Sorry for the confusion, I know it's a bit odd to have the meaning of 404 change based on the route, but it makes a bit more sense when you think about routes as identifying resources. `/namespace/ns/tables` primarily identifies a namespace and is interacting with the table set of that namespace.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,434 @@
...
+
+class ErrorResponse(IcebergBaseModel):
+    error: ErrorResponseMessage = Field()
+
+
+class RestCatalog(Catalog):
+    token: str
+    config: Properties
+
+    host: str
```
rdblue, line 155: I still think this should be `uri`, since that better describes the behavior: this is the base URI, not a host that will be embedded in `https://{host}:443/` to build URIs. That also matches the common options used by Java and are documented as [common catalog config](https://iceberg.apache.org/docs/latest/configuration/#catalog-properties).

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,434 @@
...
+        Args:
+            name: Name to identify the catalog
+            properties: Properties that are passed along to the configuration
+            host: The base-url of the REST Catalog endpoint
+            client_id: The id to identify the client
+            client_secret: The secret for the client
+            token: The bearer token
+        """
+        self.host = host
+        if client_id and client_secret:
```
rdblue, line 179: Can these be passed as a single string? I'm fine supporting both `credential` and `client_id`/`client_secret`. I'd just like to be consistent across implementations.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,434 @@
...
+            self.token = self._fetch_access_token(client_id, client_secret)
+        elif token:
+            self.token = token
+        else:
+            raise ValueError("Either set the client_id and client_secret, or provide a valid token")
```
rdblue, line 184: I don't think we should fail if there isn't a valid token. Not all environments will require auth.

python/pyiceberg/catalog/rest.py
rdblue: I think that the current mapping is fine. But if we have a service endpoint that sends something unexpected, it would be better to have the entire error response in the error that gets raised here. As it is right now, this will discard extra information and make it harder to debug if the service is doing something strange.

I see 3 options:
* Add extra data to `RESTError` that gets printed along with it: `raise RESTError(response.error.message, response.error)`
* Serialize the error as a string and use that for the message: `raise RESTError(repr(response.error))`
* Save the original exception and use it as context to make the `type` accessible: `raise RESTError from exc`

We may want to use `from exc` anyway? I'm not sure if that's going to be automatically included as `__cause__`, since this is in a handler function.

python/pyiceberg/catalog/rest.py
```diff
@@ -0,0 +1,434 @@
...
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            self._handle_non_200_response(exc, {404: NoSuchNamespaceError})
+        return [(*table.namespace, table.name) for table in ListTablesResponse(**response.json()).identifiers]
+
+    def load_table(self, identifier: Union[str, Identifier]) -> Table:
+        response = requests.get(
+            self.url(Endpoints.load_table, prefixed=True, **self._split_identifier_for_path(identifier)), headers=self.headers
```
rdblue, line 344: Nit: `prefixed=True` is the default.

python/tests/catalog/test_rest.py
```diff
@@ -0,0 +1,612 @@
...
+    RestCatalog("rest", {}, TEST_HOST, token=TEST_TOKEN).drop_table(("example", "fokko"))
+
+
+def test_delete_table_404(rest_mock: Mocker):
+    rest_mock.delete(
+        f"{TEST_HOST}v1/namespaces/example/tables/fokko",
+        json={
+            "error": {
+                "message": "Table does not exist: fokko.fokko2 in warehouse 8bcb0838-50fc-472d-9ddb-8feb89ef5f1e",
+                "type": "NoSuchNamespaceErrorException",
```
rdblue, line 604: `NoSuchTableException`?

rdblue (approved):

rdblue:
Awesome work, @Fokko! There are still a few minor comments, but there are no blockers so I merged this. Thanks!

python/pyiceberg/exceptions.py
Fokko: I had the same concern, also others could be applicable outside of the REST contest. I'll move `NoSuchNamespaceError` and `NoSuchTableError` back to inherit from `Exception`.

python/pyiceberg/catalog/rest.py
Fokko: Good one, I've renamed this one! 👍🏻

python/pyiceberg/catalog/rest.py
Fokko: Sure thing 👍🏻

python/pyiceberg/catalog/rest.py
Fokko: I didn't think of that, updated the code, thanks! 👍🏻

python/pyiceberg/catalog/rest.py
Fokko: Less is more 👍🏻

python/tests/catalog/test_rest.py
Fokko: Not sure where that comes from 🤔

PR #5294, opened 2022-07-18
Spark: Spark SQL read from Snapshot ref

hililiwei:
Reading data from a specified Ref in the following ways.
1. USE
```sql
USE { BRANCH \ TAG } refName
```
2. table path 
```sql
SELECT * FROM table.at_branch_{refName}

SELECT * FROM table.at_branch_{refName}.at_timestamp__{timestamp}

SELECT * FROM table.at_tag_{refName}

```
This is mainly in reference to our current implementation of time travel.

```sql
SELECT * FROM table.at_timestamp__{timestamp}
```

It depends on #4428 #5364

### DIFF
--- a/api/src/main/java/org/apache/iceberg/Table.java
+++ b/api/src/main/java/org/apache/iceberg/Table.java
@@ -305,4 +305,25 @@ public interface Table {
 
   /** Returns a {@link LocationProvider} to provide locations for new data files. */
   LocationProvider locationProvider();
+
+  /**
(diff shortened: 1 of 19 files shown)


dungdm93:
I prefer `SELECT * FROM table@refName` if it's possible

hililiwei:
> I prefer `SELECT * FROM table@refName` if it's possible

Unfortunately, Spark does not support this syntax.
I could try, but not optimistic.

PR #5037, opened 2022-06-14
Core: Implement BaseMetastoreCatalog.registerTable()

Mehul2500:
For migrating tables between different catalogs, need is there for a registerTable() function on destination catalog side.
Thus, planned to implement the registerTable() function in BaseMetastoreCatalog.

Closes #4995

PR #5297, opened 2022-07-18
Core : Catalog Tables Migration API

Mehul2500:
Introducing `migrateTables()` in `CatalogUtil` which could help in the migration of Iceberg tables in any Source Catalog to any Target Catalog. Uses PR #5037 , for the `registerTable()` functionality  in `BaseMetastoreCatalog`.

I used tables migrating from Hadoop Catalog to Hive Catalog for the test case.

### DIFF
--- a/aws/src/integration/java/org/apache/iceberg/aws/dynamodb/TestDynamoDbCatalog.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/dynamodb/TestDynamoDbCatalog.java
@@ -27,8 +27,10 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.iceberg.AssertHelpers;
 import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.HasTableOperations;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.Table;
+import org.apache.iceberg.TableOperations;
 import org.apache.iceberg.aws.AwsClientFactories;
 import org.apache.iceberg.aws.AwsClientFactory;
 import org.apache.iceberg.aws.AwsIntegTestUtil;
@@ -42,6 +44,7 @@ import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.types.Types;
+import org.assertj.core.api.Assertions;
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.BeforeClass;
@@ -293,6 +296,37 @@ public class TestDynamoDbCatalog {
             .key(DynamoDbCatalog.namespacePrimaryKey(namespace))
             .build());
     Assert.assertFalse("namespace must not exist", response.hasItem());
+  }
+
+  @Test
+  public void testRegisterTable() {
+    Namespace namespace = Namespace.of(genRandomName());
+    catalog.createNamespace(namespace);
+    TableIdentifier identifier = TableIdentifier.of(namespace, catalogTableName);
+    catalog.createTable(identifier, SCHEMA);
+    Table registeringTable = catalog.loadTable(identifier);
+    Assertions.assertThat(catalog.dropTable(identifier, false)).isTrue();
+    TableOperations ops = ((HasTableOperations) registeringTable).operations();
+    String metadataLocation = ((DynamoDbTableOperations) ops).currentMetadataLocation();
+    Assertions.assertThat(catalog.registerTable(identifier, metadataLocation)).isNotNull();
+    Assertions.assertThat(catalog.loadTable(identifier)).isNotNull();
+    Assertions.assertThat(catalog.dropTable(identifier, true)).isTrue();
+    Assertions.assertThat(catalog.dropNamespace(namespace)).isTrue();
+  }
+
+  @Test
+  public void testRegisterExistingTable() {
+    Namespace namespace = Namespace.of(genRandomName());
+    catalog.createNamespace(namespace);
+    TableIdentifier identifier = TableIdentifier.of(namespace, catalogTableName);
+    catalog.createTable(identifier, SCHEMA);
+    Table registeringTable = catalog.loadTable(identifier);
+    TableOperations ops = ((HasTableOperations) registeringTable).operations();
+    String metadataLocation = ((DynamoDbTableOperations) ops).currentMetadataLocation();
+    Assertions.assertThatThrownBy(() -> catalog.registerTable(identifier, metadataLocation))
+        .isInstanceOf(AlreadyExistsException.class);
+    Assertions.assertThat(catalog.dropTable(identifier, true)).isTrue();
+    Assertions.assertThat(catalog.dropNamespace(namespace)).isTrue();
   }
 
   private static String genRandomName() {
--- a/core/src/main/java/org/apache/iceberg/CatalogUtil.java
+++ b/core/src/main/java/org/apache/iceberg/CatalogUtil.java
@@ -20,10 +20,16 @@
 package org.apache.iceberg;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.iceberg.catalog.Catalog;
+import org.apache.iceberg.catalog.Namespace;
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -346,4 +353,63 @@ public static void configureHadoopConf(Object maybeConfigurable, Object conf) {
...
+    Catalog targetCatalog;
+    try {
+      targetCatalog = loadCatalog(targetCatalogProperties.get("catalogImpl"),
+          targetCatalogProperties.get("catalogName"), targetCatalogProperties, targetHadoopConfig);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(String.format(
+          "Cannot initialize Target Catalog implementation %s: %s", targetCatalogProperties.get("catalogImpl"),
+          e.getMessage()), e);
+    }
+    List<TableIdentifier> allIdentifiers = tableIdentifiers;
```
snazy, line 394: I think, this code should probably live in `Catalog`:  A new function like `Catalog.registerTableFromCatalog()` to "move" a single table to the current catalog. The `HadoopCatalog` could then do the special-handling in its implementation.

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -346,4 +353,63 @@ public static void configureHadoopConf(Object maybeConfigurable, Object conf) {
...
+    if (tableIdentifiers == null || tableIdentifiers.isEmpty()) {
+      List<Namespace> namespaces = (sourceCatalog instanceof SupportsNamespaces) ?
+          ((SupportsNamespaces) sourceCatalog).listNamespaces() : ImmutableList.of(Namespace.empty());
+      allIdentifiers = namespaces.stream().flatMap(ns ->
+          sourceCatalog.listTables(ns).stream()).collect(Collectors.toList());
+    }
+    List<TableIdentifier> migratedTableIdentifiers = new ArrayList<TableIdentifier>();
+    allIdentifiers.forEach(tableIdentifier -> {
```
snazy, line 402: I suspect this will run for a very long time, like when there a a lot of tables.
If things fail in the meantime, it's hard to resume after the failed table.
I.e. error handling here is tricky.

Not sure whether it is actually possible to properly handle the case when `registerTable` worked, but `dropTable` failed - in such a case you'd have the same table in two catalogs.

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -346,4 +353,63 @@ public static void configureHadoopConf(Object maybeConfigurable, Object conf) {
...
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(String.format(
+          "Cannot initialize Target Catalog implementation %s: %s", targetCatalogProperties.get("catalogImpl"),
+          e.getMessage()), e);
+    }
+    List<TableIdentifier> allIdentifiers = tableIdentifiers;
+    if (tableIdentifiers == null || tableIdentifiers.isEmpty()) {
+      List<Namespace> namespaces = (sourceCatalog instanceof SupportsNamespaces) ?
+          ((SupportsNamespaces) sourceCatalog).listNamespaces() : ImmutableList.of(Namespace.empty());
+      allIdentifiers = namespaces.stream().flatMap(ns ->
```
snazy, line 398: I suspect this will run for a very long time, like when there a a lot of tables.

core/src/main/java/org/apache/iceberg/CatalogUtil.java
```diff
@@ -346,4 +353,63 @@ public static void configureHadoopConf(Object maybeConfigurable, Object conf) {
...
+   * @param sourceCatalogProperties Source Catalog Properties
+   * @param targetCatalogProperties Target Catalog Properties
+   * @param sourceHadoopConfig      Source Catalog Hadoop Configuration
+   * @param targetHadoopConfig      Target Catalog Hadoop Configuration
+   * @return list of table identifiers for successfully migrated tables
+   */
+  public static List<TableIdentifier> migrateTables(List<TableIdentifier> tableIdentifiers,
+      Map<String, String> sourceCatalogProperties, Map<String, String> targetCatalogProperties,
+      Object sourceHadoopConfig, Object targetHadoopConfig) {
+    if (tableIdentifiers != null) {
```
snazy, line 372: Let's leave out the catalog instantiation and configuration here completely. I suspect that users have at least one of these catalogs already handy - and setting up "the same" catalog twice is superfluous.

nessie/src/main/java/org/apache/iceberg/nessie/NessieTableOperations.java
```diff
@@ -152,7 +152,8 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       }
     }
 
-    String newMetadataLocation = writeNewMetadata(metadata, currentVersion() + 1);
+    String newMetadataLocation = (base == null) && (metadata.metadataFileLocation() != null) ?
```
snazy, line 155: Why is this necessary?
It's a new commit, not sure whether it is good that it re-uses an existing metadata location that is (potentially) "owned" by another catalog.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -385,6 +388,89 @@ public void testDropTable() throws IOException {
     verifyCommitMetadata();
   }
 
+  private void validateRegister(TableIdentifier identifier, String metadataVersionFiles) {
+    Assertions.assertThat(catalog.registerTable(identifier, "file:" + metadataVersionFiles)).isNotNull();
+    Table newTable = catalog.loadTable(identifier);
+    Assertions.assertThat(newTable).isNotNull();
+    TableOperations ops = ((HasTableOperations) newTable).operations();
+    String metadataLocation = ((NessieTableOperations) ops).currentMetadataLocation();
+    Assertions.assertThat("file:" + metadataVersionFiles).isEqualTo(metadataLocation);
```
snazy, line 397: Hint: Your assertions are often the "wrong" way around.
It's always `assertThat(<current state>)....`, followed by the expectations.

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -385,6 +388,89 @@ public void testDropTable() throws IOException {
...
+    Assertions.assertThat(catalog.dropTable(identifier, false)).isTrue();
+  }
+
+  @Test
+  public void testRegisterTableWithGivenBranch() {
+    List<String> metadataVersionFiles = metadataVersionFiles(TABLE_NAME);
+    Assertions.assertThat(1).isEqualTo(metadataVersionFiles.size());
```
snazy, line 404: Hint: `assertThat(metadataVersionFiles).hasSize(1)`

nessie/src/test/java/org/apache/iceberg/nessie/TestNessieTable.java
```diff
@@ -385,6 +388,89 @@ public void testDropTable() throws IOException {
...
+    ImmutableTableReference tableReference =
+        ImmutableTableReference.builder().reference("main").name(TABLE_NAME).build();
```
snazy, line 406: Please use a different branch here.
Using the default branch is not that great - and the test says `...Branch` implying it's not the default branch.

core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java
```diff
@@ -639,4 +642,29 @@ public void testConversions() {
     Assert.assertEquals(ns, JdbcUtil.stringToNamespace(nsString));
   }
 
+  @Test
```
snazy, line 645: Should these tests better live in `CatalogTests`?

aws/src/integration/java/org/apache/iceberg/aws/dynamodb/TestDynamoDbCatalog.java
```diff
@@ -295,6 +298,37 @@ public void testDropNamespace() {
     Assert.assertFalse("namespace must not exist", response.hasItem());
   }
 
+  @Test
+  public void testRegisterTable() {
```
snazy, line 302: This pair of tests is repeated (in a very similar way) across multiple catalogs. Can those be centralized somewhere? `CatalogTests` maybe?

ajantha-bhat:
@Mehul2500 : Please rebase this PR.

ajantha-bhat:
As there is no activity in this. 
I have opened https://github.com/apache/iceberg/pull/5492

PR #5300, opened 2022-07-18
Core: Add base implementations for changelog tasks

aokolnychyi:
This PR adds implementations for changelog tasks to be used in the changelog scan.
This PR currently relies on existing tests. More tests will come with the scan implementation.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
+++ b/core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
@@ -59,14 +59,13 @@ class BaseAddedRowsScanTask
     }
 
     @Override
-    public List<DeleteFile> deletes() {
-      return parentTask().deletes();
+    protected SplitAddedRowsScanTask copyWithNewLength(long newLength) {
+      return new SplitAddedRowsScanTask(parentTask(), start(), newLength);
     }
 
     @Override
-    public SplitAddedRowsScanTask merge(ScanTask other) {
-      SplitAddedRowsScanTask that = (SplitAddedRowsScanTask) other;
-      return new SplitAddedRowsScanTask(parentTask(), start(), length() + that.length());
+    public List<DeleteFile> deletes() {
+      return parentTask().deletes();
     }
   }
 }
--- a/core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
+++ b/core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
@@ -41,25 +41,20 @@ class BaseAddedRowsScanTask
   }
 
   @Override
+  protected AddedRowsScanTask newSplitTask(AddedRowsScanTask parentTask, long offset, long length) {
+    return new SplitAddedRowsScanTask(parentTask, offset, length);
+  }
+
+  @Override
   public List<DeleteFile> deletes() {
     return ImmutableList.copyOf(deletes);
   }
 
-  @Override
-  protected Iterable<AddedRowsScanTask> splitUsingOffsets(List<Long> offsets) {
-    return () -> new OffsetsAwareSplitScanTaskIteratorImpl(this, offsets);
-  }
-
-  @Override
-  protected Iterable<AddedRowsScanTask> splitUsingFixedSize(long targetSplitSize) {
-    return () -> new FixedSizeSplitScanTaskIteratorImpl(this, targetSplitSize);
-  }
-
-  private static class SplitScanTaskImpl
-      extends SplitScanTask<SplitScanTaskImpl, AddedRowsScanTask, DataFile>
+  private static class SplitAddedRowsScanTask
+      extends SplitScanTask<SplitAddedRowsScanTask, AddedRowsScanTask, DataFile>
       implements AddedRowsScanTask {
 
-    SplitScanTaskImpl(AddedRowsScanTask parentTask, long offset, long length) {
+    SplitAddedRowsScanTask(AddedRowsScanTask parentTask, long offset, long length) {
       super(parentTask, offset, length);
     }
 
@@ -69,35 +64,9 @@ class BaseAddedRowsScanTask
     }
 
     @Override
-    public SplitScanTaskImpl merge(ScanTask other) {
-      SplitScanTaskImpl that = (SplitScanTaskImpl) other;
-      return new SplitScanTaskImpl(parentTask(), start(), length() + that.length());
-    }
-  }
-
-  private static class OffsetsAwareSplitScanTaskIteratorImpl
-      extends OffsetsAwareSplitScanTaskIterator<AddedRowsScanTask> {
-
-    OffsetsAwareSplitScanTaskIteratorImpl(AddedRowsScanTask parentTask, List<Long> offsets) {
-      super(parentTask, parentTask.length(), offsets);
-    }
-
-    @Override
-    protected AddedRowsScanTask newSplitTask(AddedRowsScanTask parentTask, long offset, long length) {
-      return new SplitScanTaskImpl(parentTask, offset, length);
-    }
-  }
-
-  private static class FixedSizeSplitScanTaskIteratorImpl
-      extends FixedSizeSplitScanTaskIterator<AddedRowsScanTask> {
-
-    FixedSizeSplitScanTaskIteratorImpl(AddedRowsScanTask parentTask, long splitSize) {
-      super(parentTask, parentTask.length(), splitSize);
-    }
-
-    @Override
-    protected AddedRowsScanTask newSplitTask(AddedRowsScanTask parentTask, long offset, long length) {
-      return new SplitScanTaskImpl(parentTask, offset, length);
+    public SplitAddedRowsScanTask merge(ScanTask other) {
+      SplitAddedRowsScanTask that = (SplitAddedRowsScanTask) other;
+      return new SplitAddedRowsScanTask(parentTask(), start(), length() + that.length());
     }
   }
 }
--- a/core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
+++ b/core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.util.List;
+import org.apache.iceberg.expressions.ResidualEvaluator;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+
+class BaseAddedRowsScanTask
+    extends BaseChangelogContentScanTask<AddedRowsScanTask, DataFile>
+    implements AddedRowsScanTask {
+
+  private final DeleteFile[] deletes;
+
+  BaseAddedRowsScanTask(int changeOrdinal, long commitSnapshotId, DataFile file, DeleteFile[] deletes,
+                        String schemaString, String specString, ResidualEvaluator residuals) {
+    super(changeOrdinal, commitSnapshotId, file, schemaString, specString, residuals);
+    this.deletes = deletes != null ? deletes : new DeleteFile[0];
+  }
+
+  @Override
+  protected AddedRowsScanTask self() {
+    return this;
+  }
+
+  @Override
+  public List<DeleteFile> deletes() {
+    return ImmutableList.copyOf(deletes);
+  }
+
+  @Override
+  protected Iterable<AddedRowsScanTask> splitUsingOffsets(List<Long> offsets) {
+    return () -> new OffsetsAwareSplitScanTaskIteratorImpl(this, offsets);
+  }
+
+  @Override
+  protected Iterable<AddedRowsScanTask> splitUsingFixedSize(long targetSplitSize) {
+    return () -> new FixedSizeSplitScanTaskIteratorImpl(this, targetSplitSize);
+  }
+
+  private static class SplitScanTaskImpl
+      extends SplitScanTask<SplitScanTaskImpl, AddedRowsScanTask, DataFile>
+      implements AddedRowsScanTask {
+
+    SplitScanTaskImpl(AddedRowsScanTask parentTask, long offset, long length) {
+      super(parentTask, offset, length);
+    }
+
+    @Override
+    public List<DeleteFile> deletes() {
+      return parentTask().deletes();
+    }
+
+    @Override
+    public SplitScanTaskImpl merge(ScanTask other) {
+      SplitScanTaskImpl that = (SplitScanTaskImpl) other;
+      return new SplitScanTaskImpl(parentTask(), start(), length() + that.length());
+    }
+  }
+
+  private static class OffsetsAwareSplitScanTaskIteratorImpl
+      extends OffsetsAwareSplitScanTaskIterator<AddedRowsScanTask> {
+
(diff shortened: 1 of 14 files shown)


aokolnychyi:
@flyrain @rdblue @stevenzwu @karuppayya @szehon-ho @RussellSpitzer @anuragmantri

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
```diff
@@ -0,0 +1,140 @@
...
+  public long commitSnapshotId() {
+    return commitSnapshotId;
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("change_ordinal", changeOrdinal)
+        .add("commit_snapshot_id", commitSnapshotId)
+        .add("file", file().path())
```
szehon-ho, line 55: There seems to be some ways to have a method toStringHelper() at each level and use it for inheritance, do you think its worth it?  https://github.com/google/guava/issues/1239

core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public List<DeleteFile> deletes() {
+      return parentTask().deletes();
+    }
+
+    @Override
+    public SplitScanTaskImpl merge(ScanTask other) {
+      SplitScanTaskImpl that = (SplitScanTaskImpl) other;
+      return new SplitScanTaskImpl(parentTask(), start(), length() + that.length());
```
szehon-ho, line 74: Is it possible in  parent class to have abstract construction method:      

protected abstract ThisT newSplitScanTask(T parentTask, int start, int offset);

and then push merge() to parent, using this construction method?  Then we get merge() and canMerge() in the same class.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
```diff
@@ -0,0 +1,140 @@
...
+        .add("partition_data", file().partition())
+        .add("residual", residual())
+        .toString();
+  }
+
+  abstract static class SplitScanTask<ThisT, T extends ContentScanTask<F> & ChangelogScanTask, F extends ContentFile<F>>
```
szehon-ho, line 61: ThisT and T are a bit confusing, how about something like T => ParentT ?

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
```diff
@@ -0,0 +1,140 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.ResidualEvaluator;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+
+abstract class BaseChangelogContentScanTask<ThisT, F extends ContentFile<F>>
```
stevenzwu, line 26: nit: `ThisT` seems a little uncommon compared to just `T`

stevenzwu: does `ThisT` refers to this class of `BaseChangelogContentScanTask`?

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param <T> the Java type of tasks produced by this iterator
+ */
+abstract class OffsetsAwareSplitScanTaskIterator<T extends ScanTask> implements Iterator<T> {
+  private final T parentTask;
+  private final List<Long> offsets;
+  private final List<Long> splitSizes;
+  private int sizeIdx = 0;
+
+  OffsetsAwareSplitScanTaskIterator(T parentTask, long parentTaskLength, List<Long> offsetList) {
```
stevenzwu, line 39: do we need to validate?

1. offsetList is in strictly ascending order
2. last offset is smaller than the parentTaskLength

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * An iterator that splits tasks using split offsets such as row group offsets in Parquet.
+ *
+ * @param <T> the Java type of tasks produced by this iterator
+ */
+abstract class OffsetsAwareSplitScanTaskIterator<T extends ScanTask> implements Iterator<T> {
+  private final T parentTask;
+  private final List<Long> offsets;
+  private final List<Long> splitSizes;
+  private int sizeIdx = 0;
```
stevenzwu, line 37: nit: is `splitIndex` more accurate?

core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
stevenzwu: I didn't quite follow the merge semantics here. we just add the length but lose track of `other` in the returned object.

aokolnychyi: @szehon-ho, I am not sure that will be any cleaner as the only purpose of `merge` would be to directly call that new abstract method. I feel like it would pretty much the same.

@stevenzwu, that's the existing logic from `FileScanTask`. If two tasks are adjacent after split planning, we merge them together into one.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
aokolnychyi: The string representation would include the actual class name. I thought that would be sufficient. I don't think we include the file path or location of deletes files right now. The rest is the same across changelog tasks.

Locations may have some sensitive data. I'd probably keep it as-is for now to match the existing behavior. We can definitely reconsider the behavior in the future.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
aokolnychyi: I agree. I was trying to stay on one line as the list of params is long but let me try to find some reasonable formatting to use `ParentT`.

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
aokolnychyi: I copied this code as it was. From what I see, it looks like an index to the size of this iterator that is used to access both offsets and split sizes. Let me do some renames to make it a bit cleaner.

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
aokolnychyi: I am a bit worried to add/change the existing behavior in this class. I know we had a bug related to point 2 when the size of the file in metadata was incorrect. I guess handling point 1 may be a good idea. Let me think a bit more.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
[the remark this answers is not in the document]
aokolnychyi: @szehon-ho, instead of making the base class abstract and extending it everywhere, I can probably accept a closure to construct a split scan task in the base class. Then we won't need any subclasses. I think that's going to be cleaner. What do you think?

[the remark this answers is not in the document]
szehon-ho: That sounds great

core/src/main/java/org/apache/iceberg/BaseContentScanTask.java
```diff
@@ -0,0 +1,104 @@
...
+
+package org.apache.iceberg;
+
+import java.util.List;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.ResidualEvaluator;
+import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+
+abstract class BaseContentScanTask<ThisT, F extends ContentFile<F>>
```
flyrain, line 28: Nit: Does it make sense to use `T extends ScanTask` instead of `ThisT`? So that the type of `ThisT` is more obvious here.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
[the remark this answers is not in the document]
aokolnychyi: I managed to implement it. Could you take a look, @szehon-ho?

aokolnychyi: Changed.

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
aokolnychyi: I added validation for sorted offsets, which is required by the spec.

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
aokolnychyi: I changed the code in `next` a little bit so hopefully the purpose of `sizeIndex` is a bit clearer now.

core/src/main/java/org/apache/iceberg/BaseContentScanTask.java
aokolnychyi: We are using `ThisT` throughout the code to have access to the actual type of the object in a parent class. Usually, we don't annotate it with boundaries. However, I had to in this case. 

It became this:

```
ThisT extends ContentScanTask<F>
```

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
stevenzwu: >  it is an index of the size of this iterator that is used to access both offsets and split sizes lists

That is exactly where I found it a little confusing to be named as `sizeIndex`. This index is used to construct a split from the indexed offsets and sizes lists.

```
    long offset = offsets.get(sizeIndex);
    long splitSize = splitSizes.get(sizeIndex);
    sizeIndex += 1; // create 1 split per offset
    return createSplitTaskFunc.apply(parentTask, offset, splitSize);
```

But this is a very minor thing.

stevenzwu (approved):
This looks quite good to me

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
aokolnychyi: Yeah, I see what you mean now. I don't know what was the original interpretation but I can switch to `splitIndex`.

core/src/main/java/org/apache/iceberg/SplitScanTaskIterator.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * An iterator that splits tasks.
+ *
+ * @param <T> the Java type of tasks produced by this iterator
+ */
+interface SplitScanTaskIterator<T extends ScanTask> extends Iterator<T> {
+
+  @FunctionalInterface
+  interface CreateSplitTaskFunction<T extends ScanTask> {
```
flyrain, line 32: Minor naming suggestion:
```
 @FunctionalInterface
  interface SplitTaskCreator<T extends ScanTask> {
  T create(T parentTask, long offset, long length);
  }
```

flyrain (approved):
LGTM

core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
szehon-ho: Yea its not reducing a lot of code, just more symmetric for the reader.  The classes below (OffsetsAwareSplitScanTaskIteratorImpl and FixedSizeSplitScanTaskIteratorImpl) already have this 'newSplitTask' method, so we would just have one more just like that.

```
  private static class SplitScanTaskImpl
      extends SplitScanTask<SplitScanTaskImpl, AddedRowsScanTask, DataFile>
      implements AddedRowsScanTask {

    SplitScanTaskImpl(AddedRowsScanTask parentTask, long offset, long length) {
      super(parentTask, offset, length);
    }

    @Override
    public List<DeleteFile> deletes() {
      return parentTask().deletes();
    }

    @Override
    protected SplitScanTaskImpl newSplitTask(AddedRowsScanTask parentTask, long offset, long length) {
      return new SplitScanTaskImpl(parentTask, offset, length);
    }
  }
```

So the merge() is not just directly calling the newSplitTask but actually doing somewhat corresponding logic to canMerge(), which now is more clear in the same class.

```
  abstract static class SplitScanTask<ThisT, T extends ContentScanTask<F> & ChangelogScanTask, F extends ContentFile<F>>
      implements ContentScanTask<F>, ChangelogScanTask, MergeableScanTask<ThisT> {
   ...
       abstract ThisT newSplitTask(T parentTask, long offset, long length);
    
    @Override
    public boolean canMerge(ScanTask other) {
      if (getClass().equals(other.getClass())) {
        SplitScanTask<?, ?, ?> that = (SplitScanTask<?, ?, ?>) other;
        return changeOrdinal() == that.changeOrdinal() &&
            commitSnapshotId() == that.commitSnapshotId() &&
            file().equals(that.file()) &&
            start() + length() == that.start();

      } else {
        return false;
      }
    }

    @Override
    public ThisT merge(ScanTask other) {
      SplitScanTask that = (SplitScanTask) other;
      return newSplitTask(parentTask(), start(), length() + that.length());
    }
   ...
}
```

If we can somehow get rid of all these redundant classes with a lambda or something that would be the best, but probably its not possible.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
szehon-ho: Hm , maybe I might have been not clear, so just to clarify it wasn't suggesting to make toString in the child clases and logging other fields like location, but rather to reduce the code by having a method toStringHelper() from the base class (BaseContentScanTask) and re-use it (where we already have log file, partition_data, and residual) ?

aokolnychyi: Oh, okay, got it now. Let me explore this idea.

core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
aokolnychyi: Okay, I've changed the approach to have `merge` in the parent. Let me know if that's what you meant.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
aokolnychyi: I tried it locally. I couldn't control the order of fields in child classes and I couldn't benefit from it in split scan tasks as that one does not extend `BaseContentScanTask`. I'd just probably keep it as-is.

core/src/main/java/org/apache/iceberg/SplitScanTaskIterator.java
aokolnychyi: Missed it. That's a good idea. Let me change.

core/src/main/java/org/apache/iceberg/OffsetsAwareSplitScanTaskIterator.java
aokolnychyi: Switched.

core/src/main/java/org/apache/iceberg/SplitScanTaskIterator.java
aokolnychyi: Done.

core/src/main/java/org/apache/iceberg/BaseAddedRowsScanTask.java
szehon-ho: Yea, I didn't realize that you would remove the newSplitTask from the bottom two methods, leaving this one to be the only one.  But it's better to me.

core/src/main/java/org/apache/iceberg/BaseChangelogContentScanTask.java
szehon-ho: Sounds good.

szehon-ho (approved):

aokolnychyi:
Thanks for reviewing, @flyrain @stevenzwu @szehon-ho!

I merged this to unblock the scan implementation. If anyone has any additional comments, I'll be happy to follow up.

PR #5301, opened 2022-07-18
Core: Support building custom tasks in ManifestGroup

aokolnychyi:
This PR extends `ManifestGroup` with support of creating custom scan tasks.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/ManifestGroup.java
+++ b/core/src/main/java/org/apache/iceberg/ManifestGroup.java
@@ -156,6 +156,10 @@ class ManifestGroup {
    * @return a {@link CloseableIterable} of {@link FileScanTask}
    */
   public CloseableIterable<FileScanTask> planFiles() {
+    return planFiles(FileScanTaskFactory.builder());
+  }
+
+  public <T extends ScanTask> CloseableIterable<T> planFiles(ScanTaskFactory.Builder<T> taskFactoryBuilder) {
     LoadingCache<Integer, ResidualEvaluator> residualCache = Caffeine.newBuilder().build(specId -> {
       PartitionSpec spec = specsById.get(specId);
       Expression filter = ignoreResiduals ? Expressions.alwaysTrue() : dataFilter;
@@ -169,14 +173,18 @@ class ManifestGroup {
       select(ManifestReader.withStatsColumns(columns));
     }
 
-    Iterable<CloseableIterable<FileScanTask>> tasks = entries((manifest, entries) -> {
-      int specId = manifest.partitionSpecId();
+    LoadingCache<Integer, ScanTaskFactory<T>> taskFactoryCache = Caffeine.newBuilder().build(specId -> {
       PartitionSpec spec = specsById.get(specId);
-      String schemaString = SchemaParser.toJson(spec.schema());
-      String specString = PartitionSpecParser.toJson(spec);
       ResidualEvaluator residuals = residualCache.get(specId);
-      return CloseableIterable.transform(entries, e -> new BaseFileScanTask(
-          e.file().copy(!dropStats), deleteFiles.forEntry(e), schemaString, specString, residuals));
+      return taskFactoryBuilder
+          .dropStats(dropStats)
+          .build(spec, deleteFiles, residuals);
+    });
+
+    Iterable<CloseableIterable<T>> tasks = entries((manifest, entries) -> {
+      int specId = manifest.partitionSpecId();
+      ScanTaskFactory<T> taskFactory = taskFactoryCache.get(specId);
+      return taskFactory.createTasks(entries);
     });
 
     if (executorService != null) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -279,4 +287,84 @@ public void close() throws IOException {
...
+
+    ResidualEvaluator residuals() {
+      return residuals;
+    }
+
+    boolean shouldKeepStats() {
+      return !dropStats;
+    }
+
+    abstract static class Builder<T extends ScanTask> {
```
aokolnychyi, line 328: I am not super happy with having a builder as it adds more complexity. However, I did it this way so that we can have a loading cache of task factories per spec in `ManifestGroup`. Right now, we parse schema and spec JSON representations for each manifest, which is not required. As those JSON objects can get pretty large, I feel doing the parsing once per spec is an important optimization.

If we want to get rid of the builder, then I'll have to implement per spec caching in each task factory.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -279,4 +287,84 @@ public void close() throws IOException {
...
+      Builder<T> dropStats(boolean newDropStats) {
+        this.dropStats = newDropStats;
+        return this;
+      }
+
+      abstract ScanTaskFactory<T> build(PartitionSpec spec, DeleteFileIndex deletes, ResidualEvaluator residuals);
+    }
+  }
+
+  private static class FileScanTaskFactory extends ScanTaskFactory<FileScanTask> {
```
aokolnychyi, line 344: I'll have more such factories for changelog tasks.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -279,4 +287,84 @@ public void close() throws IOException {
...
+
+    ScanTaskFactory(PartitionSpec spec, DeleteFileIndex deletes, ResidualEvaluator residuals, boolean dropStats) {
+      this.schemaAsString = SchemaParser.toJson(spec.schema());
+      this.specAsString = PartitionSpecParser.toJson(spec);
+      this.deletes = deletes;
+      this.residuals = residuals;
+      this.dropStats = dropStats;
+    }
+
+    abstract CloseableIterable<T> createTasks(CloseableIterable<ManifestEntry<DataFile>> entries);
```
aokolnychyi, line 306: A task factory may create more than one task per entry. That's why I accept an iterable here.

aokolnychyi:
@flyrain @rdblue @stevenzwu @karuppayya @szehon-ho @RussellSpitzer @anuragmantri

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -279,4 +287,84 @@ public void close() throws IOException {
...
+    }
+
+    boolean shouldKeepStats() {
+      return !dropStats;
+    }
+
+    abstract static class Builder<T extends ScanTask> {
+      private boolean dropStats = false;
+
+      boolean dropStats() {
```
RussellSpitzer, line 331: Do we need to be able to return the dropStats value here? Seems like we can avoid this and not have two dropStats methods?

RussellSpitzer (approved):
I think this is a reasonable way of getting the change in, although the complexity of this class is very high already.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
flyrain: Can we just have an inheritable method like `ScanTaskFactory<T extends ScanTask> newInstance(PartitionSpec spec, DeleteFileIndex deletes, ResidualEvaluator residuals, boolean newDropStats)`?

flyrain (approved):
LGTM

core/src/main/java/org/apache/iceberg/ManifestGroup.java
Reo-LEI: +1 for yufei approach. Or we construct the cache for `schemaString` per spec and pass the `schemaString` from cache to `createTask(CloseableIterable<ManifestEntry<DataFile>> entries, String schemaString)`. But I prefer the former.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -279,4 +289,45 @@ public void close() throws IOException {
           }
         });
   }
+
+  @FunctionalInterface
+  interface CreateTasksFunction<T extends ScanTask> {
```
aokolnychyi, line 294: I could use `BiFunction` here but the definition would be bulky so I decided to add a functional interface just to shorten lines.

aokolnychyi:
@flyrain @Reo-LEI @RussellSpitzer, I've changed the approach to use a functional interface. I feel it is easier than before. Could you take a look?

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -156,6 +156,14 @@ ManifestGroup planWith(ExecutorService newExecutorService) {
    * @return a {@link CloseableIterable} of {@link FileScanTask}
    */
   public CloseableIterable<FileScanTask> planFiles() {
+    return plan((entries, ctx) -> CloseableIterable.transform(entries, entry -> {
+      DataFile dataFile = entry.file().copy(ctx.shouldKeepStats());
+      DeleteFile[] deleteFiles = ctx.deletes().forEntry(entry);
+      return new BaseFileScanTask(dataFile, deleteFiles, ctx.schemaAsString(), ctx.specAsString(), ctx.residuals());
+    }));
```
flyrain, line 163: Nit: I kind of prefer a normal function instead of an anonymous one for better readability since we are using the functionInterface here, which needs two steps to figure out the types of input parameters.  Or we just use `BiFunction`.

flyrain (commented):
+1 for the new approach

core/src/main/java/org/apache/iceberg/ManifestGroup.java
aokolnychyi: @flyrain, can you elaborate a bit? I am not sure I understood.
Do you mean creating an explicit `CreateFileScanTaskFunction` and using it here?

I did not use `BiFunction` to avoid a bulky definition.

```
public <T extends ScanTask> CloseableIterable<T> plan(BiFunction<CloseableIterable<ManifestEntry<DataFile>>, TaskContext, CloseableIterable<T>> createTasksFunc) {

...
}
```

core/src/main/java/org/apache/iceberg/ManifestGroup.java
aokolnychyi: No longer applies.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
aokolnychyi: No longer applies.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
flyrain: +1 for the new method `createFileScanTasks`. Thanks @aokolnychyi.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
Reo-LEI: I think it's more readable to use functional interface. +1 for this.

Reo-LEI:
+1 for this approach. LGTM.

aokolnychyi:
Thanks for reviewing, @RussellSpitzer @flyrain @Reo-LEI!

PR #5302, opened 2022-07-18
Add SparkV2Filters

huaxingao:
This PR adds `SparkV2Filters` and `TestSparkV2Filters`. In next PR, I will make `SparkScanBuilder` implement `SupportsPushDownV2Filters` and add a test `TestV2FilteredScan`. 

Currently, Spark still uses V2 Filter in `SupportsRuntimeFiltering` and `SupportsDelete`. I have a PR to add `SupportsRuntimeV2Filtering` and will have a PR to add V2 Filter in `SupportsDelete`  in Spark. Before these support are released on Spark side, we can use `Filter.toV2` to convert V1 filter to V2 filter for now.

Spark V2 Filter format:
```
Predicate(String name, Expression[] children)
```

The Predicate could be

```
name: IS_NULL
children[0] is NamedReference

name: IS_NOT_NULL
children[0] is NamedReference

name: STARTS_WITH
children[0] is NamedReference
children[1] is LiteralValue

name: IN
children[0] is NamedReference
children[1...n] is LiteralValue

name: =
children[0] is NamedReference
children[1] is LiteralValue

name: <>
children[0] is NamedReference
children[1] is LiteralValue

name: <=>
children[0] is NamedReference
children[1] is LiteralValue

name: <
children[0] is NamedReference
children[1] is LiteralValue

name: <=
children[0] is NamedReference
children[1] is LiteralValue

name: >
children[0] is NamedReference
children[1] is LiteralValue

name: >=
children[0] is NamedReference
children[1] is LiteralValue

name: AND
  public And(Predicate left, Predicate right) {
    super("AND", new Predicate[]{left, right});
  }

name: OR
  public Or(Predicate left, Predicate right) {
    super("OR", new Predicate[]{left, right});
  }

name: NOT
  public Not(Predicate child) {
    super("NOT", new Predicate[]{child});
  }

name: ALWAYS_TRUE
  public AlwaysTrue() {
    super("ALWAYS_TRUE", new Predicate[]{});
  }

name: ALWAYS_FALSE
  public AlwaysFalse() {
    super("ALWAYS_FALSE", new Predicate[]{});
  }
```

### DIFF
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
@@ -169,14 +169,13 @@ public class SparkV2Filters {
             Preconditions.checkNotNull(
                 value, "Expression is always false (eq is not null-safe): %s", predicate);
             return handleEqual(unquote(attributeName), value);
-          } else if (predicate.name().equals("<=>")) {
+          } else { // "<=>"
             if (value == null) {
               return isNull(unquote(attributeName));
             } else {
               return handleEqual(unquote(attributeName), value);
             }
           }
-          break;
 
         case IN:
           return in(
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
@@ -1,0 +1,356 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.util.NaNUtil;
+import org.apache.spark.sql.connector.expressions.LiteralValue;
+import org.apache.spark.sql.connector.expressions.NamedReference;
+import org.apache.spark.sql.connector.expressions.filter.And;
+import org.apache.spark.sql.connector.expressions.filter.Not;
+import org.apache.spark.sql.connector.expressions.filter.Or;
+import org.apache.spark.sql.connector.expressions.filter.Predicate;
+import org.apache.spark.unsafe.types.UTF8String;
+
+import static org.apache.iceberg.expressions.Expressions.and;
+import static org.apache.iceberg.expressions.Expressions.equal;
+import static org.apache.iceberg.expressions.Expressions.greaterThan;
+import static org.apache.iceberg.expressions.Expressions.greaterThanOrEqual;
+import static org.apache.iceberg.expressions.Expressions.in;
+import static org.apache.iceberg.expressions.Expressions.isNaN;
+import static org.apache.iceberg.expressions.Expressions.isNull;
+import static org.apache.iceberg.expressions.Expressions.lessThan;
+import static org.apache.iceberg.expressions.Expressions.lessThanOrEqual;
+import static org.apache.iceberg.expressions.Expressions.not;
+import static org.apache.iceberg.expressions.Expressions.notIn;
+import static org.apache.iceberg.expressions.Expressions.notNull;
+import static org.apache.iceberg.expressions.Expressions.or;
+import static org.apache.iceberg.expressions.Expressions.startsWith;
+
+public class SparkV2Filters {
+
+  private static final Pattern BACKTICKS_PATTERN = Pattern.compile("([`])(.|$)");
+
+  private SparkV2Filters() {
+  }
+
+  private static final Map<String, Expression.Operation> FILTERS = ImmutableMap
+      .<String, Expression.Operation>builder()
+      .put("ALWAYS_TRUE", Expression.Operation.TRUE)
+      .put("ALWAYS_FALSE", Expression.Operation.FALSE)
+      .put("=", Expression.Operation.EQ)
+      .put("<=>", Expression.Operation.EQ)
+      .put(">", Expression.Operation.GT)
+      .put(">=", Expression.Operation.GT_EQ)
+      .put("<", Expression.Operation.LT)
+      .put("<=", Expression.Operation.LT_EQ)
+      .put("IN", Expression.Operation.IN)
+      .put("IS_NULL", Expression.Operation.IS_NULL)
+      .put("IS_NOT_NULL", Expression.Operation.NOT_NULL)
+      .put("AND", Expression.Operation.AND)
+      .put("OR", Expression.Operation.OR)
+      .put("NOT", Expression.Operation.NOT)
+      .put("STARTS_WITH", Expression.Operation.STARTS_WITH)
+      .build();
+
+  public static Expression convert(Predicate[] predicates) {
+    Expression expression = Expressions.alwaysTrue();
+    for (Predicate predicate : predicates) {
+      Expression converted = convert(predicate);
+      Preconditions.checkArgument(converted != null, "Cannot convert predicate to Iceberg: %s", predicate);
+      expression = Expressions.and(expression, converted);
+    }
+    return expression;
+  }
+
+  @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:MethodLength"})
+  public static Expression convert(Predicate predicate) {
+    if (checkIfPredicateValid(predicate) == null) {
+      return null;
+    }
+
+    Expression.Operation op = FILTERS.get(predicate.name());
+    if (op != null) {
+      switch (op) {
+        case TRUE:
+          return Expressions.alwaysTrue();
+
+        case FALSE:
+          return Expressions.alwaysFalse();
+
+        case IS_NULL:
+          return isNull(unquote(getAttrName(predicate.children()[0])));
+
+        case NOT_NULL:
+          return notNull(unquote(getAttrName(predicate.children()[0])));
+
+        case LT:
+          if (predicate.children()[1] instanceof LiteralValue) {
+            return lessThan(unquote(getAttrName(predicate.children()[0])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+          } else {
+            return greaterThan(unquote(getAttrName(predicate.children()[1])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
+          }
+
+        case LT_EQ:
+          if (predicate.children()[1] instanceof LiteralValue) {
+            return lessThanOrEqual(unquote(getAttrName(predicate.children()[0])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+          } else {
+            return greaterThanOrEqual(unquote(getAttrName(predicate.children()[1])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
+          }
+
+        case GT:
+          if (predicate.children()[1] instanceof LiteralValue) {
+            return greaterThan(unquote(getAttrName(predicate.children()[0])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+          } else {
+            return lessThan(unquote(getAttrName(predicate.children()[1])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
+          }
+
+        case GT_EQ:
+          if (predicate.children()[1] instanceof LiteralValue) {
+            return greaterThanOrEqual(unquote(getAttrName(predicate.children()[0])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+          } else {
+            return lessThanOrEqual(unquote(getAttrName(predicate.children()[1])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
+          }
+
+        case EQ: // used for both eq and null-safe-eq
+          Object value = null;
+          String attributeName = "";
+          if (predicate.children()[1] instanceof LiteralValue) {
+            attributeName = getAttrName(predicate.children()[0]);
+            value = convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value());
+          } else {
+            attributeName = getAttrName(predicate.children()[1]);
+            value = convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value());
+          }
+          if (predicate.name().equals("=")) {
+            // comparison with null in normal equality is always null. this is probably a mistake.
+            Preconditions.checkNotNull(value,
+                "Expression is always false (eq is not null-safe): %s", predicate);
+            return handleEqual(unquote(attributeName), value);
+          } else if (predicate.name().equals("<=>")) {
+            if (value == null) {
+              return isNull(unquote(attributeName));
+            } else {
+              return handleEqual(unquote(attributeName), value);
+            }
+          }
+          break;
+
+        case IN:
+          Object[] inValues = new Object[predicate.children().length - 1];
+          for (int i = 1; i < predicate.children().length; i++) {
+            inValues[i - 1] = convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[i]).value());
+          }
+          return in(unquote(getAttrName(predicate.children()[0])),
+              Stream.of(inValues)
+                  .filter(Objects::nonNull)
+                  .collect(Collectors.toList()));
+
+        case NOT:
+          Not notFilter = (Not) predicate;
+          Predicate childFilter = notFilter.child();
+          if (childFilter.name().equals("IN")) {
+            // infer an extra notNull predicate for Spark NOT IN filters
+            // as Iceberg expressions don't follow the 3-value SQL boolean logic
+            // col NOT IN (1, 2) in Spark is equivalent to notNull(col) && notIn(col, 1, 2) in Iceberg
+            Object[] notInValues = new Object[childFilter.children().length - 1];
+            for (int i = 1; i < childFilter.children().length; i++) {
+              notInValues[i - 1] = convertUTF8StringIfNecessary(((LiteralValue) childFilter.children()[i]).value());
+            }
+            Expression notIn = notIn(unquote(getAttrName(childFilter.children()[0])),
+                Stream.of(notInValues).collect(Collectors.toList()));
+            return and(notNull(unquote(getAttrName(childFilter.children()[0]))), notIn);
+          } else if (hasNoInFilter(childFilter)) {
+            Expression child = convert(childFilter);
+            if (child != null) {
+              return not(child);
+            }
+          }
+          return null;
+
+        case AND: {
+          And andPredicate = (And) predicate;
+          Expression left = convert(andPredicate.left());
+          Expression right = convert(andPredicate.right());
+          if (left != null && right != null) {
+            return and(left, right);
+          }
+          return null;
+        }
+
+        case OR: {
+          Or orPredicate = (Or) predicate;
+          Expression left = convert(orPredicate.left());
+          Expression right = convert(orPredicate.right());
+          if (left != null && right != null) {
+            return or(left, right);
+          }
+          return null;
+        }
+
+        case STARTS_WITH: {
+          return startsWith(unquote(getAttrName(predicate.children()[0])),
+              convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()).toString());
+        }
+      }
+    }
+
+    return null;
+  }
+
+  private static Object convertUTF8StringIfNecessary(Object value) {
+    if (value instanceof UTF8String) {
+      return ((UTF8String) value).toString();
+    }
+    return value;
+  }
+
+  private static Expression handleEqual(String attribute, Object value) {
+    if (NaNUtil.isNaN(value)) {
+      return isNaN(attribute);
+    } else {
+      return equal(attribute, value);
+    }
+  }
+
+  private static String unquote(String attributeName) {
+    Matcher matcher = BACKTICKS_PATTERN.matcher(attributeName);
+    return matcher.replaceAll("$2");
+  }
+
+  private static boolean hasNoInFilter(Predicate predicate) {
+    Expression.Operation op = FILTERS.get(predicate.name());
+
+    if (op != null) {
+      switch (op) {
+        case AND:
+          And andPredicate = (And) predicate;
+          return hasNoInFilter(andPredicate.left()) && hasNoInFilter(andPredicate.right());
+        case OR:
+          Or orPredicate = (Or) predicate;
+          return hasNoInFilter(orPredicate.left()) && hasNoInFilter(orPredicate.right());
+        case NOT:
+          Not notPredicate = (Not) predicate;
+          return hasNoInFilter(notPredicate.child());
+        case IN:
+          return false;
+        default:
+          return true;
+      }
+    }
+
+    return false;
+  }
+
+  private static String getAttrName(org.apache.spark.sql.connector.expressions.Expression expr) {
+    return String.join(".", ((NamedReference) expr).fieldNames());
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
+  private static Predicate checkIfPredicateValid(Predicate predicate) {
+    Expression.Operation op = FILTERS.get(predicate.name());
+    if (op != null) {
+      switch (op) {
+        case IS_NULL:
+        case NOT_NULL:
+          if (!(predicate.children()[0] instanceof NamedReference)) {
+            return null;
+          }
+          return predicate;
+
+        case LT:
+        case LT_EQ:
+        case GT:
+        case GT_EQ:
+        case EQ:
+          if (!(((predicate.children()[1] instanceof LiteralValue &&
+                  predicate.children()[0] instanceof NamedReference)) ||
+              ((predicate.children()[0] instanceof LiteralValue &&
+                  predicate.children()[1] instanceof NamedReference)))) {
+            return null;
+          }
+          return predicate;
+
+        case IN:
+          if (!(predicate.children()[0] instanceof NamedReference)) {
+            return null;
+          } else {
+            for (int i = 1; i < predicate.children().length; i++) {
+              if (!(predicate.children()[i] instanceof LiteralValue)) {
+                return null;
+              }
+            }
+            return predicate;
+          }
+
+        case NOT:
+          Not notFilter = (Not) predicate;
+          return checkIfPredicateValid(notFilter.child());
+
+        case AND:
+          And andFilter = (And) predicate;
+          if (checkIfPredicateValid(andFilter.left()) == null || checkIfPredicateValid(andFilter.right()) == null) {
+            return null;
+          } else {
+            return predicate;
+          }
+
+        case OR:
+          Or orFilter = (Or) predicate;
+          if (checkIfPredicateValid(orFilter.left()) == null || checkIfPredicateValid(orFilter.right()) == null) {
+            return null;
+          } else {
+            return predicate;
+          }
+
+        case STARTS_WITH: {
+          if (!((predicate.children()[1] instanceof LiteralValue &&
+              predicate.children()[0] instanceof NamedReference))) {
+            return null;
+          }
+          return predicate;
+        }
+
+        case TRUE:
+        case FALSE:
+          return predicate;
+      }
+    }
+
+    return null;
+  }
+}
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
@@ -103,7 +103,6 @@ public class SparkV2Filters {
   @SuppressWarnings("unchecked")
   private static <T> T leftChild(Predicate predicate) {
     org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
-    Preconditions.checkArgument(children.length == 2, "%s should have two children", predicate);
     return (T) children[0];
   }
 
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
@@ -36,15 +36,15 @@ import static org.apache.iceberg.expressions.Expressions.startsWith;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Objects;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.iceberg.expressions.Expression;
+import org.apache.iceberg.expressions.Expression.Operation;
 import org.apache.iceberg.expressions.Expressions;
+import org.apache.iceberg.relocated.com.google.common.base.Joiner;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
 import org.apache.iceberg.util.NaNUtil;
-import org.apache.spark.sql.connector.expressions.LiteralValue;
+import org.apache.spark.sql.connector.expressions.Literal;
 import org.apache.spark.sql.connector.expressions.NamedReference;
 import org.apache.spark.sql.connector.expressions.filter.And;
 import org.apache.spark.sql.connector.expressions.filter.Not;
@@ -54,9 +54,7 @@ import org.apache.spark.unsafe.types.UTF8String;
 
 public class SparkV2Filters {
 
-  private static final Pattern BACKTICKS_PATTERN = Pattern.compile("([`])(.|$)");
-
-  private SparkV2Filters() {}
+  private static final Joiner DOT = Joiner.on(".");
 
   private static final String TRUE = "ALWAYS_TRUE";
   private static final String FALSE = "ALWAYS_FALSE";
@@ -74,37 +72,50 @@ public class SparkV2Filters {
   private static final String NOT = "NOT";
   private static final String STARTS_WITH = "STARTS_WITH";
 
-  private static final Map<String, Expression.Operation> FILTERS =
-      ImmutableMap.<String, Expression.Operation>builder()
-          .put(TRUE, Expression.Operation.TRUE)
-          .put(FALSE, Expression.Operation.FALSE)
-          .put(EQ, Expression.Operation.EQ)
-          .put(EQ_NULL_SAFE, Expression.Operation.EQ)
-          .put(GT, Expression.Operation.GT)
-          .put(GT_EQ, Expression.Operation.GT_EQ)
-          .put(LT, Expression.Operation.LT)
-          .put(LT_EQ, Expression.Operation.LT_EQ)
-          .put(IN, Expression.Operation.IN)
-          .put(IS_NULL, Expression.Operation.IS_NULL)
-          .put(NOT_NULL, Expression.Operation.NOT_NULL)
-          .put(AND, Expression.Operation.AND)
-          .put(OR, Expression.Operation.OR)
-          .put(NOT, Expression.Operation.NOT)
-          .put(STARTS_WITH, Expression.Operation.STARTS_WITH)
+  private static final Map<String, Operation> FILTERS =
+      ImmutableMap.<String, Operation>builder()
+          .put(TRUE, Operation.TRUE)
+          .put(FALSE, Operation.FALSE)
+          .put(EQ, Operation.EQ)
+          .put(EQ_NULL_SAFE, Operation.EQ)
+          .put(GT, Operation.GT)
+          .put(GT_EQ, Operation.GT_EQ)
+          .put(LT, Operation.LT)
+          .put(LT_EQ, Operation.LT_EQ)
+          .put(IN, Operation.IN)
+          .put(IS_NULL, Operation.IS_NULL)
+          .put(NOT_NULL, Operation.NOT_NULL)
+          .put(AND, Operation.AND)
+          .put(OR, Operation.OR)
+          .put(NOT, Operation.NOT)
+          .put(STARTS_WITH, Operation.STARTS_WITH)
           .build();
 
-  private static final int FIRST_ORDINAL = 0;
-  private static final int SECOND_ORDINAL = 1;
+  private SparkV2Filters() {}
 
-  public static Expression convert(Predicate[] predicates) {
-    Expression expression = Expressions.alwaysTrue();
-    for (Predicate predicate : predicates) {
-      Expression converted = convert(predicate);
-      Preconditions.checkArgument(
-          converted != null, "Cannot convert predicate to Iceberg: %s", predicate);
-      expression = Expressions.and(expression, converted);
-    }
-    return expression;
+  @SuppressWarnings("unchecked")
+  private static <T> T child(Predicate predicate) {
+    org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
+    Preconditions.checkArgument(children.length == 1, "%s should have one child", predicate);
+    return (T) children[0];
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <T> T leftChild(Predicate predicate) {
+    org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
+    Preconditions.checkArgument(children.length == 2, "%s should have two children", predicate);
+    return (T) children[0];
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <T> T rightChild(Predicate predicate) {
+    org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
+    Preconditions.checkArgument(children.length == 2, "%s should have two children", predicate);
+    return (T) children[1];
+  }
+
+  private static String toColumnName(NamedReference ref) {
+    return DOT.join(ref.fieldNames());
   }
 
   @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:MethodLength"})
@@ -113,7 +124,7 @@ public class SparkV2Filters {
       return null;
     }
 
-    Expression.Operation op = FILTERS.get(predicate.name());
+    Operation op = FILTERS.get(predicate.name());
     if (op != null) {
       switch (op) {
         case TRUE:
@@ -123,92 +134,85 @@ public class SparkV2Filters {
           return Expressions.alwaysFalse();
 
         case IS_NULL:
-          return isNull(unquote(predicate.children()[FIRST_ORDINAL].toString()));
+          return isNull(toColumnName(child(predicate)));
 
         case NOT_NULL:
-          return notNull(unquote(predicate.children()[FIRST_ORDINAL].toString()));
+          return notNull(toColumnName(child(predicate)));
 
         case LT:
-          if (predicate.children()[SECOND_ORDINAL] instanceof LiteralValue) {
+          if (rightChild(predicate) instanceof Literal) {
             return lessThan(
-                unquote(predicate.children()[FIRST_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+                toColumnName(leftChild(predicate)),
+                convertUTF8StringIfNecessary(rightChild(predicate)));
           } else {
             return greaterThan(
-                unquote(predicate.children()[SECOND_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(
-                    ((LiteralValue) predicate.children()[FIRST_ORDINAL]).value()));
+                toColumnName(rightChild(predicate)),
+                convertUTF8StringIfNecessary(leftChild(predicate)));
           }
 
         case LT_EQ:
-          if (predicate.children()[SECOND_ORDINAL] instanceof LiteralValue) {
+          if (rightChild(predicate) instanceof Literal) {
             return lessThanOrEqual(
-                unquote(predicate.children()[FIRST_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+                toColumnName(leftChild(predicate)),
+                convertUTF8StringIfNecessary(rightChild(predicate)));
           } else {
             return greaterThanOrEqual(
-                unquote(predicate.children()[SECOND_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(
-                    ((LiteralValue) predicate.children()[FIRST_ORDINAL]).value()));
+                toColumnName(rightChild(predicate)),
+                convertUTF8StringIfNecessary(leftChild(predicate)));
           }
 
         case GT:
-          if (predicate.children()[SECOND_ORDINAL] instanceof LiteralValue) {
+          if (rightChild(predicate) instanceof Literal) {
             return greaterThan(
-                unquote(predicate.children()[FIRST_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(
-                    ((LiteralValue) predicate.children()[SECOND_ORDINAL]).value()));
+                toColumnName(leftChild(predicate)),
+                convertUTF8StringIfNecessary(rightChild(predicate)));
           } else {
             return lessThan(
-                unquote(predicate.children()[SECOND_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(
-                    ((LiteralValue) predicate.children()[FIRST_ORDINAL]).value()));
+                toColumnName(rightChild(predicate)),
+                convertUTF8StringIfNecessary(leftChild(predicate)));
           }
 
         case GT_EQ:
-          if (predicate.children()[SECOND_ORDINAL] instanceof LiteralValue) {
+          if (rightChild(predicate) instanceof Literal) {
             return greaterThanOrEqual(
-                unquote(predicate.children()[FIRST_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+                toColumnName(leftChild(predicate)),
+                convertUTF8StringIfNecessary(rightChild(predicate)));
           } else {
             return lessThanOrEqual(
-                unquote(predicate.children()[SECOND_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(
-                    ((LiteralValue) predicate.children()[FIRST_ORDINAL]).value()));
+                toColumnName(rightChild(predicate)),
+                convertUTF8StringIfNecessary(leftChild(predicate)));
           }
 
         case EQ: // used for both eq and null-safe-eq
           Object value;
           String attributeName;
-          if (predicate.children()[SECOND_ORDINAL] instanceof LiteralValue) {
-            attributeName = predicate.children()[FIRST_ORDINAL].toString();
-            value = convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value());
+          if (rightChild(predicate) instanceof Literal) {
+            attributeName = toColumnName(leftChild(predicate));
+            value = convertUTF8StringIfNecessary(rightChild(predicate));
           } else {
-            attributeName = predicate.children()[SECOND_ORDINAL].toString();
-            value =
-                convertUTF8StringIfNecessary(
-                    ((LiteralValue) predicate.children()[FIRST_ORDINAL]).value());
+            attributeName = toColumnName(rightChild(predicate));
+            value = convertUTF8StringIfNecessary(leftChild(predicate));
           }
 
           if (predicate.name().equals("=")) {
             // comparison with null in normal equality is always null. this is probably a mistake.
             Preconditions.checkNotNull(
                 value, "Expression is always false (eq is not null-safe): %s", predicate);
-            return handleEqual(unquote(attributeName), value);
+            return handleEqual(attributeName, value);
           } else { // "<=>"
             if (value == null) {
-              return isNull(unquote(attributeName));
+              return isNull(attributeName);
             } else {
-              return handleEqual(unquote(attributeName), value);
+              return handleEqual(attributeName, value);
             }
           }
 
         case IN:
           return in(
-              unquote(predicate.children()[FIRST_ORDINAL].toString()),
+              toColumnName(leftChild(predicate)),
               Arrays.stream(predicate.children())
                   .skip(1)
-                  .map(val -> convertUTF8StringIfNecessary(((LiteralValue) val).value()))
+                  .map(val -> convertUTF8StringIfNecessary(((Literal) val)))
                   .filter(Objects::nonNull)
                   .collect(Collectors.toList()));
 
@@ -222,13 +226,13 @@ public class SparkV2Filters {
             // Iceberg
             Expression notIn =
                 notIn(
-                    unquote(childFilter.children()[FIRST_ORDINAL].toString()),
+                    toColumnName(leftChild(childFilter)),
                     Arrays.stream(childFilter.children())
                         .skip(1)
-                        .map(val -> convertUTF8StringIfNecessary(((LiteralValue) val).value()))
+                        .map(val -> convertUTF8StringIfNecessary(((Literal) val)))
                         .filter(Objects::nonNull)
                         .collect(Collectors.toList()));
-            return and(notNull(unquote(childFilter.children()[FIRST_ORDINAL].toString())), notIn);
+            return and(notNull(toColumnName(leftChild(childFilter))), notIn);
           } else if (hasNoInFilter(childFilter)) {
             Expression child = convert(childFilter);
             if (child != null) {
@@ -260,24 +264,20 @@ public class SparkV2Filters {
           }
 
         case STARTS_WITH:
-          {
-            return startsWith(
-                unquote(predicate.children()[FIRST_ORDINAL].toString()),
-                convertUTF8StringIfNecessary(
-                        ((LiteralValue) predicate.children()[SECOND_ORDINAL]).value())
-                    .toString());
-          }
+          return startsWith(
+              toColumnName(leftChild(predicate)),
+              convertUTF8StringIfNecessary(rightChild(predicate)).toString());
       }
     }
 
     return null;
   }
 
-  private static Object convertUTF8StringIfNecessary(Object value) {
-    if (value instanceof UTF8String) {
-      return ((UTF8String) value).toString();
+  private static Object convertUTF8StringIfNecessary(Literal literal) {
+    if (literal.value() instanceof UTF8String) {
+      return ((UTF8String) literal.value()).toString();
     }
-    return value;
+    return literal.value();
   }
 
   private static Expression handleEqual(String attribute, Object value) {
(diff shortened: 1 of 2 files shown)


spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,356 @@
...
+
+    return false;
+  }
+
+  private static String getAttrName(org.apache.spark.sql.connector.expressions.Expression expr) {
+    return String.join(".", ((NamedReference) expr).fieldNames());
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
+  private static Predicate checkIfPredicateValid(Predicate predicate) {
```
huaxingao, line 284: I think Spark should have this check to make sure all the V2 Filters are in valid format. Otherwise, this kind of invalid [filter](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark-extensions/src/test/java/org/apache/iceberg/spark/extensions/TestDelete.java#L433) will get though. 

I used to have this check on Spark side, but somehow it got removed during refactor. I will add this check back. Before this is available, I will have this check here. I will remove once this is in Spark.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,356 @@
...
+        case NOT_NULL:
+          return notNull(unquote(getAttrName(predicate.children()[0])));
+
+        case LT:
+          if (predicate.children()[1] instanceof LiteralValue) {
+            return lessThan(unquote(getAttrName(predicate.children()[0])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+          } else {
+            return greaterThan(unquote(getAttrName(predicate.children()[1])),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
```
huaxingao, line 119: I think this kind of normalization (e.g. `1 < a => a > 1`) should be done in Spark. I used to have this, but I just found out it was removed. I will add this back in Spark. But before this is available in Spark, I will have it here.

huaxingao:
cc @rdblue @aokolnychyi @RussellSpitzer 
Could you please take a look when you have time? Thanks!

huaxingao:
gentle ping @rdblue @aokolnychyi @RussellSpitzer

huaxingao:
also cc @flyrain @szehon-ho @kbendick

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+import org.apache.spark.sql.connector.expressions.filter.Predicate;
+import org.apache.spark.unsafe.types.UTF8String;
+
+public class SparkV2Filters {
+
+  private static final Pattern BACKTICKS_PATTERN = Pattern.compile("([`])(.|$)");
+
+  private SparkV2Filters() {}
+
+  private static final Map<String, Expression.Operation> FILTERS =
```
RussellSpitzer, line 61: I find it kind of gross that Spark is making us deal with this, seems like this should be generated rather then us having to literally write out all the strings. Not that we have anything we can improve here. Still kind of hoping we can have some Substrait.io style solution for this kind of thing in the future.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+          converted != null, "Cannot convert predicate to Iceberg: %s", predicate);
+      expression = Expressions.and(expression, converted);
+    }
+    return expression;
+  }
+
+  @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:MethodLength"})
+  public static Expression convert(Predicate predicate) {
+    if (checkIfPredicateValid(predicate) == null) {
+      return null;
```
RussellSpitzer, line 94: I'm a little confused why we use the return value here to signal error. The Code above iterates through the predicates one at a time so we don't collect all of the "non-convertable" predicates, I could see if we wanted to generate a large error message with all the filters we couldn't convert but it seems like we could just assert here rather than in the parent "convert" method?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: I think the main reason that Spark uses a String for the V2 Filter name is because Spark introduces [`GeneralScalarExpression`](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/java/org/apache/spark/sql/connector/expressions/GeneralScalarExpression.java) in 3.3 and pushes down a lot of functions such as `+`, `-`, `*`, `/`, `ABS`, `LOG`, ... Spark doesn't want to have a class for each of these Expressions so it uses a more genetic way. V2 [`Predicate`](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/java/org/apache/spark/sql/connector/expressions/filter/Predicate.java) is a subclass of `GeneralScalarExpression` so it also uses the field `name` for Filter name. 

I am actually wondering if any of these function push down can benefit Iceberg. It seems to me that for both the data predicate push down and partition pruning, Iceberg only supports predicate `attri cmp lit`, so we can't really take advantage of any of these new function push down in Spark. In Spark built-in file source such as Parquet, partition pruning uses Catalyst Expression so Spark can prune partition such as `partCol1 + partCol2 > 3`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: The idea is that if the Spark V2 filter is not valid, then we simply return a `null` and skip the convert code, then later on we don't need to bind and push down this predicate (https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java#L120)

RussellSpitzer: But we throw an exception here? Shouldn't the behavior just be to ignore the predicate then?
https://github.com/apache/iceberg/pull/5302/files#diff-d21a6d1f3fb11bbc8d5e9ced54b7fd025e04a70a54c8263ae49313ea3cb274beR84-R85

huaxingao: I see what you mean now. Sorry I wasn't realize that you were talking about the convert method that takes the `Predicate` array.

This method `convert(Predicate predicate)` will be used by [`SparkScanBuilder`.`pushPredicates`](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java#L111) directly. If the Spark v2 filter can't be converted to Iceberg Expression, we just ignore the filter and not push it down.

The method `convert(Predicate[] predicates)` will be used by [`SparkWriteBuilder`.`overwrite`](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkWriteBuilder.java#L122) and [`SparkTable`.`deleteWhere`](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java#L311). I follow the same logic as [`SparkFilter.convert(Filter[] filters)`](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java#L311). By definition, `overwrite` can only replace data iff all the filters match and `deleteWhere` can only delete rows iff all the filter expressions match. That's why we throw Exception in `convert(Predicate[] predicates)` if any of the filter can't be converted to Iceberg Expression.

huaxingao: I think this over, I probably should call `convert(Predicate predicate)` one by one in [SparkWriteBuilder.overwrite](https://github.com/apache/iceberg/blob/master/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkWriteBuilder.java#L122). If I get null, then throw Exception saying `Cannot overwrite files using filters because not all the filter[s] can be converted correctly`. This is probably better than the error message `Cannot convert predicate to Iceberg`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+        default:
+          return true;
+      }
+    }
+
+    return false;
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
+  private static Predicate checkIfPredicateValid(Predicate predicate) {
```
szehon-ho, line 297: I think we usually prefer shorter method names, so maybe valid(Predicate predicate) is better?  Also, kind of related to the previous comment, why do we have the return value be a boolean instead of predicate?  (valid seems to be a true/false question)

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+            return null;
+          }
+          return predicate;
+
+        case LT:
+        case LT_EQ:
+        case GT:
+        case GT_EQ:
+        case EQ:
+          if (predicate.children().length != 2
```
szehon-ho, line 313: Just a personal preference but maybe its easier to read (if we go with the boolean return value approach):
```
if (predicate.children.length() != 2) {
  return false;
}
return (predicate.children[0] instanceof NamedReference && predicate.children[1] instanceof LiteralReference) ||
(predicate.children[0] instanceof LiteralRefernce && predicate.children[1] instanceof NamedReference]);
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+                      && predicate.children()[1] instanceof NamedReference)))) {
+            return null;
+          }
+          return predicate;
+
+        case IN:
+          if (!(predicate.children()[0] instanceof NamedReference)) {
+            return null;
+          } else {
+            for (int i = 1; i < predicate.children().length; i++) {
```
szehon-ho, line 326: Also optional style suggestion, can use predicate.children.stream().allMatch()

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+            return predicate;
+          }
+
+        case NOT:
+          Not notFilter = (Not) predicate;
+          return checkIfPredicateValid(notFilter.child());
+
+        case AND:
+          And andFilter = (And) predicate;
+          if (checkIfPredicateValid(andFilter.left()) == null
```
szehon-ho, line 340: Same if we change, we could have an easier to read:
return valid(andFilter.left) && valid(andFilter.right);

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+          return isNull(unquote(predicate.children()[0].toString()));
+
+        case NOT_NULL:
+          return notNull(unquote(predicate.children()[0].toString()));
+
+        case LT:
+          if (predicate.children()[1] instanceof LiteralValue) {
+            return lessThan(
+                unquote(predicate.children()[0].toString()),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
```
szehon-ho, line 116: I'm not very familiar but In the old SparkFilters code there used to be a 'convertLiteral' functionality.  Do we need that for literals?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+          } else {
+            return lessThanOrEqual(
+                unquote(predicate.children()[1].toString()),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
+          }
+
+        case EQ: // used for both eq and null-safe-eq
+          Object value = null;
```
szehon-ho, line 157: Nit: looks like we dont need null and "" (java defaults)

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+              return isNull(unquote(attributeName));
+            } else {
+              return handleEqual(unquote(attributeName), value);
+            }
+          }
+          break;
+
+        case IN:
+          Object[] inValues = new Object[predicate.children().length - 1];
+          for (int i = 1; i < predicate.children().length; i++) {
```
szehon-ho, line 183: Why not just make this a stream (as the next line anyway converts it to stream).  Ie, predicate.children.stream().map(c -> convertUTF8StringIfNecessary(...))

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,375 @@
...
+        case NOT:
+          Not notFilter = (Not) predicate;
+          Predicate childFilter = notFilter.child();
+          if (childFilter.name().equals("IN")) {
+            // infer an extra notNull predicate for Spark NOT IN filters
+            // as Iceberg expressions don't follow the 3-value SQL boolean logic
+            // col NOT IN (1, 2) in Spark is equivalent to notNull(col) && notIn(col, 1, 2) in
+            // Iceberg
+            Object[] notInValues = new Object[childFilter.children().length - 1];
+            for (int i = 1; i < childFilter.children().length; i++) {
```
szehon-ho, line 200: Same comment (use a stream)

szehon-ho (commented):
Hi @huaxingao I guess most of the code is direct from SparkFilters.java, I did an early code pass.

It would be really nice if Spark did the normalization like in the previous version (the order of comparison operations), so the code mostly stays the same as before.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: We don't need this `convertLiteral` any more. This is the conversion that V2 filter is trying to avoid.

huaxingao:
@szehon-ho Thank you very much for taking time to review. I have addressed the comments. Could you please take another look when you are free? Thanks a lot!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,353 @@
...
+                value, "Expression is always false (eq is not null-safe): %s", predicate);
+            return handleEqual(unquote(attributeName), value);
+          } else if (predicate.name().equals("<=>")) {
+            if (value == null) {
+              return isNull(unquote(attributeName));
+            } else {
+              return handleEqual(unquote(attributeName), value);
+            }
+          }
+          break;
```
szehon-ho, line 179: Nit: maybe we should return null explicitly like the other case instead of break.  Also curious, will this ever happen?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.spark.sql.connector.expressions.filter.Not;
+import org.apache.spark.sql.connector.expressions.filter.Or;
+import org.apache.spark.sql.connector.expressions.filter.Predicate;
+import org.apache.spark.sql.types.DateType$;
+import org.apache.spark.sql.types.IntegerType$;
+import org.apache.spark.sql.types.TimestampType$;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestSparkV2Filters {
```
szehon-ho, line 41: Do we need some test of the cases of the expression the other way ( literal, attribute)

szehon-ho (commented):
Looks better to me , thanks @huaxingao .  Just curious, any chance we can avoid having the normalization code in Iceberg, or will it take too long as we need to wait for a Spark pr and release?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: This will not happen. I removed this.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
huaxingao: Added some tests. Thanks!

huaxingao:
> any chance we can avoid having the normalization code in Iceberg, or will it take too long as we need to wait for a Spark pr and release?

I have already added the normalization code in Spark, but we need to wait for 3.4 release.

szehon-ho (approved):
Cool, looks good to me.  May wait a little to see if anyone else wants to take a look, maybe @aokolnychyi

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
kbendick: Agreed. But I see why it's done.

The V2FunctionCatalog works the same way. Notice the `bucket` v2 function transform opt works the same way. With a string typed `numCols` and a string typed `bucket` function that Spark searches for on the implementing library's side. The name technically doesn't matter for the field but it 100% does for the function.

https://github.com/apache/spark/blob/126870eddc57fff7408942897dd59a2c19c91c4a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/V2ExpressionUtils.scala#L90-L100

I hope @huaxingao is right though that we will be able to use Spark to parse more expressions this way though =)

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,352 @@
...
+                unquote(predicate.children()[0].toString()),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[1]).value()));
+          } else {
+            return greaterThan(
+                unquote(predicate.children()[1].toString()),
+                convertUTF8StringIfNecessary(((LiteralValue) predicate.children()[0]).value()));
+          }
+
+        case LT_EQ:
+          if (predicate.children()[1] instanceof LiteralValue) {
```
kbendick, line 124: Nit / non-blocking style question: Would it make sense to use named constant for the ordinals? They're used quite a lot.

Maybe the following?
```
private static int FIRST_PREDICATE_CHILD_ORDINAL = 0;
private static int SECOND_PREDICATE_CHILD_ORGINAL = 1;
```

Not tied to the naming but Anton suggested I do similar in the `truncate` v2 function PR and it turned out nicely. Might be overkill since they don't correspond to much besides first and second though. 

https://github.com/apache/iceberg/blob/b3062e443848c9badd7332c4b26c14f2b257cc4a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java#L56-L57

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  private static Object convertUTF8StringIfNecessary(Object value) {
+    if (value instanceof UTF8String) {
+      return ((UTF8String) value).toString();
+    }
+    return value;
+  }
+
+  private static Expression handleEqual(String attribute, Object value) {
```
kbendick, line 256: Maybe just `NaNSafeEquals` or `SafeEquals`. This is fine as well.

kbendick: Edit - Nevermind this is lifted from the original SparkFilters. Leaving it as is is the best choicee imo.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,352 @@
...
+        case LT:
+        case LT_EQ:
+        case GT:
+        case GT_EQ:
+        case EQ:
+          if (predicate.children().length != 2) {
+            return false;
+          }
+          return (predicate.children()[0] instanceof NamedReference
+                  && predicate.children()[1] instanceof LiteralValue)
```
kbendick, line 311: This pattern comes up enough (and is somewhat confusing to read).

Maybe `binaryNamedReferenceComparedToLiteral` and `binaryLiterComparedToReference`? Though I guess if Spark is moving towards inlining with strings etc, it might be better to not try to deviate from that at first.

We can always refactor so I'd save something like this for a follow up PR (if at all). We'd probably wind up building up _a lot_ of functions.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
```diff
@@ -0,0 +1,254 @@
...
+    Assert.assertEquals(
+        "Generated date expression should be correct",
+        rawExpression.toString(),
+        dateExpression.toString());
+  }
+
+  @Test
+  public void testNestedInInsideNot() {
+    NamedReference namedReference1 = FieldReference.apply("col1");
+    LiteralValue v1 = new LiteralValue(1, IntegerType$.MODULE$);
```
kbendick, line 222: Nit: I think `IntegerType$.Module$` is equivalent to `DataTypes.IntegerType`, is it not? They might be one dollar sign apart 🙃  but if they're equivalent we should use that.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
kbendick: Nit / non-blocking / for follow up if desired: We might break these up into two or three sets of tests.

I can see:
1. Literal to Attribute
2. Attribute to Literal
3. Invalid assertions.

This could totally be done in a follow up PR, but because the stringly typed nature of these areas is somewhat error prone in my opinion I think we should try to make the tests extensible. But not important for this PR imo (which has been around long enough and would be good to be using).

kbendick (approved):
Overall this looks good to me.

There are some nits that I would maybe change style wise, but I don't think it's worth holding up the PR necessarily. It's also just my opinion and not required.

Also, this should eventually cover at least the same tests as the V1 filters, but if that's already been met than that's fine.

Thanks @huaxingao! And thanks for your patience on the reviews!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
kbendick: Again, this might be overkill as they don't correspond much to anything meaningful (like the ones above in `TruncateFunction` do).

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
huaxingao: I will refactor the tests in a follow up. Thanks for the suggestion!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,352 @@
...
+          .put(">=", Expression.Operation.GT_EQ)
+          .put("<", Expression.Operation.LT)
+          .put("<=", Expression.Operation.LT_EQ)
+          .put("IN", Expression.Operation.IN)
+          .put("IS_NULL", Expression.Operation.IS_NULL)
+          .put("IS_NOT_NULL", Expression.Operation.NOT_NULL)
+          .put("AND", Expression.Operation.AND)
+          .put("OR", Expression.Operation.OR)
+          .put("NOT", Expression.Operation.NOT)
+          .put("STARTS_WITH", Expression.Operation.STARTS_WITH)
```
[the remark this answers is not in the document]
huaxingao: I have added named constants for the predicates.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed, but made the names a little shorter.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed, but also made the names a little shorter.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
huaxingao: Fixed. Thanks

huaxingao:
Thanks a lot @kbendick for taking time reviewing the PR! I have addressed the comments. Could you please take one more look when you are free? Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+import org.apache.spark.sql.connector.expressions.filter.Not;
+import org.apache.spark.sql.connector.expressions.filter.Or;
+import org.apache.spark.sql.connector.expressions.filter.Predicate;
+import org.apache.spark.unsafe.types.UTF8String;
+
+public class SparkV2Filters {
+
+  private static final Pattern BACKTICKS_PATTERN = Pattern.compile("([`])(.|$)");
+
+  private SparkV2Filters() {}
```
kbendick, line 59: Nit: Maybe move the constructor before or after all of the private static constants?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+            attributeName = predicate.children()[SECOND_ORDINAL].toString();
+            value =
+                convertUTF8StringIfNecessary(
+                    ((LiteralValue) predicate.children()[FIRST_ORDINAL]).value());
+          }
+
+          if (predicate.name().equals("=")) {
+            // comparison with null in normal equality is always null. this is probably a mistake.
+            Preconditions.checkNotNull(
+                value, "Expression is always false (eq is not null-safe): %s", predicate);
```
kbendick, line 196: Question: will this fail the query?

So if somebody writes `select a, b from foo where c = null`, the query will throw? Would we be better off logging a warning and then using `Expressions.alwaysFalse()` (or null), or is throwing more user friendly since probably they didn’t mean for that to be the query — the query basically will not match anything right?

Genuinely asking as I’m not sure if people use this pattern at all. You’re right it seems like a mistake.

Maybe we should add a unit test that this throws the specified exception?

I’m genuinely not sure which is best so just asking. I’m good with it as is.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+          .put(IS_NULL, Expression.Operation.IS_NULL)
+          .put(NOT_NULL, Expression.Operation.NOT_NULL)
+          .put(AND, Expression.Operation.AND)
+          .put(OR, Expression.Operation.OR)
+          .put(NOT, Expression.Operation.NOT)
+          .put(STARTS_WITH, Expression.Operation.STARTS_WITH)
+          .build();
+
+  private static final int FIRST_ORDINAL = 0;
+  private static final int SECOND_ORDINAL = 1;
```
kbendick, line 97: I know I suggested first_ordinal and second_ordinal, but if you’re not in love with those, after looking at it further I think “left child” and “right child” might make sense too.

But I also don’t see any need to bike shed / change further just for the sake of changing. This is plenty readable as-is for me. Just throwing it out there as another idea.

kbendick (approved):
This looks good to me. Left a few comments, but I don’t think they need to be addressed — just last minute other options I thought of in case you decide you don’t want to use something as it currently is.

But this looks good to me. I can see why Spark doesn’t want so many classes but it does add a little more burden on us. I think you did a really good job with it considering. Thanks @huaxingao!

aokolnychyi:
I'd like to take a look tomorrow or on Friday as well.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+      Preconditions.checkArgument(
+          converted != null, "Cannot convert predicate to Iceberg: %s", predicate);
+      expression = Expressions.and(expression, converted);
+    }
+    return expression;
+  }
+
+  @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:MethodLength"})
+  public static Expression convert(Predicate predicate) {
+    if (!valid(predicate)) {
```
aokolnychyi, line 112: @huaxingao, can you elaborate a bit on what is considered invalid in this case? Are we trying to validate the Spark logic or are we trying to check if the filter is supported by Iceberg?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+          } else if (hasNoInFilter(childFilter)) {
+            Expression child = convert(childFilter);
+            if (child != null) {
+              return not(child);
+            }
+          }
+          return null;
+
+        case AND:
+          {
```
aokolnychyi, line 241: nit: Do we need these extra `{}` here and 2 branches below?

aokolnychyi (commented):
This is a great effort. Thanks for doing this, @huaxingao!

I think we need to remove some of the workarounds we had for V1 filters like backticks unquoting and just use field name arrays if possible.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+    if (op != null) {
+      switch (op) {
+        case TRUE:
+          return Expressions.alwaysTrue();
+
+        case FALSE:
+          return Expressions.alwaysFalse();
+
+        case IS_NULL:
+          return isNull(unquote(predicate.children()[FIRST_ORDINAL].toString()));
```
aokolnychyi, line 126: Weird. I think my original comment wasn't posted.

I was saying that we shouldn't use unquoting workarounds we had for V1 filters. Instead, we should use field arrays in V2 predicates.

I'd consider something like this (needs to be polished further).

```
private static final Joiner DOT = Joiner.on(".");

...

@SuppressWarnings("unchecked")
private static <T> T child(Predicate predicate, int index) {
  return (T) predicate.children()[index];
}

@SuppressWarnings("unchecked")
private static <T> T child(Predicate predicate) {
  org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
  Preconditions.checkArgument(children.length == 1, "TBD");
  return (T) children[0];
}

private static String toColumnName(NamedReference ref) {
  return DOT.join(ref.fieldNames());
}
```

And using it like this:

```
...

case IS_NULL:
  return isNull(toColumnName(child(predicate)));

case NOT_NULL:
  return notNull(toColumnName(child(predicate)));

...
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+
+        case NOT_NULL:
+          return notNull(unquote(predicate.children()[FIRST_ORDINAL].toString()));
+
+        case LT:
+          if (predicate.children()[SECOND_ORDINAL] instanceof LiteralValue) {
```
aokolnychyi, line 132: Isn't `LiteralValue` private to Spark? Should we use the public interface?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+                        ((LiteralValue) predicate.children()[SECOND_ORDINAL]).value())
+                    .toString());
+          }
+      }
+    }
+
+    return null;
+  }
+
+  private static Object convertUTF8StringIfNecessary(Object value) {
```
aokolnychyi, line 276: Can this method accept `Literal` from Spark and produce the Java object?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: This is actually borrowed from `SparkFilters`

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: I took @aokolnychyi's suggestion to use `leftChild` and `rightChild`

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+  private static final String IN = "IN";
+  private static final String IS_NULL = "IS_NULL";
+  private static final String NOT_NULL = "IS_NOT_NULL";
+  private static final String AND = "AND";
+  private static final String OR = "OR";
+  private static final String NOT = "NOT";
+  private static final String STARTS_WITH = "STARTS_WITH";
+
+  private static final Map<String, Expression.Operation> FILTERS =
+      ImmutableMap.<String, Expression.Operation>builder()
```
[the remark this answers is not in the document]
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,385 @@
...
+          .put(AND, Expression.Operation.AND)
+          .put(OR, Expression.Operation.OR)
+          .put(NOT, Expression.Operation.NOT)
+          .put(STARTS_WITH, Expression.Operation.STARTS_WITH)
+          .build();
+
+  private static final int FIRST_ORDINAL = 0;
+  private static final int SECOND_ORDINAL = 1;
+
+  public static Expression convert(Predicate[] predicates) {
```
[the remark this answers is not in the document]
huaxingao: Removed.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Mainly trying to validate the Spark filter logic. In binary comparison predicate, Expression cmp Expression is also valid in Spark, but I think Iceberg only accepts Attribute cmp Literal, so I also make sure it's either attribute cmp lit or lit cmp attribute.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: This is required. Otherwise, I can't use the same variable names (`left` and `right`) in difference case blocks.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Changed to `Literal`. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
kbendick: Oh big +1 I like this a lot.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
kbendick: Ah then absolutely keep it :)

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
```diff
@@ -0,0 +1,252 @@
...
+    Not filter = new Not(new And(equal, in));
+    Expression converted = SparkV2Filters.convert(filter);
+    Assert.assertNull("Expression should not be converted", converted);
+  }
+
+  @Test
+  public void testNotIn() {
+    NamedReference namedReference = FieldReference.apply("col");
+    LiteralValue v1 = new LiteralValue(1, DataTypes.IntegerType);
+    LiteralValue v2 = new LiteralValue(2, DataTypes.IntegerType);
```
kbendick, line 240: Nit: Can `Literal` be used in place of `LiteralValue` here?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,380 @@
...
+          .put(NOT, Operation.NOT)
+          .put(STARTS_WITH, Operation.STARTS_WITH)
+          .build();
+
+  private SparkV2Filters() {}
+
+  @SuppressWarnings("unchecked")
+  private static <T> T child(Predicate predicate) {
+    org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
+    Preconditions.checkArgument(children.length == 1, "%s should have one child", predicate);
```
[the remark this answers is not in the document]
huaxingao: Will fix. Thanks

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
huaxingao: Good catch! Will fix it.

huaxingao: Oh actually `Literal` is an abstract class, will have to use `LiteralValue` here.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
aokolnychyi: I see. It makes sense but what do you think about adapting the conversion logic a bit instead of doing two passes? We anyway need to check if left or right child is literal/reference. You could define helper methods like this.

```
private static boolean isRef(org.apache.spark.sql.connector.expressions.Expression expr) {
  return expr instanceof NamedReference;
}

private static boolean isLiteral(org.apache.spark.sql.connector.expressions.Expression expr) {
  return expr instanceof Literal;
}
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+    if (op != null) {
+      switch (op) {
+        case TRUE:
+          return Expressions.alwaysTrue();
+
+        case FALSE:
+          return Expressions.alwaysFalse();
+
+        case IS_NULL:
+          return isNull(toColumnName(child(predicate)));
```
aokolnychyi, line 138: To have this as one pass, we could have:

```
return isRef(child(predicate)) ? isNull(toColumnName(child(predicate))) : null;
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+
+        case NOT_NULL:
+          return notNull(toColumnName(child(predicate)));
+
+        case LT:
+          if (rightChild(predicate) instanceof Literal) {
```
aokolnychyi, line 144: To have this as one pass, we could have:

```
if (isRef(leftChild(predicate)) && isLiteral(rightChild(predicate))) {
  String columnName = toColumnName(leftChild(predicate));
  return lessThan(columnName, convertLiteral(rightChild(predicate)));

} else if (isRef(rightChild(predicate)) && isLiteral(leftChild(predicate))) {
  String columnName = toColumnName(rightChild(predicate));
  return greaterThan(columnName, convertLiteral(leftChild(predicate)));

} else {
  return null;
}
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          return startsWith(
+              toColumnName(leftChild(predicate)),
+              convertUTF8StringIfNecessary(rightChild(predicate)).toString());
+      }
+    }
+
+    return null;
+  }
+
+  private static Object convertUTF8StringIfNecessary(Literal literal) {
```
aokolnychyi, line 277: nit: Just `convertLiteral`?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
aokolnychyi: It is unexpected and I think we wrap the conversion logic into try-catch, so we should be fine.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          } else { // "<=>"
+            if (value == null) {
+              return isNull(attributeName);
+            } else {
+              return handleEqual(attributeName, value);
+            }
+          }
+
+        case IN:
+          return in(
```
aokolnychyi, line 212: To have this as one pass, we could have:

```
if (isSupportedInPredicate(predicate)) {
  return in(
      toColumnName((NamedReference) predicate.children()[0]),
      Arrays.stream(predicate.children())
          .skip(1)
          .map(child -> convertLiteral(((Literal<?>) child)))
          .filter(Objects::nonNull)
          .collect(Collectors.toList()));
} else {
  return null;
}
```

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+                  .filter(Objects::nonNull)
+                  .collect(Collectors.toList()));
+
+        case NOT:
+          Not notFilter = (Not) predicate;
+          Predicate childFilter = notFilter.child();
+          if (childFilter.name().equals("IN")) {
+            // infer an extra notNull predicate for Spark NOT IN filters
+            // as Iceberg expressions don't follow the 3-value SQL boolean logic
+            // col NOT IN (1, 2) in Spark is equivalent to notNull(col) && notIn(col, 1, 2) in
```
aokolnychyi, line 226: nit: What about replacing `equivalent` with `equal` so that the last line of the comment would fit on one line?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          return in(
+              toColumnName(leftChild(predicate)),
+              Arrays.stream(predicate.children())
+                  .skip(1)
+                  .map(val -> convertUTF8StringIfNecessary(((Literal) val)))
+                  .filter(Objects::nonNull)
+                  .collect(Collectors.toList()));
+
+        case NOT:
+          Not notFilter = (Not) predicate;
```
aokolnychyi, line 221: nit: `notPredicate`?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          Predicate childFilter = notFilter.child();
```
aokolnychyi, line 222: nit: `childPredicate`?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          if (childFilter.name().equals("IN")) {
```
aokolnychyi, line 223: nit: Use `IN` constant defined above?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+            // infer an extra notNull predicate for Spark NOT IN filters
+            // as Iceberg expressions don't follow the 3-value SQL boolean logic
+            // col NOT IN (1, 2) in Spark is equivalent to notNull(col) && notIn(col, 1, 2) in
+            // Iceberg
+            Expression notIn =
+                notIn(
+                    toColumnName(leftChild(childFilter)),
+                    Arrays.stream(childFilter.children())
+                        .skip(1)
+                        .map(val -> convertUTF8StringIfNecessary(((Literal) val)))
```
aokolnychyi, line 233: nit: `Literal<?>` to avoid warnings about raw usage of parameterized types?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+            if (left != null && right != null) {
+              return or(left, right);
+            }
+            return null;
+          }
+
+        case STARTS_WITH:
+          return startsWith(
+              toColumnName(leftChild(predicate)),
+              convertUTF8StringIfNecessary(rightChild(predicate)).toString());
```
aokolnychyi, line 270: Hm, do we need `toString` or a cast would work too to be sure we have a string?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+        case TRUE:
+        case FALSE:
+          return true;
+      }
+    }
+
+    return false;
+  }
+
+  private static boolean namedRefCompLit(Predicate predicate) {
```
aokolnychyi, line 372: I think splitting this into `isRef` and `isLiteral` should allow us to combine the validation and conversion into one pass. Then we won't have to jump between methods when looking at the code.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          String attributeName;
+          if (rightChild(predicate) instanceof Literal) {
+            attributeName = toColumnName(leftChild(predicate));
+            value = convertUTF8StringIfNecessary(rightChild(predicate));
+          } else {
+            attributeName = toColumnName(rightChild(predicate));
+            value = convertUTF8StringIfNecessary(leftChild(predicate));
+          }
+
+          if (predicate.name().equals("=")) {
```
aokolnychyi, line 198: nit: Shall we use `EQ` constant defined above?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
aokolnychyi, line 277: nit: `Literal` -> `Literal<?>`

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+        case OR:
+          Or orFilter = (Or) predicate;
+          return valid(orFilter.left()) && valid(orFilter.right());
+
+        case STARTS_WITH:
+          if (predicate.children().length != 2) {
+            return false;
+          }
+          return namedRefCompLit(predicate)
+              && ((Literal<?>) rightChild(predicate)).value() instanceof String;
```
aokolnychyi, line 361: Hm, was it actually correct? Wouldn't it be `UTF8String`?

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
```diff
@@ -0,0 +1,252 @@
...
+import org.apache.spark.sql.connector.expressions.filter.Or;
+import org.apache.spark.sql.connector.expressions.filter.Predicate;
+import org.apache.spark.sql.types.DataTypes;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestSparkV2Filters {
+
+  @Test
+  public void testV2Filters() {
```
aokolnychyi, line 42: Can we add a test for `startsWith`? I think the was a problem in the conversion and tests did not catch it.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+        default:
+          return true;
+      }
+    }
+
+    return false;
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
+  private static boolean valid(Predicate predicate) {
```
aokolnychyi, line 317: I kept jumping from `convert` to `valid` when looking at the code. I feel having both the validation and conversion in one function would be a bit easier.

aokolnychyi (commented):
I did a detailed pass. I think this PR is really close. I'd probably combine the validation and conversion into one method. It also seems we have a bug for `starts_with`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Changed. Thanks for the suggestion!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+  private static <T> T child(Predicate predicate) {
+    org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
+    Preconditions.checkArgument(
+        children.length == 1, "Predicate should have one child: %s", predicate);
+    return (T) children[0];
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <T> T leftChild(Predicate predicate) {
+    org.apache.spark.sql.connector.expressions.Expression[] children = predicate.children();
```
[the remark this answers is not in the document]
huaxingao: Added `childAtIndex`. Thanks for the suggestion!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,381 @@
...
+          .put(AND, Operation.AND)
+          .put(OR, Operation.OR)
+          .put(NOT, Operation.NOT)
+          .put(STARTS_WITH, Operation.STARTS_WITH)
+          .build();
+
+  private SparkV2Filters() {}
+
+  @SuppressWarnings("unchecked")
+  private static <T> T child(Predicate predicate) {
```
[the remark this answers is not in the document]
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks for the suggestion!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks for the suggestion!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Changed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Changed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Seems this `toString` is needed.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Agree!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Good catch! It should be `UTF8String`

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/TestSparkV2Filters.java
huaxingao: Added. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Agree!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,345 @@
...
+          .put(OR, Operation.OR)
+          .put(NOT, Operation.NOT)
+          .put(STARTS_WITH, Operation.STARTS_WITH)
+          .build();
+
+  private SparkV2Filters() {}
+
+  @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:MethodLength"})
+  public static Expression convert(Predicate predicate) {
+
```
aokolnychyi, line 98: nit: This empty line is probably unnecessary.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,345 @@
...
+                    .filter(Objects::nonNull)
+                    .collect(Collectors.toList()));
+          } else {
+            return null;
+          }
+
+        case NOT:
+          Not notPredicate = (Not) predicate;
+          Predicate childPredicate = notPredicate.child();
+          if (childPredicate.name().equals(IN)) {
```
aokolnychyi, line 200: Shall we add also `isSupportedInPredicate(childPredicate)` to the condition?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    return false;
+  }
+
+  private static boolean isSupportedInPredicate(Predicate predicate) {
+    if (!isRef(childAtIndex(predicate, 0))) {
+      return false;
+    } else {
+      return Arrays.stream(predicate.children()).skip(1).allMatch(val -> isLiteral(val));
```
aokolnychyi, line 342: nit: `val -> isLiteral(val)` -> `SparkV2Filters::isLiteral`?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
```diff
@@ -0,0 +1,345 @@
...
+            Expression left = convert(orPredicate.left());
+            Expression right = convert(orPredicate.right());
+            if (left != null && right != null) {
+              return or(left, right);
+            }
+            return null;
+          }
+
+        case STARTS_WITH:
+          if (isRef(leftChild(predicate))
```
aokolnychyi, line 244: Is it realistic that Spark can pass something other than a reference followed by String? I think the other validation in this method is reasonable as we could technically have other valid expressions with structs inside IN or other predicates, which Iceberg does not support yet. I am not sure about this one, though.

I am fine keeping it too but I'd move it to `isSupportedStartsWithPredicate`. Up to you here, @huaxingao.

aokolnychyi (approved):
Looks great! A few final nits and is ready to go! Thanks for the work, @huaxingao!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Added. Thanks!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkV2Filters.java
huaxingao: Fixed. Thanks!

aokolnychyi (approved):

aokolnychyi:
Thanks, @huaxingao! Thanks for reviewing, @szehon-ho @kbendick!

huaxingao:
Thank you all very much for your help!! @aokolnychyi @szehon-ho @kbendick

PR #5303, opened 2022-07-18
Add a BoundBooleanExpressionVisitor for visiting bound expressions

samredai:
This adds the groundwork for a `BoundBooleanExpressionVisitor`. It only implements visiting a `BoundIn` expression since that's the only bound predicate implemented so far.

As more expressions are added, logic needs to be added to the `visit_bound_predicate` method.


samredai:
cc: @CircArgs @Fokko @dramaticlly

python/tests/expressions/test_expressions_base.py
```diff
@@ -634,3 +651,16 @@ def test_not_expression_binding(unbound_not_expression, expected_bound_expressio
...
+    bound_expression = base.BoundIn[str](
+        term=base.BoundReference(
+            field=NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
+            accessor=Accessor(position=0, inner=None),
+        ),
+        literals=(StringLiteral("foo"), StringLiteral("bar")),
+    )
+    visitor = FooBoundBooleanExpressionVisitor()
+    result = base.visit(bound_expression, visitor=visitor)
+    assert result == ["IN"]
```
samredai, line 666: Just a simple test here for now using the test implementation `FooBoundBooleanExpressionVisitor`. Once more predicates are added we include more robust tests.

python/pyiceberg/expressions/base.py
```diff
@@ -327,30 +328,26 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundIn[T]:
         return BoundIn(bound_ref, tuple(lit.to(bound_ref.field.field_type) for lit in self.literals))  # type: ignore
 
 
-class BooleanExpressionVisitor(Generic[T], ABC):
```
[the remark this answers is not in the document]
rdblue: I don't think this is a good idea. Don't we want to ensure that all of the cases are handled? What is the return value of visit methods that are not implemented? `None`?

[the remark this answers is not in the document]
samredai: Yes it would return `None`. What prompted me to do this is that `BoundBooleanExpressionVisitor` needs to inherit from `BooleanExpression` in order to use the `visit` function defined here. But that means it would need to override each method it doesn't use and return `None` (this is my understanding of what the Java one does).

But I guess there's no reason that needs to be a default behavior from the parent class and it's better to explicitly define that for this visitor. I'll change this back and implement those methods.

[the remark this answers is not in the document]
Fokko: I would also prefer to keep it abstract. What's also common is to raise a `NotImplementedError`, see https://docs.python.org/3/library/exceptions.html#NotImplementedError. This way you still need to define all the methods, but you can just raise the exceptions on the methods that don't make any sense.

python/pyiceberg/expressions/base.py
```diff
@@ -474,3 +474,77 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+
+        Args:
+            predicate (UnboundPredicate[T]): An unbound predicate
+
+        Raises:
+            TypeError: This always raises since an unbound predicate is not expected in a bound boolean expression
+        """
+        raise TypeError(f"Not a bound predicate: {predicate}")
+
+    @singledispatchmethod
```
Fokko, line 523: Keep in mind that the `singledispatchmethod` is much slower than just single dispatch: https://bugs.python.org/issue40988

I'm not sure how big these expressions trees can grow. In the case of auto-generated SQL from some BI tool, I think it can become significant.

samredai: That's right, I forgot about the performance hit when using this. I replaced it with a simple if block that does an `isinstance` check.

python/pyiceberg/expressions/base.py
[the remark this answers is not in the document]
samredai: Ok updated to revert this back to an ABC class. Since `BoundBooleanExpressionVisitor` is also an ABC class, the abstractmethods are just carried forward. I'll keep in mind the `NotImplementedError` option as we add concrete implementations. 👍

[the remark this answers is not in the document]
rdblue: > BoundBooleanExpressionVisitor needs to inherit from BooleanExpression in order to use the visit function defined here

I'm not sure I'm following. Did you mean `BooleanExpression`? I don't see why the visitor would need to be an expression.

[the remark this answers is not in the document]
samredai: @rdblue sorry that was a mistype, I meant it inherits from `BooleanExpressionVisitor`.

python/pyiceberg/expressions/base.py
```diff
@@ -431,7 +431,13 @@ def _(obj: And, visitor: BooleanExpressionVisitor[T]) -> T:
 @visit.register(In)
 def _(obj: In, visitor: BooleanExpressionVisitor[T]) -> T:
     """Visit an In boolean expression with a concrete BooleanExpressionVisitor"""
-    return visitor.visit_unbound_predicate(predicate=obj)
+    return visitor.visit_unbound_predicate(obj)
+
+
+@visit.register(BoundIn)
+def _(obj: BoundIn, visitor: BooleanExpressionVisitor[T]) -> T:
```
rdblue, line 438: I don't think this should be registered for the regular expression visitor, just the bound visitor. The regular visitor should always call `visit_bound_predicate` for all bound predicates.

samredai: Ah right, makes sense--the whole purpose for the inheritance. Updated!

rdblue:
@samredai, can you rebase?

samredai:
> @samredai, can you rebase?

Rebased, thanks for the ping! 👍

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,55 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
 
     def visit_bound_predicate(self, predicate) -> BooleanExpression:
         raise TypeError(f"Found already bound predicate: {predicate}")
+
+
+class BoundBooleanExpressionVisitor(BooleanExpressionVisitor[T], ABC):
+    @abstractmethod
+    def visit_in(self, ref: BoundReference[T], literals: set[Literal[Any]]) -> T:
```
rdblue, line 690: If we update this to Term then we won't need to implement `handle_non_reference` because visitors will be built to handle non-references.

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,42 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+    def visit_unbound_predicate(self, predicate: UnboundPredicate[T]):
+        """Visit an unbound predicate
+        Args:
+            predicate (UnboundPredicate[T]): An unbound predicate
+        Raises:
+            TypeError: This always raises since an unbound predicate is not expected in a bound boolean expression
+        """
+        raise TypeError(f"Not a bound predicate: {predicate}")
+
+    def visit_bound_predicate(self, predicate: BoundPredicate[T]) -> T:
```
rdblue, line 710: I think this should call another singledispatch method that accepts a bound predicate type and a bound expression visitor, rather than calling `isinstance` itself.

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,42 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+        """Visit a bound predicate
+        Args:
+            predicate (BoundPredicate[T]): A bound predicate
+        Raises:
+            TypeError: When an unsupported predicate is found
+        Returns:
+            R: The return type defined by the concrete bound boolean expression visitor
+        """
+        if not isinstance(predicate.term, BoundReference):
+            raise TypeError(f"{type(self).__name__} does not support non-reference: {predicate.term}")
```
rdblue, line 720: I don't think that we need to fail here. This is most likely an artifact of not having `Transform` expressions at first in the Java implementation.

python/pyiceberg/expressions/base.py
rdblue: Rather than `ref` I think this should accept `term: BoundTerm[T]`

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,42 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+    def visit_in(self, ref: BoundReference[T], literals: set[Literal[Any]]) -> T:
+        """Visit the ref and literals from an In boolean expression
+        Args:
+            ref (BoundReference[T]): The ref used in the In boolean expression
+            literals (Tuple[Literal[T], ...]): The literals used in the In boolean expression
+        Raises:
+            NotImplementedError: If the concrete bound boolean expression visitor does not override this method
+        Returns:
+            R: The return type defined by the concrete bound boolean expression visitor
+        """
```
rdblue, line 699: Do you want to add methods for the other expression types?

samredai:
Updated this PR to address the remaining feedback. Thanks for the reminder @Fokko!

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,166 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+        """Visit a bound predicate
+        Args:
+            predicate (BoundPredicate[T]): A bound predicate
+        """
+        return visit_bound_predicate(predicate, self)
+
+
+@singledispatch
+def visit_bound_predicate(obj, visitor: BooleanExpressionVisitor[T]) -> T:  # pylint: disable=unused-argument
+    pass
```
rdblue, line 776: Don't we want to throw an exception if this case is hit? It is an unknown predicate so we can't handle it.

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,166 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+
+@visit_bound_predicate.register(BoundNotNull)
+def _(obj: BoundNotNull, visitor: BoundBooleanExpressionVisitor[T]) -> T:
+    """Visit a bound NotNull predicate"""
+    return visitor.visit_not_null(term=obj.term)
+
+
+@visit_bound_predicate.register(BoundEqualTo)
+def _(obj: BoundEqualTo, visitor: BoundBooleanExpressionVisitor[T]) -> T:
+    """Visit a bound EqualTo predicate"""
```
rdblue, line 817: Are these doc strings visible anywhere? It seems pretty clear what's happening and since this is a `@singledispatch` method I'm not sure when we'd ever see them.

python/pyiceberg/expressions/base.py
```diff
@@ -672,3 +683,166 @@ def visit_unbound_predicate(self, predicate) -> BooleanExpression:
...
+
+
+@visit_bound_predicate.register(BoundIn)
+def _(obj: BoundIn, visitor: BoundBooleanExpressionVisitor[T]) -> T:
+    """Visit a bound In predicate"""
+    return visitor.visit_in(term=obj.term, literals=obj.literals)
+
+
+@visit_bound_predicate.register(BoundNotIn)
+def _(obj: BoundNotIn, visitor: BoundBooleanExpressionVisitor[T]) -> T:
```
rdblue, line 786: Nit: I'd probably rename `obj` to `expr` since it is an expression.

python/tests/expressions/test_expressions_base.py
```diff
@@ -841,3 +924,290 @@ def test_not_expression_binding(unbound_not_expression, expected_bound_expressio
     """Test that visiting an unbound NOT expression with a bind-visitor returns the expected bound expression"""
     bound_expression = base.visit(unbound_not_expression, visitor=base.BindVisitor(schema=table_schema_simple))
     assert bound_expression == expected_bound_expression
+
+
+def test_bound_boolean_expression_visitor_ignore_is_nan_for_string_type():
+    """Test ignoring IsNan expressions when the field type is a String"""
```
rdblue, line 930: This is suspicious. You can't bind `IsNaN` to a string, so this should never happen. In addition, I don't see why the visitor would handle this any differently. There may be a runtime failure because the expression is improperly bound (users should never do this) but the visitor shouldn't simply ignore it.

python/tests/expressions/test_expressions_base.py
```diff
@@ -841,3 +924,290 @@ def test_not_expression_binding(unbound_not_expression, expected_bound_expressio
...
+        ),
+    )
+    visitor = FooBoundBooleanExpressionVisitor()
+    result = base.visit(bound_expression, visitor=visitor)
+    assert result == ["TRUE"]
+
+
+def test_bound_boolean_expression_visitor_always_false():
+    """Test visiting an And expression with a bound boolean expression visitor that resolves to AlwaysFalse"""
+    bound_expression = base.And(
```
rdblue, line 1049: Why not test the false/true expressions directly?

python/tests/expressions/test_expressions_base.py
```diff
@@ -841,3 +924,290 @@ def test_not_expression_binding(unbound_not_expression, expected_bound_expressio
...
+            term=base.BoundReference(
+                field=NestedField(field_id=2, name="bar", field_type=StringType(), required=False),
+                accessor=Accessor(position=1, inner=None),
+            ),
+            literal=StringLiteral("foo"),
+        ),
+    )
+    visitor = FooBoundBooleanExpressionVisitor()
+    result = base.visit(bound_expression, visitor=visitor)
+    assert result == ["NOT_EQUAL", "EQUAL", "AND"]
```
rdblue, line 1023: Can you add separate tests that validate not-equal and equal in isolation?

samredai: Done! Much clearer

python/tests/expressions/test_expressions_base.py
samredai: Agreed and udated!

python/tests/expressions/test_expressions_base.py
samredai: I see, I noticed the behavior so included this test but what you're saying makes sense. So is it safe to just remove this test and expect that the higher-level APIs would prevent an `IsNaN` expression on a string type? Or do you feel that we need to add some validation here? (sounds like we don't need that validation but just want to confirm)
```py
@visit_bound_predicate.register(BoundIsNaN)
def _(expr: BoundIsNaN, visitor: BoundBooleanExpressionVisitor[T]) -> T:
    # validate that `expr.term.field.field_type` is an appropriate type
    return visitor.visit_is_nan(term=expr.term)
```

python/pyiceberg/expressions/base.py
samredai: Updated 👍

python/pyiceberg/expressions/base.py
samredai: Good point, removed.

python/pyiceberg/expressions/base.py
samredai: I think I misinterpreted your previous comment about not needing to fail, but I see that this here is different. Added raising of a `TypeError`.
```py
@singledispatch
def visit_bound_predicate(expr, visitor: BooleanExpressionVisitor[T]) -> T:  # pylint: disable=unused-argument
    raise TypeError(f"Unknown predicate: {expr}")
```

python/tests/expressions/test_expressions_base.py
rdblue: We don't need validation in the visitor. We should assume that the bound expression has been created correctly by the binding process and not by people instantiating the expressions directly. Maybe we should also move these into `bound.py` to avoid exposing them?

We also need to understand why the expression was skipped. That sounds like a bug to me. I don't think there's anything that checks.

rdblue: @samredai, we still need to find out what's going on here. This appears to be ignoring parts of the bound expression.

python/tests/expressions/test_expressions_base.py
```diff
@@ -104,17 +109,95 @@ def visit_test_expression_b(self) -> List:
...
 
 
 @base.visit.register(ExpressionB)
-def _(obj: ExpressionB, visitor: BooleanExpressionVisitor) -> List:
+def _(obj: ExpressionB, visitor: FooBooleanExpressionVisitor) -> List:
     """Visit a ExpressionB with a BooleanExpressionVisitor"""
     return visitor.visit_test_expression_b()
 
 
+class FooBoundBooleanExpressionVisitor(base.BoundBooleanExpressionVisitor[List]):
```
rdblue, line 123: Nit: Rather than `Foo` it would be better to name this `TestVisitor` or something.

samredai: I originally had it as `TestBoundBooleanExpressionVisitor` but since classes that start with `Test` is a pytest convention, the test suite was getting confused. How about `ExampleVisitor`?

python/tests/expressions/test_expressions_base.py
rdblue: Okay, I looked at what's happening here and it is a bit misleading because we are rewriting expressions in the expression constructors. `BoundIsNaN(...)` actually returns `AlwaysFalse()` because it cannot be true for a string. Then `Not(AlwaysFalse())` returns `AlwaysTrue()` and `And` returns its left argument because the right argument is a constant.

In the end, this is testing expression construction that simplifies automatically, not the visitor. I'm going to just remove this test since it looks like everything functions properly.

rdblue (approved):

rdblue:
Thanks, @samredai!

PR #5304, opened 2022-07-19
AWS: handle_s3_glue_exception

xingfanx:
handle s3 and other glue exception in commit phase

Tested with unittests

### DIFF
--- a/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
@@ -39,10 +39,12 @@ import org.mockito.Mockito;
 import software.amazon.awssdk.services.glue.model.AccessDeniedException;
 import software.amazon.awssdk.services.glue.model.ConcurrentModificationException;
 import software.amazon.awssdk.services.glue.model.EntityNotFoundException;
+import software.amazon.awssdk.services.glue.model.GlueException;
 import software.amazon.awssdk.services.glue.model.ValidationException;
 import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
 import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
 
 public class TestGlueCatalogCommitFailure extends GlueTestBase {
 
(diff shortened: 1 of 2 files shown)


aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
```diff
@@ -317,6 +319,49 @@ public void testGlueValidationException() {
...
+    TableMetadata metadataV1 = ops.current();
+    TableMetadata metadataV2 = updateTable(table, ops);
+
+    GlueTableOperations spyOps = Mockito.spy(ops);
+    failCommitAndThrowException(spyOps, S3Exception.builder().build());
+
+    AssertHelpers.assertThrows(
+            null,
+            S3Exception.class,
+            () -> spyOps.commit(metadataV2, metadataV1));
```
amogh-jahagirdar, line 336: Indentation

aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
```diff
@@ -317,6 +319,49 @@ public void testGlueValidationException() {
...
+    TableMetadata metadataV1 = ops.current();
+    TableMetadata metadataV2 = updateTable(table, ops);
+
+    GlueTableOperations spyOps = Mockito.spy(ops);
+    failCommitAndThrowException(spyOps, GlueException.builder().build());
+
+    AssertHelpers.assertThrows(
+            null,
+            GlueException.class,
+            () -> spyOps.commit(metadataV2, metadataV1));
```
amogh-jahagirdar, line 358: Same indentation

amogh-jahagirdar (commented):
Thanks! Other than some nit comments, do you mind updating the title and description so it's a bit more clear what we're trying to handle in terms of s3 and glue exception, and why we should handle these exceptions differently then the other Runtime exceptions.

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -144,6 +145,10 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
           "Cannot commit %s because Glue encountered a validation exception " +
               "while accessing requested resources",
           tableName());
+    } catch (S3Exception e) {
+      throw e;
+    } catch (software.amazon.awssdk.services.glue.model.GlueException e) {
```
jackye1995, line 150: nit: I don't think we need the full class path here, can import the class

aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
xingfanx: I think this aligns with the current indentation scheme

aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogCommitFailure.java
xingfanx: I think this aligns with the current indentation scheme

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -144,6 +146,10 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
           "Cannot commit %s because Glue encountered a validation exception " +
               "while accessing requested resources",
           tableName());
+    } catch (S3Exception e) {
+      throw e;
+    } catch (GlueException e) {
+      throw e;
```
jackye1995, line 152: actually we can do better here, instead of catching each case, we can change the `catch (RuntimeException persistFailure) {` block, and check `if (persistFailure instanceof AwsServiceException)` then we check the `statusCode`. If that is 5xx then we should go to the `FAILURE` case so the error gets retried. Otherwise it should directly throw. This can fully prevent any dependent service error without us having to add different types of error handling as we evolve the service.

xingfanx: sgtm

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -149,6 +150,15 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
           fullTableName, persistFailure);
       commitStatus = checkCommitStatus(newMetadataLocation, metadata);
```
jackye1995, line 151: `checkCommitStatus` is a heavy operation, I think the check of AwsServiceException can happen before that?

jackye1995 (approved):
looks good to me, thanks for the fix!

amogh-jahagirdar (approved):
LGTM, thanks for the fix!

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -147,7 +148,17 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     } catch (RuntimeException persistFailure) {
       LOG.error("Confirming if commit to {} indeed failed to persist, attempting to reconnect and check.",
           fullTableName, persistFailure);
-      commitStatus = checkCommitStatus(newMetadataLocation, metadata);
+
+      if (persistFailure instanceof AwsServiceException) {
```
jfz, line 152: Can this be a separate catch block to simplify the exception handling flow? like 

>   } catch (AwsServiceException serviceException) {
> ...
>   } catch (RuntimeException runtimeException) {
> ...
>

aws/src/main/java/org/apache/iceberg/aws/glue/GlueTableOperations.java
```diff
@@ -147,7 +148,17 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     } catch (RuntimeException persistFailure) {
       LOG.error("Confirming if commit to {} indeed failed to persist, attempting to reconnect and check.",
```
jfz, line 149: I think this logging should be moved around since `checkCommitStatus` is now conditional

PR #5305, opened 2022-07-19
Spark - Implement FunctionCatalog and Truncate

kbendick:
Implements `FunctionCatalog` for Spark 3.3 and implements all variants of `Truncate`.

## FunctionCatalog

This allows users of `SparkCatalog` and `SparkSessionCatalog` to use `truncate` without having to register it as a UDF.

All Iceberg functions that we register into the function catalog are accessible when used with an Iceberg spark catalog and:

1. No namespace is referenced - the storage partitioned joins implementation requires this.
e.g. `my_catalog.truncate(width, value)`. 
**Note** - Using `truncate(width, value)` typically does not work, as Spark adds the namespace to the call. `system.truncate` should be preferred.
2. The `system` namespace is referenced, to match called procedure syntax. **Note** this only works right now with the `SparkCatalog`, as the `SparkSessionCatalog` has logic in Spark [to verify the namespace exists](https://github.com/apache/iceberg/pull/5305#discussion_r927178316).
 e.g. `my_catalog.system.truncate(width, value)` or `system.truncate(6, column)`

## Truncate

The truncate function also allows for a dynamic width or the width to come from a column - though typically the width will likely be static for one given call as it's mostly intended to be used to match partition transforms (specifically with joins or on non-partition columns to create a new column in the data without needing to partition on it).

This PR refactors the definition of the transform functions into a utility class where needed so that Spark’s magic functions can call them via the static `invoke` function and not duplicate logic. This allows Spark to include the functions in codegen.

### Special Considerations for Using Function Catalog Efficiently via Magic Functions and Code Gen

The requirements for magic functions to be used with codegen include that:
1. `invoke` is a static function
4. `invoke` takes in the primitive types / native Spark types corresponding to each of Spark's input DataTypes (e.g. int for IntegerType and UTF8String for StringType).

Further documentation on the magic functions is found [here in the ScalarFunction JavaDoc](https://spark.apache.org/docs/latest/api/java/index.html?org/apache/spark/sql/connector/catalog/functions/ScalarFunction.html)

This partially closes https://github.com/apache/iceberg/issues/5349

### DIFF
--- a/api/src/main/java/org/apache/iceberg/transforms/Truncate.java
+++ b/api/src/main/java/org/apache/iceberg/transforms/Truncate.java
@@ -31,6 +31,7 @@ import org.apache.iceberg.expressions.UnboundPredicate;
 import org.apache.iceberg.relocated.com.google.common.base.Objects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.types.Type;
+import org.apache.iceberg.util.TruncateUtil;
 import org.apache.iceberg.util.UnicodeUtil;
 
 abstract class Truncate<T> implements Transform<T, T> {
@@ -87,7 +88,7 @@ abstract class Truncate<T> implements Transform<T, T> {
         return null;
       }
 
-      return value - (((value % width) + width) % width);
+      return TruncateUtil.truncateInt(width, value);
     }
 
     @Override
@@ -171,7 +172,7 @@ abstract class Truncate<T> implements Transform<T, T> {
         return null;
       }
 
-      return value - (((value % width) + width) % width);
+      return TruncateUtil.truncateLong(width, value);
     }
 
     @Override
@@ -391,9 +392,7 @@ abstract class Truncate<T> implements Transform<T, T> {
         return null;
       }
 
-      ByteBuffer ret = value.duplicate();
-      ret.limit(Math.min(value.limit(), value.position() + length));
-      return ret;
+      return TruncateUtil.truncateByteBuffer(length, value);
     }
 
     @Override
@@ -480,16 +479,7 @@ abstract class Truncate<T> implements Transform<T, T> {
         return null;
       }
 
-      BigDecimal remainder =
-          new BigDecimal(
-              value
-                  .unscaledValue()
-                  .remainder(unscaledWidth)
-                  .add(unscaledWidth)
-                  .remainder(unscaledWidth),
-              value.scale());
-
-      return value.subtract(remainder);
+      return TruncateUtil.truncateDecimal(unscaledWidth, value);
     }
 
     @Override
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
@@ -34,7 +34,11 @@ import org.apache.spark.sql.connector.iceberg.catalog.Procedure;
 import org.apache.spark.sql.connector.iceberg.catalog.ProcedureCatalog;
 
 abstract class BaseCatalog
-    implements StagingTableCatalog, ProcedureCatalog, SupportsNamespaces, HasIcebergCatalog, FunctionCatalog {
+    implements StagingTableCatalog,
+        ProcedureCatalog,
+        SupportsNamespaces,
+        HasIcebergCatalog,
+        FunctionCatalog {
 
   @Override
   public Procedure loadProcedure(Identifier ident) throws NoSuchProcedureException {
@@ -55,7 +59,8 @@ abstract class BaseCatalog
 
   @Override
   public Identifier[] listFunctions(String[] namespace) throws NoSuchNamespaceException {
-    if (namespace.length == 0 || (namespace.length == 1 && namespace[0].equalsIgnoreCase("system"))) {
+    if (namespace.length == 0
+        || (namespace.length == 1 && namespace[0].equalsIgnoreCase("system"))) {
       return SparkFunctions.list().stream()
           .map(name -> Identifier.of(namespace, name))
           .toArray(Identifier[]::new);
@@ -71,9 +76,11 @@ abstract class BaseCatalog
     String[] namespace = ident.namespace();
     String name = ident.name();
 
-    // Allow for empty namespace as Spark's storage partitioned joins look up the corresponding transform
-    // functions, like `bucket`, with an empty namespace
-    if (namespace.length == 0 || (namespace.length == 1 && namespace[0].equalsIgnoreCase("system"))) {
+    // Allow for empty namespace as Spark's storage partitioned joins look up
+    // the corresponding functions to generate transforms for partitioning
+    // with an empty namespace, such as `bucket`.
+    if (namespace.length == 0
+        || (namespace.length == 1 && namespace[0].equalsIgnoreCase("system"))) {
       UnboundFunction func = SparkFunctions.load(name);
       if (func != null) {
         return func;
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
@@ -55,13 +55,11 @@ abstract class BaseCatalog
 
   @Override
   public Identifier[] listFunctions(String[] namespace) throws NoSuchNamespaceException {
-    if (namespace.length == 1 && namespace[0].equalsIgnoreCase("system")) {
+    if (namespace.length == 0 || (namespace.length == 1 && namespace[0].equalsIgnoreCase("system"))) {
       return SparkFunctions.list().stream()
           .map(name -> Identifier.of(namespace, name))
           .toArray(Identifier[]::new);
-    }
-
-    if (namespace.length == 0 || namespaceExists(namespace)) {
+    } else if (namespaceExists(namespace)) {
       return new Identifier[0];
     }
 
@@ -73,7 +71,9 @@ abstract class BaseCatalog
     String[] namespace = ident.namespace();
     String name = ident.name();
 
-    if (namespace.length == 1 && namespace[0].equalsIgnoreCase("system")) {
+    // Allow for empty namespace as Spark's storage partitioned joins look up the corresponding transform
+    // functions, like `bucket`, with an empty namespace
+    if (namespace.length == 0 || (namespace.length == 1 && namespace[0].equalsIgnoreCase("system"))) {
       UnboundFunction func = SparkFunctions.load(name);
       if (func != null) {
         return func;
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
@@ -18,18 +18,23 @@
  */
 package org.apache.iceberg.spark;
 
+import org.apache.iceberg.spark.functions.SparkFunctions;
 import org.apache.iceberg.spark.procedures.SparkProcedures;
 import org.apache.iceberg.spark.procedures.SparkProcedures.ProcedureBuilder;
 import org.apache.iceberg.spark.source.HasIcebergCatalog;
+import org.apache.spark.sql.catalyst.analysis.NoSuchFunctionException;
+import org.apache.spark.sql.catalyst.analysis.NoSuchNamespaceException;
 import org.apache.spark.sql.catalyst.analysis.NoSuchProcedureException;
+import org.apache.spark.sql.connector.catalog.FunctionCatalog;
 import org.apache.spark.sql.connector.catalog.Identifier;
 import org.apache.spark.sql.connector.catalog.StagingTableCatalog;
 import org.apache.spark.sql.connector.catalog.SupportsNamespaces;
+import org.apache.spark.sql.connector.catalog.functions.UnboundFunction;
 import org.apache.spark.sql.connector.iceberg.catalog.Procedure;
 import org.apache.spark.sql.connector.iceberg.catalog.ProcedureCatalog;
 
 abstract class BaseCatalog
-    implements StagingTableCatalog, ProcedureCatalog, SupportsNamespaces, HasIcebergCatalog {
+    implements StagingTableCatalog, ProcedureCatalog, SupportsNamespaces, HasIcebergCatalog, FunctionCatalog {
 
   @Override
   public Procedure loadProcedure(Identifier ident) throws NoSuchProcedureException {
@@ -46,5 +51,35 @@ abstract class BaseCatalog
     }
 
     throw new NoSuchProcedureException(ident);
+  }
+
+  @Override
+  public Identifier[] listFunctions(String[] namespace) throws NoSuchNamespaceException {
+    if (namespace.length == 1 && namespace[0].equalsIgnoreCase("system")) {
+      return SparkFunctions.list().stream()
+          .map(name -> Identifier.of(namespace, name))
+          .toArray(Identifier[]::new);
+    }
+
+    if (namespace.length == 0 || namespaceExists(namespace)) {
+      return new Identifier[0];
+    }
+
+    throw new NoSuchNamespaceException(namespace);
+  }
+
+  @Override
+  public UnboundFunction loadFunction(Identifier ident) throws NoSuchFunctionException {
+    String[] namespace = ident.namespace();
+    String name = ident.name();
+
+    if (namespace.length == 1 && namespace[0].equalsIgnoreCase("system")) {
+      UnboundFunction func = SparkFunctions.load(name);
+      if (func != null) {
+        return func;
+      }
+    }
+
+    throw new NoSuchFunctionException(ident);
   }
 }
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/SparkFunctions.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/SparkFunctions.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.functions;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.spark.sql.connector.catalog.functions.UnboundFunction;
+
+public class SparkFunctions {
+
+  private SparkFunctions() {
+  }
+
+  private static final Map<String, UnboundFunction> FUNCTIONS = ImmutableMap.of(
+      "truncate", new TruncateFunction()
+  );
+
+  private static final List<String> FUNCTION_NAMES = ImmutableList.copyOf(FUNCTIONS.keySet());
+
+  // Functions that are added to all Iceberg catalogs get registered under the `system` namespace, so a list of
+  // names alone is returned.
+  public static List<String> list() {
+    return FUNCTION_NAMES;
+  }
+
+  public static UnboundFunction load(String name) {
+    // function resolution is case insensitive to match the existing Spark behavior for functions
+    UnboundFunction func = FUNCTIONS.get(name.toLowerCase(Locale.ROOT));
+    return func;
+  }
+
+}
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/SparkFunctions.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/SparkFunctions.java
@@ -16,7 +16,6 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
 package org.apache.iceberg.spark.functions;
 
 import java.util.List;
@@ -28,17 +27,15 @@ import org.apache.spark.sql.connector.catalog.functions.UnboundFunction;
 
 public class SparkFunctions {
 
-  private SparkFunctions() {
-  }
+  private SparkFunctions() {}
 
-  private static final Map<String, UnboundFunction> FUNCTIONS = ImmutableMap.of(
-      "truncate", new TruncateFunction()
-  );
+  private static final Map<String, UnboundFunction> FUNCTIONS =
+      ImmutableMap.of("truncate", new TruncateFunction());
 
   private static final List<String> FUNCTION_NAMES = ImmutableList.copyOf(FUNCTIONS.keySet());
 
-  // Functions that are added to all Iceberg catalogs get registered under the `system` namespace, so a list of
-  // names alone is returned.
+  // Functions that are added to all Iceberg catalogs should be accessed with either the `system`
+  // namespace or no namespace at all, so a list of names alone is returned.
   public static List<String> list() {
     return FUNCTION_NAMES;
   }
@@ -48,5 +45,4 @@ public class SparkFunctions {
     UnboundFunction func = FUNCTIONS.get(name.toLowerCase(Locale.ROOT));
     return func;
   }
-
 }
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
@@ -1,0 +1,388 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.functions;
+
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.util.ByteBuffers;
+import org.apache.iceberg.util.TruncateUtil;
+import org.apache.iceberg.util.UnicodeUtil;
+import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.connector.catalog.functions.BoundFunction;
+import org.apache.spark.sql.connector.catalog.functions.ScalarFunction;
+import org.apache.spark.sql.connector.catalog.functions.UnboundFunction;
+import org.apache.spark.sql.types.BinaryType;
+import org.apache.spark.sql.types.ByteType;
+import org.apache.spark.sql.types.CharType;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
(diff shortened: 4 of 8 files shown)


kbendick:
After this, I'll add `bucket` and `zorder` as well.

This is to facilitate the usage of the various transforms from PySpark as well as SQL.

Additionally, having a `zorder` function will make it possible for people to pre-sort their data on input, vs having to zorder sort it when running a data compaction job.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+import org.apache.spark.sql.types.Decimal;
+import org.apache.spark.sql.types.DecimalType;
+import org.apache.spark.sql.types.IntegerType;
+import org.apache.spark.sql.types.LongType;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.apache.spark.unsafe.types.UTF8String;
+
+/**
+ * Implmenetation of {@link UnboundFunction} that matches the {@code truncate} transformation.
```
rdblue, line 46: Typo: implementation.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
```diff
@@ -46,4 +54,56 @@ public Procedure loadProcedure(Identifier ident) throws NoSuchProcedureException
...
+   *     The namespace is not specified,
+   *     i.e. {@code SELECT truncate(1, 4)} or {@code SELECT my_catalog.truncate(1, 4)}
+   *   </li>
+   *   <li>
+   *     The implicit <b>system</b> namespace is used, to mirror call procedure syntax,
+   *     i.e. {@code SELECT system.truncate(1, 4)} or {@code SELECT my_catalog.system.truncate(1, 4)}
+   *   </li>
+   *   <li>
+   *     A namespace that exists within the catalog is referenced,
+   *     i.e. {@code SELECT ns.truncate(1, 4)} or {@code SELECT my_catalog.ns.truncate(1, 4)}
```
rdblue, line 97: Similar to my feedback on the Flink PR, this should not expose functions in every database. This should choose a database where Iceberg functions will be exposed. I recommend `system` since that's what we use for stored procedures.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+ * The various implementations are registered with the {@link org.apache.iceberg.spark.SparkCatalog}
+ * such that the function can be used as {@code truncate(col, width)} or {@code truncate(col, 2)}.
+ */
+public class TruncateFunction implements UnboundFunction {
+
+  public TruncateFunction() {
```
rdblue, line 52: Isn't this the default? You should be able to omit empty public constructors.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/SparkFunctions.java
```diff
@@ -0,0 +1,54 @@
...
+  private static final Map<String, UnboundFunction> FUNCTIONS = initFunctions();
+
+  private SparkFunctions() {
+  }
+
+  // The caller needs to validate that the namespace exists and is an Iceberg namespace when using these results.
+  //
+  // Functions that are added to all Iceberg catalogs and their associated namespaces get registered
+  // under Namespace.empty(), allowing the calling code to add the namespace.
+  public static String[] list() {
```
rdblue, line 38: Why does this return an array?

If it were a list, this could produce an `ImmutableList` and make that a static field, rather than getting the key set and converting it to an array every call.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+import org.apache.spark.sql.types.IntegerType;
+import org.apache.spark.sql.types.LongType;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.apache.spark.unsafe.types.UTF8String;
+
+/**
+ * Implementation of {@link UnboundFunction} that matches the {@code truncate} transformation.
+ * The various implementations are registered with the {@link org.apache.iceberg.spark.SparkCatalog}
+ * such that the function can be used as {@code truncate(col, width)} or {@code truncate(col, 2)}.
```
rdblue, line 48: I don't think this is accurate. Isn't this one unbound function returned by name, not "various implementations"?

Also, the function arguments aren't handled by the catalog lookup. Those are handled by `bind`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+
+  @Override
+  public BoundFunction bind(StructType inputType) {
+    if (inputType.fields().length != 2) {
+      throw new UnsupportedOperationException(
+          String.format("Invalid input type. Expected 2 fields but found %s", inputType.fields().length));
+    }
+
+    StructField fieldToTruncate = inputType.apply(0);
+    StructField truncationWidth = inputType.apply(1);
```
rdblue, line 81: The Spark [built-in `bucket` function](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/functions.scala#L4976-L4978) makes the width the first argument. I think we should probably stick with that convention for `bucket` and `truncate`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public boolean isResultNullable() {
+      return true;
+    }
+
+    @Override
+    public boolean isDeterministic() {
+      return true;
```
rdblue, line 131: `true` is the default for both `isDeterministic` and `isResultNullable`, so these implementations shouldn't be necessary.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+
+    DataType toTruncateSparkType = fieldToTruncate.dataType();
+    Type sourceIcebergType = SparkSchemaUtil.convert(fieldToTruncate.dataType());
+
+    validateSourceTypeUsingSparkTypes(fieldToTruncate.dataType());
+    validateTruncationWidthType(truncationWidth.dataType());
+
+    switch (sourceIcebergType.typeId()) {
+      case INTEGER:
+        return new TruncateInteger();
```
rdblue, line 91: I think this is going to cause problems. Both `tinyint` and `smallint` are mapped to Iceberg's `int`. It's fine to use the same bound function for all of them, but right now `TruncateInteger` assumes that the input type is an int and calls `getInt`, which is going to produce incorrect values when reading an `UnsafeRow` for the smaller types. It would also fail when using a `GenericRow`.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestSparkTruncateFunction.java
```diff
@@ -0,0 +1,246 @@
...
+
+  public TestSparkTruncateFunction(String catalogName, String implementation, Map<String, String> config) {
+    super(catalogName, implementation, config);
+    this.fullNamespace = ("spark_catalog".equals(catalogName) ? "" : catalogName + ".") + NS;
+  }
+
+  @Before
+  public void createNamespace() {
+    sql("CREATE NAMESPACE IF NOT EXISTS %s", fullNamespace);
+    sql("USE %s", fullNamespace);
```
rdblue, line 49: I don't think you need a namespace if you always reference `system.truncate`.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: When reading the documentation for `ScalarFunction`, it seemed like Spark wouls perform the required cast before invoking the magic function.

kbendick: https://spark.apache.org/docs/latest/api/java/index.html?org/apache/spark/sql/connector/catalog/functions/ScalarFunction.html

I might have that wrong though.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  public static class TruncateInteger extends TruncateBase<Integer> {
+    // magic method used in codegen
+    public static int invoke(int value, int width) {
+      if (width <= 0) {
+        throw new IllegalArgumentException(String.format("Invalid truncate width: %s (must be > 0)", width));
+      }
+
+      return value - (((value % width) + width) % width);
```
rdblue, line 142: I'd prefer to use the actual truncate function. I talked with Dan and heard that the motivation for this was that width might change and we don't want to create a new instance of the transform every time. I think it's good to catch that possibility, but there are other options that avoid copying the transform implementation.

The option I like best is to assume that the width is not going to change across calls and check that width is the same as the last row's width:

```java
  private int lastWidth = -1;
  private Transform<Integer, Integer> transform = null;
  public static int invoke(int width, int value) {
    ...

    if (lastWidth != width) {
      this.lastWidth = width;
      this.transform = Transforms.truncate(IntegerType.get(), width);
    }

    return transform.apply(value);
  }
```

For constant values, this will work great. There's some overhead for changing widths, but I don't think that's going to be a common case since it makes very little sense.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Oh actually you're right. I think I need to implement one for `tinyint` and `smallint` if we want to keep them using the magic funcitons.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+    public boolean isDeterministic() {
+      return true;
+    }
+  }
+
+  public static class TruncateInteger extends TruncateBase<Integer> {
+    // magic method used in codegen
+    public static int invoke(int value, int width) {
+      if (width <= 0) {
+        throw new IllegalArgumentException(String.format("Invalid truncate width: %s (must be > 0)", width));
```
rdblue, line 139: I probably wouldn't bother with this check.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Agreed that it's not likely to be a common case. That's not really the intended purpose of the function and I don't care for the code duplication.

Will go with that approach then. Thank you!

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public String canonicalName() {
+      return "org.apache.iceberg.spark.functions.truncate(int,int)";
+    }
+
+    @Override
+    public Integer produceResult(InternalRow input) {
+      return invoke(input.getInt(0), input.getInt(1));
```
rdblue, line 162: Does this need to handle null `value` args?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    @Override
+    public DataType resultType() {
+      return DataTypes.StringType;
+    }
+
+    @Override
+    public String canonicalName() {
+      return "org.apache.iceberg.spark.functions.truncate(string,int)";
```
rdblue, line 223: I'd probably make this `iceberg.truncate[width](string)`, but it's minor.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
rdblue: Actually, I'm reconsidering this. The option of having a static function that performs this would allow us to avoid boxing/unboxing and the lookup entirely. That's probably the better thing to do.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: It does, but since it has the magic function, it should already return `null` when it receives any `null` input.

I'm going to add counters to make sure the magic functions are the only ones being called though and then possibly remove `produceResult`.

kbendick: For now I'm updating it to handle `null` though 👍

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
kbendick: So the only problem with this is that within SQL (or even the dataframe API), the `SparkSessionCatalog` checks that the namespace exists.

So calling `system.truncate` results in:

kbendick: So the only problem with this is that within SQL (or even the dataframe API), the `SparkSessionCatalog` checks that the namespace exists.

So calling `system.truncate` results in
```
org.apache.spark.sql.AnalysisException: Undefined function: 'truncate'. This function is neither a registered temporary function nor a permanent function registered in the database 'system'.
```

I'm seeing if I can find a way around this, but it is part of the SessionCatalog's code.

kbendick: Full stack trace:

```
Undefined function: 'truncate'. This function is neither a registered temporary function nor a permanent function registered in the database 'system'.; line 1 pos 7
org.apache.spark.sql.AnalysisException: Undefined function: 'truncate'. This function is neither a registered temporary function nor a permanent function registered in the database 'system'.; line 1 pos 7
	at app//org.apache.spark.sql.catalyst.catalog.SessionCatalog.failFunctionLookup(SessionCatalog.scala:1561)
	at app//org.apache.spark.sql.catalyst.catalog.SessionCatalog.resolvePersistentFunctionInternal(SessionCatalog.scala:1704)
	at app//org.apache.spark.sql.catalyst.catalog.SessionCatalog.resolvePersistentFunction(SessionCatalog.scala:1673)
	at app//org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveFunctions$.org$apache$spark$sql$catalyst$analysis$Analyzer$ResolveFunctions$$resolveV1Function(Analyzer.scala:2172)
	at app//org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveFunctions$$anonfun$apply$25$$anonfun$applyOrElse$103.$anonfun$applyOrElse$109(Analyzer.scala:2122)
	at app//scala.Option.getOrElse(Option.scala:189)
	at app//org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveFunctions$$anonfun$apply$25$$anonfun$applyOrElse$103.$anonfun$applyOrElse$108(Analyzer.scala:2119)
```

Link to the code in the analyzer that's being hit: https://github.com/apache/spark/blob/f74867bddfbcdd4d08076db36851e88b15e66556/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L2118-L2126

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/SparkFunctions.java
kbendick: Made it an immutable list and converted to a static field.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Updated, though the more I think about it, the more I think I might throw this out.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Changed width to be the first argument.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Removed the unnecessary overrides.

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestSparkTruncateFunction.java
kbendick: That's correct. I'm still working on how to handle the `system` namespace.

I think probably we just register it as a special namespace within the `SparkSessionCatalog`? The analyzer specifically checks if the namspace exists when using the session catalog.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Should we just let the code fail? For some calls, it seems the code would do things that don't necessarily work (based on the existing transforms).

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Yeah I like that more. Updated.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Looking into this further, it seems that Spark will add a cast for us to Integer if we allow `smallint` or `tinyint` but only define a function for Integer type.

When calling `EXPLAIN EXTENDED SELECT system.truncate(2, -1S)`, there is a `(CAST -1 as int)` in the logical plan:

```
== Parsed Logical Plan ==
'Project [unresolvedalias('testhadoop.system.truncate(2, -1), None)]
+- OneRowRelation

== Analyzed Logical Plan ==
staticinvoke(class org.apache.iceberg.spark.functions.TruncateFunction$TruncateInteger, IntegerType, invoke, 2, -1, IntegerType, IntegerType, false, true, true): int
Project [staticinvoke(class org.apache.iceberg.spark.functions.TruncateFunction$TruncateInteger, IntegerType, invoke, 2, cast(-1 as int), IntegerType, IntegerType, false, true, true) AS staticinvoke(class org.apache.iceberg.spark.functions.TruncateFunction$TruncateInteger, IntegerType, invoke, 2, -1, IntegerType, IntegerType, false, true, true)#98]
+- OneRowRelation

== Optimized Logical Plan ==
Project [-2 AS staticinvoke(class org.apache.iceberg.spark.functions.TruncateFunction$TruncateInteger, IntegerType, invoke, 2, -1, IntegerType, IntegerType, false, true, true)#98]
+- OneRowRelation

== Physical Plan ==
*(1) Project [-2 AS staticinvoke(class org.apache.iceberg.spark.functions.TruncateFunction$TruncateInteger, IntegerType, invoke, 2, -1, IntegerType, IntegerType, false, true, true)#98]
+- *(1) Scan OneRowRelation[]
```

However, the resulting row type will still be `IntegerType`, so we should consider still defining them.

But I'm ok with allowing the cast for the width field from tinyint / smallint -> int as that doesn't affect the resulting input type and the static function is still called.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
rdblue: It won't fail because these are primitives.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
kbendick: We spoke offline and we're just giong to hold off on the spark session catalog for now.

Most likely in a follow up we will add a namespace `system` that will then be resolvable (as these issues go away when the namespace is resolvable). That matches other databases, as well as the experience of using the `ProcedureCatalog`.

I've removed or `@Ignore`d the tests against `SparkSessionCatalog` in this PR.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: I added `tinyint` and `smallint` support, so the return type would be correct.

For the width, as shown in the explain, Spark will cast it (and there is logic to ensure that width is only one of `ByteType` / `ShortType` / `IntType` aka `tinytint` / `smallint` / `int`).

spark/v3.3/spark/src/test/java/org/apache/iceberg/spark/sql/TestSparkTruncateFunction.java
kbendick: Removed the namespace from here as it's not needed.

As mentioned here https://github.com/apache/iceberg/pull/5305#discussion_r927178316, we will hold off on `SparkSessionCatalog` and then follow up with handling the need for `system` namespace in spark session catalog in another PR.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Buyt the width could be negative. The range for Integer is signed integer range. That's my concern (not so much that we'll get null for the width).

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Have updated all to handle null. Also have tests that check for calling truncate on `null` of each specific type (via casting).

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
rdblue: Yeah, if you use weird values, you get weird results.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  public abstract static class TruncateBase<T> implements ScalarFunction<T> {
+    @Override
+    public String name() {
+      return "truncate";
+    }
+  }
+
+  // tinyint
```
rdblue, line 138: Nit: this comment duplicates what is already in the class name.

kbendick: Thanks. I noticed that now that I renamed them. Moving the functions to be in a utility class and then this will be good to take a look at =)

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/BaseCatalog.java
```diff
@@ -46,4 +52,30 @@ public Procedure loadProcedure(Identifier ident) throws NoSuchProcedureException
...
+
+  @Override
+  public Identifier[] listFunctions(String[] namespace) throws NoSuchNamespaceException {
+    if (namespace.length == 1 && namespace[0].equalsIgnoreCase("system")) {
+      return SparkFunctions.list().stream()
+          .map(name -> Identifier.of(namespace, name))
+          .toArray(Identifier[]::new);
+    }
+
+    throw new NoSuchNamespaceException(namespace);
```
rdblue, line 64: This should return an empty list if the database exists.

kbendick: Fixed

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: I moved the truncation functions into `TruncateUtil` in api.

Some things already had utility functions handling their truncation, such as `UnicodeUtil.truncateString` so I didn't add functions for that.

The code in `TruncateUtil` explicitly assumes that the input has been validated, to allow for more efficient codegen or for skipping validation if it's already been done.
We pre-validate:
1. The Truncate `Transform` validates the `width` in their constructor so it doesn't need to be done again.
2. The Spark Truncate `BoundFunction` will immediately return `null` for `null` input for primitives.

Not doing duplicate validation was chosen because this code is in the hot path (namely because it gets included in Spark's codegen when used with `FunctionCatalog`).

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: I'll add a note in the JavaDoc and then remove that (as throwing exceptions in generated code is somewhat of an anti-pattern if possible).

api/src/main/java/org/apache/iceberg/transforms/Truncate.java
```diff
@@ -394,9 +395,7 @@ public ByteBuffer apply(ByteBuffer value) {
         return null;
       }
 
-      ByteBuffer ret = value.duplicate();
-      ret.limit(Math.min(value.limit(), value.position() + length));
-      return ret;
+      return TruncateUtil.truncateByteBuffer(length, value);
```
kbendick, line 398: There's `BinaryUtil#truncateBinary(ByteBuffer, int)`, but it doesn't exactly have the same code as is used here and I think the logic is different in that it allows for 0 length.

I can try to merge the two implementations in a follow up if we want.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/functions/TruncateFunction.java
kbendick: Extracted out a `TruncateUtil`, at least for the calls to be shared between `Truncate` transform and the generated code. As both `invoke` and the `TruncateUtil` are both static functions, will the Spark generated code remove the nested static function call?

kbendick:
@rdblue PTAL. I was going to tag other people as well.

kbendick:
@rdblue Do you think I should break this into a 2 PRs?

kbendick:
Looking at the storage partition joins, it looks like the function `bucket` will need to be resolvable on the empty namespace. I can update that here or a follow up:

https://github.com/apache/spark/blob/e9eb28e27d10497c8b36774609823f4bbd2c8500/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/V2ExpressionUtils.scala#L84-L100

aokolnychyi:
I'd love to take a look as well. I should have some time in a day. We also had some progress on bucketed joins internally.

kbendick:
> I'd love to take a look as well. I should have some time in a day. We also had some progress on bucketed joins internally.

Thanks Anton. Was going to tag you today now that it's cleaned up. Also cc @huaxingao @flyrain @nastra @Fokko

kbendick:
Right not we're requiring the call be to the `system` namespace, but the storage partitioned join implementation looks for a function called `bucket` in the `FunctionCatalog` using an empty array for the namespace per [this diff in the original Storage Partitioned Joins PR in `V2ExpressionUtils#toCatalystTransform`.](https://github.com/apache/spark/commit/20ffbf7b308c3dc90a49dbdcb8d7b972eeb53bc4#diff-715d0c2d59a4ddb8a4b5952c1b05be9f035b6d9b0d9670c70b58989dc722b252R79-R95)

So we should probably allow the empty namespace to resolve functions as well.

kbendick:
Link to the code that resolves out `bucket` implementation in case the link above doesn't resolve: https://github.com/apache/spark/blob/47f0303944abb11d3018186bc125113772eff8ef/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/V2ExpressionUtils.scala#L84-L100

kbendick:
Because this PR is so big, I'm going to separate out the `FunctionCatalog` implementation from `Truncate`.

I'm going to add a very simple function to be able to test it but keep the code to review a lot smaller. 👍

kbendick:
I've opened https://github.com/apache/iceberg/pull/5377 to cover _just_ the `FunctionCatalog`.

This PR is too big, and this way we can focus on just the `FunctionCatalog` business without having to worry about the details of `truncate`.

I've added an `iceberg_version` function in the other PR to assist with testing.

kbendick:
This PR is closed in favor of https://github.com/apache/iceberg/pull/5377 and https://github.com/apache/iceberg/pull/5411.

I'll open a PR for `Truncate` and link it shortly.

PR #5306, opened 2022-07-19
API: CloseableIterable.concat evaluates first item twice

nastra:
While working on Scan reporting (#5268) I noticed that this entire chain https://github.com/apache/iceberg/blob/eff879daa492641ac8bc829e7c4625dcdfac05fa/core/src/main/java/org/apache/iceberg/ManifestGroup.java#L218-L237 would be evaluated twice.
It turns out we eventually use `CloseableIterable.concat` with those
`CloseableIterables` in
https://github.com/apache/iceberg/blob/eff879daa492641ac8bc829e7c4625dcdfac05fa/core/src/main/java/org/apache/iceberg/ManifestGroup.java#L198.

When `CloseableIterable.concat` would check for emptyness via
```
if (!iterables.hasNext()) {
  return empty();
}
```
it would actually evaluate the first chain of Iterables twice.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
+++ b/api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
@@ -118,12 +118,7 @@ public interface CloseableIterable<T> extends Iterable<T>, Closeable {
   }
 
   static <E> CloseableIterable<E> concat(Iterable<CloseableIterable<E>> iterable) {
-    Iterator<CloseableIterable<E>> iterables = iterable.iterator();
-    if (!iterables.hasNext()) {
-      return empty();
-    } else {
-      return new ConcatCloseableIterable<>(iterable);
-    }
+    return new ConcatCloseableIterable<>(iterable);
   }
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/apache/iceberg/io/CloseableIterable.java
```diff
@@ -148,8 +143,6 @@ public CloseableIterator<E> iterator() {
 
       private ConcatCloseableIterator(Iterable<CloseableIterable<E>> inputs) {
         this.iterables = inputs.iterator();
-        this.currentIterable = iterables.next();
-        this.currentIterator = currentIterable.iterator();
```
rdblue, line 152: Why was it necessary to change how these are initialized? Isn't the problem that `iterator` is called twice on the `iterable` passed to `concat`? If that is avoided, then I think this can stay the same.

nastra: The init code here already implied that there would be at least one element, so this was necessary to make the code work even with empty iterables

rdblue (approved):

rdblue:
Thanks, @nastra!

Issue #5308, opened 2022-07-19
ManifestGroup entries() reads the manifest 2 times when ignoreDeleted is set. This causes perf issues.


palaniappa:
ManifestGroup entries() reads the manifest 2 times when ignoreDeleted is set. Since every manifest is read 2 times, if the table contains more manifest the planFiles time keeps increasing.

nastra:
I wonder if this is related to what's being addressed in https://github.com/apache/iceberg/pull/5306

PR #5309, opened 2022-07-19
#5308 - Avoid 2 reads on manifest

palaniappa:


palaniappa:
@rdblue can you please take a look at this small change ?

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -249,9 +249,11 @@ public CloseableIterator<T> iterator() {
                 .caseSensitive(caseSensitive)
                 .select(columns);
 
-            CloseableIterable<ManifestEntry<DataFile>> entries = reader.entries();
+            CloseableIterable<ManifestEntry<DataFile>> entries = null;
```
rdblue, line 252: You can omit the `= null` in this case because it is set in all control flow branches below.

core/src/main/java/org/apache/iceberg/ManifestGroup.java
```diff
@@ -249,9 +249,11 @@ public CloseableIterator<T> iterator() {
...
             if (ignoreDeleted) {
               entries = reader.liveEntries();
+            } else {
+              entries = reader.entries();
```
rdblue, line 256: Good catch!

rdblue (approved):
I think this looks good, but I'd prefer to remove the useless default value.

rdblue (approved):

Issue #5310, opened 2022-07-19
Flink - Data File completed and added to Manifest but data file NotFound in S3 location


abmo-x:
We have seen multi occurrences of this issue in our Flink pipeline where Iceberg Manifest has datafile which was never uploaded to S3.

This could be related to https://github.com/apache/iceberg/issues/4168

```
logger_name:  org.apache.iceberg.aws.s3.S3OutputStream 
    message:  S3OutputStream initialized with S3_PATH_TO_FILE_THAT_DOESN'T_EXIST

2022-07-15T15:04:38.220+00:00
Iceberg.flink.sink.IcebergFilesCommitter
      message:  Start to flush snapshot state to state backend,

2022-07-15T15:04:08.447+00:00 
logger_name:  org.apache.iceberg.io.BaseTaskWriter 
Complete data file GenericDataFile{content=data, file_path=PATH_TO_FILE_THAT_DOESN'T_EXIST file_format=PARQUET, spec_id=0, partition=PartitionData{date=2022-07-15, hour=14}
```

We have seen this for 5-6 data files in various partitions. One common pattern for all these files I found is that snapshot was triggered which triggered the Data File to be completed. But looks like the file never gets uploaded as I don't see any S3 PutObject or multi part upload requests for these data files.

Issue #4168, opened 2022-02-18
Flink S3Fileio incorrect commit


krisdas:
Flink version : 1.13.2
Iceberg version : 0.12
We were trying S3Fileio instead of default hadoop fileio, to ingest data into iceberg on S3.

While commit/saving checkpoint, we see below call stack while uploading data file. So looks datafile (for example : s3/bucket/.../iceberg_db.db/iceberg_table/data/file.parquet) upload fails.

```
exception: { [-] 
      exception_class:  java.util.concurrent.CompletionException 
      exception_message:  java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /tmp/s3fileio-224113294985193160.tmp 
      stacktrace:  java.util.concurrent.CompletionException: java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /tmp/s3fileio-224113294985193160.tmp
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1702)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /tmp/s3fileio-224113294985193160.tmp
	at software.amazon.awssdk.utils.FunctionalUtils.asRuntimeException(FunctionalUtils.java:180)
	at software.amazon.awssdk.utils.FunctionalUtils.lambda$safeSupplier$4(FunctionalUtils.java:110)
	at software.amazon.awssdk.utils.FunctionalUtils.invokeSafely(FunctionalUtils.java:136)
	at software.amazon.awssdk.core.sync.RequestBody.fromFile(RequestBody.java:88)
	at software.amazon.awssdk.core.sync.RequestBody.fromFile(RequestBody.java:99)
	at org.apache.iceberg.aws.s3.S3OutputStream.lambda$uploadParts$1(S3OutputStream.java:237)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
[...]

PR #1749, opened 2020-11-11
AWS: fix bugs around using S3FileIO for table operations

jackye1995:
When used in `BaseMetastoreTableOperations`, in `TableMetadataParser.internalWrite()`, an output stream is closed twice because of the closure of the output stream and its writer:

```
try (OutputStream ou = isGzip ? new GZIPOutputStream(stream) : stream;
         OutputStreamWriter writer = new OutputStreamWriter(ou, StandardCharsets.UTF_8)) {
   ...
}
```

This causes issue in `S3OutputStream` because the stream uploads the file from local to S3 during the call to close.

Also, `S3FileIO` cannot be properly initialized when there are multiple HTTP client builders available in the Java class path at runtime, so `AwsClientUtil` is used to provide a stable default client for `S3FileIO` to use.

PR #5311, opened 2022-07-20
S3OutputStream - failure to close should persist on subsequent close calls

abmo-x:
Fix for https://github.com/apache/iceberg/issues/5310 and https://github.com/apache/iceberg/issues/4168

**Issue**

When S3OutputStream fails to upload a file successfully on call to **close** due to some failure, IcebergStreamWriter in Flink still ends up adding the file to completedDataFiles from **BaseTaskWriter** resulting in table metadata pointing to a s3 data file which was never uploaded to s3. 

**Steps to Reproduce**

-  Flink 1.14 pipeline with Iceberg 0.13
-  Customer implemented **ProcessFunction<FlinkRecord, Row>** function with catch all exceptions in **processElement**
      - This is important as this is what leads to close() called twice from:
         - [shouldRollToNewFile](https://github.com/apache/iceberg/blob/bf6242fe57605a7b38b9d01ee33ae325687fb3a5/core/src/main/java/org/apache/iceberg/io/BaseTaskWriter.java#L257) --> [closeCurrent](https://github.com/apache/iceberg/blob/bf6242fe57605a7b38b9d01ee33ae325687fb3a5/core/src/main/java/org/apache/iceberg/io/BaseTaskWriter.java#L288)
         - [close](https://github.com/apache/iceberg/blob/bf6242fe57605a7b38b9d01ee33ae325687fb3a5/core/src/main/java/org/apache/iceberg/io/BaseTaskWriter.java#L309)

- configure pipeline to use S3FileIO and file size according to your test data so that the file will roll to new file 
- S3 failure on putObject(should be reproducible for MultiPartUpload as well) call to  [shouldRollToNewFile](https://github.com/apache/iceberg/blob/bf6242fe57605a7b38b9d01ee33ae325687fb3a5/core/src/main/java/org/apache/iceberg/io/BaseTaskWriter.java#L257) which calls [close --> completeUploads](https://github.com/apache/iceberg/blob/c8f93dfad1925c7192d2d01bbfd8b790b364b54e/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java#L242) 
    
<details>
  <summary> StackTrace from failure</summary>
<p>

```
org.apache.flink.streaming.runtime.tasks.ExceptionInChainedOperatorException: Could not forward element to next operator
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.pushToOperator(ChainingOutput.java:101)
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.collect(ChainingOutput.java:80)
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.collect(ChainingOutput.java:39)
	at org.apache.flink.streaming.api.operators.CountingOutput.collect(CountingOutput.java:56)
	at org.apache.flink.streaming.api.operators.CountingOutput.collect(CountingOutput.java:29)
	at org.apache.flink.streaming.api.operators.TimestampedCollector.collect(TimestampedCollector.java:51)
	...
	at org.apache.flink.streaming.api.operators.ProcessOperator.processElement(ProcessOperator.java:66)
	at org.apache.flink.streaming.runtime.tasks.OneInputStreamTask$StreamTaskNetworkOutput.emitRecord(OneInputStreamTask.java:233)
	at org.apache.flink.streaming.runtime.io.AbstractStreamTaskNetworkInput.processElement(AbstractStreamTaskNetworkInput.java:134)
	at org.apache.flink.streaming.runtime.io.AbstractStreamTaskNetworkInput.emitNext(AbstractStreamTaskNetworkInput.java:105)
	at org.apache.flink.streaming.runtime.io.StreamOneInputProcessor.processInput(StreamOneInputProcessor.java:65)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.processInput(StreamTask.java:496)
	at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxLoop(MailboxProcessor.java:203)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.runMailboxLoop(StreamTask.java:809)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:761)
	at org.apache.flink.runtime.taskmanager.Task.runWithSystemExitMonitoring(Task.java:958)
	at org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke(Task.java:937)
	at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:766)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:575)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: org.apache.flink.streaming.runtime.tasks.ExceptionInChainedOperatorException: Could not forward element to next operator
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.pushToOperator(ChainingOutput.java:101)
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.collect(ChainingOutput.java:80)
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.collect(ChainingOutput.java:39)
	at org.apache.flink.streaming.api.operators.CountingOutput.collect(CountingOutput.java:56)
	at org.apache.flink.streaming.api.operators.CountingOutput.collect(CountingOutput.java:29)
	at org.apache.flink.streaming.api.operators.StreamMap.processElement(StreamMap.java:38)
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.pushToOperator(ChainingOutput.java:99)
	... 21 more
Caused by: software.amazon.awssdk.core.exception.SdkClientException: Unable to load credentials from service endpoint.
	at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:98)
	at software.amazon.awssdk.auth.credentials.HttpCredentialsProvider.refreshCredentials(HttpCredentialsProvider.java:110)
	at software.amazon.awssdk.utils.cache.CachedSupplier.refreshCache(CachedSupplier.java:132)
	at software.amazon.awssdk.utils.cache.OneCallerBlocks.prefetch(OneCallerBlocks.java:38)
	at software.amazon.awssdk.utils.cache.CachedSupplier.prefetchCache(CachedSupplier.java:116)
	at software.amazon.awssdk.utils.cache.CachedSupplier.get(CachedSupplier.java:91)
	at java.base/java.util.Optional.map(Optional.java:265)
	at software.amazon.awssdk.auth.credentials.HttpCredentialsProvider.resolveCredentials(HttpCredentialsProvider.java:146)
	at software.amazon.awssdk.auth.credentials.AwsCredentialsProviderChain.resolveCredentials(AwsCredentialsProviderChain.java:85)
	at software.amazon.awssdk.auth.credentials.internal.LazyAwsCredentialsProvider.resolveCredentials(LazyAwsCredentialsProvider.java:45)
	at software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider.resolveCredentials(DefaultCredentialsProvider.java:104)
	at software.amazon.awssdk.awscore.client.handler.AwsClientHandlerUtils.createExecutionContext(AwsClientHandlerUtils.java:76)
	at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.createExecutionContext(AwsSyncClientHandler.java:68)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.lambda$execute$1(BaseSyncClientHandler.java:97)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.measureApiCallSuccess(BaseSyncClientHandler.java:167)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:94)
	at software.amazon.awssdk.core.client.handler.SdkSyncClientHandler.execute(SdkSyncClientHandler.java:45)
	at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:55)
	at software.amazon.awssdk.services.s3.DefaultS3Client.putObject(DefaultS3Client.java:8350)
	at org.apache.iceberg.aws.s3.S3OutputStream.completeUploads(S3OutputStream.java:396)
	at org.apache.iceberg.aws.s3.S3OutputStream.close(S3OutputStream.java:256)
	at org.apache.parquet.io.DelegatingPositionOutputStream.close(DelegatingPositionOutputStream.java:38)
	at org.apache.parquet.hadoop.ParquetFileWriter.end(ParquetFileWriter.java:1106)
	at org.apache.iceberg.parquet.ParquetWriter.close(ParquetWriter.java:239)
	at org.apache.iceberg.io.DataWriter.close(DataWriter.java:82)
	at org.apache.iceberg.io.BaseTaskWriter$BaseRollingWriter.closeCurrent(BaseTaskWriter.java:288)
	at org.apache.iceberg.io.BaseTaskWriter$BaseRollingWriter.write(BaseTaskWriter.java:254)
	at org.apache.iceberg.io.PartitionedFanoutWriter.write(PartitionedFanoutWriter.java:58)
	at org.apache.iceberg.flink.sink.IcebergStreamWriter.processElement(IcebergStreamWriter.java:74)
	at org.apache.flink.streaming.runtime.tasks.ChainingOutput.pushToOperator(ChainingOutput.java:99)
	... 27 more
Caused by: software.amazon.awssdk.core.exception.SdkServiceException: Unauthorized
```
</p>
</details>

- Pipeline should keep running even on above failure, then [snapshot barrier](https://github.com/apache/iceberg/blob/3584c79022ec70f79b326550736b4600d249e4a2/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/sink/IcebergStreamWriter.java#L65) gets triggered 
  - This calls close and ends up adding the datafile which was never uploaded to S3

**Testing**
- Unit tests added
- Testing on our dev pipeline, will update the results after the pipeline runs for a little bit


core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
```diff
@@ -0,0 +1,97 @@
...
+        Mockito.when(appender.length()).thenReturn(1L);
+        Mockito.when(appender.metrics()).thenReturn(Mockito.mock(Metrics.class));
+        Mockito.when(appender.splitOffsets()).thenReturn(Arrays.asList(1L));
+        dataWriter.close();
+        DataFile dataFile = dataWriter.toDataFile();
+        Assert.assertEquals(file.getPath(), dataFile.path());
+    }
+
+    @Test(expected=IllegalStateException.class)
+    public void closeFailureShouldNotReturnDataFile() throws IOException {
```
RussellSpitzer, line 80: 👍

aws/src/test/java/org/apache/iceberg/aws/s3/TestS3OutputStream.java
```diff
@@ -27,21 +27,16 @@
 import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.Comparator;
-import java.util.List;
-import java.util.Random;
```
RussellSpitzer, line 32: no wildcard imports

abmo-x: cleaned them up.

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
```diff
@@ -240,6 +242,23 @@ private void newStream() throws IOException {
 
   @Override
   public void close() throws IOException {
+
+    /**
+     * As calls to close() are used to mark a stream as completed and add the file
+     * to the completed data files in BaseTaskWriter we cannot just return if this
```
RussellSpitzer, line 248: comma between "BaseTaskWriter" and "we cannot"

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
```diff
@@ -240,6 +242,23 @@ private void newStream() throws IOException {
...
+     * was already closed with failure to upload.
+     *
+     * As close 'completes' the upload and staging files are cleanup, any failure to
+     * upload should be persisted and thrown back to the caller to make sure
+     * the underlying file represented by this stream is not used in the completedDataFiles
+     * of a manifest.
+     * see: https://github.com/apache/iceberg/issues/5310
+      */
+    if(closeFailureState != null){
+      throw new IllegalStateException("close called on a closed stream which failed to close completely earlier at: "+closeFailureState.getFailureAt()
```
RussellSpitzer, line 258: Few style errors on this line, Checkstyle will mark them as errors though. (Missing space between + and string, "comma after a newline"). Instructions on running our current checkstyle rules are in the dev docs

RussellSpitzer (commented):
Is it feasible for us to fix this in BaseTaskWriter? It sounds like a failed "close" call should just stop the commit process. It feels like a bit of a workaround to have subsequent "close" calls throw an exception when we probably shouldn't be making subsequent "close" calls

I could be misunderstanding this though

rdblue:
I agree with @RussellSpitzer, I think we should avoid the double close, since that is what is causing the problem (at least as far as I understand).

core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
```diff
@@ -0,0 +1,97 @@
...
+        Mockito.doThrow(new IllegalStateException("mock close failure of appender"))
+                .when(appender).close();
+        try {
+            dataWriter.close();
+        } catch (Exception e) {
+            // ignore mocked failure on first call to close
+        }
+        dataWriter.close();
+        dataWriter.toDataFile();
```
krisdas, line 89: L89 : will this line execute? line 88 will always throw exception, right?

abmo-x:
@rdblue @RussellSpitzer 
Added a commit to clear currentWriter on close in BaseTaskWriter and added 2 test cases around failure to close and complete.

I agree close should be only called once and we are relying on that behavior quite strongly and adding the data files. 
However I have found the writers are held and closed more than once in various scenarios which causes this issue where a close resulted in failure and writers were in a bad state.

1. when user defined functions catch all exceptions and ignore failures on write as seen in Flink's processElement which internally triggers a roll to new file.
2. This behavior was also observed before and fix was made in https://github.com/apache/iceberg/pull/1749 

Let me know your thoughts.

core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
abmo-x: yes, L89 will execute and we are verifying that here with expectation that call on L89 will throw an exception as expected.

abmo-x:
reverted changes to S3OutputStream to keep close api consistent

core/src/main/java/org/apache/iceberg/io/DataWriter.java
```diff
@@ -78,7 +80,14 @@ public long length() {
...
   public void close() throws IOException {
-    if (dataFile == null) {
+    /* As close is called couple times from BaseTaskWriter
+      on rollToNewFile and close, we don't want to create a data file
+      if appender.close was already called and failed due to an internal close failure.
+      Subsequent call to close after failure should not create a data file and keep it null.
+      We have observed appender.close failed when S3OutputStream.close fails and subsequent call
+      from flink tries to close the writer and use the data file which was never uploaded.
+    */
+    if (closed.compareAndSet(false, true)) {
```
RussellSpitzer, line 90: Do we need a new variable here or can we make dataFile into an Atomic reference?

core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
```diff
@@ -0,0 +1,105 @@
...
+    dataWriter.close();
+    DataFile dataFile = dataWriter.toDataFile();
+    Assert.assertEquals(file.getPath(), dataFile.path());
+  }
+
+  @Test
+  public void closeFailureShouldNotReturnDataFile() throws IOException {
+    Mockito.doThrow(new IllegalStateException("mock close failure of appender"))
+        .when(appender).close();
+    Assertions.assertThatThrownBy(() -> {
```
stevenzwu, line 89: nit: it is unclear whether exception is expected from `dataWriter.close()` or `dataWriter.toDataFile()`

core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
```diff
@@ -0,0 +1,105 @@
...
+      } catch (Exception e) {
+        // ignore mocked failure on first call to close
+      }
+      dataWriter.close();
+      dataWriter.toDataFile();
+    }).isInstanceOf(IllegalStateException.class);
+  }
+
+  @Test
+  public void toDataFile() {
```
stevenzwu, line 101: nit: maybe make the test method name more informative. is `toDataFileOnEmptyWriterShouldFail` accurate?

core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
abmo-x: reverted these changes.

core/src/test/java/org/apache/iceberg/io/TestDataWriter.java
abmo-x: reverted these changes.

abmo-x:
After further discussion with @RussellSpitzer, brought back my changes to S3OutputStream.

As failure to close a S3 stream leaves it in a bad state which cannot be recovered, any future calls to that stream should continue to fail. Changes now are simple and just in S3OutputStream.

cc @rdblue

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
```diff
@@ -240,6 +241,13 @@ private void newStream() throws IOException {
 
   @Override
   public void close() throws IOException {
+
+    // A failed s3 close removes state that is required for a successful close.
+    if (closeFailureException != null) {
+      throw new IOException(
+          "Attempted to close a S3 stream that failed to close earlier", closeFailureException);
```
RussellSpitzer, line 248: @rdblue We had a discussion about this offline and I think this is probably the only safe thing to do. While double-closing shouldn't throw an exception, if the close failed that means that any future close is also a failure. The S3OutputStream loses it's internal state on close which means that a subsequent close call can never succeed if the IOException is handled.

Because we can't ever retry successfully we decided that the stream should keep throwing exceptions to allow a caller to not falsely believe that a retry close had succeeded when the original close failed.

stevenzwu: This makes sense to me.

Nit: the error msg doesn't read smoothly. what about `Attempted to close a S3 output stream that has earlier close failed already`.

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
```diff
@@ -257,6 +258,13 @@ private void newStream() throws IOException {
 
   @Override
   public void close() throws IOException {
+
+    // A failed s3 close removes state that is required for a successful close.
```
stevenzwu, line 262: nit: wondering if the following comment is a little more clear based on Russel's comment

```
A previously failed close left the S3 output stream in a bad state that is non-recoverable.
Any future close on this stream should fail.
```

stevenzwu (approved):
This change looks good to me. It is defensive for more robustness.

stevenzwu:
@abmo-x Based on the description, it seems to me that application shouldn't have caught and swallowed all exceptions in the process function. I know it tries to catch "bad" record and route to a dead-letter table. It probably should just catch and handle the serialization specific exceptions.

abmo-x:
> @abmo-x Based on the description, it seems to me that application shouldn't have caught and swallowed all exceptions in the process function. I know it tries to catch "bad" record and route to a dead-letter table. It probably should just catch and handle the serialization specific exceptions.

Yes, the application code will be updated to handle exceptions correctly as well. As there are [other cases](https://github.com/apache/iceberg/pull/1749) where the S3 output stream can be still used after its closed I think this defensive code helps enforce bad data files are not added to metadata.

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
abmo-x: Thanks for the suggestion @stevenzwu, made few changes based on the suggestion.

aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
abmo-x: updated based on suggestion.

abmo-x (commented):
updated comments based on recommendations.

RussellSpitzer (approved):

Issue #5218, opened 2022-07-07
Support the bloom filter for ORC formats


a49a:
Since this Parquet PR merged https://github.com/apache/iceberg/pull/5035 , we can also support ORC.

We could support the following options.
`orc.bloom.filter.columns`
`orc.default.bloom.fpp`

Should we change `orc.bloom.filter.columns` to `write.orc.bloom.filter.columns` for aligning this Parquet PR  https://github.com/apache/iceberg/pull/5035 ?

deadwind4:
I will take this

PR #5313, opened 2022-07-20
Orc: Support row group bloom filters

deadwind4:
Issue link: https://github.com/apache/iceberg/issues/5218

Add `write.orc.bloom.filter.columns` and `write.orc.bloom.filter.fpp` options.
Enable these options in ORC class.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -177,6 +177,10 @@ public class TableProperties {
   public static final String AVRO_COMPRESSION_LEVEL_DEFAULT = null;
 
   public static final String ORC_STRIPE_SIZE_BYTES = "write.orc.stripe-size-bytes";
+
+  public static final String ORC_BLOOM_FILTER_COLUMNS = "write.orc.bloom.filter.columns";
+  public static final String ORC_BLOOM_FILTER_FPP = "write.orc.bloom.filter.fpp";
+
   public static final String DELETE_ORC_STRIPE_SIZE_BYTES = "write.delete.orc.stripe-size-bytes";
   public static final long ORC_STRIPE_SIZE_BYTES_DEFAULT = 64L * 1024 * 1024; // 64 MB
 
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -179,7 +179,9 @@ public class TableProperties {
   public static final String ORC_STRIPE_SIZE_BYTES = "write.orc.stripe-size-bytes";
 
   public static final String ORC_BLOOM_FILTER_COLUMNS = "write.orc.bloom.filter.columns";
+
   public static final String ORC_BLOOM_FILTER_FPP = "write.orc.bloom.filter.fpp";
+  public static final double BLOOM_FILTER_FPP_DEFAULT = 0.05;
 
   public static final String DELETE_ORC_STRIPE_SIZE_BYTES = "write.delete.orc.stripe-size-bytes";
   public static final long ORC_STRIPE_SIZE_BYTES_DEFAULT = 64L * 1024 * 1024; // 64 MB
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -179,9 +179,10 @@ public class TableProperties {
   public static final String ORC_STRIPE_SIZE_BYTES = "write.orc.stripe-size-bytes";
 
   public static final String ORC_BLOOM_FILTER_COLUMNS = "write.orc.bloom.filter.columns";
+  public static final String ORC_BLOOM_FILTER_COLUMNS_DEFAULT = "";
 
   public static final String ORC_BLOOM_FILTER_FPP = "write.orc.bloom.filter.fpp";
-  public static final double BLOOM_FILTER_FPP_DEFAULT = 0.05;
+  public static final double ORC_BLOOM_FILTER_FPP_DEFAULT = 0.05;
 
   public static final String DELETE_ORC_STRIPE_SIZE_BYTES = "write.delete.orc.stripe-size-bytes";
   public static final long ORC_STRIPE_SIZE_BYTES_DEFAULT = 64L * 1024 * 1024; // 64 MB
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,6 +64,8 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.orc.block-size-bytes         | 268435456 (256 MB) | Define the default file system block size for ORC files |
 | write.orc.compression-codec        | zlib               | ORC compression codec: zstd, lz4, lzo, zlib, snappy, none |
 | write.orc.compression-strategy     | speed              | ORC compression strategy: speed, compression |
+| write.orc.bloom.filter.columns     | (not set)          | Comma separated list of column names for which a Bloom filter must be created |
+| write.orc.bloom.filter.fpp         | 0.05               | False positive probability for Bloom filter (must > 0.0 and < 1.0) |
 | write.location-provider.impl       | null               | Optional custom implementation for LocationProvider  |
 | write.metadata.compression-codec   | none               | Metadata compression codec; none or gzip           |
 | write.metadata.metrics.default     | truncate(16)       | Default metrics mode for all columns in the table; none, counts, truncate(length), or full |
--- a/orc/src/main/java/org/apache/iceberg/orc/ORC.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORC.java
@@ -315,6 +315,9 @@ public class ORC {
                 config, OrcConf.BLOOM_FILTER_FPP.getAttribute(), BLOOM_FILTER_FPP_DEFAULT);
         bloomFilterFpp =
             PropertyUtil.propertyAsDouble(config, ORC_BLOOM_FILTER_FPP, bloomFilterFpp);
+        Preconditions.checkArgument(
+            bloomFilterFpp > 0.0 && bloomFilterFpp < 1.0,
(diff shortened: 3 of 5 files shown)


deadwind4:
cc @rdblue @RussellSpitzer @huaxingao @kbendick  Please review it, thanks a lot.

docs/configuration.md
```diff
@@ -64,6 +64,8 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.orc.block-size-bytes         | 268435456 (256 MB) | Define the default file system block size for ORC files |
 | write.orc.compression-codec        | zlib               | ORC compression codec: zstd, lz4, lzo, zlib, snappy, none |
 | write.orc.compression-strategy     | speed              | ORC compression strategy: speed, compression |
+| write.orc.bloom.filter.columns     | (not set)          | Comma separated list of column names for which a Bloom filter must be created |
+| write.orc.bloom.filter.fpp         | 0.05               | False positive probability for Bloom filter (must > 0.0 and < 1.0) |
```
kbendick, line 68: You might want to match the parquet configurations a bit more closely.

They are
```
| write.parquet.bloom-filter-enabled.column.col1          | (not set) | Enables writing a bloom filter for the column: col1|
| write.parquet.bloom-filter-max-bytes | 1048576 (1 MB)   | The maximum number of bytes for a bloom filter bitset |
```

So you could do `write.orc.bloom-filter-enabled.column.col1`. This also matches other config value formatting that's per column, such as `write.metadata.metrics.column.col1`.

For the `fpp`, as that seems to be how the value is set on the ORC bloom filter, I would suggest keeping it that way. But if the parquet implementation is translating from the max-bytes to fpp, then possibly setting the config that way for consistency (but I doubt that it is).

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -220,6 +224,10 @@ private static class Context {
       private final CompressionKind compressionKind;
       private final CompressionStrategy compressionStrategy;
 
+      private final String bloomFilterColumns;
+
+      private final double bloomFilterFpp;
```
huaxingao, line 229: nit: shall we remove the blank lines in L226 and L228 so `bloomFilterColumns` and `bloomFilterFpp` will be grouped together with the rest of the properties?

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -285,9 +305,24 @@ static Context dataContext(Map<String, String> config) {
         strategyAsString =
             PropertyUtil.propertyAsString(config, ORC_COMPRESSION_STRATEGY, strategyAsString);
         CompressionStrategy compressionStrategy = toCompressionStrategy(strategyAsString);
+        String bloomFilterColumns =
```
huaxingao, line 308: nit: add a blank line before this?

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -285,9 +305,24 @@ static Context dataContext(Map<String, String> config) {
...
+            PropertyUtil.propertyAsString(config, OrcConf.BLOOM_FILTER_COLUMNS.getAttribute(), "");
+        bloomFilterColumns =
+            PropertyUtil.propertyAsString(config, ORC_BLOOM_FILTER_COLUMNS, bloomFilterColumns);
+
+        double bloomFilterFpp =
+            PropertyUtil.propertyAsDouble(config, OrcConf.BLOOM_FILTER_FPP.getAttribute(), 0.05);
```
huaxingao, line 314: Is `0.05` the default fpp for Orc? Can we define a constant for this `0.05`?

huaxingao:
For the test, shall we also check the read path to make sure bloom filters are there?

orc/src/main/java/org/apache/iceberg/orc/ORC.java
kbendick: +1 to that. None of the other private fields have blank lines.

deadwind4:
> For the test, shall we also check the read path to make sure bloom filters are there?

I find the ORC SDK will auto recognize the bloom filter in the following code. Is it necessary to test the logic of ORC itself? IMO, we don't have to test it. @huaxingao 

https://github.com/apache/orc/blob/a49f87d492aa62ec2be2d4bce2fcfe1f53ca05d9/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L890-L912

docs/configuration.md
deadwind4: IMO, in this PR, we align `write.parquet.bloom-filter-enabled.column` and `write.orc.bloom.filter.columns` with native parquet and orc options. 

We can create a new issue to unify these options. 
In the new issue, we discuss a new option of no parquet and orc tag like `write.rowgroup.bloom-filter.columns` or `write.rowgroup.bloom-filter-enabled.column` to unify file format options. 
Maybe ·write.parquet.bloom-filter-enabled.column = id, name· or `[id, name]` is more elegant than `write.rowgroup.bloom-filter.columns.id = true`.

PS: 
I find the Trino iceberg connector use the style like `orc_bloom_filter_columns = [id, name]`
https://trino.io/docs/current/connector/iceberg.html#iceberg-table-properties

I'm happy to diver this.
@kbendick What do you think?

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -286,8 +305,25 @@ static Context dataContext(Map<String, String> config) {
...
+        String bloomFilterColumns =
+            PropertyUtil.propertyAsString(config, OrcConf.BLOOM_FILTER_COLUMNS.getAttribute(), "");
+        bloomFilterColumns =
+            PropertyUtil.propertyAsString(config, ORC_BLOOM_FILTER_COLUMNS, bloomFilterColumns);
+
+        double bloomFilterFpp =
+            PropertyUtil.propertyAsDouble(
+                config, OrcConf.BLOOM_FILTER_FPP.getAttribute(), BLOOM_FILTER_FPP_DEFAULT);
+        bloomFilterFpp =
+            PropertyUtil.propertyAsDouble(config, ORC_BLOOM_FILTER_FPP, bloomFilterFpp);
```
huaxingao, line 317: nit: Shall we add a `Preconditions.checkArgument` to make sure the `bloomFilterFpp` is `> 0.0 and < 1.0`?

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -177,6 +177,12 @@ private TableProperties() {}
   public static final String AVRO_COMPRESSION_LEVEL_DEFAULT = null;
 
   public static final String ORC_STRIPE_SIZE_BYTES = "write.orc.stripe-size-bytes";
+
+  public static final String ORC_BLOOM_FILTER_COLUMNS = "write.orc.bloom.filter.columns";
+
+  public static final String ORC_BLOOM_FILTER_FPP = "write.orc.bloom.filter.fpp";
+  public static final double BLOOM_FILTER_FPP_DEFAULT = 0.05;
```
huaxingao, line 184: nit: `BLOOM_FILTER_FPP_DEFAULT` => `ORC_BLOOM_FILTER_FPP_DEFAULT`?

huaxingao:
LGTM. Thanks a lot @deadwind4 for working on this!
cc @kbendick @rdblue

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -177,6 +177,12 @@ private TableProperties() {}
   public static final String AVRO_COMPRESSION_LEVEL_DEFAULT = null;
 
   public static final String ORC_STRIPE_SIZE_BYTES = "write.orc.stripe-size-bytes";
+
+  public static final String ORC_BLOOM_FILTER_COLUMNS = "write.orc.bloom.filter.columns";
```
kbendick, line 181: Nit: Ideally there would be a default value, `ORC_BLOOM_FILTER_COLUMNS_DEFAULT`. Even if it’s null (like Avro compression level default) or empty string, I think it’s better for readability and safe refactoring later to have the default defined where the rest are.

core/src/main/java/org/apache/iceberg/TableProperties.java
kbendick: +1. Other configs are format specific for their defaults, and parquet doesn’t use this configuration which could cause confusion.

docs/configuration.md
kbendick: I’m good with following up on anything in a new issue.

This is almost there and we should leave it as is (minus any nits). Using configs similar to those offered by native file formats is a good call imo.

kbendick (approved):
This LGTM, outside of the two nits on the config values.

Thank you for working on this @deadwind4!

deadwind4:
Thank a lot @huaxingao @kbendick for reviewing this.

deadwind4:
@kbendick Could we merge it?

pvary:
@deadwind4: I understand, that the ORC supports these filters and tests the functions, but I would still like to see some tests validating, that the filters are there. I could envision a situation where ORC changes the way to support bloom filters, and our tests fails to recognize this.
If, and only if, there is an easy way to check it, then it would be good to have a test here as well to validate that the filters are created. 

Thanks, Peter

deadwind4:
@pvary Thank you for your feedback. I have added a test case to validate bloom filters in ORC files and push a new commit. 
@kbendick @pvary Please review this. Thanks a lot.

pvary:
@deadwind4: Could you please check the failed tests.

Thanks,
Peter

deadwind4:
@pvary All checks have passed after I rebased the code and reran the CI.

pvary:
Thanks @deadwind4 for the PR and @kbendick for the review!

chenwyi2:
ORC_BLOOM_FILTER_COLUMNS this property will work on spark? when i set write.orc.bloom.filter.columns=xx and used spark to write data, i found that bloomfilter had no effect on querying.

PR #5316, opened 2022-07-20
AWS: Remove sorting of parts in multi-part upload

amogh-jahagirdar:
Currently in the MPU logic, the parts are sorted when completing the upload, which is not necessary. Sorting can be removed; this is more to simplify the code a bit, from a performance perspective sorting is likely negligible in most practical cases. Assume a 1 GB object and 32 mb part sizes, it's sorting only 32 elements, which is negligible

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
@@ -329,7 +329,6 @@ class S3OutputStream extends PositionOutputStream {
         multiPartMap.values()
             .stream()
             .map(CompletableFuture::join)
-            .sorted(Comparator.comparing(CompletedPart::partNumber))
             .collect(Collectors.toList());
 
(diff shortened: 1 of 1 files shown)


amogh-jahagirdar:
Ah, it looks like the unit tests are failing due to not being in ascending order? I didn't think was required. I'm validating this against S3 itself in the integration tests.

org.apache.iceberg.aws.s3.TestS3OutputStream > testWriteWithChecksumEnabled FAILED
    software.amazon.awssdk.services.s3.model.S3Exception: The list of parts was not in ascending order. The parts list must be specified in order by part number. (Service: S3, Status Code: 400, Request ID: )

amogh-jahagirdar:
Closing this, I realized I missed something fundamental with MPU. In a concurrent environment, it's very possible the part numbers are out of order in the current implementation and as of now we need to explicitly provide the parts in ascending order. In my head probably a way with atomic counters and a synchronized list which contains the parts in monotonically increasing order of id but it seems more complex than the value added.

PR #5323, opened 2022-07-21
Build: Upgrade caffeine to 3.1.1

nastra:


nastra:
looks like we can't upgrade to 3.x because it requires JDK 11

PR #5326, opened 2022-07-21
Python: Remove .python-version file

Fokko:
I think this is something still from the tox days.


rdblue:
I think this is for `pyenv`, but I don't think that we want to force a specific set of python versions. This should probably be local config and not checked in so I agree with removing it.

Issue #5163, opened 2022-06-29
Support catalog method to set table metadata


szehon-ho:
**Background**: 

#3851 and further prs added support to register a new table in catalog pointing to existing metadata.json.  It may be good to support setting an existing table in catalog to another metadata.json.

**Justification**:  

There have been instances in the past where such a method would have been welcome, the workaround unfortunately is to manually update the catalog (ie, update Hive Metastore directly), or even drop and recreate.

1. Disaster recovery from outdated catalog backup with a table pointing to an older metadata.json
2. Bugs where table becomes in a bad state due to bad metadata modification (in particular, partition field dropping bugs that made the table impossible to query):
a.  https://apache-iceberg.slack.com/archives/C025PH0G1D4/p1654877950329549 (just one instance of trying to help user)
b.  https://github.com/trinodb/trino/issues/8284
c.  https://github.com/apache/iceberg/pull/3411
d.  https://github.com/apache/iceberg/pull/4560
4. Catalog consistency problems where metadata.jsons get overriden from each other:
a.  #5036
b.  some problems in custom catalog impelementations

**Ideas:**

1. Add a force option to catalog::registerTable 
2. Add a different method to catalog (reset table?)

**Concerns:**

This could of course be dangerous, but metadata.json does seem a user-exposed concept already in catalog API and in some utils.  User can achieve this today by dropping table with purge=false, and registering the metadata.json to new table with same name.

We could potentially run some kind of table consistency check (explore all the reachable graph and even historic metadata graph to validate that the new table metadata is consistent).  This could be another utility that could be just generally useful.

RussellSpitzer:
Here is the other register PR, I forget where we were discussing the "force" issue https://github.com/apache/iceberg/pull/4810

flyrain:
@RussellSpitzer and me had a discussion around #4810 if we want to "update" a table with a metadata.json. 

+1 for this idea. I happened to resolve a table corruption case caused by catalog inconsistency. I had to update the table's metadata.json location manually at that time. A force register or update/reset tool will be pretty helpful in that case.

yabola:
@szehon-ho @RussellSpitzer 
I also have a strong need to call register table on the same table to update meta location.
My background:
We write using hadoop catalog, but need to read from hive catalog for special reasons (for example, we need to create a hive view to read table, hadoop catalog can't create hive view). Hive is read only here.
But in order to update the metadata, we need to update the meta location information regularly.

[...]

PR #5327, opened 2022-07-21
Support force option on RegisterTable procedure

yabola:
fix https://github.com/apache/iceberg/issues/5163
Add 'force' option to register existing table, this is optional
usage: `CALL mycatalogname.system.register_table('mycatalogname.mydb.mytablename','xx://xxxxx/metadata/xx.metadata.json','force')`

### DIFF
--- a/api/src/main/java/org/apache/iceberg/catalog/Catalog.java
+++ b/api/src/main/java/org/apache/iceberg/catalog/Catalog.java
@@ -336,14 +336,17 @@ public interface Catalog {
   default void invalidateTable(TableIdentifier identifier) {}
 
   /**
-   * Register a table with the catalog if it does not exist.
+   * Register a table with the catalog.
    *
    * @param identifier a table identifier
    * @param metadataFileLocation the location of a metadata file
+   * @param force If true will register the table even if the table already existed, otherwise
+   *     it will throw table already existing error. Default is false.
    * @return a Table instance
    * @throws AlreadyExistsException if the table already exists in the catalog.
    */
-  default Table registerTable(TableIdentifier identifier, String metadataFileLocation) {
+  default Table registerTable(
+      TableIdentifier identifier, String metadataFileLocation, boolean force) {
     throw new UnsupportedOperationException("Registering tables is not supported");
   }
 
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
@@ -45,6 +45,7 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.NamespaceNotEmptyException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
+import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.hadoop.HadoopFileIO;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
@@ -169,10 +170,19 @@ public class HiveCatalog extends BaseMetastoreCatalog implements SupportsNamespa
 
     TableOperations ops = newTableOps(identifier);
     TableMetadata lastMetadata;
-    if (purge && ops.current() != null) {
-      lastMetadata = ops.current();
-    } else {
-      lastMetadata = null;
+    try {
+      if (purge && ops.current() != null) {
+        lastMetadata = ops.current();
+      } else {
+        lastMetadata = null;
+      }
+    } catch (Exception e) {
+      if (e instanceof NotFoundException) {
+        LOG.warn("The metadata file doesn't exist any more. Continue to drop the table.", e);
+        lastMetadata = null;
+      } else {
+        throw e;
+      }
     }
 
     try {
--- a/aws/src/integration/java/org/apache/iceberg/aws/dynamodb/TestDynamoDbCatalog.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/dynamodb/TestDynamoDbCatalog.java
@@ -308,7 +308,7 @@ public class TestDynamoDbCatalog {
     Assertions.assertThat(catalog.dropTable(identifier, false)).isTrue();
     TableOperations ops = ((HasTableOperations) registeringTable).operations();
     String metadataLocation = ((DynamoDbTableOperations) ops).currentMetadataLocation();
-    Assertions.assertThat(catalog.registerTable(identifier, metadataLocation)).isNotNull();
+    Assertions.assertThat(catalog.registerTable(identifier, metadataLocation, false)).isNotNull();
     Assertions.assertThat(catalog.loadTable(identifier)).isNotNull();
     Assertions.assertThat(catalog.dropTable(identifier, true)).isTrue();
     Assertions.assertThat(catalog.dropNamespace(namespace)).isTrue();
@@ -323,7 +323,9 @@ public class TestDynamoDbCatalog {
     Table registeringTable = catalog.loadTable(identifier);
     TableOperations ops = ((HasTableOperations) registeringTable).operations();
     String metadataLocation = ((DynamoDbTableOperations) ops).currentMetadataLocation();
-    Assertions.assertThatThrownBy(() -> catalog.registerTable(identifier, metadataLocation))
+    catalog.registerTable(identifier, metadataLocation, true);
+    Assertions.assertThat(catalog.loadTable(identifier)).isNotNull();
+    Assertions.assertThatThrownBy(() -> catalog.registerTable(identifier, metadataLocation, false))
         .isInstanceOf(AlreadyExistsException.class);
     Assertions.assertThat(catalog.dropTable(identifier, true)).isTrue();
     Assertions.assertThat(catalog.dropNamespace(namespace)).isTrue();
--- a/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogTable.java
+++ b/aws/src/integration/java/org/apache/iceberg/aws/glue/TestGlueCatalogTable.java
@@ -445,7 +445,7 @@ public class TestGlueCatalogTable extends GlueTestBase {
     Table table = glueCatalog.loadTable(identifier);
     String metadataLocation = ((BaseTable) table).operations().current().metadataFileLocation();
     Assertions.assertThat(glueCatalog.dropTable(identifier, false)).isTrue();
-    Assertions.assertThat(glueCatalog.registerTable(identifier, metadataLocation)).isNotNull();
+    Assertions.assertThat(glueCatalog.registerTable(identifier, metadataLocation, false)).isNotNull();
     Assertions.assertThat(glueCatalog.loadTable(identifier)).isNotNull();
     Assertions.assertThat(glueCatalog.dropTable(identifier, true)).isTrue();
     Assertions.assertThat(glueCatalog.dropNamespace(Namespace.of(namespace))).isTrue();
@@ -459,7 +459,9 @@ public class TestGlueCatalogTable extends GlueTestBase {
     TableIdentifier identifier = TableIdentifier.of(namespace, tableName);
     Table table = glueCatalog.loadTable(identifier);
     String metadataLocation = ((BaseTable) table).operations().current().metadataFileLocation();
-    Assertions.assertThatThrownBy(() -> glueCatalog.registerTable(identifier, metadataLocation))
+    glueCatalog.registerTable(identifier, metadataLocation, true);
+    Assertions.assertThat(glueCatalog.loadTable(identifier)).isNotNull();
+    Assertions.assertThatThrownBy(() -> glueCatalog.registerTable(identifier, metadataLocation, false))
         .isInstanceOf(AlreadyExistsException.class);
     Assertions.assertThat(glueCatalog.dropTable(identifier, true)).isTrue();
     Assertions.assertThat(glueCatalog.dropNamespace(Namespace.of(namespace))).isTrue();
--- a/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
@@ -82,7 +82,7 @@ public abstract class BaseMetastoreCatalog implements Catalog {
       }
       TableMetadata current = ops.current();
       if (metadataFileLocation.equals(current.metadataFileLocation())) {
-        LOG.info("The metadata file location is the same as the existing table, ignore it.");
+        LOG.warn("The metadata file location is the same as the existing table, ignore this operation.");
         return new BaseTable(ops, identifier.toString());
       }
       dropTable(identifier, false);
--- a/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
+++ b/core/src/main/java/org/apache/iceberg/BaseMetastoreCatalog.java
@@ -66,19 +66,28 @@ public abstract class BaseMetastoreCatalog implements Catalog {
   }
 
   @Override
-  public Table registerTable(TableIdentifier identifier, String metadataFileLocation) {
+  public Table registerTable(
+      TableIdentifier identifier, String metadataFileLocation, boolean force) {
     Preconditions.checkArgument(
         identifier != null && isValidIdentifier(identifier), "Invalid identifier: %s", identifier);
     Preconditions.checkArgument(
         metadataFileLocation != null && !metadataFileLocation.isEmpty(),
         "Cannot register an empty metadata file location as a table");
 
-    // Throw an exception if this table already exists in the catalog.
+    TableOperations ops = newTableOps(identifier);
     if (tableExists(identifier)) {
-      throw new AlreadyExistsException("Table already exists: %s", identifier);
+      if (!force) {
+        // Throw an exception if this table already exists in the catalog.
+        throw new AlreadyExistsException("Table already exists: %s", identifier);
+      }
+      TableMetadata current = ops.current();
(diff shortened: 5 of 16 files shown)


yabola:
@flyrain CC~

flyrain:
Hi @RussellSpitzer @szehon-ho, could you approve the workflows?

api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
```diff
@@ -147,10 +147,11 @@ public Map<String, String> properties() {
    * @param context session context
    * @param ident a table identifier
    * @param metadataFileLocation the location of a metadata file
+   * @param force default is false. If true, do register table even if table exists otherwise it will throw error.
    * @return a Table instance
    * @throws AlreadyExistsException if the table already exists in the catalog.
    */
-  Table registerTable(SessionContext context, TableIdentifier ident, String metadataFileLocation);
+  Table registerTable(SessionContext context, TableIdentifier ident, String metadataFileLocation, boolean force);
```
flyrain, line 154: It is a bit confusing for user by reusing the name `registerTable` for updating the metadata location, even with the `force` option. Also `updating metadata location` is a risky operation, we need to separate it from a safe operation(register a table), so that user won't accidentally do that. Suggested names would be `resetTable` or `resetMetadataLocation`.

api/src/main/java/org/apache/iceberg/catalog/Catalog.java
```diff
@@ -356,10 +356,11 @@ default void invalidateTable(TableIdentifier identifier) {
    *
    * @param identifier a table identifier
    * @param metadataFileLocation the location of a metadata file
+   * @param force default is false. If true, do register table even if table exists otherwise it will throw error.
    * @return a Table instance
    * @throws AlreadyExistsException if the table already exists in the catalog.
    */
-  default Table registerTable(TableIdentifier identifier, String metadataFileLocation) {
+  default Table registerTable(TableIdentifier identifier, String metadataFileLocation, boolean force) {
```
flyrain, line 363: The same concern here.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -235,19 +235,21 @@ public void renameTable(TableIdentifier from, TableIdentifier originalTo) {
...
     Preconditions.checkArgument(isValidIdentifier(identifier), "Invalid identifier: %s", identifier);
 
-    // Throw an exception if this table already exists in the catalog.
-    if (tableExists(identifier)) {
+    // Throw an exception if this table already exists in the catalog and no force option.
+    if (tableExists(identifier) && !force) {
       throw new org.apache.iceberg.exceptions.AlreadyExistsException("Table already exists: %s", identifier);
     }
+    HiveTableOperations ops = (HiveTableOperations) newTableOps(identifier);
+    TableMetadata base = ops.current();
```
flyrain, line 247: Do we need to get the `base`? Passing the null should be fine, right?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -215,8 +215,9 @@ protected void doRefresh() {
   @SuppressWarnings("checkstyle:CyclomaticComplexity")
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
-    String newMetadataLocation = base == null && metadata.metadataFileLocation() != null ?
-        metadata.metadataFileLocation() : writeNewMetadata(metadata, currentVersion() + 1);
+    boolean isTableForRegister = metadata.isForRegister();
+    String newMetadataLocation = isTableForRegister ? metadata.metadataFileLocation()
```
flyrain, line 219: Do we need to check if `metadata.metadataFileLocation()` is null?

api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
flyrain: I think a new procedure will be better. Would like to hear from others. @szehon-ho @RussellSpitzer.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -215,8 +215,9 @@ protected void doRefresh() {
   @SuppressWarnings("checkstyle:CyclomaticComplexity")
   @Override
   protected void doCommit(TableMetadata base, TableMetadata metadata) {
-    String newMetadataLocation = base == null && metadata.metadataFileLocation() != null ?
-        metadata.metadataFileLocation() : writeNewMetadata(metadata, currentVersion() + 1);
+    boolean isTableForRegister = metadata.isForRegister();
```
szehon-ho, line 218: Should probably be called 'forReset()' or something more descriptive.  This is a bit ugly but couldn't think of a good way to avoid it.  I suppose just having a metadata with a location != null , regardless of whether base is null or not , is not enough to signal its a reset? @aokolnychyi @RussellSpitzer @rdblue for thoughts?

szehon-ho, line 218: Also seems like we should implement this for all the catalogs, or at least error there if this path is chosen.

api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
RussellSpitzer: I liked it being just an additional parameter, but I think I'm outvoted here :)

RussellSpitzer: Based on @szehon-ho 's suggestion below perhaps this should be "resetTable" based on the `forReset` comments below

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
RussellSpitzer: I'm not sure we should go through the operation.commit pathway here. Feels like what we are doing here is different than every other kind of commit. Really all we want to do is transactionally swap one parameter for another. I'm gonna think more about this tomorrow

api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java
yabola: ok

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -235,19 +235,21 @@ public void renameTable(TableIdentifier from, TableIdentifier originalTo) {
   }
 
   @Override
-  public org.apache.iceberg.Table registerTable(TableIdentifier identifier, String metadataFileLocation) {
+  public org.apache.iceberg.Table registerTable(TableIdentifier identifier,
```
ajantha-bhat, line 238: please rebase.

https://github.com/apache/iceberg/pull/5037 has moved register table to `BaseMetastoreCatalog`

yabola:
@flyrain @RussellSpitzer @szehon-ho I had separated the resetTable API and do not reuse the `commit` method (Feels like what we are doing here is different than every other kind of commit.)
Do you think it is more clear? If it looks good to you, I will finish some other catalog resetTable API.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -345,6 +345,30 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
   }
 
+  @Override
+  public void reset(String metadataFileLocation) {
+    Optional<Long> lockId = Optional.empty();
+    ReentrantLock tableLevelMutex = commitLockCache.get(fullName, t -> new ReentrantLock(true));
+    tableLevelMutex.lock();
+    try {
+      lockId = Optional.of(acquireLock());
+      Table tbl = loadHmsTable();
```
pvary, line 355: I do not like this. Too big part of the commit code is duplicated, and this is especially true for the error handling. This part is complicated and error prone, so we might want to try to find a better way to wave the to parts of the code together.

yabola: I want to abstract this part of the locking logic, let me think about it~

yabola: @pvary I abstract the part of the metadata locking logic. Please have a look.

api/src/main/java/org/apache/iceberg/catalog/Catalog.java
```diff
@@ -347,6 +347,17 @@ default Table registerTable(TableIdentifier identifier, String metadataFileLocat
...
   }
 
+  /**
+   * Reset the table with a meta file.
+   *
+   * @param identifier a table identifier
+   * @param metadataFileLocation the location of a metadata file
+   * @return a Table instance
+   */
+  default Table resetTable(TableIdentifier identifier, String metadataFileLocation) {
```
rdblue, line 357: Do we really need to add another method to `Catalog`? What is the difference between this and `dropTable` followed by `registerTable`?

rdblue:
I think we should be very careful about introducing new methods to the `Catalog` API. There needs to be a strong justification for adding to it, and I don't think merging this PR is a good idea.

api/src/main/java/org/apache/iceberg/catalog/Catalog.java
yabola: But It may have a time gap, when there are consecutive queries. And it doesn't seem guaranteed to be a transactional operation

yabola:
> I think we should be very careful about introducing new methods to the `Catalog` API. There needs to be a strong justification for adding to it, and I don't think merging this PR is a good idea.

How about reuse `registerTable` procedure?

rdblue:
> How about reuse registerTable procedure?

If you mean to add more arguments to `registerTable` so that it is an atomic operation, I don't really see much point to it. If this is a recovery operation for cases like the ones @szehon-ho listed in the issues, then why does it need to be atomic? The table is already broken.

We added `registerTable` as a way to migrate to a new metastore. At the time, there were concerns about people misusing it to fork tables and do other unreliable things. I still have that concern for this operation. Reusing `registerTable` as-is seems like the best way to avoid those problems.

szehon-ho:
I think that's a great idea, to have the Catalog's regiserTable just first call dropTable (purge=false) , instead of throw an error, if table exists, it could be a very simple way to do this.

yabola:
@szehon-ho I don't quite understand. I think @rdblue  means don't change the existing logic. Do you mean `registerTable` can be changed to drop first(if existed) and then register?

szehon-ho:
So still to me it makess sense to add a force option to registerTable.  Initially we were debating between adding a new Catalog API , or re-using registerTable with force option.  I initially preferred the first as its more clear to user, but I guess @rdblue is not a fan of adding the new API.  

So we could do this, today we throw an error if the table exists, but if force==true, we just call drop table.  And clarify in the docs what force option does.  I would probably also in the code move some the logic to read the metadataFile up and then add a Precondition that the file exists, before the if (table.exists() block) because now it will drop if force=true.  

@RussellSpitzer @flyrain does that sound like it will work ?

RussellSpitzer:
> So we could do this, today we throw an error if the table exists, but if force==true, we just call drop table. And clarify in the docs what force option does. I would probably also in the code move some the logic to read the metadataFile up and then add a Precondition that the file exists, before the if (table.exists() block) because now it will drop if force=true.
> 
> @RussellSpitzer @flyrain does that sound like it will work ?

I prefer this approach (I think I said so before :) ) I still want the default of "don't crash a table that's already there" but I like having the option of "succeed no matter what". I think "delete" + "register" is fine even if we don't swap it atomically. My original thought was for folks doing more dangerous live swaps where we are shifting between two valid metadata.json s but even a small amount of downtime for that wouldn't be a big deal.

szehon-ho:
> I prefer this approach (I think I said so before :) ) 

Yes fine you win :).  Yea the only kind of downside is if we error re-creating the table, the table is dropped as the two are not atomic.  But register can just be called again in that case.  Agree its not worth the complexity to implement atomicity for a small gain.

flyrain:
My concern for dropping table is that a catalog like HMS usually has hooks to react on event like table is dropped. We don't really want to drop a table here, but it may trigger something unexpected.

RussellSpitzer:
> > I prefer this approach (I think I said so before :) )
> 
> Yes fine you win :). Yea the only kind of downside is if we error re-creating the table, the table is dropped as the two are not atomic. But register can just be called again in that case. Agree its not worth the complexity to implement atomicity for a small gain.

I thought we could just switch the "create" to an "alter" and keep atomicity but I didn't look too much into that.

yabola:
@szehon-ho as the discussion above, please take a look on my latest change

rdblue:
@szehon-ho, what I originally meant was that it would be fine for the caller to manually call `dropTable` first. I'm hesitant to add a `force` option because it's basically the same thing as a reset, we just hide it a bit.

Thinking about the problem a bit more, I'm not sure that the use cases we're trying to satisfy make this a catalog operation. If this is about replacing the metadata for a single existing table, why not make this a table operation? Or better yet, why not use the existing table operation to do this? If the table exists, then you can always call `table.operations().commit(table.operations().current(), replacementMetadata)`.

RussellSpitzer:
@rdblue The issue with those solutions for our use case is that they only work if the current metadata is valid. If not, we cannot drop the table via the Catalog api (requires loading the metadata first) which has led me to constantly showing folks how to access the hive via the Spark external client wrapper. The same issue occurs if we want to make this a table operation, we can't perform an operation on a table which doesn't have a valid metadata.json.

Our issue is that in Spark we can't Drop the table or point to a new metadata without using an api for the metastore directly. Ideally I think the Iceberg catalog implementation should provide an exit valve for these situations so users aren't required to manually connect to the underlying catalog system themselves. 

I am open to other solutions here but I'm not very worried about folks overusing this api, it tends to only come up when the user is already in a really bad state.

yabola:
But call `table.operations().commit(table.operations().current(), replacementMetadata)` can't update the metadata location (There is some check logic in this method that prevents changes).
 I think this PR provides a command line way to accomplish this function.

rdblue:
@RussellSpitzer, I see what you mean. What about fixing the Hive catalog so that the commit succeeds if table metadata can't be loaded? It sounds like we should have better handling for broken tables.

RussellSpitzer:
> @RussellSpitzer, I see what you mean. What about fixing the Hive catalog so that the commit succeeds if table metadata can't be loaded? It sounds like we should have better handling for broken tables.

That would be ok for me as well, I'm really just looking for a blessed way of fixing these states. I think allowing for dropping an non-iceberg or iceberg table with invalid metadata is probably ok too but that does seem a little scary to me than having an explicit register command which we know isn't part of normal operations.

szehon-ho:
> What about fixing the Hive catalog so that the commit succeeds if table metadata can't be loaded? It sounds like we should have better handling for broken tables.

Are we suggesting change dropTable to continue if table metadata cannot be loaded?  (I assume we wont add any force flag?).  That sounds like a great feature, its one of the pain points using Iceberg at least on Hive catalog, dropping a tables where metadata doesnt exist, would love to see a pr on that.

It's not the complete solution to the issue I initially filed (rollback corrupt metadata) though, and agree it still would be nice if there's a blessed way somehow

rdblue:
@szehon-ho, in part, yes. I think that if we have situations where a table is corrupted then we should handle those in a better way than just having a table that requires manual intervention. Making DROP work for those tables sounds like a good idea to me. Maybe we could also update `registerTable` to overwrite a corrupt Iceberg table using similar logic. Then we wouldn't need a flag.

flyrain:
+1 to detect if the table is corrupt first, then register the table with the new metadata file location. No flag is needed.

szehon-ho:
Sounds good.  The only corruption detection we can tell now is if the metadata.json no longer exists, I think, unless there are other things we can easily detect.

My initial thought for this pr was actually for non-corrupt case (ie, I have a metadata json that loads fine, but the metadata is wrong, eg  issues listed in #5163).  But I guess the consensus is , the user just writes a script to drop and register the Iceberg table with the right metadata json in that case, and we don't want to extend Iceberg Catalog at this point to handle those cases.

@yabola sorry about that, but I guess if you want to instead make a patch for the issue (make HiveCatalog.dropTable able to handle non-exist metadata json).

yabola:
@szehon-ho Sure, I will make another PR for drop corrupt table. Thanks for all your discussions !

GabeChurch:
I'm going to bump this as well. 

It is potentially extremely helpful in cases where we want to "extend-read-only" s3-access to an iceberg table across environments, but have a separate hive metastores. For instance -- say we have a sandbox environment where we want to make a table from a production environment accessible that doesn't depend on atomicity, Extending read-only access would save a HUGE expense in replication. 

With the existing procedure the only option is to force drop the table before re-registering a new metafile, which can break other things attempting to access the table. It would be much better to maintain a lock on the table until the metadata has changed (drop+register at API level) so that other queries do not fail if they attempt to query while new metadata is being written via the force option mentioned. 

If we can't add a "force" option to the "iceberg.system.register_table" we should add a new method to update metadata without requiring end users to drop a table via another system process -- maybe something like "iceberg.system.override_registered_table"

rushilshah1:
I am also interested in this functionality as well! Are there still plans to add this?

abfisher0417:
I am also interested in this capability. I have a use case where I create/update Iceberg tables in an isolated cloud environment separate from consumers/readers of those tables. Consumers would periodically sync all metadata/data files from the producing cloud storage bucket to their own cloud storage bucket; consumers would also have their own catalog(s) that point to the underlying table metadata/data files.

The RegisterTable procedure today would allow partners in this scenario to add tables to their own catalog. However, there's not a "clean" way to point an existing table to a newer version of the same table without first dropping the table the re-adding it.

yabola:
@abfisher0417 Thank you. If community agrees, I can complete this PR again.

PR #5328, opened 2022-07-21
API: Deprecate Counter#count() / Add Counter#value() 

nastra:
The rationale behind this change is that usually a Counter should always
be able to report the actual count it is counting. The reason
`count()` was made `Optional` initially is because in the case of Hadoop
we are reporting the count to Hadoop's file system statistics and don't
necessarily want to report the count in that case.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
@@ -50,6 +50,11 @@ class IntCounter implements MetricsContext.Counter<Integer> {
   }
 
   @Override
+  public Integer value() {
+    return counter.get();
+  }
+
+  @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
         .add("counter", counter)
--- a/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
@@ -66,9 +66,20 @@ public interface MetricsContext extends Serializable {
      * Reporting count is optional if the counter is reporting externally.
      *
      * @return current count if available
+     * @deprecated Use {@link Counter#value()}
      */
+    @Deprecated
     default Optional<T> count() {
       return Optional.empty();
+    }
+
+    /**
+     * Reports the current count.
+     *
+     * @return The current count
+     */
+    default T value() {
+      throw new UnsupportedOperationException("Count is not supported.");
     }
 
     default Unit unit() {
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/apache/iceberg/metrics/IntCounter.java
```diff
@@ -45,8 +44,8 @@ public void increment(Integer amount) {
   }
 
   @Override
-  public Optional<Integer> count() {
-    return Optional.of(counter.get());
+  public Integer count() {
```
rdblue, line 47: Since this is in API, we can't remove it without deprecation. For now, can you add it back and just always return a non-empty optional? Then we can add a new method that always returns the value, like `value()`.

danielcweeks: Why did the `revapi` check pass with this?  Shouldn't it catch this type of change?

rdblue: Yes, it should!

nastra: I looked a bit why Revapi didn't fail here. Turns out https://github.com/apache/iceberg/blob/master/.palantir/revapi.yml#L2 still compares the API to the 0.13.0 version, which didn't have the `Counter` API.

@rdblue can you please create `release-base-0.14.0`? 
I created https://github.com/apache/iceberg/pull/5336 that tells RevAPI to check API compatibility against `release-base-0.14.0`. Additionally, `release-base-0.14.0` is required so that nightly snapshot publishing uses `0.15.0-SNAPSHOT` (currently it still does `0.14.0-SNAPSHOT`, because of https://github.com/apache/iceberg/blob/8300b219ec8962bd22af69adf67a4ab72dec1f48/build.gradle#L682).

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
```diff
@@ -66,11 +66,20 @@ default void initialize(Map<String, String> properties) {
...
     default Optional<T> count() {
       return Optional.empty();
     }
 
+    /**
+     * Reports the current count.
+     *
+     * @return The current count
+     */
+    T value();
```
nastra, line 81: we could make this a breaking API change and add the justification to RevAPI (as it is here) or (as @danielcweeks 
 suggested) change this to 
```
default T value() {
  throw new UnsupportedOperationException("Count is not supported.");
}
```

rdblue: I'd say let's add the default.

rdblue (approved):
Looks good, but I'd prefer to add the default implementation rather than make an exception for revapi.

nastra:
@rdblue pushed the version with the default implementation for `value()`

PR #5329, opened 2022-07-21
Improve bloom filter test

huaxingao:
I saw bloom filter test `Assert.assertFalse("Should not read: ...", shouldRead)` failed with false positive.

This PR is to make the bloom filter test less flaky by taking consideration of fpp.

### DIFF
--- a/parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
+++ b/parquet/src/test/java/org/apache/iceberg/parquet/TestBloomRowGroupFilter.java
@@ -22,7 +22,6 @@ package org.apache.iceberg.parquet;
 import java.io.File;
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.math.BigInteger;
 import java.time.Instant;
 import java.util.List;
 import java.util.Random;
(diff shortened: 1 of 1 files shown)


szehon-ho:
While this makes sense, I wonder if its even worth to assert that number of false positives is less than some number?  Each test asserting there's no false negatives may be good enough?

I assume that there is still a failure chance, and it may not really be suitable for unit test to be run each time?  Curious what other parquet or other projects do.  Also wondering what others think.

huaxingao:
I think besides asserting no false negative, we probably still need to have a negative test for each of the data types?

szehon-ho:
Yea I'm not a huge fan of having random failures still possible in build tests but could go either way on it.  Also curious with your buffer what is the estimated chance of failure ?  (like is it astronomically small )

Wonder any thoughts from @kbendick @rdblue @RussellSpitzer

rdblue:
Rather than adding code for false positives, wouldn't it be better to make the test use a consistent random seed that doesn't have a false positive? You'd need to replace `UUID.randomUUID` with a UUID built from random longs instead, but I think that would be better.

huaxingao:
@rdblue Thanks for the suggestion! Done.

rdblue (approved):

rdblue:
Thanks, @huaxingao!

huaxingao:
Thank you very much! @rdblue @szehon-ho

PR #5331, opened 2022-07-21
WIP: Adding support for Delta to Iceberg migration

ericlgoodman:
**This is a WIP PR (adding unit tests and still testing edge cases) but wanted to put this out here to get initial feedback.**

It adds support to migrate a Delta Lake table to an Iceberg table. Currently this will only create a new table and effectively compress the history of the table into a single commit. In the future, we can add functionality to optionally carry over the history of the Delta Lake table.

It also assumes that the user's current Delta Lake resides in what has now been configured to be an Iceberg catalog. Thinking of adding an optional `destinationCatalog` for scenarios where users are attempting to move between catalogs.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/actions/BaseMigrateDeltaLakeTableActionResult.java
+++ b/core/src/main/java/org/apache/iceberg/actions/BaseMigrateDeltaLakeTableActionResult.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.actions;
+
+public class BaseMigrateDeltaLakeTableActionResult implements MigrateDeltaLakeTable.Result {
+
+  private final long importedDataFilesCount;
+
+  public BaseMigrateDeltaLakeTableActionResult(long importedDataFilesCount) {
+    this.importedDataFilesCount = importedDataFilesCount;
+  }
+
+  @Override
+  public long importedDataFilesCount() {
+    return importedDataFilesCount;
+  }
+}
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java
@@ -338,7 +338,7 @@ public class SparkSchemaUtil {
     return identitySpec(schema, names);
   }
 
-  private static PartitionSpec identitySpec(Schema schema, List<String> partitionNames) {
+  public static PartitionSpec identitySpec(Schema schema, List<String> partitionNames) {
     if (partitionNames == null || partitionNames.isEmpty()) {
       return null;
     }
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
@@ -1,0 +1,383 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.spark.actions;
+
+import io.delta.standalone.DeltaLog;
+import io.delta.standalone.VersionLog;
+import io.delta.standalone.actions.Action;
+import io.delta.standalone.actions.AddFile;
+import io.delta.standalone.actions.RemoveFile;
+import java.io.File;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.hadoop.fs.Path;
+import org.apache.iceberg.AppendFiles;
+import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DataFiles;
+import org.apache.iceberg.DeleteFiles;
+import org.apache.iceberg.FileFormat;
+import org.apache.iceberg.Metrics;
+import org.apache.iceberg.MetricsConfig;
+import org.apache.iceberg.OverwriteFiles;
+import org.apache.iceberg.PartitionField;
+import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotSummary;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.TableProperties;
+import org.apache.iceberg.actions.BaseMigrateDeltaLakeTableActionResult;
+import org.apache.iceberg.actions.MigrateDeltaLakeTable;
+import org.apache.iceberg.data.TableMigrationUtil;
+import org.apache.iceberg.exceptions.AlreadyExistsException;
+import org.apache.iceberg.exceptions.NoSuchNamespaceException;
+import org.apache.iceberg.mapping.NameMapping;
+import org.apache.iceberg.mapping.NameMappingParser;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+import org.apache.iceberg.spark.SparkSchemaUtil;
+import org.apache.iceberg.spark.source.StagedSparkTable;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.catalyst.analysis.TableAlreadyExistsException;
+import org.apache.spark.sql.connector.catalog.CatalogPlugin;
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.apache.spark.sql.connector.catalog.StagingTableCatalog;
+import org.apache.spark.sql.connector.expressions.LogicalExpressions;
+import org.apache.spark.sql.connector.expressions.Transform;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.collection.JavaConverters;
+
+/**
+ * Takes a Delta Lake table and attempts to transform it into an Iceberg table in the same location
+ * with the same identifier. Once complete the identifier which previously referred to a non-Iceberg
+ * table will refer to the newly migrated Iceberg table.
+ */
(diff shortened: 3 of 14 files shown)


core/src/main/java/org/apache/iceberg/actions/BaseMigrateDeltaLakeTableActionResult.java
```diff
@@ -0,0 +1,34 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg.actions;
+
+public class BaseMigrateDeltaLakeTableActionResult implements MigrateDeltaLakeTable.Result {
+
+  private final long numFilesImported;
+
+  public BaseMigrateDeltaLakeTableActionResult(long numFilesImported) {
```
amogh-jahagirdar, line 26: Would it make sense to capture aspects like the total data size and how long the migration procedure took in the result?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java
```diff
@@ -293,7 +293,7 @@ private static PartitionSpec identitySpec(Schema schema, Collection<Column> colu
     return identitySpec(schema, names);
   }
 
-  private static PartitionSpec identitySpec(Schema schema, List<String> partitionNames) {
+  public static PartitionSpec identitySpec(Schema schema, List<String> partitionNames) {
```
amogh-jahagirdar, line 296: Why is the change in access modifiers?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
```diff
@@ -0,0 +1,311 @@
...
+                  fullPath,
+                  // Delta tables only support parquet
+                  "parquet"
+              );
+        }
+        )
+        .collect(Collectors.toList());
+  }
+
+  private static PartitionSpec getPartitionSpecFromDeltaSnapshot(
```
amogh-jahagirdar, line 203: I get what the function is doing but does it need to be static?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
```diff
@@ -0,0 +1,311 @@
...
+    } catch (org.apache.spark.sql.catalyst.analysis.NoSuchTableException e) {
+      LOG.error("Cannot restore the original table, the backup table {} cannot be found", backupIdent, e);
+
+    } catch (org.apache.spark.sql.catalyst.analysis.TableAlreadyExistsException e) {
+      LOG.error("Cannot restore the original table, a table with the original name exists. " +
+              "Use the backup table {} to restore the original table manually.", backupIdent, e);
+    }
+  }
+
+  private static List<SparkTableUtil.SparkPartition> getSparkPartitionsFromDeltaSnapshot(
```
amogh-jahagirdar, line 174: same, does it need to be static?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -380,7 +381,9 @@ private static Iterator<ManifestFile> buildManifest(SerializableConfiguration co
    */
   public static void importSparkTable(SparkSession spark, TableIdentifier sourceTableIdent, Table targetTable,
                                       String stagingDir, Map<String, String> partitionFilter,
-                                      boolean checkDuplicateFiles) {
+                                      boolean checkDuplicateFiles,
+                                      PartitionSpec nullableSpec,
+                                      List<SparkPartition> nullablePartitions) {
```
amogh-jahagirdar, line 386: I get we need to pass in the computed partition spec and spark partitions, but don't think we should change any public API signatures, we can add a new one.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
```diff
@@ -0,0 +1,311 @@
...
+              return new SparkTableUtil.SparkPartition(
+                  entry.getKey(), // Map containing name and values of partitions
+                  fullPath,
+                  // Delta tables only support parquet
+                  "parquet"
+              );
+        }
+        )
+        .collect(Collectors.toList());
+  }
```
amogh-jahagirdar, line 201: Should we parallelize this computation? For example, concurrently across files update a grouping based on partition values. Could be overkill (depends on the number of files in the snapshot)

amogh-jahagirdar (commented):
Left some comments and questions. Also suggest, rebase and running the spotlessApply so discussion can be more on the fundamental aspects and less on nit style stuff.

ericlgoodman:
Adding here my primary concern with this PR - and in general a concern going forward with Spark and using multiple tables such as Delta Lake and Iceberg.

Spark reads tables through whatever catalog is located at the first part of a table's identifier. There can only be 1 catalog per identifier, and different catalogs have different capabilities. For example, the `DeltaCatalog` can read Delta Lake and generic Hive tables, and the `SparkSessionCatalog` can read Iceberg + Hive tables.

In theory, in order to read from multiple table types in one Spark session, a user would initialize a `DeltaCatalog`, at say, `delta` and then the `SparkSessionCatalog` at `iceberg`. Then all their Delta Lake tables would be located at `delta.my_delta_database.my_delta_lake_table` and all their Iceberg tables at `iceberg.my_iceberg_database.my_iceberg_table`. Unfortunately, this doesn't work out of the box. Both of these catalog implementations are designed to be used by overriding the default Spark catalog, which is located at `spark_catalog`. `CatalogExtension`, from which `DeltaCatalog` and `SparkSessionCatalog` both inherit from, contains a method `setDelegateCatalog(CatalogPlugin delegate)`. As the Javadoc reads:

```java
 /**
   * This will be called only once by Spark to pass in the Spark built-in session catalog, after
   * {@link #initialize(String, CaseInsensitiveStringMap)} is called.
   */
  void setDelegateCatalog(CatalogPlugin delegate);
```

A user can fix this issue by manually calling this method during Spark setup and setting the delegate to the one in the default Spark catalog. But most users presumably are not doing this, and some users might face difficulty depending on their service provider and how much abstraction/configuration has been taken away from them during setup.

This basically means that in today's world, it doesn't seem realistic that users currently have a simple way to use one Spark session to read/migrate between different table types. This solution might make make sense to implement first, as users may find that a Delta/Iceberg/Hudi table makes sense for them in one context but another one is preferable in another.

When it comes to migration, there are basically two options:
1. Create a more abstract Catalog implementation that can read Iceberg/Delta/Hudi/Hive tables dynamically, similar to what happens in the Trino Hive connector. The connector inspects the table properties and determines at runtime whether to redirect to another connector. Similarly, a Spark catalog could simply delegate to specific catalogs if it sees certain table type specific properties.
2. Provide an easier method for users to not have to override the default catalog for these table type specific catalog implementations. If the Delta catalog was located at `delta`, and Iceberg at `iceberg`, then users could just keep their different table types in different catalogs and migration could take an optional parameter of the new desired catalog.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkTableUtil.java
```diff
@@ -393,16 +396,17 @@ public static void importSparkTable(SparkSession spark, TableIdentifier sourceTa
...
       } else {
-        List<SparkPartition> sourceTablePartitions = getPartitions(spark, sourceTableIdent,
-            partitionFilter);
-        Preconditions.checkArgument(!sourceTablePartitions.isEmpty(),
+        List<SparkPartition> partitions = nullablePartitions != null ? nullablePartitions
+                : getPartitions(spark, sourceTableIdent, partitionFilter);
+        Preconditions.checkArgument(!partitions.isEmpty(),
             "Cannot find any partitions in table %s", sourceTableIdent);
-        importSparkPartitions(spark, sourceTablePartitions, targetTable, spec, stagingDir, checkDuplicateFiles);
+        importSparkPartitions(spark, partitions, targetTable, nullableSpec, stagingDir, checkDuplicateFiles);
```
jackye1995, line 409: prefer to not change variable name if possible

jackye1995 (commented):
So based on my understanding, the difference of this action vs the existing migrate for Hive table is that it uses DeltaLog to get partition spec, schema and partitions of the table. However, I don't see the logic of reproducing the Delta transactions as Iceberg snapshots, which is the key value of such migration.

I understand the challenge of doing this under the context of Spark, but I think we can actually do better and do not rely on Spark context at all. What we can do is to have a BaseMigrateDeltaLakeTableAction that takes in a `icebergCatalog` instance and a Delta Lake table location. With those 2 information, we can fully replay the DeltaLog information and create an Iceberg table out of that.

Then we can seek for a Spark-based implementation, where the `icebergCatalog` can be retrieved from `SparkSessionCatalog.icebergCatalog()`, and `location` can be retrieved from a Delta table name + Spark default catalog's Delta table serde property `path`

What do you think?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateTableSparkAction.java
```diff
@@ -121,7 +121,7 @@ private MigrateTable.Result doExecute() {
 
       Some<String> backupNamespace = Some.apply(backupIdent.namespace()[0]);
       TableIdentifier v1BackupIdent = new TableIdentifier(backupIdent.name(), backupNamespace);
-      String stagingLocation = getMetadataLocation(icebergTable);
+      String stagingLocation = SparkTableUtil.getIcebergMetadataLocation(icebergTable);
```
ericlgoodman, line 124: Need to remove this change

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/SnapshotTableSparkAction.java
```diff
@@ -124,7 +124,7 @@ private SnapshotTable.Result doExecute() {
       ensureNameMappingPresent(icebergTable);
 
       TableIdentifier v1TableIdent = v1SourceTable().identifier();
-      String stagingLocation = getMetadataLocation(icebergTable);
+      String stagingLocation = SparkTableUtil.getIcebergMetadataLocation(icebergTable);
```
ericlgoodman, line 127: Same here

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
```diff
@@ -0,0 +1,383 @@
...
+        .withRecordCount(metrics.recordCount())
+        .build();
+  }
+
+  private Metrics getMetricsForFile(Table table, String fullFilePath) {
+    MetricsConfig metricsConfig = MetricsConfig.forTable(table);
+    String nameMappingString = table.properties().get(TableProperties.DEFAULT_NAME_MAPPING);
+    NameMapping nameMapping =
+        nameMappingString != null ? NameMappingParser.fromJson(nameMappingString) : null;
+    return TableMigrationUtil.getParquetMetrics(
```
jackye1995, line 256: we should do a check to make sure the file is indeed Parquet

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
```diff
@@ -0,0 +1,383 @@
...
+
+  private PartitionSpec getPartitionSpecFromDeltaSnapshot(StructType structType) {
+    Schema schema = SparkSchemaUtil.convert(structType);
+    PartitionSpec spec =
+        SparkSchemaUtil.identitySpec(
+            schema, deltaLog.snapshot().getMetadata().getPartitionColumns());
+    return spec == null ? PartitionSpec.unpartitioned() : spec;
+  }
+
+  private StructType getStructTypeFromDeltaSnapshot() {
```
jackye1995, line 288: why not just convert directly from Delta schema to Iceberg schema? Here we ended up converting to Spark and then convert to Iceberg, and as a result you had to make a few Spark classes and methods public.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/actions/MigrateDeltaLakeTableSparkAction.java
```diff
@@ -0,0 +1,383 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(MigrateDeltaLakeTableSparkAction.class);
+
+  private final Map<String, String> additionalProperties = Maps.newHashMap();
+  private final SparkSession spark;
+  private final DeltaLog deltaLog;
+  private final StagingTableCatalog destCatalog;
+  private final String deltaTableLocation;
+  private final Identifier newIdentifier;
+
+  MigrateDeltaLakeTableSparkAction(
```
jackye1995, line 95: Can we have a class `BaseMigrateDeltaLakeTableActon`? Because lots of logic could be shared when we want to extend this feature to other engines like Flink and Trino.

jackye1995: and that class can live in iceberg-core library.

JonasJ-ap:
Hi @ericlgoodman. My name is Rushan Jiang, a CS undergrad at CMU. I am interested in learning and contributing to this migration support. I saw you did not update this PR for some time. Would you mind allowing me to continue your work? 

I appreciate your time and consideration.

ericlgoodman:
> Hi @ericlgoodman. My name is Rushan Jiang, a CS undergrad at CMU. I am interested in learning and contributing to this migration support. I saw you did not update this PR for some time. Would you mind allowing me to continue your work?
> 
> I appreciate your time and consideration.

Followed up with you on the Iceberg Slack.

jackye1995:
Close since the PR is merged.

PR #5332, opened 2022-07-22
[Python] FsspecFileIO, a FileIO that wraps any fsspec compliant filesystem

samredai:
This adds a FileIO implementation using [s3fs](https://github.com/fsspec/s3fs) as well as a test running environment using a docker-compose config.


python/tests/io/test_s3fs.py
```diff
@@ -0,0 +1,194 @@
...
+
+@pytest.mark.s3
+def test_checking_if_a_file_exists(s3fs_fileio):
+    """Test checking if a file exists"""
+
+    non_existent_file = s3fs_fileio.new_input(location="s3://testbucket/does-not-exist.txt")
+    assert not non_existent_file.exists()
+
+    filename = str(uuid.uuid4())
+    output_file = s3fs_fileio.new_output(location=f"s3://testbucket/{filename}")
```
dramaticlly, line 150: nit: maybe we can also assert `output_file` does not exist before it get created?

dramaticlly (approved):
LGTM, I want to give it a try on both minIO S3 mock and actual S3 buckets myself!

python/tests/io/test_s3fs.py
samredai: Fixed! Thanks @dramaticlly

python/Makefile
```diff
@@ -17,13 +17,22 @@
...
+	poetry run coverage run --source=pyiceberg/ -m pytest tests/ -m "not s3" ${PYTEST_ARGS}
+	poetry run coverage report -m --fail-under=90
+	poetry run coverage html
+	poetry run coverage xml
+test-all:
 	poetry run coverage run --source=pyiceberg/ -m pytest tests/ ${PYTEST_ARGS}
 	poetry run coverage report -m --fail-under=90
 	poetry run coverage html
 	poetry run coverage xml
+test-s3:
```
kbendick, line 34: Eventually I think we can maybe use some kind of arguments to generify but this is fine imo for now. And maybe later too depending.

python/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+      - AWS_ACCESS_KEY_ID=admin
+      - AWS_SECRET_ACCESS_KEY=password
+      - AWS_REGION=us-east-1
+    entrypoint: >
+      /bin/sh -c "
+      until (/usr/bin/mc config host add minio http://minio:9000 admin password) do echo '...waiting...' && sleep 1; done;
+      /usr/bin/mc rm -r --force minio/testbucket;
+      /usr/bin/mc mb minio/testbucket;
+      /usr/bin/mc policy set public minio/testbucket;
+      exit 0;
```
kbendick, line 45: Question: would it be better to add the intégration tests in their own PR?

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+
+    def __init__(self, location: str, s3):
+        self._s3 = s3
+        super().__init__(location=location)
+
+    def __len__(self) -> int:
+        """Returns the total length of the file, in bytes"""
+        object_info = self._s3.info(self.location)
+        if object_info.get("Size"):  # s3fs versions seem inconsistent on the case used for size
```
kbendick, line 116: I guess this is a hot path bit maybe some kind of variant of `return object_info[size] for size in ["Size", "size"] if object.(size)`?

I'm good with this but maybe it's cheaper just to try to lower case it and avoid the if else in the length method? Just a thought.

kbendick (commented):
Some initia thoughts

python/pyiceberg/io/s3fs.py
samredai: I like that but you also reminded me that I could just do `return object_info.get("Size") or object_info.get("size")` and let python figure out which one is not `None` :)

python/docker-compose.yml
samredai: Good idea, let me remove this!

python/Makefile
```diff
@@ -17,13 +17,22 @@
...
 
 lint:
 	poetry run pre-commit run --all-files
-
 test:
+	poetry run coverage run --source=pyiceberg/ -m pytest tests/ -m "not s3" ${PYTEST_ARGS}
+	poetry run coverage report -m --fail-under=90
+	poetry run coverage html
+	poetry run coverage xml
+test-all:
```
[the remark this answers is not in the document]
samredai: Definitely not necessary, I'll make sure to keep them in when I move this out to a separate PR.

python/Makefile
samredai: That's a great idea, we already use `PYTEST_ARGS` and I think we could just pass in the marked tests we want to run using that 👍

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class S3fsFileIO(FileIO):
+    """A FileIO implementation that uses s3fs"""
+
+    def __init__(self, client_kwargs: Optional[dict] = None, **s3fs_filesystem_kwargs):
+        if client_kwargs:
+            self._s3_filesystem = S3FileSystem(client_kwargs=client_kwargs, **s3fs_filesystem_kwargs)
+        else:
+            self._s3_filesystem = S3FileSystem(**s3fs_filesystem_kwargs)
```
rdblue, line 197: The default value of `client_kwargs` is `None`, so you shouldn't need the `if` / `else`. You can always call the one above.

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def to_input_file(self) -> S3fsInputFile:
+        """Returns a new S3fsInputFile for the location at `self.location`"""
+        return S3fsInputFile(location=self.location, s3=self._s3)
+
+
+class S3fsFileIO(FileIO):
+    """A FileIO implementation that uses s3fs"""
+
+    def __init__(self, client_kwargs: Optional[dict] = None, **s3fs_filesystem_kwargs):
```
rdblue, line 193: I think we should have a convention for passing these arg groups. Catalogs accept an arbitrary property map and that is primarily what will be used to create `FileIO` instances. That means we should translate from a single map for all of these options.

For these two, I would create a `client` group that we pull from the properties dict and pass here, then pass the rest of the properties somewhere. We may want to also standardize the S3 args from Java and pass them correctly here. What do you think?

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+            OutputFile instance is provided, the location attribute for that instance is used as the location
+            to delete
+        """
+        # Create a ParseResult from the uri
+        if isinstance(location, (InputFile, OutputFile)):
+            str_location = location.location  # Use InputFile or OutputFile location
+        else:
+            str_location = location
+
+        self._s3_filesystem.rm(str_location)
```
rdblue, line 235: Should we explicitly pass `recursive=False`? I know it's the default but it doesn't seem like something we want to mess up if the underlying library changes.

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+            raise FileExistsError(f"Cannot create file, file already exists: {self.location}")
+        self._s3.touch(self.location)
+        return S3fsOutputStream(s3_object=self._s3.open(self.location, "wb"))
+
+    def to_input_file(self) -> S3fsInputFile:
+        """Returns a new S3fsInputFile for the location at `self.location`"""
+        return S3fsInputFile(location=self.location, s3=self._s3)
+
+
+class S3fsFileIO(FileIO):
```
rdblue, line 190: This appears like it would support all `fsspec` file systems. I think we should probably change this to `FSSpecFileIO` instead of `S3FileIO`.

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+        super().__init__(location=location)
+
+    def __len__(self) -> int:
+        """Returns the total length of the file, in bytes"""
+        object_info = self._s3.info(self.location)
+        if object_info.get("Size"):  # s3fs versions seem inconsistent on the case used for size
+            return object_info["Size"]
+        elif object_info.get("size"):
+            return object_info["size"]
+        raise RuntimeError(f"Cannot retrieve object info: {self.location}")
```
rdblue, line 156: Should we have a base implementation to dedup?

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+        Note:
+            If overwrite is set to False, a check is first performed to verify that the file does not exist.
+            This is not thread-safe and a possibility does exist that the file can be created by a concurrent
+            process after the existence check yet before the output stream is created. In such a case, the default
+            behavior will truncate the contents of the existing file when opening the output stream.
+        """
+        if not overwrite and self.exists():
+            raise FileExistsError(f"Cannot create file, file already exists: {self.location}")
+        self._s3.touch(self.location)
+        return S3fsOutputStream(s3_object=self._s3.open(self.location, "wb"))
```
rdblue, line 183: What is thrown if this is called when the file already exists? There's a race condition if we check `exists`, so it would be great if we could just rely on an exception and re-raise the right one.

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+from pyiceberg.io.base import (
+    FileIO,
+    InputFile,
+    InputStream,
+    OutputFile,
+    OutputStream,
+)
+
+
+class S3fsInputStream(InputStream):
```
Fokko, line 32: Ideally, we don't want this class since it is almost a copy of the Stream of S3FS itself.

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        return self._s3_object.read(length=size)
+
+    def seek(self, offset: int, whence: int = 0) -> int:
+        return self._s3_object.seek(loc=offset, whence=whence)
+
+    def tell(self) -> int:
+        return self._s3_object.tell()
+
+    def closed(self) -> bool:
```
Fokko, line 59: We don't have closed anymore (this is actually a property, but did not play well with the protocol).

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+        return self._s3_object.closed
+
+    def close(self) -> None:
+        self._s3_object.close()
+
+
+class S3fsOutputStream(OutputStream):
```
Fokko, line 66: Same as above, the calls map 1-to-1.

Fokko (commented):
Great work @samredai, a few comments, but looks great in general 👍🏻

python/pyiceberg/io/s3fs.py
rdblue: Even if it isn't in the protocol, doesn't it make sense to have a method that exposes this since the underlying implementation does?

We may want to make this a `@property`, though. That's what is normal in Python and why we couldn't include it in the Protocol.

python/pyiceberg/io/s3fs.py
rdblue: I agree. As far as I can tell, this is a pass-through for the stream from fsspec. Can we just return that stream instead?

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+
+    def __init__(self, location: str, s3):
+        self._s3 = s3
+        super().__init__(location=location)
+
+    def __len__(self) -> int:
+        """Returns the total length of the file, in bytes"""
+        object_info = self._s3.info(self.location)
+        if size := object_info.get("Size"):  # s3fs versions seem inconsistent on the case used for size
```
rdblue, line 116: Since you're using the walrus operator, you can just return `size`.

python/pyiceberg/io/s3fs.py
```diff
@@ -0,0 +1,235 @@
...
+            return object_info["Size"]
+        elif object_info.get("size"):
+            return object_info["size"]
```
rdblue, line 119: Should this be `elif size := object_info.get("size"):` to match the first case?

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+
+    def __init__(self, location: str, fs):
+        self._fs = fs
+        super().__init__(location=location)
+
+    def __len__(self) -> int:
+        """Returns the total length of the file, in bytes"""
+        object_info = self._fs.info(self.location)
+        if object_info.get("Size"):  # fsspec versions seem inconsistent on the case used for size
```
samredai, line 75:
```suggestion
        if object_info.get("Size"):
```

samredai:
@Fokko @rdblue I've updated this to `FsspecFileIO` that uses s3fs for locations with an s3 scheme. A couple questions I have is:
1. The class caches the first filesystem it initializes when a new input file or output file is requested using that uri. Do we want to handle a situation where a single FsspecFileIO instance is handling locations of varying schemes?
2. In order to use the `Dict[str, str]` type that's been added to the parent FileIO interface, I've set a `client_` prefix for client kwargs and `fs_` prefix for filesystem kwargs. They're then plucked out in the `_fs_properties()` and `_client_properties()` methods. Should this be where we also look for the java property keys and plug them in? For example check for `s3.endpoint` and plug that into the client kwargs as `endpoint_url`.

python/pyproject.toml
```diff
@@ -78,6 +80,12 @@ pyarrow = ["pyarrow"]
 snappy = ["python-snappy"]
 python-snappy = ["zstandard"]
 hive = ["thrift"]
+s3fs = ["s3fs"]
```
Fokko, line 83: I think we want to install `fsspec` always, and then we have the option to install different implementations `s3fs`, `gcsfs`, `adlfs` etc

python/tests/io/test_fsspec.py
```diff
@@ -0,0 +1,191 @@
...
+    assert isinstance(output_file, fsspec.FsspecOutputFile)
+    assert output_file.location == f"s3://testbucket/{filename}"
+
+
+@pytest.mark.s3
+def test_fsspec_write_and_read_file(fsspec_fileio):
+    """Test writing and reading a file using FsspecInputFile and FsspecOutputFile"""
+    filename = str(uuid.uuid4())
+    output_file = fsspec_fileio.new_output(location=f"s3://testbucket/{filename}")
+    f = output_file.create()
```
Fokko, line 52: Can you use the context manager in the tests? Just to give a good example:
```python
with output_file.create() as f:
    f.write(b"foo")
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+
+class FsspecInputFile(InputFile):
+    """An input file implementation for the FsspecFileIO
+
+    Args:
+        location(str): A URI to a file location
+        fs: An fsspec filesystem instance
+    """
+
+    def __init__(self, location: str, fs):
```
Fokko, line 34:
```suggestion
    def __init__(self, location: str, fs: AbstractFileSystem):
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class FsspecInputFile(InputFile):
+    """An input file implementation for the FsspecFileIO
+
+    Args:
+        location(str): A URI to a file location
+        fs: An fsspec filesystem instance
+    """
+
```
Fokko, line 33: _fs: AbstractFileSystem

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+            from s3fs import S3FileSystem
+
+            self._fs = S3FileSystem(client_kwargs=self._client_properties() or None, **self._fs_properties())
+            return self._fs
+        else:
+            raise ValueError(f"Cannot determine fsspec implementation to use with scheme: {scheme}")
+
+    def _fs_properties(self):
+        """Get fs properties from the file-io property map"""
+        return {k[3:]: v for k, v in self.properties.items() if k.startswith("fs_")}
```
Fokko, line 178: If we want to do it like this, then we should do something like:
```suggestion
FS_ = 'fs_'

...

        return {k[len(FS_):]: v for k, v in self.properties.items() if k.startswith(FS_)}
```
But I think we can make this nicer. In the configuration, we have a recursive dictionairy.

This would allow us to set properties like:
```
catalog:
  production:
   uri: https://...
   credentials: token
   py-file-io: pyiceberg.io.fsspec.FsspecFileIO
   fs:
     access_key: abc
     secret_key: def
   client_kwargs:
     client_url: https://...
```

We could leverage that to avoid parsing all the properties. WDYT?

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
+            str_location = location
+
+        uri = urlparse(str_location)
+        fs = self._fs or self._resolve_fs_type(scheme=uri.scheme)
+        fs.rm(str_location)
+
+    def _resolve_fs_type(self, scheme: str):
+        """Initialize the proper filesystem for the location's scheme"""
+        if scheme in ("s3", "s3n", "s3a"):
```
Fokko, line 168:
```suggestion
        if scheme in {"s3", "s3n", "s3a"}:
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+            from s3fs import S3FileSystem
+
+            self._fs = S3FileSystem(client_kwargs=self._client_properties() or None, **self._fs_properties())
```
Fokko, line 171: If we want to leverage the configuration, then we can drop the constructor, and replace this by:
```suggestion
            self._fs = S3FileSystem(client_kwargs=self.properties.get('client_kwargs'), **self.properties.get('fs', {}))
```

rdblue:
> Do we want to handle a situation where a single FsspecFileIO instance is handling locations of varying schemes?

Yes. I think that we want to have a mapping from URI scheme to file system, like we do for catalog URI. The FSSpecFileIO should use the correct backing FileSystem for a given URI. I think that we should also allow plugging in mappings and overriding.

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Note:
+            If overwrite is set to False, a check is first performed to verify that the file does not exist.
+            This is not thread-safe and a possibility does exist that the file can be created by a concurrent
+            process after the existence check yet before the output stream is created. In such a case, the default
+            behavior will truncate the contents of the existing file when opening the output stream.
+        """
+        if not overwrite and self.exists():
+            raise FileExistsError(f"Cannot create file, file already exists: {self.location}")
+        self._fs.touch(self.location)
```
rdblue, line 105: Why is this needed?

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def to_input_file(self) -> FsspecInputFile:
+        """Returns a new FsspecInputFile for the location at `self.location`"""
+        return FsspecInputFile(location=self.location, fs=self._fs)
+
+
+class FsspecFileIO(FileIO):
+    """A FileIO implementation that uses fsspec"""
+
+    def __init__(self, client_kwargs: Dict[str, str] = EMPTY_DICT, **fs_kwargs):
```
rdblue, line 116: I think this should be `def __init__(self, **properties: str):` That matches what we use for catalogs.

This is also where we want to pull out properties to configure the `FileIO`. Since the properties will differ between file systems, I think you may want to just pass this map to a method that instantiates the right file system:

```python
def _s3(properties: Properties) -> FileSystem:
    from s3fs import S3FileSystem

    # build client_kwargs from properties
    client_kwargs = ...

    return S3FileSystem(client_kwargs=client_kwargs, ...)

SCHEME_TO_FS = {
    "s3": _s3,
    "s3a": _s3,
    "s3n": _s3,
    ...
}
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,182 @@
...
+        """Get an FsspecInputFile instance to read bytes from the file at the given location
+
+        Args:
+            location(str): A URI or a path to a local file
+
+        Returns:
+            FsspecInputFile: An FsspecInputFile instance for the given location
+        """
+        uri = urlparse(location)
+        fs = self.properties.get("fs") or self._resolve_fs_type(scheme=uri.scheme)
```
rdblue, line 133: I don't think it should be possible to pass in a specific FS. Callers should register a method to create a FS (even if that just returns a configured one). I'm not sure if that should be done in the constructor or not, since we are passing properties as `kwargs`.

@Fokko, should we pass properties as `properties: Properties` rather than `**properties: str` so that we can pass other optional arguments?

python/pyiceberg/io/fsspec.py
rdblue: I think we need to construct `client_kwargs` rather than passing them through. That way we can make configuration the same for both Java and Python FileIO.

python/pyiceberg/io/fsspec.py
rdblue: I agree that we should use configuration, but I think we should limit it to `Properties` so that it is compatible with the properties used by Java. The config for FileIO would be:

```yaml
catalog:
  production:
    uri: https://...
    credential: token
    s3.endpoint: https://...
    s3.access-key-id: abc
    s3.secret-access-key: def
```

python/pyproject.toml
samredai: Added `fsspec` as a dependency 👍

python/tests/io/test_fsspec.py
samredai: Updated! Changed all the tests to use context managers. Thanks @Fokko

python/pyiceberg/io/fsspec.py
samredai: Fixed!

python/pyiceberg/io/fsspec.py
samredai: So should we not allow passing in additional filesystem-specific kwargs? [Here's](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem) a list of the s3fs kwargs which has stuff like `default_block_size` and `version_awarebool`.

python/pyiceberg/io/fsspec.py
samredai: Updated to explicitly construct client kwargs using the example in your [other comment](https://github.com/apache/iceberg/pull/5332#discussion_r956257603).

python/pyiceberg/io/fsspec.py
samredai: You're right this isn't needed, removed. 👍

python/pyiceberg/io/fsspec.py
samredai: Updated to use this, thanks @rdblue!

python/pyiceberg/io/fsspec.py
samredai: `properties: Properties` is what the parent class `FileIO` uses. I used that here and it worked nicely. @Fokko let me know if that works.

python/pyiceberg/io/fsspec.py
rdblue: Was it? I think what @Fokko wanted was a declaration of the name, like this:

```python
class FsspecInputFile(InputFile):
    """ docs """

    _fs: AbstractFileSystem    # <-- this

    def __init__(self, location: str, fs: AbstractFileSystem):
        ...
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,201 @@
...
+from pyiceberg.typedef import EMPTY_DICT, Properties
+
+
+def _s3(properties: Properties, **fs_properties) -> AbstractFileSystem:
+    from s3fs import S3FileSystem
+
+    client_kwargs = {
+        "endpoint_url": properties.get("s3.endpoint"),
+        "aws_access_key_id": properties.get("s3.access-key-id"),
+        "aws_secret_access_key": properties.get("s3.secret-access-key"),
```
rdblue, line 34: Are there any others that we should support?

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,201 @@
...
+        Args:
+            location(str): A URI or a path to a local file
+
+        Returns:
+            FsspecOutputFile: An FsspecOutputFile instance for the given location
+        """
+        uri = urlparse(location)
+        if uri.scheme not in self._scheme_to_fs:
+            raise ValueError(f"No registered filesystem for scheme: {uri.scheme}")
+        fs = self._scheme_to_fs[uri.scheme](self.properties, **self._fs_properties())
```
rdblue, line 167: I think these need to be cached. It looks like this is going to create a new `S3FileSystem` for every input or output file. We should build just one and reuse it every time we see the scheme.

python/pyiceberg/io/fsspec.py
rdblue: We can create a prefix for passing implementation-specific arguments.

python/tests/io/test_fsspec.py
```diff
@@ -0,0 +1,186 @@
...
+    f = input_file.open()
+    assert not f.closed
+    f.close()
+    assert f.closed
+
+    fsspec_fileio.delete(f"s3://warehouse/{filename}")
+
+
+@pytest.mark.s3
+def test_fsspec_converting_an_outputfile_to_an_inputfile(fsspec_fileio):
```
rdblue, line 181: Can we have a test that writes an Avro file and reads it back? It would be good to have a bit more real world test cases.

samredai: Good idea, I added a test called `test_writing_avro_file` that reads a local avro file, writes it to a location with an `s3` scheme using fsspec-file-io, then reads it back and asserts that the bytes match.

python/pyiceberg/io/fsspec.py
samredai: Ah right, I added `self._fs_cache: Dict[str, AbstractFileSystem] = {}` to the `FsspecFileIO` class to cache the filesystem based on the uri scheme and then I check that first before creating a new one.

python/pyiceberg/io/fsspec.py
samredai: Some other options that may be good to add are:
- **aws_session_token** (string) -- The session token to use when creating the client. Same semantics as aws_access_key_id above.
- **use_ssl** (boolean) -- Whether or not to use SSL. By default, SSL is used. Note that not all services support non-ssl connections.
- **region_name** (string) -- The name of the region associated with the client. A client is associated with a single region.

There's also an `s3_additional_kwargs` map that lets you set things like sse. I'll look closer at those options and map them to the java ones.

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Returns:
+            FsspecInputFile: An FsspecInputFile instance for the given location
+        """
+        uri = urlparse(location)
+
+        if uri.scheme not in self._scheme_to_fs:
+            raise ValueError(f"No registered filesystem for scheme: {uri.scheme}")
+
+        if cached_fs := self._fs_cache.get(uri.scheme):
```
Fokko, line 156: I think it would be better to cache this using `functools`: https://docs.python.org/3/library/functools.html#functools.cache Moving to a separate function `def _get_fs(scheme: str)` would just cache for each fo the schemas.

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,211 @@
...
+    "s3n": _s3,
+}
+
+
+class FsspecInputFile(InputFile):
+    """An input file implementation for the FsspecFileIO
+
+    Args:
+        location(str): A URI to a file location
+        _fs(AbstractFileSystem): An fsspec filesystem instance
```
Fokko, line 52:
```suggestion
        fs(AbstractFileSystem): An fsspec filesystem instance
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        return self._fs.open(self.location, "rb")
+
+
+class FsspecOutputFile(OutputFile):
+    """An output file implementation for the FsspecFileIO
+
+    Args:
+        location(str): A URI to a file location
+        fs: An fsspec filesystem instance
```
Fokko, line 86:
```suggestion
        fs(AbstractFileSystem): An fsspec filesystem instance
```

python/pyiceberg/io/fsspec.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+
+    def __init__(self, location: str, fs):
```
Fokko, line 89:
```suggestion
    def __init__(self, location: str, fs: AbstractFileSystem):
```

python/pyiceberg/io/fsspec.py
Fokko: The code of initializing an FS is also defined both in the InputFile, OutputFile and delete, would be nice if we can consolidate this as well.

python/pyiceberg/io/fsspec.py
Fokko: If we don't want to mix certain properties, then we should pass in different dicts. I like how we have it now. We could even use TypedDicts in the future if we want to be more explicit: https://peps.python.org/pep-0589/

Fokko (commented):
Some small stuff, looking good @samredai

python/pyiceberg/io/fsspec.py
samredai: Bugbear flagged this with:

> B019: Use of functools.lru_cache or functools.cache on methods can lead to memory leaks. The cache may retain instance references, preventing garbage collection.

The solution I used was to just initialize a cache in the class init and add the instance method to it. This localizes the cache so that it's destroyed with the instance.
```py
self.get_fs: Callable = lru_cache(self._get_fs)
```

Let me know what you think.

Fokko (approved):
@samredai can you resolve the conflicts? :)

Fokko:
Thanks @samredai 👍🏻

Issue #5333, opened 2022-07-22
Docs: Add doc of the upsert option


a49a:
Currently, there isn't `write.upsert.enabled` in the doc, we need to add it.

PR #5334, opened 2022-07-22
Docs: Add doc of the upsert option

a49a:
Issue link: https://github.com/apache/iceberg/issues/5333

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -79,6 +79,7 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.object-storage.enabled       | false              | Enables the object storage location provider that adds a hash component to file paths |
 | write.data.path                    | table location + /data | Base location for data files |
 | write.metadata.path                | table location + /metadata | Base location for metadata files |
+| write.upsert.enabled               | false              | Enables upsert mode (currently flink only) |
 | write.delete.mode                  | copy-on-write      | Mode used for delete commands: copy-on-write or merge-on-read (v2 only) |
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -79,6 +79,7 @@ Iceberg tables support table properties to configure table behavior, like the de
 | write.object-storage.enabled       | false              | Enables the object storage location provider that adds a hash component to file paths |
 | write.data.path                    | table location + /data | Base location for data files |
 | write.metadata.path                | table location + /metadata | Base location for metadata files |
+| write.upsert.enabled               | false              | Enables upsert mode (currently flink only) |
```
rdblue, line 82: I think this should be in a Flink section, not in general table properties. Is there a Flink section for these?

deadwind4: https://github.com/apache/iceberg/blob/5951ece2cedf82c69ae717a20c6b392b0b76237c/docs/flink-getting-started.md?plain=1#L570

The `upsert-enabled` option in the Flink section is not general table property, but the `write.upsert.enabled` option is general table property. Maybe, in the future, we will implement `write.upsert.enabled` in Spark.

The `upsert-enabled` option has been in the doc, but there is no `write.upsert.enabled` in the doc. It could get confused for users.

rdblue: I think it's a problem that this is a general property that is unsupported by all engines other than Flink.

Since we can't really fix the property name, let's move it to a Flink section so it is clear that only Flink supports it.

deadwind4: I have moved this option into the Flink section.

deadwind4:
@rdblue I have moved this option into the Flink section.

PR #5335, opened 2022-07-22
Ignore case for partition transform

southernriver:
Transform is not supported for upper case, otherwise throw SQLException as follow:
```
java.sql.SQLException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.lang.UnsupportedOperationException: Transform is not supported: HOURS(timestamp_)
at org.apache.iceberg.spark.Spark3Util.toPartitionSpec(Spark3Util.java:423)
at org.apache.iceberg.spark.SparkCatalog.createTable(SparkCatalog.java:156)
at org.apache.iceberg.spark.SparkCatalog.createTable(SparkCatalog.java:98)
at org.apache.iceberg.spark.SparkSessionCatalog.createTable(SparkSessionCatalog.java:139)
at org.apache.spark.sql.execution.datasources.v2.CreateTableExec.run(CreateTableExec.scala:41)
```

Since Spark side never check upper or lower case and SQL parse is ok,  we could ignore case for iceberg side.

### DIFF
--- a/spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/Spark3Util.java
+++ b/spark/v3.0/spark/src/main/java/org/apache/iceberg/spark/Spark3Util.java
@@ -23,6 +23,7 @@ import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


rdblue (approved):

rdblue:
Thanks, @southernriver! Good catch.

PR #5337, opened 2022-07-22
Core: Print DateTime strings always with +00:00 Zone offset

### LINKED ISSUE #5321: The error message of 'org.apache.iceberg.util.SnapshotUtil#snapshotIdAsOfTime' can be misleading to users.
https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L320

The error message of 'org.apache.iceberg.util.SnapshotUtil#snapshotIdAsOfTime' shows the timestamp in UTC timezone, this can be misleading to users who is in another timezone.

e.g:
When I snapshotIdAsOfTime('2022-06-17 11:00:00'), the error message show me 'Cannot find a snapshot older than 2022-06-17 09:00:00', because I'm not in UTC timezone.

nastra:
This should make it fairly clear that all timestamps are in UTC.
Additionally, this makes also sure that a timestamp string can be parsed
by `ZonedDateTime.parse(timestamp)`.

fixes #5321

### DIFF
--- a/core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
+++ b/core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
@@ -25,14 +25,11 @@ import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
-import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 
 public class DateTimeUtil {
   private DateTimeUtil() {
   }
-
-  private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS");
 
   public static final OffsetDateTime EPOCH = Instant.ofEpochSecond(0).atOffset(ZoneOffset.UTC);
   public static final LocalDate EPOCH_DAY = EPOCH.toLocalDate();
@@ -86,6 +83,6 @@ public class DateTimeUtil {
   }
 
   public static String formatTimestampMillis(long millis) {
-    return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC));
+    return Instant.ofEpochMilli(millis).toString().replace("Z", "+00:00");
   }
 }
--- a/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImpl.java
+++ b/flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImpl.java
@@ -38,7 +38,7 @@ import org.apache.iceberg.flink.source.StreamingStartingStrategy;
 import org.apache.iceberg.flink.source.split.IcebergSourceSplit;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
-import org.joda.time.format.DateTimeFormat;
+import org.apache.iceberg.util.DateTimeUtil;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.ClassRule;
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
```diff
@@ -86,6 +83,10 @@ public static long microsFromTimestamptz(OffsetDateTime dateTime) {
   }
 
   public static String formatTimestampMillis(long millis) {
-    return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC));
+    return Instant.ofEpochMilli(millis)
+        .atOffset(ZoneOffset.UTC)
+        .toZonedDateTime()
+        .toString()
+        .replace("Z", "+00:00");
```
dimas-b, line 90: Why not keep `Z`? It's a standard designation for UTC, I think :thinking:

core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
```diff
@@ -86,6 +83,10 @@ public static long microsFromTimestamptz(OffsetDateTime dateTime) {
   }
 
   public static String formatTimestampMillis(long millis) {
-    return DATE_FORMAT.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC));
+    return Instant.ofEpochMilli(millis)
+        .atOffset(ZoneOffset.UTC)
```
dimas-b, line 87: `Instant.ofEpochMilli(...).toString()` already formats in UTC, why bother with `.atOffset(UTC)`?

core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
nastra: see https://github.com/apache/iceberg/issues/5321#issuecomment-1191968188 as a reference.

core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
nastra: you're right. updated

core/src/main/java/org/apache/iceberg/util/DateTimeUtil.java
dimas-b: ok, both notations are standard, but `Z` is simpler, IMHO :)

dimas-b (approved):

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImpl.java
```diff
@@ -396,7 +396,7 @@ public void testIncrementalFromSnapshotTimestampWithEmptyTable() throws Exceptio
 
     AssertHelpers.assertThrows("Should detect invalid starting snapshot timestamp",
         IllegalArgumentException.class,
-        "Cannot find a snapshot older than 1970-01-01 00:00:00.001",
+        "Cannot find a snapshot older than 1970-01-01T00:00:00.001+00:00",
```
rdblue, line 399: Do we want to add the `T`? Seems like that makes it harder to read. Since this string is primarily for people, making it easy to read seems worth not conforming to ISO-8601 exactly. Maybe we should add a space before the zone offset, too?

nastra: It feels more natural to adhere to the ISO standard as that is pretty well understood. Also people could really just copy/paste this (not sure if anyone would actually do that) and parse it. I don't see any clear benefit to not adhere to the ISO standard, but happy to change if others think we should use something non-standard here.

dimas-b: +1 to ISO 8601.

Also the `T` avoids a space, which IMHO makes copy-paste and automated processing easier.

rdblue (approved):

flink/v1.14/flink/src/test/java/org/apache/iceberg/flink/source/enumerator/TestContinuousSplitPlannerImpl.java
rdblue: Seems reasonable to me.

rdblue:
Merged. Thanks, @nastra!

PR #5338, opened 2022-07-22
Spec: Inconsistency around files_count

### LINKED ISSUE #8684: Spec disagrees with implementation on field names
### Apache Iceberg version

1.3.1 (latest release)

### Query engine

Spark

### Please describe the bug 🐞

The [spec](https://github.com/apache/iceberg/blob/3f2884c6d2e514f1112bd440fedfbb7d30227a96/format/spec.md?plain=1#L545) calls the following fields as follows:

 504 added_files_count 
505 existing_files_count 
506 deleted_files_count

The [implementation](https://github.com/apache/iceberg/blob/3f2884c6d2e514f1112bd440fedfbb7d30227a96/api/src/main/java/org/apache/iceberg/ManifestFile.java#L55) calls them as follows:

    optional(504, "added_data_files_count"
      optional(505, "existing_data_files_count"
      optional(506, "deleted_data_files_count"
      
   
   (The `_data` is in the implementation, but not in the spec)

And hence the produced iceberg use these field names, which deviate from the names in the spec. Either the spec or the implementation should be updated.

Fokko:
When building the Manifest mappers for Python, @rdblue  noticed that the `added_data_files_count` should be `added_files_count` according to the spec. 

However, this field is written in Java as `added_data_files_count` to Avro by the Java implementation:  
https://github.com/apache/iceberg/blob/8104769f81ba79338fd3c94d5bd9267f22d31ed7/api/src/main/java/org/apache/iceberg/ManifestFile.java#L44-L49

Luckily this doesn't affect the reading/writing because it is position based.

However, it is confusing. I think we should resolve this. We could either do this by changing the Java impl, which probably works, but we could also change the spec. I know that this isn't something lightweight, but we could take it into consideration.

I think we should also update the references in the code to `{added,existing,deleted}_data_files_count` to make everything consistent and avoid confusion in the future.

Closes #8684

### DIFF
--- a/api/src/main/java/org/apache/iceberg/ManifestFile.java
+++ b/api/src/main/java/org/apache/iceberg/ManifestFile.java
@@ -52,11 +52,11 @@ public interface ManifestFile {
       optional(
           503, "added_snapshot_id", Types.LongType.get(), "Snapshot ID that added the manifest");
   Types.NestedField ADDED_FILES_COUNT =
-      optional(504, "added_data_files_count", Types.IntegerType.get(), "Added entry count");
+      optional(504, "added_files_count", Types.IntegerType.get(), "Added entry count");
   Types.NestedField EXISTING_FILES_COUNT =
-      optional(505, "existing_data_files_count", Types.IntegerType.get(), "Existing entry count");
+      optional(505, "existing_files_count", Types.IntegerType.get(), "Existing entry count");
   Types.NestedField DELETED_FILES_COUNT =
-      optional(506, "deleted_data_files_count", Types.IntegerType.get(), "Deleted entry count");
+      optional(506, "deleted_files_count", Types.IntegerType.get(), "Deleted entry count");
   Types.NestedField ADDED_ROWS_COUNT =
       optional(512, "added_rows_count", Types.LongType.get(), "Added rows count");
   Types.NestedField EXISTING_ROWS_COUNT =
--- a/core/src/test/java/org/apache/iceberg/TestManifestListVersions.java
+++ b/core/src/test/java/org/apache/iceberg/TestManifestListVersions.java
@@ -162,12 +162,11 @@ public class TestManifestListVersions {
     Assert.assertEquals("Length", LENGTH, generic.get("manifest_length"));
     Assert.assertEquals("Spec id", SPEC_ID, generic.get("partition_spec_id"));
     Assert.assertEquals("Snapshot id", SNAPSHOT_ID, (long) generic.get("added_snapshot_id"));
+    Assert.assertEquals("Added files count", ADDED_FILES, (int) generic.get("added_files_count"));
     Assert.assertEquals(
-        "Added files count", ADDED_FILES, (int) generic.get("added_data_files_count"));
+        "Existing files count", EXISTING_FILES, (int) generic.get("existing_files_count"));
     Assert.assertEquals(
-        "Existing files count", EXISTING_FILES, (int) generic.get("existing_data_files_count"));
-    Assert.assertEquals(
-        "Deleted files count", DELETED_FILES, (int) generic.get("deleted_data_files_count"));
+        "Deleted files count", DELETED_FILES, (int) generic.get("deleted_files_count"));
     Assert.assertEquals("Added rows count", ADDED_ROWS, (long) generic.get("added_rows_count"));
     Assert.assertEquals(
         "Existing rows count", EXISTING_ROWS, (long) generic.get("existing_rows_count"));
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/apache/iceberg/ManifestFile.java
```diff
@@ -41,11 +41,11 @@ public interface ManifestFile {
       "Lowest sequence number in the manifest");
   Types.NestedField SNAPSHOT_ID = optional(503, "added_snapshot_id", Types.LongType.get(),
       "Snapshot ID that added the manifest");
-  Types.NestedField ADDED_FILES_COUNT = optional(504, "added_data_files_count", Types.IntegerType.get(),
+  Types.NestedField ADDED_DATA_FILES_COUNT = optional(504, "added_data_files_count", Types.IntegerType.get(),
```
rdblue, line 44: These changes look good to me. I think we can also update the field names. Those won't break anything because the only names that matter are the ones in the read schema.

api/src/main/java/org/apache/iceberg/ManifestFile.java
```diff
@@ -41,11 +41,11 @@ public interface ManifestFile {
       "Lowest sequence number in the manifest");
   Types.NestedField SNAPSHOT_ID = optional(503, "added_snapshot_id", Types.LongType.get(),
       "Snapshot ID that added the manifest");
-  Types.NestedField ADDED_FILES_COUNT = optional(504, "added_data_files_count", Types.IntegerType.get(),
```
aokolnychyi, line 44: Ugh, is this now used to track the number of added delete files in our delete manifests too?

Fokko: It was already mandatory for v2. But this was already the case.

rdblue: > Ugh, is this now used to track the number of added delete files in our delete manifests too?

Yeah, we decided to use the existing metadata fields for the same purpose in delete manifests. I think it makes sense -- is there a problem?

.palantir/revapi.yml
```diff
@@ -1,6 +1,17 @@
...
+    org.apache.iceberg:iceberg-api:
+    - code: "java.field.removed"
+      old: "field org.apache.iceberg.ManifestFile.ADDED_FILES_COUNT"
+      justification: "Removes inconsistency between code and spec"
+    - code: "java.field.removed"
+      old: "field org.apache.iceberg.ManifestFile.DELETED_FILES_COUNT"
+      justification: "Removes inconsistency between code and spec"
+    - code: "java.field.removed"
+      old: "field org.apache.iceberg.ManifestFile.EXISTING_FILES_COUNT"
+      justification: "Removes inconsistency between code and spec"
```
rdblue, line 14: Since we are going to be releasing 1.0 and can't deprecate these properly, can we leave the static field name intact and make it an alias for the correct one?

Fokko: I like that a lot, thanks 👍🏻

rdblue:
@Fokko, sorry, but I think I was wrong in my initial review. I think instead of renaming these, we actually want to rename the fields that we write so that they match the spec. Because we use the same column for both `DataFile` and `DeleteFile` to count the number of `ADDED` rows, we should use a neutral name so `added_files_count` is correct.

Looks like the problem here is that the original `added_data_files_count` is used instead of `added_files_count`, not that the variable names don't match.

api/src/main/java/org/apache/iceberg/ManifestFile.java
```diff
@@ -52,11 +52,11 @@ public interface ManifestFile {
...
           503, "added_snapshot_id", Types.LongType.get(), "Snapshot ID that added the manifest");
   Types.NestedField ADDED_FILES_COUNT =
-      optional(504, "added_data_files_count", Types.IntegerType.get(), "Added entry count");
+      optional(504, "added_files_count", Types.IntegerType.get(), "Added entry count");
   Types.NestedField EXISTING_FILES_COUNT =
-      optional(505, "existing_data_files_count", Types.IntegerType.get(), "Existing entry count");
+      optional(505, "existing_files_count", Types.IntegerType.get(), "Existing entry count");
   Types.NestedField DELETED_FILES_COUNT =
-      optional(506, "deleted_data_files_count", Types.IntegerType.get(), "Deleted entry count");
+      optional(506, "deleted_files_count", Types.IntegerType.get(), "Deleted entry count");
```
ajantha-bhat, line 59: I thought we change the spec document instead of changing the schema (field names) here.
I saw the comments down below. Agree that this doesn't break compatibility. 

But If we follow this path, `manifest_entry` has a field called `data_file` which represents both data and delete file (https://github.com/apache/iceberg/blame/master/format/spec.md#L421) May be we need to rename that aswell?

Fokko: Hey @ajantha-bhat Thanks, that's actually a good point. However, If we go that route, I'd prefer to do that separately because of two reasons:
- Looking at the code, I also see that we occasionally fetch the struct by name.
- The code and spec are consistent, so we have to change them both

ajantha-bhat: ok. Sounds reasonable to me.

rdblue (approved):

Fokko:
Thanks for the review @rdblue 🙌

PR #5340, opened 2022-07-22
Python: Add dependabot for keeping the dependencies up to date

Fokko:
Avoid me having to bump the versions once a while: https://github.com/apache/iceberg/pull/5325

Dependabot is much better at this, and will also check for CVEs and such.

Maybe Python is nice to give this a try.

cc @nastra

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


nastra (approved):
works in my fork, so +1

Fokko:
An example how it looks on @nastra his fork: https://github.com/nastra/iceberg/pull/16

rdblue (approved):

rdblue:
Looks good to me. We'll want to be careful when we merge updates, though. I think we generally want to only update dependencies in point releases, and avoid breaking updates between major releases. I'm sure that's normal, so maybe it's already handled by dependabot?

PR #5345, opened 2022-07-23
Parquet: Add option to set page row count limit

bryanck:
Iceberg v0.14 uses Parquet 1.12, which limits page sizes by row count, with the default set to 20k. In some cases where data compresses well, this can cause many more pages to be created than with older versions of Parquet, which can reduce compression and scan performance.

This PR introduces a table property that allows setting the page row count limit, with a default of 20k.


core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -143,6 +143,10 @@ private TableProperties() {
   public static final String DELETE_PARQUET_PAGE_SIZE_BYTES = "write.delete.parquet.page-size-bytes";
   public static final int PARQUET_PAGE_SIZE_BYTES_DEFAULT = 1024 * 1024; // 1 MB
 
+  public static final String PARQUET_PAGE_ROW_COUNT_LIMIT = "write.parquet.page-row-count-limit";
```
rdblue, line 146: Could this be `page-row-limit` instead of `count-limit`? I think count is probably unnecessary.

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -143,6 +143,10 @@ private TableProperties() {
...
+  public static final String DELETE_PARQUET_PAGE_ROW_COUNT_LIMIT = "write.delete.parquet.page-row-count-limit";
+  public static final int PARQUET_PAGE_ROW_COUNT_LIMIT_DEFAULT = 20_000;
```
rdblue, line 148: Is this the Parquet default? How did you decide on 20k?

docs/configuration.md
```diff
@@ -46,45 +46,46 @@ Iceberg tables support table properties to configure table behavior, like the de
...
+| write.update.mode                              | copy-on-write              | Mode used for update commands: copy-on-write or merge-on-read (v2 only)                                                                                                                           |
+| write.update.isolation-level                   | serializable               | Isolation level for update commands: serializable or snapshot                                                                                                                                     |
+| write.merge.mode                               | copy-on-write              | Mode used for merge commands: copy-on-write or merge-on-read (v2 only)                                                                                                                            |
+| write.merge.isolation-level                    | serializable               | Isolation level for merge commands: serializable or snapshot                                                                                                                                      |
```
rdblue, line 88: Can you just update the rows needed for this change? We try to avoid reformatting tables unless they're really messy.

core/src/main/java/org/apache/iceberg/TableProperties.java
bryanck: Yes, that's the Parquet default, so this shouldn't change any behavior if it isn't set.

docs/configuration.md
bryanck: Sure this is done (my editor got a little aggressive)

rdblue (approved):

core/src/main/java/org/apache/iceberg/TableProperties.java
bryanck: This is done

core/src/main/java/org/apache/iceberg/TableProperties.java
bryanck: [Here](https://github.com/apache/parquet-mr/blob/0819356a9dafd2ca07c5eab68e2bffeddc3bd3d9/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java#L58) is where the default it set:

core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -143,6 +143,10 @@ private TableProperties() {
   public static final String DELETE_PARQUET_PAGE_SIZE_BYTES = "write.delete.parquet.page-size-bytes";
   public static final int PARQUET_PAGE_SIZE_BYTES_DEFAULT = 1024 * 1024; // 1 MB
 
+  public static final String PARQUET_PAGE_ROW_LIMIT = "write.parquet.page-row-limit";
+  public static final String DELETE_PARQUET_PAGE_ROW_LIMIT = "write.delete.parquet.page-row-limit";
```
stevenzwu, line 147: what about `page-row-count-limit`?

Just `row-limit` can also be interpreted as row size/bytes limit.

bryanck: That's what I had originally (and how it is defined in parquet-mr), @rdblue preferred this naming. I'm OK with either one.

stevenzwu: sorry. missed the earlier context. `page-row-count-limit` is a little more clear to me.

rdblue: I think it is pretty clear from context that this is the number of rows and not the size of a row. For these formats, types can be size limited, but rows aren't.

danielcweeks: I think `page-row-limit` is clear enough.  There is also a page size limit, but I feel the count is implied given it's rows.

rdblue (approved):

rdblue:
Thanks, @bryanck!

PR #5346, opened 2022-07-24
Substitute the method of PropertyUtil#propertyAsLong for IcebergSource#propertyAsLong

zhaomin1423:
The IcebergSource#propertyAsLong is redundant, it can be replaced by PropertyUtil#propertyAsLong.

### DIFF
--- a/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/IcebergSource.java
+++ b/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/IcebergSource.java
@@ -122,8 +122,8 @@ public class IcebergSource implements DataSourceRegister, SupportsCatalogOptions
     setupDefaultSparkCatalogs(spark);
     String path = options.get(SparkReadOptions.PATH);
 
-    Long snapshotId = propertyAsLong(options, SparkReadOptions.SNAPSHOT_ID);
-    Long asOfTimestamp = propertyAsLong(options, SparkReadOptions.AS_OF_TIMESTAMP);
+    Long snapshotId = PropertyUtil.propertyAsLong(options, SparkReadOptions.SNAPSHOT_ID, -1);
+    Long asOfTimestamp = PropertyUtil.propertyAsLong(options, SparkReadOptions.AS_OF_TIMESTAMP, -1);
     Preconditions.checkArgument(
         asOfTimestamp == null || snapshotId == null,
         "Cannot specify both snapshot-id (%s) and as-of-timestamp (%s)",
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/iceberg/util/PropertyUtil.java
```diff
@@ -57,7 +57,7 @@ public static int propertyAsInt(Map<String, String> properties,
   }
 
   public static long propertyAsLong(Map<String, String> properties,
-                                    String property, long defaultValue) {
+                                    String property, Long defaultValue) {
```
szehon-ho, line 60: Not a huge fan of changing signature , even this minor way.

Would a better option be just to change the consuming code to expect a different non-null default value?  (like -1)

zhaomin1423: Thanks for your opinion，have updated.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/IcebergSource.java
```diff
@@ -122,8 +122,8 @@ private Spark3Util.CatalogAndIdentifier catalogAndIdentifier(CaseInsensitiveStri
     setupDefaultSparkCatalogs(spark);
     String path = options.get(SparkReadOptions.PATH);
 
-    Long snapshotId = propertyAsLong(options, SparkReadOptions.SNAPSHOT_ID);
-    Long asOfTimestamp = propertyAsLong(options, SparkReadOptions.AS_OF_TIMESTAMP);
+    Long snapshotId = PropertyUtil.propertyAsLong(options, SparkReadOptions.SNAPSHOT_ID, -1);
+    Long asOfTimestamp = PropertyUtil.propertyAsLong(options, SparkReadOptions.AS_OF_TIMESTAMP, -1);
```
szehon-ho, line 126: In this case you have to change the following code right to expect -1?

zhaomin1423: Thanks for your help, have updated.

szehon-ho (approved):
Sounds good, I feel we should propagate this to other Spark versions to keep the code consistent if possible.

szehon-ho:
@zhaomin1423 actually do you mind doing the same change for other spark versions?  I feel normally we do feature  for specific spark version(s), but for something like this that's purely cosmetic, would prefer to keep the code synchronized so its easier to backport changes if necessary.

szehon-ho:
@zhaomin1423 yea its a good question, it seems it was done this way in the initial split of the Spark modules to different versions.  I think most of the spark code is actually the same across versions :(.  I guess the other option is to keep have just one spark module and have a  shim layer to take care of different things, but not sure that's easy either.

zhaomin1423:
> @zhaomin1423 yea its a good question, it seems it was done this way in the initial split of the Spark modules to different versions. I think most of the spark code is actually the same across versions :(. I guess the other option is to have just one spark module and have a shim layer to take care of different things, but not sure that's easy either.

Keeping a shim layer is difficult with the addition of versions，because a lot of reflect will be added, it is obscure when reading, and some question can't be found in compile stage. Adding a spark common module to put same code, it might be more easier.

PR #5348, opened 2022-07-24
API: add Histogram metric type

stevenzwu:
@rdblue here is the metrics PR. As we discussed, Flink Iceberg sink would need Gauge (e.g. last commit duration) and Histogram (e.g. file size distribution) metrics.

cc @danielcweeks @nastra for additional reviews

### DIFF
--- a/api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
@@ -22,6 +22,7 @@ import java.util.concurrent.TimeUnit;
 
 /** A default {@link MetricsContext} implementation that uses native Java counters/timers. */
 public class DefaultMetricsContext implements MetricsContext {
+  private static final int DEFAULT_HISTOGRAM_RESERVOIR_SIZE = 10_000;
 
   @Override
   @Deprecated
@@ -46,5 +47,10 @@ public class DefaultMetricsContext implements MetricsContext {
   @Override
   public org.apache.iceberg.metrics.Counter counter(String name, Unit unit) {
     return new DefaultCounter(name, unit);
+  }
+
+  @Override
+  public Histogram histogram(String name) {
+    return new FixedReservoirHistogram(DEFAULT_HISTOGRAM_RESERVOIR_SIZE);
   }
 }
--- a/api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
@@ -54,16 +54,12 @@ public class FixedReservoirHistogram implements Histogram {
    */
   @Override
   public Statistics statistics() {
-    int size = measurements.length;
-    if (count < (long) measurements.length) {
-      size = count;
+    int size = (int) Math.min(count, (long) measurements.length);
+    if (size == 0) {
+      return UniformWeightStatistics.EMPTY;
     }
 
     long[] values = new long[size];
-    if (size == 0) {
-      return new UniformWeightStatistics(values, 0.0, 0.0);
-    }
-
     double sum = 0.0d;
     double sumSquares = 0.0d;
     for (int i = 0; i < values.length; ++i) {
@@ -86,6 +82,9 @@ public class FixedReservoirHistogram implements Histogram {
   }
 
   private static class UniformWeightStatistics implements Statistics {
+    private static final UniformWeightStatistics EMPTY =
+        new UniformWeightStatistics(new long[0], 0.0, 0.0);
+
     private final long[] values;
     private final double mean;
     private final double stdDev;
--- a/api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+import java.util.Arrays;
+import java.util.Random;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/** A {@link Histogram} implementation with reservoir sampling. */
+public class FixedReservoirHistogram implements Histogram {
+  private final Random rand;
+  private final long[] measurements;
+  private int count;
+
+  public FixedReservoirHistogram(int reservoirSize) {
+    this.rand = new Random();
+    this.measurements = new long[reservoirSize];
+    this.count = 0;
+  }
+
+  @Override
+  public synchronized int count() {
+    return count;
+  }
+
+  @Override
+  public synchronized void update(long value) {
+    count += 1;
+    int index = count <= measurements.length ? count - 1 : rand.nextInt(count);
+    if (index < measurements.length) {
+      measurements[index] = value;
+    }
+  }
+
+  /**
+   * Naive algorithm for calculating variance:
+   * https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+   */
+  @Override
+  public Statistics statistics() {
+    int size = measurements.length;
+    if (count < (long) measurements.length) {
+      size = count;
+    }
+
+    long[] values = new long[size];
+    if (size == 0) {
+      return new UniformWeightStatistics(values, 0.0, 0.0);
+    }
+
+    double sum = 0.0d;
+    double sumSquares = 0.0d;
+    for (int i = 0; i < values.length; ++i) {
+      synchronized (this) {
+        values[i] = measurements[i];
+      }
+
+      sum += values[i];
+      // Convert to double value to avoid potential overflow of square
+      double value = (double) values[i];
+      sumSquares += value * value;
+    }
+
+    double mean = sum / size;
+    // Use the uncorrected estimator. Bessel's correction is not a good fit in this context.
+    // https://en.wikipedia.org/wiki/Bessel%27s_correction#Caveats
+    double variance = (sumSquares - (sum * sum) / size) / size;
+    double stdDev = Math.sqrt(variance);
+    return new UniformWeightStatistics(values, mean, stdDev);
+  }
+
+  private static class UniformWeightStatistics implements Statistics {
+    private final long[] values;
+    private final double mean;
+    private final double stdDev;
+
+    private UniformWeightStatistics(long[] values, double mean, double stdDev) {
+      this.values = values;
+      this.mean = mean;
+      this.stdDev = stdDev;
+
+      // since the input values is already a copied array,
+      // there is no need to copy again.
+      Arrays.sort(this.values);
+    }
+
+    @Override
+    public int size() {
+      return values.length;
+    }
+
+    @Override
+    public double mean() {
+      return mean;
+    }
+
+    @Override
+    public double stdDev() {
+      return stdDev;
+    }
+
+    @Override
+    public long max() {
+      return values.length == 0 ? 0L : values[values.length - 1];
+    }
+
+    @Override
+    public long min() {
+      return values.length == 0 ? 0L : values[0];
+    }
+
+    @Override
+    public long percentile(double percentile) {
+      Preconditions.checkArgument(
+          !Double.isNaN(percentile) && percentile >= 0.0 && percentile <= 1.0,
+          "Percentile point cannot be outside the range of [0.0 - 1.0]: %s",
+          percentile);
+      if (values.length == 0) {
+        return 0L;
+      } else {
+        double position = percentile * values.length;
+        int index = (int) position;
+        if (index < 1) {
+          return values[0];
+        } else if (index >= values.length) {
+          return values[values.length - 1];
+        } else {
+          double lower = (double) values[index - 1];
+          double upper = (double) values[index];
+          double interpolated = lower + (position - Math.floor(position)) * (upper - lower);
+          return (long) interpolated;
+        }
+      }
+    }
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
@@ -60,13 +60,13 @@ public class FixedReservoirHistogram implements Histogram {
     }
 
     long[] values = new long[size];
+    synchronized (this) {
+      System.arraycopy(measurements, 0, values, 0, size);
+    }
+
     double sum = 0.0d;
     double sumSquares = 0.0d;
     for (int i = 0; i < values.length; ++i) {
-      synchronized (this) {
-        values[i] = measurements[i];
-      }
-
       sum += values[i];
       // Convert to double value to avoid potential overflow of square
       double value = (double) values[i];
--- a/api/src/main/java/org/apache/iceberg/metrics/Histogram.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/Histogram.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+public interface Histogram {
+  /** Update the histogram with a new value observed. */
+  void update(long value);
+
+  /** Return the number of observations. */
+  int count();
+
+  /** Calculate the statistics of the observed values. */
+  Statistics statistics();
+
+  interface Statistics {
+    /**
+     * Return the number of values that the statistics computation is based on.
+     *
+     * <p>If the number of sampling is less than the reservoir size, the sampling count should be
+     * returned. Otherwise, the reservoir size is returned.
+     */
+    int size();
+
+    /** Returns the mean value of the histogram observations. */
+    double mean();
+
+    /** Returns the standard deviation of the histogram distribution. */
+    double stdDev();
+
+    /** Returns the maximum value of the histogram observations. */
+    long max();
+
+    /** Returns the minimum value of the histogram observations. */
+    long min();
+
+    /**
+     * Returns the percentile value based on the histogram statistics.
+     *
+     * @param percentile percentile point in double. E.g., 0.75 means 75 percentile. It is up to the
+     *     implementation to decide what valid percentile points are supported.
+     * @return Value for the given percentile
+     */
+    long percentile(double percentile);
+  }
+}
--- a/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
+++ b/api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
@@ -140,6 +140,10 @@ public interface MetricsContext extends Serializable {
     throw new UnsupportedOperationException("Timer is not supported.");
   }
 
+  default Histogram histogram(String name) {
+    throw new UnsupportedOperationException("Histogram is not supported.");
+  }
+
   /**
    * Utility method for producing no metrics.
    *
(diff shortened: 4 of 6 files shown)


api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
```diff
@@ -42,4 +42,21 @@ public <T extends Number> Counter<T> counter(String name, Class<T> type, Unit un
...
+  @Override
+  public <T extends Number> Gauge<T> gauge(String name, Class<T> type) {
+    if (Integer.class.equals(type)) {
+      return (Gauge<T>) new IntGauge();
+    }
+
+    if (Long.class.equals(type)) {
+      return (Gauge<T>) new LongGauge();
+    }
+    throw new IllegalArgumentException(String.format("Counter for type %s is not supported", type.getName()));
```
nastra, line 55: nit: should this say `Gauge` instead of `Counter`?

stevenzwu: good catch. fixed.

api/src/test/java/org/apache/iceberg/metrics/TestDefaultMetricsContext.java
```diff
@@ -83,4 +85,44 @@ public void timer() {
...
+    gauge.set(1L);
+    Assertions.assertThat(gauge.get()).isEqualTo(1L);
+  }
+
+  @Test
+  public void histogram() {
+    MetricsContext metricsContext = new DefaultMetricsContext();
+    int reservoirSize = 128;
+    Histogram histogram = metricsContext.histogram("test", reservoirSize);
+    for (int i = 0; i < reservoirSize; ++i) {
```
nastra, line 110: nit: I don't think it matters whether we do ++i or i++ here :)

nastra: unless you intended to add `1..128`, then you'd probably want to rewrite the loop to `for (int i = 1; i <= reservoirSize; i++)`. That would also make it sligthly easier to reason about the different percentiles below. For example, the 50th percentile would then be `Assertions.assertThat(statistics.percentile(0.50)).isEqualTo(64)` instead of `63`

api/src/main/java/org/apache/iceberg/metrics/DefaultHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+  public synchronized void update(long value) {
+    int index = (int) (count % measurements.length);
+    measurements[index] = value;
+    count++;
+  }
+
+  @Override
+  public Statistics statistics() {
+    int size = measurements.length;
+    if (count < (long) measurements.length) {
```
nastra, line 52: what happens if a different thread calls `update()`? should we add a multi-threaded test similar to `TestDefaultTimer#multiThreadedStarts()` maybe?

nastra (commented):
APIs LGTM but I added a few comments around `DefaultHistogram` and testing

api/src/main/java/org/apache/iceberg/metrics/DefaultHistogram.java
stevenzwu: `update` is a `synchronized` method. 

the array clone is protected by `synchronized` too when reading array element
```
    long[] values = new long[size];
    for (int i = 0; i < values.length; ++i) {
      synchronized (this) {
        values[i] = measurements[i];
      }
    }
```

stevenzwu: Regarding multiple-thread testing, I will add similar test method.

api/src/test/java/org/apache/iceberg/metrics/TestDefaultMetricsContext.java
stevenzwu: sure, I am fine either way. will switch to `1..128`

stevenzwu: actually switched to `[1, 100]` which will make the assertion number easier to read.

api/src/main/java/org/apache/iceberg/metrics/DefaultHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public long min() {
+      return values.length == 0 ? 0L : values[0];
+    }
+
+    @Override
+    public double percentile(double percentile) {
+      Preconditions.checkArgument(
+          !Double.isNaN(percentile) && percentile >= 0.0 && percentile <= 1.0,
+          "Percentile point cannot be outside the range of [0.0 - 1.10]: %s",
```
kbendick, line 126: Nit: `range of [0.0 - 1.0]`, not `1.10`. Either the string has a typo or the boolean logic  should be `percentile <= 1.10` but I assume it's a string typo.

api/src/main/java/org/apache/iceberg/metrics/DefaultHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+          percentile);
+      if (values.length == 0) {
+        return 0.0;
+      } else {
+        double position = percentile * values.length;
+        int index = (int) position;
+        if (index < 1) {
+          return values[0];
+        } else if (index >= values.length) {
+          return values[values.length - 1];
```
kbendick, line 136: Are these two blocks needed? Particularly the `index < 1` block.

I'm ok with it as defensive coding, but wondering how practical it is or what made you think to add these blocks (e.g. can these situations occur?).

kbendick, line 136: Are these two blocks needed? Particularly the `index < 1` block.

I'm ok with it as defensive coding, but wondering how practical it is or what made you think to add these blocks (e.g. can these situations occur?).

api/src/main/java/org/apache/iceberg/metrics/MetricsContext.java
```diff
@@ -106,6 +106,14 @@ default Timer timer(String name, TimeUnit unit) {
     throw new UnsupportedOperationException("Timer is not supported.");
   }
 
+  default <T extends Number> Gauge<T> gauge(String name, Class<T> type) {
+    throw new UnsupportedOperationException("Gauge is not supported.");
+  }
+
+  default Histogram histogram(String name, int reservoirSize) {
```
rdblue, line 113: Should we have these?

```
public <T extends Counter & Histogram> histCounter(String name, int reservoirSize) {...}
public <T extends Timer & Histogram> histTimer(String name, int reservoirSize) {...}
```

stevenzwu:
Had a discussion with @rdblue. 

Iceberg `MetricContext` should be used as a way for iceberg-core to expose metrics. Engine specific metrics (like Flink reader or writer) don't need to add an indirection and translate from Iceberg `MetricContext` to Flink metrics. There is not much benefit for the indirection. 

For that reason, I will remove the `Gauge` metric from this PR. I will also send another PR for Flink FLIP-27 source, which currently translate from Iceberg `MetricsContext` to Flink metrics.

For simplicity, we will remove the `reservoirSize` param from `MetricsContext#histogram(name)` and hard-code the reservoirSize (probably to 1,024) for now. In the future, when we want to provide more flexibility, maybe we can add a new interface (like `ObservationsTracker`) that allows users to specify the data structure for tracking observations (e.g. FixedReservoirTracker, SketchTracker, etc.)

api/src/main/java/org/apache/iceberg/metrics/DefaultHistogram.java
stevenzwu: good catch. it is a string typo

api/src/main/java/org/apache/iceberg/metrics/DefaultHistogram.java
stevenzwu: @kbendick yes, there are necessary. For the edge scenarios (index = 0 or length-1), we don't need to apply the proration calculation in the else block.

```
          double lower = (double) values[index - 1];
          double upper = (double) values[index];
          return lower + (position - Math.floor(position)) * (upper - lower);
```

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+import java.util.Arrays;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+
+/**
+ * A {@link Histogram} implementation that uses a fixed size reservoir like circular buffer. Oldest
+ * observation is replaced by latest observation.
```
rdblue, line 26: Okay, after reading this comment I understand what you were talking about more clearly. I was wondering why the sampling would be biased toward the last collected values, but I see that it's because those are the values that you're collecting.

What I would expect from a reservoir algorithm is to choose the final set of values using reservoir sample so that we know that $k$ values from the entire input set, size $n$, are selected with a uniform probability, $k/n$, using the [simple sampling algorithm](https://en.wikipedia.org/wiki/Reservoir_sampling).

I think that we probably want to use the uniform sampling by default. That way it isn't biased toward the order of values. That should also make the choice of reservoir size easier, since it needs to be proportional to accuracy and the target number of ntiles, rather than proportional to the size of the input data.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+    this.count = 0L;
+  }
+
+  @Override
+  public synchronized long count() {
+    return count;
+  }
+
+  @Override
+  public synchronized void update(long value) {
```
rdblue, line 43: I think this should be:

```java
  public synchronized void update(long value) {
    count += 1;
    index = count < measurements.length ? count : rand.nextInt(count);
    if (index < measurements.length) {
      measurements[index] = value;
    }
  }
```

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+
+    return new UniformWeightStatistics(values);
+  }
+
+  private static class UniformWeightStatistics implements Statistics {
+    private final long[] values;
+
+    private UniformWeightStatistics(long[] values) {
```
rdblue, line 69: Rather than copying all of the values to a new array and lazily computing stats, wouldn't it make more sense to compute stats over the sampled values by iterating here once?

Right now, if you want the stddev and mean, this would:
1. Allocate a new array of length `L = min(measurements.length, count)`
2. Copy `L` measurements into the new array
3. Iterate over `L` measurements to produce a mean (to return)
4. Iterate over `L` measurements to produce a mean (for stddev calculation)
5. Iterate over `L` measurements to produce `sumOfDiffSquares`

I think it would be a lot better to calculate all of the stats in one pass over the values in the constructor:

```java
  private UniformWeightStatistics(long[] measurements, int count) {
    double sumSquares = 0.0d;
    double sum = 0.0d;
    this.min = Long.MAX_VALUE;
    this.max = Long.MIN_VALUE;
    for (int i = 0; i < count; i += 1) {
      sumSquares += measurements[i] * measurements[i];
      sum += measurements[i];
      if (measurements[i] < min) {
        this.min = measurements[i];
      }
      if (measurements[i] > max) {
        this.max = measurements[i];
      }
    }
    this.mean = sum / count;
    this.stddev = Math.sqrt((sumSquares / count) - (mean * mean));
  }
```

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+          "Percentile point cannot be outside the range of [0.0 - 1.0]: %s",
+          percentile);
+      if (values.length == 0) {
+        return 0.0;
+      } else {
+        double position = percentile * values.length;
+        int index = (int) position;
+        if (index < 1) {
+          return values[0];
+        } else if (index >= values.length) {
```
rdblue, line 135: I think this is correct.

The only time that we would expect `index` to be `equal to `values.length` is when `percentile` is `1.0`, but I think it's reasonable to be extra safe here and check the actual index value produced to avoid an `ArrayIndexOutOfBoundsException`.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+          return values[values.length - 1];
+        } else {
+          double lower = (double) values[index - 1];
+          double upper = (double) values[index];
+          return lower + (position - Math.floor(position)) * (upper - lower);
```
rdblue, line 140: It would be good to note in Javadoc that the percentile will be interpolated when it is between positions

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,145 @@
...
+      } else {
+        double position = percentile * values.length;
+        int index = (int) position;
+        if (index < 1) {
+          return values[0];
+        } else if (index >= values.length) {
+          return values[values.length - 1];
+        } else {
+          double lower = (double) values[index - 1];
+          double upper = (double) values[index];
```
rdblue, line 139: After double-checking this with a few values, I think it looks good.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
rdblue: Ah, I missed that you're doing the sort on the values in the copy. It's probably fine to copy still, but you can calculate the mean and stddev while you copy, before sorting the values.

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
```diff
@@ -0,0 +1,61 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iceberg.metrics;
+
+public interface Histogram {
+  /** Update the histogram with a new value observed. */
+  void update(long value);
```
rdblue, line 23: I don't see a reason why this needs to be a `long`. It could easily be any primitive number, right?

api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
```diff
@@ -22,6 +22,7 @@
 
 /** A default {@link MetricsContext} implementation that uses native Java counters/timers. */
 public class DefaultMetricsContext implements MetricsContext {
+  private static final int DEFAULT_HISTOGRAM_RESERVOIR_SIZE = 1024;
```
rdblue, line 25: How about using 10k, rather than 1k? If we use reservoir sampling, then that would give us about 100 data points for percentiles, which seems like a good enough default.

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
```diff
@@ -0,0 +1,61 @@
...
+    long min();
+
+    /**
+     * Returns the percentile value based on the histogram statistics.
+     *
+     * @param percentile percentile point in double. E.g., 0.75 means 75 percentile. It is up to the
+     *     implementation to decide what valid percentile points are supported.
+     * @return Value for the given percentile
+     */
+    double percentile(double percentile);
```
rdblue, line 59: Should this produce a long if the input values were longs?

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
stevenzwu: I see where the difference comes from. I was mostly thinking about the long-running streaming job, where metrics are periodically polled and reported to metrics backend. Hence we only want to report the measurements based on latest values.

You were talking about the batch jobs where uniform sampling is more preferable without bias toward the recent values.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
stevenzwu: yeah. avoiding multiple loops makes sense.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
stevenzwu: as mentioned in another comment, we may need to provide two ways of tracking observations
* circular buffer mode with bias toward recent data, which is good for streaming jobs
* uniform sampling with like the code snippet above, which probably works better for batch jobs

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
stevenzwu: agree. will add

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
stevenzwu: I think you are right. we can make it a generic type as `Histogram<T extends Number>`. then we just need to change `min` and `max` return type to `T`. everything else (like `mean` and `percentile`) can stay with `double` return type.

api/src/main/java/org/apache/iceberg/metrics/DefaultMetricsContext.java
stevenzwu: I am fine with 10K as default value. 10k size long[] array would cost 80 KB per metric. 100 histogram metrics would be 8 MB, which isn't too bad.

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
stevenzwu: because we do `interpolation` when position (`double`) is btw indexes, the return value needs to be `double`.

```
return lower + (position - Math.floor(position)) * (upper - lower);
```

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
stevenzwu: while trying the generic type, I realized there are a couple of small performance implications
1) we will need to define impl classes like `IntHistogram`, `LongHistogram` and `DoubleHistogram`, as we can avoid create `T[]` using the wrapper/box class. the primitive type array is 50% smaller.
2) there will be unboxing overhead.

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
rdblue: I know the interpolation result will be a double, but we could still cast or round to a long value before returning. I think that would make sense. Is there a case where you wouldn't want that?

api/src/main/java/org/apache/iceberg/metrics/Histogram.java
rdblue: Let's just leave it as long for now.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
rdblue: Are we going to have cases where we want to use these metrics in long-running streaming jobs? Wouldn't we use Flink's metrics directly for that?

What is a case where we want to use `MetricsContext` and have Iceberg report metrics where we'd want to bias toward the latest values?

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
stevenzwu: Looking at the standard deviation again, I don't think `Math.sqrt((sumSquares / count) - (mean * mean))` is correct. 

We need to calculate the mean first in order to compute `sumOfDiffSquares`. But we can avoid computing the mean twice.

rdblue: The `sumSquares` approach is correct and equivalent. It is the ["Naive algorithm" described on Wikipedia](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance). We should probably also apply Bessel's correction as well since this is a sample (as noted on that page).

There's an [alternative algorithm](https://math.stackexchange.com/questions/198336/how-to-calculate-standard-deviation-with-streaming-inputs) mentioned on stack exchange, too. I've always used the naive one, since it's easy.

stevenzwu: Thanks for the wikipedia link. I updated the impl.

I didn't apply Bessel's correction. According to the [caveats section](https://en.wikipedia.org/wiki/Bessel%27s_correction#Caveats), it is probably not the right fit here. Also the Bessel's correction can't work with `n == 1`.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /**
+   * Naive algorithm for calculating variance:
+   * https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+   */
+  @Override
+  public Statistics statistics() {
+    int size = measurements.length;
+    if (count < (long) measurements.length) {
```
rdblue, line 58: What about `int size = (int) Math.min(count, measturements.length)`

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,152 @@
...
+      size = count;
+    }
+
+    long[] values = new long[size];
+    if (size == 0) {
+      return new UniformWeightStatistics(values, 0.0, 0.0);
```
rdblue, line 64: Minor: This could be a constant.

stevenzwu: thx. updated along with the comment for size.

also rebased to resolve merge conflict.

api/src/main/java/org/apache/iceberg/metrics/FixedReservoirHistogram.java
```diff
@@ -0,0 +1,151 @@
...
+   * https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+   */
+  @Override
+  public Statistics statistics() {
+    int size = (int) Math.min(count, (long) measurements.length);
+    if (size == 0) {
+      return UniformWeightStatistics.EMPTY;
+    }
+
+    long[] values = new long[size];
```
rdblue, line 62: I think this should probably use just one `synchronized` here with `System.arraycopy` rather than locking for each value in the loop below.

rdblue (approved):

rdblue:
Thanks, @stevenzwu! Looks great.

PR #5350, opened 2022-07-24
Add a close tag </p> to LoadTableResponse and ConfigResponse JavaDoc

price-qian:


rdblue:
Javadoc doesn't need end tags for paragraphs. We only use `<p>` and the end tags are automatically generated.

PR #5353, opened 2022-07-26
Core: Remove TestEnvironmentUtil#testEnvironmentSubstitution() as it is bui…

stevenzwu:
…ld env dependent. It will fail with `ImmutableMap.of("user-test", System.getenv().get("USER"))`, if the build env doesn't contain USER env variable, which is the case for our internal build env.

```
 java.lang.NullPointerException: null value in entry: user-test=null
         at org.apache.iceberg.relocated.com.google.common.collect.CollectPreconditions.checkEntryNotNull(CollectPreconditions.java:33)
         at org.apache.iceberg.relocated.com.google.common.collect.SingletonImmutableBiMap.<init>(SingletonImmutableBiMap.java:43)
         at org.apache.iceberg.relocated.com.google.common.collect.ImmutableBiMap.of(ImmutableBiMap.java:81)
         at org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap.of(ImmutableMap.java:126)
         at org.apache.iceberg.util.TestEnvironmentUtil.testEnvironmentSubstitution(TestEnvironmentUtil.java:31)
```

Although I can supply the env variables for our internal build env, I found that this test only covers a trivial method. Hence I think it is probably better just remove the unit test for this trivial method.
```
  public static Map<String, String> resolveAll(Map<String, String> properties) {
    return resolveAll(System.getenv(), properties);
  }
```

the other non-public `resolveAll(env, properties)` is already covered by `TestEnvironmentUtil#testMultipleEnvironmentSubstitutions()`.

### DIFF
--- a/core/src/test/java/org/apache/iceberg/util/TestEnvironmentUtil.java
+++ b/core/src/test/java/org/apache/iceberg/util/TestEnvironmentUtil.java
@@ -26,10 +26,21 @@ import org.junit.jupiter.api.Test;
 class TestEnvironmentUtil {
   @Test
   public void testEnvironmentSubstitution() {
-    Assertions.assertEquals(
-        ImmutableMap.of("user-test", System.getenv().get("USER")),
-        EnvironmentUtil.resolveAll(ImmutableMap.of("user-test", "env:USER")),
-        "Should get the user from the environment");
+    String userFromEnv = System.getenv().get("USER");
+    Map<String, String> resolvedProps =
+        EnvironmentUtil.resolveAll(ImmutableMap.of("user-test", "env:USER"));
+    if (userFromEnv == null) {
+      // some build env may not have the USER env variable set
+      Assertions.assertEquals(
+          ImmutableMap.of(),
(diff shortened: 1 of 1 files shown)


rdblue:
@stevenzwu, is it possible to fix this test instead? Can we ensure that a value is present for an environment variable? I think it is good to ensure that the properties are actually loaded from the environment.

Alternatively, maybe we could use something other than an `ImmutableMap` so that a `null` value is okay.

stevenzwu:
@rdblue I thought about sth similar to your proposal earlier. We can make the test to assert on different expected output depends on the whether the `USER` env variable exists or not. But then I looked the public `resolveAll` method and found it is so trivial and simply calls the other non-public `resolveAll` method. And the non-public `resolveAll` is already covered by the unit test. Hence, I think it is not necessary to keep this test for a trivial method.

rdblue (approved):

rdblue:
Thanks, @stevenzwu!

PR #5357, opened 2022-07-26
Flink: change precondition check to allow table pass in

chenjunjiedada:
This change precondition to allow passing in the table instance. 

Sometimes the table is already loaded and accessed before building the sink, this change can avoid load the table again.


stevenzwu (changes requested):
I understand the need to pass in the `table` directly to avoid double loading on the jobmanager side. It is already supported in the current API.

`tableLoader` should always be provided, as it needs to be serializable to be used by the committer operator. `SerializableTable` won't work for committer, as it is a ready-only copy of the `Table` and can't be used to change table state.

I think we should not make the change in this PR.

chenjunjiedada:
Thanks @stevenzwu for the explanation.

PR #5360, opened 2022-07-26
[Python] Remove all base.py, use __init__.py where it works

samredai:
This is a PR to replace all of the `base.py` by either moving the contents into `__init__.py` or renaming them to something more meaningful.

As an example, the contents of `pyiceberg/io/base.py` has been moved to `pyiceberg/io/__init__.py`:
```py
from pyiceberg import io
dir(io)
# ['ABC', 'FileIO', 'InputFile', 'InputStream', 'OutputFile', 'OutputStream', 'Protocol', 'SEEK_SET', 'Union', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'abstractmethod', 'runtime_checkable']
```

To avoid circular imports in the expression sub-module, `base.py` was renamed `expression.py` and the public API is exposed via `__all__ = [...]` defined in the `__init__.py` for the expression module.


python/tests/catalog/test_catalog.py
```diff
@@ -20,43 +20,43 @@
...
 from pyiceberg.table.metadata import INITIAL_SPEC_ID
 from pyiceberg.table.partitioning import PartitionSpec
 
 
 class InMemoryCatalog(Catalog):
     """An in-memory catalog implementation for testing purposes."""
 
-    __tables: Dict[Identifier, Table]
-    __namespaces: Dict[Identifier, Properties]
+    __tables: Dict[Tuple[str, ...], Table]
```
Fokko, line 45: Was this change intentional?

samredai: Unless I messed something up here, this was causing a circular import:
```
ImportError while loading conftest '/Users/sam/iceberg/python/tests/conftest.py'.
tests/conftest.py:55: in <module>
    from tests.catalog.test_catalog import InMemoryCatalog
tests/catalog/test_catalog.py:28: in <module>
    from pyiceberg.catalog import Catalog, Identifier, Properties
pyiceberg/catalog/__init__.py:24: in <module>
    from pyiceberg.tableimport Table
pyiceberg/table/__init__.py:22: in <module>
    from pyiceberg.catalog.base import Identifier
E   ImportError: cannot import name 'Identifier' from partially initialized module 'pyiceberg.catalog' (most likely due to a circular import) (/Users/sam/iceberg/python/pyiceberg/catalog/__init__.py)
```
Essentially from [this](https://github.com/apache/iceberg/blob/master/python/pyiceberg/table/base.py#L22) line. If there's a way to avoid this and still use `__init__.py` I can update this. Although I do wonder how informative a type of `Identifier` is versus `Tuple[str, ...]`.

python/pyiceberg/expression/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    FloatLiteral,
+    LongLiteral,
+    StringLiteral,
+    TimeLiteral,
+    TimestampLiteral,
+    UUIDLiteral,
+    literal,
+)
+
+__all__ = [
```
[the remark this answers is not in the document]
samredai: I thought about this but the expression related stuff does have enough of a logical grouping with things like literals, evaluators, all of the visitors, etc. so it seems valuable to have all of this contained in a submodule for context. I'm ok with either way here though so if nobody else chimes in I'll update this. 👍

python/pyiceberg/catalog/__init__.py
```diff
@@ -1,20 +1,246 @@
-# Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
-# distributed with this work for additional information
-# regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
-# "License"); you may not use this file except in compliance
-# with the License.  You may obtain a copy of the License at
+#  Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
samredai: Fixed!

python/pyiceberg/catalog/__init__.py
```diff
@@ -14,7 +14,233 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from typing import Dict, Tuple
 
-Identifier = Tuple[str, ...]
-Properties = Dict[str, str]
```
rdblue, line 20: Why did these get removed? I think they are good shortcuts to keep.

python/pyiceberg/catalog/__init__.py
```diff
@@ -14,7 +14,233 @@
...
+
+    The catalog namespace APIs follow a similar convention wherein they also accept a namespace identifier that can be a string
+    or tuple of strings.
+
+    Attributes:
+        name(str): Name of the catalog
+        properties(Dict[str, str]): Catalog properties
+    """
+
+    def __init__(self, name: str, properties: dict[str, str]):
```
rdblue, line 41: I don't think that the contents should change when moved. Can you make sure that this is as close as possible to the original?

python/pyiceberg/expression/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    Not,
+    NotEqual,
+    NotIn,
+    NotNaN,
+    NotNull,
+    Or,
+)
+from pyiceberg.expression.literals import (
+    AboveMax,
+    BelowMin,
```
rdblue, line 36: `AboveMax` and `BelowMin` don't need to be pulled in. Those are used in the binding process, but aren't needed afterward.

python/pyiceberg/expression/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    DateLiteral,
+    DecimalLiteral,
+    DoubleLiteral,
+    FixedLiteral,
+    FloatLiteral,
+    LongLiteral,
+    StringLiteral,
+    TimeLiteral,
+    TimestampLiteral,
+    UUIDLiteral,
```
rdblue, line 48: Do we need the literal classes? I think the `literal` method should be enough. And we should also coerce values in predicates.

python/pyiceberg/expression/expressions.py
```diff
@@ -489,88 +449,88 @@ def bind(self, schema: Schema, case_sensitive: bool) -> BoundNotIn[T]:
         return BoundNotIn(bound_ref, *tuple(lit.to(bound_ref.field.field_type) for lit in self.literals))  # type: ignore
 
 
-class BoundEq(BoundPredicate[T]):
-    def __invert__(self) -> BoundNotEq[T]:
-        return BoundNotEq(self.term, *self.literals)
+class BoundEqual(BoundPredicate[T]):
```
rdblue, line 452: Can these renames happen in a separate PR?

python/pyiceberg/expression/__init__.py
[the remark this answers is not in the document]
rdblue: I don't think literals really need to be exposed. Does that affect this decision?

python/pyiceberg/io/__init__.py
```diff
@@ -14,3 +14,187 @@
...
+
+The FileIO abstraction includes a subset of full filesystem implementations. Specifically,
+Iceberg needs to read or write a file at a given location (as a seekable stream), as well
+as check if a file exists. An implementation of the FileIO abstract base class is responsible
+for returning an InputFile instance, an OutputFile instance, and deleting a file given
+its location.
+"""
+from abc import ABC, abstractmethod
+from io import SEEK_SET
+from typing import Protocol, Union, runtime_checkable
```
rdblue, line 27: Are there additional changes here that prevented git from recognizing that it was moved from `base.py`?

python/pyiceberg/table/__init__.py
```diff
@@ -1,16 +1,29 @@
...
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
 #
-#   http://www.apache.org/licenses/LICENSE-2.0
+#    http://www.apache.org/licenses/LICENSE-2.0
```
rdblue, line 9: Unnecessary changes here as well.

python/pyiceberg/table/__init__.py
```diff
@@ -1,16 +1,29 @@
...
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from __future__ import annotations
+
+from abc import ABC
+
+
+class Table(ABC):
```
rdblue, line 23: Since this is so small, should we move this all to `table.py` instead of a package directory?

python/tests/catalog/test_catalog.py
rdblue: Maybe we should have a `typedef.py` file somewhere to handle these? I like the definitions.

python/pyiceberg/expression/__init__.py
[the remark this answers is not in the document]
samredai: > I don't think literals really need to be exposed. Does that affect this decision?

If we replace the `expression` module with a single `expressions.py` file, should the contents of `literals.py` just be moved to the top of that file then? It would still be exposed if someone wanted to import it from `pyiceberg.expressions`.

python/pyiceberg/table/__init__.py
samredai: There are still [other](https://github.com/apache/iceberg/tree/master/python/pyiceberg/table) files in the package directory such as `snapshots.py`, `partitioning.py`, etc.

python/pyiceberg/io/__init__.py
samredai: I'm not sure it's not being recognized but this was a simple `mv pyiceberg/io/base.py pyiceberg/io/__init__.py`

samredai:
@rdblue I removed all unnecessary changes
@Fokko in a separate [commit](https://github.com/apache/iceberg/pull/5360/commits/5374120c1cb94fc661aaf437b5c0e845a819df8c) I pulled flattened the expressions package directory into a top level `expressions.py` and `literals.py`. Let me know what you think.

python/pyiceberg/io/__init__.py
Fokko: You could try `git mv pyiceberg/io/base.py pyiceberg/io/__init__.py`

Fokko (approved):
Thanks @samredai I like this a lot! 🚀

rdblue:
@samredai, I think the next steps on this are to open PRs for the io and catalog modules. We need to think a bit more about what we want to expose from the expressions module. I think we may want something like the `Expressions` class in Java that hides creating references and literals.

samredai:
@rdblue I opened #5456 and #5457 which breaks out the `io` and `catalog` portion of this PR.

PR #5361, opened 2022-07-26
Build: Do not let Iceberg build fail with `-DscalaVersion=2.13`

snazy:
Flink does not provide artifacts for Scala 2.13. It _seems_ that
there is no Scala at play here, so it feels legit to silently
"downgrade" to 2.12 for the build.

I.e. when running `./gradlew jar -DscalaVersion=2.13` the Iceberg build fails.

@rdblue I am absolutely unsure whether my assumption that Scala isn't really used here is correct, but it seems that the use of Scala degraded for more recent Flink versions.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -52,16 +52,27 @@ project(':nessie').name = 'iceberg-nessie'
 project(':gcp').name = 'iceberg-gcp'
 project(':dell').name = 'iceberg-dell'
 
+List<String> knownScalaVersions = System.getProperty("knownScalaVersions").split(",")
+String scalaVersion = System.getProperty("scalaVersion", System.getProperty("defaultScalaVersion"))
+if (!knownScalaVersions.contains(scalaVersion)) {
+  throw new GradleException("Found unsupported Scala version: " + scalaVersion)
+}
+
 List<String> knownFlinkVersions = System.getProperty("knownFlinkVersions").split(",")
-String flinkVersionsString = System.getProperty("flinkVersions") != null ? System.getProperty("flinkVersions") : System.getProperty("defaultFlinkVersions")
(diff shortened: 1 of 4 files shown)


rdblue:
What about handling this like we do for unsupported JVM versions? If you're running with Java 11, we'll automatically remove Spark 2.4. Maybe we should do that here as well instead of downgrading?

snazy:
Yup - makes more sense w/ the current build.

BTW: In Nessie we let "pure" Gradle runs build for and test against all Scala versions. It would be a bit more of a change in the settings.gradle + build.gradle files, but it works nicely. IDE imports however only get one Scala version (because otherwise IDEs get confused w/ multiple projects referencing the same source roots).

rdblue:
@snazy, when we added 2.13 support, I didn't think it was possible to have Scala projects that use multiple versions. If you can get it working in Iceberg, that would be a great addition.

settings.gradle
```diff
@@ -52,37 +52,41 @@ project(':nessie').name = 'iceberg-nessie'
 project(':gcp').name = 'iceberg-gcp'
 project(':dell').name = 'iceberg-dell'
 
+List<String> knownScalaVersions = System.getProperty("knownScalaVersions").split(",")
+String scalaVersion = System.getProperty("scalaVersion", System.getProperty("defaultScalaVersion"))
+if (!knownScalaVersions.contains(scalaVersion)) {
+  throw new GradleException("Found unsupported Scala version: " + scalaVersion)
```
rdblue, line 58: Rather than throwing an exception, I think this should just exclude the projects from the build. Here's an example: https://github.com/apache/iceberg/blob/master/settings.gradle#L175-L193

PR #5362, opened 2022-07-26
Python: Refactor unary and set expressions

rdblue:
This refactors unary and set expressions to use base classes. In addition, this updates binding to produce `AlwaysTrue()` or `AlwaysFalse()` values based on the bound type. For example, `IsNull(a)` when bound to a required field a will produce `AlwaysFalse()` because the value is required.

This also adds missing tests.

### DIFF
--- a/python/pyiceberg/expressions/base.py
+++ b/python/pyiceberg/expressions/base.py
@@ -138,72 +138,6 @@ class Reference(UnboundTerm[T], BaseReference[T]):
         return BoundReference(field=field, accessor=accessor)
 
 
-class BoundPredicate(Bound[T], BooleanExpression):
-    _term: BoundTerm[T]
-    _literals: tuple[Literal[T], ...]
-
-    def __init__(self, term: BoundTerm[T], *literals: Literal[T]):
-        self._term = term
-        self._literals = literals
-        self._validate_literals()
-
-    def _validate_literals(self):
-        if len(self.literals) != 1:
-            raise AttributeError(f"{self.__class__.__name__} must have exactly 1 literal.")
-
-    @property
-    def term(self) -> BoundTerm[T]:
-        return self._term
-
-    @property
-    def literals(self) -> tuple[Literal[T], ...]:
-        return self._literals
-
-    def __str__(self) -> str:
-        return f"{self.__class__.__name__}({str(self.term)}{self.literals and ', '+str(self.literals)})"
-
-    def __repr__(self) -> str:
-        return f"{self.__class__.__name__}({repr(self.term)}{self.literals and ', '+repr(self.literals)})"
-
-    def __eq__(self, other) -> bool:
-        return id(self) == id(other) or (
-            type(self) == type(other) and self.term == other.term and self.literals == other.literals
-        )
-
-
-class UnboundPredicate(Unbound[T, BooleanExpression], BooleanExpression, ABC):
-    _term: UnboundTerm[T]
-    _literals: tuple[Literal[T], ...]
-
-    def __init__(self, term: UnboundTerm[T], *literals: Literal[T]):
-        self._term = term
-        self._literals = literals
-        self._validate_literals()
-
-    def _validate_literals(self):
-        if len(self.literals) != 1:
-            raise AttributeError(f"{self.__class__.__name__} must have exactly 1 literal.")
-
-    @property
-    def term(self) -> UnboundTerm[T]:
-        return self._term
-
-    @property
-    def literals(self) -> tuple[Literal[T], ...]:
-        return self._literals
-
-    def __str__(self) -> str:
-        return f"{self.__class__.__name__}({str(self.term)}{self.literals and ', '+str(self.literals)})"
-
-    def __repr__(self) -> str:
-        return f"{self.__class__.__name__}({repr(self.term)}{self.literals and ', '+repr(self.literals)})"
-
-    def __eq__(self, other) -> bool:
-        return id(self) == id(other) or (
-            type(self) == type(other) and self.term == other.term and self.literals == other.literals
-        )
-
-
 class And(BooleanExpression):
     """AND operation expression - logical conjunction"""
 
@@ -325,23 +259,37 @@ class AlwaysFalse(BooleanExpression, Singleton):
 
 
 @dataclass(frozen=True)
-class UnaryPredicate(Unbound[T, BooleanExpression], BooleanExpression):
+class BoundPredicate(Bound[T], BooleanExpression):
+    term: BoundTerm[T]
+
+    def __invert__(self) -> BoundPredicate[T]:
+        raise NotImplementedError
+
+
+@dataclass(frozen=True)
+class UnboundPredicate(Unbound[T, BooleanExpression], BooleanExpression):
     as_bound: ClassVar[type]
     term: UnboundTerm[T]
 
-    def __invert__(self) -> UnaryPredicate[T]:
+    def __invert__(self) -> UnboundPredicate[T]:
         raise NotImplementedError
 
+    def bind(self, schema: Schema, case_sensitive: bool = True) -> BooleanExpression:
+        raise NotImplementedError
+
+
+@dataclass(frozen=True)
+class UnaryPredicate(UnboundPredicate[T]):
     def bind(self, schema: Schema, case_sensitive: bool = True) -> BooleanExpression:
         bound_term = self.term.bind(schema, case_sensitive)
         return self.as_bound(bound_term)
 
+    def __invert__(self) -> UnaryPredicate[T]:
+        raise NotImplementedError
+
 
 @dataclass(frozen=True)
-class BoundUnaryPredicate(Bound[T], BooleanExpression):
-    inverse: ClassVar[type]
-    term: BoundTerm[T]
-
+class BoundUnaryPredicate(BoundPredicate[T]):
     def __invert__(self) -> BoundUnaryPredicate[T]:
         raise NotImplementedError
 
@@ -425,9 +373,7 @@ class NotNaN(UnaryPredicate[T]):
 
 
 @dataclass(frozen=True)
-class SetPredicate(Unbound[T, BooleanExpression], BooleanExpression):
-    as_bound: ClassVar[type]
-    term: UnboundTerm[T]
+class SetPredicate(UnboundPredicate[T]):
     literals: tuple[Literal[T], ...]
 
     def __invert__(self) -> SetPredicate[T]:
@@ -439,8 +385,7 @@ class SetPredicate(Unbound[T, BooleanExpression], BooleanExpression):
 
 
 @dataclass(frozen=True)
-class BoundSetPredicate(Bound[T], BooleanExpression):
-    term: BoundTerm[T]
+class BoundSetPredicate(BoundPredicate[T]):
     literals: set[Literal[T]]
 
     def __invert__(self) -> BoundSetPredicate[T]:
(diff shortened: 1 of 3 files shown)


python/pyiceberg/expressions/base.py
```diff
@@ -204,6 +210,7 @@ def __eq__(self, other) -> bool:
         )
 
 
+# TODO: this should still exist as a parent, but term should be abstract
```
rdblue, line 213: The `UnaryPredicate` and `BoundUnaryPredicate` classes introduced in this PR should implement the `UnboundPredicate` and `BoundPredicate` interfaces, but I didn't want to refactor the entire set of expressions in this PR so I added this TODO as a reminder.

python/pyiceberg/expressions/base.py
```diff
@@ -342,107 +349,104 @@ def __str__(self) -> str:
 
 
 @dataclass(frozen=True)
-class AlwaysTrue(BooleanExpression, ABC, Singleton):
+class AlwaysTrue(BooleanExpression, Singleton):
```
rdblue, line 352: These aren't actually abstract.

python/pyiceberg/expressions/base.py
```diff
@@ -342,107 +349,104 @@ def __str__(self) -> str:
...
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> BoundIsNaN[T]:
-        bound_ref = self.term.bind(schema, case_sensitive)
-        return BoundIsNaN(bound_ref)
+IsNull.inverse = NotNull
+IsNull.as_bound = BoundIsNull
+NotNull.inverse = IsNull
+NotNull.as_bound = BoundNotNull
+BoundIsNull.inverse = BoundNotNull
+BoundNotNull.inverse = BoundIsNull
```
rdblue, line 417: This was my main goal because I think it is easy to validate that the behavior is correct. Reimplementing `__inverse__` etc. in every class seemed like it was going to cause trouble later on.

python/pyiceberg/expressions/base.py
```diff
@@ -342,107 +349,104 @@ def __str__(self) -> str:
...
-    def __invert__(self) -> IsNaN:
-        return IsNaN(self.term)
 
-    def _validate_literals(self):  # pylint: disable=W0238
-        if self.literals:
-            raise AttributeError("NotNaN is a unary predicate and takes no Literals.")
+class BoundIsNaN(BoundUnaryPredicate[T]):
+    def __new__(cls, term: BoundTerm[T]):
+        bound_type = term.ref().field.field_type
+        if bound_type == FloatType() or bound_type == DoubleType():
```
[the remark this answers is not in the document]
rdblue: Done.

python/pyiceberg/expressions/base.py
```diff
@@ -342,107 +349,104 @@ def __str__(self) -> str:
...
+        return AlwaysFalse()
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> BoundNotNaN[T]:
-        bound_ref = self.term.bind(schema, case_sensitive)
-        return BoundNotNaN(bound_ref)
 
+class BoundNotNaN(BoundUnaryPredicate[T]):
+    def __new__(cls, term: BoundTerm[T]):
+        bound_type = term.ref().field.field_type
+        if bound_type == FloatType() or bound_type == DoubleType():
```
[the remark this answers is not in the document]
rdblue: Done.

python/pyiceberg/expressions/base.py
rdblue: With the other refactoring, I think it was reasonable to put the `__invert__` implementations back in since they're one-line changes. Thanks, @Fokko!

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
-class BoundNotIn(BoundPredicate[T]):
-    def _validate_literals(self):  # pylint: disable=W0238
-        if not self.literals:
-            raise AttributeError("BoundNotIn must contain at least 1 literal.")
+class In(SetPredicate[T]):
+    as_bound = BoundIn
 
-    def __invert__(self) -> BoundIn[T]:
-        return BoundIn(self.term, *self.literals)
+    def __new__(cls, term: UnboundTerm[T], literals: tuple[Literal[T], ...]) -> BooleanExpression:
```
Fokko, line 480: How about accumulating the positional arguments like:
```suggestion
    def __new__(cls, term: UnboundTerm[T], *literals: Literal[T]) -> BooleanExpression:
```

This way we can write:
```python
def test_in_to_eq():
    assert base.In(
        base.Reference("x"),
        literal(34.56),
        literal(19.25)
    ) == base.Eq(base.Reference("x"), literal(34.56))
```
instead of:
```python
def test_in_to_eq():
    assert base.In(
        base.Reference("x"),
        (literal(34.56), literal(19.25))
    ) == base.Eq(base.Reference("x"), literal(34.56))
```
I think this makes the API a bit friendlier and less verbose.

You can easily expand the arguments when you want to supply a list:
```python
class NotIn(SetPredicate[T]):
    as_bound = BoundNotIn
...
    def __invert__(self) -> BooleanExpression:
        return In(self.term, *self.literals)
```

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
-            raise AttributeError("In must contain at least 1 literal.")
 
-    def __invert__(self) -> NotIn[T]:
-        return NotIn(self.term, *self.literals)
+class BoundNotIn(BoundSetPredicate[T]):
+    def __new__(cls, term: BoundTerm[T], literals: set[Literal[T]]) -> BooleanExpression:
+        count = len(literals)
+        if count == 0:
+            return AlwaysTrue()
+        if count == 1:
```
Fokko, line 468:
```suggestion
        elif count == 1:
```

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
 
-    def __invert__(self) -> In[T]:
-        return In(self.term, *self.literals)
+class NotIn(SetPredicate[T]):
+    as_bound = BoundNotIn
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> BoundNotIn[T]:
-        bound_ref = self.term.bind(schema, case_sensitive)
-        return BoundNotIn(bound_ref, *tuple(lit.to(bound_ref.field.field_type) for lit in self.literals))  # type: ignore
+    def __new__(cls, term: UnboundTerm[T], literals: tuple[Literal[T], ...]) -> BooleanExpression:
```
Fokko, line 496:
```suggestion
    def __new__(cls, term: UnboundTerm[T], literals: *Literal[T]) -> BooleanExpression:
```
See the comment at `class In(SetPredicate[T]):`

python/pyiceberg/expressions/base.py
```diff
@@ -19,58 +19,22 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from functools import reduce, singledispatch
-from typing import Generic, TypeVar
+from typing import (
+    ClassVar,
+    Generic,
+    Literal,
```
Fokko, line 25: I think we want to import our own literal from:
```python
from pyiceberg.expressions.literals import Literal
```
instead of the one from typing.

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
     """FALSE expression"""
 
     def __invert__(self) -> AlwaysTrue:
         return AlwaysTrue()
 
 
-class IsNull(UnboundPredicate[T]):
-    def __invert__(self) -> NotNull:
-        return NotNull(self.term)
+@dataclass(frozen=True)
```
Fokko, line 331: An abstract class should not be a dataclass as well, since it cannot be instantiated, it does not carry any data.

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
 
-class BoundIn(BoundPredicate[T]):
-    def _validate_literals(self):  # pylint: disable=W0238
-        if not self.literals:
-            raise AttributeError("BoundIn must contain at least 1 literal.")
 
-    def __invert__(self) -> BoundNotIn[T]:
-        return BoundNotIn(self.term, *self.literals)
+class BoundIn(BoundSetPredicate[T]):
+    def __new__(cls, term: BoundTerm[T], literals: set[Literal[T]]) -> BooleanExpression:
```
Fokko, line 450: We could also consider wrapping it into positional arguments:
```python
class BoundIn(BoundSetPredicate[T]):
    def __new__(cls, term: BoundTerm[T], *literals: Literal[T]) -> BooleanExpression:
        literals_set = set(literals)
        count = len(literals)
        if count == 0:
            return AlwaysFalse()
        if count == 1:
            return BoundEq(term, literals.pop())
        else:
            return super().__new__(cls)

    def __invert__(self) -> BooleanExpression:
        return BoundNotIn(self.term, self.literals)
```
This way we also don't have to care about mutating the set. See comment below 👍🏻 

This way we don't have to pass in a set:
```python
BoundIn[str](
    base.BoundReference(
        field=NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
        accessor=Accessor(position=0, inner=None),
    ),
    StringLiteral("foo"), StringLiteral("bar"), StringLiteral("baz"),
)
```
Instead of:
```python
BoundIn[str](
    base.BoundReference(
        field=NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
        accessor=Accessor(position=0, inner=None),
    ),
    {StringLiteral("foo"), StringLiteral("bar"), StringLiteral("baz")},
)
```

Fokko (commented):
Hey @rdblue looking good! It is a bit more verbose than the previous approach, but the mypy errors shows that it is able to infer types 👍🏻

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
+        count = len(literals)
+        if count == 0:
+            return AlwaysFalse()
+        if count == 1:
+            return BoundEq(term, literals.pop())
```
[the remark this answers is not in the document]
rdblue: Good catch.

python/pyiceberg/expressions/base.py
rdblue: I wanted to do this, but I didn't know how to make it work with `@dataclass`. I looked it up and saw that you have to implement `__init__`. But maybe we don't want to use `@dataclass` anyway? It seems to have strange mypy errors:

```
327: error: Only concrete class can be given where "Type[UnaryPredicate[T]]" is expected
```

python/pyiceberg/expressions/base.py
rdblue: Ah, I was wondering what the problem was. Thanks!

python/pyiceberg/expressions/base.py
```diff
@@ -342,151 +313,197 @@ def __str__(self) -> str:
...
+class NotNull(UnaryPredicate[T]):
+    as_bound = BoundNotNull
+
+    def __invert__(self) -> IsNull:
+        return IsNull(self.term)
+
 
+class BoundIsNaN(BoundUnaryPredicate[T]):
+    def __new__(cls, term: BoundTerm[T]):
+        bound_type = term.ref().field.field_type
```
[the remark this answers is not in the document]
rdblue: I'm actually not sure about this. There are mypy errors if you declare the return type from `__new__` to be correct. This is from `In.__new__` for example:

```
451: error: Incompatible return type for "__new__" (returns "BooleanExpression", but must return a subtype of "BoundIn[Any]")
```

python/pyiceberg/expressions/base.py
rdblue: If we make `__invert__` abstract, then we can't use a frozen `@dataclass` without duplicating a lot more.

The problem is, as you pointed out, `@dataclass` can't be abstract. If I move this to the child class, `IsNull`, then the constructor no longer supports `term` because it is defined here in the parent. Then we have to implement `__invert__` in each child class and define the `term` separately in each child class.

That basically removes the benefit of using a common base class. We can no longer rely on anything higher than the leaf class having a `term` or a set of `literals`.

I think I'd rather trade the mypy hints for `__invert__` than not be able to rely on inheritance for expressions. That is, if I have a `UnaryPredicate`, I want to be able to use its `term` instead of handling everything at the leaf level.

rdblue: I solved this by implementing `__invert__` and raising `NotImplementedError`. It's overridden in subclasses just like before, but now `@dataclass` works because these aren't abstract.

python/pyiceberg/expressions/base.py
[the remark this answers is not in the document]
Fokko: Yes, the `__new__` expects to return to the class it belongs to. I think we can ignore this one because I like the trick of returning a different type very much.

python/pyiceberg/expressions/base.py
```diff
@@ -342,235 +243,321 @@ def __str__(self) -> str:
...
+    def __invert__(self) -> BoundNotNull[T]:
+        return BoundNotNull(self.term)
 
-    def bind(self, schema: Schema, case_sensitive: bool) -> BoundNotNull[T]:
-        bound_ref = self.term.bind(schema, case_sensitive)
-        return BoundNotNull(bound_ref)
 
+@dataclass(frozen=True)
+class BoundNotNull(BoundUnaryPredicate[T]):
+    def __new__(cls, term: BoundTerm[T]):
```
samredai, line 310: pylint for the 3.8 CI seems to not like that the argument signature is different between this and the default `super().__new__`. You can change this to match the `*args, **kwargs` from super, but then you'd have to pull out the kwargs and it won't look as clean plus won't allow positional arguments.
```py
@dataclass(frozen=True)
class BoundNotNull(BoundUnaryPredicate[T]):
    def __new__(cls, *args, **kwargs):
        term: BoundTerm[T] = kwargs["term"]
        if term.ref().field.required:
            return AlwaysTrue()
        return super().__new__(cls)
```
This is just a linting error though and the assumptions made when using the super type hold well so I think it'd be better to just ignore it for these lines.
```py
@dataclass(frozen=True)
class BoundNotNull(BoundUnaryPredicate[T]):
    def __new__(cls, term: BoundTerm[T]):  # pylint: disable=W0611
```

python/pyiceberg/expressions/base.py
```diff
@@ -342,235 +243,321 @@ def __str__(self) -> str:
...
+    def bind(self, schema: Schema, case_sensitive: bool = True) -> BooleanExpression:
+        raise NotImplementedError
 
 
-class NotNull(UnboundPredicate[T]):
-    def __invert__(self) -> IsNull:
-        return IsNull(self.term)
+@dataclass(frozen=True)
+class UnaryPredicate(UnboundPredicate[T]):
+    def bind(self, schema: Schema, case_sensitive: bool = True) -> BooleanExpression:
```
samredai, line 283: Should this return type be `BoundUnaryPredicate`? Same question for `SetPredicate.bind(...)` and `LiteralPredicate.bind(...)` which can have return types of `BoundSetPredicate` and `BoundLiteralPredicate`.

python/tests/expressions/test_expressions_base.py
```diff
@@ -139,210 +217,138 @@ def test_strs(op, string):
...
+def test_bind_in(request):
+    schema = request.getfixturevalue("table_schema_simple")
+    bound = base.BoundIn(
+        base.BoundReference(schema.find_field(1), schema.accessor_for_field(1)), {literal("hello"), literal("world")}
+    )
+    assert base.In(base.Reference("foo"), (literal("hello"), literal("world"))).bind(schema) == bound
+
+
+def test_bind_dedup(request):
+    schema = request.getfixturevalue("table_schema_simple")
```
samredai, line 261: It's definitely not required but instead of using `request` here you can just use the fixture name and pytest will automatically grab it for you.
```py
def test_bind_dedup(table_schema_simple):
    bound = base.BoundIn(
        base.BoundReference(table_schema_simple.find_field(1), table_schema_simple.accessor_for_field(1)), {literal("hello"), literal("world")}
    )
    assert base.In(base.Reference("foo"), (literal("hello"), literal("world"), literal("world"))).bind(table_schema_simple) == bound
```

samredai (commented):
A few small comments but otherwise this LGTM!

python/pyiceberg/expressions/base.py
rdblue: Binding may produce a different type of expression. For example, if `IsNull` is bound to a required type, binding will return `AlwaysFalse`

python/pyiceberg/expressions/base.py
rdblue: Thanks! I suppressed the false positives.

danielcweeks (approved):

Issue #3891, opened 2022-01-13
`org.apache.iceberg.util.SnapshotUtil#snapshotIdAsOfTime` could make use of binary search if `org.apache.iceberg.Table#history` is guaranteed to be ordered


findinpath:
Currently the `SnapshotUtil` is going over all the history entries of a table (with O(n) complexity).
```
    Long snapshotId = null;
    for (HistoryEntry logEntry : table.history()) {
      if (logEntry.timestampMillis() <= timestampMillis) {
        snapshotId = logEntry.snapshotId();
      }
    }
```

In case that the `org.apache.iceberg.Table#history` would be guaranteed to be ordered, a binary search algorithm could be used instead (O(log n) complexity).

RussellSpitzer:
Unfortunately I do not believe we are guaranteed that History is ordered (WAP commits) or Complete (History has a max length) so I do not believe we can rely on it. The only true way to view the lineage is the recursive approach being used.

xndai:
> Unfortunately I do not believe we are guaranteed that History is ordered (WAP commits) or Complete (History has a max length) so I do not believe we can rely on it. The only true way to view the lineage is the recursive approach being used.

This is correct that we cannot assume the history is ordered. I don't know what you mean by "the recursive approach". The current implementation is still incorrect due to the same reason - history may not be strictly ordered.

xndai:
I open an issue here - https://github.com/apache/iceberg/issues/17358

PR #5364, opened 2022-07-27
Core, API: Support scanning from refs

amogh-jahagirdar:
This change adds support for scanning from refs, where the scan can either be a time travel from a tag or a time travel from the tip of a branch.


core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -322,6 +325,32 @@ public static long snapshotIdAsOfTime(Table table, long timestampMillis) {
...
+  public static long snapshotIdAsOfTime(Table table, String branch, long timestampMillis) {
+    SnapshotRef ref = table.refs().get(branch);
+    Preconditions.checkArgument(ref != null, "Branch %s does not exist", branch);
+    Preconditions.checkArgument(ref.isBranch(), "Ref %s is a tag, not a branch", branch);
+    Long snapshotId = null;
+    for (Snapshot snapshot : ancestorsOf(ref.snapshotId(), table::snapshot)) {
+      if (snapshot.timestampMillis() <= timestampMillis) {
+        snapshotId = snapshot.snapshotId();
+      }
+    }
```
amogh-jahagirdar, line 347: I don't think this is right. Here we're iterating over the ancestors rather than the log entries like we do for time travel on the main branch. 

To be consistent with previous time travel behavior I think we need to leverage history logs but I think that means in the metadata we should actually maintain a history log for each branch (requiring a spec change). @rdblue @jackye1995 

Also we're not guaranteed timestamp ordering due to WAP commits, so I think the linear search is required as opposed to using a binary search. https://github.com/apache/iceberg/issues/3891

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -53,6 +53,30 @@ public interface TableScan extends Scan<TableScan, FileScanTask, CombinedScanTas
...
+  /**
+   * Create a new {@link TableScan} from this scan's configuration that will use the most recent snapshot
+   * as of the given time in milliseconds.
+   *
+   * @param branch branch name
+   * @param timestampMillis  a timestamp in milliseconds.
+   * @return a new scan based on this with the given branch and timestamp
+   * @throws IllegalArgumentException if the branch or snapshot cannot be found
+   */
+  default TableScan useBranch(String branch, long timestampMillis) {
```
jackye1995, line 76: what about `asOfTimeAtBranch`?

amogh-jahagirdar: How about useBranchAsOfTime?

core/src/main/java/org/apache/iceberg/DataTableScan.java
```diff
@@ -61,6 +61,24 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
+    SnapshotRef ref = tableOps().current().ref(name);
+    Preconditions.checkArgument(ref != null, "Branch %s does not exist", name);
+    Preconditions.checkArgument(ref.isBranch(), "Ref %s is a tag, not a branch", name);
+    long branchTip = tableOps().current().ref(name).snapshotId();
+    return useSnapshot(branchTip);
+  }
+
+  @Override
+  public TableScan useBranch(String name, long timestampMillis) {
+    Preconditions.checkArgument(snapshotId() == null,
```
amogh-jahagirdar, line 77: This check isn't needed, already done in useSnapshot

core/src/main/java/org/apache/iceberg/DataTableScan.java
```diff
@@ -61,6 +61,24 @@ public TableScan appendsAfter(long fromSnapshotId) {
     return appendsBetween(fromSnapshotId, currentSnapshot.snapshotId());
   }
 
+  @Override
+  public TableScan useBranch(String name) {
+    Preconditions.checkArgument(snapshotId() == null,
+            "Cannot override snapshot, already set to id=%s", snapshotId());
```
amogh-jahagirdar, line 67: Same as below

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -322,6 +325,32 @@ public static long snapshotIdAsOfTime(Table table, long timestampMillis) {
     return snapshotId;
   }
 
+  /**
+   * Returns the ID of the most recent snapshot in the branch
```
amogh-jahagirdar, line 329: Need to update comment

api/src/main/java/org/apache/iceberg/TableScan.java
amogh-jahagirdar: Updated to useBranchAsOfTime. I think it's good ot have the use just because we already have the useSnapshot method.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -322,6 +323,33 @@ public static long snapshotIdAsOfTime(Table table, long timestampMillis) {
...
+  public static long snapshotIdAsOfTime(Table table, String branch, long timestampMillis) {
+    SnapshotRef ref = table.refs().get(branch);
+    Preconditions.checkArgument(ref != null, "Branch %s does not exist", branch);
+    Preconditions.checkArgument(ref.isBranch(), "Ref %s is a tag, not a branch", branch);
+    Long snapshotId = null;
+    for (Snapshot snapshot : ancestorsOf(ref.snapshotId(), table::snapshot)) {
+      if (snapshot.timestampMillis() <= timestampMillis) {
+        snapshotId = snapshot.snapshotId();
+      }
+    }
```
amogh-jahagirdar, line 346: This doesn't find the most recent snapshot correctly; I should compute the difference in timestamps in the case the timestamp is <= and keep track of the minimum one.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -322,6 +323,39 @@ public static long snapshotIdAsOfTime(Table table, long timestampMillis) {
...
+    Preconditions.checkArgument(ref.isBranch(), "Ref %s is a tag, not a branch", branch);
+    Long snapshotId = null;
+    long minimumTimeDifference = Long.MAX_VALUE;
+    for (Snapshot snapshot : ancestorsOf(ref.snapshotId(), table::snapshot)) {
+      if (snapshot.timestampMillis() <= timestampMillis) {
+        if (timestampMillis - snapshot.timestampMillis() <= minimumTimeDifference) {
+          minimumTimeDifference = timestampMillis - snapshot.timestampMillis();
+          snapshotId = snapshot.snapshotId();
+        }
+      }
```
amogh-jahagirdar, line 349: This does not seem right in the presence of WAP. If we have a staged commit on a branch then we should not include that in time travel; but with traversing through ancestorsOf we would get that. The reason why time travel relies on the snapshot log is because the snapshot log is the source of truth for how the table (main) state evolved over time. Looking through ancestors and comparing timestamps would count staged commits which we don't want (we should only count snapshots which are part of that branch's state).


This may involve having to maintain some kind of extra metadata like instead of a single snapshot log for the main table state, there is a map<string, list<historyEntry>> where the key is the ref and the value is the list of logs. When a snapshot is produced on a branch, and is not staged then we would update this metadata.

amogh-jahagirdar: I think I was overthinking this. Lets take a simple case:

staged snapshots should not be considered when doing time travel on a branch. 

```

    S4 (staged at ts-400)
  /  
S1 (committed at ts-100) - S2 (committed at ts-200) - S3 (committed at ts-300)
```

Now we do the following: createBranch("newBranch", S4)

The graph looks like: 

```
    S4 (staged at ts-400, tip of newBranch)
  /  
S1 (committed at ts-100) - S2 (committed at ts-200) - S3 (committed at ts-300)
```
Originally I was thinking that snapshotIdAsOfTime(table, newBranch, 500) should return S1 because S4 is staged. But snapshot 4 is part of the branch state! So we should consider that when doing time travel.

Now, let's take this case where 4 is actually committed on newBranch.

```
    S4 (committed at ts-400, tip of newBranch)
  /  
S1 (committed at ts-100) - S2 (committed at ts-200) - S3 (committed at ts-300)
```




Now we want to stage a snapshot 5 onto newBranch.

```
    S4 (committed at ts-400, tip of newBranch) - S5 (staged at 500)
  /  
S1 (committed at ts-100) - S2 (committed at ts-200) - S3 (committed at ts-300)
```

When we produce the staged snapshot onto a branch we will make sure the ancestor is set correctly for S5,
    but the ref pointer should not be advanced to the staged commit. In this case snapshotIdAsOfTime(table, newBranch, 500) will return  S4 and not S5. So in this case we are good; we'll need to
    ensure this in our snapshot producer changes.

api/src/main/java/org/apache/iceberg/TableScan.java
hililiwei: +1 for useBranchAsOfTime

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -322,6 +323,39 @@ public static long snapshotIdAsOfTime(Table table, long timestampMillis) {
...
+    }
+
+    Preconditions.checkArgument(snapshotId != null,
+        "Cannot find a snapshot older than %s on branch %s",
+        DateTimeUtil.formatTimestampMillis(timestampMillis), branch);
+
+    return snapshotId;
+  }
+
```
namrathamyske, line 358: Can we make reuse `snapshotIdAsOfTime` here? We will also need to make necessary changes for this function to accept a branch.

namrathamyske (changes requested):

core/src/main/java/org/apache/iceberg/DataTableScan.java
```diff
@@ -61,6 +61,19 @@ public TableScan appendsAfter(long fromSnapshotId) {
     return appendsBetween(fromSnapshotId, currentSnapshot.snapshotId());
   }
 
+  @Override
+  public TableScan useBranch(String name) {
```
namrathamyske, line 65: @amogh-jahagirdar do theseneed to be implemented in BaseTableScan? Just following acc to `asOfTime`

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
amogh-jahagirdar: I think we need a separate method which accepts the branch parameter (which is what is done here). This is because the existing method is a public utility, so we should not break compatibility there by requiring a branch parameter.

core/src/main/java/org/apache/iceberg/DataTableScan.java
amogh-jahagirdar: Yes good point. we should have a default implementation (well it's an abstract class, so just an implementation) which throws Unsupported. Right now DataTableScan would be the only table scan which supports branching and tagging.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
jackye1995: We can let `snapshotIdAsOfTime(Table table, long timestampMillis)` call the new method right?

api/src/main/java/org/apache/iceberg/TableScan.java
jackye1995: +1

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
amogh-jahagirdar: @jackye1995 That's true, The existing snapshotIdAsOfTime can call the new method with the main branch.

core/src/main/java/org/apache/iceberg/DataTableScan.java
amogh-jahagirdar: I finally got a chance to come back to this PR,  I actually decided on the opposite approach. There's fewer table scans where we want to fail, so I think it makes sense to have the implementation in the parent class and then in the child classes which don't support (like all metadata table scan, or incremental scan) we throw an unsupported exception.

.palantir/revapi.yml
```diff
@@ -49,20 +58,25 @@ acceptedBreaks:
...
+      justification: "Interface is backward compatible, very unlikely anyone implements\
+        \ this Result bean interface"
     - code: "java.method.addedToInterface"
       new: "method long org.apache.iceberg.actions.DeleteReachableFiles.Result::deletedPositionDeleteFilesCount()"
-      justification: "Interface is backward compatible, very unlikely anyone implements this Result bean interface"
+      justification: "Interface is backward compatible, very unlikely anyone implements\
+        \ this Result bean interface"
+    - code: "java.method.addedToInterface"
+      new: "method long org.apache.iceberg.actions.ExpireSnapshots.Result::deletedEqualityDeleteFilesCount()"
+      justification: "Interface is backward compatible, very unlikely anyone implements\
```
amogh-jahagirdar, line 74: I'll need to undo this

amogh-jahagirdar:
@namrathamyske @rdblue @jackye1995 I've updated the PR based on our discussion on this [thread](https://github.com/apache/iceberg/pull/5150/files#r939702061) . Let me know what you think!

jackye1995:
Overall looks good to me, can we add some tests?

amogh-jahagirdar:
@jackye1995  Definitely, I'm still working on adding tests, wanted to get a discussion on the code in parallel.

amogh-jahagirdar:
Java CI build check and revapi check both fail with:

* What went wrong:
Plugin [id: 'nebula.dependency-recommender', version: '11.0.0'] was not found in any of the following sources:

https://plugins.gradle.org/ is down for me (get a 1016 error from cloudflare) so most likely can't resolve any plugins. I'll retrigger the build when it becomes available again.

core/src/main/java/org/apache/iceberg/TableScanContext.java
```diff
@@ -60,6 +61,7 @@ final class TableScanContext {
     this.planExecutor = null;
     this.fromSnapshotInclusive = false;
     this.scanReporter = new LoggingScanReporter();
+    this.ref = null;
```
amogh-jahagirdar, line 64: Technically with the current scan support we don't even need to store ref in the context. at the end of the day with the ref scanning we're scanning from a single snapshot, so we could just resolve the snapshot and set it as before. I'll think some more if this is really needed.

core/src/test/java/org/apache/iceberg/TestDataTableScan.java
```diff
@@ -18,6 +18,17 @@
...
+import java.io.IOException;
+import java.util.List;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.assertj.core.util.Lists;
+import org.junit.Assert;
+import org.junit.Test;
+
 public class TestDataTableScan extends ScanTestBase<TableScan, FileScanTask, CombinedScanTask> {
```
amogh-jahagirdar, line 32: Need to add more tests for the unhappy paths

amogh-jahagirdar: Added

amogh-jahagirdar:
Updated with more tests @namrathamyske @jackye1995  @hililiwei @rdblue let me know what you think!

rdblue:
I should have time to look today. Thanks, @amogh-jahagirdar!

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -86,20 +90,56 @@ public TableScan appendsAfter(long fromSnapshotId) {
   public TableScan useSnapshot(long scanSnapshotId) {
     Preconditions.checkArgument(
         snapshotId() == null, "Cannot override snapshot, already set to id=%s", snapshotId());
+    Preconditions.checkArgument(
+        ref() == null, "Cannot override snapshot, already set to id=%s", ref());
```
rdblue, line 94: Looks like this should update `id=` to `ref=`.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -86,20 +90,56 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
     Preconditions.checkArgument(
         tableOps().current().snapshot(scanSnapshotId) != null,
         "Cannot find snapshot with ID %s",
         scanSnapshotId);
     return newRefinedScan(
-        tableOps(), table(), tableSchema(), context().useSnapshotId(scanSnapshotId));
+        tableOps(),
+        table(),
+        tableSchema(),
+        context().useRef(SnapshotRef.MAIN_BRANCH).useSnapshotId(scanSnapshotId));
```
rdblue, line 103: Why does this set `useRef`? The precondition above checks that `ref` is `null`, so it seems like it should never be set if the snapshot ID is set.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -86,20 +90,56 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
+  }
+
+  @Override
+  public TableScan useRef(String name) {
+    Preconditions.checkArgument(
+        snapshotId() == null, "Cannot override ref, already set to id=%s", snapshotId());
```
rdblue, line 109: `already set snapshot id=`?

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -86,20 +90,56 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
+    Preconditions.checkArgument(ref() == null, "Cannot override ref, already set to id=%s", ref());
+    SnapshotRef ref = tableOps().current().ref(name);
+    Preconditions.checkArgument(ref != null, "Cannot find ref %s", name);
+    long scanSnapshotId = ref.snapshotId();
+    return newRefinedScan(
+        tableOps(), table(), tableSchema(), context().useRef(name).useSnapshotId(scanSnapshotId));
```
rdblue, line 115: Why set `ref` and the snapshot ID? It seems like this is going to cause conflicts with `asOfTime`

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -86,20 +90,56 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
+
+    if (ref() == null) {
+      Preconditions.checkArgument(
+          table().currentSnapshot() != null,
+          "Cannot find a snapshot older than %s in table",
+          DateTimeUtil.formatTimestampMillis(timestampMillis));
+    }
 
-    return useSnapshot(SnapshotUtil.snapshotIdAsOfTime(table(), timestampMillis));
+    String ref = ref() == null ? SnapshotRef.MAIN_BRANCH : ref();
```
rdblue, line 132: I think there's a problem here: `asOfTime` selects a snapshot from table history, which we don't keep for refs other than `main`. That means we can't really have consistent behavior because the only way to implement this for branches would be to use ancestors and not the actual branch history. Because of that issue, I don't think that we should allow combining `useRef` and `asOfTime` after all.

core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java
```diff
@@ -322,29 +322,60 @@ public static Snapshot snapshotAfter(Table table, long snapshotId) {
...
    * @param table a {@link Table}
+   * @param branch a {@link String}
    * @param timestampMillis the timestamp in millis since the Unix epoch
    * @return the snapshot ID
-   * @throws IllegalArgumentException when no snapshot is found in the table older than the
-   *     timestamp
+   * @throws IllegalArgumentException when no snapshot is found in the table, on the given branch
+   *     older than the timestamp
    */
-  public static long snapshotIdAsOfTime(Table table, long timestampMillis) {
```
rdblue, line 333: We cannot change the behavior of this method to use ancestors instead of table history.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -86,20 +90,56 @@ public TableScan appendsAfter(long fromSnapshotId) {
...
     return newRefinedScan(
-        tableOps(), table(), tableSchema(), context().useSnapshotId(scanSnapshotId));
+        tableOps(),
+        table(),
+        tableSchema(),
+        context().useRef(SnapshotRef.MAIN_BRANCH).useSnapshotId(scanSnapshotId));
+  }
+
+  @Override
+  public TableScan useRef(String name) {
```
rdblue, line 107: Because we can't combine this we `asOfTime` (see my other comment), I think the implementation can be simpler. The context doesn't need to know the `ref`. This can act just like `asOfTime` and set a specific snapshot that his the head of the branch or tag.

core/src/main/java/org/apache/iceberg/TableScanContext.java
```diff
@@ -75,7 +77,8 @@ private TableScanContext(
       Long toSnapshotId,
       ExecutorService planExecutor,
       boolean fromSnapshotInclusive,
-      ScanReporter scanReporter) {
+      ScanReporter scanReporter,
+      String ref) {
```
rdblue, line 81: I don't think it is necessary to add this, since the ref can really only select a snapshot. We can use the existing `snapshot` behavior.

rdblue (changes requested):
@amogh-jahagirdar, I was wrong about being able to support `useRef` and `asOfTime` because `asOfTime` uses table history, not snapshot ancestry. I think we can simplify this PR quite a bit because changes to `SnapshotUtil` and `TableScanContext` should no longer be needed. And testing will be simpler as well since we only have to test that `useRef`, `useSnapshot`, and `asOfTime` are incompatible with one another.

FYI @jackye1995.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
amogh-jahagirdar: I think we can actually remove this validation; if a ref is set we always set a snapshot as well. This snapshot can either be a tag, tip of a branch, or the snapshot from a time travel on the branch. 

The snapshot check on the line before is sufficient validation; although the error message won't indicate the ref is set. I can include that in the error message if that's desired?

core/src/main/java/org/apache/iceberg/BaseTableScan.java
amogh-jahagirdar: For the scan from ref cases, we set both a ref and snapshot so setting it here to be consistent with that. It maintains the invariant that every scan context will have a ref and a snapshot.

 As for the null ref check, I think we can remove that because if useRef is called, we'll be setting both, and then the existing snapshot check will fail.

amogh-jahagirdar:
@rdblue I considered this when I changed the implementation to use snapshot timestamps instead of History entries. Just for my understanding, what are the cases where the timestamp on the history entry as a result of a new snapshot, is different than the timestamp on the snapshot itself? It seems like they would always be the same.

core/src/main/java/org/apache/iceberg/TableScanContext.java
amogh-jahagirdar: It was required before because if we do useRef("branch").asOfTime(someTime) the context acts as a placeholder to store the information for branches, so that in the asOfTime we can choose the right ref. If we don't do time travel with branches then yes we just need to set a snapshot.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
amogh-jahagirdar: Updated

core/src/main/java/org/apache/iceberg/BaseTableScan.java
amogh-jahagirdar: I was following the existing snapshot id already set error messaging, I'm definitely open to change it but then I think we should change it in the other places, let me know what you think!

amogh-jahagirdar:
Discussed offline with @rdblue , I got an understanding of how history entries are different than the snapshot order, posting the information here so folks referring to the PR have context:

"History is the state of main at any given time. That can be changed, much like master in git can be changed.
Ancestors can't be changed."

The current main can change when rollback is performed. The snapshot which gets resolved for a given time T should remain constant regardless of what is the current main ancestor lineage. History entries capture all the changes on the main table state and are what enable this ability to time travel to a snapshot which are outside the current table ancestors. It also doesn't make sense to treat main as a separate case, and have a different semantic for non-main branches when it comes to time travel. 

Maybe we could maintain separate history metadata for branches, but most likely it would need to be opt-in by users and would need to not add too much metadata to manage.

So for now, it makes sense just to do time travel to a branch tip or time travel to a given tag.

core/src/main/java/org/apache/iceberg/BaseTableScan.java
amogh-jahagirdar: https://github.com/apache/iceberg/pull/5364#issuecomment-1227902420 clarified here

namrathamyske (approved):

core/src/main/java/org/apache/iceberg/BaseTableScan.java
jackye1995: I think `set snapshot id=` is more clear, otherwise people might misunderstand and think there is a ref id.

amogh-jahagirdar: Good point, I agree. I've updated the error messages in all the places.

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -39,13 +39,25 @@ public interface TableScan extends Scan<TableScan, FileScanTask, CombinedScanTas
    */
   TableScan useSnapshot(long snapshotId);
 
+  /**
+   * Create a new {@link TableScan} from this scan's configuration that will use the given
+   * reference.
```
jackye1995, line 44: nit: use the snapshot ID of the given reference?

api/src/main/java/org/apache/iceberg/TableScan.java
```diff
@@ -39,13 +39,25 @@ public interface TableScan extends Scan<TableScan, FileScanTask, CombinedScanTas
...
   /**
    * Create a new {@link TableScan} from this scan's configuration that will use the most recent
-   * snapshot as of the given time in milliseconds.
+   * snapshot as of the given time in milliseconds on the branch in the scan or main if no branch is
+   * set.
    *
    * @param timestampMillis a timestamp in milliseconds.
    * @return a new scan based on this with the current snapshot at the given time
-   * @throws IllegalArgumentException if the snapshot cannot be found
+   * @throws IllegalArgumentException if the snapshot cannot be found or time travel is attempted on
```
jackye1995, line 59: I think the doc change here is no longer needed?

core/src/main/java/org/apache/iceberg/BaseTableScan.java
```diff
@@ -85,7 +85,7 @@ public TableScan appendsAfter(long fromSnapshotId) {
   @Override
   public TableScan useSnapshot(long scanSnapshotId) {
     Preconditions.checkArgument(
-        snapshotId() == null, "Cannot override snapshot, already set to id=%s", snapshotId());
+        snapshotId() == null, "Cannot override snapshot, already set snapshot id=%s", snapshotId());
```
jackye1995, line 88: nit: "set to snapshot id=%s"?

jackye1995 (approved):
Overall looks good to me. However I am a bit concerned about the fact that we cannot do time travel in branches, I see that as a pretty important feature for many use cases. I agree with the argument of the difference of tracing ancestor vs history entry, but in that case it seems like we need to update the spec to accommodate that change and record history entries in a way that allows time travel by branch. 

@amogh-jahagirdar I think after this PR we should raise another PR against spec to discuss what is the best way to solve the time travel by branch use case.

core/src/main/java/org/apache/iceberg/IncrementalDataTableScan.java
```diff
@@ -46,6 +46,14 @@ public TableScan asOfTime(long timestampMillis) {
             timestampMillis, context().fromSnapshotId(), context().toSnapshotId()));
   }
 
+  @Override
+  public TableScan useRef(String ref) {
+    throw new UnsupportedOperationException(
```
rdblue, line 51: This seems fine for now, but we may have a case where we want incremental to be able to specify a branch. It doesn't matter currently because the snapshot IDs are always explicit.

rdblue (approved):

rdblue:
Looks great. Thanks, @amogh-jahagirdar! I'll merge when tests are passing. I had to fix a conflict in revapi.

amogh-jahagirdar:
Thanks @rdblue @yzhaoqin @namrathamyske @hililiwei for the review!

PR #5366, opened 2022-07-27
AWS: Call abortUpload only once when any of the completable future fails

singhpk234:
### About the Change :
Presently whenever a failure is observed in any of the completableFutures in `uploadParts()`, it causes further completable futures fail with `FileNotFoundException`, when it tries to create a request for it using `RequestBody.fromFile(file)`, as all the stagings files deleted as part of `abortUpload` which is called on failure of first completableFuture.

https://github.com/apache/iceberg/blob/7a92bf5e31a995c61dc598d2a36f598d5c120756/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java#L306

I think this is what was also highlighted in @stevenzwu 's observation here (https://github.com/apache/iceberg/issues/4168#issuecomment-1123025380)

This change attempts : 
* To avoid unnecessary deletes being called for all staging files on every `completableFutures` which fail due to fileNotFoundException.
* Cancels the remaining futures in which would not be required as the abort is called, reduces number of calls to abort multi-part upload as well.

### Testing Done

Existing UT's +  a UT to see abort is called only once rather than atleast once.

cc @rdblue @jackye1995 @RussellSpitzer @stevenzwu @amogh-jahagirdar @rajarshisarkar

### DIFF
--- a/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
+++ b/aws/src/main/java/org/apache/iceberg/aws/s3/S3OutputStream.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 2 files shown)


saswata-dutta (approved):

singhpk234:
This PR can also fix the flakyness of TestS3OutputStream#testAbortAfterFailedPartUpload in recent runs of master.

RussellSpitzer (approved):

RussellSpitzer:
Thanks @singhpk234  and thanks @saswata-dutta for review!

singhpk234:
Thanks @RussellSpitzer !

nastra (commented):
LGTM, thanks for fixing this

